PR #384, opened 2012-09-25
A try to implement a makeUrl override 

Evengard:
see http://forums.modx.com/thread/79430/makeurl-override-wanted


OptimusCrime:
Perhaps time to pull the plug on this one? A better approach was never provided. Suggest killing it.

rtripault:
pulling...

PR #620, opened 2013-09-22
[#10222] Move modResource.cleanAlias to modX.cleanAlias

Mark-H:
See http://tracker.modx.com/issues/10222

> The cleanAlias function is super useful in 3PCs that deal with custom urls, as it does all the translit magic and makes sure the urls are safe to use. Unfortunately you need a modResource object (which for example is not available in the OnPageNotFound event) to use it, which is sort of hacky and doesn't feel like a general-purpose function.
> I'd like to propose moving the cleanAlias function to the main modX class to encourage usage in 3PCs. To maintain backwards compatibility, the modResource.cleanAlias function could be deprecated and simply pass along the properties to $this->xpdo->cleanAlias.

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -470,120 +470,15 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
     /**
      * Transforms a string to form a valid URL representation.
      *
+     * @deprecated
+     * @see modX.cleanAlias
+     *
      * @param string $alias A string to transform into a valid URL representation.
      * @param array $options Options to append to or override configuration settings.
      * @return string The transformed string.
      */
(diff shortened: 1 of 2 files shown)


soulcreate:
+1
I want use modResource::cleanAlias() but it not work

OptimusCrime:
@soulcreate The changes mentioned in this issue was never done. Because of that you'll need to call cleanAlias on an instance of modResource instead.

soulcreate:
@OptimusCrime I find best solution:
```
modResource::filterPathSegment($modx, $alias);
```
https://modx.pro/help/5862/#comment-52751

Mark-H:
There's $modx->filterPathSegment since https://github.com/modxcms/revolution/pull/634

Similar results, but slightly different implementation than proposed in this issue :)

PR #11125, opened 2014-02-09
bugfix installed package list in its detail's page

goldsky:
bugfix installed package list in its detail's page when the package's name has spaces.
http://forums.modx.com/thread/88734/package-version-check#dis-post-489531

### DIFF
--- a/core/model/modx/processors/workspace/packages/version/getlist.class.php
+++ b/core/model/modx/processors/workspace/packages/version/getlist.class.php
@@ -31,7 +31,7 @@ class modPackageVersionGetListProcessor extends modObjectGetListProcessor {
         /* get packages */
         $criteria = array(
             'workspace' => $this->getProperty('workspace',1),
-            'package_name' => $signatureArray[0],
+            'package_name' => urldecode($this->getProperty('package_name',$signatureArray[0])),
         );
         $limit = $this->getProperty('limit');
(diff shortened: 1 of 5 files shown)


goldsky:
I want to raise this issue again.
This commit is not dealing with the remote provider, but to check the package in local.

**The processor is using [incorrect criteria ](https://github.com/modxcms/revolution/blob/master/core/model/modx/processors/workspace/packages/version/getlist.class.php#L34)**.
`'package_name' => $signatureArray[0],`

[The AJAX is not providing `package_name`](https://github.com/modxcms/revolution/blob/master/manager/assets/modext/workspace/package.grid.js#L350).

* `$signatureArray[0]` is coming from [`signature` field](https://github.com/modxcms/revolution/blob/master/core/model/schema/modx.transport.mysql.schema.xml#L47)
* `package_name` has [its own field](https://github.com/modxcms/revolution/blob/master/core/model/schema/modx.transport.mysql.schema.xml#L58)

So,
> Sorry this problem is in extras provider and it can't be fixed by using package name to search.

Please, test this again.

Again, this is not about the remote provider regardless about its naming problem, but rather the bug of getting the package on local system.

Mark-H:
@rico I'm unable of reopening the pull request, maybe the branch has since been removed on your fork? Can you create a new one?

More details would be good, maybe like steps to reproduce because I'm also unsure about what's actually broken.

goldsky:
@Mark-H https://github.com/modxcms/revolution/pull/12551

PR #11239, opened 2014-03-26
Allows comma-separated list of Form Customization constraints

OCDCoder:


Mark-H:
Hi @OCDCoder! Still here? ;)

I'd like to merge this into 2.4 as it seems like a good improvement to form customisation. I don't see a CLA (Contributors License Agreement) for you in the system though. Could you let me know your MODX username, and if you haven't yet, please sign the CLA at modx.com/cla? With that in place, I can hit the merge button and make sure this is in the upcoming 2.4 release :)

OCDCoder:
Yep, still here! And I'm SO glad to hear this is finally being added! I'm up to around 40 sites I have to patch now. :-)

I'm sure I submitted one ages ago before I did this, but just in case, I just submitted it again. I remember having a conversation with someone about which one I should fill out, since I'm doing this solo rather than as part of my company, although I was using my company domain in my email address. But I do now finally have ocdcoder.org if you want me to change the email address. (Although that would be submission x 3 since I didn't think of it until just now.)

In any case, I'm OCDCoder both places. Let me know if you need anything else!

Cherie

OCDCoder:
And of course right after I wrote this, I found the original. March 4, 2014.

Mark-H:
Merged into develop for inclusion in 2.4, thanks! 6c07d38f4ca3e5728e6fe94e1bf9ebed8b3bcd40

PR #11279, opened 2014-04-14
Enhancement: query string for postfields in modRestCurlClient

exside:
Added possibility for the native REST cURL client to send postfields as query string for APIs (like Google Closure Compiler) that cannot handle a normal passed params array. Now when the contentType in the options array is set to string, the params/postfields array is transformed to param1=value&param2=http://www.url.tld/&param3=somethingelse. The array_map() stuff has to be done because http_build_query is not suitable here as it encodes special characters like slashes in a url etc., which are then also not understood by such APIs...so it has unfortunately to be done like that:

```
$this->modx->getService('getfile','rest.modRestCurlClient');

$url = 'http://www.apiurl.com/';
$path = null;
$method = 'POST';
$params = array(
    'param1' => 'value',
    'param2' => 'http://www.url.tld/',
    'param3' => 'somethingelse'
);
$options = array('contentType' => 'string');
$response = $this->modx->getfile->request($url, $path, $method, $params, $options);
```

### DIFF
--- a/core/model/modx/rest/modrestcurlclient.class.php
+++ b/core/model/modx/rest/modrestcurlclient.class.php
@@ -86,7 +86,7 @@ class modRestCurlClient extends modRestClient {
                         break;
                     case 'string':
                         curl_setopt($ch,CURLOPT_HTTPHEADER, array('Content-Type: application/x-www-form-urlencoded'));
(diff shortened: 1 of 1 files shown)


Mark-H:
Merged, thank you Luk!

PR #11296, opened 2014-04-20
Enhancement: Allow request() method to return curlinfo in modRestCurlClient

exside:
Very small change, but quite helpful when working with the modRestCurlClient. If the param "returninfo" is set to true via the $options array like

```
$this->modx->getService('getfile','rest.modRestCurlClient');
$response = $this->modx->getfile->request($options['url'], $options['path'], $options['method'], $options['params'], $options['options'] = array('returninfo' => 1));
```

the response contains all information about the request from curl_getinfo()

### DIFF
--- a/core/model/modx/rest/modrestcurlclient.class.php
+++ b/core/model/modx/rest/modrestcurlclient.class.php
@@ -45,6 +45,11 @@ class modRestCurlClient extends modRestClient {
 
         /* execute request */
         $result = trim(curl_exec($ch));
+
+        /* if returninfo option is set replace the response with curlinfo */
+        if ($options['returninfo']) {
+            $result = curl_getinfo($ch);
+        }
         
         /* make sure to close connection */
         curl_close($ch);


Mark-H:
Hmm I like the idea here,  but wondering if maybe it would be better if there were a way to get both the result, and the curl info?

Thinking of use cases where you might want to check the curl info for the HTTP response header to see if it went through alright, and then checking the result. With this particular implementation, you can only get one of them. 

Maybe a variable passed by reference that gets filled with the info (kinda like $matches with regex)? Or a new method on the curl client getLastInfo? Or...?

Mark-H:
Thoughts on that last comment @exside?

exside:
@Mark-H I absolutely agree, it should return both, the response and the info, but how to make that possible without killing backwards-compatibility? If that wasn't an issue we could just return an array or so, e.g. the new $result would be something like:

```php
$result = array(
    'response' => 'whatever data came back from the request',
    'info' => curl_getinfo($ch)
);
```

not sure if I get how to do that with "... a variable passed by reference" as you  mentioned and also a separate method is not clear to me from an implementation standpoint as the curl_getinfo() method needs the curl-handle passed after the actual work is done, so how could we get that info out of the context of the actual request?

exside:
Any help on this one, I'm stuck =)...

theboxer:
@exside you could change the signature of the method and add a new parameter to the end, something like `&$info = array()` and then you won't have to check for the option passed and just assign `$info = curl_getinfo($ch);`.

Mark-H:
John explained the variable by reference. For the approach of a separate method, you'd do something like `$this->lastInfo = curl_info($ch);` in the request method, and add a new `getLastInfo()` method to the client class that returns the contents of `$this->lastInfo`.

Mark-H:
Closing as this has been pending improvements/feedback for almost 6 months. Feel free to reopen/open a new PR.

PR #11352, opened 2014-05-09
Feature: Add archive extraction functionality to filetree-menu

exside:
Based on the forum discussion here: http://forums.modx.com/thread/84971/unzip-function-for-file-manager#dis-post-497458

This adds a filetree context menu item "Extract file" for .zip files (~~zip only for now~~), the PR includes also an additional permission and de + en lexicon entries plus all necessary modifications in the core to make it work.

Looks like that:

bildschirmfoto 2014-05-09 um 19 35 51

Not tested as fresh installation (so maybe I missed something in the _build stuff?), working on 2.2.14pl-advanced (core outside webroot) on a LAMP environment (would be nice if somebody could test this in a Windows environment as I don't have one available)

Credits also go to Susan Ottwell for providing the right idea and approach to get this running!


Mark-H:
Marked as awaiting feedback because dependant PRs have not been merged.

exside:
I think I'll have to redo this when I have a bit air to breathe...one of my typically way too big PRs...I think I'm gonna try to avoid the xpdo modifications and just use what's already there (e.g. ZIP only) to implement this functionality. guess for most people zip only extraction would be more than enough and is far better than not having any extraction possibilities...but gonna start over with a new PR for this.

TheBoxer:
Closing and waiting for the new one :)

PR #11371, opened 2014-05-14
Enhancement: Add download() method to modFileHandler

exside:
This is a updated and properly branched version of this PR: https://github.com/modxcms/revolution/pull/680

As of @MarkH s comment from that PR

> Is this perhaps more appropriate/useful when added to media sources, so it's not limited to files? Speaking of which, I think it's already possible to download files from the Files tab, so this has to be implemented somewhere already..?

I did a bit of research and found a lot of repeating code which mostly does, what the new method download() does, e.g. setting headers etc. As already commented in the old PR: modMediaSource btw. modFileMediaSource already make use of the modFileHandler class, so it's easy to access this through the mediaSource as can be seen in the refactored download.class.php processor at core/processors/browser/ which handles the downloading of files from the file tree.

Together with the new method I also improved (hopefully) the two existing methods getSize() and getContents() to be able to return something, before a file really exists in the filesystem. This is helpful when  offering a download from data that is created on the fly but doesn't have to be saved somewhere in the filesystem, like a csv/json export etc. Like this we can set the file content via setContent($content) but never have to really save/create a physical file and can do:

```

$fileobj = $modx->file->make('path/to/the/file.ext'); // in this case this just doesn't matter, this would be the path where the file was created if we do $fileobj->save()/$fileobj->create() though

$fileobj->setContent($content);
$fileobj->download();
// or
$fileobj->download(array('mimetype' => 'image/jpeg'));
// or
$fileobj->download(array('filename' => 'someotherfilename.jpg');
```

the changed getSize() method now also allows to return the size of temporarily set content which is not possible by filesize() alone.

### DIFF
--- a/core/model/modx/modfilehandler.class.php
+++ b/core/model/modx/modfilehandler.class.php
@@ -501,7 +501,7 @@ class modFile extends modFileSystemResource {
      * 
      * @return downloadable file
      */
-    public function download($mimetype = 'application/octetstream') {
+    public function download($mimetype = 'application/octet-stream') {
         $output = $this->getContents();
 
         header('Content-type: ' . $mimetype);
--- a/core/model/modx/modfilehandler.class.php
+++ b/core/model/modx/modfilehandler.class.php
@@ -382,7 +382,7 @@ class modFile extends modFileSystemResource {
     }
 
     /**
-     * Temporarly set (but not save) the content of the file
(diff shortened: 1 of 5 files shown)


Mark-H:
Thanks @exside for your contribution, and sorry it took so long to get this integrated! Will be included in 2.4 :)

PR #11372, opened 2014-05-14
Enhancement: Add saveCSV() method to modFileHandler

exside:
This is an updated and properly branched version of this PR: https://github.com/modxcms/revolution/pull/681

The added method allows easy saving of a multidimensional array as CSV file, which is very handy when generating an export of any data that can be stored in an array (ie . Collections of MODx objects or whatever objects etc.).
Actually it involves a bit more than just this method (need to generate headers if desired, catch encoding issues etc.), but when the data is prepared nicely in a multidimensional array, it can just be passed to this method and voila...

As an example let's say we have (ya, i know) very simple data like this:

```
$data = array(
    array(
        'name' => 'Name 1',
        'email' => 'email@test.com',
        'address' => 'Test Street 3',
        'mobile' => '240024040420',
    ),
    array(
        'name' => 'Name 2',
        'email' => 'spam@test.com',
        'address' => 'Test Alley 5',
        'mobile' => '058028948220',
    ),
);

// what we can do now to save this (without any further processing etc.)

$fileobj = $modx->file->make(MODX_ASSETS_PATH . 'test.csv');

$fileobj->saveCSV($data);
```

and now we have a nice csv file at assets/test.csv

### DIFF
--- a/core/model/modx/modfilehandler.class.php
+++ b/core/model/modx/modfilehandler.class.php
@@ -431,6 +431,37 @@ class modFile extends modFileSystemResource {
         return $result;
     }
 
+     /**
+     * Writes a CSV file.
+     *
+     * @param array $content An array that holds one csv line per key, each column is a sub-array item
+     * @param string $mode The mode in which to write
+     * @param array $options Optional options for fputcsv, here you can specify custom delimiter and enclosure
+     * @return boolean true when the operation was successful
+     */
+    public function saveCSV($content = array(), $mode = 'w+', $options = array()) {
+        $options = array_merge(array(
+            'delimiter' => ',',
+            'enclosure' => '"',
+        ), $options);
+
+        $result = false;
+
+        if (!empty($content)) {
+            $fp = @fopen($this->path, $mode);
+            if ($fp) {
+                foreach($content as $line) {
+                    if (fputcsv($fp, $line, $options['delimiter'], $options['enclosure']) !== false) {
+                        $result = true;
+                    }
+                }
+                @fclose($fp);
+            }
+        }
+
+        return $result;
+    }
+
     /**
      * Gets the size of the file
      *


OptimusCrime:
Why did nothing happen to this?

core/model/modx/modfilehandler.class.php
```diff
@@ -431,6 +431,37 @@ public function save($content = null, $mode = 'w+') {
...
+    public function saveCSV($content = array(), $mode = 'w+', $options = array()) {
+        $options = array_merge(array(
+            'delimiter' => ',',
+            'enclosure' => '"',
+        ), $options);
+
+        $result = false;
+
+        if (!empty($content)) {
+            $fp = @fopen($this->path, $mode);
```
Alroniks: Oh no! Please, never use @ for suppressing errors.

sottwell:
Indeed. You need to trap the error and generate a log entry for it.

core/model/modx/modfilehandler.class.php
hansek: :+1:

exside: haha, that wasn't me...have a look at all the other "original" modFileHandler methods, they are full of @'s for error suppression, I just used the normal save() method as a template when I wrote this...

exside:
Agree with replacing @'s error suppression with real error handling and error logging, but for that the whole modFileHandler needs to be refactored as this kind of non-error-handling is all over the place in there...

exside:
actually not sure about that saveCSV method anymore, it's too specific (still useful though) for such an abstract class...when re-reading the code it's way too redundant / similar to the existing save() method, so maybe it would be more appropriate to have the possibility to "configure" the save method on how to save the content passed in or do that kind of processing beforehand and then pass the ready to save content to the save() method, e.g. create the csv before passing it as content to the save() method... I'm not sure, but I think there was a reason for that method, maybe csv-export from the manager (via extjs) without always rewriting the same code in some kind of custom processor...

Mark-H:
Not a fan of this either, so I'll agree with you that it's a too specific method for an abstract class and mark this pull request as close. If a different day comes where it does seem relevant, or there's a better implementation, feel free to open a new pull request.

PR #11394, opened 2014-05-22
First attempt to make multilanguage TVs (in Backend) possible

exside:
This was inspired by the forum threads here:

http://forums.modx.com/thread/90958/lexicon-setting-of-the-template-variables-transport-package#dis-post-498390

and subsequently here:

http://forums.modx.com/thread/31482/multi-language-tv

This could come in handy in some scenarios (especially with the envato marketplace available now...so authors could submit multilanguage themes, but in a multilanguage manager...!)

I took @MarkH s idea of making it work like it does for action lexicon entries, e.g. namespace.lexiconkey, so if you do something like that in a TV setup:

bildschirmfoto 2014-05-22 um 03 08 55

and add the appropriate lexicon files under core/components/namespace/lexicons/CULTUREKEY/default.inc.php + a namespace with core path value of {core_path}components/namespace/ in the manager...

an example lexicon file at core/components/tplstrings/lexicon/en/default.inc.php (english) for the image above could look like:

```
/**
 * Custom English language file for manager TVs
 */

$_lang['tv.tvname'] = 'TV Name';
$_lang['tv.tvname.desc'] = 'That\'s a TV description';
```

I'm not sure about the performance hit of something like this (as the $modx->lexicon->load() method doesn't return something I could check in an if statement...which then unavoidably triggers a debug message in the if check on $modx->lexicon()...but yeah, if you set log_level_debug you get many of exactly these messages anyway...see this as an idea, improvements and discussions are very welcome!

for the more visual people, this is how it works...

modx multilangtvs

### DIFF
--- a/manager/controllers/default/resource/resource.class.php
+++ b/manager/controllers/default/resource/resource.class.php
@@ -377,6 +377,16 @@ abstract class ResourceManagerController extends modManagerController {
             $this->setPlaceholder('showCheckbox',1);
         }
 
+        if ($namespace = stristr($tv->get('caption'), '.', true)) {
+            $this->modx->lexicon->load($namespace . ':default');
+            if ($this->modx->lexicon(ltrim(stristr($tv->get('caption'), '.'), '.'))) {
+                $tv->set('caption', $this->modx->lexicon(ltrim(stristr($tv->get('caption'), '.'), '.')));
+            }
+            if ($this->modx->lexicon(ltrim(stristr($tv->get('description'), '.'), '.'))) {
+                $tv->set('description', $this->modx->lexicon(ltrim(stristr($tv->get('description'), '.'), '.')));
+            }
+        }
+
         $tvOutput = $this->fetchTemplate('resource/sections/tvs.tpl');
         if (!empty($this->tvCounts)) {
             $this->setPlaceholder('tvOutput',$tvOutput);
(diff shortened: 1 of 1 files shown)


Mark-H:
> Absolutely agree with that, but what if the lexicon file or objects? (would that be better?) only gets created when somebody enters a string prefixed with the (to be defined) default namespace for tv translations, eg. "tv."?

That sounds a bit too "magical" for my personal tastes.. maybe I've just got too many developer hats on, but adding a lexicon file shouldn't be too hard, and if people want to use the lexicon manager (which in 2.4 learned a new trick to only show certain namespaces) they can also add them there.

mindeffects:
As a first step it would be great if the description of a TV could contain a lexicon tag. This would be enough magic and the fastest way to deliver a multi-language interface for my users.

Mark-H:
I like the idea of just letting people use lexicon tags. No magic there, and people can follow whatever scheme they prefer.

OptimusCrime:
+1

Keep it simple, avoid too much magic.

Mark-H:
There's been a lot of back and forth in this pull request, perhaps it's a good idea to make a fresh one that summarises the discussion?

OptimusCrime:
I think so too. In favor of closing this off.

mindeffects:
Yep. Keeping the simplest solution in mind: Lexicon tags. ;-)

OptimusCrime:
@Mark-H Was this closed as "wont happen" or "start anew"? The feature itself could be pretty cool.

Mark-H:
Closed as
> There's been a lot of back and forth in this pull request, perhaps it's a good idea to make a fresh one that summarises the discussion?

PR #11434, opened 2014-05-30
[2.3] - Added ability to define a default package provider id...

rtripault:
...using `default_provider` setting


5koder:
I can't see this default_provider setting in 2.5.5

Mark-H:
It looks like it wasn't added to the build, but if you create a setting with that key it should work just fine. Might be worth a pull request to add it to the build, but it seems like a kinda niche thing..

PR #11561, opened 2014-06-23
Enhancement: Make TV categories order configurable by a new rank field

exside:
PR solves issues: https://github.com/modxcms/revolution/issues/800, https://github.com/modxcms/revolution/issues/11121

~~This PR doesn't work as it should yet~~...all that is needed to set and save a rank for a category is in place, e.g. new field in schema/map/table + the necessary extJS for the additional field in create/update windows, looks like this then:

create:
bildschirmfoto 2014-06-23 um 17 24 55

update/rename (we can discuss to rename the "rename" stuff to update as it would be more than rename now):
bildschirmfoto 2014-06-23 um 17 24 47

my problem that I couldn't solve is to actually apply this rank to the order of the vertical tabs in the resource tv edit tab...
bildschirmfoto 2014-06-23 um 17 15 11

~~where or what t*\* ... is generating this? Searched everywhere, but can't figure it out...I can see where it's rendered, but not where it gets the data from?? Where is the Category name fetched that is later renderd as a v-tab title or asked differently, where are the categories sorted like they are (alphabetically?)? Anybody?~~
EDIT: Found it, it's done in the loadTVs() method at /manager/controllers/default/resource/resource.class.php

I guess if I knew where to look this would be quite easy to add...


mindeffects:
### Problem
The ranking optin for categories is super awesome! Fact! But if you wanna use the mouse for reordering, it becomes a little complicate: The drag'n'drop only works fine between different levels but not inside a level.

### Example
- Lev 1
- Lev 2
  - Lev B
  - Lev C
  - Lev A
- Lev 3

If you want to reorder the categories inside "Lev 2" you cannot simply drag "Lev A" on the first position.

### Workaround – not really
Draging "Lev B" and "Lev C" out into "Lev 3" and then back into "Lev 2" is a workaround but, well, nope, not really.

### Idea
Perhaps having a guiding line like with resources would be a way to improve the ordering feature? Users are familiar with it.

PR #11682, opened 2014-07-14
Grunt minify javascript proof of concept

OptimusCrime:
I made this really as a proof of concept, but it worked so well I wanted to submit this pull so you could take a look.

What this does is adding two new grunt commands. `grunt concat` which concates a long list of .js-files into one single file (manager.js). The other command is `grunt uglify` which minifies manager.js into manager.min.js.

I've replaced the group minification in the manager and removed the .php-file that does this. It is (imo) redundant. I tried getting all the javascript-files into the minification, but it looks like it is impossible due to the way extjs and l18n is made possible in the manager. I guess that is the price you pay when you build everything around dynamic javscript-solutions.

Feel free to comment on this.

And please read the forum-thread: http://forums.modx.com/thread/91496/why-compress-css-and-compress-js-are-silly


jpdevries:
@Mark-H if semver allows it I think in addition to this the server-side minification should be completely removed. That's the key. Otherwise, the files could still be getting mutilated by the server side minification if `compress_js` is on.

@OptimusCrime might need to resubmit the PR minding whitespace settings in your editor. it looks like there are a lot of unnecessary changes where it's probably just spaces vs tabs

Mark-H:
I like the idea and think it could be beneficial to combine & minify any files that are used throughout the manager from a pre-built minified file. 

> if semver allows it I think in addition to this the server-side minification should be completely removed. That's the key. Otherwise, the files could still be getting mutilated by the server side minification if compress_js is on.

This manager.js file is loaded separately from the "current" /manager/min/ compress_js minification, so any page-specific javascript could still get compressed after this. So I don't think taking out the /manager/min/ is necessary or desired. It's easier to disable during setup now if it is incompatible with the server. Maybe a debug setting of sorts could be used as flag for loading manager.js or the raw files. 

One thing this implementation doesn't seem to handle yet is updating the manager controllers to not register the shared files. This is in modManagerController->registerBaseScripts. With the files there they will still be injected with the minification. This PR only seems to deal with the compress_js_groups which is off by default. 

I'm thinking this PR might need to be refreshed and resolve that ^, but then I don't really see any reason we can't include this in 2.4?

jpdevries:
I'm having another look at this. One note @OptimusCrime I wouldn't recommend prettifying any file in a PR because it is almost guaranteed to make it impossible for integrators to merge the branch
https://github.com/modxcms/revolution/commit/f959f68565aac8f3b6f7d8d0ee9dc6413f40661b

Is there any way you can revert that commit?

I think it's better to get your PR merged and maybe request an integrator prettify it when they merge it or maybe do the prettification in a separate PR.

OptimusCrime:
@jpdevries This was really ment as a proof of concept to see if anyone with more experience with Grunt could come up with something.

However, if this is something worth looking into, I can bake a new PR and try to make thing s a bit smoother.

jpdevries:
@OptimusCrime cool. This is really great, but I am actually working on a slightly different PR 
https://github.com/modxcms/revolution/compare/master...jpdevries:grunt-concat

The grunt-concat branch I'm working on
 - uses grunt to concat the "js group files" just like 2.x always has. Same order and everything.
 - minify those files without mangling them
 - update the templates so that
     - `{if $_config.compress_js_groups}` serves
       - jsgrp-1.js
       - jsgrp-2.js
       - jsgrp-3.js 
     - `{else}` serves
       - jsgrps.js 

That's just done to respect backwards compatibility for the js groups settings. It's kind of silly and I know @opengeek probably isn't fond of adding undo extra weight to the download. I'd be fine not breaking it out into multiple files I'm just trying to keep as much backwards compatibility for now even though in this case it seems kind of pointless. I'll wait to here back if it is ok to nuke support for compressing JS in groups and then if possible:
 - updates templates to just serve jsgrps.js

I'd also be curious if we could nuke the settings entirely so a last PR to follow could be
 - remove the minification directory :tada: 
 - removes compress_css, compress_js, and compress_js_groups settings entirely. For the setup screen, system settings, everything. Like they never happened.

Issue #11117, opened 2014-02-05
überbar advanced search syntax


rthrash:
As a Manager user for a website with many resources, I want to be able to have relevant and good results every time so I can manage my website with ease and not get frustrated with the tool I use to create and update content.

Originally from rthrash/revolution#62

While it needs testing, there may be sites that are large enough that the search isn't performant. 

Romain and I discussed minus-filters, like -resources -plugins -chunks to exclude those from the results … though -elements would be cool to skip all those collectively

Also could just use the first initial -r for skip resources, -c for skip chunks, -p (plugins), -t (templates), -s (snippets) or -pst for skip plugins, snippets and templates. 

Perhaps in:100 to only search descendants of ID 100.

Ibochkarev:
### Triage (current `3.x`)

Still an open **proposal**. None of the advanced query syntax from this issue shipped.

#### What überbar does today

Frontend: `manager/assets/modext/widgets/core/modx.searchbar.js` (`#modx-uberbar`) → connector `Search/Search`.

Backend: `Processors/Search/Search.php` runs a plain `LIKE '%query%'` (plus exact `id` match) across resources, chunks, templates, TVs, snippets, plugins, and users — each gated by the matching `edit_*` permission. Tunables:

- `quick_search_result_max` (default 10 per type)
- `quick_search_in_content` (include element/resource body)

There is **no** parser for `-resources` / `-r` / `-pst`, no `-elements` collective skip, and no `in:100` descendant scope. A typed minus-token is searched as literal text.

Exact resource (and other) ID lookup did land later via [#12783](https://github.com/modxcms/revolution/pull/12783) (merged 2015).

#### Related work that never landed

| PR | Intent | Outcome |
| --- | --- | --- |
| [#11775](https://github.com/modxcms/revolution/pull/11775) | MySQL FULLTEXT / boolean-ish search (tied to this issue + @the-dunnock’s comments) | Closed, not merged |
| [#12533](https://github.com/modxcms/revolution/pull/12533) | “Advanced uberbar” granular search (#12515 / #12283 / this issue) | Closed, not merged |

So the performance path (indexes / FULLTEXT) and the UX path (type filters / `in:`) both stalled.

#### Relevance

Still useful for large trees: content `LIKE` across every type remains expensive, and power users still cannot narrow the result set from the query string. Label `urgent` looks stale for a 2014 RFC that never got a maintainer decision.

#### Suggested direction (if revisited)

[...]

PR #11775, opened 2014-07-22
MySQL Uberbar fixed to use fulltext indexes

the-dunnock:
Also resolves Ryan's feature request : https://github.com/modxcms/revolution/issues/11117

### DIFF
--- a/core/model/modx/mysql/modchunk.map.inc.php
+++ b/core/model/modx/mysql/modchunk.map.inc.php
@@ -175,6 +175,28 @@ $xpdo_meta_map['modChunk']= array (
         ),
       ),
     ),
+    'uber_ft_idx' => 
+    array (
+      'alias' => 'uber_ft_idx',
+      'primary' => false,
+      'unique' => false,
+      'type' => 'FULLTEXT',
+      'columns' => 
+      array (
+        'name' => 
+        array (
+          'length' => '',
+          'collation' => 'A',
+          'null' => false,
+        ),
+        'description' => 
+        array (
+          'length' => '',
+          'collation' => 'A',
+          'null' => false,
+        ),
+      ),
+    ),
   ),
   'composites' => 
   array (
--- a/core/model/modx/mysql/modplugin.map.inc.php
+++ b/core/model/modx/mysql/modplugin.map.inc.php
@@ -144,6 +144,28 @@ $xpdo_meta_map['modPlugin']= array (
         ),
       ),
     ),
+    'uber_ft_idx' => 
+    array (
+      'alias' => 'uber_ft_idx',
+      'primary' => false,
+      'unique' => false,
+      'type' => 'FULLTEXT',
+      'columns' => 
+      array (
+        'name' => 
+        array (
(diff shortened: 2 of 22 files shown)


amdbuilder:
It might be worth noting that MySQL 5.5 will be EOL December 31, 2015 - at least from a cPanel perspective.  I do know some hosts are transitioning to MySQL 5.6 / MariaDB 10 before that time for performance improvements.

Mark-H:
So what should we do with this PR? Get back to it in a few years?

pixelchutes:
Great points re: MySQL 5.5 EOL. Is it something worth waiting for? I'd hate to see this get neglected unnecessarily, as it probably makes a great candidate for breaking change release (2.4?), or whenever the minimum MySQL version officially gets bumped next.

Unless somebody knows a way to make it conditional on MySQL version *(didn't see how to achieve this per the associated map files)*, I worry about the potential impact of the FULLTEXT. Does anyone know what would happen to MySQL 5.5 / InnoDB users with this PR? Would it degrade gracefully and still be useable, minus the new indices?

I also wonder if this may be related to #12515, and whether or not it would impact the efforts there *(or help improve it.)*

OptimusCrime:
> It might be worth noting that MySQL 5.5 will be EOL December 31, 2015 - at least from a cPanel perspective. I do know some hosts are transitioning to MySQL 5.6 / MariaDB 10 before that time for performance improvements.

MySQL 5.5 EOL has passed. Any reason not to merge this now?

rtripault:
I guess the main reason for not merging this is Backward Compatibility.
We should be able to do so for the 3.x branch, so assigning to this milestone

OptimusCrime:
I was not aware that this broke BC.

rtripault:
Sorry, i may have not used the most appropriate term... my point being (from my understanding), merging this PR means rising the MySQL minimum supported version, which could be considered as a breaking change

OptimusCrime:
Ah, then we agree. However, does anyone know what happens if you tried to run this on a 5.5 version? Would the query fail? Is there any way to check if the engine has FT and enable it if it has? That way we could keep the old minimum supported version and at the same time give this functionality too.

pixelchutes:
@OptimusCrime It's my understanding that FULLTEXT indices cannot be created for that version/engine combo. I presume this would result in a MySQL error and a failed install. That said, I have not personally tried it against this PR w/ MySQL 5.5 + InnoDB.

To your point, it would be good to have someone verify the behavior here. Does it fail? Is it "fatal" or is the index simply not created, with the system still functional (only less performant) on MySQL < 5.6?

I'd love to see this support merged one way or another. Is [MySQL 4.1.20](https://develop.modx.com/develop/revolution/requirements/) really the current minimum database requiremed?

Mark-H:
"A few years later", this is once again relevant per #13462. Given the conflicts in this PR and the required minimum version bump, I think we should close it and look towards a fix in 3.0 or after #13462 has been resolved.

Mark-H:
Closing as proposed.

Issue #11225, opened 2014-03-21
Real-time resource alias


eladnova:
When we give Resources a title, the contents of the Title field (1) are mirrored in real-time at the top of a Resource (2). 

Is it difficult to have the same thing happen in the Resource Alias field (3) and just the separate words with a dash?

![real-time-alias](https://f.cloud.github.com/assets/1786476/2484639/9224f238-b10f-11e3-8149-b3f50bcca67c.png)

rthrash:
Since the PR for this issue has been merged for 2.4, can this be closed?

Mark-H:
Yep! My bad, should've closed it when I did the merge.

PR #11799, opened 2014-07-24
[2.3] Enhancement: Real-Time resource alias field

exside:
This is a proposal to implement what was discussed here: https://github.com/modxcms/revolution/issues/11225

The alias field is a bit more complicated than the resource header (the big title), as it should reflect as good as possible what will actually be saved, which means I had to replicate the filterPathSegment() from the modresource.class.php method with all its regexes in JavaScript. Not all is possible (for example trim(regex) is a problem...thanks to IE8), but I'm quite ok with the results.

Tested:
- lowercase only yes / no
- maxlength
- restrict chars: alphanumeric, alpha, legal (=pattern), pattern
- multiple delimiters replaced by one

Caution:
- There is no transliteration available in Javascript, so characters like ä ö ü etc. get transformed on save!

Looks like that:
modx realtime alias

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -674,6 +674,15 @@ $settings['friendly_alias_max_length']->fromArray(array (
   'area' => 'furls',
   'editedon' => null,
 ), '', true, true);
+$settings['friendly_alias_realtime']= $xpdo->newObject('modSystemSetting');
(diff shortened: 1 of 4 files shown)


Mark-H:
Very cool, merged to develop for inclusion in 2.4 :)

OptimusCrime:
Very cool indeed!

exside:
yay! thanks @Mark-H

PR #11861, opened 2014-07-31
[feature] Added a system event to allow developers to assign Smarty variables

rtripault:
### What does it do ?

Adds a system event to enable developers to assign Smarty variables to be used in user navigation.
### Why is it needed ?

Smarty variables could be used, ie., to display a notification count number on a menu entry.
### To do before merge
- [ ] triple check it is not possible to override critical variables (ie. `$_config`)
- [ ] make sure Smarty variables are usable/parsed in lexicon strings

### DIFF
--- a/_build/data/transport.core.events.php
+++ b/_build/data/transport.core.events.php
@@ -868,6 +868,12 @@ $events['OnElementNotFound']->fromArray(array (
   'service' => 1,
   'groupname' => 'System',
 ), '', true, true);
+$events['OnBeforeMenuPlaceholders']= $xpdo->newObject('modEvent');
+$events['OnBeforeMenuPlaceholders']->fromArray(array (
+  'name' => 'OnBeforeMenuPlaceholders',
+  'service' => 1,
+  'groupname' => 'System',
+), '', true, true);
 
 /* Settings */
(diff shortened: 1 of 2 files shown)


Mark-H:
Any update on those to dos? Looks like it could be a cool feature.

rtripault:
Nope. I guess i was looking for some kind of feedback before digging in those :)

Mark-H:
Well the use case you mentioned (notification counts) is really appealing and I can already think of a few extras I could consider adding it too (Scheduler with the number of scheduled runs, SimpleCart with the number of new orders etc). 

Thinking about it some more I'm not sure if adding "global variables" to smarty is the best approach to dealing with something like that.. maybe (from an extras dev perspective) adding a simple `<span class="scheduler-run-count"></span>` to the menu lexicon and using Ext.get().update (maybe with some ajax) would be a bit cleaner? It then also doesn't restrict you to using that lexicon in smarty, the same lexicon key from the menu could also be used for a component header or something, which is js/ext/`_`..

rtripault:
You are definitely right, the use case (counts) would be best suited/filled with some ajax. And since i don't really remind the other awesome use cases i could have thought about when submitting the PR (and the possible drawbacks to it), i'd be up to close it

PR #11865, opened 2014-08-01
Nullify some methods in ResourceDataManagerController which could potentially cause issues

rtripault:
### What does it do ?

Nullify some methods in `ResourceDataManagerController`
### Why is it needed ?

The controller fires some events (`On*Doc*`) which is not appropriate since most of those events are used on resource (creation/edition) form.


Mark-H:
Merged into 2.4 in 815339651d2c60f981c4c7a58c2825a88ccc5e8e

PR #11866, opened 2014-08-01
Auto "add package" when listing manager actions, when appropriate

### LINKED ISSUE #8747: error-log entries after calling Reports/Manager Actions
erhard created Redmine issue ID 8747

After calling Reports/Manager Actions there are many log-entries in the error log like this:
<pre>
[2012-09-21 13:36:12] (ERROR @ /connectors/system/log.php) Could not load class: unknown from mysql.unknown.
[2012-09-21 13:36:12] (ERROR @ /connectors/system/log.php) unknown::load() is not a valid static method.
</pre>
Even Revo 2.2.4-pl in MODX-Cloud produces this error(-log-entry).

rtripault:
### What does it do ?

Automatically calls `addPackage` when listing manager actions (`?a=system/logs`) if an action is prefixed with a namespace.
### Why is it needed ?

Unless the third party package registers itself (extension_packages), when listing manager actions dealing with third party tables, packages aren't loaded, thus we can't get appropriate objects (filling the logs with loads of `Could not load class: ClassName from mysql.classname. className::load() is not a valid static method.`).
### To do before merge
- [ ] make sure most (all ?) third party packages follow the file structure convention (`namespace/model/`)
### Related issue(s)/PR(s)

Closes #8747

### DIFF
--- a/core/model/modx/processors/system/log/getlist.class.php
+++ b/core/model/modx/processors/system/log/getlist.class.php
@@ -102,7 +102,18 @@ class modSystemLogGetListProcessor extends modProcessor {
      */
     public function prepareLog(modManagerLog $log) {
         $logArray = $log->toArray();
-        if (!empty($logArray['classKey']) && !empty($logArray['item']) && $logArray['item'] !== 'unknown') {
+        if (strpos($logArray['action'], '.') !== false) {
+            // Action is prefixed with a namespace, assume we need to load a package
+            $exp = explode('.', $logArray['action']);
+            $ns = $exp[0];
+            $path = $this->modx->getOption(
+                "{$ns}.core_path",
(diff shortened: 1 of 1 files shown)


Mark-H:
Hi @rtripault what do you think about this one, ready to go into 2.4? I'd say any package that doesn't follow that file convention doesn't deserve to have their package loaded :P

rtripault:
I'd say that should work! The only issue i could see so far is the potential case where a namespace have multiple "models" (not aware of any component doing so).
So should not harm being in 2.4...

Mark-H:
Merged into 2.5.0-rc1 in b4b6aada7b62398f22d9c53ada2c62419a236d42, thanks!

PR #11908, opened 2014-08-18
Update modfilemediasource.class.php

iseredov:
[2.3.1] Error when using absolute Base URL in Media Sources. Duplicate part of image url.


TheBoxer:
Fixed by another PR.

PR #11911, opened 2014-08-18
[WIP] Add destructors to JS objects (required by AjaxManager plugin)

danyaPostfactum:
Work in progress. There is a issue with TV fields - they and their TreeDrops (created by MODx.makeDroppable) should be destroyed properly. Otherwise Drag'n'Drop is broken (ExtJS while dragging tries to access removed DOM nodes).


theboxer:
Feel free to reopen with more commits that finishes the PR.

PR #11923, opened 2014-08-20
Prevent save/remove on modAccessibleObject (and derivative classes) if appropriate policy is not met

### LINKED ISSUE #9757: modAccessibleObject::remove and ::save permissions error
fi1osof created Redmine issue ID 9757

While user have not permissions, process not breaks, and object save and remove.
@    public function remove(array $ancestors= array ()) {
        $removed = false;
        if (!$this->checkPolicy('remove')) {
            $this->xpdo->error->failure($this->xpdo->lexicon('permission_denied'));
        }
        $removed = parent :: remove($ancestors);
        return $removed;
    }

```
public function save($cacheFlag = null) {
    $saved = false;
    if (!$this->checkPolicy('save')) {
        $this->xpdo->error->failure($this->xpdo->lexicon('permission_denied'));
    }
    $saved = parent :: save($cacheFlag);
    return $saved;
}
```

@

rtripault:
### What does it do ?

Prevent save/remove on modAccessibleObject (and derivative classes) if appropriate policy is not met
### Why is it needed ?

The current code does not "break" execution if appropriate policy is not met, resulting in object removal/save even if authenticated user does not have the right to.
### Related issue(s)/PR(s)

Closes #9757

### DIFF
--- a/core/model/modx/modaccessibleobject.class.php
+++ b/core/model/modx/modaccessibleobject.class.php
@@ -168,6 +168,7 @@ class modAccessibleObject extends xPDOObject {
         $saved = false;
         if (!$this->checkPolicy('save')) {
             $this->xpdo->error->failure($this->xpdo->lexicon('permission_denied'));
+            return $saved;
         }
         $saved = parent :: save($cacheFlag);
         return $saved;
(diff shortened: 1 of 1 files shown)


alroniks:
@opengeek Why it can not be done this way?

PR #11927, opened 2014-08-21
Fixed a few issues in modX->_loadExtensionPackages preventing to load some services & handle table prefix

### LINKED ISSUE #9573: MODx::_loadExtensionPackages() ->addPackage() not working
fi1osof created Redmine issue ID 9573

For continue of http://tracker.modx.com/issues/9572
We have:
@if (!empty($package['serviceName']) && !empty($package['serviceClass'])) {
    $this->getService($package['serviceName'],$package['serviceClass'],$package['path']);
}@
But modExtensionPackage have not columns serviceName or serviceClass. Hi have service_name and service_class.
In this case this condition return true never, and getService() do in MODx::_loadExtensionPackagesDeprecated()

rtripault:
### What does it do ?

Fixed a few issues in `modX->_loadExtensionPackages` preventing to load some services & handle table prefix
### Why is it needed ?

"Auto loading" services fail without this, as well as packages using table prefix.
### To do next

While digging on that issue, it also appeared there is some inconsistencies about path... (mostly about services).
If `modExtensionPackage.path` is empty, it will fallback to `modNamespace.path` (see https://github.com/modxcms/revolution/blob/master/core/model/modx/modcachemanager.class.php#L466).

Later, we add `model/` to the extension path (see https://github.com/modxcms/revolution/blob/master/core/model/modx/modcachemanager.class.php#L471), no matter if `model` is already present or not (ie. being set in `modExtensionPackage.path`).

Then, if a `modExtensionPackage.service_class` & `modExtensionPackage.service_name` is found, we try to load the appropriate service, using the package path... (ending with `model/`).
This will result in an error, since no service will be found : the path should have been `../model/$namespace`, unless the service_class is `Namespace.ServiceClass`.

Hope it makes sense.
### Related issue(s)/PR(s)

Closes #9573

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -564,10 +564,10 @@ class modX extends xPDO {
         $cache = $this->call('modExtensionPackage','loadCache',array(&$this));
         if (!empty($cache)) {
             foreach ($cache as $package) {
-                $package['tablePrefix'] = !empty($package['tablePrefix']) ? $package['tablePrefix'] : null;
-                $this->addPackage($package['namespace'],$package['path'],$package['tablePrefix']);
-                if (!empty($package['serviceName']) && !empty($package['serviceClass'])) {
-                    $this->getService($package['serviceName'],$package['serviceClass'],$package['path']);
+                $package['table_prefix'] = !empty($package['table_prefix']) ? $package['table_prefix'] : null;
+                $this->addPackage($package['namespace'],$package['path'],$package['table_prefix']);
+                if (!empty($package['service_name']) && !empty($package['service_class'])) {
+                    $this->getService($package['service_name'],$package['service_class'],$package['path']);
                 }
             }
         }


opengeek:
This changes how extension packages load and would break existing packages. The properties defined in the extension packages are set using the keys you changed. Shouldn't you just define your package with the mixed case property names?

rtripault:
As far as i understand the code, those changes would only apply for extension packages stored in the DB (`modExtensionPackage`) and not to the ones set the "old way" using a system setting.

Therefore, we have no control of the properties, since they are defined by the [table columns](https://github.com/modxcms/revolution/blob/master/core/model/schema/modx.mysql.schema.xml#L741-767), which make use of "underscores".

All extension packages added the "old way" would then still be processed by [_loadExtensionPackagesDeprecated](https://github.com/modxcms/revolution/blob/master/core/model/modx/modx.class.php#L574), which makes use of "mixed case" property names.

opengeek:
My god, what a mess — ok then, seems you know more about it than I do. I didn't even know we had a database table for that...

rtripault:
Please, don't assume my understanding is correct ;)

I'm not sure what drove the choice/idea of storing the "packages" into their own table, but in addition of the "drawbacks" i see (not being able to define the "loading order", not being able to use different configuration per context), its implementation is not complete (`modX::addExtensionPackage` & `modX::removeExtensionPackage` still make use of the system setting, while they should use the table for BC purpose).
But that's another story :)

opengeek:
I'm not sure what drove that decision either and this is worrying me a bit. But if this doesn't break anything, I'll go ahead and merge it for 2.3.3.

Fi1osof:
@rtripault please, have a look this: https://github.com/modxcms/revolution/pull/12359#issuecomment-180780998

Issue #3395, opened 2011-01-11
changing policy template for access policies


u007:
u007 created Redmine issue ID 3395

Hi,

ive tried to add in a custom policy in the access policies, and i realized that i couldnt do it without adding it in the policy template.
Is there a way to change the policy template of an existing access policies without editing the database table?

Or maybe its possible to add in custom policy name into access policies without involving changes on the policy template..

please advice,

thank you.

alroniks:
You can just add (at least in current 2.6.4 version) new Policy Template and add to it any permissions that you want. And after that create new Policy based on this Template and select needed permissions. So templates for creating lists of permissions, policies for configuring such lists.
I am closing it because nothing to fix. But if you have questions, please, ask. I will try to help.

PR #11940, opened 2014-08-26
Enhancement: Make access policy template editable for access policies via grid editor

exside:
This is a proposal to change an access policy's template via inline grid editor/combobox, it works so far, but an access policy with many permissions (e.g. based on AdministratorTemplate for example) will still have them after switching the template (for example if you switch the Content Editor Policy from AdministratorTemplate to ElementTemplate it will still have 24 permissions enabled instead of only the ones available in Element Template...)

I don't know yet how to solve this problem, but maybe somebody else has an idea or approach that could be tried out and added to this PR!

This PR is related to https://github.com/modxcms/revolution/issues/11937, #7383 and #3395

### DIFF
--- a/manager/assets/modext/widgets/security/modx.grid.access.policy.js
+++ b/manager/assets/modext/widgets/security/modx.grid.access.policy.js
@@ -70,7 +70,20 @@ MODx.grid.AccessPolicy = function(config) {
             ,editor: { xtype: 'textfield' }
         },{
             header: _('policy_template')
-            ,dataIndex: 'template_name'
+            ,dataIndex: 'template'
+            ,editor: { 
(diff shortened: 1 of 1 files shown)


Mark-H:
I'm not sure if we need to allow changing the template a policy is based on, after all the available permissions depend on the chosen template, so if you change that you'll also need to make sure those permissions are fixed. Have a feeling this might result in very hard to track down ACL related issues with orphaned permissions!

If this would also clean up permissions that aren't in the new template I might be on board with this as then it can be useful for the use case described in #7383 without leaving stuff behind that shouldn't be there.

Mark-H:
Closing as concerns from over a year and a half ago haven't been fixed. Feel free to open a new pull request if you think it's still relevant.

Issue #11850, opened 2014-07-30
Updated search position (includes reintroduction of site name/version).


eladnova:
I really like how OSX handles the search (screengrab below). 
• User clicks on the search icon
• Search "drawer opens"
• Search results displayed under this.

![osx-search](https://cloud.githubusercontent.com/assets/1786476/3749688/81af7c84-17ee-11e4-973e-e847ebe91c00.png)

The benefits of this are primarily that the search field can be as long as required without affecting other menu items. In MODX 2.3, the search expands which can affect custom menus and it also has knocked the sitename and version off the manager. 

## How this might work in MODX 2.3 

We tweak the current search to mimic the OSX drawer style but displaying just the icon and no search field by default. Clicking icon reveals drawer with search field and results. 
Happily, this allows space for us to reintroduce the sitename / version too. Again, we're not restricted by the search field. Granted, its an extra click to reveal the search field but I think benefits outweigh it.

Mockup below is for illustration. I know it's not pretty!

![search-v1](https://cloud.githubusercontent.com/assets/1786476/3749723/e1c13f36-17ee-11e4-9a1f-632806d8e9d9.png)

## Alternative approach 
Alternatively, I think a neater approach would be to add a search icon in a tab beside the Resources, Elements and Files tabs. The search field sits just underneath this and we categorise all the results within the existing Templates, TVs, Chunks etc toggle bars. We'd add an extra one for Resources and remove the plus and refresh icons. But as a quick mockup. I think this is a very natural fit.

![search-v2](https://cloud.githubusercontent.com/assets/1786476/3749812/a845ddc4-17ef-11e4-9822-bf931db4c2a4.png)

OptimusCrime:
This implemented in 2.4. Any objections on closing this ticket?

PR #11958, opened 2014-09-04
Uberbar improvements, makes room for site information in the header (site name / MODX version)

pixelchutes:
Based off ideas from #11850

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -56,7 +56,7 @@
 
 #modx-topnav {
   /* refactor :not selector to work in IE8 */
-  > li:not(#modx-home-dashboard) {
+  > li:not(#modx-home-dashboard):not(#modx-manager-search-icon) {
     > a {
       @include media($desktop) {
         text-indent: -99999em;
@@ -312,4 +312,12 @@
 .ext-ie7 #modx-header {
   position: relative;
   z-index: 10;
+}
+
+#modx-topnav #modx-version-info {
+  padding: 13px 15px 0 15px;
+  border-right: 0;
+
+  .site_name { color:#fff }
+  .full_appname { color:#ccc; font-style: italic }
 }
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -314,10 +314,23 @@
   z-index: 10;
 }
 
-#modx-topnav #modx-version-info {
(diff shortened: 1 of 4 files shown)


jaygilmore:
Is this PR still being considered? It's not been closed and this seems like there is some value to at least parts of it.

pixelchutes:
I'd hope so, @jaygilmore -- My original motivation was purely reintroduction of site name / version, which many still seem on the fence about, but ended up improving / fixing a few minor issues along the way.

Also, this is an attempt at creating just a little more room in an already cramped space...especially on smaller devices, like @gpsietzema has illustrated above.

OptimusCrime:
I +1 for this merged in the 2.4-branch

Mark-H:
This was just merged into develop for 2.4. Great work @pixelchutes :)

theboxer:
The search doesn't close when you click somewhere else or on `esc` key press, so you have to hit the search icon again...

pixelchutes:
It's been awhile since I've tested that, but I was pretty sure it was accounting for `esc` / clicking outside. I believe in the animated gifs above you can even see results go away when I am pressing escape (hard to illustrate).

pepebe:
This is colliding with my current refurbishing of the uberbar. 

bildschirmfoto 2015-06-30 um 16 09 41

I agree that the site information should be shown in a prominents space as I tend to work in more than one MODX project at the same time and it is very hard to tell them apart.

I'll have a look how to integrate this. PR will hopefully be coming up in the next hours.

pixelchutes:
@theboxer I'm not sure why yet (or if it has to due with merging to `develop`) but it doesn't seem like the original `Ext.KeyMap` is being applied, so the following is not being triggered as it should:
```javascript
Ext.getCmp('modx-uberbar').hideBar();
```

pixelchutes:
Not sure if it's related, but it may conflict with this change: https://github.com/modxcms/revolution/commit/5aae7ea6fc3a1ef61ce3e91b667d4a6e1d1859e3#diff-13ba45e0161390fbee6c1a566a7d4a88

pixelchutes:
This was just fixed in `develop` via 1e37e91da0c0820a004ed80663cf46f79239680d

The aforementioned PR (previous post) was inadvertently disabling _both_ keymap listeners instead of just the offending uberbar keyboard shortcut. The new fix restores the ability to close uberbar via  <kbd>Esc</kbd>

PR #12020, opened 2014-10-04
apply Country Codes in lexicons

goldsky:
At the moment, if you save country in user's profile from Manager, it will save string value.
This commit is to replace that value to county's iso code instead, based on http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
There are some additional new countries as well.
Please check english version for detail.


pixelchutes:
Any progress on the status of this PR?

pixelchutes:
@goldsky This is great! 

The only "issue" I ran into is it appears to be **Case-Sensitive**. For example, when my country is defined as `US` in the database, `United States` does **not** get pre-selected in the country dropdown list. If I change it to lowercase `us` matching the lexicon keys, it works as expected.

For me, I typically deal with the uppercase version of the ISO code *( for example, when combining`[[!UpdateProfile]]` with `[[!FormItCountryOptions]]`)*, but it makes sense to keep the lexicon entries lowercase. 

Are there any options to update the PR to support both UPPERCASE and lowercase scenarios?

goldsky:
Because this is javascript problem, so I would rather to offer [javascript solution](https://github.com/modxcms/revolution/pull/12020/files#diff-7c4a9c30257e28d5f4a5a582b0fc221bR60).

pixelchutes:
I think there are ultimately two problems, your update solves one of them. (Thanks!)

However, if the value is stored as lowercase in the database, it becomes more than a client-side JavaScript issue. For example, issues pre-populating with `[[!FormItCountryOptions]]`.

I've always known valid ISO country codes to be uppercase, but perhaps it's more a matter of preference than an actual standard.

pixelchutes:
It may sound like overkill, but what about a system setting to allow the choice of which way the Country ISO code gets stored: `UPPERCASE` vs. `lowercase`? Apparently there are some use cases for storing *both* ways.

OptimusCrime:
Sounds like overkill to me

pixelchutes:
Yeah, I think I'd have to agree...

That said, how do we settle on the preferred approach? My vote is obviously for Uppercase *(US, CA, MX, etc)*

goldsky:
I'll wait for others' opinions

pixelchutes:
Let's rally the troops! :smile:

pixelchutes:
@opengeek brought up a good point in favor of Uppercase ISO codes, allowing the potential to be used in language culture tags.

http://www.w3.org/International/articles/language-tags/
https://msdn.microsoft.com/en-us/library/ee825488(v=cs.20).aspx

@TheBoxer are you actively working on this? If not, feel free to assign to me, as we also discussed the necessary conversion during upgrade for existing user's country records to ISO _(ideally via batch update using `updateCollection()`)_

Mark-H:
I'd also be mildly in favour of uppercase for selfish reasons, but that's not at all a blocker imo. Using it as uppercase can be done by `[[+country:ucase]]`... 

You're working on merging this one right @pixelchutes? ;)

pixelchutes:
Yes, I still need to test through / optimize the required update handling for  *large* user sets.

pixelchutes:
### WIP:
- [x] Account for possible country string variations *(based on manager_language)*
- [x] Account for invalid UTF-8 sequence *(fixed encoding error for 3 of the countries from PR)*
- [x] Ensure ISO code exists during conversion *(e.g skip numeric legacy `_country_lang` keys until alternate language country lexicons land from Crowdin)*
- [x] Convert existing country strings in User Profile to uppercase ISO country codes during 2.4.0 upgrade

OptimusCrime:
Is this ready for testing now?

Great work btw! :D

pixelchutes:
Not merged yet, but will be soon!

PR #12034, opened 2014-10-08
Enable path style on AWS driver if the bucket's name contains a dot

Omeryl:
The issue lies in how wildcard certificates work. If a bucket has dots in it, https://some.bucket.name.s3.amazonaws.com/ fails SSL verification, resulting in a cURL PHP error. So we need to tell the driver to use path style to access the bucket. This changes the url to https://s3.amazonaws.com/some.bucket.name/ which passes SSL verification.

Fixes #12017


aesop1:
This is still an issue. Buckets without dots work; buckets with dots do not.

PR #12041, opened 2014-10-10
First step on allowing developers to "hook" into the search bar

rtripault:
### What does it do ?

Introduce a new system event `OnSearch`.
### Why is it needed ?

This would allow developers to plugins to display their results (ie. from external DB/component).
### Usage/notes

The event sends 3 parameters : 

| Key | Type | Description |
| --- | --- | --- |
| `query` | string | the raw query |
| `limit` | int | the maximum number of results to return |
| `actionToken` | string | the string "token" used to indicate user is looking for an action (currently not supported in "core search") |

The search processor expect a multidimensional array as output (`$modx->event->_output`).
#### Single record array :

| Key | Type | Description |
| --- | --- | --- |
| `_action` | string/int | the action URI (minus the `a=`) |
| `label` | string | the label to display for the result section (ie. "Resources", "Snippets", "My custom objects") |
| `icon` | string | the icon CSS class (currently suppose FontAwesome), ie. `heart` to produce a `icon icon-heart` result |
| `name` | string | the record name/title/label |
| `description` | string | the record description |
| `type` | string | a string to define the "section type" (to group results), (ie. "resources", "my-objects"...) |
#### Example

```
array(
        'name' => $item->get('field_used_as_name'),
        '_action' => 'edit&namespace=mycomponent&id=' . $item->get('id'),
        'description' => $item->get('field_used_as_description'),
        'type' => 'my-custom-object',
        'label' => 'My Custom Objects',
        'icon' => 'cube',
    )
```
#### Full plugin example

```
<?php
/**
 * Sample plugin to handle custom search
 *
 * @var modX $modx
 *
 * @event OnSearch
 */

$query = $modx->event->params['query'];
$limit = $modx->event->params['limit'];

$c = $modx->newQuery('Object');
$c->where(array(
    'name:LIKE' => "%{$query}%",
    'OR:description:LIKE' => "%{$query}%",
));
$c->limit($limit);

$results = array();
$collection = $modx->getCollection('Object', $c);
/** @var xPDOObject $item */
foreach ($collection as $item) {
    $results[] = array(
        'name' => $item->get('name'),
        '_action' => 'update&namespace=component&id=' . $item->get('id'),
        'description' => $item->get('description'),
        'type' => 'my-object',
        'label' => $modx->lexicon('my-objects-label'),
        'icon' => 'cube',
    );
}

if (!empty($results)) {
    $modx->event->_output = $results;
}

return '';
```
### To do before merge
- [ ] Discuss, clean, optimize the search processor
- [ ] Think about and "advanced" search page/UI (using the same processor) and how to handle advanced criteria (to prevent breakage if refactoring is needed)
- [ ] Do not assume icons will use FontAwesome with the `icon icon-*` syntax
- [ ] Create the system event during upgrade

### DIFF
--- a/core/model/modx/processors/search/search.class.php
+++ b/core/model/modx/processors/search/search.class.php
@@ -5,11 +5,36 @@
  **/
 class modSearchProcessor extends modProcessor
 {
+    /**
+     * The default maximum results for each result "type"
+     *
+     * @var int
+     */
     public $maxResults = 5;
+    /**
+     * The character used to identify an action search/execution
+     *
+     * @var string
+     */
     public $actionToken = ':';
+    /**
+     * Temporary fake actions pool to search against
+     *
+     * @var array
+     */
     private $actions = array();
 
+    /**
+     * The search query
+     *
+     * @var string
+     */
     protected $query;
+    /**
+     * An array of found results matching the query
+     *
+     * @var array
+     */
     public $results = array();
 
     /**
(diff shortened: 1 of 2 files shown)


rtripault:
Closing since the search bar is taking a different direction

PR #12099, opened 2014-10-29
Fix remove and update from window context settings

theboxer:
### What does it do ?

Fixes removing context settings and updating context settings from a window.
### Why is it needed ?

Because it doesn't work :)
### Related issues

Resolves #12086


maherelaridi:
@TheBoxer it worked for me! thanks :)

PR #12110, opened 2014-11-03
Update memberof output filter to return integer (0 or 1)

pixelchutes:
This change mirrors the other conditional output filters that all return an integer `0` or `1` instead of a `boolean`

Also, this fixes #5612 related issue mentioned near the bottom.

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -142,9 +142,9 @@ class modOutputFilter {
                             /** @var $user modUser */
                             $user = $this->modx->getObject('modUser',$output);
                             if ($user && is_object($user) && $user instanceof modUser) {
-                                $condition[]= $user->isMember($grps);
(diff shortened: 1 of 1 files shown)


goldsky:
This is odd.
Even with this change, this output filter keeps returning the user's ID instead.
I'm on v2.3.2-pl

```
<ol>
    <li>[[+modx.user.id]]</li>
    <li>[[+modx.user.id:mo=`Member`]]</li>
    <li>[[+modx.user.id:mo=`DoesNotExistGroup`]]</li>
</ol>
```

TheBoxer:
It's conditional modifier, so it's not modifying output.

```
<ol>
    <li>[[+modx.user.id]]</li>
    <li>[[+modx.user.id:mo=`Member`:then=`ok`:else=`nok`]]</li>
    <li>[[+modx.user.id:mo=`DoesNotExistGroup`:then=`ok`:else=`nok`]]</li>
</ol>
```

goldsky:
ah, ic.
thx.

PR #12196, opened 2014-12-04
Update cache.manifest.php

Kidokyfe:
Small fix to an actual status.


OptimusCrime:
Bump?

Mark-H:
Wow that's a long list of files right there! From a quick scan that looks roughly like all files in the manager folder. I'm not very well-versed in the html5 cache manifest, but wouldn't including that many files cause issues with too many files being cached? Many of those files are probably not requested directly because of assets being served combined and minified, and there's also a bunch of files that belong to an older manager design that we should probably get rid of..

Pathologic:
@Mark-H Maybe this PR can be done as extra?

Mark-H:
Because the manifest has to be referred to in the head of the page that might actually be really tricky to do @Pathologic, though with a lot of PRs I would definitely agree :P There's also some of this in the core already, so improvements to that are definitely welcome I guess...

OptimusCrime:
> Deprecated
> This feature has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Do not use it in old or new projects. Pages or Web apps using it may break at any time.
- https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache

@modxbot close

OptimusCrime:
Looks like I dont have the permissions to close this. @theboxer, would you mind?

OptimusCrime:
Cheers @pixelchutes :)

Issue #11096, opened 2014-01-31
Add ID number to manager pages (resources and elements)


frogabog:
Could we view the ID of whatever we're viewing as a page in the manager, somewhere on the page itself?  Perhaps in the title area in parentheses? 

Example: If you've navigated elsewhere through the resource tree for any purpose, fetching an ID can prove to be a little awkward having to go back and re-open the tree to see them.  

Working reasons aside, it'd just be nice to immediately know what the ID is every time the resource is opened.  It would do wonders for ID association for developers, because after a while you'd remember the ID as well as the title/name just because you see it every time you open it.

Elements need this less, but they too may as well show the IDs.  Why not?

OptimusCrime:
Should this be done with or without a system setting for toggling the behavior?

gpsietzema:
Please do it with a system setting, because we personally don't want to show it to our clients.

pepebe:
System setting would be fine with me.

wuuti:
Note: there are already permissions "tree_show_element_ids" and "tree_show_resource_ids". It makes sense to use permissions for that instead of a setting and would be wise then to add a permission "form_show_resource_ids" or something like that.

OptimusCrime:
Good catch. I think that makes sense as well.

JoshuaLuckers:
This would be a nice addition. Maybe it could be added to MODX 3?
If needed I would like to help getting this feature in MODX 3.

OptimusCrime:
Isnt there an id there today? Seem like noone has complained since we rolled out the new design (version 2.3?). In favor of closing this.

JoshuaLuckers:
There isn't. See the screenshot posted earlier in this issue.

OptimusCrime:
You are right, my bad. I thought perhaps the ID was added after the 2.3 redesign.

PR #12197, opened 2014-12-05
Quick update title

argnist:
### Related to

Partly to #11096 and @frogabog in ambassadors channel
### What it does

It shows name of element and pagetitle of resource with ID in Quick Update window title.
### Why is it needed

Open two quick updates, anything. Then collapse them. They both say the same thing. Not the actual chunk name, etc.

### DIFF
--- a/manager/assets/modext/widgets/element/modx.tree.element.js
+++ b/manager/assets/modext/widgets/element/modx.tree.element.js
@@ -248,16 +248,20 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
             }
             ,listeners: {
                 'success': {fn:function(r) {
+                    var nameField = (type == 'template') ? 'templatename' : 'name';
                     var w = MODx.load({
                         xtype: 'modx-window-quick-update-'+type
                         ,record: r.object
                         ,listeners: {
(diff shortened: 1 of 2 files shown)


Mark-H:
Hmm should it look like this? Looks a bit weird to me, maybe there needs to be some kind of visual separator? Or perhaps we can lose the "normal" title?
dashboard___modx_revolution

TheBoxer:
@Mark-H How about `:` as a separator before the element name?

screen shot 2015-06-18 at 13 44 47

TheBoxer:
Also, @argnist I see changing window title in `success` callback a bit weird solution... You can pass additional param to the quick update window, which will be added to the title. Something as:

```js
var nameField = (type == 'template') ? 'templatename' : 'name';
var w = MODx.load({
    xtype: 'modx-window-quick-update-'+type
    ,titleElement: r.object[nameField] + ' (' + r.object.id + ')'
```

and then in the quick update window definition:

```js
config.titleElement = ': ' + config.titleElement || '';
    
Ext.applyIf(config,{
    title: _('quick_update_snippet') + config.titleElement
```

TheBoxer:
Merged

PR #12198, opened 2014-12-05
Fix html5 chache

Kidokyfe:
Any more data when editing resources won't be cached.


Jako:
Closing this because the html5 cache is deprecated: https://github.com/modxcms/revolution/pull/12196#issuecomment-214554561

PR #12201, opened 2014-12-07
Add check for unpublish_document

argnist:
and fix error 500 in recently edited resources grid
### How to reproduce error 500
1. Go to user Profile (?a=security/profile) to Recently edited resources tab.
2. Try to change something.
3. Values in grid are changed but request returns error 500. After reload grid changes are dissapeared.


TheBoxer:
Merged to master & develop

PR #12203, opened 2014-12-07
Make Gravatars always load over https

victorhaggqvist:
Omit the setting `url_scheme` since there is no default for it AFAIK.

This change would make every thing in the manager load over https, since this is the only thing not beeing https as fas as I have found. Also there is not really a point of not loading it over https anyway right?

### DIFF
--- a/manager/controllers/default/header.php
+++ b/manager/controllers/default/header.php
@@ -111,7 +111,7 @@ class TopMenu
                     trim($userProfile->email)
                 )
             );
-            $gravsrc = $this->modx->getOption('url_scheme', null, 'http://') . 'www.gravatar.com/avatar/'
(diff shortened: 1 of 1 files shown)


rtripault:
I guess the only "drawback" is the response speed (https being supposed to be slightly slower than http), besides that, i'm seeing no reason not to do so, so :+1:

pixelchutes:
+1

What about protocol agnostic approach:

```php
$gravsrc = '//www.gravatar.com/avatar/'
```

amdbuilder:
You would likely encounter connection problems running MODX locally using the protocol agnostic approach.  It's a great idea otherwise.

sepiariver:
Looks like everyone here at least does not see a problem with it. @amdbuilder 's comment is enough reason to stick with setting the protocol.

PR #12210, opened 2014-12-13
Custom Icons for Contexts

jpdevries:
If this feature already exists please disregard.
Was under the impression it
already did but doesn’t appear to (i know it does for resource class_key)

the ternary may see overkill but this wasn’t working as expected:

``` php
'iconCls' => $context->getOption('mgr_tree_icon',null,$this->modx->getOption('mgr_tree_icon_context', null, 'tree-context')),
```

To use custom icons on context tree icons, create a `mgr_tree_icon` Context Setting with a value of a [Font Awesome Icon](http://fortawesome.github.io/Font-Awesome/icons/) within the given context.


TheBoxer:
Yep, the proposal from Romain works fine.

Issue #11980, opened 2014-09-15
syncsite checkbox configurable


sottwell:
It would be nice if the syncsite chechbox state were configurable instead of hard-coded...

```
'hidemenu' => $this->context->getOption('hidemenu_default', 0, $this->modx->_userConfig),
'published' => $this->context->getOption('publish_default', 0, $this->modx->_userConfig),
'searchable' => $this->context->getOption('search_default', 1, $this->modx->_userConfig),
'cacheable' => $this->context->getOption('cache_default', 1, $this->modx->_userConfig),
'syncsite' => true,
```

whitebyte:
Since related PR is merged, i suggest this issue should be closed? Pinging @sottwell

sottwell:
It also needs to be done in the update processor
manager/controllers/default/resource/update.class.php line 122
	
	        $this->resourceArray['syncsite'] = isset($this->resourceArray['syncsite']) && intval($this->resourceArray['syncsite']) == 1 ? true : $this->context->getOption('syncsite_default', 1, $this->modx->_userConfig);

PR #12218, opened 2014-12-17
Respect default setting for "syncsite" ("Empty cache" on resource save)

rtripault:
### What does it do ?

Takes care of the `syncsite_default` setting when updating a resource.
### Why is it needed ?

`syncsite` is never stored, unless you define a Form Customization rule doing so.

Before this PR, the behavior was that the "Empty cache" checkbox was checked if no Form Customization rule dealing with "syncsite" was found, not matter what was defined as `syncsite_default` setting.
### Related issue(s)/PR(s)
- Should take care of #11980


rtripault:
Updated, ready for review!

PR #12230, opened 2014-12-23
Fixed issue with large number of media sources in sources.getlist

### LINKED ISSUE #12215: Bugs in sources.getlist with large number of media sources
By default, the first 20 media sources are fetched. However, since the sources are removed _after_ they are fetched, anywhere from 0 to 20 sources could be in the final list.

This causes a few problems, particularly in the resources browser.

1) If a user has access to none of the first 20 sources, then the combobox will not let him select another one. (Just disappears). The combobox will also disappear if the user navigates to any page number that yields zero results.

2) ModX also fetches the name of the media source from the list of sources. If it is not present in the list, then the ID is displayed instead.

This means that _sources.getlist_ should remove invalid sources _before_ pagination in order to work as expected.

Here is the hack I've applied to _source/getlist.class.php::prepareQueryBeforeCount()_:

``` php
        // Only show media sources the user has permission for
        $user = $this->modx->getUser();
        if ($user != null && !$user->sudo) {
            $c->leftJoin('sources.modAccessMediaSource', 'ams', 'modMediaSource.id = ams.Target');
            $c->leftJoin('modUserGroupMember', 'ugm', "ams.principal = ugm.user_group");
            $c->where('ugm.member = ' . $user->id);
        }
```

modx_ms_bug

**Left**: Only one of the two valid media sources are available.
**Right**: Fixed version

OptimusCrime:
Fixes #12215, not tested.

### DIFF
--- a/core/model/modx/processors/source/getlist.class.php
+++ b/core/model/modx/processors/source/getlist.class.php
@@ -47,6 +47,15 @@ class modMediaSourceGetListProcessor extends modObjectGetListProcessor {
             $c->where(array('modMediaSource.name:LIKE' => '%'.$query.'%'));
             $c->orCondition(array('modMediaSource.description:LIKE' => '%'.$query.'%'));
         }
+        
+        // Only show media sources the user has permission for
+        $user = $this->modx->getUser();
+        if ($user != null && !$user->sudo) {
+            $c->leftJoin('sources.modAccessMediaSource', 'ams', 'modMediaSource.id = ams.Target');
+            $c->leftJoin('modUserGroupMember', 'ugm', "ams.principal = ugm.user_group");
+            $c->where('ugm.member = ' . $user->id);
+        }
+
         if ($this->getProperty('streamsOnly')) {
             $c->where(array(
                 'modMediaSource.is_stream' => true,


OptimusCrime:
Bump

Mark-H:
The "not tested" remark there is scary @OptimusCrime :P

OptimusCrime:
I'll test it ;)

Mark-H:
@OptimusCrime any luck with your testing? The code looks solid from a code review and I'll do some testing of my own in a few, but it would help add a bit more confidence to the merge if the person submitting it did some testing with it too ;) To get this into 2.4 we'll need this merged today.

OptimusCrime:
Ah, soso sorry. Have not had the time to test this just yet, but the code was tested by the original author, see #12215.

If you don't have the time right away I can prioritize to test it this week or something?

Mark-H:
I've just tried to apply this patch, but it looks like it's causing none of my media sources to show up. 

My setup now has 21 media sources (all very basic file system ones), and one of them is locked down to users in the Administrators group. All the others have no ACLs. 

Before applying the patch, I get 19 results on the first page (20 per page, minus one that is locked down), and 1 on the second page. After the patch, I get none. :(

OptimusCrime:
Could you please assign this to me? I'll take a look at it

CheeseSucker:
I've posted a comment in the original bug report: https://github.com/modxcms/revolution/issues/12215#issuecomment-128085333

OptimusCrime:
Darn, I understand more of the issue addressed here. Good thing this was not merged :P

PR #12234, opened 2014-12-24
Unset modx.user.userGroups in joinGroup()

BobRay:
The user's userGroups $_SESSION variable is used by other methods to get the user's groups. It needs to be cleared in addition to userGroupNames.


pixelchutes:
@BobRay,

Would the same change need to be applied for [leaveGroup()](https://github.com/modxcms/revolution/blob/9d392dadada571ccfba352781a12464f9301c0b3/core/model/modx/moduser.class.php#L711-L737)?

Issue #8357, opened 2012-08-03
Cannot Sort by Template in Form Customization Profiles


modxbot:
nuwebstudios created Redmine issue ID 8357

It appears you cannot sort by 'Template' within the Form Customization Profiles view.

We currently run 3 web sites through multiple contexts under one installation and I foresee us having 20-30 profiles in the future. Our naming convention would provide a way to easily sort each site's templates alphabetically, but it seems to be not functioning as expected.

The expected result would be when you click on the header "templates" that it alternates all rows between 'ASC' and 'DESC'. This currently works for 'ID', 'Action', and 'Description'.

PR #12246, opened 2015-01-03
Allow sorting by template in FC profiles

whitebyte:
Related issue - #8357.

### DIFF
--- a/manager/assets/modext/widgets/fc/modx.grid.fcset.js
+++ b/manager/assets/modext/widgets/fc/modx.grid.fcset.js
@@ -34,6 +34,7 @@ MODx.grid.FCSet = function(config) {
             header: _('template')
             ,dataIndex: 'templatename'
             ,width: 150
+            ,sortable: true
         },{
             header: _('constraint')
             ,dataIndex: 'constraint_data'


whitebyte:
I wonder, why 2.4.0 milestone? The fix is so tiny and definitely will not break anything, so i thought i's suitable for 2.3.3.

rtripault:
Because that's a "feature", not a "bug fix".
But i might not correctly understand the "Semantic Versioning" and am open to being corrected.

whitebyte:
I think this PR should be treated as bugfix, because column sorting is a default feature of Ext grids, and exactliy this sorting is probably disabled by mistake (so it's actually a bug). However, I do not insist.

rthrash:
An improvement to existing functionality IMO falls into a patch release as an enhancement. Sorting already was working, just not well. If it had been a request to _add_ sorting that would be a new feature. It's my $0.02, not a mandate, but I'd definitely consider enhancements/fixes to existing things as patch-level worthy.

rthrash:
Acknowledging an important foundational principal from http://semver.org/:

```MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.```

It still seems to me that semver leaves room for discretion in this case, especially since it is about the public _API_. In this case it's about changing the UI to the public API, not the API of MODX itself:

```Minor version Y (x.Y.z | x > 0) ... MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.```

Note, that doesn't say _MUST_, and they make note that MUST/MAY are two very different beasts. It seems the author of semver further leaves room for interpretation on how to handle things. In a discussion about inadvertent breaking changes of an API:

```Use your best judgment. If you have a huge audience that will be drastically impacted by changing the behavior back to what the public API intended, then it may be best to perform a major version release, even though the fix could strictly be considered a patch release. Remember, Semantic Versioning is all about conveying meaning by how the version number changes. If these changes are important to your users, use the version number to inform them.```

Therefore, I lobby for UX/UI improvements that use existing/fixed APIs at a given minor level release would fall safely within the spirit of Semver and shouldn't delay releasing improvements until a minor release. If it requires new underlying functionality, it would rightfully belong in a minor release.

whitebyte:
From now, i'm not just a random person hanging around and "suggesting closing", i'm officially a core contributor! P

OptimusCrime:
Hurray. Gratz, mate!

PR #12254, opened 2015-01-10
Remove not needed PHP headers due to PHP nature preventing caching

rtripault:
### What does it do ?

Remove manually set PHP headers from manager.
### Why is it needed ?

Those headers are not needed since the default behavior of PHP is to prevent caching.
In case one might want to implement some custom caching in the manager, having those headers might make it harder...


Mark-H:
Caching in the manager sounds like a really bad idea :D But as you say, PHP already seems to do that by default even without those header() calls, so no objections to taking 'm out.

Issue #11531, opened 2014-06-17
Feed timeout when offline


alanpich:
__The problem:__
Trying to use the MODx manager when not connected to the internet is ludicrously slow, almost to the point of nervous breakdown.

__Reproduction Steps:__
Try to use the manager of a local installation while not connected to the internet

__My understanding of the issue__
After speaking to @opengeek the issue seems to be caused by the dashboard feeds trying to update themselves (on every request, not just the dashboard???). This needs optimisation, and preferably shifting this process to an asychronous task (AJAX request probably)

PR #12259, opened 2015-01-10
Check if host have an internet access before trying to retrieve RSS feeds for dashboard widgets

rtripault:
**NOTES** : 
- not tested (no local installation available)
- method `checkdnsrr` is not available before PHP 5.3 for Windows **ONLY** (see http://php.net/manual/en/function.checkdnsrr.php). If that is a critical issue, we might consider making use of `gethostbyname()`
- source : http://stackoverflow.com/questions/1696202/check-if-host-computer-is-online-with-php
### What does it do ?

Check if we can find any DNS record for google.com, to assume "server" have an internet access, before trying to load RSS feeds.
### Why is it needed ?

When being "offline", feeds take a huge time trying to fetch RSS feeds, resulting in slow manager dashboard.
### Related issue(s)/PR(s)
- Should take care of #11531

### DIFF
--- a/manager/controllers/default/dashboard/widget.modx-news.php
+++ b/manager/controllers/default/dashboard/widget.modx-news.php
@@ -17,6 +17,9 @@ class modDashboardWidgetNewsFeed extends modDashboardWidgetInterface {
      * @return string
      */
     public function render() {
+        if (!checkdnsrr('google.com', 'ANY')) {
+            return '';
+        }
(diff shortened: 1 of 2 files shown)


opengeek:
Should we add a `function_exists()` check to avoid fatal errors in < 5.3 on Windows? Just in case?

opengeek:
I added the function_exists() check.

Issue #3981, opened 2011-03-03
Unable to copy/paste from Reports | System Information page


ukwiz:
ukwiz created Redmine issue ID 3981

When trying to extract data from Reports | System Information page, only the field description part is copied, not the values.

rtripault:
I am closing this one as "duplicate" of #2516 & #2508.

Also, a PR has been sent (#12261) to allow copy/paste of the system information.

whitebyte:
Historical moment: with the closing of this issue, the number of open tickets no longer 4-digit number!

PR #12261, opened 2015-01-11
Some TLC for the system information page

### LINKED ISSUE #2516: Save to TXT or Copy to Clipboard for System Info
smashingred created Redmine issue ID 2516

It would be very helpful if you could easily "save to text" or "copy to clipboard" the output on the System Info page for posting into the forums or to the bugtracker.

rtripault:
### What does it do ?
- make use of `defaultType` to define the default xtype of a panel's items (prevent "duplicate" xtype declarations)
- use `displayfield` xtype instead of `statictextfield`
### Why is it needed ?

Mostly to allow copy/paste of the data
### Related issue(s)/PR(s)
- Fixes #2516

### DIFF
--- a/manager/assets/modext/sections/system/info.js
+++ b/manager/assets/modext/sections/system/info.js
@@ -27,109 +27,104 @@ var viewPHPInfo = function() {
 MODx.panel.SystemInfo = function(config) {
     config = config || {};
     var info = [{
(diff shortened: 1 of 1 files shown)


whitebyte:
I believe it also affects #3981

PR #12263, opened 2015-01-12
Make syncsite checkbox a system setting

ghost:
Changes to manager/controllers/default/resource/create.class.php and update.class.php, _build/data/transport.core.system_settings.php and core/lexicon/en/setting.inc.php to make the 'Clear Cache' checkbox when editing resources respect a system setting.

I didn't do this with elements because 1. elements aren't created and edited as often, especially once development is finished, and 2. the Clear Cache checkbox is right there on the same tab as the other element editing fields, making it much simpler to clear it as desired.


TheBoxer:
This is a duplicate PR for those sent for issue #11980

PR #12264, opened 2015-01-12
Grid head/content column width fix

Jako:
The grid head and content have different widths because the head has a border left and the content has no border. This is only the scss fix.

Wrong:
wrong

Right:
right


Jako:
There is another bug with the grid css. ExtJS hides a column by setting display: none and does not change the assignment of x-grid3-cell-first/-last cell during this. Since this is an ExtJS issue, the first visible column could get a border left with the current css. Sadly it is not simple solvable by styling :first-child pseudo class.

PR #12265, opened 2015-01-13
Update modx.js

rogaldh:
Prevent to get hash value at last parameter


rtripault:
Hi @proxyfabio and thanks for your contribution.

But i fear your proposal is duplicate of #11957...

proxyfabio:
Didn't notice that, sorry

rtripault:
NP, and thanks for reminding me this, i'll try to put the PR in the radar for upcoming 2.3.x release since it seems it does not matter only to me :)

Issue #6092, opened 2011-11-22
Translation for default email templates in the system settings


modxbot:
elastic created Redmine issue ID 6092

It would be nice if the translation will be added to the default message templates in the system settings.
Below is a list of system settings:
site_unavailable_message
emailsubject
forgot_login_email
signupemail_message
webpwdreminder_message
websignupemail_message

pepebe:
This is an old one and it is still nagging me.

Question to the more knowledgeable people here:

Would it be possible to add these messages to core/lexicon/de/mail.inc.php and core/lexicon/de/site.inc.php and use [[%forgot_login_email]] to get the right translation depending on the current cultureKey?

Cheers,

pepebe

rtripault:
Somehow related to #3969

pepebe:
@rtripault. You never sleep, don't you? Anyway, as far as I understand your PR this would be exactly what we need to pull multilingual forgot_login_email messages from the lexicon. Thanks.

## 2do list
I did a quick search for the required system settings (but I didn't look very hard at the code).

### site_unavailable_message
core/model/modx/modrequest.class.php 
### emailsubject
manager/controllers/default/security/login.class.php 
### forgot_login_email
manager/controllers/default/security/login.class.php (done)
### signupemail_message
core/model/modx/processors/security/user/create.class.php 
core/model/modx/processors/security/user/update.class.php
### webpwdreminder_message
Odd, I can only find this one to be used in system settings...
### websignupemail_message
Dito...

In addition all of these setting would have to be added to the lexicon and referenced in _build/data/transport.core.system_settings.php

Did I miss something obvious?

rtripault:
@pepebe i do sleep, do not worry :smile: 

The PR indeed only takes care of the `forgot_login_email`, (at least for the manager "forgot password" function), but the idea/code could indeed be replicated for some other settings (thank you for your list/hints so far).

pepebe:
@rtripault I've noticed your heroic efforts to beat some sense into the long list of issues. I'm sure you don't sleep that much. ;)

There will be a hackton in Cologne next month. Every German developer is effected by this in some way. I guess that many of us want to get rid of that beast. I'll add add it to the long list of things I want to solve during this event.

Mark-H:
With #12266 merged, this request is a step closer, at least for the forgotten login ;)

JoshuaLuckers:
It would be nice if the other items from the todo list would be available for translations before MODX 3 is released.

[...]

PR #12266, opened 2015-01-13
Make use of the parser to process 'forgot_login_email' system setting value

### LINKED ISSUE #3969: Use chunks for signup, forgot, reminder emails body
gadamiak created Redmine issue ID 3969

The content of signup, forgot and reminder emails is set in a textfield in system settings, which is not handy. The field is small and standard placeholder can't be used.

That would be much more usable to specify a chunk name with the mail body instead and allow to use all MODx magic available in chunks.

rtripault:
### What does it do ?

Make use of the parser to process `forgot_login_email` system setting value.
### Why is it needed ?

To : 
1. ease the edition of the email (ie. chunks might be more convenient)
2. allow i18n
3. use additional tags (snippets/chunks/setting...)
### To do before merge
- [x] make sure `unsetPlaceholders` will not break anything
### Related issue(s)/PR(s)
- Fixes #3969
- Related to #6092

### DIFF
--- a/manager/controllers/default/security/login.class.php
+++ b/manager/controllers/default/security/login.class.php
@@ -236,9 +236,13 @@ class SecurityLoginManagerController extends modManagerController {
             $placeholders['manager_url'] = $this->modx->getOption('manager_url');
             $placeholders['hash'] = $activationHash;
             $placeholders['password'] = $newPassword;
+            // Store previous placeholders
+            $ph = $this->modx->placeholders;
+            // now set those useful for modParser
             $this->modx->setPlaceholders($placeholders);
             $this->modx->getParser()->processElementTags('', $message, false, false);
-            $this->modx->unsetPlaceholders($placeholders);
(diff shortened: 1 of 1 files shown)


Mark-H:
Re your to do, I guess `unsetPlaceholders` could unintentionally remove placeholders set before the handleForgotLogin method that were overwritten. One hack-ish way I've prevented something like that in another project was something along these lines:

```` php
$originalPlaceholders = $this->modx->placeholders();
$this->modx->setPlaceholders($placeholders);
$this->modx->getParser()->processElementTags(...);
$this->modx->placeholders = $originalPlaceholders;
````

rtripault:
Yes, that would do the trick i guess. Thanks

rtripault:
PR updated using the given recommendation.
Thanks

Mark-H:
Mucho gracias for a very cool enhancement :)

PR #12268, opened 2015-01-13
Add ability to search resources by template

13hakta:
Add template in condition list.

### DIFF
--- a/core/model/modx/processors/resource/search.class.php
+++ b/core/model/modx/processors/resource/search.class.php
@@ -39,6 +39,7 @@ class modResourceSearchProcessor extends modObjectGetListProcessor {
         $conditions = $this->getProperties();
         if (!empty($conditions['id'])) $where['id'] = $conditions['id'];
         if (!empty($conditions['parent'])) $where['parent'] = $conditions['parent'];
+        if (!empty($conditions['template'])) $where['template'] = $conditions['template'];
         if (!empty($conditions['pagetitle'])) $where['pagetitle:LIKE'] = '%'.$conditions['pagetitle'].'%';
         if (!empty($conditions['longtitle'])) $where['longtitle:LIKE'] = '%'.$conditions['longtitle'].'%';
         if (!empty($conditions['introtext'])) $where['introtext:LIKE'] = '%'.$conditions['introtext'].'%';


pepebe:
I'm working on a more versatile searchbar. If you want to participate, you can find it here:

https://github.com/modxcms/revolution/issues/12283

Cheers

rtripault:
While this would be valid, my understanding is that this processor is only used when on the "search" controller (`?a=search`), which would need to display a `template` field, in order to pass the "template condition".

pepebe:
We are talking about the uberbar search, right? This could be a simple use case for the extended search.

:r:template:1
(Find resources with template 1)

If this is not specific enough, we could use the extended json search

:r:json:{'template':'3','AND:content':'*xxx*'}
(find all resources with template 3 where the string xxx can be found inside content)

To digest this kind of search we either need to extend the search processor or build a new one from scratch.

I'd really like to see this happen.

rtripault:
My understanding is that we are not talking about the "uberbar", since `resource/search` processor is only used in the [search controller](https://github.com/modxcms/revolution/blob/master/manager/controllers/default/search.class.php) (see [panel](https://github.com/modxcms/revolution/blob/master/manager/assets/modext/widgets/modx.panel.search.js#L222), which would need a "template" field for the processor changes to work).

Hope that makes sense.

pepebe:
Ok, this is embarrassing. I just did a quick look at the path and confused the processor with something else. Can we forget about this?

Mark-H:
This can still be useful if someone is doing $modx->runProcessor('resource/search') so I'm proposing to merge this into 2.4, but I agree with @rtripault that it could be useful to also expose this in the Search page of the manager. That could be another pull request though.

@13hakta in order to merge this, we will need to have a Contributors License Agreement on file for you. This is as simple as going to http://develop.modx.com/contribute/cla/ and filling in the form. If you did that, please let me know your modx.com username (if it's different from your github username) and I'll get this merged.

13hakta:
Hello! My modx login is vchekryzvev and I've accepted contributor agreement.

Mark-H:
Thank you @13hakta for your contribution and making MODX better, one line of code at a time ;)

PR #12276, opened 2015-01-17
clean "onAjaxException" to remove full HTML document tag to avoid breaking manager

goldsky:
**### What does it do ?**
Clean returning ajax message by removing full HTML contents from server.

``` html
<xml ...>
<!DOCTYPE html ...>
<html>
    <header>
         <!-- html headers -->
    </header>
    <body>
         <!-- error body -->
    </body>
</html>
```

**### Why is it needed ?**
Currently, if server returns FATAL ERROR on an AJAX request, the pop up error window also breaks Manager, because it resets HTML headers.


rtripault:
While not tested, i'm not seeing anything wrong here, so :+1:

PR #12277, opened 2015-01-17
Add "Google Cloud Storage" media source option

goldsky:
**### What does it do ?**
Add Google Cloud Storage media source into core

**### Why is it needed ?**
We need to add optional media source.
Google Cloud Storage is a very good option.


goldsky:
After discussion with john, this is going to be an extra instead.

PR #12279, opened 2015-01-18
Children must be composites

Fi1osof:
I have changed only this (moved to aggregates):
'Children' =>
- array (
- 'class' => 'modUserGroup',
- 'local' => 'id',
- 'foreign' => 'parent',
- 'cardinality' => 'many',
- 'owner' => 'local',
- ),

Other - whitespaces.

### DIFF
--- a/core/model/modx/mysql/modusergroup.map.inc.php
+++ b/core/model/modx/mysql/modusergroup.map.inc.php
@@ -148,6 +148,14 @@ $xpdo_meta_map['modUserGroup']= array (
       'cardinality' => 'many',
       'owner' => 'local',
     ),
+    'Children' => 
+    array (
+      'class' => 'modUserGroup',
+      'local' => 'id',
+      'foreign' => 'parent',
+      'cardinality' => 'many',
+      'owner' => 'local',
+    ),
   ),
   'aggregates' => 
   array (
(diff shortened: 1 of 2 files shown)


TheBoxer:
@Fi1osof can you please update this PR to follow https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md

Fi1osof:
### What does it do ?
Set child relations as composites

### Why is it needed ?
Without this, children does not will remove on parent removeing.

@TheBoxer OK?

rtripault:
Thanks @Fi1osof ,

My only concern so far are : 

* xml schema have not been updated (mssql AND mysql)
* you should have used https://github.com/modxcms/revolution/blob/master/core/model/schema/build.modx.php to build the map (which should have prevented that whitespace issue)

I will leave others the pleasure to discuss about the idea of removing "nested" user groups (while i understand the reason/idea, i'm not sure that is something "safe" and would probably better offer an option to deleted "children" when removing a parent instead)

Fi1osof:
@rtripault created #12323

TheBoxer:
@Fi1osof Why you created another PR (and again without template) instead of updating this one?

Fi1osof:
Because I do not know how to do it. Should i stop to send PRs?

TheBoxer:
No, just please try to learn how to do it properly.

To update PR, just push again to the branch you created, it will automatically appear here.

Fi1osof:
Is now OK?

TheBoxer:
Yup, thanks for the update.

Fi1osof:
Not at all!

PR #12282, opened 2015-01-20
Some ui changes to uber searchbar

pepebe:
- added search by a single field name

```
id::31,
pagetitle::home, 
pagetitle:LIKE::%sch%
class_key::mgResource, 
template::3
```
- added id to search result, 
- omitted mdash after page title.

bildschirmfoto 2015-01-20 um 20 48 18

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1106,6 +1106,15 @@ $settings['modx_browser_default_viewmode']->fromArray(array (
   'area' => 'manager',
   'editedon' => null,
 ), '', true, true);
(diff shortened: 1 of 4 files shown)


pepebe:
Might be worth thinking about adding JSON support for more complex searches. Not sure about RL application/use cases though...

pepebe:
I'll rewrite this PR bases on the discussion here: https://github.com/modxcms/revolution/issues/12283

As I hope to harvest some more ideas for a good solution, I'll wait for a couple of days more before I post anything new.

In the meantime, I'm not sure if I should completely withdraw the PR or let it simply stay like it is. Any advice?

PR #12287, opened 2015-01-21
Do not remove || from value

Fi1osof:
http://joxi.ru/n2Y87DVh0pRpm6

### DIFF
--- a/core/model/modx/processors/element/tv/renders/web/output/default.class.php
+++ b/core/model/modx/processors/element/tv/renders/web/output/default.class.php
@@ -6,10 +6,6 @@
 class modTemplateVarOutputRenderDefault extends modTemplateVarOutputRender {
     public function process($value,array $params = array()) {
         $value= $this->tv->parseInput($value);
-        if ($this->tv->get('type') == 'checkbox' || $this->tv->get('type') == 'listbox-multiple') {
-            // remove delimiter from checkbox and listbox-multiple TVs
-            $value= str_replace('||', '', $value);
-        }
         $o= (string) $value;
         return $o;
     }


rtripault:
The default output renderer basically just removes the separator, which could make the TV value unusable due to being "merged".

A quick workaround would be to use the "delimiter" renderer.

Fi1osof:
@rtripault, thanks for the tip! But i still think it`s bug. TV have value '123||323' and if i got raw data, it`s must be raw date '123||323', not '123323'.

rtripault:
I'm with you on that one (thinking that's a bug too), but i'm not sure what was the original intent behind this `str_replace`, so i was just adding my thought so it could be reviewed by others.

Fi1osof:
@rtripault i think all will be OK :)

opengeek:
I'm not sure I concur that this is a bug, but I do not want to change this without additional time to review the consequences.

rtripault:
Not sure either that is a bug.

After discussing the issue with @TheBoxer, we found a valid use case of the current behavior : holding MODX tags.

Let's say you have a multi select TV with options `My chunk==[[$chunk]]||My other chunk==[[$my-other-chunk]]`, selecting both chunks should produce a TV value like `[[$chunk]]||[[$my-other-chunk]]`.

In that situation, i doubt anyone would expect the separator (`||`) to be part of the output.

When needing a separator, the "delimiter" output type is what we need.

Suggesting to decline the proposal then.

Fi1osof:
@rtripault 
> In that situation, i doubt anyone would expect the separator (||) to be part of the output.

I`m a programmer and i expect there will be a raw value, because in my TV output property are setted "default", not "with convertion".

OptimusCrime:
Close?

Fi1osof:
As you wish.

whitebyte:
-1 for closing. Raw value should be "raw". @rtripault shouldn't you use something like `renderOutput` for mentioned case?

Fi1osof:
@whitebyte +1

rainbowtiger:
When you use "get", you'll get a rendered output. When you don't, you won't, as in your screen capture above. Same difference when you use $modx->resource->get('publishedon'), versus $modx->resource->published. I use the latter when I just want the exact timestamp in the database rather than then "formatted" version MODX sends out when using "get()".

Fi1osof:
@rainbowtiger Problem not in modResource::get/set. Problem in modResource::getTVValue

rainbowtiger:
I'm including "getTVValue" when I said "get" above. In your screen capture, "$modx->resource->related_goods" returns an array, including the raw value. That's what I'm talking about. $modx->resource->getTVValue() returns a processed version of the data, which seems to be how this is designed. So, it is "expected" that people will use the output tab when defining a TV to specific exactly what gets returned by getTVValue, unless those in the know have something different to say about this.

Fi1osof:
@rainbowtiger 

> Raw value should be "raw"

rainbowtiger:
The question is whether get() or getTVValue() are SUPPOSED to return raw values or not. get('publishedon') certainly doesn't, and getTVValue() doesn't either. I've had a hard time finding any definitive answer to this in the documentation. I'd love to hear the reasoning of any developers who have created MODX concerning the difference between $modx->resource->[fieldname] and $modx->resource->get([fieldname]), and the same for getTVValue() or any other "get"ting.

rainbowtiger:
Maybe for TVs we need another option for output, "raw".

OptimusCrime:
This PR is from january of 2015. Time to decide what to do with this.

PR #12295, opened 2015-01-26
Correct syncsite behavior when updating a resource

### LINKED ISSUE #12275: syncsite_default is not honored. When enabled, Resource edits no longer clearing cache.
### Summary
A recent commit 4bb7ec2fb9721f5a6eaa16671e674735d8aeff92 seems to be causing issues when testing the new `syncsite_default` setting. When enabled, the "Empty Cache" checkbox is not automatically checked, requiring manually clearing the cache after Resource edits.

### Step to reproduce
1. Checkout `master` at least to 1c1fc58157d4591e4170cdddbe794ec4b9826109
2. Install instance and ensure `syncsite_default` is enabled
3. Edit an existing, cacheable resource. Click "Save", then "View"
4. Any changes are not immediately visible.

### Observed behavior
"Empty Cache" checkbox was **not** checked (under "Settings" tab.) Manually clearing the cache was required. 

### Expected behavior
When `syncsite_default` is enabled, the "Empty Cache" checkbox should be automatically checked while editing a Resource.

### Environment
MODX 2.3.3-dev (git, latest master), Chrome 41.0.2272.3 dev (64-bit) / OS X 10.10.1

pixelchutes:
Fix submitted by @sottwell

Resolves #12275

### DIFF
--- a/manager/controllers/default/resource/update.class.php
+++ b/manager/controllers/default/resource/update.class.php
@@ -119,7 +119,7 @@ class ResourceUpdateManagerController extends ResourceManagerController {
         $this->resourceArray['cacheable'] = intval($this->resourceArray['cacheable']) == 1 ? true : false;
         $this->resourceArray['deleted'] = intval($this->resourceArray['deleted']) == 1 ? true : false;
         $this->resourceArray['uri_override'] = intval($this->resourceArray['uri_override']) == 1 ? true : false;
-        $this->resourceArray['syncsite'] = isset($this->resourceArray['syncsite']) && intval($this->resourceArray['syncsite']) == 1 ? true : $this->context->getOption('syncsite_default', 1, $this->modx->_userConfig);
+        $this->resourceArray['syncsite'] = intval($this->context->getOption('syncsite_default', 1, $this->modx->_userConfig)) == 1 ? true : false;
         if (!empty($this->resourceArray['parent'])) {
             if ($this->parent->get('id') == $this->resourceArray['parent']) {
                 $this->resourceArray['parent_pagetitle'] = $this->parent->get('pagetitle');


manager/controllers/default/resource/update.class.php
```diff
@@ -119,7 +119,7 @@ public function process(array $scriptProperties = array()) {
         $this->resourceArray['cacheable'] = intval($this->resourceArray['cacheable']) == 1 ? true : false;
         $this->resourceArray['deleted'] = intval($this->resourceArray['deleted']) == 1 ? true : false;
         $this->resourceArray['uri_override'] = intval($this->resourceArray['uri_override']) == 1 ? true : false;
-        $this->resourceArray['syncsite'] = isset($this->resourceArray['syncsite']) && intval($this->resourceArray['syncsite']) == 1 ? true : $this->context->getOption('syncsite_default', 1, $this->modx->_userConfig);
+        $this->resourceArray['syncsite'] = intval($this->context->getOption('syncsite_default', 1, $this->modx->_userConfig)) == 1 ? true : false;
```
rtripault: My understanding is that this won't take care of form customization.
You might want to check my proposed [PR](https://github.com/modxcms/revolution/pull/12218) (which "wrong behavior" was trying to get the resource context while it was already available, see https://github.com/modxcms/revolution/issues/12275#issuecomment-71428150)

sottwell: I know nothing at all about form customization.

rtripault: Once again, this is just my understanding of the code : 

1. we grab the `modResource` object, and make it available as an [array](https://github.com/modxcms/revolution/blob/master/manager/controllers/default/resource/update.class.php#L110)
2. then, we check for [form customization rules](https://github.com/modxcms/revolution/blob/master/manager/controllers/default/resource/update.class.php#L111-112), which allow to "override" some fields values (`syncsite` being one fo those)
3. then, we [process](https://github.com/modxcms/revolution/blob/master/manager/controllers/default/resource/update.class.php#L114-121) the array data so it become usable in ExtJS

Again, my interpretation is that overridden values in FC should have an higher priority (be used/applied) than settings.

sottwell: I certainly can't argue with that. The thing is, the other settings are saved in the database and get put into the resourceArray. What this block of code is doing is to change them from 1 and 0 to true and false. The syncsite setting is not already in the resourceArray since it's not saved in the database, but it still has to be converted from the integer values in the system setting.

TheBoxer:
Should be fixed in d38d91177b38611b259430c8dd313f9eb61e584a.

PR #12298, opened 2015-01-26
Update package.windows.js

bertoost:
Make it possible to execute Javascript on update of the setup-options window. Usefull to make the things look right...

### DIFF
--- a/manager/assets/modext/workspace/package.windows.js
+++ b/manager/assets/modext/workspace/package.windows.js
@@ -167,7 +167,7 @@ MODx.window.SetupOptions = function(config) {
 };
 Ext.extend(MODx.window.SetupOptions,MODx.Window,{
 	fetch: function(content){
-		Ext.getCmp('modx-setupoptions-form').getForm().getEl().update(content);
+		Ext.getCmp('modx-setupoptions-form').getForm().getEl().update(content, true);
 	}
 
 	,install: function(btn, ev){


rtripault:
Hi Bert,

I'm not sure to understand the problem you are solving, would you mind providing a simple example of what's achievable with your proposal, that is not currently possible ?

When i looked at Sencha documentation, i saw nothing about that second argument in [Ext.Element#update](http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.Element-method-update), but it appears there is one for [Ext.Component#update](http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.Component-method-update). Unfortunately, i'm still not certain of its usage, so your enlightenment would be welcomed!

Thanks

Also, for what is worth, there's now a [contribution guidelines](https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request) you might want to check ;)

bertoost:
With this you're able to fire some javascript from your setup options. For example to make the window a bit wider or higher or add elements to the form panel?
The current setup-window is terrible.

Even better would be to support an array with xtypes or such to provide fields for the setup-options window.

rtripault:
Thanks, i really see the benefit of being able to stick with JS/ExtJS, but i still have not the picture on how that would be doable. Is the first argument becoming a callback ? 
ie.
```JS
update(function(form) {
    form.setWidth(750); 
});
```

bertoost:
Maybe.. I have tried to make it work with an array.. so "content" should be checked.
If it is a string then, use the current method (with true as second parameter) and otherwise assume that it is an array with xtypes :)

Mark-H:
Summary of what this does (mostly for myself): the second parameter allows scripts to be executed per http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.Component-method-update, so script tags in the setup options window will be parsed to make potentially very useful setup options windows.

Mark-H:
Thanks for the contribution @bertoost, sorry it took so long to get merged ;)

Merged into develop for inclusion in 2.4 in ed7711aa8c1ac920a95db325af365c38602bbdfd

PR #12317, opened 2015-02-01
Anonymous Username

ghost:
Use System Setting for anonymous user username

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1267,8 +1267,7 @@ class modX extends xPDO {
             $this->user = $this->newObject('modUser');
             $this->user->fromArray(array(
                 'id' => 0,
-                'username' => $this->config['anonymous_username']
-            ), '', true);
+                'username' => $this->getOption('anonymous_username', null, '(anonymous)');
         }
         ksort($this->config);
         $this->toPlaceholders($this->user->get(array('id','username')),'modx.user');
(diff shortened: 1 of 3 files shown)


rtripault:
Hi Susan, and thanks for your contribution.
Just to let you know we recently added some [contribution guideline](https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request), and while your contribution code was easy to read/understand, we would appreciate if you could follow those guidelines. Much thanks.

core/model/modx/modx.class.php
```diff
@@ -1267,7 +1267,7 @@ public function getUser($contextKey= '',$forceLoadSettings = false) {
             $this->user = $this->newObject('modUser');
             $this->user->fromArray(array(
                 'id' => 0,
-                'username' => '(anonymous)'
+                'username' => $this->config['anonymous_username']
```
rtripault: Wouldn't it be "safer" to use `$this->getOption('anonymous_username', null, '(anonymous)')` in case the system setting is deleted ? (which would provide a "default" value)

sottwell: Undoubtedly.

TheBoxer: Can you please change the PR to follow Romain's suggestion?

sottwell: I don't know how.

TheBoxer:
Closing, as #12418 was merged.

PR #12322, opened 2015-02-02
Fixing tooltip full viewport width

Jako:
### What does it do ?

Two css rules are added. max-width: 400px; & min-width: 200px;
### Why is it needed ?

The Tooltip width i.e. in a MODX.window is set by .x-tip { width: auto !important; /\* override ExtJS inline width */  } to auto. Because of that a tooltip has full viewport width if there is enough text in it. This is a very bad UX.

As quick fix there should be a max-width (and a min-width) set.

### DIFF
--- a/_build/templates/default/sass/_xtheme-modx.scss
+++ b/_build/templates/default/sass/_xtheme-modx.scss
@@ -538,6 +538,8 @@ td.x-grid3-hd-menu-open .x-grid3-hd-inner {
   border-radius: $borderRadius;
   padding: 5px;
   width: auto !important; /* override ExtJS inline width */
+  max-width: 400px;
+  min-width: 200px;
 }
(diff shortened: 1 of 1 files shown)


Jako:
The tooltip width should be limited somehow. Is there any reason why the calculated width of ExtJS does not fit?

rtripault:
Hi, and thanks for your contribution. We recently added [contribution guidelines](https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request), would you mind reformatting yours following those recommendations please ? Thanks.

Also, about your question, my understanding is that ExtJS will force the width/height it needs (using inline style), overriding any stylesheet (which make it fairly complicated to style).

Jako:
Done.

PR #12335, opened 2015-02-07
Make handlers on menu actions be applied together with normal actions

bertoost:
### What does it do ?

Changed the logout menu item using the security/logout action. The handler is also be applied on menu items, together with the action. When the JS handler is returning false, it will not fire the action request URL at all.
### Why is it needed ?

When having JS issues, you cannot logout the manager anymore. Which is necessary in some cases. This will prevent (at least) the logout menu item to work.
### Related issue(s)/PR(s)

https://github.com/modx-ccc-2015/whishlist/issues/65


bertoost:
Closed because doing it the right way from my own fork :$ sorry! Will be back soon...

PR #12337, opened 2015-02-07
Add regexp validation into text TV

pepimpepa:
### What does it do ?

It is adding regexp validation to text TV
### Why is it needed ?

E.g. allow only invoice numbers like "2015-0001".


TheBoxer:
Merged to develop.

PR #12339, opened 2015-02-07
The DIV creates creates a duplicate ID and is not needed anayway

mindeffects:
The DIV has the same ID as the input element:
https://github.com/modxcms/revolution/blob/master/manager/assets/modext/widgets/element/modx.panel.tv.renders.js#L30

Deleting the DIV has no effect when editing the resource.


mindeffects:
Sorry for creating this dupe of #12338. I am doing my first steps with git/GitHub and I am afraid that this will not be my last bloody nose. ;-)

PR #12340, opened 2015-02-07
Conding standards: editorconfig file #64

pixelfanatiker:
A editorconfig file for the project for better codeformatting consistency.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+tab_width = 4
+
+[**.js]
+tab_width = 4
+
+[**.coffee]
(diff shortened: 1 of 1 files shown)


Mark-H:
Thanks for your contribution @pixelfanatiker, sorry it took so long to merge.

Mark-H:
While reviewing editorconfig, I also noticed it supports specifying the charset - what do you think about also adding that one to utf-8?

rtripault:
Could not harm having the charset in. Also, wondering why all "overriding" some values which are already defined in the main wildcard, furthermore for file types not used in the project ?

```
root = true

[*] <- this is for all files
end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true
indent_style = space
tab_width = 4

[**.js]
tab_width = 4 <- same as [*]

[**.coffee] <- no coffeescript files in Revo
...
```

Issue #12289, opened 2015-01-22
modObjectCreateProcessor misses saveObject method


Jako:
Maybe it is not really necessary and could be done by beforeSave method, but it is a different behaviour between modObjectCreateProcessor and modObjectUpdateProcessor.

PR #12343, opened 2015-02-08
Feature/equalize object processor

bertoost:
### What does it do ?

It updates a couple of modObject processors to use saveObject and removeObject etc. to make the flow the same as the modObjectCreateProcessor does.
### Why is it needed ?

Not really necessary, but it will keep things the same.
Also possible to use afterSave and that kind of methods, but now you can actually overwrite the savings itself.
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/issues/12289


bertoost:
Damn.. wrong branch

Issue #11848, opened 2014-07-29
modx->sendRedirect() no longer works during OnManagerLogout event


pixelchutes:
If you'd like to redirect users to a custom page once they logout of the MODX Manager, previously you could create a simple Plugin bound to `OnManagerLogout` with a `logoutURL` property of where to send them:

``` php
<?php
$e =& $modx->event;
switch( $e->name ) {
    case 'OnManagerLogout':
        if( isset( $logoutURL ) ) {
            $modx->sendRedirect( $logoutURL );
            exit;
        }
        break;
}
```

However, this approach no longer seems to work... I believe this could be related to changes introduced in #11650

Mark-H:
With #12344 merged this works for non-XHR requests (i.e. the fallback if javascript in the manager is broken), but not with the normal ajax redirect. 

A way to fix this might be to update the MODx.logout() function:

```` js
    ,logout: function() {
        if (this.fireEvent('beforeLogout')) {
            MODx.msg.confirm({
                title: _('logout')
                ,text: _('logout_confirm')
                ,url: MODx.config.connector_url
                ,params: {
                    action: 'security/logout'
                    ,login_context: 'mgr'
                }
                ,listeners: {
                    'success': {fn:function(r) {
                        if (this.fireEvent('afterLogout',r)) {
                            location.href = './';
                        }
                    },scope:this}
                }
            });
        }
    }
````
to instead of redirecting to './' redirects the user to a link that was provided by the processor, like 

```` js     ,logout: function() {
        if (this.fireEvent('beforeLogout')) {
            MODx.msg.confirm({
                title: _('logout')
                ,text: _('logout_confirm')
                ,url: MODx.config.connector_url
                ,params: {
                    action: 'security/logout'
                    ,login_context: 'mgr'
                }
                ,listeners: {
                    'success': {fn:function(r) {
                        if (this.fireEvent('afterLogout',r)) {
                            location.href = r.data.logoutUrl;
                        }
                    },scope:this}
                }
            });
        }
    }
````

To do that, the security/logout processor and possibly the plugin will need to be updated to allow setting a logoutUrl. Maybe by looking at the output from `$this->modx->invokeEvent` (so that the plugin would need to call `$modx->event->output('url/to/redirect/to')`)? 

[...]

PR #12344, opened 2015-02-08
Make handlers on menu actions be applied together with normal actions

bertoost:
### What does it do ?

Changed the logout menu item using the security/logout action. The handler is also be applied on menu items, together with the action. When the JS handler is returning false, it will not fire the action request URL at all.
### Why is it needed ?

When having JS issues, you cannot logout the manager anymore. Which is necessary in some cases. This will prevent (at least) the logout menu item to work.
### Related issue(s)/PR(s)

Should resolve https://github.com/modx-ccc-2015/whishlist/issues/65

### DIFF
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -403,6 +403,7 @@ $children[2]->fromArray(array (
   'description' => 'logout_desc',
   'parent' => 'user',
   'permissions' => 'logout',
+  'action' => 'security/logout',
   'handler' => 'MODx.logout(); return false;',
 ), '', true, true);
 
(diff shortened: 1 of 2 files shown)


pixelchutes:
IIRC, There was a reason this got changed to `javascript:`, something to do with client-side event handlers? _(Not sure really, have not personally used to my knowledge)_

That said, this PR may resolve #11848

Mark-H:
Nice one @bertoost, thanks for your contribution. Merged into 2.4.0-dev. 

@pixelchutes I don't think this resolves #11848 just yet; only in the case of javascript being broken/disabled. Will comment on #11848.

Issue #12239, opened 2014-12-30
Allow empty pagetitle in UI


inreti:
The manager UI does not allow you to save a resource without a pagetitle. Therefore it is not possible to write a plugin which automatically sets a pagetitle when saving (like with alias or other fields). 
There are cases where you may want to hide the pagetitle via forms customiziation from the editor (not forcing the editor to give a useless title just to be able to save it, e.g. in Collections where pagetitle is not used). But if you hide the pagetitle field via fc, saving will not be possible. 

So my proposal: strip the pagetitle verification from the UI. If a resource with empty pagetitle is saved, automatically choose a pagetitle (say "resource_[[+id]]"). If the id is included it is guaranteed that all pagetitles are unique (and so are the corresponding aliases).

TheBoxer:
Should be resolved via extra and not be part of the core.

PR #12347, opened 2015-02-08
Allowing empty resource pagetitle. Will be set automatically to "Untitled Resource {id}"

bertoost:
### What does it do ?

It checks if you submit empty pagetitles and when you do, it uses Lexicon "resource_untitled" concatenated with the new ID of the resource and also updates the alias when you create a new resource.
At update, it is checking the pagetitle again to, also updating the title, not the alias.
Also added a system setting "resource_allow_empty_pagetitle" to enable/disable this feature.
### Why is it needed ?

It's not really necessary but it's a cool feature to allow empty pagetitles at creation and updating. Especially for those who sets the pagetitle hidden in FC
### Related issue(s)/PR(s)

Should resolve https://github.com/modxcms/revolution/issues/12239

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2032,4 +2032,13 @@ $settings['auto_isfolder']->fromArray(array (
     'area' => 'site',
     'editedon' => null,
 ), '', true, true);
+$settings['resource_allow_empty_pagetitle']= $xpdo->newObject('modSystemSetting');
+$settings['resource_allow_empty_pagetitle']->fromArray(array (
+  'key' => 'resource_allow_empty_pagetitle',
+  'value' => false,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'manager',
+  'editedon' => null,
+), '', true, true);
 return $settings;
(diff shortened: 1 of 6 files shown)


Mark-H:
Hiya Bert, thanks for the contribution. 

In create.class.php there is now a preparePageTitle and setPageTitle method.. could that be consolidated? Or is it useful to have both?

TheBoxer:
IMO this should be an extra and not part of the core.

bertoost:
@Mark-H Because mine has a setting it's good to have both I think? And both are fired on different moments :$ setPageTitle() can be out, but I thought not to break existing features :-)

@TheBoxer If that's possible at all; since this is checked in core files? Then you should override the core files or such?

TheBoxer:
@bertoost I was thinking about two options:

1. OnDocFormPrerender plugin, which will set & hide the pagetitle field via JS
2. Combination of FC (hide pagetitle & set some default value - for example XXX) and OnDocFormSave plugin, which will set pagetitle to whatever you wish (GeneratedPagetitle 1)

bertoost:
@TheBoxer But you also need to set "allowBlank" to "false" then.. which of course can be handled with some JS magic...
I leave this up to you.. it's a nice to have and not a must I think :-)

TheBoxer:
In both options I wrote you don't have to worry about allowBlank field, because you'll set the value to that field (so it will not be empty after submitting the form)

TheBoxer:
Closing as it should be an extra :)

PR #12348, opened 2015-02-08
Add new option to tv date "hide time for user"

### LINKED ISSUE #9473: DateTV with input options
mindeffects created Redmine issue ID 9473

Sometimes I only need a date to be stored in a TV without a time – sometimes it's the other way around. Example: When I want to store the release day of a movie I just nedd the date and not the time, but it gets displayed anyway.

I wouldn't ask for that feature if my clients did not complain about this or, to be more precise, where confused and asked my what to enter here or if it was possible to hide the time. I had to say "No." – at least not possible without some "dirty plugin tricks". ;-)

Having the possibility to customize the backend appearence of dateTVs would be awesome, much more elegent and availale for every user – setting the time to a default value, like 00:00:00, or just don't show the time at all. Maybe there could be a date formatter string that controls the _visual_ appearance of dateTV?

A system like MODX should give its users this little interface freedom. Don't you think so, too? PLZ?!

cheers
Oliver

pepimpepa:
### What does it do ?

It is possible to hide time for users in tv date
### Why is it needed ?

It is uncomfortable for users in some cases
### Related issues

Resolves #9473

### DIFF
--- a/core/lexicon/en/tv_widget.inc.php
+++ b/core/lexicon/en/tv_widget.inc.php
@@ -127,6 +127,7 @@ $_lang['textareamini'] = 'Textarea (Mini)';
 $_lang['textbox'] = 'Textbox';
 $_lang['time_increment'] = 'Time Increment';
 $_lang['time_increment_desc'] = 'The number of minutes between each time value in the list (defaults to 15).';
+$_lang['hide_time'] = 'Hide time option for user';
 $_lang['title'] = 'Title';
 $_lang['upper_case'] = 'Upper Case';
 $_lang['url'] = 'URL';
--- a/manager/assets/modext/util/datetime.js
+++ b/manager/assets/modext/util/datetime.js
@@ -72,6 +72,7 @@ Ext.ux.form.DateTime = Ext.extend(Ext.form.Field, {
     ,maxTimeValue: null
     ,minTimeValue: null
     ,disabledDates: null
+    ,hideTime: false
 
 
     // {{{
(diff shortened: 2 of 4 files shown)


Mark-H:
Thank you for your MODXCCC contribution :D

bequadrat:
:+1:

OptimusCrime:
Neat!

mindeffects:
Yes, neat.

But it saves the current time! This makes date comparisons a little tricker (not toooo tough, I know ;-) ). If only the value for "earlies time" would be used instead of the current time ... Or just set it to 00:00:00 ... Well, next version maybe. ;-)

inreti:
Would be create to have on option to disable saving of the current time. Because at the moment you cannot have a single TV with the semantics (e.g. for an appointment): either have the time explicitly given or have no special time (for the whole day)...
From my point of view, the automatic injection of the current time is totally useless. I can't imagine right know any real use case for that...

sottwell:
Well, it’s easy enough to use an output modifier on the TV when outputting its value. Generally it has to be modified anyway to get the month-day-year to be displayed as desired.

As far as I know, the date/time widget being used is the one that came with the old version of ExtJS. I don't know if it can be easily modified or not.

mindeffects:
A "random" time is becoming a problem when using the api directly. Comparisons take the time into account. Sure, you can code something to make it work but why should you? What if you want to store  "time only" data? We need TVs for date, time and date-time. Or an intelligent combo type!

PR #12349, opened 2015-02-08
Beautify of the Manager errors (like resource with ID x isn't found)

bertoost:
### What does it do ?

I have changed how the error looks like.. Just a bit more stylish for the manager.
Also changed Lexicon for the "Resource not found with ID: x" to "Resource with ID x not found"
### Why is it needed ?

Because the old, non-styled is very ugly :grin: 
### Related issue(s)/PR(s)

Non related issues as far I know.


bertoost:
image

How it looks like...

bertoost:
And multiple even support...

image

Mark-H:
Thank you! :D

Little note that the missing controller (e.g. manager/?a=resource/updatesss) doesn't yet have this styling, but that's for a next PR.

PR #12350, opened 2015-02-08
Feature/manager css cleanup

DESIGNfromWITHIN:
Replaced RGBA color variables with HEX values.


Mark-H:
As discussed at the CCC due to the conflicts this one wont be merged but we'll sit down some other time to sort it out :D

PR #12353, opened 2015-02-08
Enhancement of branding in Setup

mindeffects:
### What does it do ?

Changes the logo PNG to SVG and removes the grayscale shade which makes the logo look dull
### Why is it needed ?

Because: NEVER TOUCH THE LOGO! :-D


OptimusCrime:
+1 for svg, but it is missing the copyright icon?

mindeffects:
Nope. 1:1 like the one from modx.com
Actually, it is identical because I took it from there. ;-)

OptimusCrime:
Ah, okey, I guess that should be okey then ;)

TheBoxer:
Cherry picked to master & merged to develop.

PR #12354, opened 2015-02-09
Namespace permissions

### LINKED ISSUE #11448: [2.3] Assign Namespaces to a Category
There are days in the life of a decent MODXpert, when he gets asked by a client to change some sort of "translation", "system setting" or whatever. That happens like every day. The usual response is (spoken out loud): "Sure, no thing." but honestly he/she thinks (silently): "C'mon, that's just SUCH a simple task that they should/could do themselves!". But clients cannot do this themselves because we don't dare to let them near "System Settings" or the "Lexicon(s)".

But what if Namespaces could be assigned to a Category? Wow, think of that: Clients changes those setting they are allowed to change, because they cannot break anything. They can now edit translations for those little asides headline you used in a carefully crafted chunk. Clients could change stuff just like they can change some of their user setting.

Just create a namespace e.g. "clientstuff" with an empty "core path" and an "assets path" pointing to "{assets_path}assets/clientstuff/" and a category "Client editable". Done. Now, when a client enters the "System Settings" he only sees the settings he is allowed to change. Same for the Lexicon: only the translations he should be able to do on his own are visible.

Yes, we all want that, right? Please comment and "vote" for this idea to become reality in MODX 2.3!

theboxer:
### What does it do ?

Add ability to specify namespace access (allows hiding namespaces from system settings / lexicons).
### Why is it needed ?

Currently it's not possible to limit access to namespaces for specific users. This should help with giving users access just to specific namespace, so they can't break site by changing wrong system setting.
### Screenshots

screen shot 2015-02-09 at 14 35 04
screen shot 2015-02-09 at 14 35 14
### Setup

Run _build/transport.core.php
Run setup
After those steps you should see **Hidden namespace** policy
### Related issues

Resolves #11448
Resolves modx-ccc-2015/whishlist#47

Review would be nice :)
@modxcms/integration-managers-revolution


mindeffects:
No integrator around for this super awesome next step for MODX? ;-)

OptimusCrime:
This is awesome. Great work! :D

PR #12359, opened 2015-02-11
Fix problem load ExtensionPackage from cache

Fi1osof:
### What does it do ?

Fix problem load ExtensionPackage from cache
### Why is it needed ?

Because we have errors: Problem getting service basket, instance of class Basket, from path

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -567,7 +567,8 @@ class modX extends xPDO {
                 $package['table_prefix'] = !empty($package['table_prefix']) ? $package['table_prefix'] : null;
                 $this->addPackage($package['namespace'],$package['path'],$package['table_prefix']);
                 if (!empty($package['service_name']) && !empty($package['service_class'])) {
-                    $this->getService($package['service_name'],$package['service_class'],$package['path']);
+                    $packagePath = str_replace('//','/',$package['path'].$package['name'].'/');
+                    $this->getService($package['service_name'],$package['service_class'],$packagePath);
                 }
             }
         }


Mark-H:
I've seen this error quite a bit with MoreGallery (which loads an extension package for the custom resource), but I can't seem to recreate it on a clean install.. maybe the error has something to do with upgrades? Your commit mentions the cache, maybe you remember the steps needed to trigger the issue? If you don't remember, I don't blame you, this PR has been here for a while..

Fi1osof:
@Mark-H Maybe fixed in last version. After 2.3.3 publishing many people reports this. http://modxclub.ru/topics/oshibka-podgruzki-klassov-1566.html

Mark-H:
Hmmm so 2.3.3... maybe related to this https://github.com/modxcms/revolution/commit/ceac1177d88abeb0b3621076d6e9014d36e5d275 which was pull #11927

Fi1osof:
Yes, it seems.

Fi1osof:
Problem exists after update from modx2.2 to modx2.3

Fi1osof:
I understood. Problem exists when setted serviceName and serviceClass, for example:
```json
{"shopModx":{"serviceName":"shopModx","serviceClass":"shopModx","path":"[[++core_path]]components/shopmodx/model/"}}
```
Try to install [shopModx](http://modx.com/extras/package/shopmodx)

Fi1osof:
@Mark-H, hi! 
Is this will merged?
I tired to fix this on every site. http://joxi.ru/ZrJVWD0u1kLgPr
PR created 11 Feb 2015. Almost a year ago.

christianseel:
I have an extension package specified with serviceName and serviceClass in a project and I don't get those error messages. So maybe there's something wrong with your configuration anywhere @Fi1osof?

This is my example:
```
    {
        "phpconsole": {
            "serviceName": "phpconsole",
            "serviceClass": "phpconsoleX",
            "path": "[[++core_path]]components/phpconsole/model/"
        }
    }
```
…and the class file is located at `/core/components/phpconsole/model/phpconsole/phpconsolex.class.php`

Fi1osof:
@christianseel MODX version?

christianseel:
@Fi1osof 2.4.2-pl

Fi1osof:
@christianseel i use 2.4.2 too.
I tested this bug many times and know this is bug, not my error (and not only i catch this error).
My config {"basket":{"serviceName":"basket","serviceClass":"Basket","path":"[[++core_path]]components/basket/model/"}}

It`s happens only on MODX init. Then OK.

christianseel:
What do you mean by "MODX init"? Loading frontend with an empty cache?

Could you provide full steps to reproduce the bug? Because for me it looks like https://github.com/modxcms/revolution/pull/11927 should have fixed those issues. I remember I had similar issues with my phpconsole package, was that was because of incorrect paths and configs and not a bug in MODX.

Fi1osof:
@christianseel @Mark-H i finded mistake. 
It`s happends when extension_packages stored in database. http://joxi.ru/8AnXDv9uqKB7Zm
I dont know how they got there. modX::addExtensionPackage() writes sistem settins only, do not create modExtensionPackage objects.

modX::_loadExtensionPackages() [get modExtensionPackages](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L564), and if got, load them. And those have broken pathes. 

@opengeek, why this for?

christianseel:
Great! Maybe @rtripault can clarify here, since he was also doing #11927 which dealed with extension packages in the database.

rtripault:
I unfortunately won't be able to provide more details on why the extension packages are stored in a dedicated table (besides it's happening when migrating from Revo 2.2, see https://github.com/modxcms/revolution/blob/2.x/setup/includes/upgrades/common/2.3-extension-packages.php).
My best "bet" (other than those [comments](https://github.com/modxcms/revolution/blob/2.x/core/model/schema/modx.mysql.schema.xml#L764)) on this decision is the limitation of `modSystemSetting.value` size (text type) in case there are loads of packages.

In fact, i feel like the current implementation of `modExtensionPackage` (which is not complete, as @Fi1osof [noticed](https://github.com/modxcms/revolution/pull/12359#issuecomment-180780998)) is a "regression", since we are not able to define the order of loaded packages/services since they are loaded by namespace name (see https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modcachemanager.class.php#L448) while, with some trick (ordering the JSON data stored in the system setting), we could influence that order using the system setting implementation.

Hope that will help a bit anyway

Fi1osof:
@rtripault 

> since we are not able to define the order of loaded packages/services since they are loaded by namespace name (see https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modcachemanager.class.php#L448) while, with some trick (ordering the JSON data stored in the system setting), we could influence that order using the system setting implementation.

I wrote this too almost 3 years ago http://joxi.ru/DrlaPn9i4nQodm

It`s sad that ExtensionPackages functionality is not centralized

OptimusCrime:
This PR is over two years old. Can it be closed?

PR #827, opened 2014-01-29
Fatal error

Fi1osof:
Fatal error: Cannot redeclare class modTemplateVarInputRenderText in /var/www/evakuator.ru/public_html/core/model/modx/processors/element/tv/renders/mgr/input/text.class.php on line 10

OR add if(class_exists) here: https://github.com/modxcms/revolution/blob/develop/core/model/modx/modtemplatevar.class.php#L420
like here: https://github.com/modxcms/revolution/blob/develop/core/model/modx/modtemplatevar.class.php#L445

PR #12369, opened 2015-02-19
Fatal error: Cannot redeclare class modTemplateV..

Fi1osof:
Related to [#827](https://github.com/modxcms/revolution/pull/827)

### DIFF
--- a/core/model/modx/processors/element/tv/renders/mgr/input/text.class.php
+++ b/core/model/modx/processors/element/tv/renders/mgr/input/text.class.php
@@ -3,9 +3,11 @@
  * @package modx
  * @subpackage processors.element.tv.renders.mgr.input
  */
-class modTemplateVarInputRenderText extends modTemplateVarInputRender {
-    public function getTemplate() {
-        return 'element/tv/renders/input/textbox.tpl';
+if(!class_exists('modTemplateVarInputRenderText')){
+    class modTemplateVarInputRenderText extends modTemplateVarInputRender {
+        public function getTemplate() {
+            return 'element/tv/renders/input/textbox.tpl';
+        }
     }
-}
+} 
 return 'modTemplateVarInputRenderText';


TheBoxer:
Hi @Fi1osof. 
Thanks for trying to help with this issue. From my understanding, your PR simply hides the error, but doesn't solve the underlying problem. Can you confirm?

Fi1osof:
@TheBoxer you understood it wrong. This PR excludes error. Hide error - when error exsists. Here error not exsists. Is there not class_exists in MODX? Check it here: https://github.com/modxcms/revolution/search?utf8=%E2%9C%93&q=class_exists&type=Code

TheBoxer:
Let's step back,
can you please update the PR to follow contributing guide and provide steps how to reproduce this problem?

I don't consider wrapping a class in class_exists method as a proper solution (and you will not find that in the link you provided), because in an edge case you can do that with all classes to prevent their in-proper inclusion.

Fi1osof:
@TheBoxer Looks like already fixed. IF note mistake again, I will inform you.

TheBoxer:
Ok, closing then. Reopen if you'll need to.

Fi1osof:
@TheBoxer OK

Fi1osof:
I have error again. I think it`s because text.class.php is included for different methods [if class is not found] (https://github.com/modxcms/revolution/blob/master/core/model/modx/modtemplatevar.class.php#L448-L450).
MODX 2.2.16
I think it`s actual for MODX 2.3 too

pixelchutes:
Interesting, does changing to `include_once` solve the issue for you?

Fi1osof:
@pixelchutes Not, because $TV will not know $className without including (https://github.com/modxcms/revolution/blob/master/core/model/modx/modtemplatevar.class.php#L449). 
In modTemplateVar exists method [checkForRegisteredRenderMethod](https://github.com/modxcms/revolution/blob/master/core/model/modx/modtemplatevar.class.php#L466), but i think it`s work not correctly.

PR #12371, opened 2015-02-19
Auto suggest setting key in contexts and users settings

### LINKED ISSUE #12325: [Feature] Auto-suggest for context settings keys
### Summary
I run into this for years and thought I'd fix it in an issue to eventually address it: The key field for a context setting should be auto-suggesting existing keys from system settings when typing, this would prevent annoying copy and paste marathons when creating a new context and adding necessary settings to run it (for example as an independent site with own domain).

### Step to reproduce
Create a new context and add some basic context settings

### Observed behavior
We have to copy and paste (or remember) the setting key when adding it to a context

### Expected behavior
When starting to type in the "key" field a list of existing system setting keys is presented.

### Environment
MODX Revolution

theboxer:
### What does it do ?

Changes input type to select box with auto suggestion support based on selected namespace in user's and context's settings.
### Why is it needed ?

Prevent user error in copying system settings keys.
### Related issue(s)/PR(s)

Resolves #12325


OptimusCrime:
This is really cool. Nice work mate!

PR #12396, opened 2015-03-24
Set default background color for thumbs to WHITE instead of BLACK

mindeffects:
The default background color for thumbs of "PNGs with transparency" used to be black, which does not play well together with the white background of the resource editing page. Black is NEVER the right choice.

image

Another solution would have been to change the default filetype $f="png", but this would lead to bigger thumbnail files.

### DIFF
--- a/core/model/phpthumb/phpthumb.class.php
+++ b/core/model/phpthumb/phpthumb.class.php
@@ -39,7 +39,7 @@ class phpthumb {
 	var $sh   = null;     // Source crop Height
 	var $zc   = null;     // Zoom Crop
 	var $bc   = null;     // Border Color
-	var $bg   = null;     // BackGround color
+	var $bg   = "ffffff"; // BackGround color
 	var $fltr = array();  // FiLTeRs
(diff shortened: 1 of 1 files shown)


Mark-H:
Nice one Oliver :D 

I'm just wondering if it might be better to add this to the modPhpthumb class instead? Given the file you've edited may get overwritten when the core is updated to a newer version.

mindeffects:
Thanks for the hint (and the flowers), @Mark-H . I didn't know/think about this and went for the injection point I found first.

If there is a better place to add the white background and make it update-safe, it would be the wise to choise this one instead.

But I might struggle with finding it. :-D

mindeffects:
Wait, @Mark-H , you ment this one?
https://github.com/modxcms/revolution/blob/master/core/model/phpthumb/modphpthumb.class.php

mindeffects:
With the help of @Mark-H I replaced this pull with an update save one:
https://github.com/modxcms/revolution/pull/12411

PR #12411, opened 2015-04-10
Set default background color for thumbs to WHITE instead of BLACK v1.1

mindeffects:
The default background color for thumbs of "PNGs with transparency" used to be black, which does not play well together with the white background of the resource editing page. 
Black is NEVER the right choice for a thumb's background. ;-)

cb9fc740-d26f-11e4-8f29-c4696b795b02

Another solution would have been to change the default filetype $f="png", but this would lead to bigger thumbnail files.


pixelchutes:
My personal preference is for alternating grid of white/gray (to help indicate transparency), but until then, `#ffffff` > `#000000` :smile:

mindeffects:
The most elegant way would be to create PNG thumbs with alpha transparency and a selectable color or pattern for the background via system settings. But until then, indeed: #fff wins. ;-)

PR #12415, opened 2015-04-13
Implement new use_frozen_parent_uris option to respect frozen parent URIs in generating child URIs

Mark-H:
### What does it do ?

Previously, generating the uri for a child resource would go through each of the parents, grab it alias, and stick that together to form the URI at which the child is accessible. 

With this patch, disabled by default through use_frozen_parent_uris setting, the uri generating for child resources will still go through each of the parents to build the uri. The difference is that it now checks if a parent has the uri_override flag. If it does, the uri is built from the frozen URI instead of whatever parents may be higher up in the tree.

To illustrate, imagine this tree from the modmore site:

```
- Home
- About
- Extras
--- ContentBlocks
----- Pricing
----- Documentation
--- Redactor 
```

Assuming no frozen URIs and each alias to be the lowercase of the title, the URI for the "Pricing" page would be `/extras/contentblocks/pricing.html`. 

Now imagine we set the ContentBlocks resource to have a frozen uri of `awesome-extra/`, so we have a nice short link to a page slightly deeper in the hierarchy. Prior to this patch (or if disabled), the Pricing page would still be located at `/extras/contentblocks/pricing.html`. When enabling this patch, it will instead follow the parent URI and end up at `/awesome-extra/pricing.html`. 
### Why is it needed ?

Covered above, but primarily it adds more flexibility in building URIs while keeping a semantic structure in the resource tree.

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -782,6 +782,15 @@ $settings['friendly_urls_strict']->fromArray(array (
   'area' => 'furls',
   'editedon' => null,
 ), '', true, true);
+$settings['use_frozen_parent_uris']= $xpdo->newObject('modSystemSetting');
+$settings['use_frozen_parent_uris']->fromArray(array (
+  'key' => 'use_frozen_parent_uris',
+  'value' => '0',
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'furls',
+  'editedon' => null,
+), '', true, true);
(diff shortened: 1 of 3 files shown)


christianseel:
Yay! :+1:

TheBoxer:
@Mark-H Can you please add EN lexicon for the new system setting?

Mark-H:
@TheBoxer Added! Not 100% sure about the description, it's not a super easy feature to explain in one sentence :P

PR #12418, opened 2015-04-14
Anonymous username as System Setting

ghost:
Changed the hard-coded (anonymous) username to take a System Setting named 'default_username' in the 'session' area, and added the setting to the build data script.

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1531,6 +1531,15 @@ $settings['session_cookie_domain']->fromArray(array (
   'area' => 'session',
   'editedon' => null,
 ), '', true, true);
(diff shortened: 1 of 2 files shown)


TheBoxer:
@sottwell Why it's called `default_username`, when it's used for setting username for anonymous user?
Also there is missing english translation for that setting.

sottwell:
Because it is the default username for an unauthenticated user. At least that was my reasoning. However, feel free to call it anything you like. Or just cancel it and drop the whole thing if you don't like the idea.

sottwell:
What area should this be in? I was thinking session, but now I think it should be in authentication.

TheBoxer:
Thanks for clarification. 
Merged to develop.

sottwell:
I'll try to do it better next time.

PR #12420, opened 2015-04-14
Add JSONP support to the modConnectorResponse

Mark-H:
When returning an array from a processor->process function, the modConnectorResponse can turn that into a nice JSON response. With this patch, it can also support JSONP by looking at a "callback" get parameter, and returning the json wrapped in that.

This makes it easier to use the connector/processor approach for cross-site stuff, while also allowing the $modx->runProcessor response (modProcessorResponse) to remain useful from within snippets or stuff like that. The approach to JSONP prior to this patch would be to prepare the specific JSONP format in the process() function of the processor, after which you'll need to strip it out again if you want to re-use the processor through $modx->runProcessor(), which sucks.  

I have some slight concerns if there may be security implications by adding this, discussed below, but all considered I think it is a safe addition. 
1. One aspect is using the `$_GET['callback']` without sanitisation - that could potentially be used to inject arbitrary javascript into the response, but given that javascript would be executed on the requester, any hypothetical attacker would already have access to run arbitrary code or XSS on that requester. Also, sanitisation could break a legitimate callback, so some comments you see online are that you should just make sure it is valid, and return a 400 Bad Request if it isn't. 
2. Given JSONP allows cross-origin requests where regular AJAX doesn't, it might be easier to execute CSRF attacks across hosts with this feature. That is what the HTTP_MODAUTH validation is for though, so with that in place I don't think that is an actual security issue either.

### DIFF
--- a/core/model/modx/modconnectorresponse.class.php
+++ b/core/model/modx/modconnectorresponse.class.php
@@ -155,17 +155,22 @@ class modConnectorResponse extends modResponse {
         }
         if (is_array($this->body)) {
             @session_write_close();
-            die($this->modx->toJSON(array(
+            $json = $this->modx->toJSON(array(
                 'success' => isset($this->body['success']) ? $this->body['success'] : 0,
                 'message' => isset($this->body['message']) ? $this->body['message'] : $this->modx->lexicon('error'),
                 'total' => (isset($this->body['total']) && $this->body['total'] > 0)
-                        ? intval($this->body['total'])
-                        : (isset($this->body['errors'])
-                                ? count($this->body['errors'])
-                                : 1),
+                    ? intval($this->body['total'])
+                    : (isset($this->body['errors'])
+                        ? count($this->body['errors'])
+                        : 1),
                 'data' => isset($this->body['errors']) ? $this->body['errors'] : array(),
                 'object' => isset($this->body['object']) ? $this->body['object'] : array(),
-            )));
+            ));
+
+            if (!empty($_GET['callback'])) {
+                $json = $_GET['callback'] . '(' . $json . ')';
+            }
+            die($json);
         } else {
             @session_write_close();
             die($this->body);


christianseel:
:+1:

opengeek:
@Mark-H I am a bit concerned about the security implications here, probably because I do not understand the problem this solves in the first place. Can you provide some example of the issue this addresses? As is, I want to reject this, but I'd like to better understand the risk versus benefit here before I make any decision.

pixelchutes:
Re: `callback` sanitization, not sure if there's a defined spec to consider, but what about stripping all non-alphanumerics, and only allowing a few limited characters: (`-`, `_`, `.`, etc)

Mark-H:
The problem this solves is being able of talking to a connector/processor from another domain which isn't possible with "normal" AJAX because of the cross origin request. With CORS headers this can sometimes be fixed but for the project I looked into this I don't think that was an option.. though I don't recall exactly why not now. I think it had something to do with an ad server of sorts.

Because JSONP requests are basically <script> tags that are injected into the page, they can be used across domains without requiring CORS-related changes on the server. 

In terms of security, to the best of my knowledge after spending a fair bit of time searching the web, this does not expose anything new. The direct returning of $_GET sure feels gross, but this is specified by the _client_. If an attacker can trick the client into requesting a carefully formed JSONP request, there's a lot more they can do to do more harm (reading cookies, loading any external javascript, sending requests to other servers etc).

There are discussions online about sanitising the callback function to only support alphanumerical characters, but in the way I understand those, that does not add any actual security because the client provided it, and if the client can provide a malicious string as callback, it can do much worse things. 

This change also does not affect anything server-side like authentication and the $_GET['callback'] is not passed into any function, so if there _is_ any risk, it is to the client sending the request.

opengeek:
Fair enough @Mark-H and thanks for the additional explanation — that makes me feel more comfortable with the change.

PR #12092, opened 2014-10-28
Composer

rtripault:
Note : this is mainly a PoC to be tested & reviewed
### What does it do ?
- removes dependencies from repository
- make use of Composer to handle dependencies (xPDO included) & some auto loading
- update dependencies (see composer.json)

Notes : 
- xPDO is shipped with predis as well as [Garry's PR](https://github.com/modxcms/xpdo/pull/40)
- Composer requires PHP >= 5.3.2
### What to expect from this ?
1. breakage (hence why some testing is needed)
2. everything **must** work as before (if not, please comment with as much details as possible to help debug/fix, thanks!)
### How to install ?

pull, (make sure you have composer installed) composer install --no-dev, run setup as you would normally do.

I encountered an issue with Smarty dependency : 
`Could not scan for classes inside "path/to/core/vendor/smarty/smarty/distribution/libs/Smarty.class.php" which does not appear to be a file nor a folder`

To manually fix, head over to `vendor/composer/installed.json`, search for

```
        "classmap": [
            "distribution/libs/Smarty.class.php",
            "distribution/libs/SmartyBC.class.php",
            "distribution/libs/sysplugins/smarty_security.php"
        ]
```

replace with

```
        "classmap": [
            "libs/Smarty.class.php",
            "libs/SmartyBC.class.php",
            "libs/sysplugins/smarty_security.php"
        ]
```

run `composer install --no-dev` again.

If you plan to try this on an existing installation, make sure to wipe your `core/cache` folder (encountered some issues with cached Smarty templates)
### What needs to be done/tested ?
- [x] make the setup process work
- [ ] test with old components (non class based processors/controllers)
- [ ] performances benchmarks ?
- [ ] S3 Media Sources (SDK has been upgraded an no bucket to test here)
- [ ] Revolution upgrade
- [x] triple check Smarty templates, Smarty upgrade seems to introduce lots of exceptions (mostly due to templates trying to access undefined variables) when error_reporting(E_ALL)
### What needs to be done if merge is considered ?
- [ ] handle cases where `core` or `manager` folder are moved (would impact the setup process -- to find the auto loader), currently, editing `composer.json` & running `composer dumpautoload` is required
- [ ] think about future & potential usage of PHP namespaces
- [ ] think how third party components could define dependencies (how to handle "conflicts" ?)
- [ ] find a BC way to move 1 class per file (ie. no multiple classes definitions in a single file)
- [ ] update xPDO to make use of the loader too (no more "require_once")
[...]

PR #12422, opened 2015-04-17
Create composer.json

PatchRanger:
### What does it do ?

Turns modX into [Composer](http://getcomposer.org) library. No autoloading by namespaces, no core split into separate packages - just making modX available to be installed via Composer, nothing more and nothing less.
### Why is it needed ?

It gives modX an opportunity to be installed via the most popular dependency manager for PHP.
### Related issue(s)/PR(s)

This pull-request could be considered as a first step to https://github.com/modxcms/revolution/pull/12092 .

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,17 @@
+{
+    "name": "modx/revolution",
+    "description": "Content Management System and Application Framework",
+    "keywords": [
+        "modx",
+        "cmf"
+    ],
+    "homepage": "http://www.modx.com/",
+    "type": "library",
+    "license": "GPLv2",
+    "require": {
+        "php": ">=5.3"
+    },
+    "config": {
+        "preferred-install": "dist"
+    }
+}


zaigham:
After this is merged, MODX will be automatically available through https://packagist.org/?

OptimusCrime:
@zaigham No, the modx people still need to submit it as a package.

PatchRanger:
@zaigham @OptimusCrime It is true, the modx people still need to submit it as a package, but as a workaround someone could add a repository to the application's `composer.json`, something like this: http://stackoverflow.com/questions/12954051/use-php-composer-to-clone-git-repo .

pixelchutes:
`3.x` branch now has [composer.json](https://github.com/modxcms/revolution/blob/3.x/composer.json) file. Will this eventually conflict by adding this one to the `2.x` series?

OptimusCrime:
I don't see how. It does not brake anything backwards?

Alroniks:
Master branch now outdated (exists for BC). So,  2.x and 3.x absoluttly different branches, so I think it can me merged into 2.x. But I don't know why. I can't imagine cases when I can use modx from composer without installing. It's actual for xPDO, but not for modx.

theboxer:
Composer.json won't be present in 2.x version of Revo.

PR #12437, opened 2015-05-12
Take care of added records while counting the total number of widgets placed in a dashboard

rtripault:
### What does it do ?

Calculate the real total of dashboard widgets (to "mimic" a rank/placement), instead of the total returned by the server.
### Why is it needed ?

When editing/placing widgets into a dashboard, `Ext.data.Store#getTotalCount()` was used. This method only takes care of "cached" records (as in not counting newly added records -- see the note at the end of description http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.data.Store-method-getTotalCount).

Basically, when editing a dashboard which only had 1 record, every records added/placed later (without save in between) would have a rank of `1`.

Making use of `store.data.length` appears to be more reliable.

### DIFF
--- a/manager/assets/modext/widgets/system/modx.panel.dashboard.js
+++ b/manager/assets/modext/widgets/system/modx.panel.dashboard.js
@@ -236,9 +236,8 @@ Ext.extend(MODx.grid.DashboardWidgetPlacements,MODx.grid.LocalGrid,{
 
     ,onAfterRowMove: function(dt,sri,ri,sels) {
         var s = this.getStore();
-        var sourceRec = s.getAt(sri);
-        var belowRec = s.getAt(ri);
-        var total = s.getTotalCount();
+        var sourceRec = s.data.items[sri];
+        var total = s.data.length;
 
         sourceRec.set('rank',sri);
         sourceRec.commit();
@@ -246,7 +245,7 @@ Ext.extend(MODx.grid.DashboardWidgetPlacements,MODx.grid.LocalGrid,{
         /* get all rows below ri, and up their rank by 1 */
         var brec;
         for (var x=(ri-1);x<total;x++) {
-            brec = s.getAt(x);
+            brec = s.data.items[x];
             if (brec) {
                 brec.set('rank',x);
                 brec.commit();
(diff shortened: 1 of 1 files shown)


rtripault:
I guess this could have been targeted at master, since that's just a minor "fix"...

manager/assets/modext/widgets/system/modx.panel.dashboard.js
```diff
@@ -324,7 +324,7 @@ Ext.extend(MODx.window.DashboardWidgetPlace,MODx.Window,{
             fld.markInvalid(_('dashboard_widget_err_placed'));
             return false;
         }
-        var rank = s.getTotalCount();
+        var rank = s.data.length;
```
theboxer: @rtripault I think it should be `s.data.length + 1`, otherwise when I add another widget, it has same rank as the previous.

screen shot 2015-06-26 at 15 24 56

rtripault: As far as i remind, ranks should be starting at 0, which should make the `+ 1` useless.
All in all, when ranks are messed up, increment on the number of records is not the solution (we should rather try to find the highest rank then increment that one).

I might triple check that

theboxer: Yeah, maybe grabbing the last row and incrementing it's rank?

rtripault:
PR updated following your advise.
Thanks

theboxer:
@rtripault If you have empty widgets grid, this will not allow you to add a widget in it, because `s.getAt(s.getTotalCount() - 1)` will be undefined

rtripault:
Good call, will take care of that.
Sorry for being so "hasty".

rtripault:
Finally took care of that... good thing my last update was causing issue, since i was reproducing the mistake of checking the store (loaded from ajax) rather than the actual content... while being there, i also did some tweak to fix drag & drop reordering while records were not saved.

Hoping this time things will be fine!

theboxer:
Merged to master & develop

PR #12455, opened 2015-05-23
Log Manager login

ghost:
Added a parameter to the logManagerAction function and use it to set the userId property. Add a call to the logManagerAction function in login.class.php using the new parameter.

When a user is first logged in, invoking the OnManagerLogin event, the user object is already instantiated as the "anonymous" user with the userId of 0. So there is no way for the logManagerAction function to have the newly logged-in user's Id. The event would be logged in the database, but with a userId of 0, and of course could not be meaningfully displayed in the Manager Actions list. 

By being able to pass the userId to the API function, and having that passed value override the default property of 0, now the login.class.php file or even a plugin using the OnManagerLogin event.

Because the new parameter is added to the end of the parameter list in the function definition, it can be ignored in a function call, so this is backwards-compatible and does not affect existing usage.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1887,10 +1887,11 @@ class modX extends xPDO {
      * @param mixed $item The primary key id or array of keys to grab the object with.
      * @return modManagerLog The newly created modManagerLog object.
      */
-    public function logManagerAction($action, $class_key, $item) {
-        $userId = 0;
-        if ($this->user instanceof modUser) {
+    public function logManagerAction($action, $class_key, $item, $userId= 0) {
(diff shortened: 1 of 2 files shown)


TheBoxer:
Hello @sottwell 
I don't think that it's necessary to update `logManagerAction` method to achieve logging logins.
You can do it via:
```php
$this->modx->logManagerAction('login_mgr','modUser',$this->user->get('id'));
```

Do you agree with this change?

sottwell:
Whatever works. You certainly know the core better than I do.

TheBoxer:
Actually no, it does not work, I see the problem you are solving with the additional param now

PR #12466, opened 2015-06-12
Get rid of useless "language string not found"

wuuti:
The debug message "Language string not found:" is quite useless here, because you don't see where it comes from. 
And, even worse these log lines poison your logfile making it hard to debug any other manager issues.

### DIFF
--- a/core/model/modx/modlexicon.class.php
+++ b/core/model/modx/modlexicon.class.php
@@ -389,7 +389,7 @@ class modLexicon {
         $language = !empty($language) ? $language : $this->modx->getOption('cultureKey',null,'en');
         /* make sure key exists */
         if (!is_string($key) || !isset($this->_lexicon[$language][$key])) {
-            $this->modx->log(xPDO::LOG_LEVEL_DEBUG,'Language string not found: "'.$key.'"');
+        //    $this->modx->log(xPDO::LOG_LEVEL_DEBUG,'Language string not found: "'.$key.'"');
             return $key;
         }
         /* if params are passed, allow for parsing of [[+key]] values to strings */


TheBoxer:
Not sure why it's useless, you can search for the reported key so you'll find where it is used.

inreti:
Where am I supposed to search for the key then? Grepping the core files? This results for most keys in multiple locations
But even if you can use it with searching, it still poisons the log file, e.g. there are a lot of entries for "JPG" and so on. Which will, I am quite sure, never be translated. At least it should be configurable if you want to output that message, say with an option like system.output_lang_debug.
Or - even better - have an extra log file for translation specific logging.

TheBoxer:
I think if those warnings are mostly caused by the core it should be resolved in the core. Disabling those warnings doesn't solve the problem, it will just hide the problem and remove an option to get a message when lexicon string is missing (for example when developing a component, it's a nice way how to know you didn't forgot to translate aomething).

Mark-H:
Those messages are of level modX::LOG_LEVEL_DEBUG (=4), so setting your log_level setting to a lower value (e.g. 1 for only actual errors and not debug statements, or 2 to include warnings) should hide them.

inreti:
Yeah, I now about the log level. But as I said before, it makes it hard to debug the logs. Meaning, I have log level debug to find out some special bugs somewhere else, and those lines fill up my logfile and make it quite impossible to find it. It is easy when you now what to look for - you can use grep then. But it is not easy if you really need to debug all the logs looking for suspicious messages...

So what about the idea of having different logfiles for different purposes? We mentioned a security log at the ccc, and a special log file for translation purposes should do the trick as well. Imho it seems a good idea to split up the log, esp. for a security log. Shouldn't be that hard to implement, just another parameter at the log call - default goes to error log. This could also open the chance for extras to have their custom logfiles then...

Mark-H:
What if the "special bug" you're looking for is related to a lexicon not working? Being able of seeing a debug log that mentions the lexicon key not being found would be incredibly useful then ;)

Mark-H:
As for making logs smarter, I'd love to see monolog being adopted for that in the future, but that might no be easy to do in a backwards compatible matter. Worth investigating though.

Jako:
The most annoying part of it `Language string not found: ""` is solved by #12545

Maybe process should check for an empty key too and return an empty value then without processing further. Otherwise I would not have found the #12545 issue without that warning.

PR #12468, opened 2015-06-15
Enhance and fix configcheck

wuuti:
### What does it do ?

It adds a rule for the configuration check. This rule checks if the PHP version is outdated. The version to check can be set by system setting 'configcheck_min_phpversion', default is 5.4, which is the currently maintained minimum version. 
A small html bug is also fixed, and styling in the widget is beautified. 
### Why is it needed ?

MODX does not check the PHP version. At least (if not becoming a requirement), it should warn the admins about their outdated environment. This helps to secure MODX sites in the wild.
It also may reduce support load - the warning message could be enhanced maybe with concrete advices for problems with specific versions. 

screenshot from 2015-06-15 15 01 15
_(Note: versions in the screenshot are only for testing purpose. Default min version is 5.4 in the code)_


TheBoxer:
Hello, thanks for this update, it looks really nice.

Can you please update default value for min php to `5.3` and remove changes from german lexicons (all translations should be done via crowdin)?

inreti:
German translations are not really changed - I removed previous changes when I read about the crowdin thing. 
Is there a reason for setting default value to 5.3? This version is not maintained any more as well and imho should at least give a warning then.

TheBoxer:
5.3 will be the official minimum supported version.

PR #12469, opened 2015-06-16
Paging in modx-combo-category

argnist:
Also why all combos use hard-coded 20 per page instead of using MODx.config.default_per_page ?


Mark-H:
That's a very good question, maybe a good idea to start being more consistent and start using that setting more often? :D

Mark-H:
Hi @argnist, would you like to update this PR to make it use `MODx.config.default_per_page` instead? If not I can do that while merging.

Mark-H:
Thanks @argnist for the contribution, I've merged this one in. It would probably be a good idea to start using the MODx.config.default_per_page value more often.

PR #12470, opened 2015-06-18
Enhance and fix configcheck

wuuti:
### What does it do ?

It adds a rule for the configuration check. This rule checks if the PHP version is outdated. The version to check can be set by system setting 'configcheck_min_phpversion', default is 5.3, which is the version MODX is supposed to support now.
A small html bug is also fixed, and styling in the widget is beautified. 
### Why is it needed ?

MODX does not check the PHP version. At least (if not becoming a requirement), it should warn the admins about their outdated environment. This helps to secure MODX sites in the wild.
It also may reduce support load - the warning message could be enhanced maybe with concrete advices for problems with specific versions. 

screenshot from 2015-06-15 15 01 15
_(Note: versions in the screenshot are only for testing purpose. Default min version is 5.3 in the code)_


Mark-H:
Ohh I like what you did with the styling!

rthrash:
very nicely done indeed. thanks!

PR #12481, opened 2015-06-23
Compile Sass with libsass

jpdevries:
Switches to using `grunt-sass` which uses `libsass`, a much faster Sass compiler. 

The `grunt sass` task now takes a fraction of the time.

Also addresses Sass warnings by removing the `box-sizing` mixin which is no longer needed (we have autoprefixer)

If you get a collision on the .css files just discard those changes and run `grunt build`.

Please

``` bash
cd _build/templates/default
npm install #install grunt-sass
grunt build
```


renekopcem:
:+1:

PR #12496, opened 2015-06-30
Show MODX version (and flavor) as a tooltip for the MODX logo

mindeffects:
### Summary

Show MODX version (and flavor) as a tooltip for the MODX logo. Ends the long journey to "Manage > Reports > System Info" to find out which version of MODX is used.
Note: The line break inside the "title" of the "a" tag is important to make the tooltip use 2 lines.
### Expected behavior

image

### DIFF
--- a/manager/templates/default/header.tpl
+++ b/manager/templates/default/header.tpl
@@ -52,7 +52,8 @@
 
             <ul id="modx-topnav">
                 <li id="modx-home-dashboard">
-                    <a href="?" title="{$_lang.dashboard}">{$_lang.dashboard}</a>
+                    <a href="?" title="MODX {$_config.settings_version} ({$_config.settings_distro})
+{$_lang.dashboard}">{$_lang.dashboard}</a>
                 </li>
                 {if $_search}
                 <li id="modx-manager-search"></li>


Mark-H:
#11958 already accomplishes this as well, which was just merged into develop (MODX 2.4). 

Wondering if it makes sense to include this PR separately for master (2.3) if 2.4 is supposed to be coming soon? Thoughts @modxcms/integration-managers-revolution?

mindeffects:
The goal is: Making it easier to find out which version of MODX is being used.
If there is a 2.3.6, I am fine to put my blunt solution in there.
If 2.4 comes with the great solution of @pixelchutes, I would be more then happy to see this one in there instead!

pixelchutes:
I don't think this PR will conflict with anything in 2.3, or even in 2.4 with the recent merge. :+1: from me.

mindeffects:
Thanks, @pixelchutes / Mike.

OptimusCrime:
+1 for merge of this

Mark-H:
Thanks for your contribution @mindeffects  :)

eighthday:
This is handy for devs, but for editors it not friendly, I would appreciate a way to turn it off...

mindeffects:
Well, since the MODX version is now displayed below the site name, we don't really need this tooltip any more.

eighthday:
sorry! yes I thought this was about the version number under the logo, but it is about the tooltip (which is a much better idea)

mindeffects:
No, not if you are an admin and need/want to see if a site needs an update. ;-)
(I could start the "MODX needs a core-integrated self-updater and not an extra"-discussion again, but I don't feel ready for it for the kicks. :-D )

I would like 2 things:
1. make the version info smaller
2. eing able to turn it on/off via a system setting

whitebyte:
Please, not another system setting! "There is system settings for everything" policy turns MODX into Windows registry.

pixelchutes:
It could probably be hidden with some CSS and a custom Plugin.

mindeffects:
Nobody is forcing you to use a system setting, @whitebyte. I, personally, like to have a choice. 
And why should I do it with a plugin (for every(!) installation), when a system setting does the job with a single click, @pixelchutes? And the handling can easily be altered for contexts and usergroups. Yes, I am a fan of system settings! ;-)

jpdevries:
I think a System Setting makes sense. Also, maybe a Dashboard Widget that shows the current version?

eighthday:
I would also like the option to replace the modx logo with the sites favicon.

whitebyte:
@eighthday AFAIK this can be done with custom manager theme (never saw one in the wild though).

eighthday:
@whitebyte I used to always make custom manager themes, but it became a pain to update them with new modx versions. I am happy with the manager layout, just like to make everything as non techie as possible

jpdevries:
Until/unless this gets changed in the core you can create a simply plugin that includes the following CSS.



#### CSS
```css
#modx-site-info {
  font-size: 1.2em;
  line-height: 2em;
}
#modx-site-info .full_appname {
  /* inaccessible way of hiding */
  display: none;
}
```

eighthday:
thanks jpdevries, I never thought changing the manager css via a plugin. It works great and is easier to transport.

PR #12498, opened 2015-06-30
Fiting portrait and landscape imageTV thumbs into a max. area of 400x400px

mindeffects:
### Summary

This is an improvement for #12491 which only limited the width to 400px, the height remained unlimited.
Now the script also handles very tall images like skyscraper banners (120x600px) and fits them into a 400x400px area. This saves a lot of vertical space for portrait format images.
Smaller images remain unaltered.
### Step to reproduce

Just select an image for a TV. Here are examples for 120x120, 600x120, 120x600 and 600x600px:
##### 120x120

image
##### 600x120

image
##### 120x600

image
##### 600x600

image
### Observed behavior

The maximum thumbnail size is now fitted into 400x400px, no matter what kind or proportions the image has.
### Expected behavior

Just as observed. ;-)


inreti:
+1

Mark-H:
Another good contribution @mindeffects ;) Thanks for helping make MODX better. 

The idea of making the 400px configurable also sounds like a good improvement!

Issue #12490, opened 2015-06-29
/core/packages should not be visible to public


mdxthemes:
/core/packages should not be visible to public. I can simply put blank index.html file but it should come default with modx installation. 

Google search for "index of /core/packages"

https://www.google.co.in/search?q=index+of+/core/packages/&ie=utf-8&oe=utf-8&gws_rd=cr&ei=rqyQVYCFLYXGuASylYLwCQ

A default robot.txt would be good. 

# Robots.txt
User-agent: *
Disallow: /connectors/
Disallow: /core/
Disallow: /manager/
Disallow: /setup/

Mark-H:
I like the idea of a default robots.txt, though it is also a good way for nasty people to find out what they shouldn't be looking at. Given the number of results on that search, making them harder to find in the first place probably outweighs that though..

lektriq:
What about renaming the ht.access file in /core to .htaccess when installing?

We already have a way to deal with this on Apache at least.

    IndexIgnore */*
    
    <Files *.php>
        Order Deny,Allow
        Deny from all
    </Files>


While the results list on a search are not great at just under 7000 results (search with quotes for exact matches), it’s not as bad as the 33,10,000 results shown using the search linked above.

mdxthemes:
The results may be low but it displays which extra version the website is using. If any extra has any security issues, hackers might use it to gain access. Robots wont allow search engines to index but it is still visible to public. .htaccess should be placed in all such directories by default.

OptimusCrime:
Once again, flexibility is what Modx stands for. Supply an example robots.txt. Let people use it if they will, if not it is really up to the developer/host to decide how to handle this.

jpdevries:
Should a robots.txt also disallow assets/components? It seems like those looking to harden security would move the core folder to a non public area of there server anyways, but can't do that with assets.

lektriq:
What about giving an option to rename /core/ht.access if it is not already at install time?

I can’t think of good reasons anyone is keeping it in a web-accessible place and not using it, unless they are configuring access in the main config, are not using Apache, or are securing it by other means.

lektriq:
Maybe the top level .htaccess could also do with having ```Options -Indexes``` added?

amdbuilder:
I would recommend using this for the .htaccess instead, as far as I know there isn't a reason any files should be accessible.  I'd also be inclined to rename ht.access to .htaccess, or perhaps make that a setup option (let server rename it).

```
[...]

Issue #8854, opened 2012-10-03
Make error log location not accessible for anyone


modxbot:
michelle84 created Redmine issue ID 8854

Almost every MODX revo site has it's error log location at /core/cache/logs/error.log and the log is accessible for anyone. This could be a vulnerability when you use the error log to test or track sensitive information (which you shouldn't, I know, but it happens sometimes). So maybe it would be a good idea to hash the path to the error log file like: /core/cache/logs/error.82f23h0230fh2sl29sgkwlqoch7.log so nobody can access the error log when they don't know it's URL.

inreti:
Then this should at least be coded at the ht.access file inside the core as an example.

Mark-H:
Closing per fix in #12503

PR #12503, opened 2015-07-07
Deny any access to core folder from web

wuuti:
### What does it do?

Prevent sneaking around or fingerprinting, the whole core directory should be denied. 
### Why is it needed?

Otherwise it is possible to read the error log or check the changelog.txt from the web. As this is the example file, it should include the most secure set up. 
### Related issues

For example mentioned in #8854, #12490
### Enhancements

The file should give examples for other environments as well (or other example files). Set up for Nginx and IIS at least are a good advice for users.

### DIFF
--- a/core/ht.access
+++ b/core/ht.access
@@ -1,5 +1,4 @@
 IndexIgnore */*
-<Files *.php>
-    Order Deny,Allow
-    Deny from all
-</Files>
+
+# deny access to _all_ files in the core, including changelog.txt and error.log
+Deny from all
--- a/core/ht.access
+++ b/core/ht.access
@@ -1,4 +1,16 @@
-IndexIgnore */*
-
 # deny access to _all_ files in the core, including changelog.txt and error.log
-Deny from all
+# original borrowed from owncloud
+
+# line below if for Apache 2.4
+<ifModule mod_authz_core.c>
+Require all denied
+</ifModule>
+
+# line below if for Apache 2.2
+<ifModule !mod_authz_core.c>
+deny from all
+Satisfy All
+</ifModule>
+
+# section for Apache 2.2 and 2.4
+IndexIgnore *
--- a/core/ht.access
+++ b/core/ht.access
@@ -1,4 +1,16 @@
-IndexIgnore */*
-
 # deny access to _all_ files in the core, including changelog.txt and error.log
-Deny from all
+# original borrowed from owncloud
+
+# line below if for Apache 2.4
+<ifModule mod_authz_core.c>
+    Require all denied
+</ifModule>
+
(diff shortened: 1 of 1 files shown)


mindeffects:
+1 Excellent!
Maybe the file should be renamed to ".htaccess", too? Sort of "protection by default".

inreti:
There were arguments that it should not be named .htaccess by default because this would cause the file to be overwritten by every update and therefore you would lose any custom changes inside. But I do not know what changes would be in there in any customized way. Does anyone have a real life example? I am on your side, I would like to have it there as a default as well.

OptimusCrime:
Do not rename to `.htaccess`. This will, as already stated, cause problems upon upgrades. It will also cause problems for people using version control as the `.htaccess`-files are ignored (as they should be).

The best way to avoid this is to add a rule in the default `ht.access` file for hiding content in the `core/` directory.

Mark-H:
In the discussion in #12490 the following snippet is suggested by @amdbuilder:

````
Order Allow,Deny
Deny from all
````

That's not in this PR... I don't want to claim to be an expert in apache, but maybe this PR warrants a bit more discussion about the actual code to put in the ht.access?

**Comment by mindeffects** [comment]
Wouldn't it be safer to generate real 404s for everything inside "core/"?
With 403s you can spy on the structure of the folder. Correct filename/foldername guesses get a 403, wrong ones a 404.

What do you guys think about this one liner for the `.htaccess`:
```RedirectMatch 404 .^```

**Comment by krismas** [comment]
+1 @mindeffects

In my opinion sending 404 is as efficient as 403 but simpler (and compliant with the RFC26616#section-10.4.4). For Apache, I don't use `core/ht.access` but in place I add these lines in my "root" `.htaccess` : 

```
RedirectMatch 404 ^/core/.*$
```

These 404 errors are not catched by MODX, so an optional `ErrorDocument 404` may be added.

inreti:
RedirectMatch only works with mod_alias. So we should test for that and if its not there, use the full deny method as a fallback.

sottwell:
Using the global deny all directive causes a lot of entries to pile up in the server error logs, since every time the dashboard widget checks it, it triggers a forbidden error.

mindeffects:
Closed? What happend?

inreti:
Ähhh. Yes. After the modx repo cleanup I tried to cleanup my fork as well - and forgot that this PR was in the path-2 branch and had not been merged :-(
I will restore it.

Mark-H:
Merged into 2.5.0-rc1 in 4d60b7b2cad082ab3017050d30da8e597717e432

Thanks for your contribution, appreciate your patience :)

PR #12504, opened 2015-07-08
Added rule for checking .htaccess in /core folder on Apache

wuuti:
### Summary

Show a warning at the configcheck-widget, if running on apache and there is no .htaccess file under /core. 

screenshot from 2015-07-08 13 49 40

Related issues: #12490, #8854 
Pull Requests:  #12503

### DIFF
--- a/core/lexicon/en/configcheck.inc.php
+++ b/core/lexicon/en/configcheck.inc.php
@@ -21,9 +21,16 @@ $_lang['configcheck_errorpage_unavailable_msg'] = 'This means that your Error pa
 $_lang['configcheck_errorpage_unpublished'] = 'Your site\'s Error page is not published or does not exist.';
 $_lang['configcheck_errorpage_unpublished_msg'] = 'This means that your Error page is inaccessible to the general public. Publish the page or make sure it is assigned to an existing document in your site tree in the System &gt; System Settings menu.';
 $_lang['configcheck_htaccess'] = 'Core folder is accessible by web';
-$_lang['configcheck_htaccess_msg'] = 'Your MODX installation is running on a <strong>Apache</strong> webserver and there is no <em>.htaccess</em> file in the <em>core directory</em>. To secure your site, you should at least properly set this file. This can be easily done by renaming the existing ht.access file to .htaccess at the location <em>[[+fileLocation]]</em>.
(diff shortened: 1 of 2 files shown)


OptimusCrime:
Really nice, but this does not check if the system is configured somehow without the provided `.htaccess`-files and may report false positives.

inreti:
Btw: it also checks if the core folder was moved out of the base path and does not complain then assuming hardening took place.

Mark-H:
On my site I've got access to the core directory (among others) blocked from within the .htaccess in the root rather than the one in the actual directory. Even though core/docs/changelog.txt is not accessible (being routed to the 404 page), the message is still displayed on the dashboard. 

I think the idea behind this pull request is great and will help promote security more, but maybe the implementation needs some more work before this can get merged into 2.4. Here's what I personally think needs to happen to get this merged in:

* Rewrite the check logic to do a web request to core/docs/changelog.txt (or another file inside core). If the request returns a 200 response code, show the warning. If it's something else (which could be a 404, 302 or 500 depending on how the core folder is locked down) the test passed.
* Because the check then no longer depends on the .htaccess file, the check can then run on any server environment so the SERVER_SOFTWARE check can be removed.
* The lexicon will also need to be updated in that case. "Hardening MODX" could also be a link to that doc in the RTFM to help people find it quicker. 

Please let me know what you think.

inreti:
Totally agree. Would be much better and webserver independent. Unfortunately the "Hardening MODX" doc currently does not contain a section about this kind of hardening, that is why I have not linked it directly at the message there. Anyway, this document needs a rewrite.

Getting back to _this_ problem: I am struggling with constructing the right (absolute) URL for the check. How to determine the base domain? Maybe to early in the morning, but MODX does not really now about the domain a frontend context is running at, does it? 
I cannot use HTTP_HOST here, because we are running at the manager and the manager might be installed under different url location (as supposed by the hardening doc). Anyone knows something failsafe to construct the URL to the changelog then?

inreti:
What would be the advised status code to return there? IMHO it should return a 404 not found to avoid fingerprinting...

inreti:
I just changed my commits for this PR according to @Mark-H 's comments. 
I did use HTTP_HOST do guess the web-url for the check, although I am aware that this won't work when the manager is on a different location than the site. Well, it may work, if the core is accessible there too. Until now I don't see a better solution for that. Requesting comments.

Mark-H:
Thank you @inreti for this great contribution :)

mdxthemes:
Since the attacker knows which version the site is using, he could exploit the code if its buggy. They can also download the entire site backup if vapor package is compiled. 

Good to see that it will be fixed. Cant leave the user to secure website on his own. MODX should do it by default. New or even existing users dont know what to fix. There should be a list of to-do's in rtfm. The current one does not include a lot of things. 

Thanx @inreti :-)

PR #12516, opened 2015-07-16
Extra space in ru lexicon key

ViieeS:
Extra space in ru lexicon $_lang['upload.notpermitted.extension']


Mark-H:
Hi @ViieeS, all lexicons except English are managed via http://translate.modx.com/ and not directly on GitHub. Could you make the change there?

Thanks!

Issue #12457, opened 2015-05-25
(Enhancement + Code) Manager doesn't show users fullname in header


JensWolff:
Hey guys,
I was wondering why the manager doesn't actually shows the fullname as "Username" of the user who is logged into the manager (next to the gravatar-image), so I solved this issue by myself.

If the fullname-field ist empty, there is a fallback to the actual "username":
![2015-05-26_00h31_23](https://cloud.githubusercontent.com/assets/754205/7802819/a0dc6752-033e-11e5-85ef-8bbec861b869.png)

But if the fullname-field is filled, it gets used, to show the much cleaner fullname:
![2015-05-26_00h31_03](https://cloud.githubusercontent.com/assets/754205/7802832/bff8d3aa-033e-11e5-9ded-1815f6e799d9.png)

Replace the function "setPlaceholders" (starting at line 78) in the file "manager/controllers/default/header.php" with the following to get it work:
==============================================================

    public function setPlaceholders()
    {
        
        $userProfile = $this->modx->user->getOne('Profile');
        $username = $userProfile->get('fullname');
        if(empty($username)) $username = $this->modx->getLoginUserName();
        $placeholders = array(
            'username' => $username,
            'userImage' => $this->getUserImage(),
        );

        $this->controller->setPlaceholders($placeholders);
    }

==============================================================

Many greeting from germany
Jens

zaigham:
You should do a Pull Request. :+1:

JensWolff:
@zaigham The problem is, that I am still not very familiar with Git to do so :D

pixelchutes:
Not sure if it was by design, but my gut says it could be accounting for the already limited space in the navbar (see #11411 and #11430)

That aside, this could be a great candidate for a system setting for more control / consistency.

JensWolff:
@pixelchutes A system-setting to enable it would be perfect! :+1:

OptimusCrime:
System settings seemst to be the solution for almost everything in Modx. I am negative for the suggested feature too. This does not add anything crutial to Modx.

wuschigesetwas:
if anyone needs a patch https://github.com/modxcms/revolution/issues/12457

JensWolff:
Thanks, bro :-* :+1: :+1:  :+1:

wuschigesetwas:
you're welcome :) :-*

Mark-H:
Merged into 2.4

PR #12523, opened 2015-07-24
(Enhancement + Code) Manager doesn't show users fullname in header 

JaWe09:
https://github.com/modxcms/revolution/issues/12457

### DIFF
--- a/manager/controllers/default/header.php
+++ b/manager/controllers/default/header.php
@@ -77,8 +77,16 @@ class TopMenu
      */
     public function setPlaceholders()
     {
+        if($this->modx->getOption('manager_use_realname') == true) {
+            $userProfile = $this->modx->user->getOne('Profile');
+            $username = $userProfile->get('fullname');
+        } else {
+            $username = '';
+        }
+
+        if(empty($username)) $username = $this->modx->getLoginUserName();
         $placeholders = array(
-            'username' => $this->modx->getLoginUserName(),
+            'username' => $username,
             'userImage' => $this->getUserImage(),
         );
 


OptimusCrime:
Is this system setting already implemented in the setup?

wuschigesetwas:
@OptimusCrime nope sorry. But it's failsave if there is no setting it will take login as usual

Mark-H:
Hi @wuschigesetwas! Thanks for sending a pull request with that improvement. I've got a few questions/tips/feedback. 

1) Have you signed the MODX CLA before? If not, please visit http://develop.modx.com/contribute/cla/ and sign the agreement in order for us to accept your contribution. Please let me know your MODX.com username too. 

2) Please follow the contribution guidelines (see https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request) when sending a pull request. This includes a simple template to use for the pull request description to make sure all relevant information is there.

3) I would recommend creating separate feature branches on your fork for each pull request. Right now you committed directly to your master branch which will make it harder for you to send other pull requests if a pull request is not accepted or merged into another branch. Which brings me to.. 

4) This looks like a new feature rather than a bug fix, so this will need to go into develop (targeting 2.4) rather than master (targeting 2.3.x). 

5) The system setting will need to be added to the build so it is created on install. As it's a new setting, just adding it in `_build/data/transport.core.system_settings.php` will do the trick. Please also add a name and description for the setting to the English lexicon (English only - other languages are managed via translate.modx.com) in `core/lexicon/en/setting.inc.php`. 

6) Finally (sorry this became a longer list than I thought), does it make sense to use "manager_use_fullname" instead of "manager_use_realname"? A fullname doesn't have to be a real name, and the used field is called fullname too.

wuschigesetwas:
@Mark-H thank you a lot for all these helpfull tips. I wasn't aware of that all :).
CLA ist signed now modx.com username is "dunklertempler" (i known strange usernames ... ).

i think it would be better if i would reset my fork to upstream and create a new feature branch considering all you've mentioned above and after that create a new pull request ?!

Best Regards,

Jan

Mark-H:
Thanks for signing the CLA, Jan! Starting with a new branch and sending a new PR might indeed be the best option :)

wuschigesetwas:
@Mark-H okay so we'll close this one ... :) and i'll come back on that with a new pull request :)

wuschigesetwas:
Hello @Mark-H,

is #12527 okay that way ?

Mark-H:
From a first look that seems great (though there seems to be a merge conflict, which is odd, but I can't immediately see why).

wuschigesetwas:
mhm different bases may be but i think that's unlikely the reason.

PR #12525, opened 2015-07-25
fix recursive overriding

goldsky:
### What does it do ?

Fixing this:
combo-bug
It breaks combos.
### Why is it needed ?

I'm not sure the purpose of the the overriding, but I'd suggest to move the method on MODX's combo instead, because it became an infinite loop on original commit of overriding.
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/commit/f9deafa8cef3fbd3e6e3e183dc723094fd1a7beb#commitcomment-12355170

### DIFF
--- a/manager/assets/modext/widgets/core/modx.combo.js
+++ b/manager/assets/modext/widgets/core/modx.combo.js
@@ -19,8 +19,7 @@ Ext.override(Ext.form.TriggerField, {
         this.resizeEl = this.positionEl = this.wrap;
     }
 });
-/* store the original onLoad method to have acces to it in the override */
-var originalComboBoxOnLoad = Ext.form.ComboBox.prototype.onLoad;
+
 /* fixes combobox value loading issue */
 Ext.override(Ext.form.ComboBox, {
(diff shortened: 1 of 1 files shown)


Mark-H:
Hi @goldsky thanks for helping making MODX better. 

This is probably me being an idiot, but I don't really understand the reason behind this pull request even when looking at the CrossContextsSetting issue you mentioned it on. Could you update the pull request to follow the template in the [contributing guidelines](https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request) and provide some additional information?

goldsky:
updated

theboxer:
@goldsky I can't reproduce this issue with Revolution 2.5.0-dev

opengeek:
@goldsky if this is still an issue, reopen and add more detail so we can reproduce the problem in latest releases

goldsky:
hm... I think it's somehow fixed by itself.
Tested on 2.4.2.
I'll open another issue if it occurs back.

PR #12527, opened 2015-07-27
(Enhancement + Code) Manager doesn't show users fullname in header #1…

### LINKED ISSUE #12457: (Enhancement + Code) Manager doesn't show users fullname in header

JaWe09:
### What does it do ?

In #12457 @JensWolff provides a method for showing fullname instead of login name. I've added this method with system setting support.
### Why is it needed ?

Because some other people and i think  that it is a good idea :) and because modx posibilities and flexibility is awesome and should be developed further :) 
### Related issue(s)/PR(s)

fixes #12457
#12523

@Mark-H i hope thats better :) if I miss something please let me know.

### DIFF
--- a/manager/controllers/default/header.php
+++ b/manager/controllers/default/header.php
@@ -77,8 +77,17 @@ class TopMenu
      */
     public function setPlaceholders()
     {
+
+        if($this->modx->getOption('manager_use_fullname') == true) {
+            $userProfile = $this->modx->user->getOne('Profile');
+            $username = $userProfile->get('fullname');
+        } else {
+            $username = '';
+        }
+
+        if(empty($username)) $username = $this->modx->getLoginUserName();
         $placeholders = array(
-            'username' => $this->modx->getLoginUserName(),
+            'username' => $username,
             'userImage' => $this->getUserImage(),
         );
 
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1969,4 +1969,13 @@ $settings['auto_isfolder']->fromArray(array (
     'area' => 'site',
     'editedon' => null,
 ), '', true, true);
+$settings['manager_use_fullname']= $xpdo->newObject('modSystemSetting');
+$settings['manager_use_fullname']->fromArray(array (
+    'key' => 'manager_use_fullname',
+    'value' => false,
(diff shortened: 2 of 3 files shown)


manager/controllers/default/header.php
```diff
@@ -77,8 +77,17 @@ public function render()
      */
     public function setPlaceholders()
     {
+
+        if($this->modx->getOption('manager_use_fullname') == true) {
+            $userProfile = $this->modx->user->getOne('Profile');
+            $username = $userProfile->get('fullname');
+        } else {
+            $username = '';
```
rtripault: Am i right to assume that if `manager_use_fullname` is set to `false` neither the username nor the full name would be available ?

rtripault: Nevermind... line 88 takes care of that... time for another coffee (sorry for the noise)

wuschigesetwas: @rtripault no probleme. it's better to make a call than missing a problem :) :+1:

Mark-H:
Thank you for making MODX better @wuschigesetwas  :)

JensWolff:
:+1: :+1: :+1: 
Thanks, guys! :)

wuschigesetwas:
@JensWolff you're welcome :+1: 
@Mark-H  you too. I'm planing to do more on this. I already take a look on issues and will work on it in the future and I'm looking forward to a good cooperation :) :+1:

PR #12529, opened 2015-07-30
Make sure registered CSS/JS assets get loaded when using a "deprecated" manager controller

rtripault:
### What does it do ?

Execute parent method `modManagerController::registerCssJs` in `modManagerControllerDeprecated`.
### Why is it needed ?

At some point, some plugins could register additional CSS/JS/HTML using `$modx->controller->add{Javascript|Html|Css|LastJavascript}` methods.
Using those methods adds data into `modManagerController::$head`, which is never "processed" when the controller is an instance of `modManagerControllerDeprecated` and can result in "breaking" (or at least making it not consistent) a customized manager (additional JS/styling)

The controller can be an instance of `modManagerControllerDeprecated` when : 
1. using "pre 2.2 controller style" (not class based)
2. a controller is not found and the "Could not find action file at" error message is displayed
### Steps to reproduce issue
- create the following plugin

``` php
<?php
/**
 * @var modX $modx
 *
 * @event OnBeforeManagerPageInit
 */

$modx->controller->addHtml(<<<HTML
<script>alert('Want some cake ?');</script>
HTML
);
```
- browse to any non existing controller/action, ie. `manager/?a=theCakeIsALie`
- notice you did not receive the alert message (apply this PR and see the alert message)

### DIFF
--- a/_build/test/Tests/Controllers/DeprecatedControllerTest.php
+++ b/_build/test/Tests/Controllers/DeprecatedControllerTest.php
@@ -1,0 +1,32 @@
+<?php
+
+/**
+ * Tests related to the modManagerControllerDeprecated controller class
+ */
+class DeprecatedControllerTest extends MODxTestCase {
+    /**
+     * @var modManagerControllerDeprecated
+     */
+    public $controller;
+
+    public function setUp() {
+        parent::setUp();
+
+        $this->controller = new modManagerControllerDeprecated($this->modx);
+    }
+    public function tearDown() {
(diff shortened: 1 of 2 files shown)


Mark-H:
Very cool, thanks :)

(Also, great idea to provide that simple test plugin.. that definitely helps verifying if PRs work as expected!)

rtripault:
You are welcome!

> (Also, great idea to provide that simple test plugin.. that definitely helps verifying if PRs work as expected!)

PR #12532, opened 2015-08-03
Grid autosave error msg

### LINKED ISSUE #11946: An error window after grid row autosave fails
Would be good to show an error window when grid autosave of rowedit fails. 

Same behaviour is when panel save fails.

rtripault:
### What does it do ?

When a manager grid has the "autosave" attribute (meaning we could edit/save records editing the grid records), it adds a listener to display an error message when the save fails.
It also provides a new attribute, `autosaveErrorMsg`, which allow developers to define a generic error message in case the process just returns a "failure" without additional message.
Since the content type processor i used to test the feature had some errors, i tried to fix them up.
### Why is it needed ?

Just to give a better UX/visual feedback when error occurs (previously the UX could led the user think a record was properly saved, while it was not the case)
### Related issue(s)/PR(s)
- Fixes #11946

### DIFF
--- a/core/model/modx/processors/system/contenttype/updatefromgrid.class.php
+++ b/core/model/modx/processors/system/contenttype/updatefromgrid.class.php
@@ -18,7 +18,7 @@
 class modContentTypeUpdateFromGridProcessor extends modProcessor {
     /** @var array $records */
     public $records;
-    
+
     public function checkPermissions() {
         return $this->modx->hasPermission('content_types');
     }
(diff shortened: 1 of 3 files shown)


theboxer:
@rtripault This is probably not the correct approach for handling errors from processor.

It works only in one case, when the error is sent by `addFieldError`, but you can also send error message via `->failure('Error message here')`.

So the JS should check for:
- response.message.message (string)
- response.message.errors (array, so join msg values)
- response.data (array, so join msg values)

rtripault:
Makes sense, beside... i never saw cases where `response.message` would be an object (message/errors attributes). Would you mind pointing me to some sample ?
Thanks

theboxer:
Sorry, not sure what was with my Revo... I don't see message object anymore. I updated the previous message

rtripault:
OK, thanks for the clarification.

By the way, while implementing your proposed solution, i found that the listener would better be in a method/function, which would allow developers to override the logic to match their potential custom need (ie. `this.on('afterAutoSave', this.onAfterSave, this)`).

Any objection/opinion ?

theboxer:
Nice catch, it should definitely be there!

PR #12515, opened 2015-07-14
New uberbar search

pepebe:
New uberbar search with enhanced search functionality. The code still contains some legacy stuff that might be useful in the future.

Issue #12283, opened 2015-01-21
Uberbar UI enhancements


pepebe:
I have a simple PR for some minor enhancements for Uberbar: https://github.com/modxcms/revolution/pull/12282

The PR defaults to the basic search in (pagetitle,introtext,etc), but also allows searching for a specific id/template/pagetitle/classkey. 

```
id::31,
pagetitle::home, 
pagetitle:LIKE::%sch%,
parent::15 // useful if children are hidden from the tree by collections
class_key::mgResource, 
template::3
```

It also adds the id of each resource/element to the search result.

![bildschirmfoto 2015-01-20 um 20 48 18](https://cloud.githubusercontent.com/assets/667315/5833360/eae9ef84-a152-11e4-8534-efa8b14da6a1.png)

The new search will return results for almost all searches listed here 
http://rtfm.modx.com/xpdo/2.x/class-reference/xpdoquery/xpdoquery.where

Not included at the moment are IN, NOT IN and IS as well as searching for multiple parameters. 

For multiple parameters I would like to do it like is done in getResources:

```
{"template:=":1, "OR:template:=":2}
```

I can see some use cases for these features for my personal needs, but I'm not sure if anyone else will ever need them.

I need some feedback before I go all out :)

Cheers,

pepebe

pepebe:
I've added a few modifications to the initial PR.

* search elements/users by id
* added uberbar_maxresults system setting for overwriting the default number of 5 results/category.

rtripault:
While i like the idea, and that is something planned/discussed (see #11117 & #11116), i don't like the implementation, mostly because `pagetitle::$value` have absolutely no meaning for elements (they don't have a `pagetitle` column).
Keep in mind the plan is also to allow developers to "plug" into the search function (see PR #12041), to allow searching custom objects, or just put their own logic/syntax (to be discussed, but that's how i would like to see this feature going).

Also, do you really want to write a JSON criteria ? I doubt that would be useful to many users (that would certainly serve "advanced" people like you, but doubt it would be "understandable" for regular "editors").

We might (should ?) have a look at other "well established search engines" syntax (thinking ElasticSearch, SphinxSearch) to mimic theirs.

But being able to use advanced criteria would definitely be awesome, and might need some discussion upfront.

Hope that makes sense.

pepebe:
Hi,

thanks for your feedback.

[...]

PR #12533, opened 2015-08-04
Advanced uberbar 

Mark-H:
This pull request follows #12515. Existing commits by @pepebe have been cherry-picked, and new ones will be added per the discussion in #12515. 
### What does it do ?

Adds advanced capabilities to the uberbar, allowing granular search. 
### Why is it needed ?

Making the uberbar more useful than it already was, also incorporating the 
### Related issue(s)/PR(s)
#12515, #12283, #11117,

Somewhat related but not entirely: #11116, #12041
### Work in Progress
- [x] Loose search by default, typing in a partial name and getting relevant results is way more intuitive. Gotta think of the end user here too.
- [x] Don't require closing colons. `Acc` (or `Acc*`) should be a valid search, rather than needing to use `Acc*:`.
- [x] Similar to above, but when doing a specific search like `:r:pagetitle:acc` the last colon should also not be required to move to the next step; simply assume the current step is complete if there is a valid value.
- [x] Some searches (`:u:*:m*:`) seem to break the output while more specific ones work (`:u:username:m*:`)
- [x] The search processor needs to be refactored to make it more maintainable and less spaghetti


pepebe:
Thanks for advancing this. I'll have a look at the open issues (hopefully) during the weekend.

core/model/modx/processors/search/search.class.php
```diff
@@ -18,40 +33,584 @@ class modSearchProcessor extends modProcessor
...
+            );
         }
+    }
 
-        return $this->outputArray($this->results);
+    /**
+     * Validates the uberbar to check for permissions, allowed elements and fields etc.
+     *
+     * @param $query
+     * @return bool
```
rtripault: Seems like it also returns some array

core/model/modx/processors/search/search.class.php
```diff
@@ -18,40 +33,584 @@ class modSearchProcessor extends modProcessor
...
+    {
+        $mode = $this->modx->getOption('uberbar_mode');
+        $count = count($query);
+        $element = isset($query[1]) ? $query[1] : false;
+        $field = isset($query[2]) ? $query[2] : false;
+
+        // Element related stuff...
+        if ($count > 2) {
+            /* Validate elements */
+            if ($element != '*' && !array_key_exists($element, $this->elements)) {
```
rtripault: What if we want to "extend" the search and search within custom tables ? (see my attempt in #12041)

core/model/modx/processors/search/search.class.php
```diff
@@ -18,40 +33,584 @@ class modSearchProcessor extends modProcessor
...
+                foreach ($this->elements as $key => $val) {
+                    /* Only show elements you have permission to see. */
+                    if ($val['permission'] === true) {
+                        $activeClass = ($key == $query[1]) ? 'active ' : '';
+                        $options .= '<li><a data-value="' . $key . '" class="' . $activeClass . $btn . '">' . $val['label'] . '</a></li>';
+                    }
+                }
+                $result['options'] = $options;
+
+                $this->results[] = $result;
```
rtripault: I guess we could stop the execution once the result has been added (no chance `$count` will be different than `2` here, right ? ;))

Mark-H:
@rtripault please do continue reviewing, but just FYI I haven't gotten to refactoring that yet and that is definitely going to look difference once I get my hands dirty :P

core/model/modx/processors/search/search.class.php
Mark-H: I was thinking your event would be a good fit to add to the `_getElements` method (bottom of the class) - that is used to fill up the `$this->elements` array earlier on.

rtripault:
o'right, then i guess doesn't worth the trouble reviewing yet...

core/model/modx/processors/search/search.class.php
rtripault: While the idea seems good, i'd say that won't work : 

1. possibly allow "overrides" (as in, what if i register another element with the `r` trigger)
2. i would have no control over the action, since the query just append an `&id=` to `$this->elements['trigger']['action']` (see l.310-320 & l.357). objects might have a different PK.

Mark-H:
Alright @rtripault, I think I'm mostly done now - would appreciate a review :) 

In terms of making the code more manageable and not in the least extendable, I was thinking maybe the logic that is currently in the search() method could instead defer to a callback of sorts. That could be hardcoded for the core supported elements (e.g. a `searchSnippets()` method), and maybe the OnSearch plugin could pass the callable in the event output for custom elements. 

Haven't exactly fleshed out that idea yet but that might make it more flexible.

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -97,25 +110,35 @@ MODx.SearchBar = function(config) {
...
+                }
+            }
             ,scope: this
         }
+
     });
+
     MODx.SearchBar.superclass.constructor.call(this, config);
-    //this.setKeyMap();
+    this.setKeyMap();
```
pixelchutes: I worry about this re-opening these: #11794 and #11828 -- however, I also would like this functionality restored. Not sure how to prevent this issue for impacted keyboard regions while still enabling this, unless the shortcut itself is changed to something non-conflicting?

pepebe:
Hi,

I've read the comments above but I didn't have a chance to run the changes commited by Mark. Here are my five cents...

Let me start by saying that I'm perfectly aware that this code can be improved in a zillion ways. But as it is right now, it works and does what it is supposed to do. How about we fix minor issues, let the whole concept brew for a while and then come up with version 2.

@Mark-H You are a coding monster, but You have your limits. I'd rather see some awesome updates to contentblocks or redactor. ;)

> Loose search by default, typing in a partial name and getting relevant results is way more intuitive. Gotta think of the end user here too.

 It would be nice to have some feedback from end users how they react on the proposed syntax. If this really turns out as a bad idea, we can slim it down a bit but I don't have any idea how to handle pagination in that case.

> Don't require closing colons. Acc (or Acc*) should be a valid search, rather than needing to use Acc*:.

I don't agree that removing the colon trigger is a good idea because it can't hurt to think about what you want to have before going to the next step and it also prevents unnecessary request.

> Similar to above, but when doing a specific search like :r:pagetitle:acc the last colon should also not be required to move to the next step; simply assume the current step is complete if there is a valid value.

The last colon triggers the final search as well as pagination. The colons before that one trigger permission checks and other stuff. I don't see why the colons are such a pain.

> Some searches (:u:*:m*:) seem to break the output while more specific ones work (:u:username:m*:)

Thats a bug. They all should work. I'll have a look.

> The search processor needs to be refactored to make it more maintainable and less spaghetti

No doubt about that. I've started with minor changes to the original processor and ended up with a bloated monster. Guilty.

Cheers,

patrick

Mark-H:
Hey Patrick, the primary reason Ive been rushing this through is that 2.4 is really close (if this isn't merged in the next few days it might have to wait for 2.5) and I'd like to get it in there. We can definitely keep making changes to it, but need a solid base line. I think we're there. 

All the items in the to do list have been updated in this pull, so definitely check it out and report back with your thoughts. I know you've spent a lot of time on this, for which I say thanks :)

manager/assets/modext/widgets/core/modx.searchbar.js
Mark-H: That's a good point. I think it might be hard to find a shortcut that does not conflict with anything, so maybe the best fix is to get rid of the "opening uberbar" shortcut? The primary reason I renabled that was for closing the uberbar with ESC.

Mark-H: I've updated the keymap per https://github.com/modxcms/revolution/commit/739ce32b5b50752b408e42a3d442493d17ac68dc

pixelchutes: Agreed. Enabling the keymap (to allow ESC closing as you say), but remove the conflicting shortcut would cover both concerns.

theboxer:
Great work on this, unfortunately I don't think this is ready for 2.4, for few reasons:

- Search without modifiers searches only in Resources, not in everything like it was before
- UI for building advanced (with modifiers) syntax such as `:s:*:test:` is not working properly and it's not creating a valid syntax - you can end up with something like `:s:*nameid`, or `:stvctpu`
- UI for pagination doesn't work - creates invalid syntax - searching for `test` and than paginating produces `test2`, `test23`, etc. instead of `test:2` and `test:3`
- Lexicons for system settings are missing

This was just brief testing, there might be more issues.

pepebe:
I'm mighty confused about most of the points mentioned by theboxer. I'll compare the version commited with my dev version. 

Apart from this, I guess there are a few things that need explanation.

## Syntax
Either I've messed something up along the way, or I failed to explain what is possible and what is not.

Wildcards in step 1 (table) and 2 (fieldname) can only be used to search in all tables and/or all fields. It is not intended to search for (:*ippet) (sn)ippet or to look for (*agetitle) (p)agetitle. From my point of view this doesn't make sense.

## Default search
I believe the default search should be something targeted at clients/editors. Limiting it to resources, should cover almost all use cases.

That the default search isn't searching in all tables anymore is intentional. It never made any sense except quick-searching for the title of a resource, snippet, etc.

As a compromise between power and comfort It could as well be upgraded to search in all elements you have permission for and in all available fields (:*:*:Foo:1). This would limit the search depending on the current permissions. I guess this would mean for most clients searching only in resources (like it was before) and for power users searching everywhere.

## Advanced search
The advanced search offers a simplified syntax that covers many use cases and is very easy to learn.

If somebody wants to have a full sql like syntax or even a regex search, then this should be implemented into future versions or third party plugins.

rtripault:
While i did not test the code (just reviewed it), this seems a lot better to me (i would still love more PHPDoc).

@Mark-H 
> maybe the OnSearch plugin could pass the callable in the event output for custom elements.

I agree the `search` method is still to much convoluted.

To me, if i were to implement the "custom models" search, with the processor in its current state i would : 

1. fire an `OnBeforeSearch` event in `_getElements`, so that we could add additional elements "configuration" (our custom models) which would then be available/usable in `enterElement` & `enterField`. The output should then be merged in `$elements` if there is no array key "conflict" (else we would rise a notice/warning in the error log)
2. fire an `OnSearch` event in `search` method, expecting an array of results as output.

Once the `OnSearch` event is fired, it's no more the "search" processor responsibility to handle the search/query/joins/model(s) loading. I believe the plugin should just return an array of results. The processor would then "formats" the results output and the pagination.

@pepebe 
> I believe the default search should be something targeted at clients/editors. Limiting it to resources, should cover almost all use cases.

While i get the idea, i strongly disagree here. ACLs should take care of what the current user should be able to search in.

> As a compromise between power and comfort It could as well be upgraded to search in all elements you have permission for and in all available fields ( : * : * : Foo : 1 )

I believe this should be the default, but since i understand your idea/purpose of limiting to resources only, we could make use of a setting, defaulting to `:*:*:`, but configurable to anything, like `:r:*:`

#### Food for thought

While thinking about "features" i'd like to see coming next, i would love to : 

* be able to have some contextual search (ie. when on CMP x, i expect to search only for y & z models only)
* use a search engine rather than SQL queries (ie. ElasticSearch, SphinxSearch)
* be able to search in multiple fields (ie. `where modResource.template = 2 AND modResource.parent IN [3, 4]`)

Maybe this could "easily" be doable if we were making use of a configurable search "service class" (something similar to how we can change the "request"/"response" classes).

core/model/modx/processors/search/search.class.php
```diff
@@ -1,57 +1,534 @@
...
+     * - content: the field that contains the content, can be shown in the results
+     * - icon: a font awesome icon to use for results of this type
+     */
+    protected function _getElements()
+    {
+        $this->elements['s'] = array(
+            'label' => '(s)nippet',
+            'type' => 'snippets',
+            'class_key' => 'modSnippet',
+            'permission' => $this->modx->hasPermission('view_snippet'),
```
rtripault: Why not populating `$elements[$key]` only if the permission is available, ie. `if ($this->modx->hasPermission('view_snippet')) { $this->elements['s'] = [...] }`. That would prevent iterating over "models" user has no access to

core/model/modx/processors/search/search.class.php
```diff
@@ -1,57 +1,534 @@
...
+    public function enterField()
+    {
+        $result = array();
+
+        $result['header'] = $this->modx->lexicon('uberbar_pick_fieldname_header');
+        $result['msg'] = $this->modx->lexicon('uberbar_pick_fieldname_msg',
+            array('uberbar_mode' => $this->mode));
+
+        $options = '';
+        $options .= '<li><a data-value="*" class="' . $this->btnClasses . '">All (*)</a></li>';
```
rtripault: Shouldn't `All (*)` be a lexicon string ?

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -11,77 +10,86 @@ MODx.SearchBar = function(config) {
...
+                // Display header only once
+                ,       '<tpl if="this.type != values.type">'
+                ,           '<tpl exec="this.type = values.type; values.label = this.getLabel(values.type)"></tpl>'
+                ,           '<h3><i class="icon icon-{icon}"></i> {label}</h3>'
+                ,       '</tpl>'
+                        // Real result, make it use the default styles for a combobox dropdown with x-combo-list-item
+                ,       '<p class="x-combo-list-item">' // x-combo-list-item
+                ,           '<span>#{current}</span>'
+                ,           '<a class="x-btn x-btn-small x-btn-icon-small-left primary-button x-btn-noicon" href="?a={_action}">'
+                ,               'Edit'
```
rtripault: Shouldn't this (`Edit`) be a lexicon string ?

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -89,6 +97,11 @@ MODx.SearchBar = function(config) {
                         return false;
                     }
                 }
+                ,load: {
```
rtripault: Why this listener ?

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -97,45 +110,52 @@ MODx.SearchBar = function(config) {
                     this.tpl.type = null;
                 }
             }
-            ,focus: this.focusBar
-            ,blur: this.blurBar
+            ,focus: {
+                fn: function(){
+                    this.focusBar();
```
rtripault: What's reason to wrap those functions into function ?

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -198,18 +216,34 @@ Ext.extend(MODx.SearchBar, Ext.form.ComboBox, {
...
+                    if (Ext.get(t).hasClass('active')) {
+                        var uberGet = Ext.get('modx-uberbar'),
+                            uberGetCmp = Ext.getCmp('modx-uberbar'),
+                            oldValue = uberGet.getValue(),
+                            newValue = oldValue + dataValue;
+
+                        uberGetCmp.setValue(newValue).focus(false, 200);
+                        uberGetCmp.fireEvent('change', uberGetCmp);
+                    }
+                    else {
```
rtripault: Is there any need for that empty `else` ?

manager/assets/modext/widgets/core/modx.searchbar.js
pixelchutes: I am also curious about this one.

core/model/modx/processors/search/search.class.php
Mark-H: Good idea

manager/assets/modext/widgets/core/modx.searchbar.js
Mark-H: If I did that, it was probably to add the scope to it - but doesn't look like that's the case as the scope is defined one element up. Not sure.

pepebe:
Sorry for showing up late.

There is a lot of legacy stuff included in the js. There are also a couple of experimental things and 2dos. This version was never meant to be final or perfect, so I don't mind cleaning up the whole thing and stripping it down to the bare essentials. I'll have a look this week at the current state and try to clean things up where necessary.

Mark-H:
I'm going to close this (will leave the branch on my fork in case anyone wants to continue on it), in favour of taking smaller steps like in #12783.

PR #12534, opened 2015-08-05
Convert country lexicons to ISO 3166-1 alpha-2 country codes

### LINKED ISSUE #11391: Feature: Use ISO 3166-1 alpha-2 country codes in core
Gone are the days of managing a conversion list of numeric MODX Country IDs _(e.g. 223 => United States)_, but I am curious why they weren't simply replaced with the ISO 3166-1 country codes? I'm sure there was some discussion on the matter, and I'm even willing to be convinced otherwise... IMO, storing the country as a two character ISO code would be far more useful than storing the entire country name for each User record, especially given the many circumstances where a simple short-code can be easily used programmatically vs. full text.

A few simple examples:
* Country flag icons (using 2-char ISO naming convention)
* Compatibility with 3rd party services / APIs requiring ISO codes.
* Less space allocated in database

After some quick searching, I was not able to find any discussions about why MODX may have chosen not to incorporate ISO 3166-1, but I was able to find others who might just feel the same way: /cc @Mark-H @bertoost

While you can technically store a User's country as ISO 3166-1 (alpha-2) country code out-of-the-box, the experience is still disconnected. For example, the manager at least preserves a custom country value, but "US" is not displayed as "United States". Further, if you want to manually type in the ISO code for Canada (CA), you can't.

Possibly related to #2253 and http://bugs.modx.com/issues/5387 (not seeing #5387 on GH?)

pixelchutes:
### What does it do ?

Introduces ISO 3166-1 alpha-2 country code lexicons, now stored as ISO code in `modUserProfile.country` vs. localized country values.
### Why is it needed ?

Provides more versatility, specifically with regards to translations. Also, localized values in the database should be avoided.

Example:

When you need to show the manager in a different language than the front-end where User's can change their country based on their preferred language. With this change, rendered country names will correctly honor `manager_lang` and the User's language simultaneously.
### Related issue(s)/PR(s)

Fixes #11391
Fork of #12020 w/ cherry-picked commits by @goldsky 
### Note

Integrators / testers should be aware that `_country_lang` key changes not yet landed from Crowdin may result in ISO code conversion not taking place (depending on your defined `manager_lang`) Until then, test using `en` country lexicon file plus matching country values during upgrade conversion process.


core/model/modx/processors/system/country/getlist.class.php
```diff
@@ -1,37 +1,45 @@
...
-        if (empty($countryList)) return $this->failure();
+        if (empty($countryList)) {
+            return $this->failure();
+        }
 
         $countries = array();
-        foreach ($countryList as $country) {
+        foreach ($countryList as $iso => $country) {
             $countries[] = array(
-                'value' => $country,
```
Mark-H: Maybe it's worthwhile keeping the `value` option here as well for backwards compatibility, just in case some code relies on the value from the processor response?

manager/assets/modext/widgets/core/modx.combo.js
```diff
@@ -714,9 +714,9 @@ MODx.combo.Country = function(config) {
         ,baseParams: {
             action: 'system/country/getlist'
         }
-        ,displayField: 'value'
-        ,valueField: 'value'
-        ,fields: ['value']
+        ,displayField: 'country'
+        ,valueField: 'iso'
+        ,fields: ['iso', 'country']
```
Mark-H: Related to other comment too ^

core/model/modx/processors/system/country/getlist.class.php
pixelchutes: Great suggestion! I agree.

PR #12535, opened 2015-08-05
Pre fill fields in resource

jakub-gawryl:
This patch gives possibility to use $resource variable in plugin with "OnDocFormRender" event to predefine some inputs, when creating new resource (or update existing one).

So, if in plugin with "OnDocFormRender" we use:

``` php
$resource->set('content',"Hey, i'm default content! You can change me to anything you want...");
```

then we get pre filled "content" field after "New resource" or "Update Resource" click.

Fields, which can be pre filled:
- pagetitle
- longtitle
- description
- introtext
- content
- link_attributes
- alias
- menutitle

**1. Data will NOT be saved into database until click "save" button!
2. If you update resource, remember that, this overwrite your content received from database (you'll see this _Hey, i'm default content..._ text rather than received one!**

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -105,6 +105,14 @@ class ResourceCreateManagerController extends ResourceManagerController {
                 'cacheable' => $this->context->getOption('cache_default', 1, $this->modx->_userConfig),
                 'syncsite' => true,
             ));
+            
+            $newValuesArr = Array();
+            $allowedFields = Array('pagetitle','longtitle','description','introtext','content','link_attributes','alias','menutitle');
+            foreach ($allowedFields as $field) {
+                $newValuesArr[$field] = $this->resource->get($field);
+            }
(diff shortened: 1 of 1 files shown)


Mark-H:
Hi @Jacqbus that's a great addition for 2.4! I've also added a quick example to the docs on how one might use this at https://rtfm.modx.com/revolution/2.x/developing-in-modx/basic-development/plugins/system-events/ondocformrender

For the next time you send a pull request, could you follow the template in https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request? Thanks!

Jacqbus:
Yeah no problem - i'll use that template with all my future requests :)
I'm also really glad you're like this patch - thanks! :)

Jacqbus:
Should this pull request attached to v2.4.0-pl milestone? I'm unspeakably glad i could add some nice small piece to that great project, and this is a great motivation for me to still improve modx with all of You :)

PR #12545, opened 2015-08-10
Fixing Language string not found: ""

Jako:
It makes no sense to search for an empty lexicon key and it creates debug warnings `Language string not found: ""`


Mark-H:
Thanks @Jako for this and #12546 :) Both ready for 2.3.6.

PR #12551, opened 2015-08-14
bugfix installed package list in its detail's page

goldsky:
Re:  #11125
### Summary

bugfix installed package list in its detail's page when the package's name has spaces.
http://forums.modx.com/thread/88734/package-version-check#dis-post-489531
### Step to reproduce
1. Download lower version of any package that has spaces on its name, eg: [Grid Class Key](http://modx.com/extras/package/gridclasskey), [TinyMCE Rich Text Editor](http://modx.com/extras/package/tinymcerichtexteditor)
2. Search it locally.
3. Install it.
4. Open your DB admin, go to `modx_transport_packages` table, and edit its package name to the one like what you should get if it's downloaded from remote, eg: **Grid Class Key**, or **TinyMCE Rich Text Editor** (**...with spaces!**). (installing them from local searching also has another issue, it's not picking up the correct package name -- such if you download it from remote provider, but rather it's signature.).
5. Download again its higher version.
6. Upgrade it.
7. Edit its package name again.
8. Now, back to the Package list, click "view details". You won't find the lower version in the "Uploaded Versions" tab.
### Observed behavior

If you click "view details" of the package in Package Management/Installer, you can't find the lower version of those packages.
### Expected behavior

It should list the lower version of packages.
### Environment

MODX 2.3.5-pl sdk

### DIFF
--- a/core/model/modx/processors/workspace/packages/version/getlist.class.php
+++ b/core/model/modx/processors/workspace/packages/version/getlist.class.php
@@ -31,7 +31,7 @@ class modPackageVersionGetListProcessor extends modObjectGetListProcessor {
         /* get packages */
         $criteria = array(
             'workspace' => $this->getProperty('workspace',1),
-            'package_name' => $signatureArray[0],
+            'package_name' => urldecode($this->getProperty('package_name',$signatureArray[0])),
         );
         $limit = $this->getProperty('limit');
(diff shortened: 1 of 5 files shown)


theboxer:
@goldsky Would you mind resubmitting the PR again to 2.x branch? Looks like some other stuff got into it, when 2.4 was released, so it doesn't look clear.

goldsky:
I have updated my repo, try to diff it.

PR #12555, opened 2015-08-17
modParsedManagerController (base manager controller class) making of the regular parser

rtripault:
### What does it do ?

Adds a new abstract manager controller to ease the process of building CMPs with "tools" every MODX developers are used to (MODX tags, so you can build a CMP using chunks and snippets).
### Why is it needed ?

Developers often "complain" about how difficult it is to build CMPs. With this base controller, we are now able to build a CMP making use of chunks & snippets (and other elements the parser "understands").
### To do before merge
- [x] refactor `HelpManagerController` to make use of this new controller class


pixelchutes:
This looks great @rtripault! Seems so much better (+more intuitive) than some of the current approaches for bringing Elements/Lexicons into scope in order to use with Smarty / CMPs.

Mark-H:
Hi @rtripault - I assume you're going to talk about this in your talk at the MODXpo tomorrow so I've just merged it into 2.5. Nice work, I can think of a lot of people that will appreciate this!

Mark-H:
Merge commit was https://github.com/modxcms/revolution/commit/7f3679429577fcb59a4152d588a6625b1dd6dffa

Issue #12541, opened 2015-08-09
Package dependency not installing as expected


Mark-H:
Figured I'd give package dependencies a try, but so far it's not working.

The package in question is Mollie for SimpleCart, which has the following in the build script:

```` php
$attributes = array(
    'readme' => file_get_contents($sources['source_core'] . '/docs/readme.txt'),
    'changelog' => file_get_contents($sources['source_core'] . '/docs/changelog.txt'),
    'setup-options' => array(
        'source' => $sources['install_options'] . 'input.options.php',
    ),
    'requires' => array(
        'simplecart' => '>=2.2.15',
    )
);
foreach ($attributes as $k => $v) {
    $package->setAttribute($k, $v);
}
````

Full source available at https://github.com/modmore/SimpleCart_Mollie. 

It uses a slightly different way of packaging, but the requirements are properly added to the package. The package manager also shows the dependencies, as expected:

<img width="906" alt="screen shot 2015-08-10 at 00 24 20" src="https://cloud.githubusercontent.com/assets/312944/9157251/e1f3700c-3ef6-11e5-91c3-c236d1132a46.png">

Interestingly SimpleCart is already downloaded to this particular installation, it's just not installed as I wanted to try the dependencies. 

When I hit install, I get the familiar installer screen for SimpleCart:

<img width="903" alt="screen shot 2015-08-10 at 00 31 22" src="https://cloud.githubusercontent.com/assets/312944/9157254/27999c94-3ef7-11e5-95ec-f8c6964ab17a.png">

Setup options also show up as expected when clicking the button, indicating it managed to find the package and is working with the unzipped manifests. So far so good.

Then this happens:

````
Console running...
Attempting to install package with signature: simplecart_mollie-1.1.2-pl
Package found...now preparing to install.
Grabbing package workspace...
Workspace environment initiated, now installing package...
No filter found.
PHP notice: Undefined index: signature
Could not load package info from name for , not yet implemented.
Package simplecart_mollie-1.1.2-pl requires simplecart @ >=2.2.15
Could not install package with signature: simplecart_mollie-1.1.2-pl
````

It looks like it's trying to install the gateway rather than the core package as instructed. The error `Could not load package info from name for , not yet implemented.` comes from the logging statement I added in #12540 but as it already has the package locally, I'm not sure why it would be trying to find package info from the provider again. 

[...]

PR #12556, opened 2015-08-17
Pass signature via setup options

theboxer:
### What does it do ?

Passes signature via setup options window, when setup options are present.
### Why is it needed ?

Without this change, when installing dependency with setup options, it doesn't pass current signature and installer starts installing parent package, which fails, because of unresolved dependencies.
### Related issue(s)/PR(s)
- Package dependency not installing as expected #12541


Mark-H:
You da man, man!

Issue #12553, opened 2015-08-16
 Old Resources browser CSS loaded when compress_css is turned on 


vierkantemeter:
### Summary

When compress_css is turned on, the MODX resources browser is loading in an old CSS (looks like from 2.2). It will look like this:

![voila_capture 2015-08-16_12-46-47_pm](https://cloud.githubusercontent.com/assets/487567/9292961/702c0780-4415-11e5-96ad-6dbe59ff07f2.jpg)

This only happens with the browser which opens when using the insert image/file button in TinyMCE, not with the  'Media > MODX browser' function.

This is caused by:
`{$_config.manager_url}templates/default/css/xtheme-modx.css` 
in 
`<link rel="stylesheet" type="text/css" href="{$_config.manager_url}min/index.php?f={$_config.manager_url}templates/default/css/xtheme-modx.css,{$_config.manager_url}templates/default/css/index.css" />` in the file manager/templates/default/browser/index.tpl

Changing it to the following, makes all look good again:
`<link rel="stylesheet" type="text/css" href="{$_config.manager_url}min/index.php?f={$_config.manager_url}templates/default/css/index.css" />`
### Step to reproduce

Turn on compress_css and open the resources browser (using TinyMCE)
### Environment

MODX 2.3.5

I tried to create a pull request, but I don't know to which branch I should post it, so some of you have to fix it.

Mark-H:
If I tell you it should target the master branch because it's a bug fix, is that enough to convince you into sending a pull request? :P

vierkantemeter:
Thanks, just did!

vierkantemeter:
This is fixed in 2.4 I see.

Mark-H:
This was included in 2.4

PR #12558, opened 2015-08-17
Old Resources browser CSS loaded when compress_css is turned on

vierkantemeter:
When compress_css is turned on, the MODX resources browser is loading in an old CSS (looks like from 2.2). It will look like this:

voila_capture 2015-08-16_12-46-47_pm

This only happens with the browser which opens when using the insert image/file button in TinyMCE, not with the 'Media > MODX browser' function.

MODX 2.3.5

Related issue: https://github.com/modxcms/revolution/issues/12553


theboxer:
This is already included in 2.4.x branch

PR #12563, opened 2015-08-18
Improved generateConfig script for ability use specified user and pas…

alroniks:
…sword for database
### What does it do ?

I added dbuser and dbpass variables to script that generate config and install modx before tests. And fixed phpunit.xml, replace deprecated param "strict".
### Why is it needed ?

It's allow faster set up environment and run tests. User root and empty password for database is not best way for me and others, so ability to use own credentials would be good, I think. But default behavior was saved.
### Related issue(s)/PR(s)

No relations.


opengeek:
This was merged into the `2.4.x` branch at commit 8520507

PR #12572, opened 2015-08-21
Added check of "modx-resource-id" in treedrop

bezumkin:
In MODX 2.4.0-pl drag-n-drop elements to template content is broken because of usage of undefined variable on line 227.
[](https://file.modx.pro/files/0/8/d/08d37129cf625cee84d68dddc2d4127a.png)

This fix add check of "modx-resource-id" availability


theboxer:
Merged to `2.4.x`

Issue #12564, opened 2015-08-19
CLI installation notices


Jako:
### Summary

There are two PHP Notices during CLI install
### Step to reproduce

Install by CLI (maybe during setup routine too)
### Observed behavior

Two notices:

```
Undefined index: primary in /srv/www/.../html/core/xpdo/om/mysql/xpdomanager.class.php on line 462
```

```
Undefined index: unique in /srv/www/.../html/core/xpdo/om/mysql/xpdomanager.class.php on line 462
```
### Expected behavior

No notices.
### Environment

2.4.0, PHP 5.4

bezumkin:
Confirm

Alroniks:
Also confirm. 
```
PHP 5.5.9-1ubuntu4.11 (cli) (built: Jul  2 2015 15:23:08)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
   with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
```

Mark-H:
those notices sound like it may be related to an index in the model or something? Are these new installs or updates?

bezumkin:
Only new install in a cli mode. On update I don`t see these notices.

Jako:
New installation.

pixelchutes:
Curious if it could be related to `modManagerLog`? I see the following error during an upgrade from 2.3.x > 2.4.0:

> 
> Error updating table for class modManagerLog
> Array
> (
> [0] => 00000
> [1] => 
> [2] => 
> )

Not sure if the index in question has to do with this table or not?

pixelchutes:
Curious if this is fixed by #12622 ...can anyone confirm?

PR #12575, opened 2015-08-21
Fix CLI installation notices (#12564) due to undefined index type

krismas:
### What does it do ?

The index type for `modTemplateVarResourceGroup / tmplvar_template` is undefined since 5599467

Update MODX xPDO model to add index type definition (primary and unique). 

Also update `getIndexDef()` to replace these warnings by a log entry.
### Why is it needed ?

Fix CLI installation notices (#12564) due to undefined index type
### Related issue(s)/PR(s)
#12564 : CLI installation notices


krismas:
Need to be on master branch (bugfix)

Issue #11472, opened 2014-06-06
Add date joined to modUserProfile


wshawn:
I have asked for this a few times before.  It only makes sense for a a system with the power of our modUser / ACLs / Groups / etc to have access to when the account was created.

I understand this would not be retroactive, but it would really be nice in the future.

We have multiple dates for each resource, but only dates related to login and bans for users.

The blocked after could be used for subscription sites...  Etc

Nuan8:
Has anything happened with this?

wshawn:
As for me, I have had little time to pursue this. I got one of requests for xPDO knocked out and waiting for inclusion, but trying to get a multi million dollar site launched so I can relax a bit and get paid while knocking some of these out.

Issue #4591, opened 2011-05-04
Add active and added user dates


bravado:
bravado created Redmine issue ID 4591

Add date fields to database when user was made active and date when user was added to system.  Sometimes the dates will differ on when a user was added versus when they were made active.  Having this information available will greatly help in managing users in the system.  :)

whitebyte:
This issue itself is a duplicate of #4055

whitebyte:
...and should be closed in favor of #11472.

rtripault:
While not entirely "duplicate" of #11472, i'm closing this one in favor of mentioned issue.

Please feel free to add your voice/comment there.

Thanks

Issue #4055, opened 2011-03-14
Registered On date


modxbot:
danny_kay1710 created Redmine issue ID 4055

It would be useful if the login system recorded the date and time on which a user signed upto the website. This would allow you to track how many registrations have came as a result of a PR or Marketing push or something similar.

As a quick fix for my requirements I have altered the database schema - I have added a createdon field to the modx_users table with the default value set to CURRENT_TIMESTAMP. This doesn't allow me to view this information through ModX manager however it does allow me to get the information I need through an SQL query which I am more than happy to do for the time being.

It would be great if this was available out of the box.

whitebyte:
This issue should be closed in favor of #11472.

pixelchutes:
Closing as dupe.

PR #12581, opened 2015-08-23
Add `createdon` date field to modUser (cf. #11472)

krismas:
### What does it do ?

Update MODX xPDO model, Lexicon, modUser class and the ExtJS panel to deal with this new column.
### Why is it needed ?

This PR is related to "#11472 - Add date joined to modUserProfile"
### Related issue(s)/PR(s)
- #8461 : User createdon 
- #12119 : User creation date 
- #4055 : RegisteredoOn date 
- #4591 : Add active and added user dates

### DIFF
--- a/core/model/modx/moduser.class.php
+++ b/core/model/modx/moduser.class.php
@@ -15,6 +15,10 @@
  * @property json $remote_data Used for storing remote data for authentication for a User
  * @property string $hash_class The hashing class used to create this User's password
  * @property string $salt A salt that might have been used to create this User's password
+ * @property int $primary_group The user primary Group
+ * @property array $session_stale
+ * @property int $sudo If checked, this user will have full access to all the site and will bypass any Access Permissions checks
+ * @property int $createdon The user creation date
  *
  * @property modUserProfile $Profile
  * @property modUserGroup $PrimaryGroup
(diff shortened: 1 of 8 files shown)


bezumkin:
I really don`t understand why this field not exists by default from first versions of MODX.

+1

sottwell:
Just one of those little annoying things that wasn't quite annoying enough reach the 'must-do' threshold.

+1

pixelchutes:
:beers: 

+1

core/model/modx/moduser.class.php
```diff
@@ -81,6 +85,7 @@ public function setSudo($sudo) {
      */
     public function save($cacheFlag = false) {
         $isNew = $this->isNew();
+        if ($isNew) $this->set('createdon', time());
```
Mark-H: When dealing with import scripts and such, checks like this prevent you from setting a predefined createdon date. It might be better if this is changed to something like:

```` php
if ($isNew && $this->get('createdon') < 1) {
   $this->set('createdon', time());
}
````

That way if the createdon is already set, it isn't overwritten.

**Comment by Mark-H** [comment]
@krismas could you please follow the template from https://github.com/modxcms/revolution/blob/master/CONTRIBUTING.md#submit-a-pull-request when submitting pull requests?

File: core/model/modx/moduser.class.php (hunk 1 above)
Comment (krismas) [inline core/model/modx/moduser.class.php]: @Mark-H You are right. From what I see `modresource.class.php` use: 
```
        if ($this->isNew()) {
            if (!$this->get('createdon')) $this->set('createdon', time());
        }
```
What is your preference?

krismas:
> @krismas could you please follow the template from

@Mark-H OK, I will try to make a new PR ASAP

Mark-H:
@theboxer were you planning to do anything here?

theboxer:
Merged to 2.x branch

Issue #12582, opened 2015-08-23
Clear error log... not really


krismas:
After executing a "Clear" action on the error log, the file `error.log` and the return string contain a "space".

PR #12583, opened 2015-08-23
Really clear the log...

krismas:
Remove the extra space when clearing « error.log » file via `writeFile()` (cf. #12582)


krismas:
Need to be on master branch (bugfix)

Issue #12579, opened 2015-08-23
Problems with debug System Setting


BobRay:
According to the setting description (as I read it), setting debug to 1 is supposed to set the error_reporting level to -1, which would show all errors. It doesn't, because the code uses === true, and 1 doesn't pass the test.

In fact, setting debug to 1, 2, or 3, doesn't show PHP fatal syntax errors. 

A snippet with this code:

```
  $x = 1  /* no semicolon */
```

produces a blank page.

Setting the debug setting to E_ALL also produces a blank page, even though intval(E_ALL)  generates the correct value, because the setting is a textfield and we're getting intval('E_ALL'), which is 0.

Setting the debug level to -1 _does_ appear to give you E_ALL, so the description should make that clear.

The debug setting only works with integer values, but this creates another problem.  According to the PHP docs,  E_ALL is 32767 in PHP 5.4.x, 30719 in PHP 5.3.x, 6143 in PHP 5.2.x, and 2047 previously.

There really should be a way to set a custom error_reporting level with the debug System Setting without calculating the integer for your personal version of PHP, imo. Not all users have the option of doing it in php.ini.

I'm not sure what the best solution would be. Predefined integers (0,1,2,3,4,5) for common error_reporting configurations, explained in the setting description, with the constructor converting them to the PHP constant equivalents is one way to go.

krismas:
+1

Another "problem", is that (IMO) MODX->setDebug() and XPDO->setDebug() are not on the same "level", XPDO "debug" is incredibly verbose.

BobRay:
I think the problem is worse than I thought. I said that setting debug to -1 is the equivalent to E_ALL, but that's wrong. I've been unable to get E_NOTICE and E_DEPRECATED errors to show up either on the screen or in the error log regardless of the debug setting (although the MODX error handler has code that's clearly supposed to write them to the error log.

Here is some test code to put in a snippet:

    $x = array();
    $y = $x['missing'];
    ereg_replace('a','b','c');

    /* Show the enabled error_reporting options */
    $errLvl = error_reporting();
    for ($i = 0; $i < 15;  $i++ ) {
        print FriendlyErrorType($errLvl & pow(2, $i)) . "<br>\\n";
    }

    function FriendlyErrorType($type) {
        switch($type)  {
            case E_ERROR: // 1 //
                return 'E_ERROR';
            case E_WARNING: // 2 //
                return 'E_WARNING';
            case E_PARSE: // 4 //
[...]

PR #12585, opened 2015-08-24
Add support for CONSTANT format for debug settings (#12579)

krismas:
Constant E_ALL, E_ERROR… are now supported. At this point, combine form like « E_ALL & ~E_NOTICE » are not supported.  Also replace XPDO setDebug() by a setLogLevel().


krismas:
Irrelevant solution, because MODX is to verbose in DEBUG mode !

PR #12590, opened 2015-08-24
A suggestion to add support for CONSTANT format for "debug" settings (#12579)

krismas:
### What does it do ?

Add supoort for PHP constants name `E_ALL`, `E_ERROR`..., some extra checks on values and replace XPDO `setDebug()` by a `setLogLevel()` in these cases.

At this point, combine form like `E_ALL & ~E_NOTICE` are not supported.  
### Why is it needed ?

Selecting PHP log level is more simple and the `error.log` really smallest
### Related issue(s)/PR(s)
#12579 : Problems with debug System Setting


Mark-H:
Sorry it took so long for this to get handled, @krismas. I've merged it for inclusion in 2.7.

PR #12593, opened 2015-08-25
Modify delete and undelete processors to respect syncsite setting.

ghost:
## What does It do?

Makes resource delete and undelete processors respect the syncsite_default setting.
## Why is it needed?

The create and update processors were recently updated to respect a new syncsite_default setting. That did not take care of clearing the cache on delete and undelete.
## Related issues/pull requests

https://github.com/modxcms/revolution/issues/11980
https://github.com/modxcms/revolution/pull/12218

### DIFF
--- a/core/model/modx/processors/resource/delete.class.php
+++ b/core/model/modx/processors/resource/delete.class.php
@@ -224,12 +224,16 @@ class modResourceDeleteProcessor extends modProcessor {
      * @return void
      */
     public function clearCache() {
-        $this->modx->cacheManager->refresh(array(
-            'db' => array(),
-            'auto_publish' => array('contexts' => array($this->resource->get('context_key'))),
-            'context_settings' => array('contexts' => array($this->resource->get('context_key'))),
-            'resource' => array('contexts' => array($this->resource->get('context_key'))),
-        ));
+        $clear = $this->getProperty('syncsite',false) || $this->getProperty('clearCache',false);
+        if ($clear) {
+            $this->modx->cacheManager->refresh(array(
+                'db' => array(),
+                'auto_publish' => array('contexts' => array($this->workingContext->get('key'))),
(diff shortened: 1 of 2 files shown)


Mark-H:
@sottwell Unfortunately it doesn't look like this is working as expected when I try it. The delete and undelete option never pass along a syncsite or clearCache property as far as I can tell, meaning the cache is never cleared. In your description you mention the syncsite_default setting but those are not being used in the updated processors. Is this PR missing a final tweak perhaps?

theboxer:
Closing due to long inactivity. Feel free to reopen with more details or more commits.

intersel:
Too bad this PR is closed... it would give the ability to not always clear the cache when using the resource/delete and resource/undelete processors when using them as modx api. And so, would have the same ability than the resource/create and resource/update processors...

About syncsite_default setting, I have to admit that I could not find anywhere in the code where it is used... Perhaps, it may be used to init the default behaviour in these processors?... 
Is that the kind of tweak @Mark-H you had in mind?
```
public function clearCache() {
 $syncSite = $this->getProperty('syncsite',$this->modx->getOption('syncsite_default '));
....
```
Is it possible to reopen this PR?

Mark-H:
The PR can't be reopened anymore as the original branch no longer exists. Feel free to create a new one instead, though.

Without testing or diving into it, yeah making it fall back to the setting value (which presumably is enabled by default?) should do the trick. Altho perhaps the idea was to add the syncsite attribute somewhere in the UI? 

I don't know, it's been 6 years and as you can see there was not much of a discussion here. ;)

intersel:
Ok! I understand! actually, I was ready to prepare a PR when I saw this one from Sottwell, with exactly the same code modification I was to propose...
I'll prepare a new one... I have to admit that it's kind of obscure deep code to discuss !

PR #12598, opened 2015-08-26
Updating Font-Awesome to 4.4

jpdevries:
66 New Icons in 4.4
http://fontawesome.io/icons/

we don’t need a bower version and it is ok for the package to be private

Targets 2.4.x


OptimusCrime:
What will happen to all the `icon-*` classes? If I am not mistaken Font Awesome uses `fa-*` now.

pixelchutes:
The CSS prefix [is customizable](https://github.com/modxcms/revolution/blob/2.x/_build/templates/default/sass/index.scss#L4-L5), so we stuck with `icon` for BC when we first moved to `4.x`

theboxer:
Merged to 2.4.x

Issue #12291, opened 2015-01-22
Language strings missing warning


Jako:
```
(DEBUG @ /index.php) Language string not found: "basePath"
(DEBUG @ /index.php) Language string not found: "basePathRelative"
(DEBUG @ /index.php) Language string not found: "baseUrl"
(DEBUG @ /index.php) Language string not found: "baseUrlRelative"
(DEBUG @ /index.php) Language string not found: "allowedFileTypes"
(DEBUG @ /index.php) Language string not found: "imageExtensions"
(DEBUG @ /index.php) Language string not found: "thumbnailType"
(DEBUG @ /index.php) Language string not found: "PNG"
(DEBUG @ /index.php) Language string not found: "JPG"
(DEBUG @ /index.php) Language string not found: "GIF"
(DEBUG @ /index.php) Language string not found: "thumbnailQuality"
(DEBUG @ /index.php) Language string not found: "skipFiles"
```

Should be caused by core/model/modx/sources/modfilemediasource.class.php
modFileMediaSource::getDefaultProperties

There is also another warning:
```
(DEBUG @ /connectors/index.php) An error occurred while trying to cache lexicon/de/core/sources (lexicon/language/namespace/topic)
```

but I did not get located this one.

OptimusCrime:
Might be relevant for #12466.

This is either by design and won't be fixed, or this pull request will remove this issue.

OptimusCrime:
Pull request merged, not sure why this was not closed.

@modxbot close

PR #12602, opened 2015-08-26
Language string not found

Jako:
### What does it do ?

The patches add some language keys and change the topic names of two processors.
### Why is it needed ?

Removing several debug messages in a default installation:
- Language string not found
- An error occurred while trying to cache lexicon/de/core/sources
### Related issue(s)/PR(s)

See #12291, #12466


theboxer:
Merged into 2.4.x

PR #12604, opened 2015-08-26
Fix CLI installation notices due to undefined index type

krismas:
### What does it do ?

The index type for `modTemplateVarResourceGroup / tmplvar_template` is undefined since 5599467

Update MODX xPDO model to add index type definition (primary and unique). 

Also update `getIndexDef()` to replace these warnings by a log entry.
### Why is it needed ?

Fix CLI installation notices (#12564) due to undefined index type
### Related issue(s)/PR(s)
#12564 : CLI installation notices

### DIFF
--- a/core/model/modx/mysql/modtemplatevarresourcegroup.map.inc.php
+++ b/core/model/modx/mysql/modtemplatevarresourcegroup.map.inc.php
@@ -37,6 +37,8 @@ $xpdo_meta_map['modTemplateVarResourceGroup']= array (
     'tmplvar_template' => 
     array (
       'alias' => 'tmplvar_template',
+      'primary' => false,
+      'unique' => false,
       'type' => 'BTREE',
       'columns' => 
       array (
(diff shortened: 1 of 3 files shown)


pixelchutes:
@krismas, would you be willing to re-submit this PR targeting `2.4.x` branch for your patch referenced above? It can easily be cherry-picked if not, however I'm not sure if it should include the `getIndexDef()` changes here directly, as they may need to be patched upstream in [modxcms/xpdo](https://github.com/modxcms/xpdo)

Is that correct, @opengeek?

opengeek:
That's correct @pixelchutes — changes to modxcms/xpdo should be submitted to that project separately and would need to be adopted before this change could be accepted.

krismas:
I will try to do this ASAP...

Mark-H:
Hi @krismas what's the status on this? I suppose we should try to merge the schema changes in this repo, but I don't immediately see an xPDO pull request yet for fixing the issue there. Thoughts?

Mark-H:
This was fixed in xPDO via https://github.com/modxcms/xpdo/pull/76, which looks like it was tagged and included in 2.5.0.

PR #12605, opened 2015-08-26
Really clear the log...

krismas:
### What does it do ?

Remove the extra space when clearing  `errors.log` file via `writeFile()`
### Why is it needed ?

1 extra bit saved :smile: Seriously, a potential use case, may be a cron monitoring `errors.log` size.
### Related issue(s)/PR(s)
#12582 : Clear error log... not really


Omeryl:
Huh, it does seem like a large non-issue to me, but after looking in the git blame logs I see no particular reason why it was done this way. I'm curious if some systems didn't work right with a blank string, but it'd have to be a bunch of testing.

rtripault:
:+1: non issue and wondering why it's be done that way at first. But all for it (if there is no particular reason for having that white space), even just for the sake of the "cosmetic" when reading the log file.

theboxer:
Merged to 2.4.x

PR #12612, opened 2015-08-28
Update - Japanese lexicons

yama:
Thanks!


rtripault:
Thanks to you!

Unfortunately we no more manage translation here, but rather on https://crowdin.com/project/modx-revolution/ja (see https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md#translations).

Would you mind contributing there ?
Thanks

Issue #12601, opened 2015-08-26
[Request] Updatable Packages first in Package Manager


Jako:
Could the updatable packages be sorted first by a button or by a system setting in the Package Manager? Maybe with an ignore that update button for each package with two states: Ignore all new versions or ignore this version.

inreti:
+1

zaigham:
:+1:

beau-gosse:
+1

mishantrop:
+1

Jako:
As an addition to reopen #12613: The update status of all packages should be checked before the workspace grid is loaded with a 'checking for updates' modal box. Maybe with a console inside what is checked at the moment. The results should be cached in a similar way they are cached now.

Issue #12518, opened 2015-07-17
Default Per Page number of packages listed not related to the Default Per Page System Setting


boostventilator:
The System Settings has a preference called Default Per Page (default_per_page), described as “The default number of results to show in grids throughout the manager”, but this number is not being used as the default number of packages listed per page in the manager.

**Extras > Installer > Package Management > Packages > Per Page = 10**

Instead it is just set as 10 in here:

**Path:** manager/assets/modext/workspace/package.grid.js
**Line 86:** `,pageSize: 10`

netProphET:
Confirmed behavior. Going to try to fix this #modxbughunt

netProphET:
This was likely overridden for performance reasons. On discussion with opengeek and rtripault in #modxbughunt, I'm going to propose a fix that uses the `default_per_page` setting but limits it from being too high. Will hard-code that limit to 25 for now, but perhaps making that upper limit a system setting could be discussed as a separate feature request.

netProphET:
Pull request created #modxbughunt 
https://github.com/modxcms/revolution/pull/13348

Issue #12299, opened 2015-01-27
2.3.2-pl, Manager: In some grids the columns are not sortable


ottogal:
The sorting options in the Extras>Installer grid are deactivated (grayed-out); same with the Messages grid.

pepebe:
Unsortable columns can be found in many places. Before issues moved over to github, I wrote a bug report about this and I think Jason Coward wrote back that this behavior was not intended but somehow happened along the way. 

Should be easy to fix, but unfortunately this problem is not among the current top 100 list of issues.

It would be great to have a complete list of all grids that come along with unsortable columns, so somebody could do some research.

whitebyte:
Well, i partially managed to fix it. It's pretty easy to enable sorting in grid by modifying lines 32-36 in manager/assets/modext/workspace/package.grid.js (see gist - https://gist.github.com/whitebyte/58025c5d5cbf28e9fb35). You can use it as quick-fix, i'll make a proper pull request later.

pepebe:
@whitebyte Thanks for your fix.

I think its worth to have a look at all files listed by this search "sortable: false":

https://github.com/modxcms/revolution/search?utf8=%E2%9C%93&q=%22sortable%3A+false%22&type=Code

I'm sure there will be more columns suffering from this problem.

pepebe:
Related to this: https://github.com/modxcms/revolution/issues/12319

alroniks:
@Ruslan-Aleev Seems you can have a look at this.

Issue #8906, opened 2012-10-06
Set amount of list grid-items in system settings


DonQuicky:
donquicky created Redmine issue ID 8906

The default amount of packages shown in package management [grid] is set to 10. It would be nice to set the amount as an option in the system settings. Like any amount or all. For example: to examine  updates of packages are availble you have to go through many pages if you have many packages installed.

netProphET:
Superceded by #12518. This has finally been looked at - thanks for reporting! #modxbughunt

PR #12613, opened 2015-08-28
Add "Updatable" Column to packages grid

krismas:
### What does it do ?

This PR add an "Updatable" column to packages grid. This new column and the others are now sortable. Last the grid use the settings `default_per_page` in place of the previous fixed value (10).

packages-grid
### Why is it needed ?

This PR, help user with lot of packages to identify which ones are updatable.
### Related issue(s)/PR(s)
#12601 : Updatable Packages first in Package Manager
#12299 : 2.3.2-pl, Manager: In some grids the columns are not sortable
#12518 : Default Per Page number of packages listed not related to the Default Per Page System Setting
#8906 : Set amount of list grid-items in system settings

### DIFF
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -30,10 +30,11 @@ MODx.grid.Package = function(config) {
 
     var cols = [];
     cols.push(this.exp);
-    cols.push({ header: _('name') ,dataIndex: 'name', id:'main',renderer: { fn: this.mainColumnRenderer, scope: this } });
-    cols.push({ header: _('version') ,dataIndex: 'version', id: 'meta-col', fixed:true, width:120, renderer: function (v, md, record) { return v + '-' + record.data.release;} });
-    cols.push({ header: _('installed') ,dataIndex: 'installed', id: 'info-col', fixed:true, width: 160 ,renderer: this.dateColumnRenderer });
-    cols.push({ header: _('provider') ,dataIndex: 'provider_name', id: 'text-col', fixed:true, width:120 });
+    cols.push({ header: _('name')     ,dataIndex: 'name'         , id:'main'     , sortable: true,                        renderer: { fn: this.mainColumnRenderer, scope: this } });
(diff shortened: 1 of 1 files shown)


zaigham:
Wow, 4 in one shot! :+1:

manager/assets/modext/workspace/package.grid.js
```diff
@@ -83,7 +84,7 @@ MODx.grid.Package = function(config) {
                  ,'provider','provider_name','disabled','source','attributes','readme','menu'
                  ,'install','textaction','iconaction','updateable']
         ,plugins: [this.exp]
-        ,pageSize: 10
+        ,pageSize: MODx.config.default_per_page > 30 ? MODx.config.default_per_page : 30
```
Alroniks: Why so many by default? I think 20 would be enough.

Alroniks:
:+1:

sottwell:
:+1:

Mark-H:
Doesn't MODX check for update status when it retrieves the package for display in the grid, i.e. the updatable column would only apply to the packages currently shown in the grid, but not for packages on page 2?

inreti:
+1
For me even all packages can be shown at once without pagination if the updateable ones _and_ the not installed ones are listed first. All other packages in there are not interesting. Or what do you regularly do with them???

theboxer:
Having there one column that's sorting only current page will cause inconsistent behaviour and might provide misleading information. When you would have a package that needs updating on page 2 and you sort by `Update` column, this package won't appear on top of the first page -> you'll get wrong information about what needs to be updated. Actually you will think that nothing needs updating.

I'm for not including this change, unless there will be a way how to do it properly.

Jako:
In my opinion the update status of all packages should be checked before the grid is loaded with a 'checking for updates' modal box. Maybe with a console inside what is checked at the moment. The results should be cached in a similar way they are cached now.

Jako:
This PR should be opened again. It is not that bad and only has to be enhanced.

Jako:
This PR needs one addition before it could be merged: The update status of all packages should be checked before the workspace grid is loaded with a 'checking for updates' modal box. Maybe with a console inside what is checked at the moment. The results should be cached in a similar way they are cached now.

Mark-H:
@krismas Any thoughts on the above discussion on enhancing this pull request further?

PR #12616, opened 2015-08-28
Add anonymous_sessions to allow session-less access for anonymous users

opengeek:
### What does it do ?

Adds an `anonymous_sessions` system setting, enabled by default, which, when disabled, prevents sessions from being started until a user logs into a context.
### Why is it needed ?

High-traffic sites with a lot of anonymous visitors create a large number of sessions that by default are not garbage collected for at least 7 days. By disabling `anonymous_sessions`, only users who explicitly login to a context will create a session, saving overhead for the anonymous visitors and reducing load and storage requirements on the site overall.
### Related issue(s)/PR(s)

none that I could find

### DIFF
--- a/core/model/modx/moduser.class.php
+++ b/core/model/modx/moduser.class.php
@@ -341,6 +341,10 @@ class modUser extends modPrincipal {
      * @param string $context The context to add to the user session.
      */
     public function addSessionContext($context) {
+        if (!$this->xpdo->startSession()) {
+            $this->xpdo->log(xPDO::LOG_LEVEL_ERROR, "Attempt to start a session failed", '', __METHOD__, __FILE__, __LINE__);
+            return;
+        }
         if (!empty($context)) {
             $this->getSessionContexts();
             session_regenerate_id(true);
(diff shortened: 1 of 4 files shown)


core/model/modx/moduser.class.php
```diff
@@ -341,6 +341,9 @@ public function getSessionContexts() {
      * @param string $context The context to add to the user session.
      */
     public function addSessionContext($context) {
+        if (session_status() === PHP_SESSION_NONE) {
```
pixelchutes: RE: PHP_SESSION_NONE
> Since PHP 5.4.0.

pixelchutes:
This looks awesome, btw. Would officially bump minimum PHP version to 5.4.0.

core/model/modx/moduser.class.php
opengeek: Oops, I can probably do it in a 5.3 compatible way, but certainly up for debate...

opengeek:
The works without the PHP 5.4 dependencies now. I also added `$modx->startSession()` method that can be called multiple times and will simply return true or false if `$_SESSION` is available, regardless of how it was started. I needed this to properly mark the protected `$modx->_sessionStatus` when manually starting a session, or attempting to do so anywhere other than from the modX instance, but it will also come in handy for users who might want to start a session in their component explicitly. This will allow anonymous users to gain access to sessions on specific Resources even when `anonymous_sessions` is disabled.

rtripault:
This looks really cool!
Anything particular to test/check ?

opengeek:
> Anything particular to test/check ?

@rtripault I have been testing to make sure there is no $_SESSION unless I login to front-end contexts with this setting disabled. I'm also making sure you can manually start one in a snippet call for an anonymous user, e.g. using `$modx->startSession();`

theboxer:
Merged to 2.x

exside:
This is great, was looking for that feature for a long time, had the impression that all the cluttering up in the sessions table bc of the anonymous sessions sometimes makes modx slow...!

PR #12619, opened 2015-08-30
Add support for MODX Tags in autotag parent(s) parameter

krismas:
### What does it do ?

With this PR the `parent` parameter can contain MODX tags (TV, Snippet), parsed before the tags list is created.

Parent: `1,[[*parent*]],10`
### Why is it needed ?

We often use autotag in Collections, but we don't want to share tags list between our Collections nor create as many TVs as Collections.
### Related issue(s)/PR(s)

n/a

### DIFF
--- a/core/model/modx/processors/element/tv/renders/mgr/input/autotag.class.php
+++ b/core/model/modx/processors/element/tv/renders/mgr/input/autotag.class.php
@@ -24,7 +24,12 @@ class modTemplateVarInputRenderAutoTag extends modTemplateVarInputRender {
         ));
         if (!empty($params['parent_resources'])) {
             $ids = array();
-            $parents = explode(',',$params['parent_resources']);
+
+            /** check if we have some TVs in place of id */
+            $chunk = $this->modx->newObject('modChunk', array('name' => uniqid()));
+            $chunk->setCacheable(false);
+            $parent_resources = $chunk->process(null, $params['parent_resources']);
+            $parents = explode(',', $parent_resources);
 
             $currCtx = 'web';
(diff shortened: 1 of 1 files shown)


sdrenth:
@krismas I've tested this PR, and noticed that using tv names in the autotag parent resources the output is an empty string instead of the template variable value of the current resource. When the TV has a default input value, then the default input value is returned. If you could fix that, then this PR is ready to be merged.

I can confirm that using resource fields (For example: [[*parent]]) or a snippet is working.

OptimusCrime:
I dont understand how this works. Is this for dynamically setting input values for a TV?

sdrenth:
@OptimusCrime With this, the tags from the parent resource are automatically added to the tags of the child resource by setting the Parent Resources setting to [[*parent]] in the input options for the TV.

Jako:
That autotag TV type is quite ineffective in most cases (specially when resource IDs are referenced). It is working well in one direction (the resource is tagged with the following tags or tagging resources) but not in the opposite direction (which resources are tagged with the following tags or tagging resources). Tagger or a similar solution should be used for this.

The PR is ok, but it points somehow in the wrong direction. The autotag tv should better be deprecated.

OptimusCrime:
Oh, this is only for the autotag TV. I have never used it.

Mark-H:
@krismas Any thoughts on @sdrenth's comments? 

I'm kinda with Jako in that the autotag type is better replaced by Tagger these days. A lot more powerful and performant.

Mark-H:
Closing as there hasn't been a response to the points brought forward by Sander and Jako

PR #12622, opened 2015-09-01
Create table for modAccessNamespace

theboxer:
### What does it do ?

Creates table for modAccessNamespace
### Why is it needed ?

Currently the table is created only when updating for older versions, not during new installs.


hansek:
:+1: :)

PR #12627, opened 2015-09-02
$itemValue is empty even when value for tag isset

degoya:
Tag input type wasn't working with text==value, 
Value field returned empty value


OptimusCrime:
Care to elaborate? What do you mean? I have had no trouble getting this to work.

degoya:
in modx 2.5.6 on PHP-FPM 5.3.2 when i make a tag with this Values i.E. 
title1==class1||title2==class2||title3==class3

i can select the values for the TVs and everything looks right, after saving and reloading the resource the value list inputfield was empty. When i only used class1||class2||class3 everything worked as expected. after i was diggin to the code i found out the the value didn't got set and thats was the reason it was empty, after changing the line in the commit everything worked like it should.

Mark-H:
Thanks @degoya - merged to be included in 2.5.0-rc1. Please mind the contributing docs (https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md) next time, and I'll try to merge faster :D

Issue #5691, opened 2011-09-08
RTE toggle on introtext / description


goldsky:
goldsky created Redmine issue ID 5691

Please enable RTE toggle on introtext/description, so we can use that with the right HTML tags.

whitebyte:
Related forum topic: http://forums.modx.com/thread/?thread=30943&page=1
Seems that someone managed to enable TinyMCE for introtext with simple plugin.

Mark-H:
Redactor 2 has this built in through the redactor.loadIntrotext system setting, and after #12632 that will also work on non-document resources. 

I'd like to propose closing this issue as it's something extras can easily provide (either built in like Redactor 2, or via a plugin like this https://gist.github.com/christianseel/780b16f202c873b3b278).

OptimusCrime:
I really, really like this. Nice work @Mark-H

OptimusCrime:
Closing due to possible solution and no feedback in over a year.

@modxbot close

PR #12632, opened 2015-09-04
Fix loading rich text editors on non-document resource types

Mark-H:
### What does it do ?

Rich text editors are loaded by firing the OnRichTextEditorInit plugin events, which then allows third party plugins to handle their instantiation and registering required assets.

Some of those editors, including modmore's Redactor, use the `$modx->event->output($html)` approach to feeding back the resulting markup that needs to be injected into the page. This markup is then successfully set into a placeholder that is loaded in the create.tpl or update.tpl template for the various resource types.

The problem is that prior to this patch, the placeholder value is only set into the page when it is a standard modDocument resource, and not when it is a weblink, symlink or static resource.
### Why is it needed ?

Redactor 2 introduces a new option to load a rich text editor for the introtext field, which people have also been doing using a plugin, for example this one: https://gist.github.com/christianseel/780b16f202c873b3b278

Without this patch, those features only work on modDocuments' as the markup that defines the MODx.loadRTE method is not inserted into the page. 
### Related issue(s)/PR(s)
#5691 is somewhat related, though not necessarily fixed by this PR.

### DIFF
--- a/manager/templates/default/resource/create.tpl
+++ b/manager/templates/default/resource/create.tpl
@@ -7,5 +7,5 @@
 
 {$onDocFormPrerender}
 {if $resource->richtext AND $_config.use_editor}
-{$onRichTextEditorInit}
+    {$onRichTextEditorInit}
 {/if}
--- a/manager/templates/default/resource/staticresource/create.tpl
+++ b/manager/templates/default/resource/staticresource/create.tpl
@@ -2,3 +2,6 @@
 <div id="modx-resource-tvs-div" class="modx-resource-tab x-form-label-left x-panel">{$tvOutput}</div>
 
 {$onDocFormPrerender}
+{if $resource->richtext AND $_config.use_editor}
+    {$onRichTextEditorInit}
+{/if}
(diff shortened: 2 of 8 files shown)


rtripault:
:+1: for idea.

Does that mean RTEs would be loaded (if enabled) on any resource type, as default ?

Mark-H:
They are already being loaded per another pull a while back. I don't immediately have the reference for that available but can look it up if you'd like.

This just completes the integration for RTEs that require the output of the event being added to the page, like Redactor :)

jpdevries:
:+1:

pixelchutes:
@Mark-H is #12272 the one?

Mark-H:
Yeah, that's the one Mike :)

Mark-H:
@theboxer Would you mind elaborating a bit on why this is tagged for 2.5.0 instead of 2.4.x? #12272 is the new feature, this pull request only makes sure it actually works.

theboxer:
Well you submitted it to 2.x branch and it's not a real bugfix, it's adding functionality for loading RTE on other resource types

Mark-H:
Oh, I must've not been paying attention in choosing the right branch, sorry.

This does _not_ add functionality for loading RTE on other resource types. That was already added with #12272 a while back. This PR fixes that implementation to work consistently.

theboxer:
Oh, sorry, I didn't read your first comment properly. Let's include it with next patch release than.

theboxer:
Merged to 2.4.x

PR #12633, opened 2015-09-04
Restore backwards compatibility for addons interacting with modTransportProvider

Mark-H:
### What does it do?

The getClient method was previously public (see https://github.com/modxcms/revolution/blob/v2.3.6-pl/core/model/modx/transport/modtransportprovider.class.php#L73) and used by Gitify to provide a way of installing packages via the command line. When modTransportProvider was refactored, this method was marked protected, causing Gitify to break with the following error:

Fatal error: Call to protected method modTransportProvider::getClient() from context 'modmore\Gitify\Command\InstallPackageCommand' in Gitify/src/Command/InstallPackageCommand.php on line 168

Setting getClient back to public resolves that error. 
### Why is it needed?

2.4 was intended to be backwards compatible, however this change was overlooked. Restoring the public method will make it backwards compatible again in 2.4.1. 
### Related issue(s)/PR(s)

https://github.com/modmore/Gitify/issues/125

### DIFF
--- a/core/model/modx/transport/modtransportprovider.class.php
+++ b/core/model/modx/transport/modtransportprovider.class.php
@@ -229,7 +229,7 @@ class modTransportProvider extends xPDOSimpleObject {
 
             $package->parseSignature();
             $package->setPackageVersionData();
-            
+
             $locationArgs = (isset($metadata['file'])) ? array_merge($metadata['file'], $args) : $args;
             $url = $this->downloadUrl($signature, $this->arg('location', $locationArgs), $args);
             if (!empty($url)) {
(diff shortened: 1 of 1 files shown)


hansek:
:+1:

rtripault:
The diff is somehow messy... did you change the encoding or something ?

pixelchutes:
@rtripault here's the [diff minus whitespace](https://github.com/modxcms/revolution/pull/12633/files?w=1)

I think the line-endings changed from Windows style [dos] to Unix (preferred)

Mark-H:
Ah it happened again.. Phpstorm is making changes based on the .editorconfig file to make sure it follows the project standards, so like Mike said these source files must have the wrong line endings currently.

rtripault:
Yes, you are right! I was wrongly assuming the line ending would have been correct, but it's not!
So that's rather a good thing!

Let's get this merged ? ;)

OptimusCrime:
+1!

PR #12635, opened 2015-09-07
Not rendered output properties of custom TVs

Jako:
### What does it do ?

Fixing not rendered output properties of custom TVs.
### Why is it needed ?

Not rendered output properties of custom TVs.
### Related issue(s)/PR(s)

See: https://github.com/Jako/ImagePlus/issues/13


Mark-H:
Confirmed that this also fixes the SimpleCart Options TV.

PR #12611, opened 2015-08-27
Grunt away compress_js and make Manager not broken

jpdevries:
### What does it do ?

Moves the compress_css and compress_js to a front-end Grunt workflow and deletes the `manager/min` directory.
### Why is it needed ?

Performance and stability improvements.

In MODX 2.4- a server-side minifier is used to concatenate and minify CSS and Javascript files that are added with the MODX API. This PR begs we abandon that methodology completely so that we can ship a more performant, more optimized, and more stable product.

Targets MODX 2.5. ~~I can flatten the commits if you'd like~~ I tried and ran away.

I'm not sure why it can't automatically be merged. Probably because I started the PR a while ago. Let me know if you'd like me to rebase or reassemble it.

You won't find it in the code but this PR also asks that we have a conversation about how we look at HTTP Requests in regards to performance. They aren't all bad and in use cases like a dynamic Manager whose assets can change from page to page or as plugins are updated, activated, or deactivated to combine everything together is drastically worse for performance when measured in any practical test.

_I’ve gotten a lot of resistance trying to remove these settings over the past three years however they were never tested or really architected to consider how browsers cache files so it has honestly been frustrating to be expected to prove why they need to go away considering they were never proven they should be there in the first place._

Please keep in mind that browsers cache files based off the files URL. Anyone is obviously welcome to pull from this branch and post performance and optimization findings.
### What it does
- Adds a Grunt workflow that creates an unminified and a minified version of the global core Javascript assets for the MODX Manager with _nothing else added_
  - `<%= dirs.manager %>assets/modext/modx.jsgrps.js`
  - `<%= dirs.manager %>assets/modext/modx.jsgrps-min.js`
- adjusts the core backend to ignore `compress_` settings. This means extra assets will not be grouped together. Nor should they be.
- ~~removes the manager/min directory :tada:~~ 
- fixes all the sites that suffer from the below related issues
- potentially increases the number of HTTP requests a manager page may have but in doing so utilizes an architecture designed to load assets sensibly and more efficiently. https://github.com/modxcms/revolution/pull/12538#issuecomment-134946207
- takes us one step closer to a CDN hosted workflow that would mean logging into one MODX dashboard caches the assets for the 19 MODX sites you haven't logged into yet today. This would make MODX feel _much snappier_ across the web but as I believe @adamwintle brought up there must be proper fallbacks for not only offline use but in countries like China that block CDNs. Anyways I think we should shoot for that, and maybe likewise support in the API, for MODX 3 personally.

[...]

PR #12643, opened 2015-09-09
Performance improvements and corrected etag for lang.js connector #modxbughunt

wuuti:
### What does it do ?

This PR attempts to improve the performance and caching logic for the lang.js files created by connector. It also fixes an issue with etags causing never a 304 modified being sent. 

It is quite unlogical to regenerate the dictionary (=lexicon entries collection) for every manager page request according to _some_ topics maybe needed. As a result, nearly every page generates its own dictionary. Why don't we send the _whole_ dictionary for a language and the system at once to the browser? It is send only once, after that the browser does not need to reload it regardless on which page you are. 
In this PR the dictionary is generated by getting all namespaces and all topics inside, then fetching the whole lexicon for all topics. The result is cached (currently in cache partition 'lexicon-js'), so the etag computation can be done efficiently. 
### Why is it needed ?

Speeds up manager experience and fixes bug with etag-header. The requests to connectors/lang.js.php get a lot faster and re-requests get the correct 304 not modified.
### Open issues/enhancements
- purge cache partition on cache refresh and lexicon changes for that language. 
- (update) the cache partition also needs to be refreshed when an extra is installed/updated


inreti:
Did some updates to the PR. When cache refresh runs, the cache for the dictionary is refreshed too.
Note: this does not necessarily mean that on the next load a 200 is sent - if the newly generated dictionary hash matches again the browsers etag, a 304 is sent anyway, because the dictionary contents have not changed. So cache refresh only triggers a new dictionary load by the browser if something has really changed in the lexicon topics.

Mark-H:
This is interesting. I'm not 100% sold yet on the idea of including all namespaces and all topics - that could be a shit ton (technical term) of data that never gets used. Even if its cached, is it beneficial on say a mobile device (2.5 is more mobile friendly after all), or is this going to increase the download size on that first hit by so much that their out of their monthly quota? 

Any feedback from @modxcms/integration-managers-revolution?

Mark-H:
I've tried this patch earlier on the modmore site (which has, for reasons including legacy and a lot of custom stuff, _48_ namespaces) and the lang.js.php file went from 50kb (15kb gzipped) to 498kb (123kb gzipped) so it's approximately 10 times bigger in this particular example. 

In the bigger picture of the manager that's actually not too terrible though, and the caching is a nice bonus. Just wondering if there's any issues with loading all namespaces and topics..

inreti:
@Mark-H I am quite sure the data can be handled even by small mobile devices. If you dare to use the manager on those devices, these additional Bytes won't harm you compared to all the other data transferred. And most installations won't have that huge modmore-sized values, do they? 
And - if you use the manager without this PR for a while, the data will be transfered anyway over the time, only in smaller, partly redundant chunks - which are producing an overall larger amount of traffic (because of the redundancy). Monthly quotas will surely not be affected to much (and - the lexicon data can be compressed quite effectively on the wire...).

What issues do you expect with loading all namespaces and topics? Any idea? I couldn't imagine any. The only one might be that it blows the modx object a bit. Can you measure that portion in your huge example instance?

christianseel:
From my perspective a larger file with better caching is much better than a smaller file that is requested multiple files… And I'd still say if you're using the current MODX manager on a mobile device, you will need a huge mobile data quota anyway…

jpdevries:
> From my perspective a larger file with better caching is much better than a smaller file that is requested multiple files… And I'd still say if you're using the current MODX manager on a mobile device, you will need a huge mobile data quota anyway…

https://github.com/modxcms/revolution/pull/12611 proves that a larger file with shared cache becomes more performant each time the user visits a new screen while optimizing for a single page experience bloats the payload as caches are not shared. I feel the research done there carries over here.

> I've tried this patch earlier on the modmore site (which has, for reasons including legacy and a lot of custom stuff, 48 namespaces) and the lang.js.php file went from 50kb (15kb gzipped) to 498kb (123kb gzipped) so it's approximately 10 times bigger in this particular example.

Ya but measure your ~50kb weight multiplied by the number of pages they visit. It could easily be larger and continue to grow as they click around. See https://github.com/modxcms/revolution/pull/12611#issuecomment-135659086 for a comparable analogy. 

The MODX Manager is not a website that prioritizes snappy initial loads – it is a web app. Optimization wise the two are polar opposites. We should be focusing on things like advising GZIP is enabled rather than reducing HTTP requests and optimizing for single page, first impression, visits.

OptimusCrime:
I too agree with one big file is better than multiple small ones. In terms of speed improvements, is there anything you can tell us? Any tests? Numbers?

connectors/lang.js.php
```diff
@@ -5,68 +5,115 @@
...
 
+    $buf = 'MODx.lang = {';
+    $s = '';
+    while (list($k, $v) = each($entries)) {
+        $s .= "'$k': " . '"' . esc($v) . '",';
+    }
+    $s = trim($s, ',');
+    $buf .= $s;
+
+    // use compressed js here - saves some bytes :-)
```
Mark-H: Kinda ironic that this PR increases the size of the lang.js ten-fold, but we need to use a compressed js snippet to save bytes?

wuuti:
@Mark-H @rtripault Will this ever make it into the core? Thought it was quite useful and improves performance... I don't have any performance statistics, but it seems obvious that it is better in performance.
#modxbughunt

connectors/lang.js.php
rtripault: Agreed. I'd rather be up for code readability instead!

rtripault:
Not opposed to the idea here (but i'm not alone), according you "un-minify" the `_` function

wuuti:
Hm.... merging that old PR with 2.5.x seems not a good idea... any advice?

opengeek:
Do not ever merge old PR's. You need to rebase your work on top of the branch you want to target.

OptimusCrime:
Is it fixed now @opengeek ? The PR contains commits from two users, but it looks correct. Could need a squash though.

sdrenth:
Tested PR and it looks good to go!

Mark-H:
@wuuti Would you mind re-creating this PR targeting 2.x instead? I tried changing the base but because 2.5.x and 2.x have diverged, it's not quite as straightforward for this to get merged into 2.x.

As a reminder:
2.x => new features/major improvements
2.MINOR.x (e.g. 2.5.x) => bug fixes for the 2.MINOR (e.g. 2.5) series

wuuti:
@Mark-H Not sure where the problem with the 2.x base right now is? Or do you want me to create a complete new PR? Hopefully this won't get forgotten again for 2 years... :-(

Mark-H:
@wuuti I see you tried to merge 2.5.x into 2.x, but that's unfortunately not the solution here as those branches diverged.. if you could create a new PR with just your improvements, that would be amazing.

wuuti:
Done in #13623

Issue #12644, opened 2015-09-10
Different tree styles for unpublished + hidemenu


inreti:
### Summary
Not possible to differentiate hidemenu at the resource tree.

![screenshot from 2015-09-10 16 22 48](https://cloud.githubusercontent.com/assets/2852461/9790787/38dd0bb8-57d8-11e5-91e1-d94146e4f06e.png)

### Step to reproduce
Create a resource with unpublished+hidemenu and another one with unpublished+!hidemenu.
 
### Observed behavior
Both resources are displayed in the same way at the resource tree. Style italic, same color. 

### Expected behavior
Hidemenu: opaque color, unpublished: italic. The unpublished+!hidemenu style should be colored normal (resource 3 in the screenshot), and italic. 

### Environment
2.4.0-pl, 2.4.0-dev. FF and Chrome. 

Maybe this behaviour is intended, but imho the state styles _should_ be different. 

I would also like something better, though this is not a quick fix and I currently do not know exactly how to achieve. Either the unpublished or the hidemenu state should be visualized by anything else than the text style/color - perhaps an overlay icon for hidemenu?

rthrash:
I would think the Unpublished but shown in menu would match the color of the Published resource: color to indicate menu toggle, italics to indicate publish status.

inreti:
@rthrash yep, thats what I wanted to say with "colored normal". So to be exact, (3) has to be italic and with the color of (4).

rthrash:
I concur and that's what I was trying to say, too. :+1:

inreti:
Great :-)

PR #12646, opened 2015-09-11
Fix tree style bug #12644, enhance page title style

wuuti:
### What does it do ?

This PR fixes bug #12644, additionally the same bug for deleted resources and also applies the styles to the resource header in order to visualize the resources state better:

screenshot from 2015-09-11 13 42 42
### Why is it needed?

It was not possible to distinguish resources states at the resource tree because not all different combinations produced different styles. This is fixed now. 
When editing a resource, it was also not possible to see the resources' state at one glance (e.g. the tree is hidden, state is only visible over the checkboxes and the deleted checkbox is on another tab). Therefore it could happen that someone unintentionally edits a deleted resource. This is now fixed by applying the same styles to the header title as on the resource tree. 
### Related issue(s)/PR(s)
#12644

### DIFF
--- a/_build/templates/default/sass/_colors-and-vars.scss
+++ b/_build/templates/default/sass/_colors-and-vars.scss
@@ -117,11 +117,14 @@ $treePseudoRootOverBg: $lightBg;
 $treePseudoRootOverColor: #447996;
 
 $iconColor: $treeText;
-$unpublished: lighten($treeText, 10%) !important;
+$unpublished: $treeText;
 $disabled: $unpublished;
-$hidden: $unpublished;
+
+$hidden: lighten($treeText, 10%) !important;
 $unpubText: italic;
-$lockedText: italic;
+/* locked is already signaled with lock-icon, no need to use other semantic style here
+   as italic is used for "unpublished" */
+$lockedText: inherit; //italic;
 
 $delTextColor: fade-out(darken(lighten(desaturate($red,50%), 33%), 25%), 0.5) !important;
 $delTextStyle: normal;
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -226,7 +226,6 @@
 
   i.icon,
   i.icon-large {
-      color: $unpublished;
       font-style: normal;
   }
 }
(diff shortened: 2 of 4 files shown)


Jako:
I would prefer some icons before the title that show the state on a hover. The red/strikethrough/italic text is maybe fast seen, but which style means which state has to be remembered.

sottwell:
Even I have no problem remembering that. Flagged for deletion is the ultimate state of non-use, then unpublished, then hidden from menus. The existing system seems quite logical and straightforward to me.

pixelchutes:
:+1: Resource Tree node status updates

:-1: Resource Title styling.

This feels unnecessary, as we already have the checkboxes on the page to show the current state. What happens when you un-check the Published checkbox? Now the title title styling is out-of-sync, requiring more code to manage pre-submitted state changes, etc.

inreti:
@pixelchutes I do think that the resource title styling is necessary. The _deleted_-checkbox is on another tab (and cannot be simply moved by forms customization). All the checkboxes may be _not_ shown due to forms customization, nevertheless the corresponding information has to be visible to users. 
The title already has the live-review functionality, when you change the textual content the title is also replaced on the fly. It shouldn't be so complicated to implement live styling as well. But do we really need a live reflection? In fact the states do only apply after a save of the resource and that's when we need to change the title style. I am sure I can handle the two extra lines of code :-)

@jako I do also agree with you about the icons - although @sottwell mentioned that there are no problems to remember the different styles' meanings, I believe there are too many style combinations here. And from my point of view - I always have to think twice about the _italic_ and opaque semantics at the tree and can't remember quickly that opaque means not "unpublished" but "hidden from menu". Despites, "hidden from menu" seems a bit legacy at all...

It was not the best idea to combine the bug fix of the tree styling _and_ an enhancement suggestion in one PR. I will divide the PR into two of them and make a nicer suggestion for the title.

sottwell:
Actually, at first I missed the Evo way - published was green, I remember, and of course deleted was red/strikethrough, but I forget the rest of it.

Jako:
I would not like that red color and the text styles in the heading. That is my reason for suggesting icons there.

whitebyte:
This PR should be split into two.

theboxer:
@inreti Please split this into 2 PRs, I'd like to merge tree fixes, but the resource title need further discussion (I personally don't like it :))

inreti:
It split up now and the fix part is in #12699.

PR #12652, opened 2015-09-16
Update PHPMailer to v5.2.13

### LINKED ISSUE #12650: PHPMailer and PHP 5.6
in the version of phpmailer that is currently used - 5.2.8 the SSL is broken on PHP 5.6 (cannot connect to SMTP server). The latest version (5.2.12) of phpmailer is able to establish smtp ssl connections. this lib should be updated to fulle support php 5.6

pixelchutes:
### What does it do ?

Updates PHPMailer from v5.2.8 to v5.2.13
### Why is it needed ?

Intended to address SMTP server connectivity issues on PHP 5.6+ _(improving MODX `2.4.x` compatibility on PHP 5.6)_.

From [PHPMailer](https://github.com/PHPMailer/PHPMailer/blob/v5.2.13/class.phpmailer.php#L250-L256):

>    /**
>     \* Whether to enable TLS encryption automatically if a server supports it,
>     \* even if `SMTPSecure` is not set to 'tls'.
>     \* Be aware that in PHP >= 5.6 this requires that the server's [certificates are valid](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting#php-56-certificate-verification-failure).
>     \* @type boolean
>     */
### Related issue(s)/PR(s)

Should resolve #12650. Please confirm!

### DIFF
--- a/core/model/modx/mail/modphpmailer.class.php
+++ b/core/model/modx/mail/modphpmailer.class.php
@@ -69,7 +69,7 @@ class modPHPMailer extends modMail {
                 $this->mailer->Hostname= $this->attributes[$key];
                 break;
             case modMail::MAIL_LANGUAGE :
-                $this->mailer->SetLanguage($this->attributes[$key]);
+                $this->mailer->setLanguage($this->attributes[$key]);
                 break;
             case modMail::MAIL_PRIORITY :
                 $this->mailer->Priority= $this->attributes[$key];
@@ -135,16 +135,16 @@ class modPHPMailer extends modMail {
                 $type= strtolower($type);
                 switch ($type) {
                     case 'to' :
-                        $this->mailer->AddAddress($email, $name);
+                        $this->mailer->addAddress($email, $name);
                         break;
                     case 'cc' :
-                        $this->mailer->AddCC($email, $name);
+                        $this->mailer->addCC($email, $name);
                         break;
                     case 'bcc' :
-                        $this->mailer->AddBCC($email, $name);
+                        $this->mailer->addBCC($email, $name);
                         break;
                     case 'reply-to' :
-                        $this->mailer->AddReplyTo($email, $name);
+                        $this->mailer->addReplyTo($email, $name);
                         break;
                 }
             }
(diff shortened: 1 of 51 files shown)


pixelchutes:
PHPMailer underwent [reformatting to the PSR-2 coding standard](https://github.com/PHPMailer/PHPMailer/blob/v5.2.7/changelog.md) in v5.2.7. MODX v2.3.x *(and v2.4.0)* shipped with PHPMailer v5.2.8...

References [in modphpmailer.class.php](https://github.com/modxcms/revolution/blob/39ba702929acbad7c75a60321776977571975d2d/core/model/modx/mail/modphpmailer.class.php#L137-L147) failed to account for the [method](https://github.com/PHPMailer/PHPMailer/blob/v5.2.7/class.phpmailer.php#L701) [names](https://github.com/PHPMailer/PHPMailer/blob/v5.2.7/class.phpmailer.php#L2753) [changing](https://github.com/PHPMailer/PHPMailer/blob/v5.2.7/class.phpmailer.php#L644) to camelCase (inherent per [PSR-1](http://www.php-fig.org/psr/psr-1/)), though nothing broke since [PHP Functions are case-insensitive](http://fr.php.net/manual/en/functions.user-defined.php).

The 2nd commit in this PR is for consistency / good form.

pixelchutes:
**NOTE:** This could potentially be breaking for anyone using MODX distributed PHPMailer directly *(e.g. NOT using via modPHPMailer)*, since self-autoloading was [removed in v5.2.9](https://github.com/PHPMailer/PHPMailer/blob/v5.2.9/changelog.md)

opengeek:
Does 0f50b60 address the potential BC break due to the self-autoloading?

pixelchutes:
Yes, that was the intent. However, if someone was including PHPMailer class directly, they would need to manually invoke the autoloader, use modPHPMailer, or use PHPMailer via composer.

That said, as long as CA certs are configured correctly where PHP is looking, SMTP over TLS seems to work fine on PHP 5.6, even on v5.2.8.

PR #12657, opened 2015-09-17
xpdo::getOption performance improvement

ronaldex:
Only call more expensive function 'array_key_exists' when 'isset' fails, leads to ca 15-20% performance increases for server response time in our specific situation.


pixelchutes:
@ronaldex Would you mind submitting this PR upstream on the [xPDO project repository](https://github.com/modxcms/xpdo)?

PR #12662, opened 2015-09-22
Change user's grid in ACL view to be same as in update group.

theboxer:
### What does it do ?

Changes grid in ACL view to the same as in update user group.
### Why is it needed ?

Current grid is same as in manage users - you can't manipulate with users in user group. You can by mistake remove user from system instead of removing from user group.


sottwell:
I think that full-featured users grid you get when clicking on a group was to provide a way of filtering users by group for management. With the new filter added to the Manage -> Users menu item, it's no longer useful. So making it be the same feature-limited grid of users as when you update the group and open its Users tab makes sense.

sottwell:
Didn't that used to be a full list of all users so you could drag them into the group? Kind of like with Resource Groups? Which always seemed a little odd to me since you should just be able to drag directly from the Tree... but that's another matter.

theboxer:
Nope, it wasn't grid with all users :) It shows users based on what user group you click

sottwell:
I guess I had it confused with the Resource Groups then. Unless maybe in Evo...

PR #12667, opened 2015-09-23
Enhance file tree for large images and transparency

mindeffects:
### Problem

When hovering over the files tree, large images (incl. SVGs) leave their parent's container. Spotting transparency is visually poor.

image
### Solution
1. For `.x-tip img {}` the `max-width` must be `100%` to make the image fit into its parent-container. Otherwise large images will totally blow out of proportion and "leave" the container's borders. So the `66vw` had to be removed.
2. Adding a chessboard-pattern to `.x-tip img {}` – and lightening the background color of `.x-tip {}` – makes transparency more visible. It now looks like in Photoshop and other applications.

image


mindeffects:
Sorry, oeverlooked some problems with the background color...

PR #12670, opened 2015-09-24
Fix image width and add transparency pattern

Jako:
### Issue

When hovering over the files tree, large images (incl. SVGs) are displayed larger than the parent's container. Spotting transparency is visually poor.

before
)
### Solution
1. For `.x-tip img {}` the `max-width` must be `100%` to make the image fit into its parent-container. Otherwise large images will totally blow out of proportion and "leave" the container's borders. The `66vw` had to be removed, since the container width is calculated by ExtJS and the viewport width not.
2. Adding a chessboard-pattern to `.x-tip img {}` and changing the background color of `.x-tip img {}` makes transparency more visible. It now looks like in Photoshop and other applications.

after

Credits for this PR go to https://github.com/mindeffects.


mindeffects:
Thank you for helping me out of the/my GitHub dungeon!!! 
I promise to learn git(hub)! ;-)

theboxer:
Merged to 2.4.x branch

Issue #12656, opened 2015-09-17
Spaces in URI encoded as +


sottwell:
### Summary

Spaces in URI query strings are encoded as + in generated URIs.
### Step to reproduce

Using MODX link tag with spaces in its properties such as [[~42? &file=path/to/some file name.pdf]]
### Observed behavior

This will generate a URI with the spaces replaced with a + sign, which can break scripts consuming the generated URI.
### Expected behavior

The generated URI should have spaces replaced with the correct RFC 3986 compliant %20 encoding
### Environment

All versions, any environment.

The modLinkTag class, in core/model/modx/modparser.class.php, in its process() function, uses the PHP function urlencode() on line 1326 which does replace spaces with + signs. The PHP function rawurlencode uses RFT 3986 entities for replacement. So changing the use of urlencode() to rawurlencode() would solve this issue.

syts88:
I'm going to try and fix this today.

sottwell:
I'm pretty sure this was already fixed https://github.com/modxcms/revolution/pull/12675#issuecomment-216269895

sottwell:
core/model/modx/modparser.class.php on line 1338
```                        if ($qs= implode('&', $qs)) {
                            $qs= rawurlencode($qs);
                            $qs= str_replace(array('%26','%3D'),array('&amp;','='),$qs);
                        }
```

syts88:
You are right :)! So perhaps this issue should be closed official. We need to get the open issue counter down :) #modxbughunt

sonicpunk:
Awesome work and communication!!!!  Thanks!!!

PR #12675, opened 2015-09-25
use rawurlencode in modparser

ghost:
### What does it do ?

Replaced urlencode() with rawurlencode() in core/model/modx/modparser.class.php on line 1326. This causes spaces in generated URI query strings to be replaced with %20 instead of +.
### Why is it needed ?

When using properties in a MODX URL tag it will replace spaces in any resulting query string with + signs. These are only valid for form submission POST values. For URIs they should be replaced with RFC 3986 compliant encoding of %20. Scripts such as the PDF viewer will not work with + signs. PHP provides the urlencode() function to replace spaces with a + sign "for historic reasons". For correct RFC 3986 compliant encoding of %20 rawurlencode is used.
### Related issue(s)/PR(s)

Related issue: #12656
http://php.net/manual/en/function.urlencode.php

> It is encoded the same way that the posted data from a WWW form is encoded, that is the same way as in application/x-www-form-urlencoded media type. This differs from the » RFC 3986 encoding (see rawurlencode()) in that for historical reasons, spaces are encoded as plus (+) signs.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -1323,7 +1323,7 @@ class modLinkTag extends modTag {
                             $qs[]= "{$propertyKey}={$propertyValue}";
                         }
                         if ($qs= implode('&', $qs)) {
-                            $qs= urlencode($qs);
+                            $qs= rawurlencode($qs);
                             $qs= str_replace(array('%26','%3D'),array('&amp;','='),$qs);
                         }
                     }


sottwell:
Any chance for this to be merged? I'm getting tired of patching my installations whenever I want to use URL properties.

sottwell:
Apparently this is not really a bug and + signs should be accepted as replacement for spaces.

whitebyte:
@sottwell please reopen. I also think this is an issue and `rawulencode` should be preferred over `urlencode`.

sottwell:
Well, I'm certainly willing. I guess we'll just leave it up to "them".

OptimusCrime:
Any reason this was not added to 2.5?

Jako:
Merged to 2.5.x branch

sottwell:
Thank you!!!

PR #12677, opened 2015-09-26
Make sure FileCache uses correct permissions and optionally allow different permissions

christianseel:
### What does it do ?

1) Update modScript to use and force the folder/file permissions set in the `new_folder_permissions` and `new_file_permissions` settings (if those are set).

2) Introduce 2 new optional settings `new_folder_permissions_cache` and `new_file_permissions_cache` that can be set to use different folder and file permissions for the cache when using `xPDOFileCache`.
### Why is it needed ?

The file cache wasn't respecting the settings if you e.g. need `0775` permissions on folders/files. I had this now in 2 projects that I needed different file permissions for the cache files, due to restrictions and/or security concerns from the infrastructure. (E.g. multiple servers sharing the file system, each with different default permissions)
### Related issue(s)/PR(s)

https://github.com/modxcms/xpdo/pull/89

### DIFF
--- a/core/model/modx/modscript.class.php
+++ b/core/model/modx/modscript.class.php
@@ -154,7 +154,12 @@ class modScript extends modElement {
                 $script= $this->xpdo->cacheManager->generateScript($this);
             }
             if (!empty($script)) {
-                $result = $this->xpdo->cacheManager->writeFile($includeFilename, "<?php\n" . $script);
+                $options = array();
+                $folderMode = $this->xpdo->getOption('new_cache_folder_permissions', null, false);
+                if ($folderMode) $options['new_folder_permissions'] = $folderMode;
+                $fileMode = $this->xpdo->getOption('new_cache_file_permissions', null, false);
+                if ($fileMode) $options['new_file_permissions'] = $fileMode;
+                $result = $this->xpdo->cacheManager->writeFile($includeFilename, "<?php\n" . $script, 'wb' , $options);
             }
         }
         if ($result !== false) {
--- a/core/xpdo/cache/xpdocachemanager.class.php
+++ b/core/xpdo/cache/xpdocachemanager.class.php
@@ -253,10 +253,6 @@ class xPDOCacheManager {
                 }
             }
             @fclose($file);
-            if ($written !== false && $fileMode = $this->getOption('new_file_permissions', $options, false)) {
-                if (is_string($fileMode)) $fileMode = octdec($fileMode);
-                @ chmod($filename, $fileMode);
-            }
         }
         return ($written !== false);
     }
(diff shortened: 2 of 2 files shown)


opengeek:
This entire feature is based on the fact that most systems don't need to set permissions; they SHOULD use the umask of the system they are working on. This MUST be the default behavior. You ONLY set file/folder permissions if you want them to be different than the system's umask.

Also, can you explain in a little more detail why you would need separate permissions for writing cache files? That doesn't make any sense to me—those things SHOULD be managed by the server infrastructure, not the software running on it, when at all possible.

christianseel:
Sure, basically our issue in a multi-server & multi-site setup is, that umask has different values (depending on the server, the site is running on and/or the site is requested from). Basically there are frontend-servers that should only be allowed to write into the cache folder, nowhere else. And there's the master server which may write everywhere. Both servers / php processes have different users for the filesystem but they are in the same group. Which is why we need 775 in cache. But only in the cache.

It's a quite complex infastructure with extremly restrictive security settings etc…

**With my changes the umask values should still be the default!** Specific permissions are only forced if you set any of the settings like `new_file_permissions` or `new_file_permissions_cache`. From what I see the `new_file_permissions` system setting isn't set by default on new projects and needs to be optionally added as well.

opengeek:
Gotcha — I misinterpreted your "always force the desired folder permissions" commit subject to mean you were always chmod'ing and always expected those settings to have values.

christianseel:
Just renamed the settings to proper english: `new_cache_file_permissions` and `new_cache_folder_permissions`.

Also I added a check for the settings in the `loadScript()` function of the modscript class. Because previously the `includes` folder in the cache didn't use the permissions of the new files, because `loadScript()` isn't using the `set()` function (instead it direclty uses writeFile).

Mark-H:
From the looks of it, part of this pull request needs to go into xPDO (edits to the xpdocachemanager class)...

@opengeek could you comment on the process for that? Should this PR be broken up into one on modxcms/xpdo and one here? Or can changes made to xPDO in this repo be pulled back into xPDO?

opengeek:
The changes definitely need to be made to the modxcms/xpdo repository (`2.x` branch I guess, though this seems like a bug fix and could go into `2.4.x` of xPDO and Revo) and pulled into Revolution. There is no process for going the other direction, and in 3.x, xPDO is a vendor repository pulled in via composer. IOW, I don't want people to get used to making changes to dependencies in the main repo. :)

christianseel:
@Mark-H just updated the PR! I removed the modified xpdocachemanager file and created a new PR for that at https://github.com/modxcms/xpdo/pull/89

Now just both need to be merged, please :)

Mark-H:
The xPDO PR has been merged, so this could also be considered.

christianseel:
Right, ready for merging!

Mark-H:
Thank you @christianseel, merged for inclusion in 2.6.

Issue #12570, opened 2015-08-20
Help links missing target_blank


terrybarth:
The links to additional information for the settings (under + sign) open in the same window as the manager. They should open to a new window.

Alroniks:
We can check links in description field and try add `target="_blank"` to each or just add handlers to this links, but in system settings from MODX will be more easy check existing links and fix lexicons manually and for settings from extras let's left this task to authors of extras.

Alroniks:
So, I've added missed target=blank in system settings, for extras please send issues to author of extra.

terrybarth:
Great. Thank you.

Alroniks:
PR was merged for the next release, so this issue can be closed I guess. @theboxer

PR #12685, opened 2015-09-29
Added missed target=blank

alroniks:
### What does it do ?

Added missed target=blank for extranl link from system settings. 
### Why is it needed ?

It useful when you see new option and you can go to site with docs for reading in new tab and then you can return to opened tab and set value that needed.
### Related issue(s)/PR(s)

Fix for this https://github.com/modxcms/revolution/issues/12570


theboxer:
@Alroniks Not sure why, but your PR includes also some other commits, can you please clean it?

Alroniks:
@theboxer Looks like that my mirror repo not synced with current repo and if I try send commit to 2.4.x via github, git add other commits too.

Alroniks:
I've added new commit to 2.x branch https://github.com/modxcms/revolution/pull/12686

theboxer:
Thanks

PR #12688, opened 2015-10-01
Added border for grids

bezumkin:
In MODX 2.2 grids has borders, but in version 2.3 they was disappeared.

I think we should return them back.

Current grid:
[](https://file.modx.pro/files/6/a/1/6a1b9346db158cb2838f33e20b599f1b.png)

Improved grids:
[](https://file.modx.pro/files/b/7/7/b77e304b8f082e1a130ad161a399f71f.png) [](https://file.modx.pro/files/0/d/2/0d2beb4d901222972f938deefa136cba.png) [](https://file.modx.pro/files/4/6/3/463d200bc98c19c21b59b776c21a62c4.png)

### DIFF
--- a/_build/templates/default/sass/_xtheme-modx.scss
+++ b/_build/templates/default/sass/_xtheme-modx.scss
@@ -131,6 +131,7 @@
 .x-grid3 {
   background-color: transparent;
   background-image: none;
+  border: 1px solid $gridHeaderBgColor;
   border-radius: $borderRadius;
   overflow: hidden;
   padding: 0;
@@ -287,6 +288,11 @@ td.x-grid3-hd-menu-open .x-grid3-hd-inner {
   border-top: 1px solid $borderColor !important;
   color: #565550;
 }
+
+.x-grid3-row-last, .x-grid3-row-last.x-grid3-row-selected {
+  border-bottom-color: transparent !important;
+}
+
 .x-grid3-row-expanded {}
 
 .x-grid3-row-selected .x-grid3-cell-inner, .x-grid3-hd-inner {}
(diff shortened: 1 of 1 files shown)


whitebyte:
Looks much nicer!

jaygilmore:
This is purely an aesthetic design choice. Are there demonstrable UX issues that adding the border resolves? I also feel that calling the replacing of a border "improved" is your opinion and therefore not objectively improved. I'd personally be find if the borders were dropped on the rows as well but that's an opinion. I feel it gives more white space and the striping on the alternating rows and visual margins provide enough cues to indicate the edge of a table/grid.

inreti:
+1 for a border. It looks more consistently compared with the other UI areas/elements, like the buttons. Without the border the rows seem to "ghostly hang around" in the space there. This is obviously the case in @bezumkin's screenshots on the last row. Which has no border to the grid controls underneath, and the visual margin is not as indicating as I prefer to have there. 

@jaygilmore: the border for the alternating rows can imho be removed totally: 

screenshot from 2015-10-13 19 17 18

whitebyte:
I want to bring attention to another inconsistency (IMHO, of course), ispired by @inerti's screenshot:
search

We all know the purpose of the big button on the right side of the search input, aren't we?
index

It's so commonly used that it's a bit surprising to "clear filter" with it.

bezumkin:
@whitebyte This is an offtopic, but I use this search in my components


Here is [source code of widget](https://github.com/bezumkin/Tickets/blob/72be91ccd38b9c4e4b86837d3e4500c7226f8157/assets/components/tickets/js/mgr/misc/combos.js#L188-L234).

PR #12692, opened 2015-10-02
Fix problem with multiple placeholders in a system setting

ronaldex:
When using multiple placeholders ('{setting1}/{setting2}') in a single systemsetting and the first one is found and the second one isn't, then the second placeholder would be replace by '$matchValue' from the first placeholder.


Jako:
Merged to 2.5.x branch.

Issue #12690, opened 2015-10-02
Add version parameters to manager css and js


pepebe:
Just a quick idea. 

I've spend ages telling clients why the manager sometimes looks funny after updates and how to solve this problem.

How about we add the version number to script and link tags?

```
<script src="/manager/assets/modext/core/modx.js?v=2.4.1pl" type="text/javascript"></script>
```

This would force a browser to load the most recent fileversion after every update/upgrade.

Regards,

pepebe

zaigham:
This is great idea actually, we always do this with public facing websites to bust caches, so why not for manager.

:+1:

amdbuilder:
There is one potential problem with that approach.  If the resource is included on the login page, the version number would be publically revealed.  What about a random string of say four characters, generated during the setup/upgrade process?

pepebe:
@amdbuilder Good point. I like the idea of a random string. 

Taking about security. I never thought about this, but all those manager specific js and css files could also be used for finger printing.

Can anyone give a good reason why not to slim down the necessary code to a generic login.css and a generic login.js?

inreti:
Random string +1. Or random number. A 2-byte hex should be enough to mark a new css version. Could be simply stored as a system setting, whenever setup is run. As it is just a "version-like" parameter to certain urls, it shouldn't interfere with anything. 

What about using "_settings_version_token_" for the key name?

pepebe:
@inreti Sounds good to me.

Anyone volunteering for a PR? I don't see me looking into this before the end of October.

inreti:
I still have some other PRs to work on, but I can try it in the next 2 weeks.

pepebe:
If you beat me on this, I'll buy you a beer!

inreti:
Almost done. Just a few tests missing... :-)

pepebe:
Beer always gets it done :-).

pepebe:
Either pm me your address or wait for modxexpo and get it there.

inreti:
@pepebe: well, we missed the beer exchange at the modxpo :-(.

OptimusCrime:
This has been solved.

@modxbot close

PR #12700, opened 2015-10-06
Implements #12690, browser refresh for css/js after update

wuuti:
### What does it do ?

Adds a unique, short token to css/js links for manager resources. The token is computed by uuid of modx instance and the version. So after any update the version string changes therefore the token hash. Including the uuid acts like a salt to prevent recomputing and fingerprinting the version from the token. 
_Note: we could also use a random string here - but then we have to store it somewhere. I did not want to waste a system setting for a computable value, so I used the uuid+version token hash here._
### Why is it needed ?

This way you don't need to refresh your browsers cache after an upgrade (if the standard css and js have changed). Very useful because often clients forget that or don't know how to do that. 
#### Possible Enhancements

Can be used for all core js/css links in manager pages (currently not used for Extjs files for example).
Can maybe used also for package updates (if packages bring new versions of css and js with them).
### Related Issues
#12690

### DIFF
--- a/core/model/modx/modmanagercontroller.class.php
+++ b/core/model/modx/modmanagercontroller.class.php
@@ -928,6 +928,9 @@ abstract class modManagerController {
             }
         }
 
+        $versionToken = hash('adler32', $this->modx->getOption('settings_version') . $this->modx->uuid );
+        $this->setPlaceholder('versionToken', $versionToken);
+
         if (!$index) {
             $this->setPlaceholder('indexCss', $managerUrl . 'templates/default/css/index.css');
         }
--- a/manager/templates/default/header.tpl
+++ b/manager/templates/default/header.tpl
@@ -7,7 +7,7 @@
 {if $_config.manager_favicon_url}<link rel="shortcut icon" href="{$_config.manager_favicon_url}" />{/if}
 
 <link rel="stylesheet" type="text/css" href="{$_config.manager_url}assets/ext3/resources/css/ext-all-notheme-min.css" />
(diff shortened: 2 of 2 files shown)


christianseel:
Yay! Cool!

inreti:
Can this PR please be reviewed and then integrated into one of the next releases (2.4.3)? From my point of view it is a quite simple enhancement (4 lines of code) and therefore a quick review of an integrator should be possible in time. 

I recently again faced the problem of a client _not_ seeing UI changes after an update. Having to tell and explain them the necessity of a browser cache clear (and how to do it) is an absurdity today and leads to a non-professional reception of MODX by the clients. The PR provides the chance to easily get rid of this problem without any known side-effects.

jpdevries:
+1 

I think this is somewhat related to https://github.com/modxcms/revolution/pull/12611 in that https://github.com/modxcms/revolution/pull/12611 removes the server-side concatenation and minification of the common Manager files.

Another approach to this could be to change the file name itself (so serve something like `modx.jsgrps-min.2.5.0.js`), rather than add a URL parameter (the URL paramter would be stripped in 2.4.x with compress_js and compress_css enabled) but the filename will remain the same.

Mark-H:
Thanks @inreti - I've merged this for 2.4.3. I think it would be beneficial to extend it a bit further. In particular on the login page, and by also adding the version string to the other js files that are being loaded in modManagerController, possibly only if compress_js is disabled. 

The core extjs files don't really change anymore so those don't have to be reloaded, that's good as is.

PR #12702, opened 2015-10-07
TV option description for input type resource list

wuuti:
### What does it do ?

Adds examples to the option description "Where condition"
### Why is it needed ?

Was not exactly clear how to build the expression, therefore one has to google for examples. This is no more needed, examples are added to the description.

### DIFF
--- a/core/lexicon/en/tv_widget.inc.php
+++ b/core/lexicon/en/tv_widget.inc.php
@@ -110,7 +110,7 @@ $_lang['resourcelist_limit_desc'] = 'The number of Resources to limit to in the 
 $_lang['resourcelist_parents'] = 'Parents';
 $_lang['resourcelist_parents_desc'] = 'A list of IDs to grab children for the list.';
 $_lang['resourcelist_where'] = 'Where Conditions';
-$_lang['resourcelist_where_desc'] = 'A JSON object of where conditions to filter by in the query that grabs the list of Resources. (Does not support TV searching.)';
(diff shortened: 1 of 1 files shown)


core/lexicon/en/tv_widget.inc.php
```diff
@@ -110,7 +110,7 @@
 $_lang['resourcelist_parents'] = 'Parents';
 $_lang['resourcelist_parents_desc'] = 'A list of IDs to grab children for the list.';
 $_lang['resourcelist_where'] = 'Where Conditions';
-$_lang['resourcelist_where_desc'] = 'A JSON object of where conditions to filter by in the query that grabs the list of Resources. (Does not support TV searching.)';
+$_lang['resourcelist_where_desc'] = 'A JSON object of where conditions to filter by in the query that grabs the list of Resources. (Does not support TV searching.)<br/>Examples: [{"template:=":"4"}], [{"pagetitle:!=":"Home"}], [{"parent:IN":[34,56]}]';
```
pixelchutes: Wouldn't template be better treated as an `int` vs. `string`? _(e.g. no quotes)_

```json
{"template:=":4}
```

inreti: I just took the examples from http://www.sepiariver.ca/blog/modx-web/modx-quick-tip-resource-list-tvs/. Did not check if there is a difference in behaviour if you write `{"template:=":4}` instead of `{"template:=":"4"}`. Have to check that - but currently I think there is no difference due to PHPs type juggling. But you are right, examples here should be as specific as possible.
Btw: any other, maybe more complex examples are welcome.

zaigham:
:+1: Good stuff.

pixelchutes:
:+1: nice addition!

Issue #5980, opened 2011-10-29
default_template setting not used for New (child) Resource


netProphET:
netprophet created Redmine issue ID 5980

When a New Resource is created under a parent Resource via the context menu's New Resource option, the new Resource will default to the same Template as the parent. This may be desirable in some situations, but it is in contrast to the Quick Create option which uses the default_template system setting to set the Template of the new Resource.

See related ticket #5808 describing a new system setting in Evo 1.0.5 which allows the user to control which behavior they would like.

adamwintle:
@netProphET Is this possible with Form Customiser or a custom plugin?

gpsietzema:
Seems to be fixed by Matthew.

@modxbot close

sonicpunk:
#modxbughunt #1point to @gpsietzema

PR #12705, opened 2015-10-08
"inherit_parent_template" Fix for #5980

matdave:
Add option to turn off inheriting parent templates by default, and add a system setting called "inherit_parent_template" to override it.

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -185,7 +185,7 @@ class ResourceCreateManagerController extends ResourceManagerController {
      * @return int
      */
     public function getDefaultTemplate() {
-        $defaultTemplate = (isset($this->scriptProperties['template']) ? $this->scriptProperties['template'] : (!empty($this->parent->id) ? $this->parent->get('template') : $this->context->getOption('default_template', 0, $this->modx->_userConfig)));
+        $defaultTemplate = (isset($this->scriptProperties['template']) ? $this->scriptProperties['template'] : ((!empty($this->parent->id) && $this->context->getOption('inherit_parent_template', true, $this->modx->_userConfig)) ? $this->parent->get('template') : $this->context->getOption('default_template', 0, $this->modx->_userConfig)));
         $userGroups = $this->modx->user->getUserGroups();
         $c = $this->modx->newQuery('modActionDom');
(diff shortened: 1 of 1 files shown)


sottwell:
If I recall correctly, Evo has such a setting, to always use default, to use parent, or to use sibling.

rthrash:
Sibling inherit is/was useful—recall working on that back in the day. Taking a property value set on each container to toggle between those settings might be interesting. Great stuff Matt and suggestion Susan.

matdave:
If anyone wants to take on or show me where to add the "inherit_parent_template" to the system settings list, that would help

sottwell:
It's in one of the _build files, I forget which one right now, but there's one with all of the system settings.

sottwell:
_build/data/transport.core.system_settings.php

sottwell:
I just copy one of the existing ones in the same general area where I think the new one should go, then edit the relevant bits.

sottwell:
Not sure what its x-type should be, though.

inreti:
+1 for that. Regarding the child template behaviour: additionally an option per resource & template where you can list the ids of allowed templates as child resources. And a tree context menu respecting that setting and showing only those Templates/ResourceTypes instead of "Create new" everything and selecting the template then.
I don't want my customers to create a Gallery or Articles Resource anywhere in the tree (ResourceType), and I don't want them to create any Resource (Template) at places my tree setup was not expected to have them created. You can control the template selection somehow by Forms (Manager) customization, but this is a mess.

Mark-H:
Hey @matdave could you please follow the template from https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md#template-1 in your pull request? Thanks!

Mark-H:
Also it looks like you haven't added the setting to the build yet, if you need help with that just let me know.

Mark-H:
I'm tempted to close due to lack of response/follow-up, missing the data in the build, and that one line of code being so horrifically long with nested nested ternaries that I don't want to touch it with a ten foot pole. :P 

I can see some real benefit to this improvement, but it could really use some tidying up. 

@matdave think you'd be up for improving the PR a bit?

OptimusCrime:
Damn, that line should be fixed either way :(

PR #12709, opened 2015-10-10
Bugfix for failure message

Jako:
$f was removed in https://github.com/modxcms/revolution/commit/9a46c45aaf1879893d0cf162230dcdae99d94902


core/model/modx/modprocessor.class.php
```diff
@@ -1408,8 +1408,10 @@ public function cache() {
      */
     public function download() {
         $file = $this->object->get($this->nameField).'.xml';
+        $f = $this->modx->getOption('core_path', null, MODX_CORE_PATH) . 'export/' . $this->objectType . '/' . $file;
```
Alroniks: $f variable not clear. Best variant user $fileName instead $file, and $file instead $f.

PR #12717, opened 2015-10-13
processElementTags (speed) improvement

ronaldex:
Only recursively call itself when tags have been processed, there is no need to recursively call the function until depth is reached when no tags have been processed.

The depth parameter could / should be removed altogether, the recursiveness is automatically stopped when all tags are processed or none of the tags could be processed.

There is a noticeable speed gain on resources / templates / chunks with placeholders that cannot (immediately) be processed, FormIt placeholders for example.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -255,7 +255,7 @@ class modParser {
                 }
             }
             $this->mergeTagOutput($tagMap, $content);
-            if ($depth > 0) {
+            if ($processed > 0 && $depth > 0) {
                 $processed+= $this->processElementTags($parentTag, $content, $processUncacheable, $removeUnprocessed, $prefix, $suffix, $tokens, $depth);
             }
         }


exside:
me want this =) +1

OptimusCrime:
Any statistics on how big the improvement is?

ronaldex:
The improvement depends on the scenario. It varies from a few percent / ms on a clean site with just one form to ca. 15-20% on a complex site with many placeholders in templates / chunks / etc that are not always available.

Mark-H:
Sounds like the better you built your site, the less you have to gain :D 

@ronaldex I don't immediately see a CLA on your account, which is needed in order to accept the PR. Could you head over to http://modx.com/cla and submit the form, or let me know if I need to look under a different username for it? Thanks in advance!

ronaldex:
@Mark-H i've created an account, the username is 'ronaldexterkate'

Mark-H:
Thanks for your contribution @ronaldex! It has been merged in 2.5.0-rc1, I think this is one of those improvements that could really do with some real world testing on real sites, but looking good so far!

Issue #11338, opened 2014-04-30
Multiple calls to same snippet in page do not respect uncached directive.


ByteJamRob:
Tested in Revo 2.2.14 and 2.2.6.

To test, make a snippet called TEST with the following code:

return md5(uniqid(rand(),1));

Drop the following snippet calls on a page:

<p>[[!TEST]]</p>

<p>[[!TEST]]</p>

<p>[[!TEST]]</p>

Load the page. You will see the same value repeated three times. Since I have told the snippet to be uncached, this is not what I would expect to see.

A workaround is to assign a dummy parameter with different values for each snippet call:

<p>[[!TEST? &dummy=`1`]]</p>

<p>[[!TEST? &dummy=`2`]]</p>

<p>[[!TEST? &dummy=`3`]]</p>

Even though the "dummy" parameter is not referenced in the snippet's source code, this will yield the desired results.

Can Revo be updated so that these two scenarios behave the same way?

whitebyte:
MODX docs persistently treats snippets like "functions". If we pass different params to a function, we expect different results. Given that we call snipped uncached AND with different params it's pretty weird to have cached results. I suggest to treat this issue as a bug, not as a feature. It can affect parsing speed, yet this should be handled.

Bruno17:
Don't you get different outputs with different params?

HitkoDev:
@OptimusCrime It doesn't use the footprint to optimize speed, the snippet is still executed on each call, and the value from the LAST execution is inserted for each snippet tag (see https://github.com/modxcms/revolution/issues/12715)

inreti:
I have run into this years before, where a snippet had to output timing informations...

If you see the MODX request and parsing from a functional programming point of view, everything said about same footprint = same result can be considered as right. But this somehow interferes with the concept of caching. The first computes the same value everytime, a cached value just returns a previously computed value. 

[[!snippet]] should be computed every time it occurs imho. That's would people understand as "uncached" and what they expect then. Me too. So if you dare to use 10 times a snippet call like that on a page, fine, you know what you do and what performance issues result out of them. You would expect that performance thing then, so for me that's no problem here. 

If actually the uncached snippets are evaluated each time, but only the last value makes it to the result and all calls are replaced by that result, then it is a bug - depending on the concept the bug is either to evaluate the snippet more than once _or_ the bug is using only the last result for all calls.

HitkoDev:
[...]

PR #12719, opened 2015-10-14
Fix modParser identical tags issue

### LINKED ISSUE #12715: Issue: Multiple snippet tags bug
### Summary
Incorrect result when there are several identical snippet tags on the same page

### Step to reproduce
Make a snippet with the following code, then insert multiple (uncached) snippet tags on the same page.
```php
$pl = $modx->getPlaceholder('dsfsdfds');
if(!$pl) {
    $pl = 1;
} else {
    $pl++;
}
$modx->setPlaceholder('dsfsdfds', $pl);
return $pl . ' * ';
```

### Observed behavior
Output: __5 * 5 * 5 * 5 * 5 *__  (the final value of the placehoder gets outputed on each call)

### Expected behavior
Output: __1 * 2 * 3 * 4 * 5 * ...__ (the output increments with each call)

### Environment
Tested on MODX 2.4.2, probably occurs on previous versions as well. The most likely candidate for this bug is something like ```$content = str_replace($tag, $output, $content);``` in the parser.

HitkoDev:
Replaced `$tagMap` with `$tagKeys` and `$tagValues` to avoid collisions of identical tags in `modParser::processElementTags`. 
Updated `modParser::mergeTagOutput` accordingly (maintaining backward compatibility). 
Fixes https://github.com/modxcms/revolution/issues/12715  and https://github.com/modxcms/revolution/issues/11338

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -231,6 +231,7 @@ class modParser {
             $tagKeys= array ();
             $tagValues= array ();
             foreach ($tags as $tag) {
+                $token= substr($tag[1], 0, 1);
                 $value= $tag[0];
                 if (!$processUncacheable && $token === '!') {
                     if ($removeUnprocessed) {
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -227,12 +227,14 @@ class modParser {
         $content = $this->modx->documentOutput;
         unset($this->modx->documentOutput);
         if ($collected= $this->collectElementTags($content, $tags, $prefix, $suffix, $tokens)) {
-            $tagMap= array ();
+            /* using an associative array causes collisions with identical tags */
+            $tagKeys= array ();
+            $tagValues= array ();
             foreach ($tags as $tag) {
-                $token= substr($tag[1], 0, 1);
+                $value= $tag[0];
                 if (!$processUncacheable && $token === '!') {
                     if ($removeUnprocessed) {
-                        $tagMap[$tag[0]]= '';
+                        $value= '';
                     }
                 }
                 elseif (!empty ($tokens) && !in_array($token, $tokens)) {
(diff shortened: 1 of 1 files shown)


Mark-H:
Per the discussion in #11338 (in particular [Jason's summary](https://github.com/modxcms/revolution/issues/11338#issuecomment-148102103)), I've tagged this for 3.0. 

Given the impact some unrelated parsing changes had in the 2.4.3/2.4.4 releases, I think we also need to start investing some time into creating some more unit tests that cover real world examples to make sure we can make changes like these confidently.

bezumkin (approved):
Tested and looks like everything is okay, but I strongly recommend to change `mergeTagOutput` to this
```php
public function mergeTagOutput(array $tagMap, & $content, array $tagValues= array()) {
    if (!empty ($content) && is_array($tagMap) && !empty($tagMap)) {
        if (is_array($tagValues) && !empty ($tagValues)) {
            foreach ($tagMap as $idx => $key) {
                $content = preg_replace('#' . preg_quote($key) . '#', $tagValues[$idx], $content, 1);
            }
        } else {
            $content = str_replace(array_keys($tagMap), array_values($tagMap), $content);
        }
    }
}
```
It work the same way, but much more understandable.

HitkoDev:
@bezumkin Here you're running `preg_replace` in a loop, which is far less efficient `[O(n * m/2)]` than the original single-loop implementation `[O(n + m)]`

bezumkin:
@HitkoDev Yes, you are right. For 1000 tags my code will be 50 times slower (I just tested).

Okay, let it is will be longer, but faster =)

JoshuaLuckers:
@opengeek you self-assigned this PR: are we going to approve this or is there still some work to be done?

opengeek:
I unassigned myself. I personally am not a fan of this change as this will alter the intended behavior of the parser. But I've left it open because others seem to want it changed.

alroniks (commented):
I don't want to lock it by myself, but implementation is not clear for me, especially in `mergeTagOutput` method. I agree with Mark, we should cover it by unit tests before we are changing it.

HitkoDev:
### processElementTags
The basic idea here is that instead of using `array(tag => replacement)` we use two arrays, such that `$tag[i] => $replacement[i]`. This prevents collisions when `$tag[i] === $tag[j]`, but `$replacement[i] !== $replacement[j]`.

### mergeTagOutput
This one is more complex. The main idea is that tags are extracted and parsed in order in which they appear in content, so `$tag[i+1]` always comes after `$tag[i]`.

- we find `$tag[i]` starting from `$begin`
- extract content between `$begin` and `$tag[i]`
- concatenate `$replacement[i]`
- update `$begin` to point to the end of `$tag[i]`

That way we never search the same part of content twice, and ensure each occurrence of a tag gets the corresponding replacement.

---

I'll have a look at recent changes and merge conflicts, and try to clarify this in the code.

JoshuaLuckers:
We should definitely add more Unit Tests for the parser in 3.x before we should consider implementing a change like this.

Since this is a really old PR and still targeting 2.x I'm going to close it for now.

PR #12720, opened 2015-10-14
Fix Profile  Password reset to have same field labels as User panel password reset

hansek:
### What does it do ?

User panel reset form use different field labels than Profile panel reset form. This is a bit "weird".
### Why is it needed ?

MODX users should have uniform look on the same fields (they should have same labels).
##


hansek:
Wrong branch, sorry.

Issue #12645, opened 2015-09-10
modMail error - can't embed images


whitebyte:
### Summary
It's impossible to embed images using modMail service.

### Step to reproduce
Try to use `$modx->mail->embedImage();`

### Observed behavior
Image is not embedded due to the error in MODX phpMailer wrapper

### Expected behavior
Image should be embedded

### Environment
MODX 2.3.x, 2.4.0

The error is in `/core/model/modx/mail/modphpmailer.class.php`, line 251:
```
$this->mailer->AddEmbeddedImage("/".$image,$cid,$name,$encoding,$type);
```
As you see, leading slash is always added, making every path absolute. As a quick fix one should remove leading slash:
```
$this->mailer->AddEmbeddedImage($image,$cid,$name,$encoding,$type);
```
to mirror actual phpMailer AddEmbeddedImage function.

OT: what is the rationale behind modMail at all? I don't see any advantages of this wrapper. It's poorly documented and difficult to update (e.g. we can't just drop in a new version of phpMailer. Actual vesion is 5.2.12 while MODX still uses 5.2.8)

Mark-H:
> OT: what is the rationale behind modMail at all? I don't see any advantages of this wrapper. It's poorly documented and difficult to update (e.g. we can't just drop in a new version of phpMailer. Actual vesion is 5.2.12 while MODX still uses 5.2.8)

The ability to use other mailers rather than just phpmailer :) There's only one other modMail instance I can think of which is for Swift, but there might be others.

Alroniks:
I agree with @Mark-H 
It's good practice to use wrapper (adaprter design pattern or the same) for doing operations, but have ability transparently change implementation.
Nonetheless, bug with image path should be fixed.

PR #12725, opened 2015-10-17
Fix 12645

whitebyte:
This PR removes redunant slash in the modphpmailer.class.php that prevents embedding images into email. Currently every path that we provide to modmailer becomes absolute because of leading slash. This is undocumented and doesn't mirror actual phpmailer function prototype.
### Related issue(s)/PR(s)
#12645


whitebyte:
Oops, sry.

PR #12734, opened 2015-10-24
Set the leaf property to true instead of 1

sergant210:
### Problem description

If the leaf property of elements is set to 1, function TreeNode.isLeaf () stops working correctly. 

```
//TreeNode.isLeaf function
isLeaf : function(){
    return this.leaf === true;
}
```

### DIFF
--- a/core/model/modx/processors/element/getnodes.class.php
+++ b/core/model/modx/processors/element/getnodes.class.php
@@ -392,7 +392,7 @@ class modElementGetNodesProcessor extends modProcessor {
                 'id' => 'n_'.$elementIdentifier.'_element_'.$element->get('id').'_'.$element->get('category'),
                 'pk' => $element->get('id'),
                 'category' => $categoryId,
-                'leaf' => 1,
+                'leaf' => true,
                 'name' => $name,
                 'cls' => implode(' ', $class),
                 'iconCls' => 'icon ' . ($element->get('icon') ? $element->get('icon') : ($element->get('static') ? 'icon-file-text-o' : 'icon-file-o')),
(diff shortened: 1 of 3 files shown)


exside:
I found many more of these, maybe a folder search in the modx repo will reveal them to you as well =), for example in:

```
/core/model/modx/processors/security/documentgroup/getnodes.php
/core/model/modx/processors/security/documentgroup/getpairingnodes.php
/core/model/modx/processors/element/getnodes.class.php -> there is more than one as far as I can see!
```

Mark-H:
Hi @sergant210 - it doesn't look like there's a CLA for you on file. Could you please head over to modx.com/cla to get that squared away?

We also have a template for pull request descriptions at https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md#template-1 - if you follow those three questions, that gives us a bit more information on how to test your contribution and provides a useful log of things. Could you please elaborate a bit on that, even though this is such a small patch (in terms of line of code, not in effort figuring it out I'm sure!)

sergant210:
### What does it do ?
Set the leaf property of elements from the elements tree to true instead of 1. This must be also fixed in the files described by exside above.

### Why is it needed ?
Without this fix you can not check tree nodes by the function Node.isLeaf().

### Related issue(s)/PR(s)
Not found.

Mark-H:
Hi @sergant210 - would it be possible for you to update the pull request with the other files noted as well? If you don't have time for that I can get this merged, but it sounds like this is an area where we could get more consistency in general.

sergant210:
@Mark-H Ok. I'll try.

sergant210:
@Mark-H Done.

PR #12736, opened 2015-10-24
 Remove redundant code from manager/index.php 

whitebyte:
### What does it do ?

This removes some code duplications and superfluous checks from MODX manager index.php


whitebyte:
Aagrh.. sry for the noise. Wrong branch again.

Issue #12735, opened 2015-10-24
date.timezone setting should be requested on install


Jako:
The current behaviour is a bit annoying, if the date.timezone setting is missing in PHP:

```
FATAL ERROR: MODX Setup cannot continue.

To use PHP 5.3.0+, you must set the date.timezone setting in your php.ini. Please do set it to a proper timezone before proceeding. A list can be found here.
```

Two options are thinkable: use a default date.timezone setting or prepare an input for the user to generate it and try a @ini_set for this before showing the error message.

digitalpenguin:
+1

PR #12738, opened 2015-10-25
Fixing #12735 missing date.timezone setting on install

Jako:
### What does it do ?

Switch to a default timezone if no timezone is set in php.ini.
### Why is it needed ?

A big german hosting provider (1&1) does not set this setting and it is better to have a default timezone than stopping the installation and let (maybe inexperienced) users fix that by a custom php.ini (this is the answer of the hosting providers for that issue).
### Related issue(s)/PR(s)
#12735


rtripault:
Makes sense, and looks good to me, :+1:

Mark-H:
Thanks @jako for another useful contribution :)

PR #12746, opened 2015-10-31
Try to make use of the maximum available height for logs textarea

### LINKED ISSUE #5942: Make the Error Log Textarea larger
everettg_99 created Redmine issue ID 5942

It'd be nice if the textarea in the manager for the Error log were a bit larger ... so it filled out the page more... 50 rows? 100 rows?  not sure exactly, but the readout there would be more readable if the textarea were larger. 

Thanks!

### LINKED ISSUE #12743: Error Log view
Utilize full height or allow manual resizing. Otherwise text feels cramped in the text box, forcing too much scrolling.

screen shot 2015-10-30 at 3 16 50 pm

rtripault:
### What does it do ?

Sets the error log textarea height (the one holding the log content) to the maximum available client viewport height
### Why is it needed ?

The current "growth" is set to 400px height max. Meaning it leaves a lot of unused "white space" below the textarea, which could be used to display more log data.
### Steps to reproduce issue
- On a screen with more than 600px height, just head to the error log and see the textarea is not "maximized"
### Related issue(s)/PR(s)
- Fixes #12743
- Fixes #5942

### DIFF
--- a/manager/assets/modext/widgets/system/modx.panel.error.log.js
+++ b/manager/assets/modext/widgets/system/modx.panel.error.log.js
@@ -3,7 +3,7 @@ MODx.panel.ErrorLog = function(config) {
     Ext.applyIf(config,{
         url: MODx.config.connector_url
         ,id: 'modx-panel-error-log'
-		,cls: 'container'
+        ,cls: 'container'
         ,baseParams: {
             action: 'system/errorlog/clear'
         }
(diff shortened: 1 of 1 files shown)


rtripault:
FWIW, duplicate of #5942.
All in all, PR sent to fix that behavior

jpdevries:
would be great to make full use of available space for editing elements and resource content also.

Mark-H:
Very cool, nicely done @rtripault! :)

PR #12747, opened 2015-10-31
Added resource ids to OnResourceAutoPublish param

christianseel:
### What does it do ?

This PR extends the `$results` parameter of the OnResourceAutoPublish event with `$results['published_resources']` and `$results['unpublished_resources']`. Those give developers, besides the number of published/unpublished resources, also the IDs of those resources.
It actually includes the id, context_key and pub_date.
### Why is it needed ?

Now you can write a plugin that clears custom cache partitions for the auto-published resource or it's parents because you get the resource id from the event. Without getting the IDs, the event is pretty useless from my perspective.

### DIFF
--- a/core/model/modx/modcachemanager.class.php
+++ b/core/model/modx/modcachemanager.class.php
@@ -605,11 +605,24 @@ class modCacheManager extends xPDOCacheManager {
      */
     public function autoPublish(array $options = array()) {
         $publishingResults= array();
-        /* publish and unpublish resources using pub_date and unpub_date checks */
         $tblResource= $this->modx->getTableName('modResource');
         $timeNow= time();
+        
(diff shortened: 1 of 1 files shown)


rtripault:
:+1: definitely useful when dealing with custom cache partitions!

christianseel:
@rtripault any chance you can merge this as well?

Mark-H:
Thanks @christianseel for your contribution :) I've merged it into 2.x for inclusion in 2.5.

It might make sense to extend this behaviour a bit further so it also includes the `unpub_date` in `unpublished_resources`? I guess the most important parts are the ID and the the context_key though, and with those you can get any value, so not a very high priority.

christianseel:
Thanks Mark! Submitted another PR that adds unpub_date!

PR #12750, opened 2015-11-02
Fix lexicon service init order to extension packages services can use lexicons

hansek:
### What does it do ?

Fix bug which disallow to load default lexicons in component using autoloading of service by _extension_packages_ variable.
### Why is it needed ?

To allow component with automatic service initialization from _extension_packages_ to load lexicons in class constructor. Currently in 2.4.2 lexicon service is not initialized before load of Extension packages.

In main model class then you can use:

```
public function __construct(modX &$modx, array $options = array()) {
    $this->modx->lexicon->load('cmp:default');
}
```

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -530,10 +530,10 @@ class modX extends xPDO {
             $this->getCacheManager();
             $this->getConfig();
             $this->_initContext($contextKey, false, $options);
+            $this->_initCulture($options);
             $this->_loadExtensionPackages($options);
             $this->_initSession($options);
             $this->_initErrorHandler($options);
-            $this->_initCulture($options);
 
             $this->getService('registry', 'registry.modRegistry');
 


hansek:
Is here any issue with this PR or can someone merge it? :wink:

rtripault:
I'm not sure this is an issue, but it seems that `modX::_initCulture` is able to load data from the session (not sure if that's even used in the core), so i would expect the session to be initialized first.

For what it worth, to handle the issue you are aiming to solve, i usually put the following in my service class constructor

```php
if (!$this->modx->lexicon) {
    $this->modx->getService('lexicon', 'modLexicon');
}
$this->modx->lexicon->load("namespace:topic");
```

hansek:
I agree about the Session, but still if there is a way how to "autoload" service through `_loadExtensionPackages()`, it should be able load lexicon's topics as well.

And your `if` is here something like try/catch for me ;) 

After all thanks for your code snippet, I'll put it in my code as a __temporary solution__ and we'll see if someone other take a look on this PR. 

Maybe @opengeek can take a look and make it more clear to us  ;) (thanks)

rtripault:
> And your if is here something like try/catch for me ;)

Definitely! To me, services are pieces of code, not necessarily available/instantiated. So i'd say it's your job to make sure the service you depend on is available (for example, when you want to use the `modMail` service, you `getService` first -- albeit the service might already be loaded/available).

I believe having us assuming the lexicon service is already loaded and available to use in our own services is a "legacy" of old documentation...

Anyway, just trying to share my understanding of the architecture/services, nothing really related to your proposal (hoping that makes sense).

hansek:
Yeah, totaly agree with you, but still think lexicon service is one of main services in the Manager and should be loaded automatically in this place ;)

Thanks for your opinion ;)

hansek:
@opengeek Could you take a look on this one? We're not sure if this PR is appropriate to fix the bug or I miss something and PR can be closed/cancelled.

opengeek:
Did a change made recently introduce this "bug"? Or are you just calling this a bug because it never did this but you expected it should?

hansek:
You're right, it is only my expectation, and as I found this easily looking solution, I supposed, this should be fixed.

But as is discussed with Romaine sooner I agree with his solution to check if service is initiated and if not then initiate it by my self.

I'm just curious if my expectation of "a bug" is correct or not and then we can close this PR.

opengeek:
I think this would be an enhancement. Unfortunately, this is a bit of a catch-22 issue — though it would be nice to be able to do this, the change would break user-specific culture settings since you are forcing it to load before the user session is loaded. And there may even be other consequences I'm missing from a quick review of the code.

hansek:
Ok, thank you for your insights.

For now I closing this PR.

PR #12772, opened 2015-11-21
Returning "user not found." for the page title if user does not exist

tehsquidge:
### What does it do ?

Returning "user not found." for the page title if user does not exist when trying to edit.
### Why is it needed ?

Currently it will try and return a page title which references the user object, which will be null if the user does not exist. This creates a "PHP Fatal error" and a 500 error. With this fix the "user not found." message can actually appear.


Mark-H:
Hi @tehsquidge, thanks for your contribution!

This pull request should have been based off the 2.4.x branch and also targeting that branch rather than develop, but I think we'll manage to merge it. You can find out more about the branching strategy we're using, please have a look at https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md#choosing-the-correct-branch :)

Mark-H:
Merged into 2.4.3 in https://github.com/modxcms/revolution/commit/41858b2b502a418aebb1495112e7e3911cf21061, thanks again!

Issue #12698, opened 2015-10-06
Expanding instead of editing


mishantrop:
Hello. 
Could you create the system option e.g. "expand_parent = 1/0"? It would be nice, because icon for expanding is so small (one reason).
Expanding instead of editing when clicking on resource (isfolder=1) in resources tree.

whitebyte:
How will we edit container nodes in this case? Especially when we haven't "right click" option (tablets etc).

sottwell:
tablets etc "right-click" if you hold on an ordinary click, at least Adroid does. Anyway, what about double-click for edit?

sottwell:
That should be fairly easy to remember, as all of the grids edit with double-click.

mishantrop:
whitebyte, continuous tap is replacement of right click. At least that's in my case (Android 5).

vasyahuyasa:
I like behaviour with small arrow as it is now.

Mark-H:
Since #12773, clicking outside the text but inside the list item will expand instead of edit the resource.

Issue #12375, opened 2015-02-25
Feature Request: Change default action for containers in resource tree.


digitalpenguin:
Peoples first reaction to expand a conainer in the resource tree is to click on the name instead of the arrow.
This opens the edit resource screen for that container and they get confused.

Can we make it so the default action for clicking on a container's name is to expand?
Only for resources that are tagged as containers of course.
And just add an edit option to its context menu?

Cheers!

OptimusCrime:
I think this would be even more confusing. Having left click to two different actions in the menu would in most cases not make it easier for the users.

I suggest telling people, with emphasis, that they need to click on the arrow to expand/collaps.

OptimusCrime:
Closing due to lack of feedback ad age. There's some improvements done to this in the overhaul of the manager in #12773 .

Feel free to reopen or comment if you feel this issue is not resolved sufficiently enough.

@modxbot close

Issue #756, opened 2013-12-19
Please change click behaviour for nested elements in the tree


pepebe:
After a couple of years watching clients fiddling around with the resource tree I have come to the conclusion that the arrows in front of resource containers are a serious UI problem. 

Its much to easy to open a resource instead of unfolding the parent container.

I would suggest to unfold the parent container on clicking anything except the resource title (and perhaps the icon directly in front of it. This would make it much easier to navigate through the tree.

Please have a heart for people with clumsy fingers or bad eye-sight.

Regards,

pepebe

sottwell:
At least, as of 4.0 double-clicking almost opens the branch. The child resource briefly shows half-hidden by the parent, then it disappears as the page is reloaded with the editing form.

adamwintle:
I think the latest design from 2.3 handles this quite well, its certainly harder to miss-click and accidently open that branch and I can't think of a way this could further be improved. Going to close this issue.

@modxbot close

PR #12773, opened 2015-11-21
Enhancement: Improve usability of tree by preventing accidental editing of containers

exside:
### What does it do ?

This small change improves the usability of the resource tree by preventing accidental opening of the resource edit page when the intention was actually to open the container/"folder". It's a bit hacky (as it's relying on the tagName of the JS event.target)...but the problem here is, that the ExtJS node basically is the whole width of the tree and I couldn't find a better way to do this right now, suggestions of improving or making it less hacky are very welcome!

modx-treeclickarea
### Why is it needed ?

In the case of a container resource in the resource tree, the only way of opening the "folder"/container was to click that very tiny arrow icon to the left of the node, clicking anywhere else (even the folder icon) would open the resource edit page, which is an often heard complaint from clients and fellow MODXers.
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/issues/12375, https://github.com/modxcms/revolution/issues/756, https://github.com/modxcms/revolution/issues/12698

### DIFF
--- a/manager/assets/modext/widgets/core/tree/modx.tree.js
+++ b/manager/assets/modext/widgets/core/tree/modx.tree.js
@@ -459,7 +459,9 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
         if (n.attributes.page && n.attributes.page !== '') {
             if (e.button == 1) return window.open(n.attributes.page,'_blank');
             else if (e.ctrlKey == 1 || e.metaKey == 1 || e.shiftKey == 1) return window.open(n.attributes.page);
-            MODx.loadPage(n.attributes.page);
+            else if (e.target.tagName == 'SPAN') MODx.loadPage(n.attributes.page); // only open the edit page when clicking on the text and nothing else (e.g. icon/empty space)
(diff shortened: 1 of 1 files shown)


jpdevries:
Nice!!! @Mark-H this reminds me of our conversation about how the Eureka tree works, which a little different than the MODX tree but works very similar to what this PR suggests (in that clicking a folder expands or collapses it in the tree and clicking the filename itself selects it) which are two completely different actions.
http://markup.tips/eureka

My question is this, if this PR is merged could be also get rid of those tiny little arrows completely? The folder icon itself indicates if things are open or not.

exside:
@jpdevries not sure about removing the arrow, I agree with you about the folder icon showing the state, but maybe that's just for "us" devs...for a person who is used to a kind of windows-explorer type of "filesystem" interaction that little arrow probably creates a familiar interface/experience. But no opposition from my side when a majority agrees with removing that arrow!

jpdevries:
@exside ya, some people might be up in arms about removing the arrow. Who knows. I think the tree in general is a big unfamiliar to end users.

We do kind of need the arrows for expanding and collapsing things like contexts so maybe it should stay for consistency

Mark-H:
Thanks @exside - works very intuitively. Will be part of 2.5.

PR #12783, opened 2015-11-27
Enhancement: Allow search for resource ID in Uberbar

exside:
### What does it do ?

Added one line to the where query in the search processor that also looks up the ID column for resources.

modx enhancement search resid
### Why is it needed ?

Many times feedback from MODX involves resource ID's (ex. duplicate resource URIs when clearing cache etc.) and the only way to find resources like that is via the hidden ?a=search URL (you have to know that to get to the advanced search page where you could enter the resource ID)...this tiny PR just adds this very necessary and IMO useful functionality to the uberbar.
### Related issue(s)/PR(s)

This is by far not as nice as what @pepebe suggested in https://github.com/modxcms/revolution/pull/12515 or what's currently discussed in https://github.com/modxcms/revolution/pull/12533 but it's a start... also related are other issues in the two just mentioned: https://github.com/modxcms/revolution/issues/12283, https://github.com/modxcms/revolution/issues/11117

### DIFF
--- a/core/model/modx/processors/search/search.class.php
+++ b/core/model/modx/processors/search/search.class.php
@@ -153,6 +153,7 @@ class modSearchProcessor extends modProcessor
             'OR:alias:LIKE' => '%' . $this->query .'%',
             'OR:description:LIKE' => '%' . $this->query .'%',
             'OR:introtext:LIKE' => '%' . $this->query .'%',
+            'OR:id:=' => $this->query,
         ));
         $c->sortby('createdon', 'DESC');
 
--- a/core/model/modx/processors/search/search.class.php
+++ b/core/model/modx/processors/search/search.class.php
@@ -163,7 +163,7 @@ class modSearchProcessor extends modProcessor
         /** @var modResource $record */
         foreach ($collection as $record) {
             $this->results[] = array(
-                'name' => $record->get('pagetitle'),
+                'name' => $this->modx->hasPermission('tree_show_resource_ids') ? $record->get('pagetitle') . ' (' . $record->get('id') . ')' : $record->get('pagetitle'),
                 '_action' => 'resource/update&id=' . $record->get('id'),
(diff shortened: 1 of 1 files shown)


garryn:
Awesome, simple and it's a start indeed ... I had this question the other day. 👍

christianseel:
👍🏼 please merge this asap!

pepebe:
Great!
One sidenote: I think the return value should also include the id (for example "foo (34)").

exside:
@pepebe agree, but that would probably be a bit more complicated, no? I mean that would have to respect the permission if one can see the resource id's in the tree etc.

exside:
done, wasn't that hard =)

modx enhancement search resid 2

pepebe:
THANKS! Is there even a permission to see ids or is it just a system setting for the tree? Anyway. IDs are a must. You need them all the time.

exside:
as far as I know it's a permission (is there a system setting?) called "tree_show_resource_ids"

sottwell:
I thought the whole point of the "uberbar" was to also show elements? Shouldn't entering an integer also show up any elements with that ID?

OptimusCrime:
Agree with @sottwell

sottwell:
Actually it's quite simple to add the original Search feature to the menu, just create a new menu item "Search" with the action "search" and required permissions "search". Then drag it around to wherever you want it.

exside:
It's absolutely possible to also add that id-lookup for other things than resources, the question is, is it useful? I never search for an element by ID...but if you guys do that, no problem to add it =)!

davidpede:
Yes please! search by id would be really useful

exside:
erm, sorry guys, seems like I was too slow... =) @pixelchutes, I cannot push to that branch anymore once it's merged right?

Mark-H:
Please create a new branch/PR for that @exside :)

PR #12787, opened 2015-11-29
html-first autofocus on login screen

jpdevries:
currently the username receives focus via js, this does the same with
html5 by making use of the autofocus attribute


Mark-H:
Please remember to use the PR template, but thanks again ;) Merged in https://github.com/modxcms/revolution/commit/79d0d55e4a1ab9582d5a37c04c501b3a84e04302

PR #12790, opened 2015-12-01
Setting fa-css-prefix back to default

jpdevries:
### What does it do ?

Removes `$fa-css-prefix:       icon;` which changes the font-awesome prefix from the default `.fa` to `.icon`. This is a breaking change and targets 3.x.
### Why is it needed ?

So that when people include plugins in the manager our icon classes are
compatible with the assumed default `.fa`. This fixes the issue of blank/missing icons in the manager for plugins that assume icon classes of `.fa` will be present.
### Related issue(s)/PR(s)

n/a

This PR does not update all the references to `.icon` throughout the codebase as I figured this should be discussed first. It is expected that if this were tested as is all the icons in the manager would disappear. It would be an effort to update the references but if this is agreed upon I think it would be worthwhile. 

Another solution would be to ship "double CSS" so there are classes for both `.fa` and `.icon` but that feels kind of lousy.

This issue has been discussed at MODX.org https://modxcommunity.slack.com/archives/development/p1448917321000397

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1,7 +1,6 @@
 @import "bourbon/bourbon";
 $border-box-sizing:false;
 @import "neat/neat";
-$fa-css-prefix:       icon;
 @import "font-awesome/font-awesome";
 .#{$fa-css-prefix}-large { @extend .#{$fa-css-prefix}-lg; }
 @import "image-set";


Mark-H:
Slightly more permanent log of that chat is at https://logs.modx.org/%23development/2015-11-30.log, starting at [16:02:02]

Mark-H:
I like Jason's suggestion at 16:39:24:
> what's the real cost of that? i.e. would it be worth it to introduce both for now, deprecate icon-, and then remove icon- finally in 3.x ?

jpdevries:
I like Jason's suggestion of shipping both for 2.x. I was thinking given breaking changes in 3.x we could switch it back to the default.

pixelchutes:
:+1: I agree that is a good plan for `2.x`.

IIRC, we retained `icon` (vs `fa`) for BC so we could jump from FA 3.x to FA 4.x as part of the MODX 2.3 redesign. Definitely will come across as less "disruptive" if people's custom Resource icons / Template icons, etc don't "break" until 3.x.

opengeek:
Is this something someone can take over and re-prepare for 2.x as suggested?

alroniks:
I'll try tomorrow Jason

alroniks:
@opengeek I've created [new PR](https://github.com/modxcms/revolution/pull/14031) for 3.x where both variants work. MODX 3 tries to keep BC, so better to delete old style in 3.1 or higher version.
This PR can be closed, I guess.

PR #12800, opened 2015-12-05
Shows a 'please wait' message while modSystemRefreshurisProcessor runs.

### LINKED ISSUE #12642: Issue: Clear Cache > Refresh URIs doesn't trigger Console modal or any user feedback
### Summary
The standard Clear Cache function immediately triggers the Console and gives feedback of progress; during this time the user cannot move away to another page and must wait for the cache clearing process to finish. 

When selecting the Refresh URIs function there's no immediately feedback to the user, which can make it appear as if nothing has happened or that the URIs are not being refreshed. 

Then after a few seconds the Console modal appears and gives feedback.

### Step to reproduce
Go to top menu into Clear Cache, then sub-menu Refresh URIs.

### Observed behavior
No feedback to user if URIs are being refreshed or not.

### Expected behavior
There should be some feedback to the user that something is happening, even if it's just the Console modal with a "please wait..." message, or some details of the refresh progress similar to the Clear Cache readout. 

Overall this helps the UX of the Manager, and prevents users from becoming confused, or clicking the menu item multiple times to try and trigger the function. 

If this function runs in the background, and it is safe for the user to navigate away from the current page, then this should also be communicated back to the user. A suggestion could be perhaps use the Growl-like top-right notifications (also used when saving a Resource) to inform the user that the Refresh URIs is in progress in the background. 

### Environment
MODX version 2.4.0-pl

pixelchutes:
### What does it do ?

Shows a **Please Wait...** message status window while modSystemRefreshurisProcessor is running. _(via Clear Cache > Refresh URIs)_
### Why is it needed ?

Provides feedback to the user that something is happening and in progress.
### Related issue(s)/PR(s)

Fixes #12642

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -187,6 +187,11 @@ Ext.extend(MODx,Ext.Component,{
 
     ,refreshURIs: function() {
         var topic = '/refreshuris/';
+        MODx.msg.status({
+            title: _('please_wait'),
+            message: _('refreshuris_desc'),
+            dontHide: true
+        });
         MODx.Ajax.request({
             url: MODx.config.connector_url
             ,params: {
@@ -888,4 +893,3 @@ Ext.extend(MODx.HttpProvider, Ext.state.Provider, {
         Ext.Ajax.request(o);
     }
 });
-


Mark-H:
Thanks Mike, very useful :)

PR #12803, opened 2015-12-05
Fix branch for travis badge in readme

Mark-H:
### What does it do

The travis badge shows the build status. This was set to the develop branch, but after changing the branching strategy, this needs to be updated for the 2.x branch. 
### Why is it needed

Showing the accurate build status. 
### Related issue(s)/PR(s)

None, though while merging #12801 I noticed it was set wrong.

PR #12804, opened 2015-12-05
Fix travis builds on PHP 5.5

Mark-H:
### What does it do ?

Based on fix in http://andreas.heigl.org/2015/11/12/testing-code-with-phpunit-on-travis-ci-for-php-5-5-and-php7/ and the discussion in https://github.com/travis-ci/travis-ci/issues/5206#issuecomment-162041397, this adjustment makes sure that on PHP 5.5, phpunit 4.x is loaded rather than 5.x which is not compatible with PHP 5.5, but required for PHP7.
### Why is it needed ?

The travis build indicates it fails on PHP 5.5 due to an unsupported version of phpunit. 
### Related issue(s)/PR(s)

Brief discussion here: https://github.com/modxcms/revolution/pull/12803#issuecomment-162224698

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,6 @@ mysql:
   encoding: utf8
 
 before_script: 'cd _build/test && ./generateConfigs.sh'
+
+before_install:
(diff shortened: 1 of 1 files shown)


Mark-H:
Well that's weird, it works from my fork (https://travis-ci.org/Mark-H/revolution/builds/95083691) but not here (https://travis-ci.org/modxcms/revolution/builds/95084108)

Mark-H:
:tada:

Mark-H:
Took some tweaking, but this is finally green again on PHP 5.5.

Issue #12797, opened 2015-12-02
Possible PHP7 issues (php7cc report)


exside:
I ran the PHP7 Compatibility Checker https://github.com/sstalle/php7cc against the MODX git repository, the following report was the result. Not sure if all of that is actually a problem or relevant, just thought I'll post it here for reference.

```
> php7cc /Git/MODX/revolution

/Git/MODX/revolution/core/model/aws/lib/yaml/lib/sfYamlParser.php
> Line 96: Possible object property creation during by-reference assignment
    $parser->refs =& $this->refs;
> Line 108: Possible object property creation during by-reference assignment
    $parser->refs =& $this->refs;
> Line 150: Possible object property creation during by-reference assignment
    $parser->refs =& $this->refs;
> Line 202: Possible object property creation during by-reference assignment
    $parser->refs =& $this->refs;


/Git/MODX/revolution/core/model/aws/sdk.class.php
> Line 880: Function argument(s) returned by "func_get_args" might have been modified
    func_get_args();
> Line 1227: Possible object property creation during by-reference assignment
    $this->batch_object =& $this->internal_batch_object;
> Line 1232: Possible object property creation during by-reference assignment
    $this->batch_object =& $this->internal_batch_object;


/Git/MODX/revolution/core/model/aws/services/cloudfront.class.php
> Line 147: Function argument(s) returned by "func_get_args" might have been modified
    func_get_args();


/Git/MODX/revolution/core/model/aws/services/s3.class.php
> Line 355: Function argument(s) returned by "func_get_args" might have been modified
    func_get_args();


/Git/MODX/revolution/core/model/modx/mail/phpmailer/class.phpmailer.php
> Line 2222: Removed function "set_magic_quotes_runtime" called
    set_magic_quotes_runtime(0);
> Line 2231: Removed function "set_magic_quotes_runtime" called
    set_magic_quotes_runtime($magic_quotes);


/Git/MODX/revolution/core/model/modx/modcachemanager.class.php
> Line 36: Possible object property creation during by-reference assignment
    $this->modx =& $this->xpdo;


/Git/MODX/revolution/core/model/modx/modelement.class.php
> Line 139: Nested by-reference foreach loop, make sure there is no iteration over the same array
    foreach ($property['options'] as &$option) {
    }


/Git/MODX/revolution/core/model/modx/modmanagercontroller.class.php
> Line 891: Possible object property creation during by-reference assignment
[...]

PR #12807, opened 2015-12-06
Update smarty to version 3.1.27

exside:
### What does it do ?

Remove all files in core/model/smarty/ and add all files of the current smarty release 3.1.27.
### Why is it needed ?

a) To keep up to date with major improvements in the library (speed etc.) and b) to avoid possible issues with no available PHP 7.
### Related issue(s)/PR(s)

Just marginally https://github.com/modxcms/revolution/issues/12797 as the PHP 7 Compatibility checker also lists one smarty class which could be an issue in PHP7 (unconfirmed)
### Stuff to be aware of

On my git installation this switch to the current version works without any hickups, the problem for me is that I'm not very familiar with smarty and thus may not be able to test this properly...in my case the manager works exactly like before, cleared the cache folder manually and extensively browsed the manager, but yeah, that's kinda "hobby" testing so if anybody has some concerns about this update, please let me know how I can test this PR better!


Mark-H:
Thanks @exside - merged into 2.5.

rtripault:
Some Smarty method signature changed, making `modSmarty` override causing the following error : `(ERROR @ /app/vendor/modx/revolution/core/model/modx/smarty/modsmarty.class.php : 37) PHP warning: Declaration of modSmarty::display($template, $cache_id = NULL, $compile_id = NULL, $parent = NULL) should be compatible with Smarty::display($template = NULL, $cache_id = NULL, $compile_id = NULL, $parent = NULL)`.

I believe adding a default `null` value to `$template` would do the trick while still being Backward Compatible (i might be wrong, but not seeing anyone extending `modSmarty`).

Also, it appears Smarty became more "strict" and now rises warnings when trying to access undefined variables/indexes : `(WARN @ /app/vendor/modx/revolution/core/cache/mgr/smarty/app/65fb5a01a40da9eec417ba32a0f289537ef2107d_0.file.header.tpl.php : 81) PHP notice: Undefined index: a`. To fix that, i guess using the `|default` modifier (http://www.smarty.net/docs/en/language.modifier.default.tpl) could to the trick (i remind doing that in PR #12063)

exside:
@rtripault how do you get these and where (when accessing which pages in the manager)? should I set the default MODX log level to debug or warn to get them?

rtripault:
Right, my log level is at warn  (`3`).
Those errors appears on every manager page (since some/most "issues" are inside the `header.tpl`), but i believe each Smarty template should be checked (well, especially those having variables in)

exside:
it's strange, I set the log_level in the manager to 4 (e.g. show all / DEBUG) but I don't get anything =(, did you set the log level somewhere else?

opengeek:
FWIW @exside this depends on your error_reporting level in PHP as well. If PHP is not configured to show E_NOTICE level errors (typical production PHP configuration), the log_level will make no difference.

Jako:
An 'Update Extras with a CMP to Revolution 2.5' should be prepared in RTFM. I notice that WARN messages in MIGX and Formz at the moment.

exside:
@opengeek: I put 

```
error_reporting = E_ALL & ~E_NOTICE | E_STRICT
```
 in my php ini, but didn't get the notices...

PR #12808, opened 2015-12-06
Update phpmailer to version 5.2.14

exside:
### What does it do ?

Update the phpmailer class to version 5.2.14. Current version is only 5.2.13 so I hope/guess the possibility for breaking stuff is quite small, not sure though if it fixes the warning given by php7cc as mentioned in the issue below. Will check that once I have a "merged" php7 version with all related changes. Couldn't find the removed set_magic_quotes_runtime() method used anymore in the code though...
### Why is it needed ?

Keeping libs up to date and fixing a security issue =). For details see https://github.com/PHPMailer/PHPMailer/releases/tag/v5.2.14
### Related issue(s)/PR(s)

Maybe also https://github.com/modxcms/revolution/issues/12797


Mark-H:
Thanks @exside!

PR #12809, opened 2015-12-07
Make modphpthumb.class.php compatible with PHP 7

exside:
### What does it do ?

Replaces removed functions ereg() and eregi() with preg_replace() equivalents (hopefully, somebody with more PHP skills please verify this =D).
### Why is it needed ?

To make MODX PHP 7 compatible, the media browser is broken without this...
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/issues/12797, https://github.com/modxcms/revolution/pull/12810

PR #12810, opened 2015-12-07
Update phpthumb to version 1.7.14

exside:
### What does it do ?

Updates the phpthumb library to the most current code on github and adds a required constructor (for PHP 7 compatibility) to the main phpthumb.class.php. If this is not present it will fail with E_DEPRECATED in PHP 7...~~that constructor is not present in the original source code~~, a PR against that repo was made as well: https://github.com/JamesHeinrich/phpThumb/issues/45 -> PR was merged
### Why is it needed ?

To make thumbnailing work under PHP 7, mostly for the media browser and other image related areas in the manager.
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/issues/12797, https://github.com/modxcms/revolution/pull/12809

### DIFF
--- a/core/model/phpthumb/index.php
+++ b/core/model/phpthumb/index.php
@@ -1,4 +1,3 @@
 <?php
 header('Location: /');
 exit;
-?>
--- a/core/model/phpthumb/index.php
+++ b/core/model/phpthumb/index.php
@@ -1,3 +1,11 @@
 <?php
-header('Location: /');
-exit;
+if (!file_exists('phpThumb.config.php')) {
+	if (file_exists('phpThumb.config.php.default')) {
+		echo 'WARNING! "phpThumb.config.php.default" MUST be renamed to "phpThumb.config.php"';
+	} else {
+		echo 'WARNING! "phpThumb.config.php" should exist but does not';
+	}
+	exit;
+}
+header('Location: ./demo/');
+?>
--- a/core/model/phpthumb/index.php
+++ b/core/model/phpthumb/index.php
@@ -1,11 +1,4 @@
 <?php
-if (!file_exists('phpThumb.config.php')) {
-	if (file_exists('phpThumb.config.php.default')) {
-		echo 'WARNING! "phpThumb.config.php.default" MUST be renamed to "phpThumb.config.php"';
-	} else {
-		echo 'WARNING! "phpThumb.config.php" should exist but does not';
-	}
-	exit;
-}
-header('Location: ./demo/');
+header('Location: /');
+exit;
 ?>
--- a/core/model/phpthumb/phpthumb.class.php
+++ b/core/model/phpthumb/phpthumb.class.php
@@ -12,7 +12,7 @@
 ob_start();
 if (!include_once(dirname(__FILE__).'/phpthumb.functions.php')) {
 	ob_end_flush();
-	die('failed to include_once("'.realpath(dirname(__FILE__).'/phpthumb.functions.php').'")');
+	die('failed to include_once("'.dirname(__FILE__).'/phpthumb.functions.php")');
 }
 ob_end_clean();
 
@@ -154,6 +154,7 @@ class phpthumb {
 	var $config_disable_pathinfo_parsing             = false;
 	var $config_disable_imagecopyresampled           = false;
 	var $config_disable_onlycreateable_passthru      = false;
+	var $config_disable_realpath                     = false;
 
 	var $config_http_user_agent                      = 'Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.7.12) Gecko/20050915 Firefox/1.0.7';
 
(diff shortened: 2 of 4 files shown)


exside:
@Mark-H what about these two modifications: https://github.com/modxcms/revolution/commit/1d17bb45ef7ee8c297c12600aad074ae9770ad55, https://github.com/modxcms/revolution/commit/38eb388c718af8179eb15b42ccc14093fb812867

should I add them in again?

Mark-H:
Probably good to make sure the weird character in https://github.com/modxcms/revolution/commit/38eb388c718af8179eb15b42ccc14093fb812867 hasn't made its way back, but https://github.com/modxcms/revolution/commit/1d17bb45ef7ee8c297c12600aad074ae9770ad55 was needed after a merge conflict duplicated the methods, so that shouldn't be the case here.

exside:
give me a little time here before merging this, the guy over at the phpThumb repo is quite fast with merging (the constructor mod is already in the core, so it's not a "mod" anymore =D)...also proposed to replace that ° character with the word "degrees" so we don't have to keep this character in mind each time when updating...if that get's merged as well I'll update the phpthumb.class.php once more and then we're in proper sync with the original repo!

exside:
And just to mention it, while debugging yesterday I again came across something I really dislike about the media manager... this:

```
http://url.tomodx.tld/connectors/system/phpthumb.php?src=assets%2Fsite%2Fmedia%2Fmodx-treeclickarea.gif&w=100&h=80&f=png&q=90&far=1&HTTP_MODAUTH=<yeah.complete.modauth.hash.via.GET>=mgr&source=1
```
as of 2015, is it really a) necessary and b) a good idea to pass a session auth hash via GET params? Or is it just looking scary to me as a non-server-savy guy and not actually a problem? Just asking...
Besides that it performs very poorly in folders with a lot of images it could also be a potential security hole for somebody listening on a non-SSL manager somehow...no?

exside:
Done, also https://github.com/JamesHeinrich/phpThumb/pull/47 was merged, so the source repo is now the same as the code in our repo, which is a good thing =)...that PR fixed that ° character @Mark-H

Mark-H:
Hmm this update seems to be breaking thumbnails in the media manager for me on my standard dev environment running PHP 5.5. Could you check again to make sure if this is working as expected for you?

exside:
Will have a look!

exside:
@Mark-H I cannot reproduce the breaking thumbnails under php 5.5.30.

- Manually deleted all folders inside core/cache/ (standard git install)
- Tried the dedicated Media Browser CMP, seems to generate the thumbs, when clicking the same folder again they get loaded from the cache (although not visible from the frontend...)
- Same thing for an image TV (cleared the cache folder again...)

could you inspect one of the broken thumbnails and paste the URL in a separate window and see if some error pops up, if not go to core/connectors/system/ and see if there's any error_log file?

exside:
rebased on current 2.x, but still cannot reproduce it =/

exside:
and somehow now PHP 5.3 fails in Travis...any idea where that is coming from suddenly?

Mark-H:
Hmm will need to do some digging on my end then why it wasn't working for me then. 

The 5.3 failure looks safe to ignore; it failed on a dashboard test that takes too long... it complains about that every now and then.

exside:
Did you clear the cache folder before testing it (probably...)?

exside:
rebased on current 2.x, still working on my side...anybody else can confirm @Mark-H 's issue (e.g. broken thumbnails in the media browser)?

Mark-H:
Still getting the errors here when checking out this patch.. 

```` 
[2016-01-13 21:25:25] (ERROR @ /release-2.5/connectors/system/phpthumb.php) phpThumb was unable to generate a thumbnail for: /Applications/MAMP/htdocs/release-2.5/core/cache/phpthumb/localhost__src9bb932e960805a15be1331f6f6bee9f3_par02a1b054e8467fb1e234c8a8fd8d10d2_dat1449531431.png
[2016-01-13 21:25:25] (ERROR @ /release-2.5/connectors/system/phpthumb.php) Error outputting thumbnail:
OutputThumbnail() failed because !is_resource($this->gdimg_output) in file "phpthumb.class.php" on line 606
````

Reverting back brings it back so there's something about this that's breaking the thumbnails on my machine.

I'm not sure what yet, but perhaps it's related to running MODX in a subfolder? I've seen that cause issues before, though that's not been the case in a while..

exside:
the test environment is actually also in a subfolder (you mean deeper nested than "public_html"?)...but definitely have to make sure that this doesn't break thumbnailing...

Jako:
This fix/update should be tested and included as fast as possible, because there is a deprecated issue in the phpthumb class.

Mark-H:
Looks like the problem on my dev environment has pretty much gone away now.. not sure why. Looks like there's updates available from phpthumb again though, maybe you can update the pull request so we get the very latest?

Jako:
Closed in favor of #13039 (newer phpThumb version)

PR #12811, opened 2015-12-07
[3.0.0-dev]Fix bug in _build/build.properties.sample.php

Codenator81:
### What does it do ?

Add 

``` php
use xPDO\xPDO;
```

to _build/build.properties.sample.php
### Why is it needed ?

This solving error when build modx package
I use php 5.6  
### Related issue(s)/PR(s)

php 5.5 issue we dealt with in #12804

### DIFF
--- a/_build/build.properties.sample.php
+++ b/_build/build.properties.sample.php
@@ -1,4 +1,5 @@
 <?php
+use xPDO\xPDO;
 /* define some properties */
 $properties['cache_path'] = MODX_CORE_PATH . '/' . (MODX_CONFIG_KEY === 'config' ? '' : MODX_CONFIG_KEY . '/') . 'cache/';
 


Mark-H:
The failed build seems to be because of the php 5.5 issue we dealt with in #12804, which hasn't been merged into 3.x yet, so that's not a real problem.

@Codenator81 could you please update the pull request with the template from https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md#template-1, and sign the CLA at modx.com/cla if you haven't yet? If you did sign it, please let me know what username that would be under as I can't currently locate it.

Codenator81:
@Mark-H My nickname on modx ```codenator```. And I update PR as described in template you provided

Mark-H:
Thanks @Codenator81! Looks like nothing is standing in the way of this getting merged :)

PR #12813, opened 2015-12-07
Fixed fatal error re-definition function mkdirs

justenj:
This is error report, when run this processor more than once.

### DIFF
--- a/core/model/modx/processors/system/filesys/folder/create.php
+++ b/core/model/modx/processors/system/filesys/folder/create.php
@@ -3,6 +3,17 @@
  * @package modx
  * @subpackage processors.system.filesys.folder
  */
+ 
+// recursive mkdir function
+if (!function_exists('mkdirs')) {
+        function mkdirs($strPath, $mode){
+        	if (is_dir($strPath)) return true;
+        	$pStrPath = dirname($strPath);
+        	if (!mkdirs($pStrPath, $mode)) return false;
+        	return @mkdir($strPath);
+        }
+}
+
 if (!$modx->hasPermission('file_manager')) return $modx->error->failure($modx->lexicon('permission_denied'));
 
 // prevent subdir hack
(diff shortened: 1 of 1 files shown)


Mark-H:
Hi @Redcore7, thanks for contributing to MODX!

Could you please update your pull request to follow the template in https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md#template-1? 

Also, have you ever signed the MODX CLA? That's needed for us to accept your pull request, but I can't seem to find one in your name. Please visit modx.com/cla and submit the form, and let me know when you did so this fix can be included in the next MODX release. 

Thanks!

Mark-H:
Hi @Redcore7 - just checking to see if you saw my earlier message? Thanks!

exside:
@Mark-H guess if we fix this, maybe I could also rewrite it to a class based one...what about using the modCacheHandler or modFileHandler to create the directory instead of recreating that functionality with mkdir etc. inside the processor? Opinions?

exside:
@Redcore7 my question is more about how you actually get that error, when creating a directory where (what manager functionality)? The media browser actually uses the processor at /processors/browser/directory/create.class.php as far as I know...where else can we create directories?

Mark-H:
Sounds like a good idea @exside. 

As for how to trigger it, I suspect they're using $modx->runProcessor to call the processor multiple times.

exside:
mmh, good idea =)...they could also just use the one from processors/browser/ no?? Is there any way to "route" such legacy processors to another one without rewriting them entirely?

Mark-H:
@Redcore7 Could you please take a look at my reply at https://github.com/modxcms/revolution/pull/12813#issuecomment-162704037 - I'd like to merge this into 2.5.0-rc1 but at the very least you need to have the CLA signed.

theboxer:
Closing due to lack of response. Fix is comited in 55c50ef84031d0892ab936357b96c291e23bfeda.

PR #12814, opened 2015-12-08
Added unpub_date to publishingResults

christianseel:
### What does it do ?

Extend $publishingResults with `unpub_date` which is returned during the autoPublish event.
### Why is it needed ?

`unpub_date` should be handy for unpublished resources…
### Related issue(s)/PR(s)

As suggested by @Mark-H here: https://github.com/modxcms/revolution/pull/12747#issuecomment-161269192


Mark-H:
Thanks @christianseel :)

PR #12820, opened 2015-12-09
Fix PHP Undefined index warnings for ext_debug and a due to smarty update

exside:
### What does it do ?

Fixed variables in manager smarty templates that throw Undefined index PHP warnings. Fixed till now:
- [x] a
- [x] ext_debug
- [x] match smarty's display() method signature in modsmarty.class.php
### Why is it needed ?

Because smarty was updated to 3.1.27 as of https://github.com/modxcms/revolution/pull/12807 and seems to be more strict now when checking for unset variables. The Problem is, that I'm not getting these warnings, even when my log_level is set to 3...so I'm kind of blind in what else there could be that would throw such warning. The two above were reported by @rtripault 
### Related issue(s)/PR(s)

See the one above.

### DIFF
--- a/core/model/modx/smarty/modsmarty.class.php
+++ b/core/model/modx/smarty/modsmarty.class.php
@@ -130,7 +130,7 @@ class modSmarty extends Smarty {
      * @param mixed $compile_id compile id to be used with this template
      * @param object $parent next higher level of Smarty variables
      */
-    public function display($template, $cache_id = null, $compile_id = null, $parent = null) {
+    public function display($template = null, $cache_id = null, $compile_id = null, $parent = null) {
         echo $this->fetch($template, $cache_id, $compile_id, $parent);
     }
 }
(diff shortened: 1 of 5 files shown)


exside:
@rtripault Also fixed the method signature

opengeek:
Are there specific pre-conditions for these notices? In PHP 7.0.2? I'm trying to reproduce but am not sure how to get them to show up.

opengeek:
Actually, I got it showing up now... but still getting the following after pulling in the changes...

```[2016-01-11 16:40:29] (WARN @ /Users/opengeek/www/revo-2.x/core/cache/mgr/smarty/default/7ef189ec8673b41b3c2d6f5c13a2d03bc57053df_0.file.header.tpl.php : 94) PHP notice: Undefined index: a
[2016-01-11 16:40:29] (WARN @ /Users/opengeek/www/revo-2.x/core/cache/mgr/smarty/default/7ef189ec8673b41b3c2d6f5c13a2d03bc57053df_0.file.header.tpl.php : 99) PHP notice: Undefined index: a
 ```

Issue #12791, opened 2015-12-01
Stop Messing with the Browser's Default Focus outline


jpdevries:
### Summary
All browsers have a default value they apply to the currently focused item. ExtJS aggressively sets outlines on focused items to none. This is bad for a11y.
http://tjvantoll.com/2013/01/28/stop-messing-with-the-browsers-default-focus-outline/

![](http://j4p.us/image/142c000T0O19/Screen%20Shot%202015-12-01%20at%206.16.19%20AM.png)

### Step to reproduce
Try tabbing through the 2.4 Login screen. Notice when the language select receives focus there is no indication it has done so.

### Observed behavior
Users can't tell if inputs have focus.

### Expected behavior
Users should be able to tell if inputs have focus.

### Environment
MODX 2.4.x default theme

PR #12824, opened 2015-12-11
Stop Messing with the Browser's Default Focus outline

jpdevries:
### What does it do ?

All browsers have a default value they apply to the currently focused item. ExtJS aggressively sets outlines on focused items to none. This is bad for a11y. This overrides ExtJS' dirty hack of setting all focused items to `outline:none`;
### Why is it needed ?

So users can tell if elements have focus.
http://tjvantoll.com/2013/01/28/stop-messing-with-the-browsers-default-focus-outline/
### Related issue(s)/PR(s)

Closes https://github.com/modxcms/revolution/issues/12791
### Note

I figured I'd start putting a11y improvements in a separate .scss file so it makes it easier to port to things like the a11y theme project.

### DIFF
--- a/_build/templates/default/sass/_a11y.scss
+++ b/_build/templates/default/sass/_a11y.scss
@@ -1,0 +1,3 @@
+body.ext-webkit *:focus{
+    outline: auto !important;
+}
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -21,6 +21,7 @@ $fa-css-prefix:       icon;
 @import "windows";
 @import "dashboard";
 @import "help";
+@import "a11y";
 
 body {
   color: $mainText;


Mark-H:
Merged, keep an eye out for 2.5.0-rc1

Mark-H:
I can't seem to reopen this pull request, but I've reverted the merge from earlier today as this is causing issues with double focus styles popping up in random places in the manager. ExtJS is already handling focus styles on quite a lot of things across the manager.  

This will likely need to be redone in a way that the CSS is more specific to aspects that lack a focus style, like the language selector on the login page, rather than just adding a focus to everything.

jpdevries:
@Mark-H I think double focus styles for some users is better than no focus styles for some users. The former is perhaps unsightly, the later is critical.

PR #12833, opened 2015-12-22
Fire event on one resource cache update

faitno:
This event need for speed up extension "statcache" http://modx.com/extras/package/statcache from opengeek, that delete only one file from static cache, and provides freshness it


Mark-H:
It looks like a commit from #13163 made it into this pull request.

The `OnResourceCacheUpdate` event also isn't added in _build/data/transport.core.events.php. Other than that I can see this as a useful event for certain cases.

sdrenth:
@soulcreate Are you planning on working on this PR and fixing the feedback provided by Mark?

OptimusCrime:
The origin repository is unknown, so it has to be recreated. 

I could do it, it adding it to the transport core file is the only other thing that needs to be done?

Mark-H:
Closing in favour of #13590.

PR #12841, opened 2016-01-07
- fix htmlentities problem

vgrish:
htmlentities problem in extended field user panel:
example

```
<a href="http://narod.ru/disk/65201298001.3bb5c513dd4781ed3ffa0255a1a3d533/book-0002.zip.html">Скачать книгу "Канавцев Г.И. Предупреждение пожаров на нефтяных промыслах, 1940 год" с Яндекс.Народ (PDF, 24.5 Мб)</a>
```

### DIFF
--- a/manager/controllers/default/security/user/update.class.php
+++ b/manager/controllers/default/security/user/update.class.php
@@ -141,7 +141,7 @@ Ext.onReady(function() {
                     $v = substr($v,0,30).'...';
                 }
                 $field['iconCls'] = 'icon-terminal';
-                $field['text'] = $key.' - <i>'.$v.'</i>';
+                $field['text'] = $key.' - <i>'.htmlentities($v,ENT_QUOTES,$encoding).'</i>';
                 $field['leaf'] = true;
                 $field['value'] = $value;
             }


Mark-H:
Hi @vgrish, thanks for your pull request. Could you please sign the CLA at http://modx.com/cla? We need that in order to be able of merging your contribution.

Could you also elaborate a bit on what this fixes, how to reproduce the problem and mention any possibly related other issues? We have a template in the contributing docs at https://github.com/modxcms/revolution/blob/2.x/CONTRIBUTING.md that helps making sure we have all the information needed to review your pull request, so if you could follow that it would make it a lot easier for us!

Jako:
The issue should be the following: At the moment you could create a setting with an HTML tag in the extended attributes value. After saving the user it can't be edited again. The javascript console is showing 'Uncaught TypeError: Cannot read property 'stopFx' of null'. This is (probably) fixed with the patch. But the patch should be better done with htmlspecialchars.

Now the Issue is somehow deeper: At the moment you could create a extended setting i.e. with quotes in the name. Those settings can't be edited/deleted after. The chars should be restricted for the name or htmlspecialchars should be used there too.

Mark-H:
@Jako Thanks for the clarification. Could reproduce with that information and this fix indeed resolved it. 

@vgrish Thanks for the contribution and signing the CLA. :)

Merged into 2.5.x for inclusion in 2.5.1 4d51bdd991057cc9f7d4051a8e300226c31b6e5d

Issue #12846, opened 2016-01-12
[2.5] E_NOTICE errors on setup 


Mark-H:
### Summary
First step in the setup throws 2 E_NOTICE errors in setup/templates/language.tpl. One for undefined index "restarted" and one for trying to get a property of a non-object. Installation does work as expected beyond that. 

### Step to reproduce
Open setup with display_errors on or logging notices somewhere. 

### Observed behavior
<img width="729" alt="screen shot 2016-01-12 at 13 13 05" src="https://cloud.githubusercontent.com/assets/312944/12263343/8a153ad4-b92e-11e5-816a-f9d1e52163fe.png">

### Expected behavior
No errors/notices

### Environment
Code from 2.x branch as of right now. Local MAMP Pro set up, happens on both PHP 5.4.39 and 5.6.7, likely on other versions too.

PR #12847, opened 2016-01-12
Ensure $restarted in templates/language.tpl always exists

Mark-H:
### What does it do ?

Makes sure the $restarted variable in templates/language.tpl always exists. If it wasn't set, it would trigger two E_NOTICE errors. 
### Why is it needed ?

Make sure the setup doesn't throw any E_NOTICE errors
### Related issue(s)/PR(s)
#12846

### DIFF
--- a/setup/controllers/language.php
+++ b/setup/controllers/language.php
@@ -37,5 +37,8 @@ $parser->set('languages',$languages);
 if (!empty($_REQUEST['restarted'])) {
     $parser->set('restarted',true);
 }
+else {
+    $parser->set('restarted', false);
+}
 
 return $parser->render('language.tpl');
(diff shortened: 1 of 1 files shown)


setup/controllers/language.php
```diff
@@ -37,5 +37,8 @@
 if (!empty($_REQUEST['restarted'])) {
     $parser->set('restarted',true);
 }
+else {
```
Alroniks: Maybe to simplify a little?
```
$parser->set('restarted', !empty($_REQUEST['restarted']));
```

theboxer:
@Mark-H will you please update the PR with the suggestion?

Mark-H:
@theboxer Done

PR #12778, opened 2015-11-22
rm -rf manager/min

jpdevries:
### What does it do ?

Removes the manager/min directory
### Why is it needed ?

We can make breaking changes and don't need it anymore
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/pull/12611

Issue #12610, opened 2015-08-27
Manager Breaks when loading off domain assets with API and compress_js


jpdevries:
- Install MODX
 - Enable `compress_js` and `compress_css`
 - Create a simple plugin that uses something like loads an off-domain asset
 - notice the minifier will try and fopen that file and break the whole manager :+1: 

#### Example
`$modx->controller->addJavascript('https://cdnjs.cloudflare.com/ajax/libs/ace/1.1.9/ace.js');`
or
`$modx->regClientStartupScript('https://cdnjs.cloudflare.com/ajax/libs/ace/1.1.9/ace.js')`

![](https://slack-files.com/files-tmb/T025K7SUQ-F09KLHXV4-0bbdcedb2c/pasted_image_at_2015_08_25_02_53_pm_1024.png)

Mark-H:
Closing as #12611 made js compression safer, and should only be disabled for development.

PR #12857, opened 2016-01-16
Removing Disable Compress Checkbox from Setup Installer

jpdevries:
### What does it do ?

Removes the "Disable CSS/JS compression" checkbox from the Setup screen. It does not remove the associated lexicons, let me know if it should.

### Why is it needed ?

Assuming #12778 is merged, MODX 3.x will no longer have a min API. So, the setting to disable it is no longer needed.
### Related issue(s)/PR(s)

See modxcms/revolution#12778
See modxcms/revolution#12538
See modxcms/revolution#12610


Mark-H:
Thanks!

PR #12860, opened 2016-01-18
Added new output filter 'stripmodxtags'

Jako:
### What does it do ?

The output filter strips (recursive) all MODX tags in a string. Credits go to Josh Curtis (https://github.com/jcdm).
### Why is it needed ?

The default stripTags output filter does not clear MODX tags and there is no possibility to filter that tags away i.e. in a meta description with the following call for the content field.

```
<meta name="description" content="[[*description:default=`[[*content]]`]]">
```

The default regex in MODX does not catch nested tags. The proposed one does this. See: https://regex101.com/r/lW5gX6/1

```
<meta name="description" content="[[*description:default=`[[* content:stripmodxtags]]`]]">
```

strips all MODX tags before they are parsed.

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -271,7 +271,7 @@ class modOutputFilter {
                             }
                             break;
                         case 'stripmodxtags':
-                            $output = $modx->stripTags($input, '', array('/(\[\[([^\[\]]++|(?R))*?\]\])/sg', 1));
+                            $output = preg_replace("/\\[\\[([^\\[\\]]++|(?R))*?\\]\\]/s", '', $input);
                             break;
                         case 'length':
                         case 'len':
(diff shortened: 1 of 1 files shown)


jcdm:
With thanks to andufo on stackoverflow for pointing me in the right direction: http://stackoverflow.com/questions/2359921/need-help-with-regex-to-remove-square-brackets-and-anything-between

Jako:
Have to do one test before merge … maybe the whole thing is not useful.

Jako:
Test worked fine with 
```
[[*description:default=`[[*content:stripmodxtags]]`]]
```

OptimusCrime:
@Jako: I don't 100% get this, what does `[[*content]]` now output?

Jako:
All MODX tags are stripped away with that modifier. So `[[*content:stripmodxtags]]` would remove all MODX tags inside before they are parsed.

jcdm:
No change to content. This is an output modifier that can be used on any field in the form [[something:stripmodxtags]]. This will output 'something' with all double bracketed tags and their contents removed.

The example use case just happened to be using it on a [[*content]] tag.

OptimusCrime:
Yeah, that was what I ment. This tag is awesome for debugging then. Huge +1 from me.

Jako:
Thanks for the 'use for debug' idea :+1:

PR #12862, opened 2016-01-18
Fix system settings pagination issue if default_per_page is > 30

### LINKED ISSUE #12832: System Setting Paging issue
MODXCloud Hosted, revolution version 2.4.2-pl

When viewing system settings and clicking the next button, pagination doesn't work. I get redirected to page 101 of 5, instead of page 2 of 5.

screen shot 2015-12-22 at 11 56 40

Mark-H:
### What does it do ?

Cast the default_per_page setting to integer before using it in the system settings grid.
### Why is it needed ?

The default_per_page config option comes in as a string, and because it isn't cast/parsed to an integer properly, the pagination goes crazy, resulting in page "101" instead of "2".
### Related issue(s)/PR(s)

Fixes #12832


OptimusCrime:
Looks good!

Issue #12821, opened 2015-12-09
HHVM errors


pixelive:
### Summary
Having the following errors with MODX 2.4.2-pl on a HHVM machine.

```
Deprecated: ereg: As of PHP 5.3.0, preg_match is suggested instead in /var/www/vhosts/vv-domain.com/production/core/model/phpthumb/modphpthumb.class.php on line 160

Warning: Session callback expects true/false return value in /var/www/vhosts/domain.com/production/core/model/modx/modx.class.php on line 2441
```

Also noticed that the multi select tv fails on HHVM, for the time being im running the manager in PHP-FPM.

Thanks in advance.

exside:
The first one would be fixed in https://github.com/modxcms/revolution/pull/12809 (I believe, would be nice if you can test)

pixelive:
Alright gonna test as soon as possible!

Thanks!

exside:
and for the second one I'm lost...seems to be something in that block of code

```php
                $sh= false;
                if ($sessionHandlerClass = $this->getOption('session_handler_class', $options)) {
                    if ($shClass= $this->loadClass($sessionHandlerClass, '', false, true)) {
                        if ($sh= new $shClass($this)) {
                            session_set_save_handler(
                                array (& $sh, 'open'),
                                array (& $sh, 'close'),
                                array (& $sh, 'read'),
                                array (& $sh, 'write'),
                                array (& $sh, 'destroy'),
                                array (& $sh, 'gc')
                            );
                        }
                    }
                }
                if (!$sh) {
                    $sessionSavePath = $this->getOption('session_save_path', $options);
                    if ($sessionSavePath && is_writable($sessionSavePath)) {
                        session_save_path($sessionSavePath);
                    }
                }
```

pixelive:
Will try to take a look at it on Sunday!

PR #12865, opened 2016-01-19
Update modx.class.php

johanmolen:
### What does it do ?

Fixed the modx.class.php because it threw the following error:

Warning: Session callback expects true/false return value in /var/www/vhosts/domain.com/production/core/model/modx/modx.class.php on line 2441
### Why is it needed ?

Because it now triggers an warning.
### Related issue(s)/PR(s)

This should fix the following issue partially. 

https://github.com/modxcms/revolution/issues/12821


pixelive:
Replaced with:

https://github.com/modxcms/revolution/pull/12868/files

PR #12869, opened 2016-01-19
manager/cache.manifest.php

lokamaya:
### What does it do ?

HTML5 application cache. Returning 404 not found will clear all HTML5 aplication cache
### Why is it needed ?

**Deprecated** : HTML5 Application Cache has been removed from the Web standards. Manager page break at any time, when the first user logout and the second user login. The cache reside in the browser and make manager page vulnerable.


lokamaya:
Deprecated
This feature has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Do not use it in old or new projects. Pages or Web apps using it may break at any time.

Using the application caching feature described here is at this point highly discouraged; it’s in the process of being removed from the Web platform. Use Service Workers instead. In fact as of Firefox 44, when AppCache is used to provide offline support for a page a warning message is now displayed in the console advising developers to use Service workers instead

https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache

Omeryl:
Ack! I hate when stuff like that happens, after I spent so long getting personal project of mine to use the appcache stuff.. On a side not, wouldn't it make sense to combine this with a PR that does implement Service Workers though? It might be deprecated, but it's better than nothing.

PR #12872, opened 2016-01-21
FR: Patch template typeahead

goldsky:
Adding typeAhead for template on Resource edit mode.
Useful when web has countless template

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -543,6 +543,8 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,anchor: '100%'
             ,editable: true
             ,typeAhead: true
+            ,typeAheadDelay: 300
+            ,forceSelection: true
             ,baseParams: {
                 action: 'element/template/getList'
                 ,combo: '1'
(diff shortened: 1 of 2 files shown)


manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -541,7 +541,10 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,name: 'template'
             ,id: 'modx-resource-template'
             ,anchor: '100%'
-            ,editable: false
+            ,editable: true
+            ,typeAhead: true
```
theboxer: I suggest disabling `typeAhead` function which preselects the first item. When you start typing `template` and wait for suggestions, it automatically fills `BaseTemplate` and you can't continue writing

goldsky: It fills the first available template (on your case: BaseTemplate) if you type non-existed template.
If you have "**Basic**" template and you start to type "**b**", it will select the first "**b**" template when you stop typing for 250ms ([default](http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.form.ComboBox-cfg-typeAheadDelay)), but you **can** keep continue typing, the blue highlight is just **suggestion**
typeahead template

theboxer: You can continue typing, only if the first matching template is matching from start. 
Add `TemplateTest` to your templates. Then start typing `Templ` and wait for suggestions to pop up, it will preselect `BaseTemplate` and if you'll try to continue typing, you'll get nonsense text.

goldsky:
updated

Issue #12835, opened 2015-12-24
Partial caching fails when a cached element is between uncached elements


yoleg:
Normally, whenever an uncached element is called inside of a cached element, only the unprocessed tag of the uncached element gets stored in the cache: `'[[$sayHello]]' => 'Hello, [[!getUserName]]'`
 
Recently (not sure when this changed, but using MODX Revolution 2.4.2 now), the processed result of uncached tags gets stored in the cache if there is one or more cached elements between two uncached elements: `'[[$sayHello]]' => 'Hello, Jay!'`
 
* Properly cached:
    
    ```
    Cached   --> Cached   -> Cached   --> Uncached
    Cached   --> Cached   -> Uncached --> Uncached
    Cached   --> Uncached -> Uncached --> Uncached
    Uncached --> Uncached -> Uncached --> Uncached
    ```
    
* Improperly cached (uncached tag results get stored in the cache):
    
    ```
    Uncached --> Cached   -> Uncached --> Uncached
    Uncached --> Uncached -> Cached   --> Uncached 
    Uncached --> Cached   -> Cached   --> Uncached
    Cached   --> Uncached -> Cached   --> Uncached
    ```
    
The results are the same whether snippets using `$modx->getChunk` or chunks are used as the initializing element.
 
**Why this is critical**:

* Many uncachable snippets rely on `$modx->getChunk` for templating. It is unreasonable to assume that those template chunks will not contain other chunks called in the standard, cached fashion, that in turn contain uncached snippets or placeholders.

**To reproduce:**

1. Create a resource with an empty template. In the content, put:
    
    ```
    <div>Snippet start:</div>
    <div class="level1">[[getChunk? &tpl=`test_chunk` &type=`<span>Cached  </span>` &level=`1` &instance=`1`]]</div>
    <div class="level1">[[!getChunk? &tpl=`test_chunk` &type=`<span>Uncached</span>` &level=`1` &instance=`2`]]</div>
    <div> </div>
    <div>Chunk start:</div>
    <div class="level1">[[$test_chunk? &type=`<span>Cached  </span>` &level=`1` &instance=`3`]]</div>
    <div class="level1">[[!$test_chunk? &type=`<span>Uncached</span>` &level=`1` &instance=`4`]]</div>
    ```
    
2. `getChunk` snippet content:
    
    ```
    $tpl = $modx->getOption('tpl', $scriptProperties, 'test_chunk');
    return $modx->getChunk($tpl, $scriptProperties);
    ```
[...]

PR #12874, opened 2016-01-21
Proposed fix to bug #12835

yoleg:
This fixes a bug that can expose user-submitted form data and other user-specific or request-specific data to the public, particularly in sites that use many nesting levels of elements. See https://github.com/modxcms/revolution/issues/12835

Because this fix changes the tag-parsing order, this might cause some sites with
complex tag nesting to experience bugs as a result of even more complex tag nesting.
This is probably not a result of this fix but is a result of the different tag parsing
order. The parsing order changes because uncached tags vulnerable to
the bug will now remain unparsed until the next parsing pass. Thus, a system
setting is needed to allow sites to disable this fix temporarily on production sites.

Also, parsing speed may decrease. Again, this is not a result of this fix but a result of the unparsed tags that are vulnerable to this bug getting their parsing delayed until the next pass.

Sites that do not have any vulnerable unparsed elements should remain unaffected by this fix, but testing is recommended.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -450,6 +450,23 @@ class modParser {
         if ($cacheable && $token !== '+') {
             $elementOutput= $this->loadFromCache($outerTag);
         }
+        $_restoreProcessingUncacheable = $this->_processingUncacheable;
+        // temporarily disable processing uncacheable elements so they don't get put into this cached element's cache
+        // caching an uncacheable tag could lead to serious problems like caching user-submitted data and exposing it to other users
+        // See bug https://github.com/modxcms/revolution/issues/12835
+        // New system setting:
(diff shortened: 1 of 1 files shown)


opengeek:
@yoleg this was submitted against the `master` branch which is now obsolete — can you rebase your work on the `2.4.x` branch and force update this pull request so we can get this into the next patch release, assuming we don't find any regression problems?

yoleg:
OK, done: https://github.com/modxcms/revolution/pull/12884

yoleg:
I do not know how to change the branch for this pull request, so I just created a new pull request.

opengeek:
In the future if you just keep the same branch and use the `--force` option when you push, it will update the pull request automatically, rewriting the history for the rebase.

opengeek:
Oh right, you cannot change the target of the pull request, I forget these things...

PR #12878, opened 2016-01-24
Remove legacy files during upgrades

### LINKED ISSUE #12737: Legacy files not removed during setup
In MODX 2.3 and 2.4, most processors have been converted to be class based. (see #11437).

If a site began on an earlier version of MODX *(e.g. v2.2)*, it seems the original processors still remain on the FS.

Should there be some kind of process during setup put in place for the MODX instances not installed from Git to perform this cleanup?

**This is just a sample of some of the old processors that still exist:**

legacy_processors

pixelchutes:
### What does it do ?

Adds a MODX 2.5 cleanup upgrade script to remove legacy files identified.

Since I was not able to get a working version of MODX 2.0.0 installed on my system (php 5.5+), the associated list of legacy files/folders was derived for MODX 2.1+:
- Installing MODX `2.1.5-pl`
- Upgrading to `2.2.15-pl`
- Upgrading to `2.3.6-pl`
- Upgrading to `2.4.2-pl`
- Upgrading to `2.5.0-dev`

The entire instance was then cross-referenced with Git HEAD of `2.x` repo, and the difference of `git status -s` was captured.
### Why is it needed ?

Download distributions _(e.g. MODX instances not driven by [Git installation](https://rtfm.modx.com/revolution/2.x/getting-started/installation/git-installation))_ can build up legacy files that have been removed upstream, but no formal cleanup mechanism existed to remove these old files as the site instance is upgraded over time. From a general maintenance and security standpoint, this is not desirable behavior.

This provides a way to define core files/folders that should no longer exist on a system, allowing them to be automatically cleaned up.
### Related issue(s)/PR(s)

Fixes #12737


pixelchutes:
For those interested, attached is the raw file I prepared when making this PR: [cleanup.txt](https://github.com/modxcms/revolution/files/102573/cleanup.txt)

PR #12884, opened 2016-01-27
Proposed fix to bug #12835 (2.4.x branch)

yoleg:
Rebase for https://github.com/modxcms/revolution/pull/12874

This fixes a bug that can expose user-submitted form data and other user-specific or request-specific data to the public, particularly in sites that use many nesting levels of elements. See https://github.com/modxcms/revolution/issues/12835

Because this fix changes the tag-parsing order, this might cause some sites with
complex tag nesting to experience bugs as a result of even more complex tag nesting.
This is probably not a result of this fix but is a result of the different tag parsing
order. The parsing order changes because uncached tags vulnerable to
the bug will now remain unparsed until the next parsing pass. Thus, a system
setting is needed to allow sites to disable this fix temporarily on production sites.

Also, parsing speed may decrease. Again, this is not a result of this fix but a result of the unparsed tags that are vulnerable to this bug getting their parsing delayed until the next pass.

Sites that do not have any vulnerable unparsed elements should remain unaffected by this fix, but testing is recommended.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -450,6 +450,23 @@ class modParser {
         if ($cacheable && $token !== '+') {
             $elementOutput= $this->loadFromCache($outerTag);
         }
+        $_restoreProcessingUncacheable = $this->_processingUncacheable;
+        // temporarily disable processing uncacheable elements so they don't get put into this cached element's cache
+        // caching an uncacheable tag could lead to serious problems like caching user-submitted data and exposing it to other users
+        // See bug https://github.com/modxcms/revolution/issues/12835
+        // New system setting:
+        //   Because this fix changes the tag-parsing order, this might cause some sites with
+        //   complex tag nesting to experience bugs as a result of even more complex tag nesting.
+        //   This is probably not a result of this fix but is a result of the different tag parsing
+        //   order. The parsing order changes because uncached tags vulnerable to
+        //   the bug will now remain unparsed until the next parsing pass. Thus, a system
+        //   setting is needed to allow sites to disable this fix temporarily on production sites.
+        // Also, parsing speed may decrease. Again, this is not a result of this fix but
+        // a result of the unparsed tags that are vulnerable to this bug getting their
+        // parsing delayed until the next pass.
+        if ($this->_processingUncacheable && $cacheable && $this->modx->getOption('core.parser_recurse_uncacheable', null, true)) {
+            $this->_processingUncacheable = false;
+        }
         if ($elementOutput === null) {
             switch ($token) {
                 case '+':
@@ -522,6 +539,7 @@ class modParser {
             /* $this->modx->cacheManager->writeFile(MODX_BASE_PATH . 'parser.log', "Processing {$outerTag} as {$innerTag}:\n" . print_r($elementOutput, 1) . "\n\n", 'a'); */
         }
         $this->_processingTag = false;
+        $this->_processingUncacheable = $_restoreProcessingUncacheable;
         return $elementOutput;
     }
 


opengeek:
Thanks for the great catch and contribution @yoleg — I shortened the code comments for now, modified the system setting name slightly and added it into the system settings data and English lexicon so it is available to people in the manager.

yoleg:
Great, thanks!

yoleg:
@opengeek Hey, just a heads up that I found two side-effects from the fix:
 
Side effect 1 is a bug, and I am still looking for on a fix for it. The problem is that unprocessed tags with an output filter that are in turn inside another output filter break the outer output filter because they are not removed. So, this chunk would break (assuming fake and fake2 do not exist in the properties):
```
<h1>Level 1</h1>
[[+fake:notempty=`
  <h2>Level 2 Notempty</h2>
  <p>[[+fake2:notempty=`Level 3A Notempty`:default=`Level 3A Default`]]</p>
`:default=`
  <h2>Level 2 Default</h2>
  <p>[[+fake2:notempty=`Level 3B Notempty`:default=`Level 3B Default`]]</p>
`]]
```

Side effect 2 is more of a configuration issue: in complex sites, the system setting "parser_max_iterations" might need to be increased since there are more parser iterations.

yoleg:
@opengeek I found the problem, which is in `modPlaceholderTag::process()`. When the fix is enabled, this `if` statement evaluates to false for undefined cached placeholders:
```php
            if ($this->_output !== null || $this->modx->parser->isProcessingUncacheable()) {
                $this->filterOutput();
                $this->_processed = true;
            }
```
 
The output filter is skipped, so undefined tags are not removed. Because undefined tags are not removed, an undefined tag with an output filter is allowed to remain inside of another output filter. When that output filter is processed, it is split on ":", but the inner output filter also has ":", and the split creates unparsable code. 
 
For example, this:
```html
fake:notempty=`
  <h2>Level 2 Notempty</h2>
  <p>[[+fake2:notempty=`Level 3A Notempty`:default=`Level 3A Default`]]</p>
`:default=`
  <h2>Level 2 Default</h2>
  <p>[[+fake2:notempty=`Level 3B Notempty`:default=`Level 3B Default`]]</p>
`
```

is processed without the `fake2` placeholders getting removed by `filterOutput()`. Thus, the split on ":" results in:

```html
* \n  <h2>Level 2 Notempty</h2>\n  <p>[[+fake2:notempty=`Level 3A Notempty
* Level 3A Default`]]</p>\n
* \n  <h2>Level 2 Default</h2>\n  <p>[[+fake2:notempty=`Level 3B Notempty*
* Level 3B Default`]]</p>\n
```
 
There are several directions to go in that I can think of to fix this: 

1. Have two separate  `modParser::_processingUncacheable` variables: one to track if we are processing uncacheable elements NOW, and one to check if we have started processing uncacheable elements AT ALL. 
2. Remove undefined placeholders whether or not we are processing uncacheable.
3. Create a regular expression in the `modInputFilter:filter` method that skips nested tags (instead of blindly splitting by ":")
4. Force all nested tags to be processed and removed before splitting in `modInputFilter:filter`. 
5. Some other idea? 
 
Which of these strategies is most compatible with the MODX tag parsing strategy?

Mark-H:
@yoleg This should probably be a new issue to make sure it doesn't fall under the radar, could you create that? 

Option 1 sounds the most compatible to me.

pixelchutes:
@Mark-H @yoleg Perhaps already covered by #12901

yoleg:
@pixelchutes @Mark-H Submitted issue #12906 for clarity. I'm start working on fix #1

Issue #12882, opened 2016-01-26
RFC: Adding optional "X-Powered-By: MODX Revolution" header to MODX sites?


Mark-H:
Following this conversation on twitter (https://twitter.com/brandonkelly/status/691953553494028288):

<img width="338" alt="screen shot 2016-01-26 at 19 25 34" src="https://cloud.githubusercontent.com/assets/312944/12590312/9a7128f4-c462-11e5-89ec-ae3358937759.png">

I'm wondering if perhaps there would be consensus to add an X-Powered-By header to MODX installs. The primary benefit would be a better way to get MODX represented properly in website analysers like W3Techs and BuiltWith.com. The primary disadvantage would be that a site can be identified as built with MODX; if a critical vulnerability is found that could lead to highly targeted attacks*, even if it doesn't include the specific release. 

What I'd like to propose is adding the X-Powered-By header with value "MODX Revolution" or slightly more specifically "MODX Revolution 2.x". I'm not sure if it should be opt-in or opt-out; with opt-out we might get faster adoption of the header and better adoption from w3techs/builtwith, but it might be perceived as pushing it onto people. Opt-in with a checkbox during setup might be better. 

Historically, MODX has always been very clear about not messing with what the site builder puts in. As this is only a header and not anything added to the output, I don't think that has to be a problem. 

Implementation of such a header is pretty simple (here's how it can be done as a plugin: https://gist.github.com/Mark-H/bda283b073b6145cfde1). 

* Fun side note: when I added the X-Powered-By header to some of my own sites, that actually overwrote an existing X-Powered-By header that showed the specific PHP version used. In a way, this makes it more secure, but that is obviously debate-able :P

Mark-H:
Craft has the X-Powered-By header enabled by default, so opt-out (https://craftcms.com/docs/config-settings#sendPoweredByHeader)

isaacniebeling:
I would definitely want that to be opt-in. I don't like advertising what I'm using as much as possible, and MODX is the most easily-camouflaged system I've used to date.

Omeryl:
I find one of the biggest things MODX CMS does right is no branding. Once you change manager location and the first resource, you're unlikely to ever be found out. That's a huge security plus, not a negative.

bezumkin:
+1 for X-Powered-By, enabled by default. But with system option for disable.

Lack of header do not makes your site more secure. MODX can be easy revealed with one HTTP request to /manager/ with default setup. 

[...]

PR #12885, opened 2016-01-28
Implement X-Powered-By header to send "MODX Revolution 2.x" on all requests

Mark-H:
### What does it do ?

Allows MODX to respond to all requests with a "X-Powered-By" header, with the value of `MODX Revolution 2.x` (or `MODX Revolution 3.x` once that is out). It is controlled by a new setting send_poweredby_header. 

The setting is disabled by default. 

On new installations, the checkbox (pictured below) will be _checked by default_, allowing a swift opt-out if they wish not to do that. 

On upgrades, the setting value is not changed (either from the default disabled, or enabled if the user changed the setting). So the site builder is in complete control over wether this is sent or not. 

screen shot 2016-01-28 at 17 09 47
### Why is it needed ?

MODX is considered to be a small contender in the CMS market by people from outside our community, because it rarely shows up in services that track technology usages across the top X websites, like builtwith.com and w3techs.com. As I believe MODX is a lot bigger than those sites seem to indicate, this pull request will allow sites like those to more accurately track when a site is MODX. 
### Related issue(s)/PR(s)

More background and discussion of this change took place in issue #12882.


jpdevries:
Do these checkboxes get ticked to indicate if their current values? For example, if Send X-Powered-By is true in the databse will the box be initially ticked to reflect that?

Mark-H:
The checkbox is only available on a new installation, where no existing value is set.

ckburnett:
I'm not a fan of exposing this information at all (and would really hate for it to ever be anything but opt-in), and I understand "security by obscurity is no security at all" but it is _part of_ defense-in-depth... 

but all of that set aside...

Food for thought - what about offering a bit more granularity of choice? Let the site generate:
1. No x-powered-by header
2. A header with just the code name (i.e. "Revolution") only
3. A header with both the code name and the version number as you have it now

Of course, you could have options for all the choices provided by getVersionData(), but I think adding in 2 above gives a bit more flexibility that could be desirable to some.

rthrash:
This is a great idea. Thanks Mark! :+1:

jaygilmore:
After rethinking this, I wonder if anyone will ever check this box. It seems to me that since we've left the technical term in place that almost everyone will leave it unchecked as they attempt to fill or change only the necessary fields as quickly as possible to get to the finish line. 

I feel we've compromised all the way to no point to include at all. It should be opt out nor negatively phrased. 

`Disable Powered-By Headers` with the default empty state, if for new installs. This is a new feature, and a non-breaking-change.

Mark-H:
To clarify, on new installations, the checkbox controlling the setting ("Send X-Powered-By Header") is checked by default (= opt-out on new installs). 

On updates, there is no checkbox, and the setting is disabled by default (= opt-in on upgrades).

PR #12897, opened 2016-02-08
Extend SmartyBC in modSmarty to ensure backwards compatibility

### LINKED ISSUE #12895: MODX2.5RC1: Unknown method get_template_vars
**Description**
After updating MODX from 2.4.2 to 2.5 RC1, login to the manager is not possible. I get an 500 error.

**Expected behaviour**
- Manager login page is visible and login is possible

**Steps to reproduce**
- Quickstartbuttons have to installed
- Update to MODX 2.5 RC1
- Access the manager login page

**Server log**
`PHP Fatal error:  Uncaught  --> Smarty: Call of unknown method 'get_template_vars'. <-- \n  thrown in /var/www/vhosts/xxx/web/core/model/smarty/sysplugins/smarty_internal_templatebase.php on line 554`

**Short analysis**
Some dashboard widgets (SimpleCart, QuickstartButtons) accessing the method addLangTopic.
This method seems to be removed.

There should be an logic to prevent  the exeption so that accessing to the manager login page is always possible.

Mark-H:
### What does it do?

Restores backwards compatibility with the previously used version of Smarty by using the special `SmartyBC` class. 
### Why is it needed?

When Smarty was updated for 2.5.0-rc1 (to get PHP 7 compatibility), this included some unintentional breaking changes (methods were renamed). Smarty ships with a special SmartyBC class that wraps those breaking changes, we just weren't using that. 
### Related issue(s)/PR(s)

This fixes #12895 and potentially other issues that rely on the old naming structure in Smarty (none logged yet that I'm aware of).

### DIFF
--- a/core/model/modx/smarty/modsmarty.class.php
+++ b/core/model/modx/smarty/modsmarty.class.php
@@ -34,7 +34,7 @@ include_once (strtr(realpath(dirname(__FILE__)) . '/../../smarty/Smarty.class.ph
  * @package modx
  * @subpackage smarty
  */
-class modSmarty extends Smarty {
+class modSmarty extends SmartyBC {
     /**
      * A reference to the modX instance
      * @var modX


theboxer:
:+1:

exside:
great catch, thanks!

PR #12905, opened 2016-02-13
Adds missing preserve_menuindex setting

### LINKED ISSUE #12696: Duplicate does not preserve menu order
When duplicating entire folders and their children, the menu order of the items is not preserved.

To Reproduce :
- Create a MODx resource
- Create multiple resources under that resource
- Create multiple grandchildren so you end up with a folder of folders, each with some children.
- Right-click the parent resource and choose to "Duplicate"

The result is generally a mixed order, menu index is randomly set :-(

It's really a big issue when copying (sub)tree for other contexts or creating things from a tree template...

pixelchutes:
When duplicating Resources, the menu index order will also be preserved.

Fixes #12696

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2077,4 +2077,13 @@ $settings['parser_recurse_uncacheable']->fromArray(array (
     'area' => 'system',
     'editedon' => null,
 ), '', true, true);
+$settings['preserve_menuindex']= $xpdo->newObject('modSystemSetting');
+$settings['preserve_menuindex']->fromArray(array (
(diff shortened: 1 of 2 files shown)


OptimusCrime:
How is this not an existing system setting?

pixelchutes:
The linked issue above describes in more detail, but I'm not sure why the setting itself was never added. You can see it in practice [in the code here](https://github.com/modxcms/revolution/blob/8c34e7336ed05f71976e54bb24fe9488c49ab1fb/core/model/modx/modresource.class.php#L1032).

OptimusCrime:
Oh. That makes sense. Nice work tho!

Issue #12906, opened 2016-02-13
Output filters broken for undefined placeholders in 2.4.3


yoleg:
When an undefined placeholder has an output filter, the entire tag is removed instead of treating the placeholder as empty. For example, this chunk with an undefined placeholder `fake`:

```
Start [[!+fake:notempty=`Not empty`:default=`Default`]] End
```

Results in:
```
Start End
```

Instead of (as expected):
```
Start Default End
```
 
This was caused by pull request #12884. You can see the explanation in the comments to #12884. Assuming no-one fixes it sooner, I will submit a fix today. 
 
Probable duplicate of #12901

OptimusCrime:
As far as I can see this is a 100% duplicate. Any reason to keep this open?

yoleg:
Sure, you can close this. Should I change the pull request to point to the other issue too, or it doesn't matter?

pixelchutes:
I have closed the original since this one already as an associated PR.

iLexN:
Not sure it is related to this topic or not. 

    [[*longtitle:notempty=`[[*longtitle]]`:else=`[[*pagetitle]]`]]
    
this work fine with php5.6 but cause error in php7

    Parse error: syntax error, unexpected ')' in /home/dev/ihi/httpdocs/core/model/modx/filters/modoutputfilter.class.php(191) : eval()'d code on line 1

after var_dump($conditional) in line 189;

    ....

    /home/dev/ihi/httpdocs/core/model/modx/filters/modoutputfilter.class.php:189:string '1' (length=1)

    /home/dev/ihi/httpdocs/core/model/modx/filters/modoutputfilter.class.php:189:string '0' (length=1)

    /home/dev/ihi/httpdocs/core/model/modx/filters/modoutputfilter.class.php:189:string '0' (length=1)

    /home/dev/ihi/httpdocs/core/model/modx/filters/modoutputfilter.class.php:189:string '' (length=0)
    
    Parse error: syntax error, unexpected ')' in /home/dev/ihi/httpdocs/core/model/modx/filters/modoutputfilter.class.php(191) : eval()'d code on line 1
    
find that the empty string cause error in php7, https://3v4l.org/GbMP9

Jako:
@iLexN notempty is not a conditional modifier. So the else outputfilter query won't work in 5.6 and 7. I think it should show those issues in 5.6 too.

Two valid solutions for your query 
```
[[*longtitle:default=`[[*pagetitle]]`]]
[...]

PR #12907, opened 2016-02-13
Proposed temporary fix for #12906

yoleg:
Summary: Check "global" parsing state instead of the current parsing state in modPlaceholderTag:process to decide if output filters should be run.

This seems to fix the #12906 bug, but the logic in `modPlaceholderTag:process` should probably be re-thought at some point.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -259,6 +259,12 @@ class modParser {
      * @return int The number of processed tags
      */
     public function processElementTags($parentTag, & $content, $processUncacheable= false, $removeUnprocessed= false, $prefix= "[[", $suffix= "]]", $tokens= array (), $depth= 0) {
+        if ($processUncacheable and !$this->minState(self::$STATE_UNCACHED)) {
+            $this->setState(self::$STATE_UNCACHED);
+        }
+        if ($removeUnprocessed and !$this->minState(self::$STATE_REMOVE)) {
+            $this->setState(self::$STATE_REMOVE);
+        }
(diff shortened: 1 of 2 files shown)


core/model/modx/modparser.class.php
```diff
@@ -110,6 +135,28 @@ public function setProcessingElement($arg = null) {
     }
 
     /**
+     * Sets the global parsing state.
+     *
+     * Explicit global parser states for clarity and easier debugging. Called by modParser in the 3 different states.
```
yoleg: This should say "Called by modResponse in 3 different states."

core/model/modx/modparser.class.php
```diff
@@ -31,6 +31,25 @@
...
+     */
+    public static $STATE_UNCACHED = '2_uncached';
+    /**
+     * State 3: Parser parses both cached and uncached and removes undefined tags.
+     *
+     * @var string
+     */
+    public static $STATE_REMOVE = '3_remove';
+
+    /**
```
yoleg: All this extra code is for clarity for people who are unfamiliar with the parser. It took me a long time stepping through the code to figure out the MODX parsing logic, and a simple _isGloballyProcessingUncached instance variable that is used in only one place would just add to the confusion. My logic:

* modResponse is the one switching between the "global states", so modResponse should be the one to switch the states.
* Even though only one state ($STATE_UNCACHED) is needed, adding all 3 states is more orthogonal and also helps quickly determine the state while stepping through the code in a debugger.

opengeek:
This was resolved via #12939

PR #12911, opened 2016-02-19
CSS - Remove ugly outline from navbar search

muzzwood:
### What does it do ?

Removes outline property from the navbar search icon in the manager.
### Why is it needed ?

Purely aesthetics

### DIFF
--- a/manager/templates/default/css/index.css
+++ b/manager/templates/default/css/index.css
@@ -6023,7 +6023,8 @@ ul.x-tab-strip-bottom {
   float: left;
   margin: 0;
   padding: 0;
-  position: relative; }
+  position: relative;
+  outline: 0; }
 
 #modx-navbar li:hover {
   z-index: 10000; }


Mark-H:
Outlines should only be removed if there's a suitable alternative in place for keyboard users.. Have you checked to make sure there is still is?

Mark-H:
Unfortunately it won't be possible to merge this as is, because you've edited the CSS and not the source sass files. These are under _build/templates/. Can you please update your pull request, or send a new one, to edit the right file?

digitalpenguin:
Sorry about the file - first time contributing. Will know for next time.
Ah I didn't realise it was there for a reason for it.
That's unfortunate, outlines are not the most appealing sight. 
Closing PR.

Mark-H:
Like I said, it is acceptable for browser default outlines to be removed, but only if there is an alternative visual cue as to what is in focus, I haven't personally tried this patch  yet to see if that would still be there. As you mentioned the reason being purely aesthetic, I wanted to make sure  accessibility wasn't forgotten.

digitalpenguin:
No problem! I checked and yes the outline is the only indicator. I'll see if I can replace it with something nicer and make another pull request.

jpdevries:
See https://github.com/modxcms/revolution/pull/12824 and http://tjvantoll.com/2013/01/28/stop-messing-with-the-browsers-default-focus-outline/

PR #12916, opened 2016-02-23
Replace hard-coded charset with MODX setting

### LINKED ISSUE #12915: New startup theme charset meta tag
### Summary
The new default template's charset meta tag has a hard-coded value.
### Step to reproduce
Edit the new default template and look at the charset meta tag.

### Observed behavior
The value is hard-coded to be utf-8. (whether or not this tag should be explicitly self-closed is a matter of personal preference in HTML 5).
```
<meta charset="utf-8" />
```

### Expected behavior
It should be using a system setting. While it probably won't cause a problem, MODX uses the system setting 'modx_charset' when setting the HTML headers for the page. If the headers don't match the meta tag in the HTML head, at the least the W3C validator will complain.
```
<meta charset="[[++modx_charset]]" />
```

### Environment
MODX Revo 2.5.0

zaigham:
### What does it do ?

Replace hard-coded charset with MODX setting in new default template
### Why is it needed ?

Because MODX tags are dynamic and preferred over hard coded string.
### Related issue(s)/PR(s)

Resolves #12915

### DIFF
--- a/setup/templates/base_template.tpl
+++ b/setup/templates/base_template.tpl
@@ -3,7 +3,7 @@
 <head>
     <title>[[*pagetitle]] - [[++site_name]]</title>
     <base href="[[!++site_url]]" />
-    <meta charset="utf-8" />
+    <meta charset="[[++modx_charset]]" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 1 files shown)


Mark-H:
"Self explanatory" is not much better than just skipping the template entirely :/

zaigham:
Hey @Mark-H. I updated the text to be better than that. :smile:

PR #12922, opened 2016-02-25
E-Mail regex should support more than 6 chars in TLD

Jako:
### What does it do ?

Change the ExtJS regular expression validating email addresses.
### Why is it needed ?

The previous regular expression detects only a TLD with 2-6 chars. They could be longer these days. 25 chars according to the longest entry on https://www.icann.org/resources/pages/tlds-2012-02-25-en.


christianseel:
Thanks @Jako !

OptimusCrime:
25 char long TLD? Madness!

Jako:
Sorry for messing around in that pull request.

PR #12925, opened 2016-02-28
Simpler temporary fix for #12906

yoleg:
### What does it do?

Check a global variable to see if uncached parsing has _ever_ started instead of checking the current cached vs uncached parsing state in `modPlaceholderTag:process` to decide if output filters should be run.
### Why is it needed?

In case MODX devs prefer a simpler fix for issue #12906 than the one in #12907
### Why is it temporary?

This fix seems to work great, but the question is still why a global "uncached parsing has started" variable is even needed if uncached and cached parsing now switches turns (from pull request #12884). When I remove `|| $this->modx->parser->startedProcessingUncacheable()`, other stuff breaks. I have not looked into that yet. 
### Related issue(s)/PR(s)

Alternative to pull request #12907. Fixes issue #12906 (which was caused by pull request #12884).

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -55,6 +55,12 @@ class modParser {
      * @var bool $_removingUnprocessed
      */
     protected $_removingUnprocessed = false;
+    /**
+     * If the parser has ever processed uncacheable
+     *
+     * @var bool $_startedProcessingUncacheable
+     */
+    protected $_startedProcessingUncacheable = false;
 
     /**
      * @param xPDO $modx A reference to the modX|xPDO instance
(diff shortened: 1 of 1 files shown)


OptimusCrime:
In terms of speed/performance, which of these options are the best?

yoleg:
Technically, this one uses a few less method calls and assignments and so in theory would be faster, but I'm pretty sure that would not make a significant difference. They both do pretty much the same thing; #12907 is just wordier about it.

OptimusCrime:
I think this is prettier in terms of code. +1 for this

opengeek:
I like this fix better as well @yoleg , but can you re-submit this one against the 2.4.x branch instead of 2.x so I can merge it forward as I release 2.4.4 and 2.5.0?

yoleg:
@opengeek OK, done: #12939

Issue #12887, opened 2016-02-01
The account menu in Manager too close to save button


DavPah:
Many are the times I've lost changes, because moving cursor too fast, ending up opening the user menu and clicking "Edit Account" when aiming for the save button.

Maybe possible to move it or solve this in some other manor?

(Yes, most times you can use command/ctrl-Save, but not when cursor being placed in a richtext editor window – then the web browser's save function kicks in instead. That's why I tend to use the Save button.)

sottwell:
Those drop-down on hover menus drive me nuts anyway. Pain in the neck.

DavPah:
Maybe a setting to turn off the auto-open-on-hover could be added? That would be pretty nifty actually.

sottwell:
That would mean that the sub-menus would never be available. It would need some Javascript to make it open on click.

This is why when I originally designed the top menu bar for Evo I used a "dropline" menu, where the submenus are listed in a secondary menu bar rather than a dropdown.

christianseel:
Just to also have a different opinion: I never had that issue to click the user menu instead of save and I like the dropdown menus in Revo much more than the Evo menu.

But a optional setting to require a click on 1st level menu items to reveal the dropdown should not hurt.

DavPah:
Well, for me the only thing needed really, is perhaps relocate the Save button a bit on new rescources/TVs/chunks etc, because it winds up directly under the menu and you lose all things you've typed in, when accidentally mis-click.

For example have the buttons aligned left instead of right or something, keeping SAVE at the same spot every time.

madze:
Second DavPah.

Or just swapping "Save" and "Help" would probably work as well.

inreti:
In fact, I had the same problems with the account menu popping up accidently when I wanted to hit the save button. I don't have a proper solution in mind right know, but +1 for anything making the situation better.

Mark-H:
I wonder if maybe a CSS transition can delay opening the menus?

DavPah:
The delay idea is a good one. And it does not only apply to the save button. Everytime I'm reaching for the row of buttons and not being careful, I often pass over the account menu and wind up in the wrong place if not careful.

I changed the value in the manager's index.css

#modx-navbar li:hover ul.modx-subnav
{visibility:visible;
opacity:1;
filter:alpha(opacity=100);
transition:all .15s ease;
transition-delay:.2s
}

Just changing the delay from .2 to .7 made a huge difference.

mindeffects:
[...]

PR #12931, opened 2016-03-04
Increase the hover delay as suggested on #12887

Jako:
### What does it do?

Increase the hover delay for dropdown.
### Why is it needed?

See the discussion on #12887
### Related issue(s)/PR(s)
#12887


Mark-H:
Hm this almost feels like it takes too long now for the menus to open. I've changed the transition to .5s instead of .7s but if people feel like it's still too quick we can up it some more. 

Merged into 2.5.x for 2.5.1 in 362e42452be3b992eee5df642b87f2481cea7178

PR #12936, opened 2016-03-16
Enhancement: Add package events

exside:
### What does it do?

Adds 3 system events that are fired when interacting with the package manager: **OnPackageInstall** (also fired when a package is updated), **OnPackageUninstall** and **OnPackageRemove**

Added these events to the build script for proper setup as well.
### Why is it needed?

Can be helpful to trigger additional scripts when interacting with packages.
### Related issue(s)/PR(s)

None that I knew of.

### DIFF
--- a/_build/data/transport.core.events.php
+++ b/_build/data/transport.core.events.php
@@ -3,7 +3,9 @@
  *
  * @package modx
  */
+
 $events = array();
+
 
 /* Plugin Events */
 $events['OnPluginEventBeforeSave']= $xpdo->newObject('modEvent');
@@ -31,6 +33,7 @@ $events['OnPluginEventRemove']->fromArray(array (
   'groupname' => 'Plugin Events',
 ), '', true, true);
 
+
 /* Resource Groups */
 $events['OnResourceGroupSave']= $xpdo->newObject('modEvent');
 $events['OnResourceGroupSave']->fromArray(array (
@@ -56,6 +59,7 @@ $events['OnResourceGroupBeforeRemove']->fromArray(array (
   'service' => 1,
   'groupname' => 'Security',
 ), '', true, true);
+
 
 /* Snippets */
 $events['OnSnippetBeforeSave']= $xpdo->newObject('modEvent');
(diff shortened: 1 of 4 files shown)


Mark-H:
This would be nice for Gitify to use :)

rtripault:
While i'm not seeing any use case (please, feel free to shed some light), nothing against the idea.

Just some "concerns" : 

1. events appears to be triggered when actions have been performed, so i would rather go event names like `OnPackageInstalled`, `OnPackageRemoved`, `OnPackageUninstalled`
2. `OnPackageInstall` is also occurring on upgrades. Could be "interesting" to distinguish the events (`modTransportPackage::previousVersionInstalled()` could be helpful i guess)
3. why passing the `modTransportPackage` as an array instead of the object ?

Mark-H:
The use case for Gitify that came to mind is that it could keep better track of installed by packages. It has a list of packages in its `.gitify` file of packages that should be installed when deploying. This could be kept up to date easier when these events are added. 

Concern 1 seems to be inconsistent with existing naming patterns (e.g. the first in the file are OnPluginEventBeforeSave and OnPluginEventSave, not OnPluginEventSave and OnPluginEventSaved). Agreed on the other two concerns though, any thoughts @exside?

exside:
Updated the PR based on the feedback, not sure if that 'action' thing in the install event acutally does what it's intended to, didn't have time to test yet...or should I just make an "update": true||false from that?

rtripault:
Woops, sorry for the late reply here... you are absolutely right Mark, better keep the existing naming pattern.

@exside if we want to be consistent, i guess a `mode` key would be best (that's the one used for existing `On*Save` events)

Mark-H:
Thanks @exside, merged into 2.x for inclusion in 2.6. 

Also added documentation stubs at https://docs.modx.com/revolution/2.x/developing-in-modx/basic-development/plugins/system-events/onpackageremove, https://docs.modx.com/revolution/2.x/developing-in-modx/basic-development/plugins/system-events/onpackageinstall and https://docs.modx.com/revolution/2.x/developing-in-modx/basic-development/plugins/system-events/onpackageuninstall

PR #12940, opened 2016-03-21
E-Mail regex should support more than 6 chars in TLD

Jako:
### What does it do ?

Change the ExtJS regular expression validating email addresses.
### Why is it needed ?

The previous regular expression detects only a TLD with 2-6 chars. They could be longer these days. 25 chars according to the longest entry on https://www.icann.org/resources/pages/tlds-2012-02-25-en.


OptimusCrime:
:+1:

Mark-H:
I seem to recall an issue about this with some more details on reproducing it, but can't find it right now. Any idea @Jako?

Mark-H:
Merged into 2.5.x for 2.5.1, but if you remember where this was previously discussed let me know :P

Jako:
Ask @christianseel about this. He had one tld in use with 8 chars that can't be entered in an email field, i.e. email_sender, if I remember right.

christianseel:
This is required for the users email. We discussed this in slack...

OptimusCrime:
The use case is pretty straight forward. The regex allows TLDs with a length below 7. 

As stated here, the length can be much longer: http://stackoverflow.com/questions/9238640/how-long-can-a-tld-possibly-be

Issue #12937, opened 2016-03-18
Import HTML regex breaks if newlines in title


sottwell:
### Summary
When importing an HTML document the regex that parses the title tags breaks if there are newlines in the title.

### Step to reproduce
Add some extra lines to the title tag in an HTML page's head. Import the HTML document.

### Observed behavior
The title will not be used as the resource's pagetitle, it will default to the base filename.

### Expected behavior
It should use the HTML title. In order to do so the preg_match needs to use the s modifier on line 104 of core/model/modx/import/modstaticimport.class.php:
```
if (preg_match("/<title>(.*)<\/title>/siU", $file, $matches)) {
```
Now it will use the HTML title as expected, even with extra lines in the title.

### Environment
N/A

Jako:
It would be better to change the regex to trim the whitespace away.

```
if (preg_match("/<title>\s*(.*)\s*<\/title>/siU", $file, $matches)) {
```

sottwell:
Whatever works. Regular expressions are not my strong point.

Mark-H:
Fixed in #12941

PR #12941, opened 2016-03-21
Fixing #12937 Import HTML regex breaks if newlines in title

Jako:
### What does it do?

Change the regular expression that detects the title tag and extracts the title
### Why is it needed?

When importing an HTML document the regex that parses the title tags breaks if there are newlines in the title.
### Related issue(s)/PR(s)
#12937


Mark-H:
Merged into 2.5.x for inclusion in 2.5.1 (230a588), thanks!

PR #12959, opened 2016-04-13
Fix issue where site_start and default_template settings get the wrong ID on certain environments

### LINKED ISSUE #12958: site_start and default_template system settings on install
### Summary
In rare cases, `site_start` and `default_template` system settings get the wrong values on install.

### Step to reproduce
Setup a galera clustered environment. Install MODX on the various nodes. Notice how the different servers will get different IDs for the Base Template and default Resource, but the System Settings will not reflect this. This makes bad things happen.

### Observed behavior
Without a valid `site_start` you get a server error on the front-end. Without a valid `default_template` new Resources are kinda borked.

### Expected behavior
`site_start` should have the actual ID of the default installed Resource. `default_template` should have the actual ID of the default installed Base Template.

### Environment
MODX 2.4.x in specialized cluster environment.

sepiariver:
### What does it do?

Fixes #12958 by adding some calls to get and save System Settings when Base Template and default Resource are created in `setup/includes/new.install.php`
### Why is it needed?

See issue #12958 
### Related issue(s)/PR(s)
#12958

### DIFF
--- a/setup/includes/new.install.php
+++ b/setup/includes/new.install.php
@@ -151,6 +151,23 @@ $template->fromArray(array(
     'content' => $templateContent,
 ));
 if ($template->save()) {
+    
+    /** @var modSystemSetting $setting */
+    $setting = $modx->getObject('modSystemSetting',array(
+        'key' => 'default_template',
+    ));
+    if (!$setting) {
+        $setting = $modx->newObject('modSystemSetting');
+        $setting->fromArray(array(
+            'key' => 'default_template',
+            'namespace' => 'core',
+            'xtype' => 'modx-combo-template',
+            'area' => 'site',
+        ));
+    }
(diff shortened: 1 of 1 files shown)


Mark-H:
Nothing's standing in the way of this getting merged from what I can see, but as a piece of advice @sepiariver you should **always** make a new branch on your fork for **every different contribution**. 

By committing directly to your 2.x branch you're setting yourself up for some git headaches trying to get things in sync again later, because your 2.x branch might contain things that are not in the upstream 2.x, yet the upstream 2.x has also gotten new commits. 

I'd recommend creating branches like `issue-12958` or `issue-broken-settings` if you don't have an issue ID, but what you call them is up to you.

sepiariver:
@Mark-H thanks for the feedback/advice. Will heed for future 👍

PR #12960, opened 2016-04-14
Make sure uberbar resource search respects ACLs

christianseel:
### What does it do?

Adds a context_key check to the query for searching through resources.
### Why is it needed?

Previously it was possible to find resources that belong to a context where a user didn't have access permissions for. By calling `getCollection('modContext')` MODX checks the context ACLs and returns only contexts that are accessible for the current user.
### Related issue(s)/PR(s)

https://modxcommunity.slack.com/archives/general/p1460634256003370

### DIFF
--- a/core/model/modx/processors/search/search.class.php
+++ b/core/model/modx/processors/search/search.class.php
@@ -147,7 +147,7 @@ class modSearchProcessor extends modProcessor
         $typeLabel = $this->modx->lexicon('search_resulttype_' . $type);
 
         $contextKeys = array();
-        $contexts = $this->modx->getCollection('modContext', array('key:NOT IN' => array('mgr')));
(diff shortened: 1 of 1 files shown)


core/model/modx/processors/search/search.class.php
```diff
@@ -146,14 +146,25 @@ public function searchResources()
         $type = 'resources';
         $typeLabel = $this->modx->lexicon('search_resulttype_' . $type);
 
+        $contextKeys = array();
+        $contexts = $this->modx->getCollection('modContext', array('key:NOT IN' => array('mgr')));
```
Mark-H: This query could be simplified to

`array('key:!=' => 'mgr')`

christianseel: Thanks, Mark. Fixed.

Issue #12951, opened 2016-04-10
References of passed scriptProperties gets lost on its way to plugins


Bruno17:
### Summary
I'm trying to manipulate the TV-categories-array with a plugin, runninig at OnResourceTVFormRender.
It passes nicely a reference with the 'categories' - property in ResourceManagerController->loadTVs. 
But this reference gets lost in modParser->parseProperties, so it isn't possible to manipulate the array within the plugin.

### Step to reproduce
Create a plugin with the system-event OnResourceTVFormRender:
```
<?php

$value = '';
$params = array();

$category = array();
$category['category'] = 'Test';
$tv = $modx->newObject('modTemplateVar');
$tv->set('type', 'text');
$tv->_fieldMeta['id']['phptype'] = 'string';
$tv->set('caption','Test'); 
$tv->set('id', '_1');
$inputRenderPaths = $tv->getRenderDirectories('OnTVInputRenderList', 'input');
$inputForm = $tv->getRender($params, $value, $inputRenderPaths, 'input', null, $tv->get('type'));
$tv->set('formElement', $inputForm);

$category['tvs'][] = $tv; 
$scriptProperties['categories'][] = $category;

$modx->event->resetEventObject();
return ;
```

### Observed behavior
There is no new Category with a new TV-field added to the TV-Tab. 

### Expected behavior
A new Category with one TV-field should be added to the TV-Tab.

### Possible Solution:
change modParser->parseProperties to
```
    public function parseProperties($propSource) {
        $properties = array();
        if (!empty($propSource)) {
            if (is_string($propSource)) {
                $properties = $this->parsePropertyString($propSource, true);
            } elseif (is_array($propSource)) {
                foreach ($propSource as $propName => &$property) {
                    if (is_array($property) && array_key_exists('value', $property)) {
                        $properties[$propName] = $property['value'];
                    } else {
                        $properties[$propName] = &$property;
                    }
                }
            }
        }
        return $properties;
    }
```

### Environment
Revo 2.5.0-rc2

PR #12967, opened 2016-04-19
#12951 References of passed scriptProperties gets lost on its way to …

Bruno17:
### What does it do?

$property as reference
### Why is it needed?

Reference to values passed by reference got lost and could not longer be manupulated within plugins
### Related issue(s)/PR(s)
#12951


Bruno17:
Any chance, this gets merged?
Sorry for asking, but othewise I need to solve it with a javascript - solution for my Extra, while I prefer the serverside plugin-solution to add special Extra-TV-fields.

PR #12968, opened 2016-04-19
store and restore plugin-event-name before running inner plugins

Bruno17:
### What does it do?

stores modx->event->name and restores it after running plugins invoked by other plugins
### Why is it needed?

event->name gets lost, when a plugin is invoked within another plugin. 
For reference see 
https://forums.modx.com/thread/97694/support-comments-for-tinymcewrapper?page=10#dis-post-532301

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1583,6 +1583,7 @@ class modX extends xPDO {
             //$this->log(modX::LOG_LEVEL_DEBUG,'System event '.$eventName.' was executed but does not exist.');
             return false;
         }
+        $oldEventName = $this->event->name;//store current eventname        
         $results= array ();
         if (count($this->eventMap[$eventName])) {
             $this->event= new modSystemEvent();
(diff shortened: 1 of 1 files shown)


core/model/modx/modx.class.php
```diff
@@ -1583,6 +1583,7 @@ public function invokeEvent($eventName, array $params= array ()) {
             //$this->log(modX::LOG_LEVEL_DEBUG,'System event '.$eventName.' was executed but does not exist.');
             return false;
         }
+        $oldEventName = $this->event->name;//store current eventname
```
rtripault: Isn't there a case where `$modx->event` is not yet an object ? (i guess before the first event is fired)

rtripault:
Sorry for the time it took to comment here.
While i get the problem you are trying to solve here, i'm not sure about the proposed solution.

What would you think about story the whole `modSystemEvent` event instead ?

```
$previousEvent = $this->event; // $this-event is allowed to be null
...
$this->event = $previousEvent;
```

PR #12969, opened 2016-04-19
new grunt bump task

jpdevries:
### What does it do?

Added a new `grunt bump` tasks that  will update the README.md copyright messaging to
the current year and update the version to the current MODX version in package.json. It also adds a console log that whenever any grunt tasks will output something like `Grunting MODX-2.5.0-rc2`. 
### Why is it needed?

It isn't critical, but this automates the process of keeping copyright, versioning, and release info up to date.
### Related issue(s)/PR(s)

modxcms/revolution#12930

grunt bump:copyright will update the README.md copyright messaging to
the current year
grunt bump:pkg will update the version to the current MODX version in
package.json

This should be run before tagging new releases.

```
grunt bump
grunt build
```

### DIFF
--- a/_build/templates/default/gruntfile.js
+++ b/_build/templates/default/gruntfile.js
@@ -32,16 +32,34 @@ var coreJSFiles = [
 ];
 
 module.exports = function(grunt) {
+  var parseString = require('xml2js').parseString;
+  var xml = grunt.file.read('../../build.xml');
+  var coreVer,
+  coreRel;
+  parseString(xml, function (err, result) {
+      coreVer = getXMLPropByName('modx.core.version').value;
+      coreRel = getXMLPropByName('modx.core.release').value;
+      console.log('Grunting MODX ' + coreVer + '-' + coreRel);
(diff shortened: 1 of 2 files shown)


jpdevries:
If it would be useful, I could add an option to bump the version number and release in the `build.xml` file as well. Something like

```
grunt bump -ver=2.6.0 -rel=rc1
```

Mark-H:
Do we really need a grunt task to automate something that happens once a year and only really affects the readme (I don't think the package.json version is used for anything)? 

Even when discounting the time it takes to add it to documentation and making core integrators aware it exists, it seems to me that you spent more time on adding this to grunt, then it would take to update the copyright by hand for the next decade or so. 

This comes to mind.. https://xkcd.com/1205/

Mark-H:
As issuing releases already has a different workflow, the package.json change is irrelevant and a grunt task for automating changing a year really is overkill, so I'm going to close this. Appreciate thinking along with the release managers but this does not appear to make things simpler.

PR #12971, opened 2016-04-19
Add uri to mysql/modcontext.class.php

### LINKED ISSUE #12744: ResourceMap empty on mariadb installation
Hi,
I seem to have the same issue as described here: #9901 but with mariadb instead of SQL Server. I came across this in search of a reason why my wayfinder call:

`[[Wayfinder? &startId=`1` &outerTpl=`nav_ul` &rowTpl=`nav_li` &ignoreHidden=`1` &level=`1`]]`

which seems to be correct, doesn't return any items. With debug turned on, it would only return "No resources found for menu.". I tried searched the forums for similar issues, and tried several ways to fix it including [opengeek's recalculateURIs.php](https://gist.github.com/opengeek/917662) which seemed to have helped someone else on the forum. None helped.

Finally, I came across issue #9901 the and tried the same with core/model/modx/mysql/modcontext.class.php - the SQL statement generated by this function looks like this:

`SELECT `r`.`id`, `r`.`parent`, `r`.`uri` FROM `modx_site_content` `r` FORCE INDEX (`cache_refresh_idx`) LEFT JOIN `modx_context_resource` `cr` ON `cr`.`context_key` = 'web' AND `r`.`id` = `cr`.`resource` WHERE `r`.`id` != `r`.`parent` AND (`r`.`context_key` = 'web' OR `cr`.`context_key` IS NOT NULL) AND `r`.`deleted` = 0 GROUP BY `r`.`parent`, `r`.`menuindex`, `r`.`id``

and yields the following error message when executed directly in mariadb console:

ERROR 1055 (42000): 'modx.r.uri' isn't in GROUP BY

I have added the missing field in the SQL statement, and changed the affected line in modcontext.class.php, and Wayfinder started working again.

Detailed information about my installation:
MODX version: 2.4.2-pl , upgraded from 2.3.5-pl (fresh install)
Mariadb version: 10.0.21

Hmm, tried to attach a diff here and failed...and you might just not be quite happy with a diff in JPG format ;-)
Please add `, `r`.`uri`` at the end of line 17 in modcontext.class.php to fix this issue.

Regards,
Marc

OptimusCrime:
### What does it do?

Add `uri` to `GROUP BY` clause in mysql/modcontext.class.php.
### Why is it needed?

Fixes issue in mariadb.
### Related issue(s)/PR(s)

Fixes #12744

### DIFF
--- a/core/model/modx/mysql/modcontext.class.php
+++ b/core/model/modx/mysql/modcontext.class.php
@@ -17,7 +17,7 @@ class modContext_mysql extends modContext {
             $resourceFields= array('id','parent','uri');
             $resourceCols= $context->xpdo->getSelectColumns('modResource', 'r', '', $resourceFields);
             $bindings = array($context->get('key'), $context->get('key'));
-            $sql = "SELECT {$resourceCols} FROM {$tblResource} `r` FORCE INDEX (`cache_refresh_idx`) LEFT JOIN {$tblContextResource} `cr` ON `cr`.`context_key` = ? AND `r`.`id` = `cr`.`resource` WHERE `r`.`id` != `r`.`parent` AND (`r`.`context_key` = ? OR `cr`.`context_key` IS NOT NULL) AND `r`.`deleted` = 0 GROUP BY `r`.`parent`, `r`.`menuindex`, `r`.`id`";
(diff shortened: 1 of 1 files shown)


Jako:
Merged to 2.5.x branch.

Issue #12221, opened 2014-12-18
Change css for title/template labels


pepebe:
If you hide the template select field via manager customization, the rules for the hidden element don't apply to the next element that moves to the top (usually resource alias), because template select is not removed from the dom, but just hidden from view. As a consequence, the vertical rhythm is broken:

![bildschirmfoto 2014-12-18 um 10 37 07](https://cloud.githubusercontent.com/assets/667315/5486113/655e2700-86a3-11e4-9393-06a33db65044.png)

The root of this problem can be found in index.css:

```
.x-form-label-top .x-form-item:first-of-type label.x-form-item-label {
padding: 0 0 4px;
}
```
It overwrites the general css for the labels enclosed in the first .x-form-item found in each column.

```
.x-form-label-top .x-form-item label.x-form-item-label {
display: inline-block;
margin: 0;
padding: 15px 0 4px;
}
```

I can't see any benefit of this rule, so I vote for removing it from https://github.com/modxcms/revolution/blob/master/manager/templates/default/css/index.css

Cheers

OptimusCrime:
@modxbot close

PR #12973, opened 2016-04-19
Remove :first-of-type, reduce padding on container

OptimusCrime:
### What does it do?

As noted in #12221, if you hide the template dropdown using form customization, you will have a weird offline padding on the left and right panels. This is due to the line of css that reduces the padding for the `:first-of-type` element, which will still be the template dropdown.

This pull removes this reduced padding (it is only used in the resource interface), and instead removes the padding given to the entire resource container.
### Why is it needed?

Argument given above
### Related issue(s)/PR(s)
#12221

### DIFF
--- a/_build/templates/default/sass/_forms.scss
+++ b/_build/templates/default/sass/_forms.scss
@@ -252,10 +252,6 @@ input::-moz-focus-inner {
     padding-bottom: 0; /* override extjs default theme styles */
 
     &:first-of-type {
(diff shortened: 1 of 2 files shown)


Jako:
Removed a superfluous closing brace and merged into 2.5.1.

Jako:
This patch has a padding-top issue in a `.modx-window`. Maybe a `.modx-window.x-window .x-window-body { padding-top: 0 }` is enough. Could you investigate that too?

Issue #12557, opened 2015-08-17
Form Customization Title


Jako:
### Summary
Wrong title during editing a Form Customization Set.

### Step to reproduce
Edit a Form Customization Set.

### Observed behavior
Title is 'Create New Set'.

### Expected behavior
Title should be 'Edit Set'.

### Environment
MODX 2.3.5

OptimusCrime:
Now it simply says `Form Customization`. 

@modxbot close

OptimusCrime:
@modxbot open

OptimusCrime:
Too fast on the trigger there. I located the correct placement. Sorry about that.

OptimusCrime:
@modxbot close :)

OptimusCrime:
@modxbot close :)

PR #12974, opened 2016-04-19
Set correct title to edit fc set

OptimusCrime:
### What does it do?

The title for edit fc set is wrong. It says `Create New Set`. While in the popup at `security/forms/profile/edit` this is correct, it is impossible to access this view in the browser (there is no `security/forms/set/new`, only `security/forms/set/update`.
### Why is it needed?

Misleading title
### Related issue(s)/PR(s)

Reported in #12557


Jako:
Merged to 2.5.x branch.

PR #12979, opened 2016-04-23
Corrected array index for $_lang['setting_anonymous_sessions_desc']

enigmatic-user:
$_lang['setting_anonymous_sessions'] appears twice, the second one should be $_lang['setting_anonymous_sessions_desc']. I also changed "css" to the official notation "CSS" once.


Jako:
Merged to 2.5.x branch.

Issue #12694, opened 2015-10-02
Absent Snippet Logging


amdbuilder:
### Summary
Log attempts to use non-existant snippets

### Step to reproduce
Attempt to use a snippet that isn't installed.

### Observed behavior
https://twitter.com/sepiariver/status/646834408788393984

### Expected behavior
Indicate the snippet being called and the location of that call if possible (ie: resource id, chunk name, template name) in the MODX Error Log

### Environment
Any Version

whitebyte:
#12467

OptimusCrime:
Duplicate of #12467. Closing in favor of issue with more comments.

@modxbot close

Issue #12467, opened 2015-06-15
No log entries for non-existent snippet calls?


whitebyte:
### Summary

When we call a non-existent snippet, it silently fails (e.g. produces no output at all). I think it should add some kind of warning into the MODX logs, because it's a source of non-obvious errors.
### Environment

MODX 2.3.x

TheBoxer:
You can create a plugin which will let you know about non-existing snippet call. I don't think that this should be part of the core...

whitebyte:
Of course i do, but undefined snippet call is an exceptional condition and shouldn't be just silently ignored by default. This is really frustrating in certain cases. We recieve warnings and errors in PHP when trying to call some non-existent stuff, so why MODX parser shouldn't do the same? This is expected behavior.

sottwell:
+1 (for what it's worth, I think this would be a good idea)

inreti:
+1. MODX core should support development as good as possible. Logging these could help a lot.

Alroniks:
+1 for logging.

garryn:
This would be good if the error message also logged what element it was called from (and/or resource). Perhaps have it on log level INFO too.

rtripault:
:-1:  here.

This is something useful for development "only", and already possible with the existing events system (as explained by @theboxer). Should be up to the developer(s) to set their tools/environment.

On an another hand, :+1: for having more properties available as suggested by @garryn (for what it worth, `OnElementNotFound` is triggered from [modParser::getElement](https://github.com/modxcms/revolution/blob/develop/core/model/modx/modparser.class.php#L564)

mindeffects:
:+1: Actually it's a :+10: from my site (no icon for that), because I already spent hours searching for typos in snippet/chunk calls! So, yes, error log, please!

BTW: The whole error-log is for "developers only". So, either we kick it completely out or do it in a proper way! ;-)

OptimusCrime:
I just submitted a pull request with this functionality. Feel free to take a look at it: #12984.

Mark-H:
Will be included in 2.6. There's a setting to turn it off.

PR #12984, opened 2016-04-26
Implement logging when calling snippets that does not exist

OptimusCrime:
### What does it do?

This pull request adds functionality to log snippets that are called but not found in the system to the error log.

A new system setting called `log_snippet_not_found` is added, which is `No` by default to avoid breaking changes. The description for this system setting is: 

> Log snippets not found
> If set to yes, snippets that are called but not found will be logged to the error log.

If turned to `Yes` the following is added to the error.log when a snippet that is not found is called:

> Could not find snippet with name SNIPPET-NAME-HERE.
### Why is it needed?

During development, one of the most common errors to make is to misspell snippets and chunks. Misspelling of chunks can be identified by the lack of HTML output, while misspelling of snippets may be harder to debug efficiently.

This pull request offers a better way to deal with this problem. While this functionality could be handled by a plugin, it feels a bit unnecessary to write a plugin or install an Extra to keep track of misspellings during development.

The implementation suggested here is opt-in, so people that does not want to use this functionality can keep living their lives without it without having to do anything.
### Related issue(s)/PR(s)

Discussed in #12467 (duplicate of #12694).
### Sidenotes

Open for suggestions to both the system setting name, the description and the actual error.log text.

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2086,4 +2086,13 @@ $settings['preserve_menuindex']->fromArray(array (
     'area' => 'manager',
     'editedon' => null,
 ), '', true, true);
+$settings['log_snippet_not_found']= $xpdo->newObject('modSystemSetting');
+$settings['log_snippet_not_found']->fromArray(array (
+    'key' => 'log_snippet_not_found',
+    'value' => false,
+    'xtype' => 'combo-boolean',
+    'namespace' => 'core',
+    'area' => 'site',
+    'editedon' => null,
+), '', true, true);
 return $settings;
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -813,3 +813,6 @@ $_lang['setting_default_username_desc'] = 'Default username for an unauthenticat
 
 $_lang['setting_manager_use_fullname'] = 'Show fullname in manager header ';
 $_lang['setting_manager_use_fullname_desc'] = 'If set to yes, the content of the "fullname" field will be shown in manager instead of "loginname"';
+
+$_lang['log_snippet_not_found'] = 'Log snippets not found';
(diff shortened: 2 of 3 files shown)


Mark-H:
Nice, can see this being useful. Any particular concerns relating to BC that made you default it off?

OptimusCrime:
@Mark-H Only to follow semver really. If you like, I could change it to default it on. Not sure what people want.

pixelchutes:
It's purely semantics, but should this be `log_element_not_found`? Looks like it covers Chunks + TVs, too?

This is great!

OptimusCrime:
As far as I can understand, it covers only Snippets. In the switch, the default looks like it targets snippets. Look at the call to `modSnippet` below this line: https://github.com/OptimusCrime/revolution/blob/246559987cb4ff5d64aca3b7e6fb52a9e7e043ce/core/model/modx/modparser.class.php#L515

pixelchutes:
Ah, good catch. Missed that it was tied to the default/fallback handling in order to work. Still awesome.

OptimusCrime:
Any more ideas on this? Should I turn automatic logging on? Change the wording somehow, or is this more or less good to go?

whitebyte:
What about absent chunks logging?

OptimusCrime:
@whitebyte The idea as it was posted was that snippets were the most difficult ones to debug, because they fail "silently". Chunks, templates, etc are easier as they would output (or not output) something if they failed or were misspelled. But sure, I can add more logging.

pixelchutes:
LGTM! 😉

OptimusCrime:
Taking a wild guess at "Lets Get This Merged", @pixelchutes ? ;)

Anything you'd like me to change or adjust before this might be an option?

Mark-H:
Thanks @OptimusCrime - merged into 2.x for inclusion in 2.6. I did decide to change the default from false to true because Mike +1'd that, I was thinking default on would be fine, and you seemed open to either defaults :P

OptimusCrime:
No problem, thanks @Mark-H !

Issue #12355, opened 2015-02-10
Prevent empty manager theme


Lefthandmedia:
While switching from manager theme in the system settings something went wrong and I ended up with an empty manager theme setting (manager_theme). After that the manager hardly works and the only way to fix this was to edit the database directly.
When loading the theme there should be a check and default to 'default' when empty

$theme = ($manager_theme != '')? $manager_theme : 'default';

rtripault:
This is supposed to be the case, see https://github.com/modxcms/revolution/blob/master/core/model/modx/modmanagercontroller.class.php#L127

The only way i could see this going "wrong" is selecting a theme then delete its folder, or manually setting the value to a non existing theme

Lefthandmedia:
Hi Romain,
no the theme folder was not deleted the theme setting was just empty.
this:  `$this->theme = $this->modx->getOption('manager_theme',null,'default');`
seems to just check if the setting exists but it does not check if its empty.
When you clear the setting in System settings  you end up with an unstyled / unmanageable manager.
You can recreate the problem by emptying the theme setting,
the code should be something like

`$this->theme = $this->modx->getOption('manager_theme',null,'default');
if($this->theme ==''){$this->theme='default'};
`

Mark-H:
Setting a 4th param of true on the getOption would have the same effect.
Op 15 feb. 2015 01:31 schreef "Lefthandmedia" <notifications@github.com>:

Lefthandmedia:
So the line should be
$this->theme = $this->modx->getOption('manager_theme',null,'default',true)

OptimusCrime:
@Lefthandmedia did you ever open a pull request for this? Is it merged?

Lefthandmedia:
I dont hink its ever been merged

OptimusCrime:
Nice. Lets hope they merge it.

OptimusCrime:
Fixed.

@modxbot close

PR #12986, opened 2016-04-26
Avoid empty manager theme

OptimusCrime:
### What does it do?

Adds a fourth parameter to the getOption that decides what manager theme to load. With this setting it will return the "default" theme even if the setting is empty.
### Why is it needed?

Might crash the manager. Only way to fix it is via database.
### Related issue(s)/PR(s)

Reported in #12355. Looks like it was fixed in [this fork](https://github.com/Lefthandmedia/revolution/commit/63c584e8a9e01e9da52aa25430a7dc1cf4c60563), but it was never pull requested or merged.

### DIFF
--- a/core/model/modx/modmanagercontroller.class.php
+++ b/core/model/modx/modmanagercontroller.class.php
@@ -126,7 +126,7 @@ abstract class modManagerController {
             'action' => $this->config,
         ));
 
-        $this->theme = $this->modx->getOption('manager_theme',null,'default');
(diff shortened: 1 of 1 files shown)


Jako:
Closing this in favor of the 'original' #12989

OptimusCrime:
@Jako Okey, I could not find the pull request.

Jako:
@OptimusCrime The 'original' was not there when you created yours. But it was his find.

OptimusCrime:
Yeah, I understood that afterwards. It is fine :)

Issue #12966, opened 2016-04-19
Updating User Setting updates first setting in database


ishryal:
### Summary
When updating a user setting via Manage -> Users, the setting entry in the database that was entered first is updated regardless of the setting key or user id.

### Step to reproduce
1: Create User testuser. Provide full name, email, set to active, press save.
2: Once created, make a settings entry called test_id, name of Test ID, with value of 123 (all other fields default)
3: Go back to Manage -> User
4: Create another User testuser1. Provide full name email, set to active, press save.
5: Once created, make a setting entry called test_id, name of Test ID, with value of 456 (all other fields default).
6: Check database table modx_user_settings. There should be two entries with key of test_id with two different user ids.
7: In the user settings for testuser1, update setting test_id with new value of 789 (either via right click -> update setting or double clicking value in grid).
8: Check database table modx_user_settings again. The entry for testuser has been updated from 123 to 789 and NOT testuser1.
9: While in the user settings for testuser1, create a new setting called another_id, name of Another ID and value 987 (all other fields default).
10: Check database table modx_user_settings again. There will be new entry with key of another_id for testuser1.
11: Once grid has refreshed in user settings, update another_id setting with value 654.
12: Check database table modx_user_settings again. The entry 'test_id' for testuser has been updated to 654 and NOT the entry key 'another_id' for testuser1.

### Observed behavior
The wrong user and setting gets updated. This issue can also replicated with ONE user by creating a new user with two settings, then updating the second setting that was created. Result is that the first setting in the database is always updated.

### Expected behavior
The correct setting should be update for the correct user.

### Environment
MODX Revolution 2.4.4 and  2.4.3, apache 2.2.3, mysql 5.1.54, Firefox, PHP Version 5.3.5

ishryal:
Also occurs on MySQL 5.6.17 and PHP 5.5.12.

ishryal:
Also with fresh install of ModX 2.4.4 with no additional plugins, extras, code edit.

ishryal:
As per the forum post located at https://forums.modx.com/thread/100020/user-settings-bug#dis-post-540910, this issue is resolved by updating the core/model/modx/processors/security/user/setting/update.class.php to return true instead of parent::initialize();

Jako:
Maybe you should open a PR with that change in the forums. By closing this, the issue and the solution gets lost.

ishryal:
I would but it didn't let me for some reason.

Mark-H:
[...]

PR #12988, opened 2016-04-27
Updating user setting for the wrong user

Jako:
### What does it do?

Fixing updating user setting for the wrong user
### Why is it needed?

At the moment changing a user setting updates this for the first user found instead of updating the setting for the edited user.
### Related issue(s)/PR(s)

See #12966 and the discussion on https://forums.modx.com/thread/100020/user-settings-bug#dis-post-540910


BobRay:
+1 for this fix.  To identify a User Setting, you need both key and user ID. Calling the parent undoes this, since it gets the object with one argument.

Mark-H:
Thanks :)

PR #12990, opened 2016-04-28
fixed #12987

### LINKED ISSUE #12987: Fresh install 2.5.0 failed
### Summary
Fresh installation for 2.5.0 failed because `modx_system_settings` table was not created successfully.

### Step to reproduce
Fresh install

### Observed behavior
It went to 
```
500 Error

Site temporarily unavailable.
```
page.
The installation log:

```
[2016-04-27 17:30:46] (ERROR @ /var/www/html/***/www/core/xpdo/om/xpdoobject.class.php : 240) Error 42S02 executing statement: 
Array
(
    [0] => 42S02
    [1] => 1146
    [2] => Table 'modx_***.modx_system_settings' doesn't exist
)

[2016-04-27 17:30:46] (ERROR @ /var/www/html/***/www/core/xpdo/om/xpdoobject.class.php : 240) Error 42S02 executing statement: 
Array
(
    [0] => 42S02
    [1] => 1146
    [2] => Table 'modx_***.modx_system_settings' doesn't exist
)

[2016-04-27 17:30:46] (FATAL) Could not load core MODX configuration!
```

### Expected behavior
Should continue to summary page.

### Environment
MODX version 2.5.0,
Apache 2.0,
PHP Version 5.6.20-1+deb.sury.org~wily+1
mysqlnd 5.0.11-dev - 20120503

ghost:
### What does it do?

Modifed schemas in modx.mysql.schema.xml to have default values for NOT NULL date and date/time fields. Some datetime fields had a default string value that was rejected, changed them to an empty string.
### Why is it needed?

Installation fails when STRICT_TRANS_TABLES is configured
### Related issue(s)/PR(s)

Fixes #12987 

Added default values to various date fields set to be NOT NULL, changed some datetime fields to default to an empty string.
Needs to be tested by someone with STRICT_TRANS_TABLES configured.

### DIFF
--- a/core/model/schema/modx.mysql.schema.xml
+++ b/core/model/schema/modx.mysql.schema.xml
@@ -476,7 +476,7 @@
         <field key="xtype" dbtype="varchar" precision="75" phptype="string" null="false" default="textfield" />
         <field key="namespace" dbtype="varchar" precision="40" phptype="string" null="false" default="core" />
         <field key="area" dbtype="varchar" precision="255" phptype="string" null="false" default="" />
-        <field key="editedon" dbtype="timestamp" phptype="timestamp" null="false" attributes="ON UPDATE CURRENT_TIMESTAMP" />
+        <field key="editedon" dbtype="timestamp" phptype="timestamp" null="false" default="0" attributes="ON UPDATE CURRENT_TIMESTAMP" />
 
         <index alias="PRIMARY" name="PRIMARY" primary="true" unique="true">
             <column key="context_key" collation="A" null="false" />
@@ -670,7 +670,7 @@
         <field key="namespace" dbtype="varchar" precision="40" phptype="string" null="false" default="core" index="index" />
         <field key="language" dbtype="varchar" precision="20" phptype="string" null="false" default="en" index="index" />
         <field key="createdon" dbtype="datetime" phptype="datetime" />
-        <field key="editedon" dbtype="timestamp" phptype="timestamp" null="false" attributes="ON UPDATE CURRENT_TIMESTAMP" />
+        <field key="editedon" dbtype="timestamp" phptype="timestamp" null="false" default="0" attributes="ON UPDATE CURRENT_TIMESTAMP" />
 
         <index alias="name" name="name" primary="false" unique="false" type="BTREE">
             <column key="name" length="" collation="A" null="false" />
@@ -690,14 +690,14 @@
 
     <object class="modManagerLog" table="manager_log" extends="xPDOSimpleObject">
         <field key="user" dbtype="int" precision="10" attributes="unsigned" phptype="integer" null="false" default="0" />
-        <field key="occurred" dbtype="datetime" phptype="datetime" null="true" default="0000-00-00 00:00:00" />
+        <field key="occurred" dbtype="datetime" phptype="datetime" null="true" default="" />
         <field key="action" dbtype="varchar" precision="100" phptype="string" null="false" default="" />
         <field key="classKey" dbtype="varchar" precision="100" phptype="string" null="false" default="" />
         <field key="item" dbtype="varchar" precision="255" phptype="string" null="false" default="0" />
 
         <index alias="user_occurred" name="user_occurred" primary="false" unique="false" type="BTREE">
             <column key="user" length="" collation="A" null="false" />
-            <column key="occurred" length="" collation="A" null="false" />
+            <column key="occurred" length="" collation="A" null="false" default="" />
         </index>
 
         <aggregate alias="User" class="modUser" local="user" foreign="id" owner="foreign" cardinality="one" />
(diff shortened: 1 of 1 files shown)


OptimusCrime:
This is an important fix! Nice work

Jako:
Merged to 2.5.x branch.

OptimusCrime:
@Jako Did you forget to close this?

Jako:
No, leaving it open - can't check STRICT_TRANS_TABLES here.

OptimusCrime:
Ah, okey. But this is merged and ready to be tested?

Jako:
Wait for #12995

pixelchutes:
Why was this submitted / merged without being tested against MySQL Strict Mode being enabled _first_?

pixelchutes:
For me, with `explicit_defaults_for_timestamp = 0` in **my.cnf** (MySQL 5.6.30), `modx_manager_log` fails to be created due to [this line](https://github.com/modxcms/revolution/pull/12990/files#diff-f032a2068ea0a2229b531a77f448215bR693):

> ERROR 1067: Invalid default value for 'datetime'

**NOTE**: This is WITHOUT Strict Mode enabled.

pixelchutes:
I just tested building `2.x` HEAD against MySQL 5.6.30 w/ Strict Mode enabled.

**my.cnf**:
```bash
[mysqld]
sql-mode="STRICT_TRANS_TABLES"
```
I receive the same error on `setup/index.php?action=complete`:
> **500 Error**
> Site temporarily unavailable.

The following tables were NOT created:
- `modx_context_setting`
- `modx_lexicon_entries`
- `modx_system_settings`
- `modx_user_group_settings`
- `modx_user_settings`

pixelchutes:
The tables that weren't created all have `ON UPDATE CURRENT_TIMESTAMP`

This may be related: https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_explicit_defaults_for_timestamp

My testing has been done with `explicit_defaults_for_timestamp = 0`

Jako:
Maybe updating the timestamp value should be moved into the according update processor or in the save method of the xPDOObject class.

thewhiterabbit:
Having this issue with a new installation after downloading the 2.5.0 developer package today, May 3rd. Using STRICT_TRANS_TABLES mode in mySQL v5.7.10. I don't want to change my server setting to accommodate a single broken installation script for a single app. Any ETA on a fix being pulled for distro? Thanks in advance guys. Love this app and would love to get it working correctly again for security's sake.

Jako:
The important parts of the fix are already merged. The TIMESTAMP issues have to be solved somehow else. I am no SQL hero and don't know wether the deprecated clauses could be changed in SQL or have to be done in the processors or the xPDOObject.

thewhiterabbit:
If you can give me errors and line numbers I'll try to fix it. I suppose the errors are in the SQL log. I'll also look there and see if I can track it down.

Jako:
See https://github.com/modxcms/revolution/pull/12990#issuecomment-216565525

sottwell:
My apologies. I don't understand the issues well enough, I thought just giving those tables a default value would be sufficient.

Jako:
The value of the timestamp is updated to 'now' on saving a record with that ON UPDATE CURRENT_TIMESTAMP directive. Those directives are deprecated (maybe totally – the sentence in the MySQL docs is not understandable for me) and have be changed.

thewhiterabbit:
Ohh yes I had to wrk around this for another script I wrote. I ended up having to update the current timestamp field "manually" in php. This is a tough one because I am guessing multiple algorithms will need to be changed in xPDO class?

sottwell:
It looks like the problem here is the default being 0. Maybe try the default being CURRENT_TIMESTAMP instead.

sottwell:
http://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html

Jako:
That would be suitable for created records (i.e. for the `occured` column) but not for updated records (i.e. for the `editedon` column).

sottwell:
This is not on all tables.

thewhiterabbit:
No, wait according to the documentation @sottwell is correct. Setting a default value should fix the problem. I would use '0000-00-00 00:00:00' or NULL as the default and then set the table to ON UPDATE CURRENT_TIMESTAMP where applicable. That should work.

Jako:
See the note on the docs. If I am right, that issue occurs at the moment.

thewhiterabbit:
When checking to see if an editedon field has been updated I always check for NULL and if it's null then I know it hasn't been edited, then I don't display the edited date or mark, else I do and the date is set by ON UPDATE CURRENT_TIMESTAMP.

Jako:
@thewhiterabbit Maybe you could create some `create table` test php scripts and look what is working with your MySQL version.

thewhiterabbit:
@Jako The default value must be in the correct format for the field type for one thing, but also I found that the best practice is to assign a NULL value to the field rather than '0000-00-00 00:00:00'. I don't understand why mySQL prefers NULL, but it seems to work here on my server v5.7.10. I will be able to make amendments in the core, test, and add a pull request later today/tonight, unless someone beats me to it. If someone can shed more light on the exact reason v5.7.10 is preferring NULL default values over '0000-00-00 00:00:00' I would be very interested to know. It actually seems more logical for space saving because when you are storing NULL it's 4 char vs 19 char, and also the date '0000-00-00 00:00:00' can't logically exist for computers (because Unix epoch), so maybe that's the whole reason, I'm not sure.

thewhiterabbit:
Couldn't resist more testing. With strict mode enabled you must use NULL not "zero" values. So, the fix is to change the field to NULL as default value to be compatible with all versions of mySQL. As stated before, I will be able to amend the core later, other work requires me to take my attention away from this till later tonight.

thewhiterabbit:
I've changed the defaults for those affected datetime, timestamp, and date fields to NULL, which creates the tables just fine ,but now I've seemingly opened a can of worms.

I must return to my other work again, post lunch, so I'll pop in again later with another update.

We'll get it dialed in, I'm sure.

thewhiterabbit:
Alright. I hadn't opened any can of worms. When I cloned I hadn't noticed that the core.config.tpl somehow was renamed with .php.

As far as I can tell I've got the true fix now and I will submit a pull request after some more testing.

Jako:
#13001 is merged, hope the issue is gone now.

Issue #13004, opened 2016-05-05
Feature Request: Class Key and Item ID fields in manager action logs


matdave:
### Summary

The manager action logs have no way to search by a general class key or item id. 
### Step to reproduce

?a=system/logs try to find who has edited a specific chunk or page
### Observed behavior

Limited search functionality currently in manager actions
### Expected behavior

Would like more broken out functionality like this
![image](https://cloud.githubusercontent.com/assets/2213127/15047068/a7b7337a-12a8-11e6-8def-b10a36156312.png)

PR #13005, opened 2016-05-05
Improve Manager Log by adding a filter for class key and item

matdave:
### What does it do?

Adds an option to search by Class Key or Item ID to the manager logs
image
### Why is it needed?

It's incredibly hard to quickly see the history of who updated a specific element 
### Related issue(s)/PR(s)

Issue #13004


sottwell:
Looks good to me. +1

pixelchutes:
👍 Nice work! This is awesome.

OptimusCrime:
:+1:

amdbuilder:
:+1:

ilyautkin:
:+1:

Mark-H:
Hey folks, there's a new feature on GitHub where you can 👍 on the first message (or any other message). If you don't mind, using that instead of posting separate messages for each will help keep things less cluttered, especially when there's a feature or fix that needs more discussion or back-and-forth before it is merged ;) Thanks!

core/model/modx/processors/system/log/getlist.class.php
```diff
@@ -69,6 +71,8 @@ public function getData() {
         /* check filters */
         $wa = array();
         if (!empty($actionType)) { $wa['action:LIKE'] = '%'.$actionType.'%'; }
+        if (!empty($classKey)) { $wa['classKey:LIKE'] = '%'.$classKey.'%'; }
```
Mark-H: Does this have to be a LIKE? Are there situations where the class key isn't exactly what was chosen?

Jako: Maybe to search snippet instead of modSnippet

matdave: What Jako said, also just keeping with the same search standards of the other fields

OptimusCrime:
Any reason not to merge this? Should the `LIKE` be rewritten as equals instead? All in favor of merging this into the next minor release.

matdave:
The reason i really prefer `LIKE` for both class key and action is if you want to check all "save" actions across the board or just shorthand Chunk in search

OptimusCrime:
I'm in favor of this too

matdave:
*bump*

Issue #13003, opened 2016-05-05
Build error: Source file cache.manifest.php is missing


pixelchutes:
### Summary
During transport core / build process of a MODX Git installation, the following error is thrown:

> ERROR @ core/xpdo/transport/xpdovehicle.class.php (Line 362)
> Source file **manager/cache.manifest.php** is missing.

### Steps to reproduce

1. Checkout current HEAD of `2.5.x`
2. Perform MODX [Git installation prerequisites](https://rtfm.modx.com/revolution/2.x/getting-started/installation/git-installation)
3. Run `php _build/transport.core.php`
4. See the error after the "Packaged in web context" step.

### Observed behavior
Error is logged due to missing file.

> [2016-05-05 08:17:15] (ERROR @ core/xpdo/transport/xpdovehicle.class.php : 362) Source file manager/cache.manifest.php is missing

### Expected behavior
No errors thrown. Perhaps some missed references since #12985 is now merged?

### Environment
MODX 2.5.x-dev (git)

pixelchutes:
I think the fix is to simply remove [this reference](https://github.com/modxcms/revolution/blob/5324f9245a59c07abfa6f39c1d05673483cd89e5/_build/transport.core.php#L518-L522) from the build script.

OptimusCrime:
Damnit, I must have overlooked this when I removed the HTML5 references. I'll try to submit a new pull request as soon as I can. @pixelchutes solution should do it.

Jako:
Maybe a test for the build process should be added in `_build/test`

PR #13014, opened 2016-05-12
Fixing #13003

Jako:
### What does it do?

Remove required cache.manifest.php from build script.
### Why is it needed?

The file was removed during merging #12985
### Related issue(s)/PR(s)
#13003


OptimusCrime:
Very nice. Sorry about this.

Issue #13017, opened 2016-05-13
Issue: Parent Combo Select Breaks Save Process


matdave:
### Summary
If a parent resource has elements in it's page title, e.g. `<span> <strong> <sup>`, it breaks the resource save process with a Virtual Memory error.  The resource saves, but the saving screen never vanishes.

### Step to reproduce
Create a page with the title, `<span>Overproduced <sup>®</sup></span> Pagetitle`
Add a subpage and save the subpage.
The first save might work, but onces it's refreshed and you're in update mode try to save again.

### Observed behavior
The saving screen continues to spin without being able to complete.  I can confirm this on a 2.2.X site as well. 

### Expected behavior
It would be preferred if the parent title was stripped of elements before being displayed

### Environment
All versions affected.

Mark-H:
Fixed in #13018

PR #13018, opened 2016-05-13
Adding $this->modx->stripTags() to pagetitle combo box #13017

matdave:
### What does it do?

Added a strip tag process to the page title combobox from on a resource settings tab. 
### Why is it needed?

Save process doesn't stop spinning if a parent has elements in the name e.g. `<span>Overproduced <sup>®</sup></span> Pagetitle`
### Related issue(s)/PR(s)

Closes issue #13017


Mark-H:
Hm, I wonder if it might be better to throw some htmlentities at it rather than actually removing the tags?

matdave:
I removed the tags because that's how it will appear in the tree when you click to select another.

Mark-H:
Thanks @matdave, makes sense. Merged for inclusion in 2.5.1.

Issue #13012, opened 2016-05-11
extended user settings with quotes in the name


Jako:
### Summary
At the moment you could create a extended setting i.e. with quotes in the name. Those settings can't be edited/deleted after.

### Step to reproduce
Create an extended user setting with a quote in the name and try to edit/delete it.

### Expected behavior
The chars should be restricted for the name field (or htmlspecialchars should be used there).

### Environment
MODX 2.5

### Somehow Related
#12841

OptimusCrime:
This should not be too hard to fix

OptimusCrime:
Wait, has this already been fixed?

Jako:
Not in the name field, If I am right.

OptimusCrime:
I can not recreate this with the extended fields themselves, but the containers crash if they contain quotes. Is that what you meant?

OptimusCrime:
Never mind, now I see

Mark-H:
Fixed per #13019

PR #13019, opened 2016-05-13
Remove double quote from extended fields containers and attributes

OptimusCrime:
### What does it do?

Makes it impossible to add `"` (double quote) to the name of containers or attributes for users' extended fields.
### Why is it needed?

Because the name is used as a part of the id for the ext object, it is a part of the DOM. Adding something in the DOM with a `"` will break it. It does not work to escape it either (using "\"), so it has to be removed.
#### Tested:
- Adding new container with a quote in the name
- Renaming a container to contain a quote in the name
- Adding an attribute with a quote in the name
- Renaming an attribute to contain a quote in the name
### Related issue(s)/PR(s)
#13012
### Somehow Related
#12841


Mark-H:
Thanks @OptimusCrime :)

PR #13021, opened 2016-05-18
Add after(append) and before(prepend) output filters.

avanturist888:
### What does it do?

Add after(append) and before(prepend) output filters to able append or prepend values if output not empty. 
### Why is it needed?

Many times did this by separate snippets, maybe it should be built?

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -214,6 +214,7 @@ class modOutputFilter {
                         case 'append': /* appends the options value to the input value (if both not empty) */
                             if (!empty($m_val) && !empty($output))
                                 $output = $output.$m_val;
+                            break;
                         case 'before':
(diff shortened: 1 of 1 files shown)


Jako:
A `wrap` or `outer` output filter (combining both output filters) would be quite usable too: 

```
[[+string:wrap=`xxx|yyy`]]
```

vasia123:
Good idea! But which format would be better? 

    [[+string:wrap=`xxx|yyy`]]

 or:

    [[+string:wrap=`xxx||yyy`]]

Jako:
Maybe changeable by a system setting. Default should be `||` in my opinion. Similar to this (untested):

```
                        case 'wrap':
                        case 'outer': /* wraps the options value around the input value (if both not empty) */
                            if (!empty($m_val) && !empty($output))
                                $output = str_replace($this->modx->getOption('wrap_filter_delimiter',null,'||'), $output, $m_val);
                            break;

```

theboxer:
Merged to 2.x

Issue #12742, opened 2015-10-27
Resource icon in the tree ignores "isfolder" flag


whitebyte:
### Summary
It seems that resource icon in the tree ignores `isfolder` flag. Even if it's checked, resource shown as a regular one. Only actual presence/absence of children is taken into account. I think "folder" resource should be shown as "folder" icon even if it contains zero children. "isfolder" flag affects fURL suffix, so it's state should be clearly visible in the tree.

### Environment
2.4.2

pixelchutes:
Fixed in `v2.5.1`

PR #13027, opened 2016-05-24
Extended selector to show folder icon for empty containers

achterbahn:
### What does it do?

Extended selector that resources marked as container but without children show the container icon
### Why is it needed?

Even resources without children are containers. This should be reflected by the icon in the tree
### Related issue(s)/PR(s)

Fix for #12742

Ps. only the .scss file was changed. Please process to file to css.


Mark-H:
Thanks @achterbahn - merged into 2.5.x for inclusion in 2.5.1!

PR #13029, opened 2016-05-24
Fix #12910 - Duplicate button is missing when editing Weblinks

### LINKED ISSUE #12910: Duplicate button is missing when editing Weblinks
### Summary
When editing a Weblink Resource  (`modWeblink`), the <kbd>Duplicate</kbd> button is missing.

### Step to reproduce
1. Login to MODX Manager _(Administrator role is fine)_
2. Edit an existing Weblink Resource.
3. Look at the top row of buttons.

### Observed behavior
<kbd>Duplicate</kbd> button is missing.

### Expected behavior
<kbd>Duplicate</kbd> button should be present given proper permissions.

To compare, edit a normal Resource (`modResource`) and you'll see <kbd>Duplicate</kbd> next to <kbd>Save</kbd> button.

### Environment
MODX 2.4.3

labr1005:
### What does it do?

Added the missing checks for user rights "resource_duplicate" and "new_document_in_root" in the specific controllers. Those settings are passed to "revolution/manager/assets/modext/sections/resource/update.js" which adds the button when the user passes these checks.

I also did some cleanup on the code, the function $this->modx->hasPermission(...) is not needed here as it already runs in "revolution/manager/controllers/default/resource/resource.class.php".
### Why is it needed?

Duplicate button is missing when editing Weblinks and Symlinks.
### Related issue(s)/PR(s)
#12910

### DIFF
--- a/manager/controllers/default/resource/symlink/update.class.php
+++ b/manager/controllers/default/resource/symlink/update.class.php
@@ -33,9 +33,11 @@ class SymlinkUpdateManagerController extends ResourceUpdateManagerController {
                 ,locked: '.($this->locked ? 1 : 0).'
                 ,lockedText: "'.$this->lockedText.'"
                 ,canSave: '.($this->canSave ? 1 : 0).'
-                ,canEdit: "'.($this->modx->hasPermission('edit_document') ? 1 : 0).'"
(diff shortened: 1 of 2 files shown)


Mark-H:
Nice find! Merged into 2.5.x for inclusion in 2.5.1.

Issue #12363, opened 2015-02-13
Amazon S3 Buckets with Dots in Name Cannot Be connected to as Media Sources


aesop1:
Modx doesn't connect to Amazon S3 Buckets with dots in the bucket name. Amazon requires some buckets to match domains (requiring dots) when a user wants to match the bucket to a particular domain. This issue has been reported before, but it keeps on being closed without any proof (that I can see) of resolution.

Here are two of the closed reports:
https://github.com/modxcms/revolution/pull/12034
https://github.com/modxcms/revolution/issues/12017

aesop1:
Verified that when the following was commented out, the S3 Media Source bucket (with dots) started working:
 public function setBucket($bucket) {
         $this->bucket = $bucket;
//        if (strpos($bucket,'.') !== false) {
//            $this->driver->enable_path_style(true);
//        }

Omeryl:
Hmm, can you show me your bucket URL? At the time I did that change it worked, but AWS might've changed something since then.

aesop1:
//navisweb.meetnavis.com.s3-website-us-west-2.amazonaws.com

adamwintle:
I'm also getting this issue with bucket: assets.domain.co.uk

adamwintle:
@opengeek thanks for looking at this and marking it as priority.  For anyone interested I added some notes about this bug here: https://github.com/modxcms/revolution/issues/12017#issuecomment-170549972

PR #13031, opened 2016-05-25
Fix buckets with dots

theboxer:
### What does it do?

Fixes displaying files from S3 buckets with dots. New region property were added to S3 media source.
### Why is it needed?

Because it doesn't work otherwise :)
### Related issue(s)/PR(s)
#12034

👏 to @garryn for finding the solution!


Mark-H:
Also related #12363

PR #13034, opened 2016-05-27
Add sortBy and sortDir for initial grid config

13hakta:
### What does it do?

Added sortBy and sortDir for initial grid config.
### Why is it needed?

These options as said in RTFM are used for grids unconditionally, but implemented only for grouped grid.


Jako:
The default sorting should be set in the getlist processor of that grid. All other data handling like filtering, limiting etc. is done there.

PR #13035, opened 2016-05-27
Added two missing fields

Jako:
### What does it do?

Add two missing fields to the object
### Why is it needed?

Remove code inspect warnings


theboxer:
Merged to 2.5.x

PR #13037, opened 2016-05-27
Fix server port check in setup start script

MaartenW:
### What does it do?

Adds checks to the `$_SERVER['SERVER_PORT']` setting.
### Why is it needed?

If the server port is not passed or is an empty string (NGINX) then the redirect is wrong.


Jako:
I can't check this issue (no NGINX on my development host – maybe another one could jump in). But maybe this check looks a bit better (It should check the same things (assuming $_SERVER['SERVER_PORT'] can't be 0 or '0').

```
if (isset($_SERVER['SERVER_PORT']) && !empty($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] != 80)
```

PR #13038, opened 2016-05-29
Fixing padding-top issue in MODx.Window

Jako:
### What does it do?

Fixing a padding-top issue in .modx-window .x-window-body
### Why is it needed?

Since #12973 there is no reduced padding on the first label and the padding-top is too large there at the moment. The issue is visible i.e. during editing a system setting:

screen

I hope the side effects are low, since there could be a .modx-window that does not start with a label. The ones I have checked in the pages of the core starts with a label. I will look for side effects in the next days on my develop installation.
### Related issue(s)/PR(s)
#12973

### DIFF
--- a/_build/templates/default/sass/_windows.scss
+++ b/_build/templates/default/sass/_windows.scss
@@ -103,6 +103,9 @@
     overflow-y: auto;
     padding: 15px;
   }
+  &.modx-window .x-window-body {
+    padding-top: 0;
+  }
 
   .x-panel-bwrap {
     background: $winBodyBg;


OptimusCrime:
I'll take a look at this. I could not find any other places this would be applied, but apparently there was. Sorry about that.

theboxer:
Merged to 2.5.x

pixelchutes:
This PR seems to have a small/trivial impact on console window:

### BEFORE:

console_before

### AFTER:

console_padding

Jako:
Ok, then the console area needs some padding top.

PR #12999, opened 2016-05-04
Deprecated

devpastetOLD:
### What does it do?

Describe the technical changes you did.
### Why is it needed?

Describe the issue you are solving.
### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request (see https://github.com/blog/1506-closing-issues-via-pull-requests)

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; phpthumb has a deprecated constructor

PR #13039, opened 2016-05-29
Update phpThumb to version 1.7.14 

Jako:
### What does it do?
- Update the phpThumb classfiles from 1.7.13-201406261000 to 1.7.14-201604151303
- Changed the encoding of some original files to UTF-8 (they still seem to be Windows Latin 1)
- Repaired some previously wrong encoded chars
### Why is it needed?

Have the phpThumb class uptodate. The main reason is PHP7 compatibility.
### Related issue(s)/PR(s)
#12810, #12797, #12999

### DIFF
--- a/core/model/phpthumb/phpthumb.class.php
+++ b/core/model/phpthumb/phpthumb.class.php
@@ -12,7 +12,7 @@
 ob_start();
 if (!include_once(dirname(__FILE__).'/phpthumb.functions.php')) {
 	ob_end_flush();
-	die('failed to include_once("'.realpath(dirname(__FILE__).'/phpthumb.functions.php').'")');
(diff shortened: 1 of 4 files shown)


Jako:
Please check this commit on as many installations as possible …

opengeek:
@Jako does this PR replace #12810 ?

Jako:
@opengeek: yes

OptimusCrime:
What is the difference? The phpThumb versions are identical?

Jako:
The versions are not 'that' identical. My patch uses 1.7.14-201604151303 and the earlier one 1.7.14-201512071121.

OptimusCrime:
Oh. My bad. Sorry @Jako.

Adding build dates the version number...

Jako:
Yep. Strange …

Issue #13043, opened 2016-06-01
Parser issue with nested elements


christianseel:
### Summary

There seems to be a bug in the MODX parser with a special nested MODX tag structure. Issue exists if you have nested chunk calls where the chunk name contains another placeholder like `[[$chunk-[[*context_key]]]]`.
### Step to reproduce
- create a chunk called `chunk-1` and paste the following content:

```
<div class="wrapper">
    <span>Level: [[+level:default=`not set`]]</span>
    [[+nestedcontent]]
</div>
```
- edit the resource with the ID 1 (can be any other resource but then update the name of the chunk with the resource ID)
- fill the following code into the content field:

```
[[$chunk-[[*id]]?
        &level=`1`
        &nestedcontent=`
            [[$chunk-[[*id]]?
                &level=`2`
                &nestedcontent=`
                    [[$chunk-[[*id]]?
                        &level=`3`
                        &nestedcontent=`
                            this is level 3. and here it breaks!
                        `
                    ]]
                `
            ]]
        `
    ]]
```
### Observed behavior

MODX parser breaks at the 3rd level and returns wired output with broken placeholder/tags like

```
<div class="wrapper">
<span>Level: 1</span>
`
            [[$chunk-1?
                &level=`2`
                &nestedcontent=`
                    [[$chunk-1
</div>

</div>
```

If you remove the most inner chunk call (3rd level) and just leave the other 2 levels it works as expected.

Also if you replace the `[[*id]]` placeholder in the chunk name with `1` it works as expected.
### Expected behavior

Correct output with 3 `<div class="wrapper">` and a text `this is level 3. and here it breaks!` in the inner div.
### Environment

Tested with MODX 2.5.0-pl and 2.4.2-pl. Also PHP 5.6 and 7.0. Both Apache.
No special templates, no ContentBlocks…

OptimusCrime:
Do you have pdoTools installed?

christianseel:
Hey, nope. But I might be close to the issue in the source codes. Looks like it's related to `$depth` for `processElementTags()`. Seems like it's using the default value `0` for chunks instead of the system setting or default value of `10` which is used at most other function calls.

christianseel:
Changing https://github.com/modxcms/revolution/blob/2.5.x/core/model/modx/modparser.class.php#L431 to something like
```
$maxIterations= intval($this->modx->getOption('parser_max_iterations',null,10));
[...]

PR #13044, opened 2016-06-01
Make sure to call processElementTags with correct $maxIterations value

christianseel:
### What does it do?

Call `processElementTags()` in the modParser always with the `$maxIterations` / `$depth` value from the `parser_max_iterations` system setting (default 10), instead of the "function default" of 0.
### Why is it needed?

Fixes a parser bug for special nested tags. See #13043 for steps to reproduce the bug.
### Related issue(s)/PR(s)

Directly related to the issues reported in #13043

I've tested it already on 1 production site where it fixed my issue from #13043 and didn't trigger any other issues so far.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -428,7 +428,8 @@ class modParser {
         }
 
         /* collect any nested element tags in the innerTag and process them */
-        $this->processElementTags($outerTag, $innerTag, $processUncacheable);
+        $maxIterations= intval($this->modx->getOption('parser_max_iterations',null,10));
+        $this->processElementTags($outerTag, $innerTag, $processUncacheable, $this->modx->parser->isRemovingUnprocessed(), "[[", "]]", array (), $maxIterations);
         $this->_processingTag = true;
         $outerTag= '[[' . $innerTag . ']]';
 
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -429,7 +429,7 @@ class modParser {
 
         /* collect any nested element tags in the innerTag and process them */
         $maxIterations= intval($this->modx->getOption('parser_max_iterations',null,10));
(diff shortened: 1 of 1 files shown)


christianseel:
I'm not sure but I guess the travis tests are also incorrect since they are using `$depth = 0`?

I'd say
```
[[!doNotCacheMe? &but=`[[youCanCacheMe]]`]]
```
can't be correctly parsed with `$depth = 0` – correct?

opengeek:
Can you update this PR to point at the 2.x base branch @christianseel ?

opengeek:
@christianseel I have merged this but if you have a chance before I get to it, can you look into fixing the tests you broke with this change?

christianseel:
@opengeek unfortunately I have no idea how to fix the travis tests. Never worked with those before…

opengeek:
No problem @christianseel — I don't understand those particular tests anyway, so I'm just commenting them out for now.

opengeek:
This PR caused some unexpected behavior that I am afraid to include in the 2.5.6 release without a better understanding of what is going on. I have reverted these changes for now.

christianseel:
@opengeek any information what unexpected behavior your saw? We got this running in one of our installations without issues…

OptimusCrime:
Not sure if I understand the problem either. It just sets the correct value for maxDepth?

christianseel:
Exactly! And that fixes the issues described in https://github.com/modxcms/revolution/issues/13043

opengeek:
First, it broke the basic unit tests. I had some additional issues in my local environment using this. It was producing unexpected results as it changed the way things work. That is a breaking change. I will re-review it for 2.6, but changes like this CANNOT go into a patch release.

OptimusCrime:
It can not (as I see) be added for 2.6 without breaking changes? So it would either need to be considered a bug fix (and sneak it in for 2.6), or a feature and add it for 3.0.

Issue #13048, opened 2016-06-07
Installer Button


EkoOne:
Installer "Continue" button text not centered. See screenshot.

Revo 2.5.0

![installer](https://cloud.githubusercontent.com/assets/1754223/15873539/caa76c02-2cc4-11e6-97c6-2e071f5c4fd5.png)

meshkov:
Confirm
![2016-06-08_14-06-09](https://cloud.githubusercontent.com/assets/1703670/15892418/3428877a-2d82-11e6-8330-25b50d3d8942.png)

Pathologic:
I've fixed that by changing  https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/workspace/package.panels.js#L152 to:
```
            Ext.getCmp('package-install-btn').autoWidth = true;
            Ext.getCmp('package-install-btn').setText(_('continue')).syncSize();
```

Jako:
@Pathologic Thanks for your solution. I have created a slightly modified PR with that.

Pathologic:
Thanks for quick response, @Jako

PR #13057, opened 2016-06-24
Fixing #13048

Jako:
### What does it do?

Add autoWidth/autoHeight to the button and add a .syncSize() after changing the button text.
### Why is it needed?

Fixing the displayed width of the button.
### Related issue(s)/PR(s)
#13048


pixelchutes:
Not sure if it's 100% related or not yet, but just noticed this issue when testing "Setup Options" button in a package on `modx-2.5.x`:

setup_options

PR #13067, opened 2016-07-08
move OnFileManagerBeforeUpload event call

Pathologic:
### What does it do?

This PR changes the time of invoking OnFileManagerBeforeUpload event. 
### Why is it needed?

It allows more flexible pre-processing of uploaded files: skip files by any critetia or change file format, file name and so on. It also fixes this:

```
/* invoke event */
        $this->xpdo->invokeEvent('OnFileManagerBeforeUpload',array(
            'files' => &$objects,
            'file' => &$file,
            'directory' => $container,
            'source' => &$this,
        ));
// $newPath is set before event call, so nothing to do with file name in OnFileManagerBeforeUpload plugins
            if (!move_uploaded_file($file['tmp_name'],$newPath)) {
                $this->addError('path',$this->xpdo->lexicon('file_err_upload'));
                continue;
            }
        }
```

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -706,6 +706,14 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
 
         /* loop through each file and upload */
         foreach ($objects as $file) {
+            /* invoke event */
+            $this->xpdo->invokeEvent('OnFileManagerBeforeUpload',array(
+                'files' => &$objects,
+                'file' => &$file,
+                'directory' => $container,
+                'source' => &$this,
+            ));
+            
             if ($file['error'] != 0) continue;
             if (empty($file['name'])) continue;
             $ext = pathinfo($file['name'],PATHINFO_EXTENSION);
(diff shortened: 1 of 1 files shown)


sdrenth:
Looks good to me!

OptimusCrime:
Simple fix. +1. Note that this is a breaking change, so it will need to target 2.x (actually 3.x, but I think we can skip this here?)

Mark-H:
@OptimusCrime Could you elaborate on the breaking changes that you see? I don't immediately see any use cases that would be broken; it's still happening before the upload, the $newPath processing wasn't available in plugins anyway.. The only thing I can come up with that could break is that people may have been using it to change `$file['name']`, and the extension for that is now checked. If you can think of anything else, that would be good to know before this gets merged.

OptimusCrime:
You are right. I did not inspect the code closely enough. I thought the old and new time of invoking the event involved variables that were passed by reference. Now that I look at the code it is obvious that you are right. The difference should not matter.

Mark-H:
Thanks @OptimusCrime - I don't see any potential harm with the $file variable. 

Thank you @Pathologic, merged for 2.6.

Issue #13063, opened 2016-07-05
hidemenu label doesn't change using Form Customization. 


jaygilmore:
### Summary
`hidemenu` label doesn't change using Form Customization. 

### Step to reproduce

1. Create Form Customization Rule Set for Create Resource or Update Resource
2. Change the label value for `hidemenu`
3. Go create or update (depending on which you picked in step 1)
4. Note the label is still set for the default lexicon value of *Hide From Menus*

### Observed behavior
The `hidemenu` label is still set for the default lexicon value of *Hide From Menus*

### Expected behavior
Label should reflect updated Form Customization setting

### Environment
Revo 2.5.0 nginx, percona. MODX Cloud

Mark-H:
Duplicate of #13053 and #12890 and fixed in #13025 in 2.5.1.

theboxer:
@Mark-H Changelog entry is missing for this fix

jaygilmore:
Bah. I searched issues but I must have left the filters on. Thanks.

Issue #13053, opened 2016-06-12
FC's new name don't work for hidemenu


Realetive:
### Summary
If setup new label name for hidemenu field, this isn't use. If this setup for published it's ok.

### Step to reproduce
Create new Form Customization rule and try to setup new label name for hidemenu. 

### Observed behavior
Nothing to changes.

### Expected behavior
Checkbox label need to equal new label name from CMP

### Environment
MODX v2.5.0-pl

Jako:
I think you mean FC (Form Customization) instead of CMP (Custom Manager Page).

Realetive:
@Jako sorry, yes, you are right.

Mark-H:
Dupe of #12890 and fixed in #13025 for 2.5.1

Realetive:
Many thanks @Mark-H . Good news.

PR #13025, opened 2016-05-24
Solve #12890 - fixed form customization for checkbox labels

labr1005:
### What does it do?

Removed check for fieldname "published".
### Why is it needed?

Labels for checkboxes could not be overridden via form customization.
### Related issue(s)/PR(s)
#12890

Issue #12890, opened 2016-02-02
hidemenu custom label in Manager Form Customization does not work


madze:
### Summary
Custom Labels for the 'hidemenu' item in Manager Form Customization do not change the label.

### Step to reproduce
Add a custom label for the 'hidemenu' item and then view that input in the manager while logged into the corresponding user group.

### Observed behavior
The label still reads: "Hide From Menus" instead of the custom label.

### Expected behavior
Should be replaced with the custom label.

### Environment
REVO  2.4.0-pl

Mark-H:
Fixed per #13025 in 2.5.1.

PR #13068, opened 2016-07-10
Modify MODx.renameLabel to handle hidemenu field

sepiariver:
### What does it do?

`manager/assets/modext/core/modx.js` line 397 added check for "hidemenu" in `flds` argument to fetch different Cmp object and use different method for changing the field label.
### Why is it needed?

Before this patch, the hidemenu would not conform to Form Customization field label changes. Steps to reproduce:
1. Install MODX
2. Make a FC rule to change the hidemenu field label.
3. Make another FC rule to change the adjacent published field label.
4. Load the default Resource Edit View.

The hidemenu label would not be changed but the published field label would conform to the custom labels.

Why would this be? I checked the files in `manager/assets/modext/widgets/resource/` and couldn't find anything different about how the hidemenu widget is implemented, versus the published widget. I checked the DOM and they render similarly. After a couple hours of digging around I went for a patch. 
### Related issue(s)/PR(s)

n/a

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -65,7 +65,7 @@ Ext.extend(MODx,Ext.Component,{
      */
     ,add: function(cmp) {
         if (typeof cmp === 'string') {
-            cmp = { xtype: cmp }
+            cmp = { xtype: cmp };
         }
         var ctr = Ext.getCmp('modx-content');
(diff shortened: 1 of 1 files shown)


Mark-H:
Related issues #13063 #13053 #12890 and already fixed for 2.5.1 since #13025

PR #13084, opened 2016-07-26
Duplicate array index, and description is missing

enigmatic-user:
### What does it do?

Removes a duplicated phrase and adds a missing one (translation of the English version).
### Why is it needed?

The needed second phrase can't be added via crowdin or in any other way.

The second $_lang['setting_anonymous_sessions'] should be $_lang['setting_anonymous_sessions_desc'] with a different text. While this is correct in the 2.x branch, it isn't in the 2.5.x branch for some reason, and the language file delivered with MODX 2.5.1-pl contains the wrong version, too. This is why I corrected it this way, not via crowdin (which is impossible).

As I just saw, it's the same in the other language files (except the English one)! Sorry, I don't have the time to fix them all...


enigmatic-user:
$_lang['set_edit] is missing in the other translations in the 2.5.x branch, too.

enigmatic-user:
I guess $_lang['setting_manager_html5_cache'] and $_lang['setting_manager_html5_cache_desc'] (setting.inc.php) can be removed in the 2.5.x branch? It's gone in the 2.x branch and the English language file in the 2.5.x branch.

rtripault:
ping @theboxer

theboxer:
Looks like everything is in crowdin now. Ping me if any other issue appear.

PR #13088, opened 2016-08-08
package.grid.js // default_per_page items in Package-Overview-List

sebastian-marinescu:
### What does it do?

Updated package.grid.js, so that it not statically always show 10 entries, 
but acknowledges the _default_per_page_ system-setting the user provided, 
if it is larger than 10.
### Why is it needed?

The Package-Overview statically shows 10 entries. 
I was expecting it to behave as all the other grids in the manager.
### Related issue(s)/PR(s)
#12518 Default Per Page number of packages listed not related to the Default Per Page System Setting
#12613 Add "Updatable" Column to packages grid
#12601 [Request] Updatable Packages first in Package Manager
#8906   Set amount of list grid-items in system settings

### DIFF
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -83,7 +83,7 @@ MODx.grid.Package = function(config) {
                  ,'provider','provider_name','disabled','source','attributes','readme','menu'
                  ,'install','textaction','iconaction','updateable']
         ,plugins: [this.exp]
-        ,pageSize: MODx.config.default_per_page > 10 ? MODx.config.default_per_page : 10
+        ,pageSize: parseInt(MODx.config.default_per_page) > 10 ? parseInt(MODx.config.default_per_page) : 10
         ,columns: cols
         ,primaryKey: 'signature'
         ,paging: true
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -83,7 +83,7 @@ MODx.grid.Package = function(config) {
                  ,'provider','provider_name','disabled','source','attributes','readme','menu'
                  ,'install','textaction','iconaction','updateable']
         ,plugins: [this.exp]
-        ,pageSize: 10
+        ,pageSize: MODx.config.default_per_page > 10 ? MODx.config.default_per_page : 10
         ,columns: cols
         ,primaryKey: 'signature'
         ,paging: true


OptimusCrime:
Why not just use the `default_per_page` setting? Why do you need to check that it is larger than 10? Is that default behavior in the other grids?

sebastian-marinescu:
I saw it done like this in the settings-grid (with a check for 30 - [see code](https://github.com/modxcms/revolution/blob/e1d0f054d1ac5431e93008aaca517f93a39667ee/manager/assets/modext/widgets/core/modx.grid.settings.js#L154)) and thought it would also be good for the package-browser to have a minimum amount. But I think I should also `parseInt` it.

sebastian-marinescu:
Also on others I saw something like this as value: 

```js
config.pageSize || MODx.config.default_per_page || 20
```
But I don't know and didn't check if `config.pageSize` exists or where it's coming from. 
Does someone know this by any chance?

sebastian-marinescu:
Updated **Related issue(s)/PR(s)** – seems as if this has been an issue for almost 4 years.

sottwell:
I'm sure there are other grids in the Manager that are still hard-coded. They pop up every now and then.

Mark-H:
I seem to recall this being intentional because it hits up a webservice and they didn't want any potentially high limit being used there?

sebastian-marinescu:
I also saw other Grids behaving individually, but can't remember any but the settings-grid right now.

@sottwell: My PR was actually just for the Package-Overview, but we could also gather all the places where this is inconsistent and decide how to deal with it (although I think opening a separate Issue/PR for that would be better).

@Mark-H: Is the high limit still a thing? And how do you feel about it? 

If performance is still an issue, how do you all feel about a separate setting just for the package-browser-grid (which is constraint to a max value) or maybe another hardcoded `pageSize` for that particular grid?

Jako:
@Mark-H If the web service is the bottleneck, it should be enhanced to give an answer for multiple version requests. That way the number of requests could be dramatically reduced. Until that we should leave the limit untouched.

pixelchutes:
I remember the same as @Mark-H , however I wonder if recent updates to the Extras search API on modx.com can perhaps remove this intentional limit?

Jako:
Even if the improvements to the Extras API allow to set a higher limit, it would be better to add a new endpoint that retrieves more than one version information with a single request, if that is possible.

Jako:
But browsing the list of the installed extras would create a version request to the provider for each single extra, if I am right.

Jako:
Somehow related to #12601 and #12613

Mark-H:
That's a good point, it does indeed check for updates in the request there..

sebastian-marinescu:
Any news on an alternative API endpoint for multiple version-checks?

Mark-H:
I believe this was fixed with #13348.

PR #13097, opened 2016-08-22
Allow manually editing rank-order of contexts

sebastian-marinescu:
### What does it do?

Add a numberfield for the rank-value of a context: 

image
### Why is it needed?

Right now this setting can only be seen in the DB, 
it would be convenient to also see it were expected.
### Related issue(s)/PR(s)

not sure - will update if found

### DIFF
--- a/manager/assets/modext/widgets/system/modx.panel.context.js
+++ b/manager/assets/modext/widgets/system/modx.panel.context.js
@@ -57,7 +57,7 @@ MODx.panel.Context = function(config) {
 					xtype: 'numberfield'
 					,fieldLabel: _('rank')
 					,name: 'rank'
-					,allowBlank: false
+					,allowBlank: true
 					,width: 300
 				},{
 					html: MODx.onContextFormRender
--- a/manager/assets/modext/widgets/system/modx.panel.context.js
+++ b/manager/assets/modext/widgets/system/modx.panel.context.js
@@ -54,6 +54,12 @@ MODx.panel.Context = function(config) {
 					,width: 300
 					,grow: true
 				},{
+					xtype: 'numberfield'
+					,fieldLabel: _('rank')
+					,name: 'rank'
+					,allowBlank: false
+					,width: 300
+				},{
 					html: MODx.onContextFormRender
 					,border: false
 				}]


jpdevries:
Sweet. Not to change the topic, but on the note of things that are in the DB but not exposed in the UI I noticed last week that there is a photo column in the user attributes table but that doesn't seem to be exposed in the Manager.

rtripault:
> that doesn't seem to be exposed in the Manager.

Have a closer look in `?a=security/user/update&id={modUser.id}` & `?a=security/profile` ([Revo 2.4+](https://github.com/modxcms/revolution/blob/2.x/core/docs/changelog.txt#L227))

Otherwise, :+1: for the PR

sebastian-marinescu:
Thanks @jpdevries and @rtripault 

I took a look – the `user_photo` is exposed as `modx-combo-browser` in both places ("Benutzerfoto"): 

image

It's just not really prominent and there is no picture-preview, so..... done :smile:

manager/assets/modext/widgets/system/modx.panel.context.js
```diff
@@ -54,6 +54,12 @@ MODx.panel.Context = function(config) {
 					,width: 300
 					,grow: true
 				},{
+					xtype: 'numberfield'
+					,fieldLabel: _('rank')
+					,name: 'rank'
+					,allowBlank: false
```
rtripault: I guess we could allow a blank value here, since it will default to `0`

rtripault:
Should the rank field be added to the [create window](https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/system/modx.grid.context.js#L124) too ?
We might also want to enable sorting via drag & drop from the grid (but that's a more complex task, which could be addressed later)

manager/assets/modext/widgets/system/modx.panel.context.js
sebastian-marinescu: It's defaulting to `0` because it's a numberfield?

rtripault: No, because of the [DB schema](https://github.com/modxcms/revolution/blob/2.x/core/model/schema/modx.mysql.schema.xml#L451)

sebastian-marinescu:
I think adding it to the _create window_ wouldn't hurt – if only it would make this more consistent.
And sorting via drag & drop _could_ be interesting, but _could_ also be annoying if you did drag it by accident and the numbers change other than expected – but like you said this could be addressed later. 
Now with this PR I could add the `rank`-field to the context-grid as column and try to make it inline-editable: what do you think?

manager/assets/modext/widgets/system/modx.panel.context.js
sebastian-marinescu: Thanks for that!

rtripault:
I guess it makes sense to add the `rank` column indeed.

theboxer:
Added rank field to create window & context grid with ability to change it directly from grid

Issue #12826, opened 2015-12-11
Error HY000 executing statement


thomay:
### Summary
* modResource::getTemplateVarCollection fails using mysql server 5.7.9
* unclear if xPDO or modx issue

### Step to reproduce
* Install modx 2.4.2 using mysql server 5.7.9
* fail found using getResources with processing of template variables

### Observed behavior
* Template variables were not processed (empty)
* Errorlog shows:

```
(ERROR @ /index.php) Error HY000 executing statement: 
Array
(
    [0] => HY000
    [1] => 3065
    [2] => Expression #1 of ORDER BY clause is not in SELECT list, references column '$$$$.tvtpl.rank' which is not in SELECT list; this is incompatible with DISTINCT
)
```

### Expected behavior
* no error in error log
* processed template variables

### Environment
* MODX Revolution 2.4.2-pl (traditional)
* ExtJS-Version: 3.4.1
* Smarty-Version: 3.0.4
* PHPMailer-Version: 5.2.13
* MagpieRSS-Version: 0.72
* MySQL Server Version: 5.7.9
* mysql pdo driver Client API version: 5.5.46
* PHP Version: 5.6.16-2

### Notes
The issue results from the sortby statement `$c->sortby('tvtpl.rank,modTemplateVar.rank');` used in conjunction with the DISTINCT query. `tvtpl.rank`is used to sort the result set but is not part of the SELECT statement.
Additionally selecting `tvtpl.rank` seems to fix the issue from within getTemplateVarCollection, but I am not sure that this fix belongs here. However, it fixes the issue for me. Find my fix below as a patch file:
```
diff -rc modx-2.4.2-pl/core/model/modx/modresource.class.php modx-2.4.2-pl_patched/core/model/modx/modresource.class.php
*** modx-2.4.2-pl/core/model/modx/modresource.class.php 2015-10-06 13:06:16.000000000 +0200
--- modx-2.4.2-pl_patched/core/model/modx/modresource.class.php 2015-12-11 12:26:28.193976999 +0100
***************
*** 360,365 ****
--- 360,366 ----
                  'tvc.contentid' => $resource->get('id'),
              ));
          }
+       $c->select(array('tvtpl.rank'));
          $c->sortby('tvtpl.rank,modTemplateVar.rank');
          return $resource->xpdo->getCollection('modTemplateVar', $c);
      }
```

pixelchutes:
Sort of related to #12401, in that it is a MySQL 5.7 support issue.

[...]

PR #13098, opened 2016-08-22
Fixing #12826 MySQL 5.7 Error ORDER BY clause is not in SELECT list

Jako:
### What does it do?

Add a SELECT clause for the ORDER BY clause because of the use of DISTINCT
### Why is it needed?

Fixing a MySQL 5.7.+ error:
`ORDER BY clause is not in SELECT list … this is incompatible with DISTINCT`
### Related issue(s)/PR(s)
#12826


IvanTheFirst:
Thank you, this is almost fix my problem!

tjventurini:
Looking good, I think this did the trick!

OptimusCrime:
Please merge this!

Issue #10793, opened 2010-04-16
Duplicate box should also include the Caption field


xpix:
xpix created Redmine issue ID 10793

Hi, When duplicating a TV the box should also contain a caption field. If a template has 5 similar text TVs the caption box will help a lot as you do not have to click on the TV and change the caption. Maybe this can be extended to Snippets, Plugins, Templates but is not as important as for TVs.
Thx

Alroniks:
Hooray! I did it! :)

Alroniks:
@OptimusCrime Hi, I think we can close this issue as PR was merged.

PR #13100, opened 2016-08-23
Duplicating caption field for TVs

alroniks:
### What does it do?

This improvement allows change caption of TV during duplicate it.
### Why is it needed?

It can be useful when you start duplicates one based TV to new and after that, you will get a mess on resource page because all TVs will have the same captions. This change allows avoid it.
### Related issue(s)/PR(s)

https://github.com/modxcms/revolution/issues/10793

### DIFF
--- a/core/model/modx/processors/element/tv/duplicate.class.php
+++ b/core/model/modx/processors/element/tv/duplicate.class.php
@@ -140,7 +140,7 @@ class modTemplateVarDuplicateProcessor extends modElementDuplicateProcessor {
         $caption = $this->getNewCaption();
         $this->setNewCaption($caption);
 
-        return parent::beforeSet();
+        return parent::beforeSave();
     }
 }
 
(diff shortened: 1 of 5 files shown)


Alroniks:
@OptimusCrime ^^^

Jako:
Merged into 2.5.x. Thanks!

core/model/modx/processors/element/tv/duplicate.class.php
```diff
@@ -104,5 +105,43 @@ public function duplicateMediaSources() {
...
+    /**
+     * Get the new caption for the duplicate
+     * @return string
+     */
+    public function getNewCaption()
+    {
+        $caption = $this->getProperty($this->captionField);
+        $newCaption = !empty($caption)
+            ? $caption
+            : $this->modx->lexicon('duplicate_of', array('name' => $this->object->get($this->captionField)));
```
rtripault: When there is no caption (in example because we removed it with the UI when duplicating and the source has no caption), this forces a caption of `Duplicate of `.

PR #13101, opened 2016-08-26
Default value for the "unauthorized_page"

sergant210:
### What does it do?

Sets the default value of the "unauthorized_page" to the "site_start" setting instead of '401'.
### Why is it needed?

It's not easy task to create resource with ID = 401 for unauthorized users.
### Related issue(s)/PR(s)

None.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1204,7 +1204,7 @@ class modX extends xPDO {
             $options
         );
         $this->invokeEvent('OnPageUnauthorized', $options);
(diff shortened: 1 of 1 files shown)


Jako:
Nice find.

sergant210:
[The same](https://github.com/sergant210/revolution/blob/66f5c3bc87708521605b389ff741c7229f210fbc/core/model/modx/modx.class.php#L1183) for the error page. :)

rtripault:
Thanks, cherry picked into 2.5.x.

PR #13105, opened 2016-08-26
Removed duplicate element ID

rtripault:
### What does it do ?

Removes a duplicated HTML element ID
### Why is it needed ?

On system settings manager page, 2 different buttons share the same ID
### Steps to reproduce issue
-  navigate to `?a=system/settings`
- activate `system events` tab
- get back to `system setting` tab
- notice 2 `clear filters` buttons are displayed


Jako:
Merged into 2.5.x

Jako:
Thanks!

PR #13106, opened 2016-08-28
Fix: php.net link to cURL help page

### LINKED ISSUE #12634: php.net link doesn't work in Package Management
Link in the warning message (cURL must be installed and configured as an extension for PHP for you to use Package Management. You can find more information about cURL on the PHP cURL website.) to php.net/curl doesn't work in some languages (ex. Russian). 
default

AndreyPanyush:
### What does it do?

I just added slashes
### Why is it needed?

Because link doesn't work
### Related issue(s)/PR(s)

fix #12634


Mark-H:
Hi @AndreyPanyush, thanks for the fix but this needs to be adjusted in the translation system (CrowdIn) at http://translate.modx.com/ instead of directly in github. 

Could you fix it there instead? Let me know if you have any questions on how that works.

Thanks!

PR #13109, opened 2016-08-30
#13108: System event OnResourceSort to provide information whether re…

### LINKED ISSUE #13108: System event 'OnResourceSort' to provide information whether resource in tree was moved under new parent
### Summary
When hooking up to System event "OnResourceSort" (which fires when a resource was moved in the resource-tree through D&D), it is not possible to find out whether the resource's parent was changed during that operation, by the available objects provided by fireAfterSaveEvent(). This might be handy to find out - e.g. if you want to trigger further operations if the resource got a new parent (e.g. clearing cache / operations with children of that resource)

### Step to reproduce
Create a plugin and hook it to "OnResourceSort". When this event is triggered it is not possible to get information about the actual resource that became sorted so it is not possible to find out e.g. whether the parent of the resource was changed during that operation. Only the following objects are provided by the method:
`array( 
        'nodes' => &$this->nodes, 
        'nodesAffected' => &$this->nodesAffected, 
        'contexts' => &$this->contexts, 
        'contextsAffected' => &$this->contextsAffected, 
        'modifiedNodes' => &$this->nodesAffected, /* backward compat */ 
    );
`
### Observed behavior
It is not possible to determine the actual resorted element and find out whether its parent has changed during the sort operation

### Expected behavior
Provide a flag to indicate whether the parent has changed so further actions could be triggered

### Environment
Any version of modX Revolution

SiarAneas:
### What does it do?

Enhanced modResourceSortProcessor to provide sorted resource ID (the one that was dragged&dropped in resource tree) and an indication whether the sorted resources' parent got changed during that operation through fireAfterSort() invoking the 'OnResourceSort' event
### Why is it needed?

See #13108 
### Related issue(s)/PR(s)

Resolves #13108

### DIFF
--- a/core/model/modx/processors/resource/sort.class.php
+++ b/core/model/modx/processors/resource/sort.class.php
@@ -40,8 +40,10 @@ class modResourceSortProcessor extends modProcessor {
         $this->getNodesFormatted($data, $this->getProperty('parent', 0));
 
         $this->fireBeforeSort();
-        $resourceBeforeSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
-        $oldParentID = $resourceBeforeSort->get('parent');
+        if($this->getProperty('source_type')==="modResource"){
+            $resourceBeforeSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
+            $oldParentID = $resourceBeforeSort->get('parent');
+        }
 
         $target = $this->getProperty('target', '');
         $source = $this->getProperty('source', '');
@@ -65,9 +67,12 @@ class modResourceSortProcessor extends modProcessor {
         $sorted = $this->sort();
         if ($sorted !== true) return $this->failure($sorted);
 
-        $resourceAfterSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
-        $newParentID = $resourceAfterSort->get('parent');
-        if($oldParentID !== $newParentID) $this->resourceMovedUnderNewParent = true;
+        if($this->getProperty('source_type')==="modResource"){
+            $resourceAfterSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
+            $newParentID = $resourceAfterSort->get('parent');
+            if($oldParentID !== $newParentID) $this->resourceMovedUnderNewParent = true;
+        }
+        
         $this->fireAfterSort();
 
         $this->clearCache();
--- a/core/model/modx/processors/resource/sort.class.php
+++ b/core/model/modx/processors/resource/sort.class.php
@@ -12,11 +12,12 @@ class modResourceSortProcessor extends modProcessor {
     public $nodesAffected = array();
     public $contexts = array();
     public $contextsAffected = array();
+    public $resourceMovedUnderNewParent = false;
 
     public $source;
     public $target;
     public $point;
-    
+
     public $autoIsFolder = true;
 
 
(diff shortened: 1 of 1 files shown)


core/model/modx/processors/resource/sort.class.php
```diff
@@ -39,6 +40,8 @@ public function process() {
         $this->getNodesFormatted($data, $this->getProperty('parent', 0));
 
         $this->fireBeforeSort();
+        $resourceBeforeSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
```
rtripault: I believe this processor is also handling context sorting, so `source_pk` could be a `modContext.key`

SiarAneas: Just realized, that's true - bit confusing name for the processor then... I will provide a check whether $this->getProperty('source_pk') is a valid resource by checking whether $this->getProperty('source_type')==="modContext"

core/model/modx/processors/resource/sort.class.php
```diff
@@ -39,6 +40,10 @@ public function process() {
         $this->getNodesFormatted($data, $this->getProperty('parent', 0));
 
         $this->fireBeforeSort();
+        if($this->getProperty('source_type')==="modResource"){
+            $resourceBeforeSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
+            $oldParentID = $resourceBeforeSort->get('parent');
```
Mark-H: This could use a check if $resourceBeforeSort has a valid resource, otherwise this could cause a fatal error "method get on null" if for whatever reason the resource can't be loaded.

core/model/modx/processors/resource/sort.class.php
```diff
@@ -62,6 +67,12 @@ public function process() {
         $sorted = $this->sort();
         if ($sorted !== true) return $this->failure($sorted);
 
+        if($this->getProperty('source_type')==="modResource"){
+            $resourceAfterSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
+            $newParentID = $resourceAfterSort->get('parent');
+            if($oldParentID !== $newParentID) $this->resourceMovedUnderNewParent = true;
```
Mark-H: The body of the if clause should be wrapped in curly brackets `{ }` and on its own line

core/model/modx/processors/resource/sort.class.php
```diff
@@ -62,6 +67,12 @@ public function process() {
         $sorted = $this->sort();
         if ($sorted !== true) return $this->failure($sorted);
 
+        if($this->getProperty('source_type')==="modResource"){
+            $resourceAfterSort = $this->modx->getObject('modResource', array('id' => $this->getProperty('source_pk')));
+            $newParentID = $resourceAfterSort->get('parent');
```
Mark-H: Same as above, if for some weird reason the resource can't be loaded, this would cause a fatal error "method get on null".

Mark-H (changes requested):

Mark-H:
Pending the tweaks to account for potential fatal errors if something goes wrong, I think this is a nice improvement.

sdrenth:
@SiarAneas Are you planning on fixing the requested changes? Would be nice so we can get this PR merged ;)

Mark-H:
Closing as there has been no contributor feedback and I'm not comfortable merging it without those checks. If anyone else wants to recreate the PR with the requested fixes, be my guest :)

PR #13111, opened 2016-09-02
Update the function processEventResponse

sergant210:
### What does it do?

Implode() is executed only if $result is array.
### Why is it needed?

If plugin (when it changes the active status of user) doesn't return any output the response looks like an array `array(0=>'')`. This causes a warning `PHP warning: implode(): Invalid arguments passed` because the second argument of the implode function is not an array.
### Related issue(s)/PR(s)

As I know none.


Jako:
Merged into 2.5.x

Jako:
Thanks!

PR #13112, opened 2016-09-02
Update ht.access

Gallenkamp:
### What does it do?

Added browser caching for standard files and web fonts
### Why is it needed?

It will boost site performance by default
### Related issue(s)/PR(s)

None that I found

I added the ruleset for browser caching which I am using in all my sites


rtripault:
Hey Guido! Thanks for your contribution.

While i won't be affected by this, i wanted to highlight that most of default htaccess content is commented.
It would probably be better to keep it that way.

Also, some food for thought, it could be nice to check [H5BP](https://github.com/h5bp/server-configs-apache/blob/master/src/web_performance/expires_headers.conf) and/or eventually mention it in the htacess (pretty sure there are some nice chunks there).
What do you reckon ?

theboxer:
Closing due to long inactivity. Please feel free to reopen with more commits to reflect Romain's suggestions.

PR #13118, opened 2016-09-09
ExtJS Manager headers and descriptions components

rtripault:
### What does it do ?

Creates 2 new ExtJS components to help create manager pages headers & description
### Why is it needed ?

Prevent code duplication and provide ease of change (manager wise)


manager/assets/modext/core/modx.js
```diff
@@ -893,3 +893,30 @@ Ext.extend(MODx.HttpProvider, Ext.state.Provider, {
         Ext.Ajax.request(o);
     }
 });
+
+MODx.Header = function(config) {
```
rtripault: JS Doc maybe ?

PR #13119, opened 2016-09-09
Make sure things do not break when no valid json/empty string is retu…

rtripault:
### What does it do ?

Make sure things do not break when no valid json/empty string is returned OnMediaSourceGetProperties event
### Steps to reproduce
- create a plugin bound to `OnMediaSourceGetProperties` with content `return;`
- notice the manager crashed when editing a resource

### DIFF
--- a/core/model/modx/sources/modmediasource.class.php
+++ b/core/model/modx/sources/modmediasource.class.php
@@ -10,11 +10,11 @@
 interface modMediaSourceInterface
     /**
      * Initialize the source, preparing it for usage.
-     * 
+     *
      * @return boolean
      */{
     public function initialize();
-    
+
     /**
      * Return an array of containers at this current level in the container structure. Used for the tree
      * navigation on the files tree.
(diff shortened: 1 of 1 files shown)


core/model/modx/sources/modmediasource.class.php
```diff
@@ -1,729 +1,732 @@
...
+            $properties = $defaultProperties;
+        }
+        /** @var array $results Allow manipulation of media source properties via event */
+        $results = $this->xpdo->invokeEvent('OnMediaSourceGetProperties',array(
+            'properties' => $this->xpdo->toJSON($properties),
+        ));
+        if (!empty($results)) {
+            foreach ($results as $result) {
+                $result = is_array($result) ? $result : $this->xpdo->fromJSON($result);
+                if (!is_array($result)) {
```
rtripault: This is the change!

rtripault:
Sorry for the diff noise, line ending went from CRLF to LF

Jako:
I am not sure, but is this solved maybe with this: #13111

rtripault:
I thought so too when viewing the PR, but it appears not, since the PR is used in processors (which is not the case in this PR)

Jako:
Merged into 2.5.x

PR #13120, opened 2016-09-09
Removed a bunch of code in the manager "gateway" which should not be …

rtripault:
### What does it do ?

Removed a bunch of code in the manager "gateway" which should not be required


Jako:
Thanks! Merged into 2.5.x

PR #13124, opened 2016-09-11
Fixed addEventListener function

sergant210:
### What does it do?

Added the plugin ID to the plugin map as an array key.
### Why is it needed?

Now this method adds a plugin to the plugin map without any key - 
`$this->eventMap[$event][]= $pluginId;`
But the "InvokeEvent" method uses the array key to get the plugin object - 

```
foreach ($this->eventMap[$eventName] as $pluginId => $pluginPropset) {
    ...
    $plugin= $this->getObject('modPlugin', array ('id' => intval($pluginId), 'disabled' => '0'), true);
    ...
}
```

So the plugin added via the "addEventListener" method will not run.
### Related issue(s)/PR(s)

None.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1945,15 +1945,8 @@ class modX extends xPDO {
             if (!isset($this->eventMap[$event]) || empty ($this->eventMap[$event])) {
                 $this->eventMap[$event]= array();
             }
-            $ElementPropertySetTbl= $this->getTableName('modElementPropertySet');
-            $propsetTbl= $this->getTableName('modPropertySet');
-            $sql= "SELECT PropertySet.name 
-                FROM {$ElementPropertySetTbl} ElementPropSet 
-                LEFT JOIN {$propsetTbl} PropertySet ON ElementPropSet.property_set = PropertySet.id 
-                WHERE ElementPropSet.element = {$pluginId} AND ElementPropSet.element_class='modPlugin' AND PropertySet.name = '{$propertySetName}' ";
-            $propertySetName = $this->getValue($this->prepare($sql));
             $this->eventMap[$event][$pluginId]= $pluginId . (!empty($propertySetName) ? ':' . $propertySetName : '');
-            $added= true;
+            $added = true;
         }
         return $added;
     }
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1939,11 +1939,20 @@ class modX extends xPDO {
      */
     public function addEventListener($event, $pluginId) {
         $added = false;
+        $pluginId = intval($pluginId);
         if ($event && $pluginId) {
             if (!isset($this->eventMap[$event]) || empty ($this->eventMap[$event])) {
                 $this->eventMap[$event]= array();
             }
-            $this->eventMap[$event][$pluginId]= $pluginId;
+            $pluginEventTbl= $this->getTableName('modPluginEvent');
+            $propsetTbl= $this->getTableName('modPropertySet');
+            $sql= "SELECT PropertySet.name AS propertyset 
+                FROM {$pluginEventTbl} PluginEvent 
+                LEFT JOIN {$propsetTbl} PropertySet ON PluginEvent.propertyset = PropertySet.id 
+                WHERE PluginEvent.event = '{$event}' AND PluginEvent.pluginid = {$pluginId}
+            ";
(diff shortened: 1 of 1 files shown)


Jako:
It should be no bug before since $pluginId should be a number before the patch. But maybe PHP7 has a changed behaviour there. If that's the case, we have to look for similar code in the core.

Jako:
It seems to be more complicated. The $pluginPropset that is requested in https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L1584 is not added in the addEventListener method https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L1940

Is it needed somehow?

sergant210:
I don't know. But it's not difficult to add to the addEventLinstener function.

Jako:
Would be fine.

sergant210:
Done.

rtripault:
@sergant210 i'm not really sure about the "logic" here (the code looks fine).

From my understanding, this method is supposed to add an existing modPlugin to a event it is not attached/bound to (so no modPluginEvent for that event).
However your code to handle the modPropertySet supposes the plugin is already attached to the event.

My guess was that we need to modify the method signature to pass the property set name : `public function addEventListener($eventName, $pluginId, $propertysetName = '')`

What do you reckon ?

sergant210:
@rtripault You right. Fixed it. Thanks. But I can't find where these property set is used. I 'm interested does somebody use property sets for plugins?

Jako:
This are some as far as i know: https://github.com/donShakespeare/TinymceWrapper and https://github.com/BobRay/UpgradeMODX for example

sergant210:
@Jako Thanks.

rtripault:
@sergant210 i'm not using much property sets, so i've been digging a bit to understand how/where they are used. You might want to check `modPlugin::getPropertySet` (as well as `modElement::getPropertySet`)

core/model/modx/modx.class.php
```diff
@@ -1935,15 +1935,24 @@ public function removeAllEventListener() {
...
                 $this->eventMap[$event]= array();
             }
-            $this->eventMap[$event][]= $pluginId;
+            $ElementPropertySetTbl= $this->getTableName('modElementPropertySet');
+            $propsetTbl= $this->getTableName('modPropertySet');
+            $sql= "SELECT PropertySet.name 
+                FROM {$ElementPropertySetTbl} ElementPropSet 
+                LEFT JOIN {$propsetTbl} PropertySet ON ElementPropSet.property_set = PropertySet.id 
+                WHERE ElementPropSet.element = {$pluginId} AND ElementPropSet.element_class='modPlugin' AND PropertySet.name = '{$propertySetName}' ";
+            $propertySetName = $this->getValue($this->prepare($sql));
```
rtripault: I would completely remove the query (to find the property set name) since that's what the method 3rd parameter is supposed provide.

sepiariver:
Thanks for this fix. I didn't notice the addEventListener was broken? Except that when you modify Plugin properties and/or event registration, you usually have to clear cache twice to see the effects...maybe this will fix that.

For what it's worth, I use property sets on Plugins a LOT. They are very, very useful. Especially since when an Extra is updated, often it will define property set values in the package and thus overwrite the default property set. Custom propSets maintain the custom settings.

core/model/modx/modx.class.php
sergant210: @rtripault This query checks the specified property set is bound to the plugin or not. But I figured out that this check is [here](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modelement.class.php#L588).

sergant210: @rtripault My phpStorm took me the wrong way and I missed the modPlugin::getPropertySet method.
phpstorm1
.

sergant210:
@sepiariver I see. Thanks for information.

core/model/modx/modx.class.php
Jako: I agree with @rtripault. A plugin should be registered by that method to an event listener on the fly, the property set should be selectable there on the fly too.

sergant210:
Hi guys, 
What do you think about small enhancement of the "addEventListener" method to use the lambda-function like in Javascript.
`$modx->addEventListener('eventName','myFunc',function(&$modx, $var1, $var2){return $var1+$var2;}, array(1,2))`
I'm using this for some time. It works great. If it's interesting I can make a PR.

PR #13127, opened 2016-09-15
Let's not try to create a new DirectoryIterator on non existing folders

rtripault:
### What does it do?

Prevents creating a `DirectoryIterator` on non existing directory.
### Why is it needed?

Prevents issues like `Uncaught UnexpectedValueException: DirectoryIterator::__construct(/../core/components/ns-without-lexicon/lexicon/): failed to open dir: No such file or directory` (PHP 7)


Jako:
Merged into 2.5.x

PR #13128, opened 2016-09-16
Set temporary directory for files processing in phpThumb

bezumkin:
### What does it do?

Set temporary directory for files processing in phpThumb
### Why is it needed?

Otherwise you will get warnings on servers with open_basedir enabled on MODX 2.5.1

```
[2016-09-16 16:02:15] (ERROR @ /home/****/www/core/model/phpthumb/phpthumb.class.php : 1215) PHP warning: realpath(): open_basedir restriction in effect. File(/tmp) is not within the allowed path(s): (/home/****)
```


Jako:
Thanks! Merged into 2.5.x

Jako:
Thanks!

PR #13129, opened 2016-09-16
Update modX::removeEventListener method

sergant210:
### What does it do?

Removes the specified plugin from the eventMap for the specified event.
### Why is it needed?

The modX::addEventListener method registers the plugin in the eventMap. I can add as many plugins as I want. But I can't remove the certain plugin, only all.
### Related issue(s)/PR(s)

This [PR](https://github.com/modxcms/revolution/pull/13124).


rtripault:
Good catch for companion PR #13124.

PR #13130, opened 2016-09-19
Add User Group description to UserGroups grid (with row toggle)

pixelchutes:
### What does it do?

Adds User Group description (with row toggle) to the `UserGroups` grid in the Access Permissions tab when editing a user.
### Why is it needed?

When working with _a lot_ of User Groups, particularly when using a naming convention resulting in seemingly similar entries, having the User Group description available in this view can be very helpful to help distinguish / differentiate like-entries.


pixelchutes:
One benefit of this addition is that it supports showing the entire, longer descriptions, whereas the description shown in combos / dropdown gets truncated: 

user_group_desc

vs.

truncated_desc

theboxer:
Merged to 2.x

PR #13133, opened 2016-09-25
modRest refactoring

Jako:
### What does it do?

Some code refactoring
### Why is it needed?
- The return values don't match the PHPDoc definitions and the usage was different too
- Wrong phpDoc type in recursive method call
- Removed unused variables
- Updated property descriptions
- Method names are case sensitive

### DIFF
--- a/core/model/modx/rest/modrest.class.php
+++ b/core/model/modx/rest/modrest.class.php
@@ -593,7 +593,6 @@ class RestClientResponse {
      */
     public function _parse($result) {
         $headers = array();
-        $httpVer = strtok($result, "\n");
 
         while($line = strtok("\n")){
             if(strlen(trim($line)) == 0) break;
@@ -645,7 +644,6 @@ class RestClientResponse {
         if($valueKey && !is_string($valueKey)){$valueKey = '@values';}
 
         $return = array();
-        $name = $xml->getName();
         $_value = trim((string)$xml);
         if(!strlen($_value)){$_value = null;};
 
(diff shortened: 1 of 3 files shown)


core/model/modx/rest/modrest.class.php
```diff
@@ -473,7 +473,7 @@ public function toXml($parameters,$rootNode) {
      * @param array $parameters
      */
     protected function _populateXmlDoc(&$doc, &$node, &$parameters) {
-        /** @var $val DOMNode */
+        /** @var DOMnode/array $val */
```
Mark-H: I don't think that's the right format for phpdoc is it? Isn't that supposed to be `DOMnode[]`?

Mark-H (commented):
Could you elaborate on the reasoning behind changing from returning values to returning nothing? Sounds like a breaking change and I don't immediately see the benefit, so some more thoughts on what these changes accomplish would be great.

core/model/modx/rest/modrest.class.php
Jako: Maybe `DOMnode[]/DOMnode`

Jako:
The main reason for the change was that $this->success and $this->failure have no return values. The other methods use those return values. So the return was null every time (if the success and failure methods are not overridden).

pixelchutes:
Thanks @Jako, was curious about the reasoning there.

core/model/modx/rest/modrest.class.php
Alroniks: Alternatives should be separated using `|` instead of `/`. So correct way `DOMnode|array`

Jako: $parameters has now the right phpdoc type. The type warnings are away now.

Mark-H:
Thanks @Jako!

PR #13134, opened 2016-09-25
PHPDoc issue

Jako:
### What does it do?

Change the PHPDoc code
### Why is it needed?

It was wrong before

### DIFF
--- a/core/model/modx/modprocessor.class.php
+++ b/core/model/modx/modprocessor.class.php
@@ -137,7 +137,7 @@ abstract class modProcessor {
      * @param modX $modx A reference to the modX object.
      * @param string $className The name of the class that is being requested.
      * @param array $properties An array of properties being run with the processor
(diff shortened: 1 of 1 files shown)


core/model/modx/modprocessor.class.php
```diff
@@ -137,7 +137,7 @@ public function addFieldError($key,$message = '') {
      * @param modX $modx A reference to the modX object.
      * @param string $className The name of the class that is being requested.
      * @param array $properties An array of properties being run with the processor
-     * @return The class specified by $className
+     * @return string The class specified by $className
```
rtripault: Well, it appears it returns an instance of `modProcessor` or derivative class, not a string.

Jako: Sorry, you are right: the include returns the class name. This method returns the class.

PR #13138, opened 2016-09-28
Fixing the tvLabel PHP warning empty needle

yuka123:
### What does it do?

Checkin if m_val is not empty instead of isset
### Why is it needed?

To not produce the PHP warning empty needle


Jako:
Merged into 2.5.x with 1c6cb1a1accf9249fd64097c8e503a0dd4cf8800

PR #13139, opened 2016-09-28
Update resource.class.php

MaartenW:
### What does it do?

Changes `!empty` into `is_object && method_exists`
### Why is it needed?

At that point in the script the `$this->resource` can still be an Array instead a full fledged XPDOSimpleObject
### Related issue(s)/PR(s)

Just checking if a variable is not empty is not enough to conclude it has a certain method.

### DIFF
--- a/manager/controllers/default/resource/resource.class.php
+++ b/manager/controllers/default/resource/resource.class.php
@@ -149,7 +149,7 @@ abstract class ResourceManagerController extends modManagerController {
      * @return array|bool|string
      */
     public function firePreRenderEvents() {
-        $resourceId = !empty($this->resource) ? $this->resource->get('id') : (!empty($this->scriptProperties['id']) ? $this->scriptProperties['id'] : 0);
+        $resourceId = is_object($this->resource) && method_exists($this->resource, 'get') ? $this->resource->get('id') : (!empty($this->scriptProperties['id']) ? $this->scriptProperties['id'] : 0);
         $properties = array(
             'id' => $resourceId,
             'mode' => !empty($resourceId) ? modSystemEvent::MODE_UPD : modSystemEvent::MODE_NEW,


Jako (commented):
Would a typeof xPDOObject check be better than a method exists?

rtripault:
Sorry but could you point to the place where `$resource` attribute  could be an array ?
Also, there was a known issue with Revo 2.5.1 and "legacy" TinyMCE RTE (since the resource was passed by reference and TinyMCE was `$resource->toArray()`) see https://github.com/modxcms/TinyMCE/pull/20 in case that's relevant

MaartenW:
@Jako yeah, might be better. Is that the first object with the `get()` method?

@rtripault I probably don't understand you correctly but the `$resource` attribute is an array at the place visible in the filechange within this pull request. At least in my set up. I updated from 2.2.16 to 2.5.1 and got blank screens. Checked the server log which read 

```2016/09/28 21:08:45 [error] 13803#0: *250 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Call to a member function get() on a non-object in /var/www/manager/controllers/default/resource/resource.class.php on line 152" while reading response header from upstream, client: **, server: ***, request: "GET /manager/?a=resource/update&id=113 HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5-fpm.sock:", host: "***", referrer: "http://***/manager/?a=system/settings"```

I can't find a backtrace so this is all I know. Things shaped up when the change in this PR was made.

Jako:
Your issue occurs in MODX 2.5.1 in conjunction with TinyMCE before 4.3.4. It should not occur in other cases. So the fix would help with that deprecated version of TinyMCE but nowhere else.

Would suggest to close that.

MaartenW:
**The issue only occurs when skipping minor versions with upgrading.**  
When I tried to upgrade from 2.2.16 to 2.5.1, no cigar.
When upgrading step by step: 2.2.16 > 2.3.6 > 2.4.4 > 2.5.1 everything turned out hunky dory.

Sorry for the trouble. Thanks for helping me clear this up.

OptimusCrime:
That should really not be required as far as I know? A full upgrade to the latest version should always work.

rthrash:
When coming from 2.2.x, we almost never have it work without hitting 2.3.x first. Then we usually can go to 2.5.x latest. But I agree, it _should_ work.

Mark-H:
Thank you @MaartenW for sending a pull request when you found an issue :)

MaartenW:
It doesn't work. I hit a wall because the `_context` table wasn't updated to have a `name` field.
Oh, maybe good to mention: I ran into this while using CLI with `--installmode=upgrade`. It doesn't seem to run the `2.3.0-rc1.php` upgrade file. At least, that's where the database would be updated.

PR #13142, opened 2016-10-02
Fix $containerSuffix is appended twice

HitkoDev:
`$identifier = "{$identifier}{$containerSuffix}";` already appends the `$containerSuffix`. Using `$found = $this->modx->findResource("{$identifier}{$containerSuffix}");` would append it again.

### DIFF
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -322,7 +322,7 @@ class modRequest {
                     $found = $this->modx->findResource($identifier);
                 } else {
                     $identifier = "{$identifier}{$containerSuffix}";
-                    $found = $this->modx->findResource("{$identifier}{$containerSuffix}");
+                    $found = $this->modx->findResource($identifier);
                 }
                 if ($found) {
                     $parameters = $this->getParameters();


Mark-H:
Is there a related bug report or something to test this against?

Mark-H:
Hi @HitkoDev - the code change seems straightforward enough as a potential bug, but to test it before the merge it would be best if we could first reproduce the bug, and then confirm it's fixed after applying your patch. If you could please provide some additional information on how to reproduce, that would help move this PR along.

Mark-H:
The code change seems simple enough, and makes sense, but I'm really weary about merging without more detail about how to test the previous behaviour to see how it may affect unrelated use cases. 

@HitkoDev - I appreciate it has been forever, but any chance you recall how to reproduce the problem this fixes?

HitkoDev:
The easiest way to reproduce it is to set `container_suffix = '.ab'` (slash doesn't work, as sever removes duplicate slashes from the path), and navigate to `/container-resource.ab.ab` (notice the double suffix). Instead of 404, you will get the actual _container-resource_ page. Additionally, setting `friendly_urls_strict = 'Yes'` will redirect the double suffix URL to the canonical, single suffix URL.

Mark-H:
Thanks @HitkoDev! That, along with duplicates #13582 and #13719 helped me understand what's going on here :)

Merged for inclusion in 2.7.

Issue #13144, opened 2016-10-05
Missing file connectors/security/login.php prevent from renewing session


vladmirx:
### Summary
Missing folder /connectors/security/ and file connectors/security/login.php
Both needed to renew session if it is expired during your work in manager.
At this commit file was removed https://github.com/modxcms/revolution/commit/ec21e6f5e0b64aaaaf8e4f390c7f16b1efc82035

### Step to reproduce
Log in to manager, open any resource, delete session cookie and try to save resource.

### Observed behavior
If you submit modal login window, manager will do nothing. Because there is no such file there.

### Expected behavior
With this file, white notify message should popup with text: Session extended!
And you can continue to work.

### Environment
PHP 7.0
MODX Revolution 2.5.1-pl Advanced edition

Mark-H:
That file is not supposed to be needed. Instead the index.php connector should be posted to with the action security/login. That loads up the right logic. It does look like there's an issue, but I don't think your summary has the root cause right.

bezumkin:
@Mark-H But auth window [tries to send action into](https://github.com/modxcms/revolution/blob/07656896542a6b4195f65886fb43eaae366c34d1/manager/assets/modext/widgets/windows.js#L833) `MODx.config.connectors_url + 'security/login.php'`

Jako:
Hey, you have found another bug 👍  Try this change:

```
url: MODx.config.connectors_url,
action: 'security/login'
```

pixelchutes:
Fixed in 5c9e11ac61826c52e022f9e67869a24424265055

pixelchutes:
See previous PR for additional detail, but [connectors/security/login.php](https://github.com/modxcms/revolution/blob/d4ecf6106c2c86e68d44be0c79e0568122492676/connectors/security/login.php) **still exists** as of `2.5.2-dev`, and I cannot seem to reproduce this issue. _(e.g. I can renew session via this file without issue)_

Can you clarify what you mean by "missing folder" for **connectors/security/***? Does that not exist in the 2.5.1-advanced transport file, or are you suggesting it's not properly installed during setup? Have you tried re-running setup / transferring files from the 2.5.1 installer?

Lastly, what is your `compress_js` configured as? Does toggling `compress_js` setting have any effect for you?

[...]

PR #13146, opened 2016-10-06
Change to action based connector call

Jako:
### What does it do?

Changed the direct call of a connector to the action based default connector
### Why is it needed?

The old connector don't exist anymore
### Related issue(s)/PR(s)
#13144

### DIFF
--- a/manager/assets/modext/widgets/windows.js
+++ b/manager/assets/modext/widgets/windows.js
@@ -830,7 +830,8 @@ MODx.window.Login = function(config) {
     Ext.applyIf(config,{
         title: _('login')
         ,id: this.ident
-        ,url: MODx.config.connectors_url + 'security/login.php'
+        ,url: MODx.config.connectors_url
+        ,action: 'security/login'
         // ,width: 400
         ,fields: [{
             html: '<p>'+_('session_logging_out')+'</p>'


pixelchutes:
@Jako, I realize I tested this with `compress_js` setting enabled, and without running the [grunt process](https://github.com/modxcms/revolution/tree/2.x/_build/templates/default) before doing so...

Testing this change again now, with `compress_js` disabled, I am getting back **HTTP/1.1 401 Not Authorized** for this action based connector call:

```json
{"success":false,"code":401}
```

pixelchutes:
That said, I was *not* experiencing the issue described in #13144 with my compressed version of JavaScript (`compress_js` enabled). I was able to extend my session just fine.

So, perhaps there's something more going on? Perhaps a difference between clean install vs. upgrade to 2.5.x version.

pixelchutes:
To be crystal clear, I'm referring to it working via [connectors/security/login.php](https://github.com/modxcms/revolution/blob/d4ecf6106c2c86e68d44be0c79e0568122492676/connectors/security/login.php) in `2.5.2-dev` -- the file still exists.

login_php_200_ok

Jako:
Is that connector file maybe removed during running the setup by the cleaning script?

pixelchutes:
Good though, but no, I already checked. See new PR for why it was "missing" (e.g. not properly installed)

Jako:
We should look where the not action based processor is used and switch to the action based there.

pixelchutes:
Is there a correct way to invoke it without throwing `401`?

pixelchutes:
Perhaps we simply need to pass `ctx` in the request to bypass this? I confirmed in the action-based approach, it is missing.

**The current fields are:**

```
action
HTTP_MODAUTH
username
password
rememberme
```

I will test this.

PR #13151, opened 2016-10-20
Improved phpThumb InitializeTempDirSetting

Jako:
### What does it do?
- PHP internal values should be used before an environment value is retrieved
- Check each value with realPathSafe
### Why is it needed?

On one of my installations the TMPDIR points to the webroot and this folder is not writable by PHP for security reasons.
### Related issue(s)/PR(s)

Improves (replaces) #13128 for the following reasons:
- pthumb does not use modPhpThumb but phpThumb directly
- Using the cache folder for temporary images is not that useful. The system folder would be cleaned automatic, it is out of the webroot in most cases etc.

### DIFF
--- a/core/model/phpthumb/phpthumb.class.php
+++ b/core/model/phpthumb/phpthumb.class.php
@@ -4341,11 +4341,10 @@ if (false) {
 	}
 
 	function InitializeTempDirSetting() {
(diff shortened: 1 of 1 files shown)


Jako:
It was merged into the phpThumb repository too: https://github.com/JamesHeinrich/phpThumb/pull/75

PR #13154, opened 2016-10-21
Fix missing connectors/security/ directory for download distributions

### LINKED ISSUE #13144: Missing file connectors/security/login.php prevent from renewing session

pixelchutes:
### What does it do?

Ensures **connectors/security/**\* is no longer missing for download distribution based installs.
### Why is it needed?

Because the expired session / session extension handling relies on it.
### Related issue(s)/PR(s)

Fixes #13144

Reverts 5c9e11ac61826c52e022f9e67869a24424265055 (#13146)

### DIFF
--- a/_build/transport.core.php
+++ b/_build/transport.core.php
@@ -535,6 +535,10 @@ $attributes = array (
     'vehicle_class' => 'xPDOFileVehicle',
 );
 $files[] = array (
+    'source' => MODX_BASE_PATH . 'connectors/security',
+    'target' => "return MODX_CONNECTORS_PATH;",
+);
+$files[] = array (
     'source' => MODX_BASE_PATH . 'connectors/system',
     'target' => "return MODX_CONNECTORS_PATH;",
 );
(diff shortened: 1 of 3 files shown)


OptimusCrime:
I dont understand. The current form is the correct one for all connectors? Why does it not work in this case?

pixelchutes:
For us to support action based connector here, we need to add an explicit check to prevent `401 Unauthorized` error.

The alternative (this PR), is to use a legacy connector specifically to re-login.

The kicker is, for Git installs the file-based connector (login.php) is _already_ in place. For download distributions (.zip's), it never gets installed and will `404 Not Found`.

pixelchutes:
Fixed in fbf837d690a84924e17319cf134d5a2b862dacc5 using single connector approach.

pixelchutes:
Closing for good this time. Fixed by #13157 and #13158.

PR #13158, opened 2016-10-23
Add security/login support for action based connector

### LINKED ISSUE #13144: Missing file connectors/security/login.php prevent from renewing session

pixelchutes:
### What does it do?

Adds required support for `security/login` in `modConnectorResponse`
### Why is it needed?

To prevent internal 401 Unauthorized / Access Denied error for new action based connector request.
### Related issue(s)/PR(s)

Builds on #13157
Fixes #13144

### DIFF
--- a/core/model/modx/modconnectorresponse.class.php
+++ b/core/model/modx/modconnectorresponse.class.php
@@ -93,7 +93,7 @@ class modConnectorResponse extends modResponse {
         $target = str_replace('../','', htmlspecialchars($options['action']));
 
         $siteId = $this->modx->user->getUserToken($this->modx->context->get('key'));
-        $isLogin = $target == 'login';
+        $isLogin = $target == 'login' || $target == 'security/login';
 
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Thank you for explaining in #13154 @pixelchutes ! I understand the approach when I see this code. Good fix!

PR #13163, opened 2016-10-29
Make possible change 'content' in sendForward

faitno:
### What does it do?

Changes in hardcode field `content` in method sendForward
### Why is it needed?

Sometimes need to change `content`, but save `ID`, `class_key`, etc. For example logic in miniShop2 - display products in category, but some changes in `pagetitle` and`content` need.

Make possible change 'content' in sendForward MODX method.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1108,10 +1108,9 @@ class modX extends xPDO {
             $currentResource = array();
             if ($merge) {
                 $excludes = array_merge(
-                    explode(',', $this->getOption('forward_merge_excludes', $options, 'type,published,class_key')),
+                    explode(',', $this->getOption('forward_merge_excludes', $options, 'type,published,class_key,content')),
                     array(
-                        'content'
-                        ,'pub_date'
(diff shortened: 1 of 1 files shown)


Mark-H:
I too can see potential use cases for overriding the content with a `$modx->sendForward` call, however I think this pull request as it is currently may cause unintended, potentially breaking, changes. 

In the current situation `content` is always excluded. The new situation (at least the _code_) makes it configurable, *but* the setting is not changed. So when someone updates to a release that has this patch, `content` wont be excluded anymore. Similarly a new site will behave differently compared to a new site before this change. 

In order to merge this, I think we'll need: 

1) The default forward_merge_excludes setting (`_build/data/transport.core.system_settings.php`) updated to include the content
2) an upgrade script added that checks if the `forward_merge_excludes` setting was edited (editedon not being null) and if so adds `content` to it.

Mark-H:
I'm going to close this as in the current form it will change behaviour (as described in my previous comment). There's still merit to this proposal I think, so feel free to create a new PR that handles it differently.

PR #13164, opened 2016-10-31
Added stream based uploads and file handle for binary uploads

Jako:
### What does it do?
The file upload (in the post/put request) is changed to support streams. The handle for the stream is set as parameter value for the modRestController and could be handled there (at the moment for the contentType 'image/jpeg', 'image/png' and 'image/gif' (could be extended).

### Why is it needed?
The filehandle parameter is needed to support fileuploads via the REST API. Switching the file handling to streams is needed to decrease the memory usage on the server.

### Related issue(s)/PR(s)
None

### DIFF
--- a/core/model/modx/rest/modrestservice.class.php
+++ b/core/model/modx/rest/modrestservice.class.php
@@ -420,6 +420,7 @@ class modRestServiceRequest {
                 $data = stream_get_contents($filehandle);
                 fclose($filehandle);
                 $params = $this->service->modx->fromJSON($data);
+                $params = (!is_array($params)) ? array() : $params;
                 break;
             case 'application/x-www-form-urlencoded':
             default:
@@ -430,9 +431,6 @@ class modRestServiceRequest {
         }
         if ($this->service->getOption('trimParameters', false)) {
             array_walk_recursive($this->parameters, array('modRestServiceRequest', '_trimString'));
-        }
-        if (empty($params)) {
-            $params = array();
         }
         return $params;
 	}
--- a/core/model/modx/rest/modrestservice.class.php
+++ b/core/model/modx/rest/modrestservice.class.php
@@ -395,13 +395,13 @@ class modRestServiceRequest {
      * @return array
      */
 	protected function _collectRequestParameters() {
-            $filehandle = fopen('php://input', "r");
-	    $params = array();
-	    $contentType = isset($_SERVER['CONTENT_TYPE']) ? $_SERVER['CONTENT_TYPE'] : '';
-	    $spPos = strpos($contentType,';');
-	    if ($spPos !== false) {
-	        $contentType = substr($contentType,0,$spPos);
-	    }
+        $filehandle = fopen('php://input', "r");
+        $params = array();
+        $contentType = isset($_SERVER['CONTENT_TYPE']) ? $_SERVER['CONTENT_TYPE'] : '';
+        $spPos = strpos($contentType, ';');
+        if ($spPos !== false) {
+            $contentType = substr($contentType, 0, $spPos);
+        }
         switch ($contentType) {
             case 'image/jpeg':
             case 'image/png':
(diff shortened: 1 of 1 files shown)


Jako:
Some extended information: There was no image (binary upload) handling before. The data of an posted image could be retrieved before only in an array key in $_REQUEST. The key value was the binary data of that image. It must be cut at some position or is there no length limit on an array key? It did not get into $params of modRestServiceRequest. So I am quite sure that no one has ever written code to retrieve a posted/putted binary data with modRest and there should not be any BC with this.

core/model/modx/rest/modrestservice.class.php
```diff
@@ -395,32 +395,45 @@ public function setRequestParameters() {
...
+                fclose($filehandle);
                 parse_str($data, $params);
                 break;
         }
-        if ($this->service->getOption('trimParameters',false)) {
-		    array_walk_recursive($this->parameters,array('modRestServiceRequest','_trimString'));
+        if ($this->service->getOption('trimParameters', false)) {
+            array_walk_recursive($this->parameters, array('modRestServiceRequest', '_trimString'));
+        }
+        if (empty($params)) {
```
Alroniks: It does not make sense because `$params` already initialized by empty array on L399

rtripault: Indeed, i guess the check should be `!is_array()` instead

Alroniks: In the code above $params is always array so this check here not needed at all.

Jako: I have not changed the code on that part. It is only reformatted with curly braces.

$params could get null in line 422 with an invalid JSON, so the check makes sense. But it could be changed to !is_array() for sure.

sdrenth: @Jako Do you need to edit line 422 or is this PR ready to be merged?

Jako: I have modified it.

Mark-H:
Hm, I'm a bit worried about what this might do to backwards compatibility. Please correct me if I'm wrong, but if I understand this correctly, in the previous situation, the image data was available under in a request param with the name of the upload field(s). With this change applied, it is available as a stream with key filehandler? 

Wouldn't $_FILES work in this case to access the uploaded file?

Jako:
There was nothing in $_FILES. The only part where a POSTed image could be retrieved from, was a key in $_REQUEST, the key itself was 'binary image data'. I don't remember the value. I have searched the whole visible variable scope in PhpStorm with the REST debugger.

The following code is used on a client installation to upload an image and attach it to an existing object.
```
    /**
     * Handles updating of objects
     */
    public function put()
    {
        $id = $this->getProperty($this->primaryKeyField, false);
        if (empty($id)) {
            $this->failure($this->modx->lexicon('rest.err_field_ns', array(
                'field' => $this->primaryKeyField,
            )));
            return;
        }
        $c = $this->getPrimaryKeyCriteria($id);
        $this->object = $this->modx->getObject($this->classKey, $c);
        if (empty($this->object)) {
            $this->failure($this->modx->lexicon('rest.err_obj_nf', array(
                'class_key' => $this->classKey,
            )));
            return;
        }

        $realImagePath = realpath(MODX_ASSETS_PATH . $this->config['productImagesPath']);
        $realImageFile = $realImagePath . '/' . $this->object->get('article_nr') . '.jpg';
        $inputHandle = $this->getProperty('filehandle');

        if ((file_exists($realImageFile) && is_writeable($realImageFile)) ||
            (file_exists($realImagePath)) && is_writable($realImagePath)
        ) {
            $outputHandle = fopen($realImageFile, 'w+');
            $length = 0;

            while (true) {
                $buffer = fgets($inputHandle, 4096);
                $length += strlen($buffer);
                if (strlen($buffer) == 0) {
                    fclose($inputHandle);
                    fclose($outputHandle);
                    break;
                }
                fwrite($outputHandle, $buffer);
            }

            if (empty($length)) {
                $this->failure('The image file was empty');
                return;
            }
        } else {
            $this->failure('Can\'t upload the file to ' . $realImageFile);
            return;
        }

        $objectArray = array(
            'id' => $id,
            'image' => $realImageFile
        );

        $this->success('', $objectArray);
        return;
    }
```

Mark-H:
Thanks for the clarification and sample code. I have no objections against merging this with that additional information (but no time to do it myself right now).

PR #13170, opened 2016-11-04
Hide some critical setting in MODx.config

Fi1osof:
### What does it do?
Hide some system settings like table_prefix, pathes, dirs and passes.

### Why is it needed?
Critical! This info allow process SQL-injection via MODx.Ajax

### DIFF
--- a/core/model/modx/processors/system/config.js.php
+++ b/core/model/modx/processors/system/config.js.php
@@ -90,7 +90,13 @@ if (isset($scriptProperties['action']) && $scriptProperties['action'] != '' && i
 
 $c = array_merge($modx->config,$workingContext->config,$modx->_userConfig,$c);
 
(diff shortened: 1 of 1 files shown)


Jako:
But you need manager credentials to access the processor. Are you sure that no core javascript uses any of those _path|_dir|pass settings?

Mark-H:
Can you provide information about the security issue you mention to security@modx.com?

Mark-H:
Can you provide information about the security issue you mention to security@modx.com?

Fi1osof:
@Juko when we authorized in manager we can find any select-promted processor.

> Are you sure that no core javascript uses any of those _path|_dir|pass settings?

I sure. Only urls needed. pathes - it`s need on server`s side

Fi1osof:
@Mark-H sent.

rtripault:
This will break things for sure (see https://github.com/modxcms/Collections/issues/220).

I'd rather type the critical (core) system settings rather than just ditch everything ending with `_path` or `_dir`

Mark-H:
I'm closing this pull request as it was merged into 2.5.x; not sure why it didn't mark as closed automatically. 

A fix for the comment from Romain above is now in #13180

PR #13171, opened 2016-11-07
Fix truncating filename at space by downloading via filemanager

aaaatpdev:
### What does it do?
Wraps a filename to download in double quotes.

### Why is it needed?
It fixes truncating filename to download at space. For example, downloading file with name "Aaa aaa.pdf" by right-clicking in filemanager (and then "Download file") gives result filename as "Aaa". This behavior is peculiar to Firefox. Related link: https://bugzilla.mozilla.org/show_bug.cgi?id=221028.


Jako:
Merged into 2.5.x

PR #13173, opened 2016-11-13
Prevent path traversal in modConnectorResponse action param

opengeek:
### What does it do?
Prevents path traversal in the action parameter of modConnectorResponse.

### Why is it needed?
To prevent users from accessing processors in unintended locations.

### Related issue(s)/PR(s)
Not applicable.

Thanks to @Fi1osof for a security report describing the problem and the solution.

### DIFF
--- a/core/model/modx/modconnectorresponse.class.php
+++ b/core/model/modx/modconnectorresponse.class.php
@@ -24,7 +24,7 @@ class modConnectorResponse extends modResponse {
 
     public $responseCode = 200;
 
-	protected $_responseCodes = array(
+    protected $_responseCodes = array(
         100 => 'Continue',
         101 => 'Switching Protocols',
         200 => 'OK',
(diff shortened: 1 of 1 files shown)


Fi1osof:
I think it shell be where too: https://github.com/opengeek/revolution/blob/01bef4fe3da3d7238d1900c8db784f828a863e42/core/model/modx/modx.class.php#L1662-L1664

azernov:
What about modx.config.js.php and processor config.js.php? He will be never found.

Mark-H:
Merged into 2.5.x https://github.com/modxcms/revolution/commit/9f16dbb80131ec125b6713d5387cfe51edcd8889

opengeek:
Good point @azernov — we will have to improve the regex a bit...

azernov:
@opengeek As i remember - config.js.php - is one processor with dot in name. May be it will be simpler to rename it to configjs.php?

Mark-H:
There's also lang.js.php, but that's all of them these days. We managed to get a slightly different regex to work though, see https://github.com/modxcms/revolution/commit/befef7eba7b44439c4d1923bc254cf88cc09b880 and let me know if you see any further issues with that.

PR #13174, opened 2016-11-14
Make sure a bunch of processors have a permission check

Mark-H:
### What does it do?
Adds a permission check to the following processors that previously had none:

```
- workspace/getlist		set to workspaces
- workspace/theme/getlist	set to settings
- system/phpinfo			set to view_sysinfo
- system/event/grouplist 	set to events
- search/search			set to search
- resource/translit			set to view
- resource/getlist			set to view
- resource/get 			set to view
- resource/reload			set to save_document
- resource/locks/release	set to view
- element/tv/getlist 		set to view_tv
- element/template/getlist 	set to view_template
- element/plugin/getlist 	set to view_plugin
- element/chunk/getlist 	set to view_chunk
- element/category/getlist 	set to view_category
- element/snippet/getlist	set to view_snippet
```

There's a couple more processors lacking a permission check, however I'm not sure if or how those should be resolved. These are:

```
- system/downloadoutput 	possibly allows downloading files from core or cache?
- system/console 			not sure what this needs?
- system/registry/register/read same ^ 
- system/registry/register/send same ^
- system/contenttype/getlist	not sure on required permission, possibly content_types or view_document?
- system/rte/getlist			not sure on req permission
- system/phpthumb				not sure if or how to fix => possible vulnerability or breaking change => has a check on the source tho, so safe?
- element/tv/renders/* 		seems to return a php error for missing class, ignore?
- browser/directory/getlist has no permision check directly, but does check on the media source. Ignore?
```

Feedback on the permissions those processors should check for is welcome.

### Why is it needed?
Combined with a vulnerability that allows access to processors that shouldn't be accessible (#13173), some of these processors may be abused into providing access to other vulnerabilities that expose information or do other Bad Things (r). By adding in a permission check to these processors this sort of attack is prevented (however, other fixes are also already made on both ends of the hoop). 

This PR also makes sure that certain information that authenticated manager users aren't supposed to have access to, also can't be accessed by forging specific requests. 

### Related issue(s)/PR(s)
#13173

### DIFF
--- a/core/model/modx/processors/element/category/getlist.class.php
+++ b/core/model/modx/processors/element/category/getlist.class.php
@@ -15,6 +15,7 @@ class modElementCategoryGetListProcessor extends modObjectGetListProcessor {
     public $classKey = 'modCategory';
     public $languageTopics = array('category');
     public $defaultSortField = 'category';
+    public $permission = 'view_category';
 
     public function initialize() {
         $initialized = parent::initialize();
--- a/core/model/modx/processors/element/chunk/getlist.class.php
+++ b/core/model/modx/processors/element/chunk/getlist.class.php
@@ -15,5 +15,6 @@ require_once (dirname(dirname(__FILE__)).'/getlist.class.php');
 class modChunkGetListProcessor extends modElementGetListProcessor {
     public $classKey = 'modChunk';
     public $languageTopics = array('chunk','category');
+    public $permission = 'view_chunk';
 }
 return 'modChunkGetListProcessor';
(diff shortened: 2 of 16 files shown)


OptimusCrime:
Some of the files seems to have change line endings, which makes the diff somewhat noisy. Might be worth either doing the line endings change in a separate commit, or ignore it for now and take it in a later PR.

Mark-H:
I believe those are being autocorrected because they were wrong before... kinda annoying, but can be reviewed without whitespace by adding `?w=1` to the url: https://github.com/modxcms/revolution/pull/13174/files?w=1

pixelchutes:
Looks like v2.5.2 has already shipped. Bump this to milestone v2.5.3?

opengeek:
Any update on these? If not, can you split the ones you have questions on into a separate issue @Mark-H? I'd like to release 2.5.3 in the morning.

gernotebenlechner:
#13505 #modxbughunt

Issue #9607, opened 2013-03-09
Make MODx.config less verbose and/or customizable


opengeek:
opengeek created Redmine issue ID 9607

The MODx.config object that is constructed by the system/config.js processor is currently too verbose, exposing database connections, API tokens, credentials and other potentially sensitive information stored in MODX config settings. This should be customizable via a System Event where users and/or Extras could install Plugins that unset certain settings that should not be exposed.

pbowyer:
While MODx.config is still verbose, it doesn't expose database connections etc AFAICT. @opengeek is this valid, as addressed in https://github.com/modxcms/revolution/pull/13180? Otherwise be good to close.

#modxbughunt

opengeek:
I really don't deal with the modExt side of the house much. Would be better for others to provide input on this.

OptimusCrime:
The MODx.config has been trimmed a lot since this issue was first reported. As far as I can see in what is left, there is nothing that should not be exposed there now.

Closing. Feel free to comment or reopen if you disagree.

@modxbot close

PR #13180, opened 2016-11-16
Remove path related settings from MODx.config, partial revert of #13170

Mark-H:
### What does it do?
Hardcodes specific setting values that need to be removed from the MODx.config array in the manager, instead of the partial match on path/dir/pass. 

### Why is it needed?
The previous fix from #13170 was a bit too aggressive, causing a break in Collections (modxcms/Collections#220). It can't be ruled out that other extras are also affected, so part of the fix there has been reverted in favour of a more specific.

The goal of #13170 and this pull request is to keep confidential information (server file paths) away from the frontend. While not a direct vulnerability, the value of some of these settings could potentially aid future attacks by simplifying retrieving information about the server. 

### Related issue(s)/PR(s)
#13170, modxcms/Collections#220

### DIFF
--- a/core/model/modx/processors/system/config.js.php
+++ b/core/model/modx/processors/system/config.js.php
@@ -90,13 +90,7 @@ if (isset($scriptProperties['action']) && $scriptProperties['action'] != '' && i
 
 $c = array_merge($modx->config,$workingContext->config,$modx->_userConfig,$c);
 
-unset($c['password'],$c['username'],$c['mail_smtp_pass'],$c['mail_smtp_user'],$c['proxy_password'],$c['proxy_username'],$c['connections'],$c['connection_init'],$c['connection_mutable'],$c['dbname'],$c['database'],$c['table_prefix'],$c['driverOptions'],$c['dsn'],$c['session_name']);
-
-foreach($c as $k => $v){
-    if(preg_match("/(_path|_dir|pass)/u", $k)){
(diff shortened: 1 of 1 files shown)


Mark-H:
@Fi1osof Please let me know if there are other settings that should be removed here. I went through MODx.config on my site to see which ones would expose the server paths and added those to the list.

rtripault:
Seems like a less restrictive approach!

Just wanted to point this is somehow related to issue #9607

OptimusCrime:
I'm more in favor of splitting this up. Not only is the line faaar longer than suggested in standards, it also makes diffs difficult. The diff will happen now anyways, but we can spare the future an ambiguous commit.

For example: 

```
<?php
unset($c['password'],
    $c['username'],
    $c['mail_smtp_pass'],
    $c['mail_smtp_user'],
    $c['proxy_password'],
    $c['proxy_username'],
    $c['connections'],
    $c['connection_init'],
    $c['connection_mutable'],
    $c['dbname'],
    $c['database'],
    $c['table_prefix'],
    $c['driverOptions'],
    $c['dsn'],
    $c['session_name'],
    $c['assets_path'], 
    $c['base_path'], 
    $c['cache_path'], 
    $c['connectors_path'], 
    $c['core_path'], 
    $c['friendly_alias_translit_class_path'], 
    $c['manager_path'], 
    $c['processors_path']
);
```

Issue #13079, opened 2016-07-21
PHP7 and not utf8 capable system locale


Jako:
### Summary

As @kwazaro mentioned in #13002 there is an issue on some shared hosts that makes the manager unusable if the locale setting is not containing an utf8 capable locale.
### Step to reproduce

It can't be reproduced here, but he said, that the system works fine on PHP7, but only if "locale" system setting is set to i.e. "en_GB.utf8". If he leaves locale parameter empty the Package Manager doesn't work, there are empty values in reports etc.
### Observed behavior

Some parts of the manager don't work, i.e. the Package Manager because of a broken JSON.
### Expected behavior

They should not break, if the MODX locale system setting is empty.
### Environment

PHP7 on a russian shared host (as far as I know).

OptimusCrime:
Yeah, this issue is pretty critical. Weird behavior too.

oorooroot:
This error is about setLocale() php function behavior under server environment such as windows 7 as example. 

Here is some output from my apache on win7 :
```
$this->getOption('locale')
"ru_RU.UTF8"

setlocale(LC_ALL, "ru_RU.UTF8")
false

setlocale(LC_ALL, "0")
"Russian_Russia.1251"

strftime('%b %d %Y', 1471864827)
"��� 22 2016"
```

Here is the root of problem, that causes incorrect symbols output in locale-dependent functions like strftime()
https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L2374

Jako:
An approach to solve this would be: 
- Check if the current setting uses a utf8-capable locale.
- If not run some translit before the string is JSON encoded.

optim1zer:
I found temporary solution for this, while we waiting fo fix.
Edit 2 start files 
In manager/index.php:
change
`$modx->initialize('mgr');`
to
`$modx->initialize('mgr', ['setlocale'=>false]);`

And in connectors/index.php:
`$modx->initialize($ctx);`
to 
`$modx->initialize($ctx, ['setlocale' => false]);`

Jako:
You even could create a MODX system setting 'setlocale' => 0. Maybe a setmgrlocale system setting could be created. It would supercede the setlocale setting.

OptimusCrime:
Can we close this? The system setting is introduced. As far as MODX goes there is not much more can be done here.

#modxbughunt

MaartenW:
[...]

PR #13182, opened 2016-11-21
Fix listing packages with non-utf8 locales (#13079)

fn3k4:
Hello,
there is a bug present in modx revo up to last version 2.5.1.
This bug already has an issue discussion:
https://github.com/modxcms/revolution/issues/13079

Conditions:
- OS locale non-english with 8-bit codepage (ru_RU.cp1251 for example)
- Modx locale ru_RU
- modx_charset UTF-8
- PHP 5 >= 5.2.0
- PHP runs as CGI

How to replay a bug:
- Install modx
- Enter "Extras -> Install" page
- Select "modx.com" provider
- Select "Extras" on the left pane
- Select subitem of the "Extras" ("Blogging" for example)
- You should see infinite "Loading..." message and no listing of extras.

Reason why it is happened.
Thereis a code in the file
core\model\modx\processors\workspace\packages\rest\getlist.class.php
=======================
public function initialize() {
...
  $this->setDefaultProperties(array(
...
    'dateFormat' => '%b %d, %Y',
...
=======================

where "%b" date format specifier produces *localized* short month name with OS locale settings.
Then this string falls into following code in the file
core\xpdo\xpdo.class.php
=======================
    public function toJSON($array) {
        $encoded= '';
        if (is_array ($array)) {
            if (!function_exists('json_encode')) {
                if (@ include_once (XPDO_CORE_PATH . 'json/JSON.php')) {
                    $json = new Services_JSON();
                    $encoded= $json->encode($array);
                }
            } else {
                $encoded= json_encode($array); // *** about line 2412
            }
        }
        return $encoded;
    }
=======================

Then built-in function "json_encode" cannot encode 8-bit name of the month in cp1251 and returns empty result in variable "$encoded".

Possible solutions:

1. made 'dateFormat' field numeric only:
=======================
'dateFormat' => '%Y-%m-%d',
=======================
this will work everywhere.

2. add an option to call built-in "json_encode" to ignore errors:
=======================
$encoded= json_encode($array, JSON_PARTIAL_OUTPUT_ON_ERROR);
=======================
then you cannot see the released date, but you can see the extras list

3. Or do not use built-in "json_encode"

### DIFF
--- a/core/model/modx/processors/workspace/packages/rest/getlist.class.php
+++ b/core/model/modx/processors/workspace/packages/rest/getlist.class.php
@@ -26,7 +26,7 @@ class modPackageRemoteGetListProcessor extends modProcessor {
             'sorter' => false,
             'start' => 0,
             'limit' => 10,
-            'dateFormat' => '%b %d, %Y',
+            'dateFormat' => '%Y-%m-%d',
             'supportsSeparator' => ', ',
         ));
         $start = $this->getProperty('start');


Mark-H:
`JSON_PARTIAL_OUTPUT_ON_ERROR` was added in PHP 5.5, so unfortunately we can't use that until MODX 3 which will bump the minimum again. (Currently, MODX requires at least 5.3.3)

So I guess changing the date to a non-locale based one, like this pull request, is the only way to fix this currently.

Jako:
This is not a MODX issue in my opinion and I don't like this to be merged. It is an issue created by using a not UTF8 capable locale in a MODX system setting. Maybe we should introduce a system setting that activates that json_encode option as bypass (the wrong encoded array value is suppressed with that option - and the wrong encoded string content could be missing, if I'm right).

OptimusCrime:
2nd @Jako 

The solution is to use correct width charsets. UTF8 is the way to go for all things.

Jako:
Another idea to solve this, is to remove locale specific results in the connector (as suggested) **and** add a column renderer in the ExtJS grid column.

opengeek:
Should this PR be closed then?

Jako:
I will prepare a column renderer in ExtJS.

Jako:
Could you try to solve it by setting the MODX system setting `setlocale` to `0` (respective `No`).

See: https://github.com/modxcms/revolution/issues/13079#issuecomment-268982621 and https://github.com/modxcms/revolution/issues/13079#issuecomment-268984710

Jako:
Merged (and added an ExtJS column renderer) into 2.5.x

PR #13183, opened 2016-11-21
Fix unpacking

gvozdb:
### What does it do?
Fixing unpacking in MODX 2.5.2

### Why is it needed?

### Related issue(s)/PR(s)

### DIFF
--- a/core/model/modx/processors/browser/file/unpack.class.php
+++ b/core/model/modx/processors/browser/file/unpack.class.php
@@ -17,6 +17,9 @@ class modUnpackProcessor extends modProcessor {
 
     public function initialize() {
         $this->properties = $this->getProperties();
+        if (!preg_match('/^' . str_replace('/', '\/', MODX_BASE_PATH) . '/', $this->properties['path'])) {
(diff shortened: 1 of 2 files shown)


Mark-H:
From the changes it looks like this is related to #13170 where the _path settings are removed from the config object?

gvozdb:
Yes.

Jako:
Is a full path really needed? Could the source path not be used?

gvozdb:
Now in version 2.5.2 does not unpack archives. With this patch unpack.

Jako:
I meant, that a full path does not have to be used in the connector. The media source and a relative path should be enough for identifying the unpack path.

gvozdb:
OK. My code is based on this:
`,path: MODx.config.base_path + node.attributes.directory`

OptimusCrime:
Fixed in alternate PR #13238

@modxbot close

OptimusCrime:
Hmm, it looks like I don't have permissions to close this. @Jako mind closing this pending PR?

PR #13184, opened 2016-11-22
Semantic Placeholders on Login Screen

jpdevries:
### What does it do?
Changed placeholder attributes for username and password on MODX login screen to be examples of valid input.

### Why is it needed?
The placeholder attribute is not meant to describe an input.
Placeholder is meant to offer an example of a valid entry.

### Related issue(s)/PR(s)
An alternative to https://github.com/modxcms/revolution/pull/13186

Did not hook the placeholders up to Lexicons. Seems unnecessary.

### DIFF
--- a/manager/templates/default/security/login.tpl
+++ b/manager/templates/default/security/login.tpl
@@ -59,14 +59,14 @@
             <div class="x-form-item login-form-item login-form-item-first">
                 <label for="modx-login-username">{$_lang.login_username}</label>
                 <div class="x-form-element login-form-element">
-                    <input type="text" id="modx-login-username" name="username" autocomplete="on" autofocus value="{$_post.username|default}" class="x-form-text x-form-field" placeholder="{$_lang.login_username}" aria-required="true" required />
+                    <input type="text" id="modx-login-username" name="username" autocomplete="on" autofocus value="{$_post.username|default}" class="x-form-text x-form-field" placeholder="admin" aria-required="true" required />
                 </div>
             </div>
 
             <div class="x-form-item login-form-item">
                 <label for="modx-login-password">{$_lang.login_password}</label>
                 <div class="x-form-element login-form-element">
-                    <input type="password" id="modx-login-password" name="password" autocomplete="on" class="x-form-text x-form-field" placeholder="{$_lang.login_password}" aria-required="true" required />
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Isn't that just more confusing? I can see clients getting confused that their username says admin. While providing examples might be the actual use case for placeholders, I am more in favor of using them instead of labels, like many do now.

For example, logging into Google:

skarmavbild 2016-11-22 kl 12 44 01

Or logging into Facebook (Swedish language, sorry about that):
skarmavbild 2016-11-22 kl 12 45 53

jpdevries:
> Isn't that just more confusing?

As it is now, modern screen readers like VoiceOver will read 'username' and 'password' twice. If we aren't going to use placeholder correctly we should remove it as it is redundant both visually and audibly. What purpose does it serve as is?

> I can see clients getting confused that their username says admin.

If you can find any research or case studies to suggest that standard use of placeholder confuses users please share. If so I'll pass it along upstream to the W3C.

Note that Facebook and Google are hiding the label or using `aria-label` to prevent redundancy. Their forms still aren't standards compliant though.

It is possible to achieve the placeholder design semantically. You position the label on top of the input, hide the placeholder, then when the input receives focus you move the label upwards and make it smaller, let the placeholder show. 




Generally speaking, there are [many reasons placeholder attributes are problematic](https://medium.com/simple-human/10-reasons-why-placeholders-are-problematic-f8079412b960#.exe5xl77e) and another is that they are not translated by Google Translate. For example, here is Facebook login in Dutch translated to English using Chrome. Notice the placeholders are not translated.

jpdevries:
As much as I'm a fan of a proper use of placeholder, I actually prefer https://github.com/modxcms/revolution/pull/13186

It is a lighter design and can be visually deciphered quicker without placeholders at all.

OptimusCrime:
Ditto. However, having to consider what Chrome translations etc does makes things very difficult. The translations are a service, an option, not an integrated part of the WWW. If it does not work, it does not work. I would also say that using placeholders the same way that the "big" guys does, it more than justified, even though they "hack" things to get screen readers etc to work. The semantic of HTML is not perfect, and there are things that the browsers just have not caught up to yet. Making things work has long been the way of interpreting the W3C, instead of following them with no deviation.

My five cents in this. I also do prefer #13186. Mostly because it makes more sense, but also because I feel like this option is more confusing and does not really add anything to the form via the placeholders.

Mark-H:
Merged #13186 into 2.5.3.

PR #13186, opened 2016-11-22
Removing placeholders from login screen

jpdevries:
### What does it do?
Removed placeholders from the login screen inputs.

### Why is it needed?
The placeholders are redundant both audibly and visually as they are the same as the corresponding labels.

### Related issue(s)/PR(s)
An alternative to https://github.com/modxcms/revolution/pull/13184


rthrash:
+1 Fan of this one over #13184

Mark-H:
Cherry-picked into 2.5.x for 2.5.3 in 41f97afc1144ff24e45a618427ea885f816805d7. Please double check the source and target branches you choose; bug fixes can go into patch releases ;)

Issue #12596, opened 2015-08-25
Media Sources getting wrong name


sottwell:
In a site with 13 Media Sources, upon upgrade to 2.4 Media Source 1 (Filesystem) is displayed with the name of Media Source 11. It shows correctly in select drop-downs, such as when selecting a Media Source for a TV, but once selected it displays with the name of 11 (Directors).
![screen shot 2015-08-25 at 9 35 31 pm](https://cloud.githubusercontent.com/assets/1073780/9475748/53d33d66-4b71-11e5-81cb-fc1c0f086ab2.png)
![screen shot 2015-08-25 at 9 39 01 pm](https://cloud.githubusercontent.com/assets/1073780/9475863/22167aa8-4b72-11e5-9f1d-7d9da5733e01.png)
![screen shot 2015-08-25 at 9 39 15 pm](https://cloud.githubusercontent.com/assets/1073780/9475885/36c7c38a-4b72-11e5-9f2c-4d5999d4213d.png)

krismas:
Reproduced, on a fresh 2.4 install, with new Media Sources. The problem seem to appear for any Media Source if his ID in [10, 20[ and his name (sort ASC) is before "Filesystem"... A sort key/value problem !?

krismas:
At the initial display the name is valid, then after some ms this "valid label" is replaced by the "wrong label". So an ExtJS problem !?

OptimusCrime:
Guessing it is fetched from some processor and applied somehow, not sure how

krismas:
"source/getlist" is called as many times as the number of Media Sources! No cache! Is this normal?

inreti:
@krismas: that would indeed be a performance problem.

sottwell:
this is still an issue with 2.5.1 - media sources #1, #2 etc are given the name of #10, #20, etc. They appear in the alphabetical listing in the order one would expect if they had the correct names.
![screen shot 2016-08-13 at 20 22 52](https://cloud.githubusercontent.com/assets/1073780/17644603/c681b2e6-6193-11e6-8e10-3d27b8df63c7.png)

sottwell:
For example, that second Cache should be Carousel, and the second Directors should be Filesystem.

IamNeo1:
Hi,
i have an installation with 7 contexts and 3 media source for each context. I have this problem too and my customers are not very assumed about that. 
I am using  modx 2.5.1.
[...]

PR #13189, opened 2016-11-23
Fixing #12596 Media Sources getting wrong name

Jako:
### What does it do?
Change the javascript code to get the name of the right media source.

### Why is it needed?
At the moment some media sources get a wrong name from a different media source in the file tree.

### Related issue(s)/PR(s)
#12596


Mark-H:
Merged to 2.5.x in 5896a79df83ea020d8d5933e4ab6edeaabbccbcc - please double check the target branch next time ;)

Issue #13179, opened 2016-11-15
Unable to view previous versions of packages: sortby works incorrect with the IF operator


sergant210:
### Summary
xPDOQuery::sortby stops to work with conditions like this 

> $c->sortby('IF(modTransportPackage.release = "" OR modTransportPackage.release = "ga" OR modTransportPackage.release = "pl","z",IF(modTransportPackage.release = "dev","a",modTransportPackage.release)) DESC','');

It's because of [this](https://github.com/modxcms/revolution/blob/37973d2a2cdc29ca6f4894f2f9e787e33650844c/core/xpdo/om/xpdoquery.class.php#L402-L405).

### Step to reproduce
Open any Extra which has the old versions and go to the tab "Uploaded Versions". There will be nothing.

### Environment
MODX version 2.5.2-pl.

Mark-H:
Fixed per #13190 in 2.5.3

PR #13190, opened 2016-11-23
Broken "Uploaded Versions"

kudashevs:
### What does it do?
Fix Package Manager -> View Details -> Uploaded Versions view.

### Why is it needed?
No issue yet.

### Related issue(s)/PR(s)
No issue yet.

After upgrading to 2.5.2 found what if i use View Details in Package Manager i get error in error.log: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ASC, modTransportPackage.release_index DESC LIMIT 20' at line 1" and Uploaded Versions tab is empty. While investigating this problem found, what new sortby function from xpdoquery.class.php (commit 067cb74d41af6419120f87d98f576a97be820d95) now require second argument obligatory. This change back Package Manager to work normaly.


Mark-H:
Relevant issue: #13179

Jako:
Thanks for the patch!

Mark-H:
Thanks @kudashevs; this has been merged for 2.5.3.

PR #13191, opened 2016-11-24
[3.x] Fix warnings and errors in processors due to missing `use` statements

Mark-H:
### What does it do?
Resolves E_WARN errors triggered by incorrect method signatures on a few processors. These were caused by the xPDO 3 refactor and were missed in updating the processors to use the right xPDO/Om/* classes. 

### Why is it needed?
Broken functionality and warnings that pop up and/or get logged. 

### Related issue(s)/PR(s)
None. 

Logged errors/warnings that are fixed in this pull request:

- PHP warning: Declaration of modElementCategoryGetListProcessor::prepareQueryBeforeCount(xPDOQuery $c) should be compatible with modObjectGetListProcessor::prepareQueryBeforeCount(xPDO\Om\xPDOQuery $c)
- PHP warning: Declaration of modTemplateGetListProcessor::prepareQueryBeforeCount(xPDOQuery $c) should be compatible with modElementGetListProcessor::prepareQueryBeforeCount(xPDO\Om\xPDOQuery $c)
- PHP warning: Declaration of modUserGroupAccessNamespaceGetListProcessor::prepareQueryBeforeCount(xPDOQuery $c) should be compatible with modObjectGetListProcessor::prepareQueryBeforeCount(xPDO\Om\xPDOQuery $c)

### DIFF
--- a/core/model/modx/processors/element/category/getlist.class.php
+++ b/core/model/modx/processors/element/category/getlist.class.php
@@ -1,4 +1,5 @@
 <?php
+use xPDO\Om\xPDOQuery;
 /**
  * Grabs a list of Categories.
  *
--- a/core/model/modx/processors/element/template/getlist.class.php
+++ b/core/model/modx/processors/element/template/getlist.class.php
@@ -1,5 +1,6 @@
 <?php
 use xPDO\Om\xPDOObject;
+use xPDO\Om\xPDOQuery;
 
 require_once (dirname(dirname(__FILE__)).'/getlist.class.php');
 /**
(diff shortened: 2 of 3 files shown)


Mark-H:
Oh come on, whitespace issues again? I've got the editorconfig plugin enabled in PhpStorm, so it should be using the options configured in the .editorconfig (`end_of_line = lf`) but either 1) we have a bunch of incorrect files in git now, or 2) I'm doing something wrong in my environment. 

`git config core.autocrlf` shows it's set to input, which I guess should be `auto` instead? I usually commit using the phpstorm gui so I'm not sure if that uses the same config, but it might do.

OptimusCrime:
Can this be merged soon?

OptimusCrime:
@opengeek ?

PR #13193, opened 2016-11-24
[3.x] Prevent uncaught E_FATAL if core/vendor/autoload.php doesn't exist

Mark-H:
### What does it do?
In normal usage this condition shouldn't, like, ever trigger. However during development it's easy to forget to run `composer install` and you're presented with a blank manager and frontend. This fix catches that oversight and subtly reminds people to install dependencies.

### Why is it needed?
I'm familiar with using composer on different projects, and I was stumped for about a minute trying to get the 3.x branch running because I forgot to run `composer install`. Now imagine someone who never used composer before trying to get the 3.x branch running and forgetting about that; they'll spend a lot longer than a minute figuring it out. The error shown now should point them in the right direction. 

### Related issue(s)/PR(s)
N/a


OptimusCrime:
Please merge this

PR #13194, opened 2016-11-24
[3.x] Add setup upgrade to remove the manager/min/ folder on upgrade

Mark-H:
### What does it do?
Adds an upgrade file that removes the `manager/min/` folder on upgrade. 

### Why is it needed?
`manager/min/` was removed in #12778. These migrations make sure that the related files are also removed when upgrading from 2.x to 3.0. 

### Related issue(s)/PR(s)
#12778


alroniks (approved):

bezumkin (approved):

alroniks:
I've checked it before and now we have approval from @bezumkin too. So, I guess it can be safely merged. @opengeek

opengeek (commented):
Is there a reason we did not just use `xPDOCacheManager->deleteTree()` with the `deleteTop` option set to `true`?

alroniks:
@Mark-H Do you have an answer to Jason's question?

Mark-H:
Not really, other than that this appears to have been the way files/folders were cleaned up in the past. That's a terrible reason to do something, I admit that, but that's the only explanation I have at the moment. :P 

Maybe the original years-old reason is that we can't use xPDOCacheManager reliably yet in the setup? That's just a guess though.

Mark-H:
@opengeek Do you feel strongly about this needing to use the xPDOCacheManager or is the current PR acceptable to you?

JoshuaLuckers (commented):
Since it's a destructive action I'm in favour of providing a list with the actual files that where deleted instead of deleting everything.  A user/developer may have placed custom files in this directory.

Jako:
I don’t think that placing any code there makes sense without referring the other minify code. But I can be wrong.

Mark-H:
I'm with @Jako - that doesn't make a whole lot of sense to do. 

@opengeek This just needs your stamp.

opengeek (approved):

OptimusCrime:
Can not believe I lived to see this got merged. Time for celebrations you guys!

PR #13204, opened 2016-12-02
Update checkPassword

Fi1osof:
### What does it do?
Allow return error via plugin.
$modx->event->_output = true;  // Will not check password
$modx->event->_output = false; // Will check password
$modx->event->_output = 'error'; // Will return error


rtripault:
Thanks, working as expected!

Fi1osof:
@rtripault not at all!

gpsietzema:

Issue #13203, opened 2016-12-01
Wrong behaviour of output filter if resource have no such a field 


manicko:
### Summary
The output filters for the non-existent resource fields does not work as expected

### Step to reproduce
Put the code  [[*newField:default=`DEFAULT`]] in a Modx resource content or template or chunk. (Be sure that resource do not have the field with this name).

### Observed behavior
Returns nothing.

### Expected behavior
The expected output is DEFAULT

### Environment
MODX 2.5.1

This makes impossible the usage of the same chunk for Modx resources and migx tables.
E.g. if I have the field MyField in Migx table and I'm using  the same chunk for both Migx and  Modx resources with the code below [[*MyField:default=`DEFAULT`]]
I expect that for Modx resources I get otput as DEFAULT  and for Migx I expect to see the value of the field MyField.  Though for Modx resources the output is empty.

pixelchutes:
At first I thought this may be related to #11969, but after re-reading this I'm not thinking it is.

I can't say I've ever tried using MODX TV tags (for non-existent TVs) combined with output modifiers... That said _(and MIGX aside)_, I was not able to reproduce this in core.

### Example:

```
[[*abc:default=`xyz`]]
```
> Non-existent TV: `abc`

### Expected output (per the above comments)

Returns nothing.

### Actual behavior

Outputs:
> xyz

OptimusCrime:
I tested this as well and I got the same output as @pixelchutes 

@manicko perhaps this is an issue tied to MIGX? I do not use this extra, so I am not sure how it works, or just what it does. Or perhaps you have changed the default MODX parser to pdoTools or something? Either way I am pretty sure this is not a MODX core bug.

manicko:
No, you can reproduce this on the fresh installation
Please check: 
[https://c0849.paas1.fra.modxcloud.com/manager/?a=element/template/update&id=1](url)

Log: Admin
Pass: 123654789

manicko:
![1](https://cloud.githubusercontent.com/assets/1097057/20845441/2842ad02-b8d6-11e6-961c-c936a018dbf1.png)

![2](https://cloud.githubusercontent.com/assets/1097057/20845442/28441782-b8d6-11e6-89b2-a31180dde2d2.png)

OptimusCrime:
I can confirm that this is happening when building from source.

OptimusCrime:
[...]

PR #13205, opened 2016-12-03
Fix incorrect/inconsistent handling of non-existing TV

OptimusCrime:
### What does it do?
Before, if a TV (or attribute of a resource) was called, it would silently be ignored by the parser. This made it impossible to use output modifiers such as default on them. This behavior differs from how placeholders are handled.

Illustration. The following content field on a template which renders nothing but the content. Two template variables are created, `hello` and `hello2`. The former is non-empty, while the latter is. Content content:

```
Pagetitle = [[*pagetitle]].
Pagetitle with default:foo = [[*pagetitle:default=`foo`]].
Existing (non-empty) TV with default:bar = [[*hello:default=`bar`]].
Existing (empty) TV with default:bar = [[*hello2:default=`bar`]].
Non-existing TV with default:bar = [[*foo:default=`bar`]]. 
Placeholder that is not sat with default:bar = [[+foo:default=`bar`]].
```

Previous output:
```
Pagetitle = Home.
Pagetitle with default:foo = Home.
Existing (non-empty) TV with default:bar = hello.
Existing (empty) TV with default:bar = bar.
Non-existing TV with default:bar = . 
Placeholder that is not sat with default:bar = bar.
```

Output with patch:
```
Pagetitle = Home.
Pagetitle with default:foo = Home.
Existing (non-empty) TV with default:bar = hello.
Existing (empty) TV with default:bar = bar.
Non-existing TV with default:bar = bar. 
Placeholder that is not sat with default:bar = bar.
```

PS: I ignored composer.phar and the vendor dir. It made testing a lot easier. I also added a test to make sure this behavior is not altered without a heads up.

### Why is it needed?
As reported in the related issue, this behavior is inconsistent with how similar markup is parsed. It could also look like this behavior has changed recently, although I was not able to figure out how or why. In 2.4.x the behavior introduced in this PR is present.

### Related issue(s)/PR(s)
#13203 and perhaps others (?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,10 @@ _build/templates/default/sass/font-awesome
 *.css.map
 *.js.map
 
-#ignore composer files
+# ignore composer files
+composer.phar
 composer.json
 composer.lock
+
+# ignore composer vendor dir
+/vendor
--- a/_build/test/Tests/Model/modParserTest.php
+++ b/_build/test/Tests/Model/modParserTest.php
@@ -461,4 +461,11 @@ class modParserTest extends MODxTestCase {
             array("name", "name@propset:filter=`name@propset:filter`"),
         );
     }
+
+    public function testDefaultNonExistingTvValue() {
+        $output = "[[*foo:default=`bar`]]";
+        $this->modx->parser->processElementTags('', $output, true, false, '[[', ']]', array(), 10);
+        $this->assertEquals($output, "bar", "Did not parse non-existing TV with default modifier correctly");
+
+    }
 }
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -505,7 +505,16 @@ class modParser {
                         $element->setCacheable($cacheable);
                         $elementOutput= $element->process($tagPropString);
                     }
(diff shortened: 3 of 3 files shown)


Jako:
Is this somehow a BC? Or could this be merged as bugfix?

OptimusCrime:
Where it was reported it may look as if this had a different handling earlier, although I am not sure what caused the change. The handling of TVs like this differs from how everything else is handled, so it could be considered a bug fix that was too. However, it does break how it currently functions (as of the most recent version), which also creates a breaking change. I could try to locate the change that introduced the problem in the first case?

OptimusCrime:
I am wondering if it's this change that introduced the bug:

https://github.com/modxcms/revolution/commit/1c6cb1a1accf9249fd64097c8e503a0dd4cf8800

Jako:
That was just a change for the tvLabel modifier.

OptimusCrime:
You might be right, I have no idea what the code right there does. However, this is still a valid problem and was a backwards change introduced in a minor version.

OptimusCrime:
This can also be merged.

Mark-H:
Thanks @OptimusCrime!

Issue #13202, opened 2016-12-01
Duplicate Resource Alias Alert Doesn't Display unless Document Tab Visible


pyrographics:
There is a user experience issue in MODX 2.5.2 where, when saving a new resource, if you are in template variables tab when you click save and there is an error on the document tab (duplicate alias) it will not display the notification to the end-user.  I think there should be an additional alert directly under the save button.

markwillis82:
This happens on all of the required fields. We should switch the active tab to where the error is.

Forced URI is another example.

OptimusCrime:
Tried to implement a solution for this problem. I've experienced it multiple times myself without knowing what the heck is going on. I am pretty sure there are other issues similar to this that may be hidden from the users, but I could not remember any while I was coding. Feel free to check my solution out.

OptimusCrime:
There is a PR for this posted now. It also solves similar problems if you have any errors in the TV section.

PR #13206, opened 2016-12-03
Automatically change tab when saving resource and invalid values/states are present in another tab

OptimusCrime:
### What does it do?
Implement an extension of the current tab behavior when updating/creating a resource. If you happen to have a duplicated alias/uri and the processor returns an error, you will not automatically see the correct tab where the error is, if you are not already on that tab. See gif below.

**NOTE: This illustration is no longer valid. This PR now handles illegal states for all fields in the form, not only uri/alias.**

New behavior:
alias-new-behavior

### Why is it needed?
Solves an annoying issue that most of us have experienced at some point.

### Related issue(s)/PR(s)
#13202

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -137,7 +137,6 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
     }
 
     ,beforeSubmit: function(o) {
-        console.log('done');
         var ta = Ext.get('ta');
         if (ta) {
             var v = ta.dom.value;
@@ -236,7 +235,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
 
         // The next check needs the tabs index value to check if it is hidden or not
         if (tabs && tabs.items && tabs.items.keys) {
-            tabs_index = tabs.items.keys.indexOf(tab_name)
+            tabs_index = tabs.items.keys.indexOf(tab_name);
         }
 
         // We are already on a tab that has an invalid state. No need to switch
@@ -272,6 +271,8 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             var btn = Ext.getCmp('modx-abtn-save');
             if (btn) { btn.enable(); }
         }
+
+        this.fireEvent('failureSubmit');
     }
 
     ,freezeUri: function(cb) {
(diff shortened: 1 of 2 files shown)


Jako:
I think this is not enough (but it's nice for a start). A resource could be not saveable when a TV is required.

A field (should have an `x-form-field` class) with an error gets a `x-form-invalid` class. Searching `#modx-content` for those elements, looking if it is member of one or more tab content (`x-tab-panel`) and activating all tabs for the first element found should make it.

sdrenth:
Since this PR has been untouched for a while, is it perhaps an idea to accept this PR and maybe create another issue for the solution provided by @Jako? Or is @OptimusCrime planning on working on this PR any further?

I've tested the PR and I can confirm that it works.

OptimusCrime:
I was planning on trying to take a look at it. I just have way too much going on these days. If I find the time, I'll look at this first.

OptimusCrime:
I have expanded on the original PR. I implemented a new event which is fired upon failing to submit the form. This change had to be done to the `modx.component.js` file, as no events were available when a form contains an invalid state.

In the new event we are passed the entire MODX Resource component. We grab each tab form and simply loop the children recursively until we either hit the bottom, or we find an element with `x-form-invalid` class attached. Note that this approach is different from the validation done in `modx.component.js`. Their implementation uses the `validate()` method, which is -- as far as I can tell -- much more costly to call, especially on the entire form. I might be incorrect about this, however.

Some minor details about validating aliases exist, as this code was copied and pasted over in the new event handler.

I have tested the implementation, and it works like a charm!

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -136,6 +137,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
     }
 
     ,beforeSubmit: function(o) {
+        console.log('done');
```
pixelchutes: Debug log? Should this be cleaned up?

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -200,6 +202,70 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
...
+        if (tab_name === null) {
+            return;
+        }
+
+        var tabs = Ext.getCmp('modx-resource-tabs');
+        var tabs_index = null;
+
+        // The next check needs the tabs index value to check if it is hidden or not
+        if (tabs && tabs.items && tabs.items.keys) {
+            tabs_index = tabs.items.keys.indexOf(tab_name)
```
sdrenth: Missing semicolon at end of line

OptimusCrime:
Fixed comments, and implemented behavior for duplicate aliases.

See it in action below. A few notices: 

- Check and handling for duplicate aliases is only possible if the form itself was submitted. This means that any other mistakes (such as missing TVs) needs to be fixed first.
- If the TVs are categorized, this code does not switch to the correct category. This might be a future improvement.

behavior_demo

I've removed the WIP tag. This should be good for merging. I could squash the commits it you'd like?

Mark-H:
Thanks @OptimusCrime

Issue #13187, opened 2016-11-22
Category when сreating new template/tv/chunk/plugin/snippet 


legiero:
### Summary
When you create an element category becomes category with lowest id

### Step to reproduce
New installation of MODX 2.5.2 (no packages)
Create 1 category
Create 1 element ( template/tv/chunk/plugin/snippet )

### Observed behavior
When you create an element category becomes category with lowest id

### Expected behavior
When you create an element category must be "None", 0.

### Environment
MODX version: 2.5.2, Testing on modxcloud.com (2.5.1 normal behavior

UPD. 02/12/2016
When i use this buttons the category is "None" and link is "a=element/template/create"
![image](https://cloud.githubusercontent.com/assets/1620809/20840331/f0341ac8-b8c0-11e6-9ae7-abf6a101ece2.png)

When i use right click on element category the category is the category with lowest id and link is "a=element/template/create&category=0"
![image](https://cloud.githubusercontent.com/assets/1620809/20840475/6d446644-b8c1-11e6-91c3-64ec495a0c43.png)

If i use "a=element/template/create&category=**id**" where **id** is negative or non-existent, then the category of the element becomes empty as the expected behavior.

himurovich:
+1 The same issue detected on 2.5.2 modxcloud too

ychelovek:
+1

OptimusCrime:
The behavior changed between versions? That is really weird. I'll check it out.

natedin:
+1

Marduh:
+1

Mark-H:
Fixed per #13207

PR #13207, opened 2016-12-03
Make sure default category when get param is 0 is None for elements

OptimusCrime:
### What does it do?
For some reason MODX wants to load the default category as the category with the lowest ID. Given the query string `&category=0` the default category should be None. This PR fixes that, HOWEVER, someone should really look into what is going on, because there is something else wrong here I belive, but I was not able to figure out just what is was.

This line:
```
$this->category = $this->modx->getObject('modCategory',$scriptProperties['category']);
```

Did multiple times load a category that had a totally different ID than what the `scriptProperties` array contained.

Current behavior:
category-before

New behavior:
category-after2

### Why is it needed?
Fixes a bug.

### Related issue(s)/PR(s)
#13187

### DIFF
--- a/manager/controllers/default/element/chunk/create.class.php
+++ b/manager/controllers/default/element/chunk/create.class.php
@@ -70,7 +70,7 @@ class ElementChunkCreateManagerController extends modManagerController {
      */
     public function getCategory(array $scriptProperties = array()) {
         /* grab default category if specified */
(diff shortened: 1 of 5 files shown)


opengeek:
This should no be necessary any longer. The bug was caused by invalid behavior in xPDO's getObject method.

Issue #13178, opened 2016-11-15
File Manager does not check extension of renamed file


Pathologic:
### Summary
File Manager does not check extension of renamed file, so it's possible to upload php file as image and then rename.

### Step to reproduce
1. Create file image.jpg:
```
<?php
phpinfo();
```
2. Upload via File Manager
3. Rename to image.php

### Observed behavior
image.jpg turns to image.php

### Expected behavior
File renaming should be aborted if new file does not have extension allowed by Media Source or system settings. 

### Environment
MODX 2.5.1

Pathologic:
Temporary fixed with plugin:
```
$e = $modx->event;
if ($e->name == 'OnFileManagerFileRename') {
    $properties = $source->getProperties();
    $allowed = empty($properties['allowedFileTypes']['value']) ? $modx->getOption('upload_files') : $properties['allowedFileTypes']['value'];
    if (!empty($allowed)) {
        $ext = strtolower(pathinfo($path, PATHINFO_EXTENSION));
        $allowed = explode(',',$allowed);
        if (!in_array($ext,$allowed)) unlink($path);
    }
}
```

Jako:
Good catch! 👍

Pathologic:
Not my catch, that was mentioned in modx.pro community.

PR #13208, opened 2016-12-03
Validate file extension when renaming files in file browser

OptimusCrime:
### What does it do?
Do validation of file extension when renaming objects. Fixing potential security issue.

NOTE: I have not actually been able to test the S3 media source, but I belive they share the same code for this method up to the point where I do the check, so it should work as intended.

### Why is it needed?
Security.

### Related issue(s)/PR(s)
#13178

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -494,6 +494,21 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
             return false;
         }
 
+        // Check that the new file has a file type that is allowed
+        $allowedFileTypes = explode(',',$this->xpdo->getOption('upload_files',null,''));
+        $allowedFileTypes = array_merge(explode(',',$this->xpdo->getOption('upload_images')),explode(',',$this->xpdo->getOption('upload_media')),explode(',',$this->xpdo->getOption('upload_flash')),$allowedFileTypes);
+        $allowedFileTypes = array_unique($allowedFileTypes);
+
+        $file_split = explode('.', $newName);
+        $ext = $file_split[count($file_split) - 1];
+        if (count($file_split) > 1 and !in_array($ext, $allowedFileTypes)) {
+            $this->addError('path',$this->xpdo->lexicon('file_err_ext_not_allowed',array(
+                'ext' => $ext,
+            )));
+
+            return false;
+        }
+
         /* sanitize new path */
         $newPath = $this->fileHandler->sanitizePath($newName);
         $newPath = dirname($oldPath).'/'.$newPath;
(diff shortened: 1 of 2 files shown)


Jako:
Could someone check the S3 Media Source part and merge this?

core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -494,6 +494,21 @@ public function renameObject($oldPath,$newName) {
             return false;
         }
 
+        // Check that the new file has a file type that is allowed
+        $allowedFileTypes = explode(',',$this->xpdo->getOption('upload_files',null,''));
+        $allowedFileTypes = array_merge(explode(',',$this->xpdo->getOption('upload_images')),explode(',',$this->xpdo->getOption('upload_media')),explode(',',$this->xpdo->getOption('upload_flash')),$allowedFileTypes);
+        $allowedFileTypes = array_unique($allowedFileTypes);
+
+        $file_split = explode('.', $newName);
+        $ext = $file_split[count($file_split) - 1];
```
Mark-H: Interesting way to get the extension - wouldn't it be easier to just use [pathinfo() with PATHINFO_EXTENSION](http://php.net/manual/en/function.pathinfo.php)?

Mark-H:
modFileMediaSource has a `allowedFileTypes` property - should that be checked as well? S3 source doesn't have that AFAIK.

core/model/modx/sources/modfilemediasource.class.php
OptimusCrime: For about 15 years, this has been my goto way of grabbing the file extension. I was not aware of the `pathinfo()` approach, but I found it used in the same class, here: https://github.com/OptimusCrime/revolution/blob/26f59a39e238333d5185161c4c1f7d8b74610c70/core/model/modx/sources/modfilemediasource.class.php#L726

Do you want me to rewrite it? Pretty sure the implementation is more or less identical under the hood, but doing things one way may be better?

Mark-H: A lot has happened for PHP over the last 15 years, so if you don't mind I think pathinfo would be cleaner ;) That link you just posted does flag a potential issue in this PR which is the extensions are being checked in a case sensitive way, so perhaps (untested) it could use some case normalisation too.

Jako: @OptimusCrime The missing case sensitivity should at least be fixed and the `pathinfo()` method looks cleaner.

I could check the code with a S3 media source.

Jako:
@Mark-H I am a bit insure about that `allowedFileTypes`. It is just used in the getContainerList method for filtering not listable files. Is it meant to restrict uploads?

The description says the same: 

```
$_lang['prop_file.allowedFileTypes_desc'] = 'If set, will restrict the files shown to only the specified extensions. Please specify in a comma-separated list, without the dots preceding the extensions.';
```

Jako:
The S3 source has a check for the option, but it can't be set in the backend.

Jako:
All four options: allowedFileTypes, upload_images, upload_media, upload_flash should be combined in MODX 3.x

Jako:
Closing this in favour of #13240

Issue #13212, opened 2016-12-08
Dragging a file from the tree to a weblink causes undefined error


matdave:
### Summary
Unable to drag files into a weblink box.

### Step to reproduce
Create a weblink, place the cursor in the weblink box, drag a file from the tree into it.

### Observed behavior
Results in "Undefined" error and requires page refresh. 
![image](https://cloud.githubusercontent.com/assets/2213127/21020988/5567a908-bd3c-11e6-97da-516e7996b191.png)

### Expected behavior
Place the path of the file, which works if you drag it into any other boxes. 

### Environment
MODX 2.5.2, Chrome 54.0.2840.98

matdave:
Console error 
![image](https://cloud.githubusercontent.com/assets/2213127/21021093/b457d898-bd3c-11e6-8711-2e12b645499f.png)

PR #13213, opened 2016-12-08
Update modx.treedrop.js

matdave:
### What does it do?
It adds a check to weblink and symlink fields in case an item being dragged is not a resource.

### Why is it needed?
When dragging a file into a weblink it would cause an "Undefined" issue because it was looking for a resource ID that doesn't exist on files. 

### Related issue(s)/PR(s)
I want to resolve issue #13212

### DIFF
--- a/manager/assets/modext/widgets/modx.treedrop.js
+++ b/manager/assets/modext/widgets/modx.treedrop.js
@@ -99,7 +99,11 @@ Ext.extend(MODx.TreeDrop,Ext.Component,{
                         }
                         if (el.dom.id == 'modx-symlink-content' || el.dom.id == 'modx-weblink-content') {
                             Ext.getCmp(el.dom.id).setValue('');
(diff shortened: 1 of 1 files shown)


juro:
The same bug is in /manager/assets/modext/modx.jsgrps-min.js and it's still there in 2.5.4

Jako:
It will be included in the compressed files 2.5.5

pixelchutes:
@Jako will that be true for Git installations, too, or only for download distribution? (e.g. .zip)

Jako:
modx.jsgrps-min.js contains the uglified coreJSfiles (https://github.com/modxcms/revolution/blob/2.x/_build/templates/default/gruntfile.js#L214-L225) which includes manager/assets/modext/widgets/modx.treedrop.js (https://github.com/modxcms/revolution/blob/2.x/_build/templates/default/gruntfile.js#L24). It was just a forgotten `grunt build` before updating the Revolution repository and packing the distribution.

PR #13220, opened 2016-12-17
Added autoHeight in the Create/UpdateSetting window

Jako:
To make additional x-types inside possible i.e. a grid that could grow with additional content.

Example Screenshot


theboxer:
Merged to 2.5.x & 2.x

PR #13224, opened 2016-12-22
Added option to disable EVAL binding in TVs

rtripault:
Note : this is just a proposal for a first review, we might want to create a system setting at some point.

### What does it do?

Adds a setting to disable `eval` in TV binding, keeping the default behavior as it currently is.
When the option is disabled, the code/value will just be handled as regular text.

### Why is it needed?

Allowing code to be `eval()`'ed can lead to some security issue. While we always assumed people with access to the manager are trust worthy, it could be different on some other projects.

### DIFF
--- a/core/model/modx/modtemplatevar.class.php
+++ b/core/model/modx/modtemplatevar.class.php
@@ -618,7 +618,7 @@ class modTemplateVar extends modElement {
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


Mark-H:
If we can add the setting to the build, this looks 👍 to me

Issue #13210, opened 2016-12-05
Fatal error crashes manager for nonexistent Chunks


pixelchutes:
### Summary
When attempting to edit a Chunk using an ID that doesn't exist, MODX Manager crashes with a 500 fatal error instead of handling it better, like it does with Template Variables or Resources.

### Step to reproduce
1. Browse to Chunk edit page for ID that doesn't exist, e.g. https://domain.com/manager/index.php?a=element/chunk/update&id=9999999999

### Observed behavior
500 Fatal error when browsing to nonexistent Chunk ID.

> NOTICE: PHP message: PHP Fatal error:  Call to a member function get() on null in manager/controllers/default/element/chunk/update.class.php on line 134

### Expected behavior
Still usable manager with a more appropriate error message (like that of TVs and Resources):

![better_error_handling](https://cloud.githubusercontent.com/assets/352182/20904626/a7e456c2-bafd-11e6-9e82-4001e0a6a8e1.jpg)

_^ Like this, but for Chunks._

### Environment
MODX 2.5.2 / PHP 5.6

Mark-H:
Fixed per #13226

PR #13226, opened 2016-12-24
Access chunk array instead of chunk object instance

OptimusCrime:
### What does it do?
Change update controller for chunks to mirror the behavior of snippets. This fixes an issue where the controller will throw a fatal error if you try to enter a ID for a chunk that does not exist.

skarmavbild 2016-12-24 kl 16 37 30

### Why is it needed?
Fix fatal error, and replace it with an error message.

### Related issue(s)/PR(s)
Reported in #13210


Omeryl:
Did this just get merged in @opengeek?

Jako:
Seems to be merged into 2.5.x

OptimusCrime:
Ah, did not notice that.

PR #13227, opened 2016-12-27
[Security] Update PhpMailer to 5.2.19 to protect against RCE vulnerability

Mark-H:
### What does it do?
Updates the PHPMailer version included with MODX to 5.2.19. PHPMailer was found to have a remote code execution vulnerability, which could be used to get control of a server. 

The advisory can be found here: https://legalhackers.com/advisories/PHPMailer-Exploit-Remote-Code-Exec-CVE-2016-10033-Vuln.html 

[This WordFence article](https://www.wordfence.com/blog/2016/12/phpmailer-vulnerability/) also has some details on the scope of things. 

### Why is it needed?
Make 100% sure MODX is not vulnerable. 

I've not yet been able of successfully executing the public exploit in MODX, but am planning to do some more testing. As the exploit is based on the `sender` containing certain inputs, theoretically this could interact with something like FormIt where the visitor provides an email address in a contact form, and the site owner receives an email that looks as if it was sent from that email. In that case the `sender` would be provided by a third party.

[Another source](https://github.com/opsxcq/exploit-CVE-2016-10033#vulnerable-code) mentioned the exploit requiring PHP 5.2 or older. As MODX requires 5.3 or up, that would also reduce the risk of MODX falling victim to the vulnerability.
 
### Related issue(s)/PR(s)

Security inbox, discussions on slack.


digitalbricks:
Referring to [legalhackers.com](https://legalhackers.com/advisories/PHPMailer-Exploit-Remote-Code-Exec-CVE-2016-10045-Vuln-Patch-Bypass.html), the vulnerability **isn't fixed in PHPmailer 5.2.19** but in 5.2.20.

Mark-H:
5.2.20 was released after this was merged, but yes we need to update it again.

digitalbricks:
Yes, i know. Meanwhile version 5.2.21 of PHPmailer was published because in the 5.2.20 they forgot to increment the version number string.

BTW: 

> Another source mentioned the exploit requiring PHP 5.2 or older. As MODX requires 5.3 or up, that would also reduce the risk of MODX falling victim to the vulnerability.

The section of the readme, referring to PHP 5.2 or older, was [removed](https://github.com/opsxcq/exploit-CVE-2016-10033/commit/dfd7a7f93e4f0c6e5044112f0db6115e4ec73b66). So it seems they are not sure that newer versions aren't affected.

Just out of curiosity: Are there some MODX specific modifications of the PHPmailer library or could i patch existing MODX installs just by replacing the files with the new ones from the PHPmailer site?

Omeryl:
@digitalbricks Largely should just be able to drop the new files in.

PR #13232, opened 2017-01-04
Put GitHub files in own directory

OptimusCrime:
### What does it do?
Clean up things a bit. Moving all GitHub related stuff in its own directory (as supported by GitHub).

### Why is it needed?
Less files at the root is a good thing.

### Related issue(s)/PR(s)
None


theboxer:
Merged to 2.5.x & 2.x

Issue #13221, opened 2016-12-18
Wrong message on duplicate TV with existing name


himurovich:
### Summary
Wrong message lexicon is used for noticing non-unique TV name

### Step to reproduce
Please open any existing TV, click on "Duplicate" button and enter new TV name, use some other existing TV name

### Observed behavior
MODX reacts on non-unique name and returns message but with wrong lexicon **key tv_err_ae**
![tv_err_ae](https://cloud.githubusercontent.com/assets/2815640/21293793/622a271e-c53f-11e6-8f82-0106662ae49e.png)

### Expected behavior
right lexion here is **tv_err_exists_name** which contains *"A Template Variable already exists with the name "[[+name]]".*

### Environment
MODX Revolution 2.5.1-pl

Alroniks:
Spent some time for investigation. Not ready to prepare PR but leave here some info.
modTemplateVarDuplicateProcessor uses inherited method `process` and `alreadyExists` from common modObjectProcessor class. But in update and create processors used custom method that uses another lexicon, not that provided by modObjectProcessor.
modObjectProcessor build key such way:
```php
if ($this->alreadyExists($name)) {
    $this->addFieldError($this->nameField, $this->modx->lexicon($this->objectType . '_err_ae', array('name' => $name)));
}
```
As result tv_err_ae that actually does not exist in lexicons. But exists tv_err_exists_name that uses in create/update. So easiest way just add new lexions, but I guess better to adopt duplicate processor for using already existing lang strings (and already translated).

himurovich:
Thanks @Alroniks for your deep investigation! Agree with you - it's better to adopt existing functional for that.

OptimusCrime:
Would you ( @himurovich and @Alroniks ) take a look at my PR in #13233 ?

I tried following your outline. Not sure if we agreed on the final solution or not.

OptimusCrime:
@modxbot close

PR #13233, opened 2017-01-04
Unify element already exist lexicon

OptimusCrime:
### What does it do?
Unifying lexicon used when element already exists. This includes templates, template variables, chunks, snippets and plugins. Keys for these lexicon entries are changed in all languages.

The processors:

- processors/element/create.class.php
- processors/element/update.class.php

Are adjusted to using the new lexicon keys.

### Why is it needed?
As described by @Alroniks, there are two ways MODX handles checking for duplicate elements. These two ways use different lexicon keys. Throughout almost the entire core, the format ``[type]_err_ae`` is used, as you can see [here](https://github.com/modxcms/revolution/search?utf8=✓&q=_err_ae).

Element duplicate processors all inherit from [one of the processors that use this naming format](https://github.com/modxcms/revolution/blob/5f59848ab1f1805ba09f03fae35bee5b7166f37b/core/model/modx/modprocessor.class.php#L977), but at the same time also override checking for duplicates during [creation/update](https://github.com/modxcms/revolution/search?utf8=✓&q=_err_exists_name).

My suggestion is (and what this PR intend to solve):

- Use the format ``[type]_err_ae`` across the board, moving away from ``[type]_err_exists_name``, which is only used for template, TVs, chunks, snippets and plugins.
- Fix ``processors/element/[update/create].class.php`` to use ``[type]_err_ae``.

### Related issue(s)/PR(s)
Reported in #13221


Alroniks:
Hi @OptimusCrime. I agree with your solution. It better to fix the wrong behavior of the system. I was wondering about current state of translates in Crowdin but I hope John will resolve it :)

OptimusCrime:
@Alroniks I am not sure how Crowdin works like that, but in this scenario the keys are incorrect which I think is solved here(?)

Hopefully someone knows better.

Mark-H:
In terms of translations, you should only include the English lexicon in pull requests. The others are synced with crowdin.

OptimusCrime:
@Mark-H I am aware of that, but I thought because we change the KEY associated with the entries, they needed to be pulled from GitHub? If I only need to change the keys for the english language I can go ahead and undo the rest of the changes.

Mark-H:
Maybe @theboxer disagrees but the way I normally use crowdin, I let that create the keys in the different languages again. Probably results in them losing their translation though..

theboxer:
Yes, it will result in loosing translations, but I'd go that way too...

OptimusCrime:
That's fine. I'll adjust the PR as soon as I get the time.

OptimusCrime:
Updated as per requested.

PR #13236, opened 2017-01-05
Update superboxselect.js

lexsmil:
Escape regular expression special characters

### What does it do?
Escape regular expression special characters in last query string

### Why is it needed?
Solving the problem that occurs when checking for new input values

### DIFF
--- a/manager/assets/modext/util/superboxselect.js
+++ b/manager/assets/modext/util/superboxselect.js
@@ -1503,7 +1503,7 @@ Ext.ux.form.SuperBoxSelect = Ext.extend(Ext.ux.form.SuperBoxSelect,Ext.form.Comb
     },
     shouldQuery : function(q){
         if(this.lastQuery){
-            var m = q.match("^"+this.lastQuery);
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Mind elaborate on just what

> Solving the problem that occurs when checking for new input values

means?

What is a problem (input/output/behavior) that is solved by this PR? Are there any related issues?

Thank you for submitting a PR btw :)

lexsmil:
Sorry for my bad english... Correct me please if I'm somewhere wrong.

The problem is that when using superboxselect, adding new item is checked with the items previously added, and if among the previous elements (which have already been successfully added) among the symbols there are symbols *\()? and other special characters for regular expressions error occurred when checking the match function.
For example, the string "Hotel ***** when you add the next element after it will generate an error and will not allow you to add more of any 1 item until you remove the "extra" characters.

Jako:
The regex looks fine. Thanks for sharing!

OptimusCrime:
Looks good, thank you for explaining @lexsmil

Jako:
Merged into 2.x/2.5.x

Issue #13223, opened 2016-12-21
File Unzip feature broken


sottwell:
### Summary
the Files unzip feature fails due to an invalid path, the site base path is shown as "undefined"

### Step to reproduce
Upload a zip file, then from the context menu choose Unzip File.

### Observed behavior
Error message
![screen shot 2016-12-21 at 16 39 21](https://cloud.githubusercontent.com/assets/1073780/21393265/62b174ee-c79c-11e6-8c06-c7b99c37c3db.png)

### Expected behavior
Unzip the file in the same location

### Environment
MODX Cloud, Revolution 2.5.2

igorsuhinin:
Hi @sottwell,

Thanks for your report.
Could you please check again if all is okay? I've accidentally uploaded the wrong file on modx.com.
Hopefully it should work now.

Bruno17:
I can confirm that issue

sottwell:
I believe this is related to #3170 and #3180

Bruno17:
it is.
https://github.com/modxcms/revolution/blob/b9ce93658cbfd8846f89714e0658eeb0129f559c/manager/assets/modext/widgets/system/modx.tree.directory.js#L604

OptimusCrime:
Fixed in #13238

OptimusCrime:
@modxbot close

PR #13238, opened 2017-01-05
File Unzip feature broken #13223

Bruno17:
### What does it do?
Removed MODx.config.base_path, which is undefined client-side.
Added base_path to the path server-side.

### Why is it needed?
File Unzip was broken as result of  #13170 and #13180

### Related issue(s)/PR(s)
Related to #13223 and #13183


Jako:
Thanks, merged into 2.5.x

Issue #13211, opened 2016-12-05
Cannot delete elements when reviewing (removable only from tree)


pixelchutes:
### Summary
In Revo there is currently no way to Delete an element from the edit/update page, a pretty important feature missing from early MODX Evo days. This pattern is already established for Resources in Revo, for example.

### Step to reproduce
1. Browse to the update page for any element (TV, Chunk, Snippet, etc)
2. After reviewing or deciding it is safe for removal, you cannot easily remove it from the system. (Missing <kbd>Delete</kbd>)

### Observed behavior
To delete an element, today you are forced to:

1. (optional) Reveal the tree
2. Jump to Elements tab
3. (optional) Expand the [TV|Chunks|Snippets|Plugins] Tab
4. (optional) Know (and find) the Element's Category container
5. (optional) Expand the container
6. Browse through results to (hopefully) find the correct element
    - _(Especially important if lots of results or similar naming convention in effect)_
7. Right-click
8. Select **"Delete [TV|Chunk|Snippet|Plugin]"**

### Expected behavior
There should be a way to delete an Element from the edit page. A <kbd>Delete</kbd> button, just like for Resources.

Not only is this far more convenient, it's a safer way to purge content from the system since you can actively be reviewing the very Element you're considering for deletion.

### Environment
MODX 2.5.2-dev

achterbahn:
Deleting a resource this way just marks the resource as deleted in the tree view. The resource itself stays open on the right side.
As for elements there is no functionality to mark them as deleted. Discussed here #764. 
When adding a delete button, what should happen after deleting the element? Loading the dashboard? 
As of writing this: wouldn't it be convenient to show a grid when clicking on elements category headers instead of collapsing/expanding it? There is the little arrow to perform this action.

pixelchutes:
Flagging an Element for deletion (e.g. Trash) would indeed be a cool feature, and welcomed addition.

This issue is more about not restricting Element deletion to the tree, requiring knowing a potential Category, and having the right collapsed state + right-click / context menu knowledge.

There should simply be a <kbd>Delete</kbd> button on the page you edit Elements from, similar to Resources. Should Element "trashing" feature be introduced in the future, saidn button could easily be adapted to support, but for now, it should mimic the current supported behavior (immediate deletion), just as it works from the Elements tree.

pixelchutes:
Classes by cc5ed4b

PR #13245, opened 2017-01-08
Add Delete button to chunk/snippet/plugins-window

achterbahn:
### What does it do?
Add Delete Buttons to the edit windows for chunks/snippets/plugins

### Why is it needed?
For convenience. Now it's possible to delete a chunk/snippet/plugin directly from the edit window. Not only through contextual menu in the tree.

### Related issue(s)/PR(s)
#13211


Jako:
Nice! A redirect to the manager url would be ok on success.

achterbahn:
I didn't add a button for deleting Templates and TVs. Was not sure if I just can use the delete method on them..

pixelchutes:
Looking good! Yes, if possible this would cover _all_ Elements, so Templates and TVs, too. You should be able to delete just the same, mimicking the current behavior from tree.

sottwell:
What about templates that still have a resource using them? And when deleting a TV, would all of its associated records in the other TV tables be deleted as well?

achterbahn:
Deleting a template where resources are bound to it throws a warning and is not deletable. Deleting a TV removes the TV and all the entries in the table site_tmplvar_contentvalues. 
The delete buttons should reflect this behaviour.

sottwell:
site_tmplvar_templates and site_tmplvar_access?

Jako:
@sottwell He uses the default remove processors, so this is covered

https://github.com/achterbahn/revolution/blob/c43a25483b0e39a0a6e8fa700f7f96ecc33475d1/manager/assets/modext/sections/element/chunk/update.js#L77-L80

manager/assets/modext/sections/element/chunk/update.js
```diff
@@ -64,5 +70,21 @@ Ext.extend(MODx.page.UpdateChunk,MODx.Component, {
...
+				text: _('chunk_delete_confirm')
+				,url: MODx.config.connector_url
+				,params: {
+						action: 'element/chunk/remove'
+						,id: this.record.id
+				}
+				,listeners: {
+						success: {
+							fn:function(r) {
+								MODx.loadPage('?'); //not sure about this
```
Jako: Try `MODx.loadPage(MODx.config.manager_url);`

achterbahn: Thanks.

achterbahn:
The remove processor (/core/model/modx/processors/element/template/remove.class.php) of a template checks if a resource is using this template.

In contrast the remove processor of TVs is not checking. I added a check if a TV is used by a template or a resource.
`      public function beforeRemove() {

      /* get tv relational tables */
      $this->TemplateVarTemplates = $this->object->getMany('TemplateVarTemplates');
      $this->TemplateVarResources = $this->object->getMany('TemplateVarResources');
      $this->TemplateVarResourceGroups = $this->object->getMany('TemplateVarResourceGroups');

        /* check if any template uses this TV */
        $tvts = $this->object->getMany('TemplateVarTemplates',array(
          'tmplvarid' => $this->object->get('id')
        ));
        if (count($tvts) > 0) {
          foreach ($tvts as $tvt) {
            $ms .= "<br />\n Template ID: ".$tvt->get('templateid');
          }
        }

        if ($ms != "") {
          return $this->modx->lexicon('tv_inuse_template').$ms;
        } else {
          return true;
        }

    }
`

OptimusCrime:
Awesome PR, just a few things about the code style:

- PHP code is indented with two spaces, the rest of the project uses four spaces
- The JS code is indented with tabs, the rest of the project uses four spaces

achterbahn:
you're absolutely right @OptimusCrime. I'm glad I could make the clone/commit/push things in git. Next step: working on the cosmetics :-) 
There should be something like a MODX Contribute workshop from time to time. But I hope I get more fluent with all the rules and stuff.
Thanks for your input.

Jako:
Better check if that processor is used somewhere else before changing it:
core/model/modx/processors/element/tv/remove.class.php

I think it will not be a BC, but it could.

Jako:
Merged into 2.x

PR #13262, opened 2017-01-24
Fix issue - not showing resources tree in modx manager.

avanturist888:
### What does it do?
This fix for error `TypeError: s[i].search is not a function` 
in  file `manager/assets/modext/widgets/core/tree/modx.tree.js`
 on line 421 `sr = s[i].search(p);`

### Why is it needed?
This is because first argument of `s` is not string. Better check "if `s[i]` is string or not" fix this issue.
Screenshot from chrome debugger:

### DIFF
--- a/manager/assets/modext/widgets/core/tree/modx.tree.js
+++ b/manager/assets/modext/widgets/core/tree/modx.tree.js
@@ -417,7 +417,7 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
                 var f = false;
                 var sr;
                 for (i=0;i<s.length;i++) {
-                    if (s[i] == undefined || s[i] == 'undefined') { s.splice(i,1); continue; }
+                    if (s[i] == undefined || typeof s[i] != 'string' ) { s.splice(i,1); continue; }
                     sr = s[i].search(p);
                     if (sr !== -1 && s[sr]) { /* dont add if already in */
                         if (s[sr].length > s[i].length) {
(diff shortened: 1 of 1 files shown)


pixelchutes:
I haven't personally noticed this particular error. Can you please elaborate on where in the Manager you are seeing this, and under what conditions does the error occur? _(e.g. version of MODX, browser/version, steps to reproduce, etc)_

vasia123:
This error very specific - only for long time sites, that created in 2.1.6 and updated until 2.5.+
Reproduces in all browsers. Check my screenshot.

OptimusCrime:
Does it happen on page load? Or what do you do to trigger the error?

vasia123:
This happened on trying to load resource tree. Chunks/templates/snippets and files looks fine.

jaygilmore:
If these are upgrades, this sounds like the issue of the tables not getting created at 2.3. What's in the error log? We've found that while it's supposed to happen that missing tables get added for every previous release that you may find you need to upgrade to 2.3.0 and then to 2.5.x. Can you see if this process resolves it if you have a 2.1.x site not yet upgraded?

OptimusCrime:
No feedback for more than 6 months. I suggest closing this.

Mark-H:
While in this case there is no steps to reproduce, it seems to be a fairly harmless change that only makes the code a bit better.

Mark-H:
Thank you @vasia123 for your contribution! Sorry it took so long to get merged.

Issue #9030, opened 2012-10-26
Set $modx, $modx->user, and $modx->resource early in Setup


BobRay:
bobray created Redmine issue ID 9030

If an upgrade saves a user or resource and a plugin connected to an invoked event calls get() on one of these, Setup will crash if these are not set.

This affects an early 2.2.x upgrade.

If possible, xPDO::OPT_SETUP should also be set earlier in Setup, because it's set too late to help with this problem.

Mark-H:
Always check if you received what you expect from any API. `if ($modx->getUser() && $modx->resource)`.

BobRay:
The point is that a plugin with a missing sanity check shouldn't cause setup to crash. The user has no idea what's happened, little hope of fixing things,  and lost confidence in MODX. 

I don't think there's any reason for plugins to be executing during setup. 

If xPDO::OPT_SETUP were set early, the code that runs plugins could be altered to skip them if xPDO::OPT_SETUP is set.

OptimusCrime:
Can we just turn them off without any side effects?

Mark-H:
It's still seems like a developer error to me, but sure I'll reopen if you'd like to send a pull request to fix it.

BobRay:
It *is * a developer error, but not one the user should suffer for, imo.

The fix is simple and I've done a PR. There will be a tiny performance hit, but it's a constant comparison, so it should be negligible.

pixelchutes:
I recently experienced an error during setup for the first time from a plugin, and ended up tracing it down to pdoTools. I had to disable the plugin and clear all of core/cache/* in order to proceed.

+1 for the recommended approach in the PR and for preventing unnecessary events during setup.

OptimusCrime:
Feels like a hack, and I've never encountered this before. However, if it solves a real problem and I'm +1 for implementing this. Great work with sending the pR @BobRay :)

pbowyer:
+1. Can we merge and close? #modxbughunt

Mark-H:
The pull request is not complete yet. When it's merged the issue will be closed too.

BobRay:
I edited the PR. I hope it's correct.

opengeek:
Are we sure this is the proper way to proceed? It seems to me there are some pretty valid scenarios where you would want plugins to execute during setup, especially if they are attached to resources and users.

BobRay:
Excellent point. 

[...]

PR #13271, opened 2017-02-01
Prevent plugins from crashing setup (#9030)

BobRay:
### What does it do?

Checks for xPDO::OPT_SETUP === true

### Why is it needed?
If an upgrade saves a user or resource and a plugin connected to an invoked event calls get() on one of these, Setup will crash if these are not set. This change prevents plugins from executing during setup.

I've checked all event invocations in the code and the only ones where false would not be acceptable involve RTE events, which would never execute during setup.

### Related issue(s)/PR(s)
   issue #9030

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1575,8 +1575,9 @@ class modX extends xPDO {
      * @return bool|array
      */
     public function invokeEvent($eventName, array $params= array ()) {
-        if (!$eventName)
+        if (!$eventName || xPDO::OPT_SETUP === true) {
             return false;
+        }
         if ($this->eventMap === null && $this->context instanceof modContext)
             $this->_initEventMap($this->context->get('key'));
         if (!isset ($this->eventMap[$eventName])) {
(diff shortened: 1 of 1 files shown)


Mark-H:
Sorry if I'm being daft, but I don't understand what this achieves... xPDO::OPT_SETUP is a constant that's defined as a value of `setup` - that would never get a value of `true`?

pixelchutes:
I think it gets set during setup in **setup/includes/modinstall.class.php:**  [here](https://github.com/modxcms/revolution/blob/2.x/setup/includes/modinstall.class.php#L444) or maybe [here](https://github.com/modxcms/revolution/blob/2.x/setup/includes/modinstall.class.php#L473) ?

Mark-H:
The PR only tests the constant though, not an option with that key. Probably should be a check on getOption instead?

pixelchutes:
Ah, I'm with you now. Thoughts @BobRay?

Jako (changes requested):
```
if (!$eventName || $this->getOption(xPDO::OPT_SETUP) === true)
```

Mark-H:
@BobRay Could you please update the PR to do what it's supposed to do per Jako's review?

BobRay:
Mark, I just did that yesterday, unless I did something wrong.

Let me know.

Mark-H:
Thanks @BobRay!

PR #13274, opened 2017-02-03
Re-style the templated package provider thumbnail grid

Mark-H:
### What does it do?
Re-styles the thumbnail grid that is used in the package management. 

If you're thinking "wasn't that a standard table grid?!", then you are indeed correct. The thumbnail grid is only used for categories in the package providers that return a `templated` flag with a value of 1/true. With the standard MODX.com package provider, that includes the _Front End Templates_ and _Manager Templates_ categories; the main _Extras_ category uses the grid view at the moment. 

Before/after screenshots can be found below.

### Why is it needed?
Make it more appealing to the user. The standard packages grid offers very little room for extra information visible before the user downloads a package (e.g. if a user has licenses available), and the thumbnail grid is, well, pretty uninspired. 

The thumbnails grid with these changes offers a big thumbnail that people can actually see, and also shows a short 2-3 line description right there in view. These two things alone can make a huge difference in people understanding what an extra is for, and may also prompt developers to update their extra listings if it shows the all-too-common "Initial content", or "no preview".  

This restyle also adds a new bar for packages that are marked as `featured` by the package provider. At the moment that bar is rather useless as every extra is marked as featured, but [Jason/John will look into that](https://modxcommunity.slack.com/archives/general/p1486153851005634), so I expect that to be solved before this change makes it into a distribution.

If this new styling is accepted and well-received, the MODX.com team could consider changing the provider to have it use this view for standard extras as well, by changing the `templated` value to 1/true in the `/repository` response. When this ships in a release, I intend to update the modmore provider to use this view for the extras.  

### Related issue(s)/PR(s)
None that I could find.

If this is accepted, I might send another pull request in the future that makes it possible for the package provider to return an optional `featured_text`, which should then be used for the featured bar instead of the lexicon. That could be used by the provider to say things like "New!", "Now on sale!", or "Recently updated".

### Screenshot Before

screen shot 2017-02-03 at 23 27 00

### Screenshot After

screen shot 2017-02-03 at 23 34 24


rtripault:
Not really sure why, but having some issues on the "manager templates" section
screenshot_2017-03-06_12-32-23

Otherwise :heart:

Mark-H:
@rtripault I'm not seeing that anywhere I look. Is that the first page of manager templates > administration? What browser? 

The problem looks to be that "freelancewebdev" is too long of a name :p

Mark-H:
I've rebased and updated the pull request. There was a conflict, so maybe that has to do why you were seeing that issue but I wasn't.

rtripault:
Thanks Mark, will check that out ASAP.

I don't exactly recall which browser i was testing on, but changes are that was Chrome/Debian.
Will keep you posted (hopefully today)

rtripault:
All good, many thanks!

PR #13281, opened 2017-02-09
Add timings for admins into Chrome dev tools

bezumkin:
### What does it do?
This code will add special response header with timings for users, that logged into manager.

### Why is it needed?
Admins will able to quickly look to the speed of page generation.

devtools

### DIFF
--- a/core/model/modx/modresponse.class.php
+++ b/core/model/modx/modresponse.class.php
@@ -94,6 +94,13 @@ class modResponse {
             $queryTime= $this->modx->queryTime;
             $queries= isset ($this->modx->executedQueries) ? $this->modx->executedQueries : 0;
             $phpTime= $totalTime - $queryTime;
+            if ($this->modx->user->hasSessionContext('mgr')) {
+                header('Server-Timing:' . implode(', ', array(
+                    't=' . number_format($totalTime, 6) . '; "Total time"',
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Very nice! Do you know if MODX exposes more timing related variables? It could be interesting to see how long MODX used to parse the content too.

bezumkin:
I think you can install special extra for this. DebugParser for example
debugparser

Omeryl:
Can't this just be put into an extra, do we need to maintain this in core?

bezumkin:
Why not put this into core? It is one additional header for admins. 

That values has already been received, so there is no speed issues.

opengeek:
Why not put it in a plugin? That's why we have a plugin system.

bezumkin:
Okay, okay.

Sorry.

zaigham:
Come on guys, MODX is already generating this information, this code just show it on devtools. Making a plugin for this is nice, but overkill IMO.

Omeryl:
@zaigham The only reason I brought it up is because my fear is: what if Chrome decides to change how the structure is, and it breaks? That all of a sudden requires an entire core update to fix it. Whereas as an extra/plugin, it can be updated independently.

jaygilmore:
It should not be implemented in a way that it's blocking if browser behavior changes.

Issue #12842, opened 2016-01-08
Can't delete resource from resouce group from manager


Kookabura:
When there are multiple equal resources in different group you can't choose the wanted one by click. That happens because of equal id for resources.
The bug might be fixed by replacing line 59 at core/model/modx/processors/security/resourcegroup/getnodes.class.php
`'id' => 'n_'.$resource->get('id').'_'.$resourceGroup->get('id'),
`

Mark-H:
Thanks for the report, a fix is pending merge in #13283.

sdrenth:
@Mark-H This issue can be closed according to
https://github.com/modxcms/revolution/pull/13283

Mark-H:
We tend to keep issues open until a fix has been merged, and we don't merge our own pull requests ;)

rtripault:
Closing as it's merged for 2.5.6

PR #13283, opened 2017-02-09
Fix issue deleting resource from resource groups tree if it is in multiple groups (#12842)

Mark-H:
### What does it do?
Adds the resource group ID to the ID used in the tree markup. 

### Why is it needed?
Go to Content > Resource Groups and assign a resource to two different resource groups. Because the ID was solely based on the resource ID, there was an ID conflict that ExtJS can't handle, so anything you try to do is only applied to one of the resource IDs. 

By adding in the resource group ID, they have unique IDs again, and everyone's happy. Even ExtJS.

### Related issue(s)/PR(s)
Reported in #12842


Flowan:
@Mark-H Tested this and can confirm this is fixed #modxbughunt

rtripault:
This is supposedly merged for 2.5.6 (thanks), so closing

Issue #13280, opened 2017-02-09
it is not possible Injects css or js via plugin at the login page


FerX:
### Summary
it is not possible Injects css or js via plugin at the login page

### Step to reproduce
view file login.tpl 
missing the code:
```{$maincssjs}
{foreach from=$cssjs item=scr}
{$scr}
{/foreach}
```

### Environment
MODX 2.5.4

alroniks:
Closing, as proposed pull request was declined. It means that this task becomes "won't fix".

PR #13284, opened 2017-02-09
fix issue it is not possible Injects css or js via plugin at the login page (#13280)

FerX:
### What does it do?
Insert missing code.

### Related issue(s)/PR(s)
#13280 (it is not possible Injects css or js via plugin at the login page)

### DIFF
--- a/manager/templates/default/security/login.tpl
+++ b/manager/templates/default/security/login.tpl
@@ -26,6 +26,11 @@
     <script src="assets/modext/sections/login.js" type="text/javascript"></script>
 
     <meta name="robots" content="noindex, nofollow" />
+
+{$maincssjs}
+{foreach from=$cssjs item=scr}
+{$scr}
+{/foreach}
 </head>
 
 <body id="login">
--- a/manager/templates/default/security/logout.tpl
+++ b/manager/templates/default/security/logout.tpl
@@ -33,6 +33,12 @@
 	var SITE_NAME = '{$_config.site_name|strip_tags|escape}';
 	var CONNECTORS_URL = '{$_config.connectors_url}';
 	</script>
+        
+{$maincssjs}
+{foreach from=$cssjs item=scr}
+{$scr}
+{/foreach}
+
 </head>
 <body id="login" style="background-color: #fffffa;">
 


Flowan:
@Mark-H Tested this and can confirm this is fixed #modxbughunt

Mark-H:
Hm, here's an interesting one. This does what it says it does on the tin, but it also seems to be loading assets from extras that don't necessarily belong on the login page (in my test, image plus). These then trigger javascript issues because they can't get to some manager variables. Login still seems functional, but I can imagine this might cause unexpected issues depending on what gets loaded there.

I can see the benefit to this fix, but wouldn't want to see 2.6 causing all sorts of issues on the login screen. Does anyone have any thoughts?

rtripault:
When i want to tweak the login layout, i usually create a dedicate theme and change the smarty template to fit my needs. But that's definitely overkill and adding extra JS/CSS files using a plugin would probably be more convenient!

I'm also torn about the proposed solution if it loads JS/CSS from other extras (and potentially "leak" the fact the site is Revo powered).

Although i have no better proposal, for the time being.

mrhaw:
What about having 2 manager settings for CSS/JS

OptimusCrime:
How do you get CSS/JS/HTML to show up in these blocks? With the `addJavascript` and `addCSS` methods in the controllers/plugins?

If so, this could potentially be solved by adding an additional parameter:

`function addJavascript(string $file, boolean $addToLoginScreen = false)`

Also fixes BC.

rtripault:
Sounds like a sensible proposal @OptimusCrime but this would mean checking if we actually are in the defined controllers for every `addCss` `addJavascript` & co methods... not sure i'm fine with that

OptimusCrime:
Why? Why would you need that? Could it not be a new feature to add stuff to the login screen?

rtripault:
Because that's how the `add*` methods work : it stacks some html in `modManagerController::head`, which is processed in `modManagerController::registerCssJs` (to be set as a Smarty variable named `cssjs`).

So with your proposal, we would have to check if where are in the login/logout controller anytime we `add*` and only stack up the content in if `$addToLoginScreen === true` or if not in the login/logout controller.

Hoping my explanation attempt makes sense!

OptimusCrime:
I think it can be done much easier than that. For example using the method `addCss`:

```
public function addCss($script, $addToLogin = false) {
    if ($addToLogin) {
        $this->login['css'][] = $script;
    }
    $this->head['css'][] = $script;
}
```

Then the array `::login` is ordered and fixed exactly like `::head`. In the template file for the login/logout controllers we use and loop `login` and while logged in we only use `head` discarding the content in `login`.

This would not require us to check any instances or identify what controller is used. However, this may lead to unexpected behavior, as while on the login controller, a call to `addCss('path/to/file.css', false)` will not output anything.

Thoughts?

rtripault:
This is probably the cleaner option...still we would need to modify `modManagerController::registerCssJs` to handle the login/logout cases (or duplicate the method to support using your proposed `login` array )

> a call to addCss('path/to/file.css', false) will not output anything.

Like it is as of today ;)

All in all, i guess i'm not up for crafting something special to only add a small benefit (2 controllers!), guess that's the 80/20 law.

OptimusCrime:
Yes, you are right about that. Although this feature might be good, I dont know just what it is supposed to be used for? However however, the implementation I suggest should not result in significant overhead as far as I can tell.

Mark-H:
I'm going to close this PR for the reasons stated earlier (assets meant only for in the manager leaking into the login page unintentionally). I think it's a good idea to allow plugins to more easily register to the login page, it just needs to be executed differently. 

Thanks @FerX for contributing. Even though it didn't make it into the core, your help is much appreciated :)

FerX:
In this case then it would be useful to set some settings to change the logo and other currently static information in tpl

PR #13288, opened 2017-02-11
Update file.inc.php

FerX:
### What does it do?
Removed leading space in tags

### Why is it needed?
tags modx not parsed


sottwell:
Isn't there another place for dealing with lexicon files?

FerX:
@sottwell opss.. sorry.

sottwell:
I'm not sure if that's where fixes to established translations get done, that's why I asked.

Mark-H:
Yes Susan is correct, translation changes go in crowdin. Only source file changes go directly in github.

Issue #13292, opened 2017-02-14
Files Tree images broken when media source above doc root


labr1005:
### Summary
When using media sources that sit above the doc root:
- the preview images, when hovering over the filename in the Files Tree, are broken.
- the preview images in the MODX browser incl. zoom are rendered.

![bildschirmfoto-2017-02-14-um-18 36](https://cloud.githubusercontent.com/assets/1913250/22941359/76df87ca-f2e5-11e6-8b3e-ff4222b0ef4e.jpg)

### Step to reproduce
System setting phpthumb_allow_src_above_docroot Yes.

Create a media source above the doc root with
- an absolute basePath and basePathRelative No.
- an absolute baseUrl and baseUrlRelative No.

### Observed behavior
The preview images, when hovering over the filename in the Files Tree, are broken.

### Expected behavior
The preview images, when hovering over the filename in the Files Tree, should be rendered.

### Environment
MODX 2.5.5
Apache 2.2.29
PHP 7.0.8

joeke:
I'll look into this #modxbughunt

Flowan:
I'm going to try and fix this today. #modxbughunt

labr1005:
But there's already a pull request for this.

Flowan:
@labr1005 Oh wow, missed that completely. Thanks, will continue to the next issue :D

PR #13293, opened 2017-02-15
Fixed issue #13292 Files Tree images broken when media source above doc root #modxbughunt

labr1005:
### What does it do?
Images for the file tree are now rendered with the phpthumb connector.
- This will also display thumbnails for images in media sources that sit above the doc root (when system setting phpthumb_allow_src_above_docroot Yes).
- Before, the displayed images where the "original" images. Now, they are scaled down with phpthumb if larger than system setting filemanager_image_width or filemanager_image_height.

### Why is it needed?
Fix for Files Tree images broken when media source above doc root.
Scale down images with phpthumb for a faster manager.

### Related issue(s)/PR(s)
modxcms/revolution#13292

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -220,6 +220,8 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
 
                     if (in_array($ext, $imagesExts)) {
 
+                        $modAuth = $this->xpdo->user->getUserToken($this->xpdo->context->get('key'));
+
                         $imageWidth = $this->ctx->getOption('filemanager_image_width', 400);
                         $imageHeight = $this->ctx->getOption('filemanager_image_height', 300);
                         $thumbnailType = $this->getOption('thumbnailType', $properties, 'png');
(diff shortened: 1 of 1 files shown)


sottwell:
Works perfectly for me on localhost 

Revo 2.5.4-pl
MAMP - Apache, PHP 7.1.0

Jako:
Nice one! Thanks for your fix.

Mark-H:
Seems suitable to be included in a patch release, doesn't it @Jako? Or am I overlooking something that would need to push it to 2.6?

Jako:
Could go into 2.5.6 too. I was not sure if this could be marked as a bug or as an enhancement.

sottwell:
I'd call it a bug, since the Media Browser doesn't have that problem.

OptimusCrime:
I would argue that the broken images are not by design, hence its a bugfix.

labr1005:
Help.

labr1005:
Oops. Sorry.

labr1005:
Damn it!

Flowan:
@Mark-H This can be merged #modxbughunt

labr1005:
:)

PR #13294, opened 2017-02-16
Remove listener when a window form panel is destroyed

rtripault:
## What does it do?

Removes a JS listener when a `MODx.Window` form panel is destroyed

## Why is it needed?

After closing a `MODx.Window` (in example a "quick edit" window) and resizing the browser viewport, we get a JS issue because we were trying to set a style in a destroyed DOM element (see `MODx.Window#resizeWindow`)


Mark-H:
Thanks @rtripault - cherry picked into 2.5.x in 8cd4e07f91d840a4716e5646f06c09baa0f4acf8

PR #13295, opened 2017-02-16
Fix issue where Media Source relative paths were not "expanded"

rtripault:
## What does it do?

Expand Media Source path when relative

## Why is it needed?

When using a relative path with `../` or `./` the absolute path was not expanded, resulting in errors (in example not being able to upload a file)


Mark-H:
Devil's advocate here, should relative paths like that be accepted? It's in the media source configuration, which is different from accepting paths like that in a request, but what do you think?

sottwell:
I'd say they should be, I've used that kind of path notation a time or two myself.

rtripault:
Not sure if we "should", but the point (at least from my seat) is to prevent having absolute paths, which could be different depending on the project environment. Finding this could be convenient... happy not to have it in the core tho if you think that does not worth it/there are more drawbacks than convenience.

Mark-H:
Thanks @rtripault!

PR #13296, opened 2017-02-16
Make sure we have a field before trying to store the focused resource field

rtripault:
## What does it do?

Makes sure `field` is defined

## Why is it needed?

When the "field" involved is not a field (in example the resource group grid), it causes a JS error because `o.field` is `undefined`


Mark-H:
Cherry-picked into 2.5.x for 2.5.6, thanks Romain!

PR #13297, opened 2017-02-17
Fix array_key_exists PHP warning

Jako:
### What does it do?
Check `$this->aliasMap`/`$ctx->aliasMap` before it is used in array_key_exists

### Why is it needed?
Avoid the following errors in the MODX error log
`PHP warning: array_key_exists() expects parameter 2 to be array, null given`
when the aliasMap is not available.


rtripault:
Wouldn't `isset($array[$key])` fix the warning while keeping a single "check" ?

Jako:
Maybe both (the array itself and the key of the array) have to be checked by isset. isset could be faster than array_key_exists().

Jako:
Just have thought about this. The check has to run with array_key_exists to have no side effects. If `aliasMap[$uri]` is set and is null for some reason, `$resourceId` will be 0 instead of false.

PR #13299, opened 2017-02-17
Fix to dirty buggy css hack.

avanturist888:
### What does it do?
This hack very buggy, and can be fixed. 
Key of this fix - `margin-right:-170px;` and `float: left;padding-left:170px;`. 
`width: 100%;padding-right: 15px;box-sizing: border-box;` - is for better visiblity of panel.

### Why is it needed?
Because of that: 

And what happened after this fix:

### Related issue(s)/PR(s)
n/a

P.S. Not shure how to right way to fix this in index.min.css - can someone explain this?

### DIFF
--- a/manager/templates/default/css/index.css
+++ b/manager/templates/default/css/index.css
@@ -6135,10 +6135,7 @@ ul.x-tab-strip-bottom {
     border-right: 1px solid #E4E4E4 !important;
     /* ovverride extjs default theme */
     float: left;
-    margin-bottom: -10000px;
-    /* dirty hack to make vertical tabs container stretch to bottom */
-    padding-bottom: 10000px !important;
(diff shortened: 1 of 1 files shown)


Jako:
It will be grunted in, so you don't have to do that.

Jako:
But please change it here and not in the css file: https://github.com/modxcms/revolution/blob/2.5.x/_build/templates/default/sass/_tabs.scss

vasia123:
https://github.com/modxcms/revolution/pull/13300 - i create another pull request for this.

Jako:
Thanks for the other PR. If you want to check the grunted result, you could build the template it on your own by reading the following readme: https://github.com/modxcms/revolution/tree/2.x/_build/templates/default

PR #13300, opened 2017-02-17
Fix to dirty buggy css hack.

avanturist888:
### What does it do?
This hack very buggy, and can be fixed. 
Key of this fix - `margin-right:-170px;` and `float: left;padding-left:170px;`. 
`width: 100%;padding-right: 15px;box-sizing: border-box;` - is for better visiblity of panel.

### Why is it needed?
Because of that: 

And what happened after this fix:

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/_build/templates/default/sass/_tabs.scss
+++ b/_build/templates/default/sass/_tabs.scss
@@ -299,8 +299,7 @@ ul.x-tab-strip-bottom {
     background: $coreFieldBg !important; /* ovverride extjs default theme */
     border-right: 1px solid $borderColor !important; /* ovverride extjs default theme */
     float: left;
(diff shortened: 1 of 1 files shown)


Mark-H:
While this gets rid of the ugly hack, it does not keep the proper behaviour for the styles.

This is how it's supposed to look (and what `2.5.x` looks like):
screen shot 2017-02-21 at 14 52 21

This is with your patch:
screen shot 2017-02-21 at 14 51 01

Note the sidebar background and border is missing with your patch applied. 

I'm going to close this pull request, but feel free to open a new one that gets rid of the ugly hack while it still looks right :)

Issue #12666, opened 2015-09-23
New dependencies installation fails for some Extras (MIGX, SEO tab, SEO Pro)


DESIGNfromWITHIN:
When using the new dependencies feature in MODX 2.4.0-pl to install MIGX, SEO Pro or SEO Tab, The extra's get listed but when you click install they do not get installed.

The only way to install them is via the 'normal' way, after this they get listed correctly as installed under dependencies tab.

I tried different names, but no go for these three extras. Others do work fine...

joeke:
I've tested this with Modx 2.4.2 and I've got the same issue. I've dug into the problem and the issue seems to be with spaces in the dependency package name. See issue #55 at the SEO Tab repo for more info: https://github.com/Sterc/SEOTab/issues/55

joeke:
@opengeek I've had a talk with you (on Slack) about this, and as you pointed it seems to be an issue in the modx extras repository when using dependencies with a space in the package name.. Are there any plans / ideas on fixing this? Would love to help out, but I don't think the Extras repository code is publicly available, is it?

Mark-H:
Closing as this likely needs to be resolved in the provider/extras site.

opengeek:
Package names never have spaces in them; if you submit them that way, they are cleansed. They have to match the transport filename. I'm a little confused as to what the problem really is here.

opengeek:
@joeke I recall the conversation but, unfortunately, I don't recall the conversation details. If you can provide more details here that might help. But again, packages should *never* contain a space in the name. They *should* always be lowercase and not contain a dash character, just like the transport filename rules.

joeke:
@opengeek I've managed to reproduce the issue, here are the steps:
1. Build new transport zip for any extra, but add SEO Pro to 'requires' array (key: 'SEO Pro', value: '>=1.0.4'). (please note that I use the package name here, because that's how the `checkDependencies` and `checkDownloadedDependencies` functions in `modtransportpackage.class.php` handle the dependencies.)
2. Install the newly created transport zip in a clean modx install (at least without SEO Pro installed).
3. The installer screen shows the Dependencies tab, and it shows SEO Pro, with a download button.
4. Click on the download button, the request fails with following error message:
`{"success":true,"message":"Could not download and create transport package with signature: ","total":0,"data":[],"object":[]}`.

My request URL is 
[...]

PR #13304, opened 2017-02-21
Fixing modTransport checks

Jako:
### What does it do?
Changed/enhanced the xPDO Query that locates previous installed package versions and dependent packages.

### Why is it needed?
- The check for previous installed packages did not found any installed packages.
- The dependent package check has issues with package names different than the signature. The dependency has to contain the signature without the version to pass the check.

### Related issue(s)/PR(s)
#12666

### DIFF
--- a/core/model/modx/transport/modtransportpackage.class.php
+++ b/core/model/modx/transport/modtransportpackage.class.php
@@ -497,7 +497,8 @@ class modTransportPackage extends xPDOObject {
                 'modTransportPackage',
                 array(
                     array(
-                        "UCASE({$this->xpdo->escape('package_name')}) LIKE UCASE({$this->xpdo->quote($package)})"
+                        "UCASE({$this->xpdo->escape('package_name')}) LIKE UCASE({$this->xpdo->quote($package)})",
+                        'OR:signature:LIKE' => $package . '-%'
                     ),
                     'installed:IS' => null,
                 )
--- a/core/model/modx/transport/modtransportpackage.class.php
+++ b/core/model/modx/transport/modtransportpackage.class.php
@@ -724,7 +724,11 @@ class modTransportPackage extends xPDOObject {
      */
     public function previousVersionInstalled() {
         $this->parseSignature();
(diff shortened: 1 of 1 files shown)


Mark-H:
This does not seem to fix the problem... I hijacked flexibility5 and added Seo Pro and seopro to the list of required packages (by editing the manifest.php in the unzipped package for easy testing), and the download still fails. 

When [searching the repo for Seo Pro](http://rest.modx.com/extras/package/versions?package=Seo%20Pro&constraint=%3E=1.0.4) it doesn't return results, but it does when [searching for seopro](http://rest.modx.com/extras/package/versions?package=seopro&constraint=%3E=1.0.4), but this PR doesn't make it download the dependency either way.

Jako:
As far as I remember, this fixed the installation of a already downloaded package. The package download could have issues, that could only be solved on server (provider) side.

sdrenth:
@Jako Could you provide a valid test case on how to test this PR on if it is working or not?

opengeek:
I don't understand why you would change this include a LIKE query. This PR needs a further description as well as some example of why this is necessary. As is, this would seem to match packages with similar names which could be entirely wrong in some cases.

Jako:
I will prepare a use case in the next days maybe with an example package. The like addition will match the signature of a package which has a more restricted charset than the package name. Since the signature can’t contain a `-` there can’t be wrong matched packages, if I am right.

opengeek:
The signature before the first occurrence of `-` *is* the package name, case-insensitive. This is the first rule of transport packaging.

Jako:
Sorry I meant what you said. You can’t catch a wrong package with the signature before a `-`.

As far as I remember, I had issues installing Image+ as dependency. Or maybe SEO Pro. The dependency check did go though with the patch, if the ‚signature‘ package was locally available.

And if I remember right, there is some mismatch between signature and package name. There is not a real reason, why a signature has to be a lowercase version of the package name. Why could you create (or maybe rename) a package in MODX extras with differences there?

But as I said, I try to prepare a valid use case in the next days.

Mark-H:
Any update here @Jako? :)

Issue #13278, opened 2017-02-08
2.5.5 logs "Resource with id <id> not found in context <calling context>" on valid cross-context links


Mark-H:
### Summary
The 2.5.5 introduced a much appreciated log for bad links (#13268), but it's a bit too enthusiastic. Links to a resource within another context are logging an error even though they are valid.  

### Step to reproduce
Create a 2nd context, and add a resource in it. Link from web to the resource in that context with the `[[~123]]` link tag. 

### Observed behavior
Error messages in the log. For example three header links pointing to resources in another context:

````
[2017-02-08 16:46:39] (ERROR in modContext::makeUrl @ resource 605) Resource with id 322 was not found in context web
[2017-02-08 16:46:39] (ERROR in modContext::makeUrl @ resource 605) Resource with id 101 was not found in context web
[2017-02-08 16:46:39] (ERROR in modContext::makeUrl @ resource 605) Resource with id 493 was not found in context web
[2017-02-08 16:50:48] (ERROR in modContext::makeUrl @ resource 169) Resource with id 322 was not found in context web
[2017-02-08 16:50:48] (ERROR in modContext::makeUrl @ resource 169) Resource with id 101 was not found in context web
[2017-02-08 16:50:48] (ERROR in modContext::makeUrl @ resource 169) Resource with id 493 was not found in context web
````

Links are valid and functional.

Actually broken links are logged as such:

````
[2017-02-08 16:50:48] (ERROR in modContext::makeUrl @ resource 169) Resource with id 111 was not found in context web
[2017-02-08 16:50:48] (ERROR in resource 169 @ /path/to/core/model/modx/modparser.class.php : 1353) Bad link tag `[[~111]]` encountered
````

which makes me believe that we could drop the error triggered in makeUrl. Alternatively, the check in modContext->makeUrl should be moved to modX->makeUrl until after it automatically discovers the context that is being linked to, but Jason mentioned that was troublesome.

### Expected behavior
No error log entry for valid cross-context links. 

### Environment
MODX Revolution 2.5.5. Not server specific other than the requirement for multiple contexts.

sottwell:
Yes, the 'bad links' error is nice, but the 'not in context' error is just a nuisance.

sottwell:
Maybe make it an info or a debug instead of an error?

gravino81:
I also encountered the same problem
`
[...]

PR #13305, opened 2017-02-21
Lower log level for 'resource with id not found in context' to info [#13278]

Mark-H:
### What does it do?
Changes the modContext->makeUrl resource-not-found logging introduced in #13268 from an ERROR to an INFO message. 

### Why is it needed?
As detailed and discussed in #13278, the message triggers a fair amount of false-negatives on multi-context sites where you're linking from one context to the other without explicitly referring to the right context. `modX->makeUrl` will automatically fetch the right context after the message is logged. This gets annoying, fast. 

While it's a useful message, the logging in modLinkTag (also added in #13268) is more accurate for the majority of use cases and continues to be logged as an ERROR. 

### Related issue(s)/PR(s)
Introduced in #13268 
Problem discussed in #13278

### DIFF
--- a/core/model/modx/modcontext.class.php
+++ b/core/model/modx/modcontext.class.php
@@ -313,7 +313,7 @@ class modContext extends modAccessibleObject {
                 }
             } else {
                 $this->xpdo->log(
-                    xPDO::LOG_LEVEL_ERROR,
+                    xPDO::LOG_LEVEL_INFO,
                     "Resource with id {$id} was not found in context {$this->key}",
                     '',
                     __METHOD__,


digitalbricks:
@Mark-H Thanks for working on that issue! 
Only for my understanding: Lowering the log level from `LOG_LEVEL_ERROR` to `LOG_LEVEL_ERROR` doesn't prevent this message from being written in the error log completely but "formats" it as an _INFO_, not an _ERROR_? So this line will still be written in the log, causing a growth in filesize day by day? Or is there a way (a setting) to prevent _INFO_ from being written in the error log?

sottwell:
That depends on the log level you specify in the system settings. The default to log to ERROR. https://docs.modx.com/xpdo/2.x/class-reference/xpdo/xpdo.log
screen shot 2017-05-04 at 22 47 52

digitalbricks:
@sottwell Thank you very much! I've overlooked that setting.

isaacniebeling:
@Mark-H I think this also needs to be applied at `/core/model/modx/modparser.class.php : 1353`

Mark-H:
I didn't think so @isaacniebeling as that one is far less likely to cause false positives, but feel free to open a new issue with details why.

PR #13306, opened 2017-02-22
Patch parsing tag on ellipsis

goldsky:
### What does it do?
This fixes truncated modx's tag when using ellipsis output filter

### Why is it needed?
To avoid revealing modx's tag because the filter truncates the whole tag's call
eg:
In getResources's tpl, you have this output
```
[[+content:ellipsis=`100`]]
```
If there's a chunk/snippet gets cut off, then it will show
```
Morbi cursus tellus augue, nec mattis magna laoreet non. Aliquam nec [[$chunk?...
```
### Related issue(s)/PR(s)
https://forums.modx.com/thread/96866/ellipsis-not-working-as-expected#dis-post-524087

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -326,6 +326,13 @@ class modOutputFilter {
                             $limit= intval($m_val) ? intval($m_val) : 100;
                             $pad = $this->modx->getOption('ellipsis_filter_pad',null,'&#8230;');
 
+                            /* parse all elements first, avoid MODX's tags get truncated */
+                            if ($this->modx->getParser()) {
+                                $maxIterations = intval($this->modx->getOption('parser_max_iterations', array(), 10));
+                                $this->modx->parser->processElementTags('', $output, true, false, '[[', ']]', array(), $maxIterations);
+                                $this->modx->parser->processElementTags('', $output, true, true, '[[', ']]', array(), $maxIterations);
+                            }
+
                             /* ensure that filter correctly counts special chars */
                             $output = html_entity_decode($output,ENT_COMPAT,$encoding);
                             $len = $usemb ? mb_strlen($output,$encoding) : strlen($output);


Jako:
Your patch is quite 'expensive'. 

A different solution similar to `[[+placeholder:striptags:ellipsis]]` could be the following modifier chain `[[+placeholder:stripmodxtags:ellipsis]]`

goldsky:
1st of all, when in conjunction with striptags, the ellipsis must come first.
order must be: `[[+placeholder:ellipsis:striptags]]`

1. render modx's tags first, then cut the output
2. then strip all output's HTML

`[[+placeholder:stripmodxtags:ellipsis]]` itself is a different output, it removes the modx's tag entirely, not renders it.

Jako:
The striptags output filter is just a call for the php strip_tags method: https://github.com/modxcms/revolution/blob/6591097d0ad43af28582e6154392cf2549e90131/core/model/modx/filters/modoutputfilter.class.php#L266-L275. It should run before ellipsis to have no invalid HTML tags after the truncation of ellipsis.

It is quite expensive to execute the MODX parser on the placeholder content by default inside of ellipsis. It would be better to make this optional by adding an own `parse` snippet output modifier, that contains your parsing code and call it like this: `[[+placeholder:parse:ellipsis]]`

Snippet code of `parse`
```
<?php
$maxIterations = intval($modx->getOption('parser_max_iterations', array(), 10));
$modx->parser->processElementTags('', $input, true, false, '[[', ']]', array(), $maxIterations);
$modx->parser->processElementTags('', $input, true, true, '[[', ']]', array(), $maxIterations);
return $input;
}
```

goldsky:
Would that be custom or built in, then?

Jako:
Custom with the code above.

OptimusCrime:
Could it be possible to look for potential MODX tags? Like doing a search for `[[*]]` in the string? I agree that introducing the parser is a heavy patch, but it does solve something that the default behavior now breaks.

goldsky:
Well, searching the tags is what processElementTags do, isn't it?

Bruno17:
I'm with Jako. Parsing the whole thing in any case by default can lead to unpredictable performance problems. I think, chaining outputfilters, where needed, is the better solution.

OptimusCrime:
Could we have a built in output modifier called parse? It would be nice if this feature was added in the core. Note that this might be BC, so it would need to target 3.x in that case.

Other than that I'm still in favor of implementing this as suggested. I'm not sure how big performance impact the parser adds, but I can't imagine it being that significant, especially because this is for the ellipsis modifier, not then or else.

Jako:
An internal parse modifier could be possible.

Mark-H:
The parser is only invoked on the ellipsis content, not every single output filter, so I think that extra overhead wouldn't be too much. For users that rely a lot on ellipsis, being protected against breaking the page (which is what broken tags like that could do) is a good thing. 

That puts me in favour of this change, but am only a 3/10 on the [url=http://blog.capwatkins.com/the-sliding-scale-of-giving-a-fuck]SSGF[/url].

OptimusCrime:
I'm in favor of the change too.

sdrenth:
Tested this fix and I can confirm that it's working and I'm also in favor of this change.

opengeek:
This sure seems like it is circumventing the parsing process by forcing parsing of noncacheable tags regardless of when the output filter is being executed. Is no one else concerned by this?

Jako:
Good point. This should be switchable by $options.

Mark-H:
There's definitely no consensus for this change currently - even I am personally torn between whether this is an improvement or something that's going to bite us in the butt two days after it is released. People may be relying in some weird way on the current parsing behavior, and currently there is no way to turn this on or off.

I'd like to suggest we go with @Jako's proposal of a separate (built-in) `parse` output filter. For people that use the `ellipsis` on places that may contain MODX tags, the `parse` output filter can be chained before ellipsis like `[[*introtext:parse:ellipsis]]`. That makes the behaviour opt-in, and regarding Jason's point of it circumventing the regular parsing, the opt-in makes that an acceptable trade-off in my opinion. Sounds good to all?

Unless there's any other heavy weighing feedback on this, anyone interested in coming up with a PR that adds it?

Mark-H:
#13306 deals with adding the `parse` filter, so I'll take that as a "Great, let's do that!" :p

PR #13310, opened 2017-02-24
Example rule to restrict the manager to one hostname

Jako:
### What does it do?
Added an example rule to restrict the manager to one hostname.

### Why is it needed?
- Solving preview issues with the manager with/without www. prefix 
- Solving license issues with commercial package providers, checking licenses on base of the manager hostname.

### Related issue(s)/PR(s)
none known

### DIFF
--- a/ht.access
+++ b/ht.access
@@ -35,10 +35,11 @@ RewriteBase /
 
 
 
-# Redirect the manager to a specific domain
-RewriteCond %{HTTP_HOST} !^example-domain-please-change\.com$ [NC]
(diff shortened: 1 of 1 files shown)


rthrash:
Thomas, would this allow you to force previews on the actual domain? For example if your Manager was at manager.example.de but your site was www.example.de, the preview buttons/context menus would load the preview on the www url.

rthrash:
Ignore that question. Reviewed the actual code. Sorry!

PR #13311, opened 2017-02-24
Automatic flush of permissions on group update, create/update/remove policy

lawrenz1337:
### What does it do?
Added automatic flush of permission on updating a group, creating/updating/removing policy

### Why is it needed?
It is needed to not manually flush permissions every time when you update a group, create/update/remove a policy

### DIFF
--- a/core/model/modx/modcachemanager.class.php
+++ b/core/model/modx/modcachemanager.class.php
@@ -607,7 +607,7 @@ class modCacheManager extends xPDOCacheManager {
         $publishingResults= array();
         $tblResource= $this->modx->getTableName('modResource');
         $timeNow= time();
-        
+
         /* generate list of resources that are going to be published */
         $stmt = $this->modx->prepare("SELECT id, context_key, pub_date, unpub_date FROM {$tblResource} WHERE pub_date IS NOT NULL AND pub_date < {$timeNow} AND pub_date > 0");
         if ($stmt->execute()) {
@@ -760,7 +760,7 @@ class modCacheManager extends xPDOCacheManager {
     /**
      * Flush permissions for users
      *
-     * @return string|array The success response
+     * @return bool True if successful
      */
     public function flushPermissions() {
         $ctxQuery = $this->modx->newQuery('modContext');
@@ -769,9 +769,11 @@ class modCacheManager extends xPDOCacheManager {
             $contexts = $ctxQuery->stmt->fetchAll(PDO::FETCH_COLUMN);
             if ($contexts) {
                 $serialized = serialize($contexts);
-                $this->modx->exec("UPDATE {$this->modx->getTableName('modUser')} SET {$this->modx->escape('session_stale')} = {$this->modx->quote($serialized)}");
+                if ($this->modx->exec("UPDATE {$this->modx->getTableName('modUser')} SET {$this->modx->escape('session_stale')} = {$this->modx->quote($serialized)}")) {
+                    return true;
+                }
             }
         }
-        return $this->modx->error->success('');
+        return false;
     }
 }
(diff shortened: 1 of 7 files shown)


rtripault:
Thanks for your contribution!

I guess i'm getting the idea here and it can be convenient, but really not like the fact the same method is duplicated across multiple processors... don't you think the content of the `flushPermissions` methods rather belong to `modAccessPolicy` object or `modCacheManager` "service" (see `modCacheManager::refresh`) ?

lawrenz1337:
I get your idea, and to be honest I don't really like to duplicate methods across processors but `flushPermissions` affects users and I don't think it belongs to `modAccessPolicy` on the other hand I really like the idead of using cacheManager, that would be cool, i will look into it tomorrow morning and update my commit

core/model/modx/modcachemanager.class.php
```diff
@@ -756,4 +756,22 @@ public function clearCache(array $paths= array(), array $options= array()) {
...
+        $ctxQuery = $this->modx->newQuery('modContext');
+        $ctxQuery->select($this->modx->getSelectColumns('modContext', '', '', array('key')));
+        if ($ctxQuery->prepare() && $ctxQuery->stmt->execute()) {
+            $contexts = $ctxQuery->stmt->fetchAll(PDO::FETCH_COLUMN);
+            if ($contexts) {
+                $serialized = serialize($contexts);
+                $this->modx->exec("UPDATE {$this->modx->getTableName('modUser')} SET {$this->modx->escape('session_stale')} = {$this->modx->quote($serialized)}");
+            }
+        }
+        return $this->modx->error->success('');
```
rtripault: This would be "useless" since the "exec" never sets an error.

What do you think about returning `false` as default (meaning we were unable to perform the statement, and return `true` whenever `modX->exec` is executed ?

rtripault (changes requested):
Sorry to bother again @lawrenz1337, but i added a new comment.

Also, would you mind updating the `security/access/flush` processor to make use of the cache manager method too (less code duplication!).

Many thanks

core/model/modx/modcachemanager.class.php
lawrenz1337: Sure, I can do that

core/model/modx/modcachemanager.class.php
```diff
@@ -756,4 +756,22 @@ public function clearCache(array $paths= array(), array $options= array()) {
...
+     *
+     * @return string|array The success response
+     */
+    public function flushPermissions() {
+        $ctxQuery = $this->modx->newQuery('modContext');
+        $ctxQuery->select($this->modx->getSelectColumns('modContext', '', '', array('key')));
+        if ($ctxQuery->prepare() && $ctxQuery->stmt->execute()) {
+            $contexts = $ctxQuery->stmt->fetchAll(PDO::FETCH_COLUMN);
+            if ($contexts) {
+                $serialized = serialize($contexts);
```
rtripault: After some testing, turns out this should not be required, xPDO should handle the serialization on its own.

lawrenz1337: I tried removing the serialization and it didn't work. Maybe I got wrong what you said, anyway, you can check the new commit

rtripault (approved):

core/model/modx/modcachemanager.class.php
rtripault: Right, as discussed in Slack, you are absolutely right, serialization is required for the `exec` to work!

Sorry about that

rtripault:
This merged in 2.x (2.6.0) - see b10e92b6fff380483561b473fd13d06c11ed4f1d
Not sure why the PR was not automatically closed/marked as merged tho, so manually closing.

Thanks again

lawrenz1337:
Hmm, ok I can see the changes, but in app it shows that pr is still not merged and closed, but whatever, weird stuff

OptimusCrime:
If it was cherry picked and not merged through the GitHub interface the PR will not be closed automatically. It may be because of that.

Issue #3749, opened 2011-02-05
Problem when there is several active Sets in the Form Customization


Spheerys:
spheerys created Redmine issue ID 3749

I have found an anomaly when I trying to use several active Sets for a same profile, with the TV templates.
The goal, is to customize the manager for a final user. 
The final user (Bob), has to only have the necessary fields when he's editing/creating a document.
We are using TVs for the several type of documents.

_Preparation to reproduce the bug :_
- imagine there are 2 sorts of container documents : Stages (id=1) and Lessons (id=2)
- create two TVs : 
  - StageDate (if you want in a category "Stage" because this TV is only for the Stages)
  - LessonPlace (if you want in a category "Lesson" because this TV is only for the Lessons)
- create the account for Bob and make all the right access permissions and role
- go to Security > Form Customization
- create a new profile "End User" and edit it
- create a new Set in "Stage Creation" with the good parameters, for the same Template (I suppose I don't have to explain everything here)
- connect yourself with Bob and try to create a stage : everything should be OK : you can see the TV tab with the StageDate TV
- if yes, return in the Form Customization and duplicate the "Stage Creation"
- rename it "Lesson Creation" and edit it to change the good TV, the parent targeted etc...
- desactivate the "Stage Creation" Set and activate the "Lesson Creation" 
- with Bod account, try to create a new lesson and verify if the good TV is accessible (and only it)

_So normaly everything is ready to reproduce the bug :_
- return in the Form Customization and activate the both Sets
- with Bod account, try to create a Stage and verify if the good TV is accessible
- if yes, create now a new Lesson. The LessonPlace TV is not accessible while it  should !

hugopeek:
Will try to reproduce this bug during #modxbughunt.

hugopeek:
OK, I think I figured out the scenario that TS described:

- Created Editor user group
- Created user Bob in Editor group
- Create FC set "For Bob" that hides a certain TV for the Editor group
- Confirm that it works: TV visible for admin, TV hidden for Bob
- Duplicate the entire For Bob **profile** (so not just the set)
- Hide a different TV from the Editor group (and set the first one to "visible" again)
- Confirm that it works
- Enable both profiles

This results in both TVs **not** being displayed. I tested it with Pagetitle, with the same result.

Conclusion: the "hidden" state is dominant for the visibility toggle. If an element is present in multiple FC profiles, and visibility is unchecked in one of them, the element will remain hidden.

[...]

PR #13314, opened 2017-03-03
[FC] Improve description of how visibility toggle behaves in overlapping profiles

hugopeek:
### What does it do?
This addresses the issue described in #3749.

### Why is it needed?
Although I think issue 3749 is not a bug, it is a bit unclear how the visibility setting behaves. It would be good to know that any "Hidden" state for an element is dominant when the same element is listed as "Visible" in an overlapping FC profile.

### Related issue(s)/PR(s)
#3749


Mark-H:
Merged for the #modxbughunt - thanks!

Issue #12701, opened 2015-10-06
Missing Lexicon Error: MediaSource Access Policy


matdave:
![image](https://cloud.githubusercontent.com/assets/2213127/10320083/349ef49e-6c37-11e5-88a2-205b42c4cee0.png)

Just noticed when a Policy isn't selected there is an empty error on the Media Source access policy creation window.

syts88:
I'm going to try and fix this today.

syts88:
Entered a pull request with the fix. #modxbughunt

Mark-H:
Fixed in 2.5.6

PR #13316, opened 2017-03-03
Fixed issue #12701 #modxbughunt

syts88:
### What does it do?
Removed the empty error pop up. The error pop up was empty because of a missing lexicon. 
I made the choice to remove this pop up because in similar actions which you can do in Modx there is also no error pop up. Besides the error is clear and visible beneath the policy field.

### Why is it needed?
Empty error pop ups gives it a messy look.

### Related issue(s)/PR(s)
#12701


syts88:
#modxbughunt

Mark-H:
Thanks @syts88 - merged for 2.5.6! #modxbughunt

PR #13317, opened 2017-03-03
Fixed issue #13302 #modxbughunt

### LINKED ISSUE #13302: Duplicate TV has caption "undefined"
### Summary
When duplicating a TV (right-click inside the tree) the new caption is always "Duplicate of undefined" (red box), even if the source has a caption (green box):
modx_issue_duplicate-tv
(When using the "Duplicate"-button while editing a TV, the new caption field is empty and the is no "undefined" – but still, there is no duplicate of the caption.)

### Step to reproduce
Right-click on a TV an choose "Duplicate TV".

### Observed behavior
see: summary

### Expected behavior
The new caption should show "Duplicate of <caption_old>" with <caption_old> being the caption of the source TV or it should be empty, when there is nothing to duplicate (caption of source is empty)

### Environment
On any system, as far as I can say, running 2.5.5. I had this issues also on lower versions but can't recall if it was 2.5.4, 2.5.2 or even lower.

johanmolen:
### What does it do?
The caption field now has a placeholder once you click the duplicate button (at the TV screen itself not the tree).

### Why is it needed?
Because the tree duplicate option did it but the TV screen duplicate button did not.

### Related issue(s)/PR(s)
Fixes #13302 

template_variable__testname___modx_revolution

changed to:

template_variable__testname___modx_revolution

### DIFF
--- a/manager/assets/modext/sections/element/tv/update.js
+++ b/manager/assets/modext/sections/element/tv/update.js
@@ -49,6 +49,7 @@ Ext.extend(MODx.page.UpdateTV,MODx.Component, {
             id: this.record.id
             ,type: 'tv'
             ,name: _('duplicate_of',{name: this.record.name})
+            ,caption: _('duplicate_of',{name: this.record.caption})
         };
         var w = MODx.load({
             xtype: 'modx-window-element-duplicate'


Mark-H:
Merged for 2.5.6, thanks!

PR #13319, opened 2017-03-03
Fix #10299: let developers know that config files will be overwritten #modxbughunt

### LINKED ISSUE #10299: Add comments to default config file
everettg_99 created Redmine issue ID 10299

Because the core/config/config.inc.php and the config.core.php files get rewritten by the MODX install/upgrade process, it's important to alert dev users to that fact.  In many (most?) other systems, the config files are not WRITTEN by the app, they are only READ, so the config files become an important spot for developers to add customizations (e.g. defining constants, adding global settings, etc), but in MODX, only the $config_options array is safe from the update process.  As a courtesy to any developers who are used to the way other systems handle this, it is important to alert them to this behavior.  Just a short blurb of text would suffice:

<pre>
/* This file is managed by the installation process.  Any modifications to it may get overwritten.*/
</pre>

Or 

<pre>
/* Add customizations to the $config_options array.  All other modifications to this file may get overwritten by the install or upgrade process. */
</pre>

pbowyer:
### What does it do?
Adds a comment to the top of autogenerated config files

### Why is it needed?
Most frameworks do not overwrite config files when you upgrade. I have customised them in the past (substituting `__DIR__` for hardcoded paths, so they work on different servers) only to lose the changes when I upgrade. This issue doesn't fix that, but it reminds everyone not to rely on the files.

### Related issues
#10299 

#modxbughunt

### DIFF
--- a/_build/resolvers/resolve.core.php
+++ b/_build/resolvers/resolve.core.php
@@ -10,7 +10,7 @@ switch (MODX_SETUP_KEY) {
                 $configContent= "<?php\n"
                     . "/*\n"
                     . " * This file is managed by the installation process.  Any modifications to it may get overwritten.\n"
-                    . " * Add customizations to the $config_options array in `core/config/config.inc.php`.\n"
+                    . " * Add customizations to the \$config_options array in `core/config/config.inc.php`.\n"
                     . " *\n"
                     . " */\n"
                     . "define('MODX_CORE_PATH', '" . MODX_CORE_PATH . "');\n"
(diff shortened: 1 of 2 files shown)


Mark-H:
After running the setup, the file looks like this:

````
<?php
/*
 * This file is managed by the installation process.  Any modifications to it may get overwritten.
 * Add customizations to the  array in `core/config/config.inc.php`.
 *
 */
define('MODX_CORE_PATH', '/Users/mhamstra/Sites/revolution.local/core/');
define('MODX_CONFIG_KEY', 'config');
?>
````

I.e. `$config_options` is getting replaced with nothing. Maybe just getting rid of the `$` is sufficient to stop that from happening.

pbowyer:
Thanks @Mark-H, fixed!

Mark-H:
Thanks @pbowyer - merged for inclusion in 2.5.6.

Issue #13256, opened 2017-01-18
Inexact error during file or directory rename


sartatpdev999:
### Summary
When i trying to rename directory (by right-clicking in File Manager) to existing directory name i get inexact error. For example, i have two directories in `/path/to/dir`:

- `test`

- `_test`

And when i trying to rename `_test` to `test` i get the following error: 

> An unknown error occurred while trying to rename the directory.

I think it would be better something like this:

> A directory already exists with that name in that location.

This feature also applies to files.

### Step to reproduce
Checking for file or directory existence while renaming. Something like:

```
  if (file_exists($newPath)) {
    $this->addError('name',$this->xpdo->lexicon('file_folder_err_ae'));
    return false;
  }
```

For file existence:

```
$this->addError('name',$this->xpdo->lexicon('file_err_ae'));
```
In:

- core/model/modx/sources/modfilemediasource.class.php

- and all other related files

### Expected behavior
When trying to rename directory to existing name it should throw:

> A directory already exists with that name in that location

When trying to rename file to existing name it should throw:

> File %s already exists

Flowan:
I'm going to try and fix this today.

Mark-H:
Fixed in 2.5.6

PR #13320, opened 2017-03-03
Fixed issue #13256 #modxbughunt

flowan:
### What does it do?
Add check before renaming a file or directory to see if the new path not already exists and give the correct response to the client.

### Why is it needed?
To show the correct response for renaming a file or directory.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13256


Mark-H:
Thanks @Flowan - fixed and included in 2.5.6.

Issue #13277, opened 2017-02-06
Make duplicating Context Resources optional when duplicating Context


jaygilmore:
### Summary
It should be optional to duplicate the Resources under a Context when duplicating the Context either from the tree or as suggested in #13276. Add checkbox to the dialog to confirm choice to duplicate Resources.

### Step to reproduce
Trigger duplicating a Context from the Resource Tree. Dialog pops up with options. It should be another unchecked checkbox on the dialog box to confirm that they wish to duplicate the Resources as well.

### Observed behavior
Currently, it is not possible to optionally duplicate Resources when duplicating a Context. The Resources will be Duplicated.

### Expected behavior
![duplicate_child_resources](https://cloud.githubusercontent.com/assets/378199/22664363/bf14de38-ec86-11e6-91d6-fec453e9ad1a.png)
Allow the user to choose to duplicate or not duplicate the children Resources of a Context. Add the checkbox input with the text: "Duplicate child Resources". It may also make sense to disable the `Preserve duplicated resources alias` and `Preserve duplicated resources menu index` checkboxes until the `Duplicate child Resources` has been checked. 

### Environment
Current MODX Revolution Managers +

joeke:
I'll look into this. #modxbughunt

Mark-H:
Included in 2.6

PR #13321, opened 2017-03-03
Fixed issue #13277 - When duplicating context choose if resources should be duplicated #modxbughunt

joeke:
### What does it do?
Fix issue #13277 
When duplicating context allow user to choose to duplicate resources or not. When un-checking, the options 'Preserve duplicated resources alias' and 'Preserve duplicated resources menu index' are un-checked and disabled.

### Related issue(s)/PR(s)
Issue #13277


rtripault:
Not deeply reviewed yet, but just having the feeling this would rather belong to 2.6 (feature) rather than a 2.2 (bugfix)

joeke:
Yeah you're right. It's more of a new feature, so not really a bugfix.

Mark-H:
Thanks @joeke, merged into 2.6.

Issue #13298, opened 2017-02-17
Duplicating resource folder with hidden children does not ask to duplicate children


sonicpunk:
### Summary
Another issue - when you have children that are hidden from the resource tree, like when using the Collections extra for example. Clicking duplicate on the parent resource will not duplicate the children if they are not shown in the resource tree. Also the checkbox option for duplicating children does not show up in the modal.

It seems that duplicating children only works with resources that are actually shown in the resource tree.

Attribute hasChildren=false, even though it technically does have children, just that they are not shown in the tree.

### Step to reproduce
Make a parent resource with children.  Change the children show_in_tree value to 0 in the database. These children resources are no longer shown in the resource tree.

### Observed behavior
Right click on the parent resource.  The checkbox for duplicating the children does not show.

### Expected behavior
Technically the children do exist, therefore I would expect the ability to duplicate them as well.

### Environment
MODX version 2.5.5

wuuti:
Can confirm that. In the ResourceGetNodesProcessor the children count (L376) is only populated with resources not hidden tree (L170). But this value is then used in modx.tree.resource.js (L124) to configure the duplicate modal (window.js), where it is then used to show up the opton "Duplicate children" or not. 
IMHO the option should be visible _always_ if there are any children, regardless of whether they are visible or not. As the processor returned attribute "hasChildren" might be used elsewhere, it would be a good idea to add something like "hasAnyChildren" here, which is then evaluated by the duplicate modal. 
I can also think of a more verbose modal where a checkbox labelled with "Duplicate children (15 total, 3 hidden in tree)", which would warn you then if you dare to duplicate a collection with thousands of children.

sdrenth:
I'm going to try and fix this today.

sonicpunk:
#modbughunt Thanks @sdrenth !  I just added the hashtag so that we can filter it out.  Get those bugs!!!

sdrenth:
I've added a pullrequest which fixes this issue and also adds a child counter next to the duplicate children option:

https://github.com/modxcms/revolution/pull/13322

sonicpunk:
I was thinking that we shouldn't overwrite the current behaviour but maybe instead create a new value that saves the count of all children regardless of the show_in_tree value.  Then we could have a new system setting that would allow the admin to choose which behaviour they would like to have.

Mark-H:
Fixed in 2.5.6

[...]

PR #13322, opened 2017-03-03
Fixed issue #13298 #modxbughunt

sdrenth:
### What does it do?
If you duplicate a resource with resources which are hidden from the tree, the duplicate children option is now also shown. Also I've added a children count next to the duplicate children checkbox.

### Why is it needed?
It wasn't possible to duplicate children of a resource when they where set to hidden in the tree (show_in_tree) because the Duplicate children options was not shown. Now the duplicate children option is also being shown when the children are not being shown in the tree.

### Related issue(s)/PR(s)
Issue #13298


Mark-H:
Nice work 👍 Merged for 2.5.6 and the #modxbughunt of course.

Issue #6142, opened 2011-11-26
Plugin Manager: List active plugins and their actions


modxbot:
everettg_99 created Redmine issue ID 6142

This would make a great 3PC, but what I'd love to see is an easy way to view _all_ active plugins in one place, and also to have way to all actions that are being used by active plugins.  

E.g. view active plugins might show a list

_Active Plugins_
- TinyMCE: event1, event2
- Quip: event3
- My Custom Plugin: event1

And a secondary page (e.g. click "event1") and then see a display showing all plugins tied to that event:

_event1_
- TinyMCE
- My Custom Plugin

This would be a cleaner view on what's going on with events.

Alroniks:
Picked up this #modxbughunt

rtripault:
While still not perfect, there's been improvements on that area, so i'll close that one.

Feel free to comment/reopen if needed.
Thanks

PR #13324, opened 2017-03-03
#6142 Added abbility to see which plugins attached to the event #modxbughunt

alroniks:
### What does it do?
It allows to see which plugins attached to system event on the events page.

### Why is it needed?
When in the system a lot of plugins sometimes it can be useful so see which of them attached to events from events page. Or when we try to see which plugins work with specific event.

system settings - modx revolution

### Related issue(s)/PR(s)
Partially fix for https://github.com/modxcms/revolution/issues/6142


core/model/modx/processors/system/event/getlist.class.php
```diff
@@ -11,60 +11,94 @@
...
         $list = array();
         /** @var modEvent $event */
         foreach ($data['results'] as $event) {
             $eventArray = $event->toArray();
 
+            if ($eventArray['plugins']) {
+                $c = $this->modx->newQuery('modPlugin');
+                $c->leftJoin('modPluginEvent', 'modPluginEvent', 'modPluginEvent.pluginid = modPlugin.id');
+                $c->sortby('modPluginEvent.priority', 'ASC');
+                $c->where(['modPluginEvent.event' => $eventArray['name']]);
```
rtripault: While i would love to use the short array syntax, this would break compatibility to PHP 5.3.x

rtripault (changes requested):
Added a small note about short array syntax (which is PHP 5.4+).

Also, i was wondering if we should display only active plugin (`where(array('disabled' => false))` or style disabled plugins differently (rendered in the grid).

Otherwise :+1:

core/model/modx/processors/system/event/getlist.class.php
Alroniks: Agree, It's a hard to follow when it has become a habit. :)

Alroniks:
@rtripault I thought about to add marks for plugins, like `Ace (disabled, locked), pdoTools (locked)` and to avoid to change styles. My assumptions, that here we will have only readable info and disabled mostly applicable to form elements.

Mark-H:
This looks useful, but can't be merged until the short array syntax is fixed ;)

Alroniks:
@Mark-H @rtripault Fixed short arrays and add statuses for plugins (see screnshot below)
system settings - modx revolution 1

rtripault (approved):

core/model/modx/processors/system/event/getlist.class.php
```diff
@@ -11,60 +11,103 @@
...
+                        : $plugin['name'];
+
+                    $plugins[] = array('name' => $pluginName);
+                }
+                $eventArray['plugins'] = $plugins;
+            } else {
+                $eventArray['plugins'] = '';
+            }
+
+            $eventArray['plugins'] = json_encode($eventArray['plugins']);
```
rtripault: What is the benefit of json encoding here and having to parse the json client side later ?

rtripault:
All working as expected.
Just added a question about json encoding a value that would be json encoded later anyway

core/model/modx/processors/system/event/getlist.class.php
Alroniks: You are right, it's obsolete. I've fixed and removed this encoding/decoding.

rtripault:
Many thanks for your work and patience ;)

gpsietzema:

PR #13325, opened 2017-03-03
Fixes #4039 Add double encoding option to htmlent output filter #modxbughunt

### LINKED ISSUE #4039: htmlent output filter double encodes entities
joncmodx created Redmine issue ID 4039

I made a mod to the "htmlent" output filter code so that it defaults to having the double_encode param passed to the actual PHP htmlentities function set to false, but also allows you to override this with a "true" param. The reason for this is that a lot of times I was running into code that had some characters in their html entity form while other characters were not. Setting the filter to default to not double encode will resolve this.

I realize this may not be appropriate to be added to the default install because apparently PHP < 5.2.3 doesn't support the double encode param in the PHP htmlentities function.
(http://modxcms.com/forums/index.php/topic,47796.0.html)

Usage
[[+your-tag:htmlent]] - Will not double encode htmlentities (ex. "This &amp;  That" = &quot;This &amp; That &quot;)

[[+your-tag:htmlent=`true`]] - Will double encode entities (ex. "This &amp; That" = &quot;This &amp;amp; That &quot;)

Code:
In line 205 of core/model/modx/filters/modoutputfilter.class.php:

<pre>
                    case 'htmlent':
                    case 'htmlentities':
                        /* See PHP's htmlentities - http://www.php.net/manual/en/function.htmlentities.php */
                        if ($m_val == 'true') {
                            $double_encode = true;
                        } else {
                            $double_encode = false;
                        }
                        $output = htmlentities($output,ENT_QUOTES,$encoding,$double_encode);
                        break;
</pre>

Related post:
http://modxcms.com/forums/index.php/topic,62162.0.html

pbowyer:
### What does it do?
Add double encoding option to htmlent output filter

### Why is it needed?
It's useful if you have a TV that contains already-encoded data e.g. `He &amp; She` and you want to output the &amp; to be visible on the page.

Using `[[*myvar:htmlent=`true`]]` will now change the above to `He &amp;amp; She`.

### Related issue(s)/PR(s)
#4039

#modxbughunt

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -225,7 +225,7 @@ class modOutputFilter {
                         case 'htmlent':
                         case 'htmlentities':
                             /* See PHP's htmlentities - http://www.php.net/manual/en/function.htmlentities.php */
-                            $output = htmlentities($output,ENT_QUOTES,$encoding);
+                            $output = htmlentities($output,ENT_QUOTES,$encoding,($m_val == 'true'));
                             break;
                         case 'htmlspecialchars':
                         case 'htmlspecial':


gpsietzema:

Issue #10267, opened 2013-10-16
Content Type adding extra dot


sottwell:
sottwell created Redmine issue ID 10267

When using a Static Resource as an attachment and a Content Type, an extra . is added to the downloaded file's path:

assets/files/testing..pdf
assets/style/css/testing..css
assets/js/testing..js

Alroniks:
Checked, confirmed for PDS (other content types works fine) #modxbughunt
Will try to fix

Alroniks:
Fixed

OptimusCrime:
PR is merged. Closing this then.

@modxbot close

PR #13327, opened 2017-03-03
#10267 Fixed issue with double dots in the file name #modxbughunt

alroniks:
### What does it do?
Removes redundant dots from string which creates filenames using name of resouce and extension from content types.

### Why is it needed?
When you create static resouce with binary content type and mode Attachment, MODX will try to send header for downloading this file. But in case, when FURL turned off, URI param is empty and will be used alias or title, that use formula `$name .= ".{$ext}";`. But currently all content types contains extensions with dots as a result `name` + `.{$ext}` where ext is `.jar` will produce `name..jar`, but should be `name.jar`.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/10267


Mark-H:
Nice one @Alroniks, merged into 2.5.6

Issue #13094, opened 2016-08-22
[Bug?] Very strange blank manager cause, moving top menu items with handlers


exside:
### Summary

First question is: Am I the only one getting this when doing as explained later on?

The manager goes blank (Internal Server Errors behind, so clearly visible in the inspector) entirely after moving around two menu items that have JavaScript handlers attached, e.g. the one for "flush own permissions" and "reset all sessions (log out all users)", everything is fine a s long as they are in the normal/original parent menu item, but when you move them, let's say into the "User" top nav container (where "Logout" and "Profile" etc. are), an internal server error (500) will happen and the logs say as follows:

```
PHP Fatal error:  Uncaught  --> Smarty Compiler: Syntax error in template "8d6bd625d30e7f6357af45b17d7d9fe19d9a3258"  on line 17 "'success': {fn:function() { location.href = './'; },scope:this}"  - Unexpected ":", expected one of: "}"
```

to see where it actually came from you have to edit this error output of smarty in _core/model/smarty/sysplugins/smarty_internal_templatecompilerbase.php_ at around line 1017, I just added:

```
$error_text .= ', template filepath: ' . "\n" . print_r($this->template->source, 1);
```

to get a hint of what template we are talking about here...because the "path" provided is a hash in this case (for whatever reason)
### Step to reproduce

See above, just move one of the menu items to the user menu and be aware that you have to edit _manager/controllers/default/header.php_ and comment out the section _after_ you have done the menu moving:

``` php
            if (!empty($menu['handler'])) {
                $attributes .= ' onclick="'.str_replace('"','\'',$menu['handler']).'"';
            }
```

because otherwise, you don't have any access to the manager and cannot undo the menu item move...yeah, you could go into the db, but that seemed easier...

The content of both problematic handlers looks fine to me (but could be that some part of the formatting messes withs smarty? but if so, why not when they are in the original top menu?):

``` javascript
MODx.msg.confirm({
    title: _('flush_sessions')
    ,text: _('flush_sessions_confirm')
    ,url: MODx.config.connector_url
    ,params: {
        action: 'security/flush'
    }
    ,listeners: {
        'success': {fn:function() { location.href = './'; },scope:this}
    }
});


MODx.msg.confirm({
    title: _('flush_access')
    ,text: _('flush_access_confirm')
    ,url: MODx.config.connector_url
    ,params: {
        action: 'security/access/flush'
    }
    ,listeners: {
        'success': {fn:function() { location.href = './'; },scope:this}
[...]

PR #13328, opened 2017-03-03
Added functionality to allow javascript in the user-nav #modxbughunt

Oetzie:
### What does it do?
Added functionality to allow javascript in the user-nav

### Why is it needed?
Wasn't possible to use JS in the usernav.

### Related issue(s)/PR(s)
#13094

#modxbughunt

### DIFF
--- a/manager/controllers/default/header.php
+++ b/manager/controllers/default/header.php
@@ -312,7 +312,7 @@ class TopMenu
                 $attributes = ' href="?a='.$menu['action'].$menu['params'].'"';
             }
             if (!empty($menu['handler'])) {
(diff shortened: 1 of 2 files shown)


Mark-H:
Do you have some suggested steps to test this? E.g. some code that couldn't run before that can be run now?

Oetzie:
Before this fix you could not place the actions 'clear cache' or something like that under the usernav menu. Because the eval (smarty) and { triggered a php error.

gpsietzema:
#13094

Mark-H:
Thanks @Oetzie, merged into 2.x for 2.6.0.

Issue #12828, opened 2015-12-14
uberbar search result resource icons ambiguous


smaddock:
### Summary

This appears to be a repeat of #11875. The uberbar resource search results should use the same icons as the resource tree.
### Step to reproduce

Search for resources in the uberbar.
### Observed behavior

All resources use identical "file" icons
### Expected behavior
- The icon shown for Resources should match what is used in the resource tree.
- If the result is a Weblink / Symlink / Static Resource, it should uses the specific icon for that resource type.
- If the result is a CRC or has a template icons, those should be displayed instead.
### Environment

MODX 2.4.2

smaddock:
PR #12749 offers a workaround using a custom processor, but the above should still be in core.

neoneddy:
I just came here to say, weblinks should have a different icon.  Or they should match the tree styles.  Which may have been said here or in another.  That is all.

alroniks:
Note for me and other contributors: it seems one AC missed - about template icons attached to resources.

alroniks:
Checked in 2.7.0-dev, it respects template icons too.

PR #13329, opened 2017-03-03
#12828 Fixed not showing correct icons and adding max search results option #modxbughunt

andresterc:
### What does it do?
Fixed the icon issues and added some extra features like showing the contextkey and make max searchresults adjustable with a contextsetting.

### Why is it needed?
There was no difference between the resource types. The icon was all the same.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/12828


Mark-H:
Replaced by pull #13332

Issue #13049, opened 2016-06-08
OnDocFormRender set template bug


Tauntik:
In plugin with OnDocFormRender event, i can't set document template.

$resource->set('template', 3);
or
$modx->controller->setProperty('template', 3);
make no effect

MODx version: 2.5.

On version 2.2 
$modx->controller->setProperty('template', 3);
this wariant worked fine.

Realetive:
It's ok for me. See working example here: https://gist.github.com/Realetive/b92384dc615cd481e5f436d586a2d82f

Flowan:
I'm going to try and fix this today. #modxbughunt

Alroniks:
Could be this issue to be related to your fix? https://github.com/modxcms/revolution/issues/10140 @Flowan

OptimusCrime:
This worked before the patch?

Mark-H:
Merged #13330 which fixes this.

PR #13330, opened 2017-03-03
Allow overriding the resource template on OnDocFormRender #13049 #modxbughunt

flowan:
### What does it do?
Override document template in OnDocFormRender event.

### Why is it needed?
Add template to scriptproperties if set in OnDocFormRender event.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13049

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -82,6 +82,11 @@ class ResourceCreateManagerController extends ResourceManagerController {
         $placeholders['parentname'] = $this->setParent();
         $this->fireOnRenderEvent();
 
+        /* set template */
+        if (!is_null($this->resource->get('template')) && $this->resource->get('template') !== 0) {
+            $this->scriptProperties['template'] = $this->resource->get('template');
+        }
+
         /* check permissions */
         $this->setPermissions();
 


Mark-H:
Hi @Flowan! It looks like `$modx->controller->setProperty('template', 3);` works before this patch... is this needed?

Flowan:
Hi @Mark-H Correct, that works. But with the change i made you can use `$resource->set('template', 3);` to set the template.

I think this should be added. Because you can set/overwrite all values of a resource with `$resource->set()` ,except the template. Isn't that weird behaviour?

Mark-H:
That makes sense. It'd be interesting to know why it works for other fields and not the template, I suspect due to the way the template can be overridden in different ways?

Flowan:
@Mark-H The template can be overridden in different ways, that's why it didn't work before.

Mark-H:
Not sure why the PR didn't get autoclosed, but this is merged for 2.7 in https://github.com/modxcms/revolution/commit/b5fd81aa5399efec7c1d5185a4c79bf1103fc746

Mark-H:
Thanks @flowan for the contribution, sorry it took so long to merge ;)

Issue #12028, opened 2014-10-08
Remove locks on open resource does not change manager UI


wuuti:
The "remove locks" does not change the state of the currently open page if it is locked. If you try to edit a page which is locked, unfortunately you can do so (see #11929), but the save button is disabled and there is the only information for you that this resource is locked ("locked by...."). Fine, you may remove the locks from the menu, but that does not change the state of your open manager page - it will still show "locked by..." and you cannot save. So your only chance is to reload the whole page (and loose all your changes). 
The user interface _should_ change if you remove lock on the current, because that is what you want to do in that moment.: to remove the lock and keep editing the resource....

sebastian-marinescu:
:+1: 

I just had that issue yesterday, but I was lucky enough I hadn't changed too much. 
My end-users might not be so lucky, so this has somehow priority in my opinion, because it is unexpected behaviour and also it is hindering your creative freedom...

Oetzie:
I'm going to try and fix this today. #modxbughunt

Mark-H:
Fixed in 2.6

PR #13331, opened 2017-03-03
Re-enable resource save buttons after clearing the locks #12028 #modxbughunt

Oetzie:
### What does it do?
Added the UI of the resource update panel, when you remove resource lockes the save button will show up again (12028) #modxbughunting

### Why is it needed?
You could not save a resource when you removed the resources locks because the save button did not appear.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -227,7 +227,7 @@ $childrenOfClearCache[0]->fromArray(array (
 $children[1]->addMany($childrenOfClearCache, 'Children');
 
 /* Remove Locks */
-$children[2]= $xpdo->newObject('modMenu');
+/*$children[2]= $xpdo->newObject('modMenu');
 $children[2]->fromArray(array (
   'menuindex' => 2,
   'text' => 'remove_locks',
@@ -252,6 +252,17 @@ MODx.msg.confirm({
          },scope:this}
     }
 });',
+), '', true, true);*/
+
+$children[2]= $xpdo->newObject('modMenu');
+$children[2]->fromArray(array (
+  'menuindex' => 2,
+  'text' => 'remove_locks',
+  'description' => 'remove_locks_desc',
+  'parent' => 'manage',
+  'permissions' => 'remove_locks',
+  'action' => '',
+  'handler' => 'MODx.removeLocks();return false;',
 ), '', true, true);
 
 /* Flush Permissions */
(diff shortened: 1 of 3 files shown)


_build/data/transport.core.menus.php
```diff
@@ -227,7 +227,7 @@
 $children[1]->addMany($childrenOfClearCache, 'Children');
 
 /* Remove Locks */
-$children[2]= $xpdo->newObject('modMenu');
+/*$children[2]= $xpdo->newObject('modMenu');
```
Mark-H: You can just remove code that's no longer needed instead of commenting it out; it'll still be in the git history if needed to revert at some point.

manager/assets/modext/sections/resource/update.js
```diff
@@ -103,27 +103,33 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
 
     ,getButtons: function(cfg) {
         var btns = [];
-        if (cfg.canSave == 1) {
+        
+        //if (cfg.locked) {
```
Mark-H: Again same as above, just remove what's no longer needed please.

manager/assets/modext/sections/resource/update.js
```diff
@@ -103,27 +103,33 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
...
+            btns.push({
+                text: cfg.lockedText || _('locked')
+                ,id: 'modx-abtn-locked'
+                ,handler: Ext.emptyFn
+                ,hidden: (cfg.canSave == 1)
```
Mark-H: Before the check was on cfg.locked - I understand the logic of rendering the buttons despite it being locked or not, but are you sure this check is correct?

Mark-H (dismissed):
Hi @Oetzie - I added a few comments to the code. Could you please review and update the PR? Thanks!

gpsietzema:
#12028

manager/assets/modext/sections/resource/update.js
rtripault: Agreed, seems wrong from my seat too.

@Oetzie what do you reckon ?

opengeek:
@Mark-H Does this need to be re-reviewed?

Mark-H:
Thanks @Oetzie, merged into 2.6.

Mark-H:
It's not closing automatically because the base branch was set to 2.5.x (and I can't change that anymore), but it is definitely merged.

opengeek:
You should be able to change the base branch.

opengeek:
But yes, do that before merging. ;)

Mark-H:
Yeah, can't do it after merging. GitHub says there's no difference between the base and head, and just leaves you hanging like that. I've actually run into that a few times in the past two weeks, and still haven't learned to check the base first 😁

PR #13332, opened 2017-03-03
#12828 Fixed not showing correct icons and adding max search results option #modxbughunt

andresterc:
### What does it do?
Fixed the icon issues and added some extra features like showing the contextkey and make max searchresults adjustable with a contextsetting.

### Why is it needed?
There was no difference between the resource types. The icon was all the same.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/12828

### DIFF
--- a/core/model/modx/processors/search/search.class.php
+++ b/core/model/modx/processors/search/search.class.php
@@ -98,7 +98,7 @@ class modSearchProcessor extends modProcessor
 //        $c->where(array(
 //            'action:LIKE' => '%' . $this->query . '%',
 //        ));
-//        $c->limit($this->maxResults);
+//        $c->limit($this->modx->getOption('maxsearchresults',null,$maxResults));
 //
 //        $collection = $this->modx->getCollection($class, $c);
 //        /** @var modMenu $record */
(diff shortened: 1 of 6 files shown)


core/model/modx/processors/search/search.class.php
```diff
@@ -172,13 +172,14 @@ public function searchResources()
         ));
         $c->sortby('createdon', 'DESC');
 
-        $c->limit($this->maxResults);
+        $c->limit($this->modx->getOption('maxsearchresults',null,$maxResults));
```
rtripault: Instead of overriding for each "case", you could change the attribute (`$this->maxResults`) in the process method, or the like

andresterc:
@rtripault fixed the systemsetting key name

Mark-H:
Any thoughts on Romain's other suggestion? 

> Instead of overriding for each "case", you could change the attribute ($this->maxResults) in the process method, or the like

andresterc:
@Mark-H All the suggestions of Romain are fixed!

rtripault:
@andresterc as far as i can read from the diff, you are still limiting the number of result in each "case" rather than globally

Mark-H:
This is still looking up the system setting in each limit call. Once that gets resolved I see no problem merging this. :)

sdrenth:
@andresterc Can you put this on top of the process method in /modx/processors/search/search.class.php:
```
$maxSearchResults = (int) $this->modx->getOption('max_search_results');
$this->maxResults = ($maxSearchResults > 0) ? $maxSearchResults : 5;
```

And then replace for every case:
```
$c->limit($this->modx->getOption('max_search_results',null,$this->maxResults));

with:
$c->limit($this->maxResults);
```
So we can get this PR merged ;)

OptimusCrime:
Or:

```
$this->maxResults = (int) $this->modx->getOption('max_search_results');
if ($this->maxResults === 0) {
    $this->maxResults = 5;
}
```

Depends on what kind of code you find most readable. I prefer something like this.

OptimusCrime:
I have also submitted a PR for this at #13591, although I am not 100% sure how it works. This PR seem to contain changes for a couple of different issues?

opengeek:
Any chance this PR could be reduced in scope to just fixing the icon issue?

Mark-H:
With the icons already being fixed since #13705, it would be best to create a new PR for the other things this added. Sorry it took so long @andresterc! Hopefully you're still willing to help improve MODX ;)

Issue #13196, opened 2016-11-25
Save clears resource cache from all contexts


dannevang:
### Summary
When a resource is saved and the cache is refreshed resource cache is cleared from all contexts instead of only the context the resource belongs to.

### Step to reproduce
1. Create two contexts with a resource in each. 
2. View both resources on front-end and check the core/cache folder to make sure both resources has been cached. 
3. Open one resource in the manager and save. 
4. Recheck the core/cache folder and observe that all resources from all contexts has been cleared.

### Observed behavior
All contexts need new resource cache.

### Expected behavior
Only the resources from the context the resource belongs to should have been cleared.

### Environment
Tested on MODX 2.5.2 and 2.5.1 on MODX Cloud.

dannevang:
And also same behaviour with MODX 2.4.4

neoneddy:
https://modx.com/extras/package/uncacher    We use this on multi-context sites.  It helps quite a bit.

Mark-H:
I have a multi context site that fetches content (e.g. menus, blog lists) across contexts - this would prevent a standard resource save from that getting updated.

dannevang:
@Mark-H That it would. But as I have understood from opengeek the expected behaviour is what I described above.

christianseel:
I'd agree with Mark and say that the observed behavior is the expected behavior. This is how the MODX resource cache works. Clearing the cache for all contexts makes sense for the same reason why it makes sense to clear the cache of all resources within the same context.

For multilingual websites this totally makes sense and is needed. For multicontext projects, where every context is a totally different website, this of course would not be required. Therefor it might be an idea to introduce a system setting to change the behavior?

OptimusCrime:
More in favor of a system setting that can toggle the proposed behavior on and off.

rtripault:
FWIW, it seems like clearing only the relevant context(s) was the intention (see [modResourceUpdateProcessor::clearCache]( https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/resource/update.class.php#L739)).
But it seems the cache manager is not handling it (see [modCacheManager::refresh](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modcachemanager.class.php#L523) not handling the "resource" partition)

sdrenth:
I'm going to try and fix this today.

OptimusCrime:
PR submitted for this.

alroniks:
Solved.

PR #13333, opened 2017-03-03
Bug 13196

sdrenth:
### What does it do?
Added a system setting to enable clearing the resource cache partial in the cache refresh method. By default clearing the resource cache will stay unchanged. 

### Why is it needed?
By default when updating a resource the resource cache for all contexts will be cleared. This option enables the option to only delete the resource cache for the active context.

### Related issue(s)/PR(s)
Related issue: 
https://github.com/modxcms/revolution/issues/13196

#modxbughunt


sdrenth:
Closed pull request, I'm going to add this as a feature to the 2.x branch as discussed with @Mark-H

Issue #10134, opened 2013-08-27
"Prefix" Output Filter/Modifier


modxbot:
everettg_99 created Redmine issue ID 10134

It would be handy to have a built-in output filter that would prepend data onto the input.  This would be the companion to the existing "cat" output filter.

<pre>
return $options.$input;
</pre>

It really saves on space if you can have these built-in rather than having to create multiple custom Snippets to support them.

OptimusCrime:
Should be a minor addition

pbowyer:
Present in branch 2.X

Close @modxbot 
#modxbughunt

OptimusCrime:
@modxbot close

:)

PR #13335, opened 2017-03-03
Feature 10134 Add "Prefix" Output Filter/Modifier #modxbughunt

pbowyer:
### What does it do?
Adds a counterpart to the `cat` output filter. While `cat` appends the value, `prefix` prepends it.

### Why is it needed?
I wrote it because it's so simple, and the ticket has been open since 2013.

I do not think that adding to a minor release is a good idea. If someone has already defined a snipped named `filter:prefix` this addition will break their code (as it's used instead of their custom filter).

But hey, for #modxbughunt we're not discussing first what should be added, even if that would be sensible :)

### Related issue(s)/PR(s)

#10134

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -198,6 +198,10 @@ class modOutputFilter {
                             if (!empty($m_val))
                                 $output = $output . $m_val;
                             break;
+                        case 'prefix': /* prepends the options value (if not empty) to the input value */
+                            if (!empty($m_val))
+                                $output = $m_val . $output;
+                            break;
                         case 'lcase':
                         case 'lowercase':
                         case 'strtolower':


rtripault:
@pbowyer i believe you might want to target 2.5.x branch instead

pbowyer:
Thanks @rtripault, fixed!

Mark-H:
Wouldn't this be a new feature better suited for 2.x? :P

rtripault:
ahem... yeah (it's now obvious when reading the diff ;p)

/me hides

pbowyer:
Annnnd productivity comes to a grinding halt while I try to rebase my branch onto 2.x :p

pbowyer:
Already present in 2.X 😢

wuuti:
We should make sure that https://docs.modx.com/revolution/2.x/making-sites-with-modx/customizing-content/input-and-output-filters-(output-modifiers) is updated then with the new output filter.

sottwell:
Where in 2.X? It's not in the modoutputfilter.class.php file https://github.com/modxcms/revolution/blob/2.x/core/model/modx/filters/modoutputfilter.class.php

Jako:
Use `prepend` or `before`: https://github.com/modxcms/revolution/blob/2.x/core/model/modx/filters/modoutputfilter.class.php#L201-L210

But it seems that both are not included in the docs.

sottwell:
That explains it. Added to docs.

PR #13336, opened 2017-03-03
Feature 13196 - Added system setting to clear resource cache partially

sdrenth:
### What does it do?
Added a system setting where you can enable to clear resource cache partially by the provided contexts provided to the cache refresh method.

### Why is it needed?
When updating a resource the resources cache for all contexts would always be cleared. This option makes it possible to clear the cache only for the current active context. By default the new options is set to false, so it will not change anything unless the system setting is changed.

### Related issue(s)/PR(s)
Related issue:
#13196 

#modxbughunt

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -288,7 +288,7 @@ $settings['cache_resource_expires']->fromArray(array (
 $settings['cache_resource_clear_partial']= $xpdo->newObject('modSystemSetting');
 $settings['cache_resource_clear_partial']->fromArray(array (
      'key' => 'cache_resource_clear_partial',
-     'value' => '',
+     'value' => 0,
      'xtype' => 'combo-boolean',
      'namespace' => 'core',
      'area' => 'caching',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -285,6 +285,15 @@ $settings['cache_resource_expires']->fromArray(array (
   'area' => 'caching',
   'editedon' => null,
 ), '', true, true);
+$settings['cache_resource_clear_partial']= $xpdo->newObject('modSystemSetting');
+$settings['cache_resource_clear_partial']->fromArray(array (
+     'key' => 'cache_resource_clear_partial',
+     'value' => '',
+     'xtype' => 'combo-boolean',
+     'namespace' => 'core',
+     'area' => 'caching',
+     'editedon' => null,
+ ), '', true, true);
 $settings['cache_scripts']= $xpdo->newObject('modSystemSetting');
 $settings['cache_scripts']->fromArray(array (
   'key' => 'cache_scripts',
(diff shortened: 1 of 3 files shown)


Mark-H:
I like this.

OptimusCrime:
@Mark-H Any reason this can not be merged? Yet to be accepted or something? I like this feature too.

Mark-H:
Looks good to me, just didn't have enough waking time left during my last merge session to test and merge it ;)

opengeek:
The only problem I see is that this will not work with memcached. In fact, this probably only works with the file based cache as is. We might need to consider creating a resource partition per context in order to be able to do this effectively with any cache provider.

OptimusCrime:
@opengeek As someone how has never used the other cache systems. What would happen if you tried to do this with memcached or anything else than the file based one?

opengeek:
You can delete a folder in file-based cache, but memcached has no concept of folders, only individual keys, so this would delete nothing since there is no entry for the context, only entries with the context in the key.

opengeek:
In memcached you cannot efficiently delete a hierarchy of entries at all. It's flush the entire thing or delete individual entries one at a time.

OptimusCrime:
Hmm, that is interesting. Which approach would be best then? Could it be possible to prefix it? Or is there no obvious way to get around this?

sdrenth:
@opengeek Do you have any idea how we can fix the partial cache clearance for all caching handlers or is it maybe better to limit partial cache clearance only for when using the xPDOFileCache cache handler?

Because then I could add another check to only allow partial cache clearance for the xPDOFileCache handler by checking the cache_handler system setting, so then when using other caching handlers the behaviour of the resource cache clearance would remain the same as before.

sdrenth:
Since I can't edit my PR anymore, could someone perhaps replace case resource with the code below? I've changed the code so partial cache clearance only happens when using xPDOFileCache, so other caching types are not effected by this change.

```
case 'resource':
    $clearPartial = $this->getOption('cache_resource_clear_partial', null, false);
    $cacheHandler = $this->getOption('cache_handler', null, 'xPDOFileCache');

    if (!$clearPartial || $cacheHandler !== 'xPDOFileCache') {
        $results[$partition] = $this->clean($partOptions);
    } else {
        /* Only clear resource cache for the provided contexts. */
        foreach ($partOptions['contexts'] as $ctx) {
            $this->modx->cacheManager->delete(
                $ctx,
                array(
                    xPDO::OPT_CACHE_KEY => $this->modx->getOption('cache_resource_key', null, 'resource'),
                    xPDO::OPT_CACHE_HANDLER => $this->modx->getOption('cache_resource_handler', null, $this->modx->getOption(xPDO::OPT_CACHE_HANDLER)),
                    xPDO::OPT_CACHE_FORMAT => (int) $this->modx->getOption('cache_resource_format', null, $this->modx->getOption(xPDO::OPT_CACHE_FORMAT, null, xPDOCacheManager::CACHE_PHP))
                )
            );
        }
    }
    break;
```

OptimusCrime:
@modxbot close

Issue #13166, opened 2016-10-31
Empty directory shows arrow in file tree


sdrenth:
### Summary
An empty directory also has an arrow indicating the folder contains files, even though the directory is actually empty.

### Step to reproduce
- Create an empty directory
- Refresh the tree

### Observed behavior
An arrow is shown indicating the directory contains files.

Before clicking to expand directory:
![screen shot 2016-10-31 at 14 38 03](https://cloud.githubusercontent.com/assets/7802465/19855920/5724f256-9f77-11e6-8b07-64e9087bff76.png)

After clicking to expand directory:
![screen shot 2016-10-31 at 14 38 49](https://cloud.githubusercontent.com/assets/7802465/19855936/77ec25cc-9f77-11e6-9667-50a8e93ac7c8.png)

### Expected behavior
The arrow should not be shown, because the directory does not contain any files.

### Environment
- MODX: MODX 2.5.1-PL traditional

Mark-H:
I think the technical reason for this is that the tree does not yet know that the directory is empty. Without recursively going through directories just to list one level, I don't think this can be easily fixed, unless someone else has any good ideas?

Oetzie:
I'm going to try and fix this today.

OptimusCrime:
I agree with @Mark-H . I did look at this and that was the problem I encountered too. Unless @Oetzie has a good plan to fix this I'd recommend closing this off as wontfix.

OptimusCrime:
Closing as wontfix. In order to fix this issue we need an additional check for every directory to determine if it is empty or not. The I/O overhead for this operation is not worth the minimal gain of removing a few unnecessary arrows.

@modxbot close

PR #13337, opened 2017-03-03
Files tree, when the directory is empty the arrow wont appear (13309) #modxbughunt

Oetzie:
### What does it do?
Files tree, when the directory is empty the arrow wont appear

### Why is it needed?
The arrow in the file tree was visible on empty directories, changed it so it wont appear any more on empty directories

### Related issue(s)/PR(s)
n/a


rtripault:
Thanks for the contribution!

I'm just not sure that's something we want (performance wise)... otherwise :+1: for the UX

Mark-H:
That's a good point.. especially for remote media sources like S3 this might make it much slower. Hmm, difficult trade-off speed vs usability..

gpsietzema:
#13166

opengeek:
Any additional input on this from anyone?

gpsietzema:
I'm not sure if it is actually impacting performance. Checking the code it looks like one simple file-check. Not sure how heavy 'getContainerList' is though.

Mark-H:
getContainerList on every single result is going to request the contents of all subdirectories... try browsing a S3 media source and tell me again that this wont affect performance ;) 

It would probably be better to patch this directly in the media source getContainerList method. That way it could be handled differently for non-file media sources that are much slower than the file one.  

Given the small impact this has on the user (the negative aspect being that they might click on an empty directory to see it is empty), I'm going to propose this does not gets merged unless the performance impact, especially on remote media source types, is minified by not calling getContainerList just to count the children.

OptimusCrime:
I would be in favor of implementing this for file system media sources and NOT the other types.

Mark-H:
I'm going to close this for now for the reasons in my earlier comment, but feel free to suggest alternative approaches in a new PR.

Issue #8161, opened 2012-06-24
Make date output filter/modifier smarter


modxbot:
markh created Redmine issue ID 8161

The :date output filter/modifier needs the UNIX timestamp, which in some cases need the user to prepend the strtotime output filter, and sometimes it doesn't depending on the input.

I think it would be a nice convenience if the :date output filter would check the input to see if it is a unix time stamp (integer) or a other format, and use strtotime when it is in the wrong format.

Mark-H:
Included in 2.6

PR #13338, opened 2017-03-03
Output filters date/strftime now accept text-formatted datetime #modxbughunt

pbowyer:
### What does it do?
The :date output filter/modifier needs a UNIX timestamp, which in some cases need the user to prepend the strtotime output filter, and sometimes it doesn't depending on the input.

This PR changes that so the :date output filter checks the input to see if it is a unix time stamp (integer) or a other format, and use strtotime when it is in the wrong format.

### Why is it needed?
Added flexibility. Particularly now, we don't always store dates as Unix timestamps.

### Related issue(s)/PR(s)

#8161

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -438,18 +438,13 @@ class modOutputFilter {
                         case 'strftime':
                         case 'date':
                             /* See PHP's strftime - http://www.php.net/manual/en/function.strftime.php */
-                            if (empty($m_val))
+                            if (empty($m_val)) {
                                 $m_val = "%A, %d %B %Y %H:%M:%S"; /* @todo this should be modx default date/time format? Lexicon? */
-                            if (is_numeric($output)) {
-                                $value = 0 + $output;
-                            } else {
+                            }
+                            if (($value = filter_var($output, FILTER_VALIDATE_INT)) === false) {
                                 $value = strtotime($output);
                             }
-                            if ($value != 0 && $value != -1) {
-                                $output= strftime($m_val,$value);
-                            } else {
-                                $output= '';
-                            }
+                            $output = ($value !== false) ? strftime($m_val,$value) : '';
                             break;
 
                         case 'strtotime':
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Nice, but you are doing some pretty wizard-ish weak type casting here. Instead of doing ```$value = 0 + $output;``` what about the more readable ```$value = intval($output);```

pbowyer:
@OptimusCrime That's original MODX code that is! https://github.com/modxcms/revolution/pull/13338/commits/deaee44533a3b30fc7c3fc4b8478d9c438a110b2#diff-3b43f92f730c95fab1bd225d72f76928L443

Jako:
Maybe it is better to have a strict check for an integer with [filter_var](http://php.net/manual/de/filter.filters.validate.php). Numeric is too broad.

```
if (filter_var($output, FILTER_VALIDATE_INT) === false) {
```

OptimusCrime:
@pbowyer Oh, I'm sorry. Still fishy code. The core is filled with weird weak type checking and casting which is prone to errors. I'm trying to get rid of these where we can. @Jako 's idea was good. It would be nice if you could alter your PR.

pbowyer:
> The core is filled with weird weak type checking and casting which is prone to errors.

First time I've worked on a codebase without a fixed coding standard and rules around not changing what's already present 😀 

Here we go - I've expanded the rewrite to simplify down the entire `case` statement. @Jako made a great point and I've switched to `filter_var`.

OptimusCrime:
Very nice. Looks good now :)

opengeek:
@pbowyer Can you edit the base branch of this PR to target 2.x instead of 2.5.x?

opengeek:
Er, nevermind @pbowyer — apparently I can do that myself and didn't realize it until now. 🤦‍♂️

pbowyer:
@opengeek np, it's a nice new(ish) feature of GitHub

Mark-H:
Hey, I requested that feature 5 years ago! I don't remember doing that, but fun that I'm now the one to merge it in 😁 

Thanks @pbowyer, merged for inclusion in 2.6.

PR #13341, opened 2017-03-03
very basic try to fix #12380 #modbughunt.

### LINKED ISSUE #12380: Ugly error page when symlink target isn't available
When the symlink target isn't available (not published or no rights to access it) MODX isn't showing one of the configured error pages. Instead it constructs its own HTML code.

Wouldn't it more feasible to call sendErrorPage() in modx.class.php's sendForward() instead of sendError()?

achterbahn:
### What does it do?
symlink to non existent resource id leads to a sendError() instead of sendErrorPage()

### Why is it needed?
I think it is just a esthetic problem?

### Related issue(s)/PR(s)
related to Issue #12380

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1137,6 +1137,8 @@ class modX extends xPDO {
                 }
                 $this->request->prepareResponse();
                 exit();
+            } else {
+                $this->sendErrorPage();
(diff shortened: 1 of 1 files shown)


Mark-H:
Sometimes very basic is all you need ;)

Issue #13167, opened 2016-11-02
Using Memberof with conditionals generates 500 error


wandering213:
### Summary
When I try to use conditionals to output if a user is a member of a group I get a 500 error. If I don't include 'then/else the error is gone.

### Step to reproduce
```
[[!+modx.user.id:mo=`GroupName`:eq=`1`:then=`True`:else=`False`]]
```

### Observed behavior
Generates a 500 server error

### Expected behavior
I would expect it to output True or False

### Environment
MODX version 2.5.1, nginx 1.11.5, mysql 5.7.16, Ubuntu 16.04

pixelchutes:
What happens if you remove the ```:eq=`1````?

wandering213:
That resolves the error.

OptimusCrime:
Without the eq statement, is the behavior still correct?

wandering213:
Yes

Jako:
A conditional could not combined directly with the next conditional as far as I know. There has to be a some conjunction between like `and:` or `or:`. But the Internal Server Error is somehow a bug and should be removed.

OptimusCrime:
My thoughts exactly.

OptimusCrime:
Traced the bug down to lines like these: https://github.com/modxcms/revolution/blob/2.x/core/model/modx/filters/modoutputfilter.class.php#L145

Conditional arrays should be `[1, '||', 1]`, `[0, '&&', 1]`, and similar. Problem happens when they do not follow the format `[value, operand, value, operand, value]` and so on. In this case the array is `[1, 1]` (two true after each other), which fails the eval function. 

I am going to submit a PR once I get a chance, but I think the approach would be to loop the array and make sure we have a int first and last, and alternating operands and ints between those.

Mark-H:
Fixed in 2.6

PR #13343, opened 2017-03-03
Better sanitizing of conditionals before eval'ing #modbughunt

OptimusCrime:
### What does it do?
This PR fixes a crash that occur if output modifiers are used incorrectly.

The following code causes a PHP error:

```
[[!+modx.user.id:mo=`Administrator`:eq=`1`:then=`Member`:else=`Not member`]]
```

The reason is that the output modifiers uses PHPs `eval` to parse the conditions and their values. In the example above, we try to eval an expression that ends up being `1 1`. This is incorrect syntax. This PR introduces better handling of sanitizing these expressions. The expressions should be on the format `[integer, operator, integer, operator, integer, ...]` where the first and last values has to be integers (or bools), and every odd value has to be operators (either `||` or `&&`). Every even value also has to be integer/bools.

This PR also cleans up some code and removes duplicated code blocks. 

### Why is it needed?
Fixes error

### Related issue(s)/PR(s)
Reported in #13167

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -36,7 +36,7 @@ class modOutputFilter {
 
     /**
      * Filters the output
-     * 
+     *
      * @param modElement $element The element to filter
      */
     public function filter(&$element) {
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Sorry, the code in the PR description lacked a character. Could you try again now? The error I get looks like this:

```
Parse error: syntax error, unexpected '0' (T_LNUMBER) in ..../revolution/core/model/modx/filters/modoutputfilter.class.php(178) : eval()'d code on line 1
```

Mark-H:
Looks like I had to switch to PHP 7.1 to reproduce, wasn't triggering on 5.6.

OptimusCrime:
Ah, sorry. I forgot to mention that. I think this only happens in 7.x I am currently using 7.0.15-1.

Mark-H:
I can confirm this fixes the issue but this targets 2.x (and for parser changes, I think that's a very good idea), so I'll leave this open until a later time (or someone else to integrate it) for 2.x. I've only been doing 2.5.x merges today. :P

Mark-H:
Thanks @OptimusCrime, merged for 2.6.

opengeek:
Are we aware that this completely broke the `isempty` output filter? After months of trying to track down why my sites were no longer working in the 2.6/2.7 releases, I finally tracked the problem down to this commit.

The following returns the value of the placeholder instead of the `else` when the value is not empty:
```[[+placeholder:isempty:then=`foo`:else=`bar`]]```

OptimusCrime:
I was not aware of that at all. It passed all the tests, and it worked fine across all my sites. Are there any other filters similar to `isempty` that functions this way? I guess I could write a fix for this.

opengeek:
Sorry, I misstated the problem — it returns the placeholder if it is NOT empty instead of the else.

Jako:
@opengeek `isempty` is no conditional operator. Thats the reason why your construct does not work. It does not extend the $condition array, so then/else does not work properly (it seems that `then` is working, because validateConditionals of an empty array is true - that could be a bug).

`isempty` works the same as `default`. The syntax is ```[[+placeholder:isempty=`foo`]]```.

You have to change the placeholder construct to ```[[+placeholder:eq=``:then=`foo`:else=`bar`]]``` to have then/else available.

opengeek:
Horse crap. It worked perfectly up until this commit, period.

opengeek:
@Jako It worked perfectly like this up until this commit. It is a valid bug because it breaks BC.

Jako:
I don't know how to argue better, but the right syntax for the `isempty` modifier is ```[[+placeholder:isempty=`foo`]]```. It is working equal to ```[[+placeholder:default=`foo`]]```. There is no valid usage of `then/else` afterwards.

The commit changes the output of a wrong used modifier. Is that really a BC?

opengeek:
I've been using this syntax successfully since 2.0 @Jako. I'm lost on why you are saying this is improper usage.

opengeek:
Anyway, forget about it. I'm closing the issue. I'm just glad I figured out why every site I had stopped working in 2.6.

Issue #12542, opened 2015-08-09
modResource->getTVValue/setTVValue is_string needs changed to is_numeric


wshawn:
Based on this link: http://stackoverflow.com/questions/14052207/modx-revo-settvvalue-returns-old-data/31909888#31909888

I tested out the following scenario:

``` php
$changeit = 10;
//$changeit= 'somename';
// $changeit='10';

$r = $modx->getObject('modResource', 177);
$tvs = $r->getTemplateVars();
 if ($tvs) {
        foreach ($tvs as $object){
            if (is_object($object) && $object instanceof modTemplateVar){
               print_r($object->toArray());
            }
        }
    }

$r->setTVValue($changeit,'99999fhfhg');
echo '<hr>'. $r->getTVValue($changeit);
$r->setTVValue($changeit,'888885454564');
echo '<hr>'. $r->getTVValue($changeit);
$r->setTVValue($changeit,'123456789');
echo '<hr>'. $r->getTVValue($changeit);
```

If the numeric 10 is sent to the two functions it is evaluated as a non string, but due to is_string if '10' is sent it as a string and thereby becomes the name of the TV instead of the id.

I suggest changing

``` php
$byName = false;
        if (is_string($pk)) {
            $byName = true;
        }
```

To:

``` php
       $byName = (is_numeric($pk)) ? false : true;
```

Affects:
#11486
#7453

pbowyer:
Claiming this for myself :)

Mark-H:
Fixed in 2.5.6

PR #13346, opened 2017-03-03
Bug 12542 getTVValue/setTVValue consider numeric TV names as IDs #modxbughunt

pbowyer:
### What does it do?
Previously any strings received by getTVValue/setTVValue were considered to be the name of a template variable. This caused problems when an identifier like `"240"`. 

### Why is it needed?
As PHP users are used to typecasting on the fly, and as users are unlikely to name TVs as numbers, I've made it check for numeric TV names and treat them as strings.

### Related issue(s)/PR(s)

#12542

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -823,7 +823,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
      * TV is not found.
      */
     public function getTVValue($pk) {
-        $byName = (is_numeric($pk)) ? false : true;
+        $byName = !is_numeric($pk);
 
         /** @var modTemplateVar $tv */
         if ($byName && $this->xpdo instanceof modX) {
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Good! You can write the line 

```
$byName = (is_numeric($pk)) ? false : true;
```

as 

```
$byName = !is_numeric($pk));
```

Which I find more readable.

pbowyer:
Do the MODX coding standards have anything to say? I prefer to be explicit, but happy to change to whatever fits.

OptimusCrime:
There is none. I just saw the original issue and @wshawn suggested code. I prefer avoiding conditional operator in cases where the value can be directly inferred from the original call. All you really do here is negate the original value.

pbowyer:
There we go, I've modified the code

OptimusCrime:
Awesome @pbowyer :)

Mark-H:
Thanks @pbowyer!

Issue #12520, opened 2015-07-22
Improving cache generation/resource save for large sites


markwillis82:
I've been spending some time improving the save time of resources within the manager when having 100k + resources, and have tracked the problem down to 2 items. Both related to this file: https://github.com/modxcms/revolution/blob/master/core/model/modx/mysql/modcontext.class.php

The first is the JOIN for `modContextResource` and second being no index that can be used. From my findings I cannot find anywhere that the `modContextResource` table is actually populated (I've got multiple single and multi context sites) but cannot get any new rows added to this table.

I have refactored these functions to be a lot quicker (down from 16 seconds to 1 second on my local machine for 250,000 resources) but want to be sure it won't break anything by removing this table from the query.

Is this some old/dated code that can be removed?

There is a great comment here:
https://github.com/modxcms/revolution/blob/master/core/model/modx/modcontextresource.class.php#L12
`@todo Work this relationship into use in the manager and the logic of each` dating back to 2011

OptimusCrime:
Wow, this is amazing. If this works, dude.

OptimusCrime:
I did some digging myself and found the following line in the [changelog](https://github.com/modxcms/revolution/blob/9df6a1b6529ddfd14183a814366caa910653ba71/core/docs/changelog.txt#L2880):

> modCacheManager->generateContext(): Added support for Resources to be generated in multiple contexts via modContextResource.

From version 2.0.0-beta-1.

Other results are the one you linked and mentioned once in a the method for returning resources, that I can't decode or figure out.

markwillis82:
I should add... the only references to this object are:
```
./core/docs/changelog.txt
./core/model/modx/metadata.mysql.php
./core/model/modx/metadata.sqlsrv.php
./core/model/modx/modcontextresource.class.php
./core/model/modx/modrequest.class.php
./core/model/modx/mysql/modcontext.class.php
./core/model/modx/mysql/modcontext.map.inc.php
./core/model/modx/mysql/modcontextresource.class.php
./core/model/modx/mysql/modcontextresource.map.inc.php
./core/model/modx/mysql/modresource.map.inc.php
./core/model/modx/sqlsrv/modcontext.class.php
./core/model/modx/sqlsrv/modcontext.map.inc.php
./core/model/modx/sqlsrv/modcontextresource.class.php
./core/model/modx/sqlsrv/modcontextresource.map.inc.php
./core/model/modx/sqlsrv/modresource.map.inc.php
[...]

PR #13347, opened 2017-03-03
Stop calling generateContext twice on resource update #modxbughunt

christianseel:
### What does it do?
This sets the `$regenerate` flag on the `modContext:prepare()` function to false, which stops the `generateContext()` call in that case. That call is not required, as the resource update processor also calls a `clearCache` function which refreshes the context cache already.

### Why is it needed?
Before, the context cache was regenerated *twice*. This results in very (!) long execution times for larger sites and is totally useless.

### Related issue(s)/PR(s)
We discovered this while working on https://github.com/modxcms/revolution/issues/12520, through it's not related to that.

#modxbughunt


opengeek:
Nice catch guys.

Mark-H:
Thanks @christianseel - merged into 2.5.6

PR #13348, opened 2017-03-03
Use (but limit) system setting for results per page in package management grid #modxbughunt

### LINKED ISSUE #12518: Default Per Page number of packages listed not related to the Default Per Page System Setting

mkschell:
Fixes #12518 #modxbughunt

Package management results per page was hard-coded to 10. The bug report suggested using the system setting `default_per_page` but there could be performance problems if this number was really high (since that grid may cause remote requests out to the package provider). So a default cap of 25 was applied to the system setting in this case.

### DIFF
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -74,7 +74,6 @@ MODx.grid.Package = function(config) {
         };
     }
 
-    var pageSize = parseInt(MODx.config.default_per_page)
     Ext.applyIf(config,{
         title: _('packages')
         ,id: 'modx-package-grid'
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -83,7 +83,6 @@ MODx.grid.Package = function(config) {
                  ,'provider','provider_name','disabled','source','attributes','readme','menu'
                  ,'install','textaction','iconaction','updateable']
         ,plugins: [this.exp]
-        ,pageSize: 10
         ,columns: cols
         ,primaryKey: 'signature'
         ,paging: true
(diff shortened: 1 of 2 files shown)


OptimusCrime:
You introduce a new variable here that is not used as far as I can see. 

Also, instead of using the number class you can write the logic as

```
Math.min(parseInt(MODx.config.default_per_page), 25);
```

Other than that, good work :)

Mark-H:
Test failure is due to a XML request timeout, not related to the PR.

netProphET:
@OptimusCrime I had removed the variable originally, then added it back in to impose the limit. Unless you're referring to the introduction of another variable?

Thanks for the suggestion of Math.min(). Committed.

OptimusCrime:
Looks good, but you should remove the variable here: https://github.com/modxcms/revolution/pull/13348/files#diff-283367202eada3aed53c62d14f3604ddR77

rtripault:
Just wanted to link to #13088

opengeek:
@netProphET I believe they are referring to the definition of `var pageSize` which is never used.

manager/assets/modext/workspace/package.grid.js
```diff
@@ -74,6 +74,7 @@ MODx.grid.Package = function(config) {
         };
     }
 
+    var pageSize = parseInt(MODx.config.default_per_page)
```
opengeek: I don't believe this is used anywhere.

Issue #13165, opened 2016-10-31
Moving directories up/down adds dir as a subdirectory


sdrenth:
### Summary
When moving directory up/down inside the files tree, the directory gets moved inside the directory directly below it after dragging the folder.

### Step to reproduce
- Go to the Files tab inside the MODX manager
- Create directories:  
    - Test 1
    - Test 2
    - Test 3
- Move Test 2 above Test 1
- Refresh the tree

### Observed behavior
Instead of moving the directory "Test 2" above "Test 1", the directory is now added as a child directory to "Test 1".

### Expected behavior
I expected that after moving a directory above Test 1, that the directory Test 2 is shown above Test 1, and is not added as a subdirectory.

### Environment
- MODX Version: Modx Revolution 2.5.1-PL (Clean install from traditional installation package)
- Google Chrome 54.0.2840.71 (64-bit)
- Nginx

EasterPeanut:
Also noticed similar unexpected behavior of the file tree in the media Browser when dragging files. 

1. Go to the media browser
2. Drag a file from inside a directory to the root

Opgemerkt dat de file tree van de media browser raar gedrag vertoont. Als ik bijvoorbeeld een afbeelding tussen twee mappen in sleep, verplaatst de afbeelding zich soms naar in een map die onder het divider-streepje staat. Af en toe gaat het wel goed. Bij een aantal sites getest en hetzelfde ondervonden.

labr1005:
I'll try to fix this today.

labr1005:
@sdrenth @EasterPeanut 

IMO sorting of directories or files should not be possible at all. Why would you want to do this? They should always be sorted alphabetically.

It should only be possible to drag directories/files inside other directories.

sdrenth:
@labr1005 @EasterPeanut I don't think it is necessary to sort directories/files, but the problem is that the UI makes it look like it can. And as I described above, the directory is then being moved inside the other directory instead of being moved above it.

Mark-H:
Fixed in 2.5.6

PR #13349, opened 2017-03-03
Fixed issue #13165 Moving directories up/down adds dir as a subdirect…

labr1005:
…ory #modxbughunt

### What does it do?
Disables the possibility to sort directories or files.

### Why is it needed?
It should only be possible to drag directories/files into other directories (or root).

### Related issue(s)/PR(s)
modxcms/revolution#13165


Mark-H:
Thanks @labr1005 - merged for 2.5.6.

PR #12512, opened 2015-07-14
[#12508] Validate chmod value

garryn:
PR related to #12508.

Issue #12508, opened 2015-07-11
Chmod'ing a directory from file tree should check for valid value


garryn:
Chmod'ing a directory currently allows any text to be passed as the chmod value - MODX should be checking for a valid value.

Passing a non-valid value (e.g. "lockmeout") as the chmod value can lock the permissions down on the folder making it impossible to access until you chmod it back correctly via SSH/Cpanel etc.

Mark-H:
hmm I wonder if this should be checked in the browser/directory/chmod processor, or in the modMediaSource->chmodContainer method? Or both?

garryn:
I think the check should be in modFileMediaSource::chmodDirectory() and let the error bubble up through the processor (if I'm reading the code right, I haven't looked too long at this). Thoughts?

garryn:
Actually, having looked, checking it in modFileResource::chmod() makes more sense.

Mark-H:
Fixed in 2.3.6

OptimusCrime:
I can take a look at this

@OptimusCrime

OptimusCrime:
New PR opened. #modxbughunt

OptimusCrime:
This is fixed now :)

@modxbot close

PR #13352, opened 2017-03-03
Validate chmod input

OptimusCrime:
### What does it do?
Validate the chmod input. It should be integer values only.

#modxbughunt

### Why is it needed?
There is no validation as of today, allowing the user to input anything he or she wants.

### Related issue(s)/PR(s)
Originally reported in #12508 . First implementation in #12512 but it was reverted. This PR contains the suggested fix.


theboxer:
Cherry picked to 2.5.x

PR #13354, opened 2017-03-03
Fix issue 12520, #modxbughunt

wuuti:
### What does it do?
Introduces a new system setting 'use_context_resource_table' (set to true by default) which enables you to disable inefficient and unnecessary queries for context refreshes. 
It also logs a warning message when long running queries are detected and gives a hint in the log message to use the system setting.

### Why is it needed?
The mysql/modcontext class implements statements with legacy queries. Those queries contain the context_resource table, which is never populated except you do that with a custom api call (details described in #12520). On large sites with a lot of resources and multiple contexts this makes the manager quite unusable (times measured e.g. to save a resource: +6s, or to save a system setting over 60s in system with 280K resources and 12 contexts). Also the queries do not respect the cache_alias_map setting und return unnecessary data. 

### Related issue(s)/PR(s)
Loose relationship to #13347, which "doubles" the performance problem from here by doing every query twice.


christianseel:
@wuuti that PR contains commits from others. Potentially it was based on the 2.5.x branch and is now supposed to be merged into 2.x. I think you need to open a new PR based on 2.x to merge it in 2.x

PR #13358, opened 2017-03-03
Prevent inline edit on recently edited dashboard widget causing aliases to change #12129 #modxbughunt

### LINKED ISSUE #12129: Recently Edited Resources dashboard widget: Update from grid errors
The "Recently Edited Resources" dashboard widget apparently allows you to make edits from the grid, however the updates don't succeed, instead errors with a 500 status. I presume this is unintended behavior, and probably should be disabled.

syts88:
…hboard widget #modxbughunt

### What does it do?
In the recent edited resources widget you can edit the pagetitle on double click. But by doing that you also change the alias automaticaly. So i turned the editor function off for the pagetitle. Right click still works ofcourse.

### Why is it needed?
So that clients don't change their urls of pages by accident.

### Related issue(s)/PR(s)
#12129

### DIFF
--- a/manager/assets/modext/widgets/security/modx.grid.user.recent.resource.js
+++ b/manager/assets/modext/widgets/security/modx.grid.user.recent.resource.js
@@ -28,7 +28,6 @@ MODx.grid.RecentlyEditedResourcesByUser = function(config) {
             header: _('pagetitle')
             ,dataIndex: 'pagetitle'
             //,width: 150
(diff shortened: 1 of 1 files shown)


Mark-H:
Merged into 2.5.6, thanks!

pixelchutes:
Merged via 1ead28c7af6dc6b9bb37f889bbe57e621ed76306

Issue #12510, opened 2015-07-13
Selfaware chunks?


pepebe:
Just a quick idea

Can we add some default placeholders to getChunk with meta data like id and name?

Example:
To make debugging easier, I usually write my chunks like this:
```
<div data-chunk-id="1" data-chunk-name="hello world"  >
Hello World
</div>
```
This is helpful, but not always foolproof because you have to update name and id from time to time.

Writing it like this would make it easier:

```
<div data-chunk-id="[[+chunk_id]]" data-chunk-name="[[+chunk_name]]">
Hello World
</div>
```

Regards,

pepebe

rthrash:
Neat idea. I could see that being really useful.

pepebe:
I'm not ready for a pull request, but I'd like to post some code that would have to be changed to make chunks a bit smarter:

Line 486: core/model/modx/modparser.class.php

````
                case '$':
                    $tagName= substr($tagName, 1 + $tokenOffset);
                    if ($element= $this->getElement('modChunk', $tagName)) {
                        $element->set('name', $tagName);
                        $element->setTag($outerTag);
                        $element->setCacheable($cacheable);
                        
                        $tagPropString['chunkName'] = $tagName;
                        $tagPropString['chunkId'] = $element->get('id');
                        
                        $elementOutput= $element->process($tagPropString);
                    }
                    break;
````

Line : core/model/modx/modx.class.php

````
    public function getChunk($chunkName, array $properties= array ()) {
        $output= '';
        if ($this->getParser()) {
            $chunk= $this->parser->getElement('modChunk', $chunkName);
            if ($chunk instanceof modChunk) {
                $chunk->setCacheable(false);
                $properties['chunkName'] = $chunkName;
                $properties['chunkId'] = $chunk->get('id');
                $output= $chunk->process($properties);
            }
        }
        return $output;
    }
````

A new parameter for getChunk() would be nice. 

```
public function getChunk($chunkName, array $properties= array () $debug = false) {}
```

Setting debug to something other than false (for example, comment or log) could help to track down performance problems.

Things that might be useful:
* Cached on: 2016-12-12 20:34:12
* Required processing time: 0.0153324s (easy).

I guess the caching related stuff is located around process(), right?

mindeffects:
+1

Mark-H:
To avoid possible conflicts, maybe stuff like that should go into a _self placeholder, like `[[+_self.id]]` and `[[+self.name]]`.

[...]

PR #13360, opened 2017-03-03
Fix issue #12510 (mgr performance large installations), #modxbughunt

wuuti:
(cherry picked from commit 78b915ce278fd3fb9743776f732542a76ba98a94)
Replaces wrongly based PR 13354

### What does it do?
Introduces a new system setting 'use_context_resource_table' (set to true by default) which enables you to disable inefficient and unnecessary queries for context refreshes. 
It also logs a warning message when long running queries are detected and gives a hint in the log message to use the system setting.

### Why is it needed?
The mysql/modcontext class implements statements with legacy queries. Those queries contain the context_resource table, which is never populated except you do that with a custom api call (details described in #12520). On large sites with a lot of resources and multiple contexts this makes the manager quite unusable (times measured e.g. to save a resource: +6s, or to save a system setting over 60s in system with 280K resources and 12 contexts). Also the queries do not respect the cache_alias_map setting und return unnecessary data. 

### Related issue(s)/PR(s)
Loose relationship to #13347, which "doubles" the performance problem from here by doing every query twice.

### DIFF
--- a/core/model/modx/mysql/modcontext.class.php
+++ b/core/model/modx/mysql/modcontext.class.php
@@ -22,7 +22,9 @@ class modContext_mysql extends modContext {
             $resourceFields= array('id','parent','uri');
 
             // we do not need to select uri if cache_alias_map is set to false
-            if ($cache_alias_map==0) $resourceFields = array('id','parent');
+            if ($cache_alias_map == 0) {
+                $resourceFields = array('id','parent');
+            }
 
             $resourceCols= $context->xpdo->getSelectColumns('modResource', 'r', '', $resourceFields);
             $contextKey = $context->get('key');
@@ -50,7 +52,9 @@ class modContext_mysql extends modContext {
 
             // output warning if query is too slow
             $time = ((microtime(true)-$time));
-            if ($time >= 1.0 && $use_context_resource_table==1) $context->xpdo->log(2,"[modContext_mysql] Slow query detected. Consider to set 'use_context_resource_table' to false.");
+            if ($time >= 1.0 && $use_context_resource_table==1) {
+                $context->xpdo->log(modX::LOG_LEVEL_WARN,"[modContext_mysql] Slow query detected. Consider to set 'use_context_resource_table' to false.");
+            }
         }
         return $stmt;
     }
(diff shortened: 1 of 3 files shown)


wuuti:
There was an error in the fix, I corrected it in the last commit. The context key condition was missing on the reduced queries.

sdrenth:
Tested this PR with approx 30k resources and can confirm that it is working.

OptimusCrime:
This PR is awesome. +1000000!

@sdrenth Awesome work with the testing lately. Very nice work!

core/model/modx/mysql/modcontext.class.php
```diff
@@ -12,33 +12,88 @@ class modContext_mysql extends modContext {
...
+            $cache_alias_map = $context->getOption('cache_alias_map');
+            $use_context_resource_table = $context->getOption('use_context_resource_table',null,1);
+
             $tblResource= $context->xpdo->getTableName('modResource');
             $tblContextResource= $context->xpdo->getTableName('modContextResource');
+
             $resourceFields= array('id','parent','uri');
+
+            // we do not need to select uri if cache_alias_map is set to false
+            if ($cache_alias_map==0) $resourceFields = array('id','parent');
```
Mark-H: Code style should be more like

`if ($cache_alias_map == 0) {
    $resourceFields = ..
}`

core/model/modx/mysql/modcontext.class.php
```diff
@@ -12,33 +12,88 @@ class modContext_mysql extends modContext {
...
+            }
+
             $criteria = new xPDOCriteria($context->xpdo, $sql, $bindings, false);
             if ($criteria && $criteria->stmt && $criteria->stmt->execute()) {
                 $stmt =& $criteria->stmt;
             }
+
+            // output warning if query is too slow
+            $time = ((microtime(true)-$time));
+            if ($time >= 1.0 && $use_context_resource_table==1) $context->xpdo->log(2,"[modContext_mysql] Slow query detected. Consider to set 'use_context_resource_table' to false.");
```
Mark-H: Code style here as well.

Instead of log level 2, please use the xPDO level constants so it's clear from the code what level it is logging at.

core/model/modx/mysql/modcontext.class.php
```diff
@@ -12,33 +12,88 @@ class modContext_mysql extends modContext {
...
+                $sql .= "   AND `r`.`context_key` = ?";
+            }
+
             $criteria = new xPDOCriteria($context->xpdo, $sql, $bindings, false);
             if ($criteria && $criteria->stmt && $criteria->stmt->execute()) {
                 $stmt =& $criteria->stmt;
             }
+
+            $time = ((microtime(true)-$time));
+            if ($time >= 1.0 && $use_context_resource_table==1) $context->xpdo->log(2,"[modContext_mysql] Slow query detected. Consider to set 'use_context_resource_table' to false.");
```
Mark-H: Ditto as above

Mark-H (approved):
Some small changes here, mostly code style, other than that it looks good.

Issue #12714, opened 2015-10-12
Searchbar does not react to 'Enter' keypress


whitebyte:
While we able to navigate through Uberbar search results with `up` and `down` arrow keys, the only way to open currently selected entry is a mouseclick. I think it should be opened by hitting `Enter` key as well.

pixelchutes:
I cannot reproduce (v2.4.2-pl) -- Perhaps try clearing your local browser cache?

whitebyte:
Hmm, it looks like Firefox-specific issue. Tried on both Windows and Linux, 2 different PCs. Chrome works just fine, FF ignores Enter and Ctrl+Enter.
MODX 2.4.0, 2.4.2

pixelchutes:
Confirmed.
> TypeError: e is undefined

jpdevries:
This is an unfortunate bug but it does bring an accessibility concern to the surface. If the uberbar were an HTML-first semantic `<form>` that posted to a search results landing page then the search feature would still be accessible even in cases of script breakage like this one.

The uberbar is currently an `<input>` inside `<div>` and lacks an associated `<form>`.  
![](http://j4p.us/image/292w3q3N3P1m/Screen%20Shot%202015-11-29%20at%205.03.41%20AM.png)

Mark-H:
Looks like firefox simply doesn't have window.event: http://stackoverflow.com/a/9813523

EasterPeanut:
I will try to fix this today.

Mark-H:
Fixed in 2.5.6

PR #13364, opened 2017-03-03
Fixed issue #12714 #modxbughunt

EasterPeanut:
### What does it do?
It fixes the bug in the search bar in Firefox > cannot use enter to go to the page when using the search bar.

### Why is it needed?
Usability for Firefox users.

### Related issue(s)/PR(s)
#12714


Mark-H:
Thanks!

PR #13365, opened 2017-03-03
#9039 Added validation for min and max length of field #modxbughunt

alroniks:
### What does it do?
This PR adds validators for TV Text renderer for max and min length fields.

### Why is it needed?
Problem was that before you could set max more that min value and you had no ability to properly check the field, because it was always wrong. This validation rules hightligth when you try to input max value less that min value and also when you try set min value more than max value. Also, I moved min value at first, because usually, people set min value at first. So just UX enhancement.
Unfortunately, it does not prevent save TV configuration with wrong values. 

Some screenshots:
First one
template variable- childtv - modx revolution 1
Second one
template variable- childtv - modx revolution

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/9039


OptimusCrime:
I did not understand that is was two images at first. Perhaps writing before and after before the images?

But good job!

Alroniks:
@OptimusCrime Yes, two images without delimiter was a bit confusing :)
I found that the similar solution also can be expanded for numberfield with min/max value.
And the main issue - don't give ability to save form in case of error. There used standard extjs validation, but it doesn't impact to whole TV form. And I don't know yet how to solve it.

rtripault:
While testing the PR, i've been surprised by the fact you can have a non valid input properties form and still submit the form (as you noted), so i tried to find a way to prevent that.

I added a listener to the panel (inputproperties/text.tpl) as follow

```js
listeners: {
        beforeRender: {
            fn: function(me) {
                // Target the whole form panel/page
                var form = Ext.getCmp('modx-panel-tv');
                form.on('beforeSubmit', function() {
                    // Before submitting the complete form, let's try to validate our own fields, assuming it is valid as default
                    var valid = true;
                    //  Iterate each panel items, looking for a validate method
                    Ext.each(me.items.items, function(field) {
                        if (field.validate) {
                            // Seems like we can validate this component, let's do so
                            if (!field.validate()) {
                                // Component did not validate, let's prevent saving the whole form
                                valid = false;
                                console.log('field', field.getActiveError());
                                // @TODO notify user
                                // note : you can either stop iterating here, or continue the tests and return all errors to the user
                            }
                        }
                    });
                    return valid;
                })
            }
        }
    }
```

Hoping that could help you going further.

PR #13370, opened 2017-03-06
Added grunt-cli as a dev dependency rather than a global requirement

rtripault:
### What does it do?

Adds grunt-cil as a development dependency

### Why is it needed?

The current workflow involves installing grunt-cli globally. Having it as a development dependency could be convenient to automate some builds with CI

/cc @jpdevries


Mark-H:
Better separating global and local dependencies sounds like an improvement... but this is sure becoming a lot of typing to do a build :/

rtripault:
Happy to have it closed then!

Issue #13371, opened 2017-03-06
Edit context in grid adds context key multiple


Jako:
### Summary
Edit context directly in the grid adds the context key in brackets multiple.

### Step to reproduce
Edit a context directly in the context grid by i.e. changing the description and reload the grid afterwards. 

### Observed behavior
The name column contains the context key in brackets twice (or even more - each edit adds one context key in brackets).

### Expected behavior
It should contain the context key once.

### Environment
MODX 2.5.5

OptimusCrime:
Sent a PR for this.

achterbahn:
I've never came to an end with my try to fix this issue. So discard my commits and go for @OptimusCrime 's solution.

OptimusCrime:
No problem @achterbahn. It it turns out that my approach is too fragile and open for bugs, we might end up using your solution.

PR #13372, opened 2017-03-06
Fix: Edit context in grid adds context key multiple #13371

achterbahn:
The name field in the grid was concatenated from context-name and context-key. PR removes the concatenation. Only the context name is shown in the name field.

### DIFF
--- a/core/model/modx/processors/context/getlist.class.php
+++ b/core/model/modx/processors/context/getlist.class.php
@@ -59,7 +59,6 @@ class modContextGetListProcessor extends modObjectGetListProcessor {
         if (!empty($contextArray['name'])) {
             $contextArray['name'] .= ' ';
         }
-        $contextArray['name'] .= "({$contextArray['key']})";
         $contextArray['perm'] = array();
         if ($this->canEdit) {
             $contextArray['perm'][] = 'pedit';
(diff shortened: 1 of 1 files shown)


Jako:
The getlist processor could be used elsewhere, i.e a combobox could use it. Better create a different column with the concatenated column, search for those getlist occurrences and change the display field to the concatenated column.

achterbahn:
What happens when double-clicking on the name filed in the context grid? Is then the display fileld replaced with the "real" value? 
image
Or should there be two fields?

Jako:
Yes, the 'name' column is changed by the getlist processor, it is displayed in the grid cell and saved with that changed value during inline editing in the grid by the updatefromgrid processor.

Since the name column of the getlist processor better contains an unchanged value, a new column should contain the changed value, which is uses in the combo box (or the combo box could use a template - but that would have a display issue).

Jako (dismissed):
Add a new column (i.e. $contextArray['name_key']) with the concatenated value in the processor, change the combo box displayField and add the new column to the fields array in https://github.com/modxcms/revolution/blob/8ddd6a3e8c9aa43ad3aa32a58c8ee06140673cca/manager/assets/modext/widgets/core/modx.combo.js#L316-L318

sottwell:
Most of the original tables have the "name" field set to 50 chars, and I haven't heard anybody complain about that.

theboxer (approved):
I agree with this change, the grid should show the real value.

Issue #13363, opened 2017-03-03
If context_tree_sort is set to false, disable drag and drop of contexts [#modxbughunt]


OptimusCrime:
### Summary
The context "blocks" are drag and droppable. This can be used to reorder them as they have a rank system similar to resources. However, if the system setting `context_tree_sort` is false the state will not be stored. This is confusing. It would be better that in scenarios where `context_tree_sort` is set to false, dragging and dropping is disabled.

### Step to reproduce
Set `context_tree_sort` to false (system default). Reorder contexts by dragging them. Refresh site.

### Observed behavior
Reordering is not saved.

### Expected behavior
Reordering is not possible.

### Environment
modxcms/revolution@842f03fbdb7b7bfef1b91cc15db9639f53026096

jcdm:
+1 - just discovered this via a client who'd been trying to drag and drop sort but finding it didn't 'stick'. Was in 2.5.5.

Mark-H:
Fixed in 2.6

PR #13376, opened 2017-03-06
Disable dd of contexts if context_tree_sort is false

OptimusCrime:
### What does it do / Why is it needed?
With the current implementation contexts can be reordered by dragging and dropping them. However, if the system setting `context_tree_sort` is false the state will not be saved. This behavior is counter intuitive as it "fools" the user to belive dragging and dropping is enabled and works.

This PR checks the value of the system setting `context_tree_sort` and if it is disabled it will make the root nodes in the resource tree undraggable. These nodes corresponds to the context nodes (top level).

### Related issue(s)/PR(s)
#13363


Mark-H:
Thanks @OptimusCrime!

iamdave210:
I didn't know about this setting! I'm so happy to be able to reorder contexts now!

Issue #13367, opened 2017-03-05
Critical error when invoking not existant widget


MarkOdey:
### Summary

Unable to reach the manager when migrating the database of a production environement to a local developement environement.  It seems I uninstalled some package in my developpement environement that my production database is still tracking thus trying to invoke it. I believe that those dashboard widget should be silently failing inorder to avoid cataclysmic issues.

Fatal error: Call to a member function setNamespace() on null in /var/www/core/model/modx/moddashboardwidget.class.php on line 104

### Step to reproduce

Just uninstall some widget on one side of the platform and override your database again. 

### Observed behavior

Gives you a blank page if debug mode turned off. Give you this error if debug mode turned on.

Fatal error: Call to a member function setNamespace() on null in /var/www/core/model/modx/moddashboardwidget.class.php on line 104

### Expected behavior

The error should be handled so no critical php error fires. And possibly log an error in the logs to notify the developper that some package reference is missing.

### Environment
MODX 2.5.2, apache/nginx and Apache/2.4.16 (Ubuntu), mysql  Ver 14.14 Distrib 5.5.46

MarkOdey:
I don't know if that is intended behaviour or not. I am sure one of you guys know what would be the ideal solution. However this did the trick.

    ```
  if($widget == null) {
            return;
        }
    
```

Not optimal but could be a potential fix.

OptimusCrime:
The fatal error should not happen, but I am not sure how you are doing your migration? Exporting/importing a database is not all you have to do. You also need to move the corresponding files.

MarkOdey:
Very true. Might not be very relevant to provide a fix. It would help though figuring what precisly is crashing when the dashboard is loading. I don't know.

Jako:
Which dashboard widget causes the issue and could you maybe post the code of that widget? 

There should be created somewhere a new modDashboardWidget or inherited object, that is not checked after creating.

sottwell:
Could be a snippet-type widget that got its snippet deleted, but the widget is still attached to the dashboard.

MarkOdey:
Hey guys, a little update on my situation. First off, the main problem was related to file permission on some files that couldn't run. No error were apparent at that level. I believe it was my google analytics widget causing it to crash like so. However it was hard to tell which one was the guilty one since I couldn't get the source of the error. Ideally I should have gotten a 500 somewhere to help figure out the issue. Am I right?

[...]

PR #13377, opened 2017-03-06
Make sure dashboard widget successfully loaded

OptimusCrime:
### What does it do?
Adds a security check to make sure we actually have a working widget before we call member functions in it.

### Why is it needed?
Under certain conditions, the widgets fail to load. This will produce an error as the code is today.

### Related issue(s)/PR(s)
#13367


Mark-H:
Thanks @OptimusCrime, good fix.

PR #13378, opened 2017-03-07
Added ability to use bindings in default values

### LINKED ISSUE #3454: Binding as a default value for checkbox tv not working correctly
andybaughman created Redmine issue ID 3454

I have tried using @SELECT, @EVAL and @CHUNK to provide a default value for a checkbox tv, but no values will return. Even something as simple as the following fails: 

@EVAL $test = '23'; return $test;

Using '23' (as a string) for the default value works correctly. I am using @CHUNK to provide the input option values, which works correctly. Can you replicate this?

rtripault:
### What does it do?

Use the processed default value for TVs being rendered in the manager

### Why is it needed?

It allows using bindings to generate a TV default value.

### Related issue(s)/PR(s)

Should closes #3454

### DIFF
--- a/manager/controllers/default/resource/resource.class.php
+++ b/manager/controllers/default/resource/resource.class.php
@@ -313,10 +313,7 @@ abstract class ResourceManagerController extends modManagerController {
                             $tv->set('inherited',true);
                         }
                         if ($tv->get('value') === null) {
-                            $v = $tv->get('default_text');
-                            if ($tv->get('type') == 'checkbox' && $tv->get('value') == '') {
-                                $v = '';
-                            }
+                            $v = $default;
                             $tv->set('value',$v);
                         }
                     }


sdrenth:
I've changed the edited code in MODX version 2.5.6 and noticed that if I return the default option of 0 or empty, it will simply return the '@EVAL'.

So if i set the default value to 0 inside the snippet, it will return the following as the default option:
`@EVAL return $modx->runSnippet('mySnippetName');`

If I set the value to anything else then empty or 0 it is working fine!

OptimusCrime:
@sdrenth This should be reported in a new issue. Stuff like this goes "missing" way too fast if it is not organized properly. Care to create the issue?

rtripault:
When opening the issue, may i suggest you to post the content of `mySnippetName` so it could be easily testable in the same condition you did ?

Many thanks

sdrenth:
@OptimusCrime @rtripault I've created a new Issue which can be found here:
[13548](https://github.com/modxcms/revolution/issues/13548)

PR #13379, opened 2017-03-07
Allow "log_target" setting to be empty and default to "FILE"

### LINKED ISSUE #7659: log_target setting shows default is FILE
smashingred created Redmine issue ID 7659

In System Settings the log_target setting is set to FILE on a new install and in the description shows that it should default to FILE if unset. However if unset, any errors that would otherwise be written to a file seem to break the ajax in the manager. This was demonstrated by Articles having path properties that caused errors as Package Management couldn't locate the paths. This is incorrect but and expected resulting error. These errors were written to the Error Log file when the log_target was set to FILE just fine. If unset or set to ECHO it seems to cause an uncaught exception error in Ext.

rtripault:
### What does it do?

Checks for empty value when trying to get `log_target` setting.

### Why is it needed?

It allows the setting value to be empty and still default to `FILE` as stated in the setting description (xPDO default being `ECHO`), preventing to crash the connector responses.

### Related issue(s)/PR(s)

Should closes #7659

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -2363,7 +2363,7 @@ class modX extends xPDO {
         }
         if ($initialized) {
             $this->setLogLevel($this->getOption('log_level', $options, xPDO::LOG_LEVEL_ERROR));
-            $this->setLogTarget($this->getOption('log_target', $options, 'FILE'));
+            $this->setLogTarget($this->getOption('log_target', $options, 'FILE', true));
             $debug = $this->getOption('debug');
             if (!is_null($debug) && $debug !== '') {
                 $this->setDebug($debug);


Mark-H:
@rtripault Do you have any way of reproducing/test this? The issue mentions an old articles issue, since resolved, and while I'd be happy to take this at face value (it's a pretty harmless change), I do enjoy making sure a fix does what it says it does.

rtripault:
If i recall correctly, just set `log_target` setting to an empty value.

Without this patch, it should "echo" the logs, with this patch logs should end up being written in the "file"

Mark-H:
Thanks @rtripault, was easy to test with a snippet to just log something random :)

opengeek:
FYI, I cherry-picked this to 2.5.x for inclusion in 2.5.8.

PR #13380, opened 2017-03-08
gives margin to the resource's form title

goldsky:
it overlaps with buttons

### What does it do?

Fixing this.
screen shot 2017-03-08 at 7 44 19 am

To this.
screen shot 2017-03-08 at 7 45 59 am

### Why is it needed?
To avoid curses from our clients.

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -403,6 +403,7 @@ a.x-grid-link:hover, a.x-grid-link:focus{ text-decoration: underline; }
 /* panel stylings */
 .modx-page-header, .modx-page-header div {
   background-color: transparent !important;
+  margin-top: 60px;
 }
 
 #modx-content form.x-panel-body {


OptimusCrime:
Could you add a print screen of the entire manager interface? I'd like to see just how much this shift actually affects the rest of the content. Does the area above the title look slightly "empty"?

FerX:
@goldsky what is your screen's resolution?
you can not force a top margin for all resolutions.

OptimusCrime:
@FerX The title can be long. Even on 1440p screens the title can stretch all the way to the control buttons.

FerX:
we can evaluate the use of media queries, but #modx-action-buttons is fixed, and the length of title and action buttons bar is variable.
A possibile solution is use javascript code.

goldsky:
At least, the minimum space is the fixed height. Then js takes the rest.

FerX:
@goldsky 
is a complex problem. This is my screenshot. I do not have your problem.

image

It depends on these factors:

- resolution screen
- modx-leftbar width  (position fixed)
- title dimension 
- action button size

goldsky:
@FerX ,
This is not complex.
Regardless the monitor size or modx-leftbar, I suggest the Title keeps going downward.
not dynamic resizing.

Mark-H:
I'm not in favour of this change. It solves an edge case (long titles on smaller screens overlapping) by changing _everything_. The top margin will look out of place when there is no, or only a short action bar, and if this were merged I'm sure we'd get bug reports of the title being in the wrong place all of a sudden.  

If there's a way to prevent the _overlap_ without moving the title in all scenarios, I would be much more in favour of such a solution. I understand that might be tricky given its absolute positioning though.

Mark-H:
ExtJS does allow buttons to have submenus, so that could be an option to shorten some of the nav... I don't think it'll easily turn into a hamburger/show-hide menu completely tho, but can't say I've tried :P

goldsky:
If that is the solution, then there will be new issue for other packages
that inject button alongside the menu, like GridClassKey, Babel, Lingua,
Collections, etc, while user permission also takes a part on hiding some.
This button group component is subject to modified, while Title is not,
which is most likely only be changed when pagetitle is being typed.
That's why I suggest this change, not the button.

theboxer:
I'm also not in favour of this. It bumps all titles bellow the toolbar, which isn't necessary on the most of pages - extras installer, users, actually, everywhere but resources with collections (or similar) and long page titles.

I'd be for discussing more the hamburger menu, but not here :)

Closing.

PR #13387, opened 2017-03-14
Let's try to get back on tracks with PHP 7

rtripault:
## What does it do

* use PHPunit 5.7 to run tests on PHP 7.0+
* added PHP 7.1 (latest stable) & 7.2 (nightly) to the test "matrix"

## Why is it needed

We love green... and our test suite in not ready for PHPunit 6.0 (https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-6.0.0 & https://github.com/sebastianbergmann/phpunit/wiki/Preparing-for-PHPUnit-6)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,22 @@
 language: php
-script: "phpunit -c ./phpunit.xml"
+script:
+  - phpunit --version
+  - "phpunit -c ./phpunit.xml"
 
 # Set php versions
 php:
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
+  - 7.1
   - 7.0
+  - 5.6
+  - nightly
+  - 5.5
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Nice. Could it be possible to introduce a composer.json file for the 2.x branch too, or is that something we wish to avoid? In the 3.x branch we can just run composer install. I've altered this somewhat in my WIP branch "Sane Test Suit", and the code there looks like this: https://github.com/OptimusCrime/revolution/blob/sane-test-suit/composer.json

I've done this to make it easier for people to run the tests before they push their branch and/or open a PR.

rtripault:
I would love to!

opengeek:
Merged into 2.5.x via 56e44fe.

PR #13388, opened 2017-03-15
Remove cached session variables when updating usergroups

spectrum48k:
### What does it do?
Removes cached session variables after updating usergroups when updating a user.

### Why is it needed?
Need logout after flushing permissions for changes to take effect when using `$modx->isMember()`

### Related issue(s)/PR(s)
none known


spectrum48k:
This will not be fixed in #13311 because  #13311  fixes flushing for actions:  group update, create/update/remove policy. I'll try another way for fix this.

lawrenz1337:
Hmm, from the looks of it, it is a different problem, in my commit #13311  I've used `sesstion_stale` when updating policies. The OP seems to have problems when changing group of a user. I'll investigate it later but right now it seems to work this way:
1. user has group `Group1`
2. I update his group to `Group2`
3. set `session_stale`
4. Call `$user->isMember()`
Expecting to get `Group2` and getting `Group1`.

Issue #13199, opened 2016-11-27
Collect deprecated methods/classes to be removed in 3.0


Jako:
### Summary
Collect deprecated methods/classes to be removed in 3.0 in the MODX core code and add logging code in those methods/classes.

This should be done before the release of 2.6 to have some warning time for extra authors.

This issue could be used to collect those methods/classes. Please add the methods/classes with the filename, classname and method name in a new comment.

Filename|Classname|Method
---------|-----------|--------
~core/model/modx/rest/modrestclient.class.php~|modRestClient|-
~core/model/modx/rest/modrestcurlclient.class.php~|modRestCurlClient|-
core/model/modx/rest/modrestserver.class.php|modRestServer|-
core/model/modx/rest/modrestsockclient.class.php|modRestSockClient|-
core/xpdo/xpdo.class.php|xPDO|toJson
core/xpdo/xpdo.class.php|xPDO|fromJson

modRestClient is used in core for Package Management. So the core methods have to be changed to use modRestService or the deprecated notice has to be removed.

Unsure:

Filename|Classname|Method|Reason
---------|-----------|--------|-------
core/model/modx/rest/modrest.class.php|modRest|-|Is it replaced by modRestService?

Update (19/10/2021): I've gone through the topic and marked all things that have been addressed as ~crossed off~. If not crossed off, that means it may require further discussion. ^Mark

Mark-H:
I didn't realise the modRest utils were going to be removed? Got a source for that, or are you just proposing we do?

Jako:
All the classes mentioned above have a line like this:
https://github.com/modxcms/revolution/blob/2.x/core/model/modx/rest/modrestclient.class.php#L11

modRestService, modRestConrroller etc. are not touched by this.

Mark-H:
Ah.. don't think I noticed that before :D Makes me wonder what other code we have that is marked as deprecated in that way from years ago,  and how we feel about those things today. Some deprecations will be logical as things are refactored, but there might also be things that we want to take out just because it's not something MODX should provide.

Jako:
The classes of the first table seemed to be improved a lot in modRestService, so they could be deprecated. But I have no code using these classes, so I am maybe wrong.

Jako:
~modRestClient is used in core.~ ^Mark: Not anymore!

Jako:
[...]

PR #13389, opened 2017-03-20
Proper use of json_encode and error handling for outputArray() in processors #modxsnowup

christianseel:
### What does it do?
This PR replaces the `$modx->toJSON()` and the hardcoded string in `modProcessor::outputArray()` with a full `json_encode()` call. It also adds a log message to the MODX error log if `json_encode()` fails and makes sure the processor still returns valid JSON.

### Why is it needed?
If the `$array` was for any reason malformed (in our case it contained illegal UTF-8 characters), the `$modx->toJSON()` returns `false` instead of valid JSON. Due to the fact that the rest of the JSON response was hardcoded as a string the processor returned invalid JSON like
```
{"success":true,"total":"3","results":}
```
(notice the missing data after `"results":`)

### Related issue(s)/PR(s)
(none known)

PR powered by the #modxsnowup

### DIFF
--- a/core/model/modx/modprocessor.class.php
+++ b/core/model/modx/modprocessor.class.php
@@ -247,7 +247,16 @@ abstract class modProcessor {
      */
     public function outputArray(array $array,$count = false) {
         if ($count === false) { $count = count($array); }
-        return '{"success":true,"total":"'.$count.'","results":'.$this->modx->toJSON($array).'}';
+        $output = json_encode(array(
+            'success' => true,
+            'total' => $count,
+            'results' => $array
+        ));
+        if ($output === false) {
+            $this->modx->log(modX::LOG_LEVEL_ERROR, 'Processor failed creating output array due to JSON error '.json_last_error());
+            return json_encode(array('success' => false));
+        }
+        return $output;
     }
 
     /**


christianseel:
Travis probably failed not because of the changes in this PR:
```
PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/modxcms/revolution/_build/test/MODxTestCase.php on line 29
```

rtripault:
Indeed, unrelated and apparently safe to ignore

OptimusCrime:
This +100. So much better doing it this way. Note that this breaks compatibility, so it might need to go in the 3.x branch? Perhaps the 2.x could work too? No matter which version it is introduced it, it the PHP JSON extension will need to be a server requirement from the version and up.

christianseel:
@OptimusCrime json_encode is already widely used in the MODX core: https://github.com/modxcms/revolution/search?utf8=%E2%9C%93&q=json_encode
So I'd say it doesn't add breaking changes and it can be considered as a bug fix.

OptimusCrime:
I see, my bad @christianseel :)

I see that the `toJSON` method has been changed from xpdo [2.x](https://github.com/modxcms/xpdo/blob/67cce90b14b3d2a1ed10412838b3fd970249da52/xpdo/xpdo.class.php#L2405) to [3.x](https://github.com/modxcms/xpdo/blob/6473819b17b83d87f84904f510cccfa7179908ff/src/xPDO/xPDO.php#L2387).

I also see the JSON listed in the [server requirements under extensions](https://docs.modx.com/revolution/2.x/getting-started/server-requirements), which makes me wonder why the core has used `toJSON` to tasks that could just be `json_encode`/`json_decode` calls from the beginning of, as this method more or less serves not purpose other than attempting to use a user library in the 2.x version if the native json methods are missing.

Mark-H:
The toJSON/fromJSON methods date back from a time where json_encode/json_decode support was not as universal as it is today. No longer relevant, and deprecated as listed in #13199.

OptimusCrime:
Should it be deprecated and removed in 3.x then? This PR targets 2.5.x. I'm all in favor of removing this unnecessary library, I am just trying to understand how you are handling this + xPDO and semver.

opengeek:
Yes, toJSON and fromJSON should be deprecated and removed from xPDO in 3.x. We still need to target 2.6 (i.e. 2.x branch) with this change.

opengeek:
This was merged into 2.5.x and cherry-picked into 2.x — modify the base branch before merging so these PRs will automatically close.

Issue #11153, opened 2014-02-26
New Feature: Exclude resource alias from child URIs


bertoost:
This can be helpful when you have a container just for site-working purposes, like grabbing it's child resources. For example:

Footer pages
- Sitemap
- Privacy statement
- Disclaimer
- Conditions

In this case you get an automated alias like "footer-pages/sitemap.html" or "footer-pages/disclaimer.html".
This is unwanted in this case, because you don't want "footer-pages/" in your url.
Of course you can use "freeze uri" function on every child, but when you a have a container with a lot of children pages and a lot of children-children pages that's a hell of a job. In that cases it's easier to turn a parent to be excluded in child paths.

I already have this working; I tried to send a PR but cannot get only those commits to be a PR. Please look at these two commits...

https://github.com/bertoost/revolution/commit/4323cfdbdb9afca3ff23d30faf0492e524b57faa
https://github.com/bertoost/revolution/commit/1a9bd36035b6bba28ff50a2692060b8567949fd3

OR to move the field below alias (and some changes to the texts)

https://github.com/bertoost/revolution/commit/07cd3a4b8045ba6868d455c6fa7965a1f5690576

degoya:
is this features still not implemented in 2.3.3?

vierkantemeter:
The following hack will work in both 2.2 and 2.3 to hide a resource in a URL:
https://github.com/Qwarble/revolution/commit/df1902d5b68e02d760842c19f3a6e647f38ff6ab

osabate:
That's great. Are you gonna make a PR for it?

vierkantemeter:
That's up to @Qwarble

Qwarble:
I did PR, and it was closed.
https://github.com/modxcms/revolution/pull/11167

dssaf:
2.3.3
An error occurred while trying to save the resource.

gadgetto:
2.4 an still not available  :-(

jonleverrier:
:-1:

sbrin:
When this feature will be in released?

Jako:
Since @bertoost is not working on this anymore, someone else has to check this PR on 2.5.1+

rainbowtiger:
I just converted a client from Evolution, which has this feature, to Revolution, which doesn't. Why was this left out of Revolution?? The site needs to keep the same URLs for tracking purposes, and now I have no way of making that happen, short of hand-entering Freeze-uri for every single page that needs this.

[...]

PR #13231, opened 2017-01-03
[WIP]Sane test suit

OptimusCrime:
### What does it do?
This is a rewrite of the current MODX test suit. The goal is to:

- [x] Make the tests autoloaded following the approach from Slim
- [ ] Simplify tests
- [ ] Delete empty tests
- [x] Simplify running tests
- [x] Make travis test for both MySQL and SQLite (MSSQL may be difficult), check xpdo
- [ ] Introduce tests that actually runs the setup, to make sure it is not broken
- [ ] Clean code in MODxTestHarness.php

#### Tests

- [x] Tests/
- [x] Tests/Cases/Requests
- [x] Tests/Controllers/
- [x] Tests/Controllers/Context
- [ ] Tests/Model/
- [ ] Tests/Model/Dashboard
- [ ] Tests/Model/Element
- [ ] Tests/Model/Element_todo
- [ ] Tests/Model/Error
- [ ] Tests/Model/Filters
- [ ] Tests/Model/FormCustomization
- [ ] Tests/Model/Hashing
- [ ] Tests/Model/Lexicon
- [ ] Tests/Model/Mail
- [ ] Tests/Model/REST
- [ ] Tests/Model/Registry
- [ ] Tests/Model/Request
- [ ] Tests/Model/Resource
- [ ] Tests/Model/Security
- [ ] Tests/Model/Sources
- [ ] Tests/Model/Transport
- [ ] Tests/Model/Validation
- [ ] Tests/Processors/
- [ ] Tests/Processors/Browser
- [ ] Tests/Processors/Context
- [ ] Tests/Processors/Element
- [ ] Tests/Processors/Resource
- [ ] Tests/Processors/System

### Why is it needed?
The current test suit has several problems:

- The harness does a lot of magic, difficult to understand just what it does
- Tests are used with complicated and unnecessary providers that just clutter the code
- There are many empty test cases that should be removed
- Several tests should be adjusted

### Motivation
Having a good test suit for MODX has not been a priority and the current test coverage is very low. As we begin to look towards 3.x, we should try to update our test suit to simplify testing bigger changes. This PR is an attempt to start this.

### Related issue(s)/PR(s)
Needs PR #13191 merged in order to fix a few skipped tests.

PR #13395, opened 2017-03-23
[WIP]Exclude resource alias for child

OptimusCrime:
### What does it do?
Implement functionality to hide a alias in the path of children resources.

Includes:

- This PR makes sure to add the new field in the map files of both MSSQL and MySQL. I've also added scripts for each of them to be added during an upgrade. This is test and works as expected.
- Adds a new checkbox under Settings in the box at the right side, below Searchable and above Freeze URI.
- Adds a simple lexicon entry with a wording similar to that in Evo. Appears when hoving the checkbox like the other settings in this area.

It works like this. Say we have the current relationship of resources Foo -> Bar, which has aliases `foo` and Bar which has its alias to `bar`. The default value of `alias_visible` is 1, so it does not change the current behavior of any sites without alterations. If both Foo and Bar has this setting to 1 (as is the default), it would produce the URI as before `foo/` and `foo/bar`. Below are the other configurations and their URI values:

Foo `alias_visible` = 0. Bar `alias_visible` = 1
Foo URI = `foo/`. Bar URI = `bar`.

Foo `alias_visible` = 1. Bar `alias_visible` = 0
Foo URI = `foo/`. Bar URI = `foo/bar`.

Foo `alias_visible` = 0. Bar `alias_visible` = 0
Foo URI = `foo/`. Bar URI = `bar`.

As we see, the URIs ignore the PARENT alias if `alias_visible` is 0, but will always use its own alias regardless of the `alias_visible` value. This is because we always require an alias for any resource.

*Note: This PR is currently missing tests. I'd love to implement some tests to make sure this works like it should.*

### Why is it needed?
Could be useful in certain situations. See issue referenced below. Seem like a common request. The feature was present in Evo, but for some reason removed since then. I see no reason this should not be included in the core. The functionality to go around this via an extra/plugin would be rather painful to implement.

### Related issue(s)/PR(s)
#11153


Pathologic:
Thank you, I hope this try will be finished successfully.

joeke:
@OptimusCrime 
Looking good, but I do have 2 comments/questions:
1. I think the field also has to be added to the modx.mysql.schema.xml (and sqlsrv.schema)
2. Does this work in combination with the `use_frozen_parent_uris` system setting? So let's say I have a 3 levels of resources:
+ Foo (alias_visible = 1)
++ Bar (alias_visible = 0, uri_override = 1)
+++Baz (alias_visible = 1)
In this case how will the uri of Baz look like, assuming you have `use_frozen_parent_uris` set to Yes?

OptimusCrime:
@joeke Thank you for your input :) I'll add the fields to the schemas. 

If you look at the code for the implementation, it is [rather simple](https://github.com/OptimusCrime/revolution/blob/243e4ae5f929b65809f9f6490d399648cb86f00e/core/model/modx/modresource.class.php#L923-L927). No matter what the alias for a parent resource is, or how it was computed, it will not be added to the path if `alias_visible` is 0. 

So, based on this I'd guess the calculated URI for Baz would be `/foo/baz`. Do you think it should be different?

pixelchutes:
I think I have the same question as @joeke -- how are conflicts handled when a parent container's alias is omitted from the path, and you are facing a frozen or pre-existing URI conflict? 

From your 3rd example above (both `alias_visible` = 0), what if you already had a Resource with `alias`=`bar` one level up, resulting in same URI? Likewise, what if some other Resource (regardless of location in hierarchy) had activated `uri_override` resulting in similar conflict?

Not opposed to the idea or progress at all, just trying to better understand!

OptimusCrime:
This PR only alters the generation of the alias path. This means that other checks, such as duplicates, will function just as they did previously. I have no tested this, but it *should*.

Therefore @pixelchutes , both your questions would, as far as I can understand, result in duplicated URIs. This is simply because the values for the associated resource in the `site_content` table, after they are constructed, will be the same.

I am not sure if this is unwanted or not. The idea of this is really just to provide (yet) another way of handing aliases to best suit your needs. It will "conflict" with existing solutions such as `uri_override` and `use_forzen_parent_uris`, but as far as I can understand, those two options can already conflict among themselves.

Any input on alternative approaches to this is much appreciated. I did not browse the Evolution source code to see how this feature was implemented there, so I dont know if they have some of the same issues there.

pixelchutes:
If that's how it will work, that's awesome! My line of thinking was if an _established_ site hierarchy "flipped the switch" at a certain level, you could have a potential for conflicts from a totally different corner of your site. At that point, I'm not sure to what degree the potential `uri` conflict check handles, but at a minimum a Refresh URI and/or Clear Cache would report any dupes in the console like always.

OptimusCrime:
Great :)

As far as I can tell, by looking at the method(s) that handles aliases, the Resource create/update form should highlight and report a duplicate URI for this resource, just like any other duplicated URI. That part of the code has not been altered, and those methods does not care how the aliases were produced, only if they are duplicated of each other or not.

joeke:
So as far as I can see it won't add any new conflicts with existing methods.
But I do think a helpful addition is that the error message for duplicate uri's should be optimised. Say for instance you have parent Foo with child Bar, where Foo has alias_visible = 0, and you also have Bar in the root level, then saving this one has conflicting uri with Foo > Bar, but it's not directly apparent that it has to do with the alias_visible = 0 setting on Foo. So a clear help/error message would be very helpful in this case.

Mark-H:
Is this still WIP or ready for review?

OptimusCrime:
I'd still like to add some tests, but the functionality is OK. There's no need to do anything else with the code. Could it be merged without the tests and I can add them to my other PR #13231 ?

opengeek:
This is breaking existing tests. If those can be fixed we could review this for merge.

OptimusCrime:
@opengeek The tests failed because the version of PhpUnit we used were incompatible with PHP7. It was fixed recently. I will rebase and push. Just a FYI on why the tests are red.

OptimusCrime:
@opengeek Rebased and pushed. Tests should be green now.

setup/includes/upgrades/mysql/2.6.0-pl.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/*1
```
Alroniks: `1` here is obsolete

OptimusCrime:
Closing this as #13660 superseed my PR with additional fixes and follow up that I was unable to do.

PR #13400, opened 2017-03-28
new function ChangeField into the manager to change a column without …

flolopdel:
…drop the existing content, you must update the schema xml changing the field before call this function

### What does it do?
Change a existing column without drop the existing content, you must update the schema xml changing the field before call this function

### Why is it needed?
To change a column field without drop the old values

### DIFF
--- a/core/xpdo/om/mysql/xpdomanager.class.php
+++ b/core/xpdo/om/mysql/xpdomanager.class.php
@@ -375,6 +375,39 @@ class xPDOManager_mysql extends xPDOManager {
         return $result;
     }
 
+    public function changeField($class, $oldName, $newName, array $options = array()) {
+        $result = false;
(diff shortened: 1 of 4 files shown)


Mark-H:
Hi @flolopdel!

Thanks, this looks really interesting and I've wondered why we didn't have a method like that before...

Unfortunately though this is the wrong place to send that pull request... all changes to xPDO need to go into the main xPDO repo at https://github.com/modxcms/xpdo (probably the 2.x branch - the 3.x branch is newer but wont be incorporated until MODX 3). 

The xPDO repo gets pulled into MODX at a later point. 

Maybe worth changing the method name to `renameField`? There's already an `alterField` which can be used to change the position of the field on the table, so rename might be a bit more descriptive in what this does to avoid confusion.

Jako:
The PR was opened on [xPDO](https://github.com/modxcms/xpdo/pull/114) now. So I close it here.

PR #13405, opened 2017-03-30
Make search bar work as expected on Chrome & Firefox

### LINKED ISSUE #13404: Unable to click items in search uber bar
### Summary
In MODX 2.5.6, when I search for something I'm unable to click and goto the entry.

### Step to reproduce
Search for something in the uber bar

### Observed behavior

image

### Expected behavior
That when I click, I can edit the resource or element

### Environment
MODX 2.5.6 , Apache, PHP7 , racksapce / liquidweb cloud sites

rtripault:
### What does it do?

Get the DOM event object the way it should be

### Why is it needed?

Fix the regression introduced in #13364

### Related issue(s)/PR(s)

Closes #13404

### DIFF
--- a/manager/assets/modext/widgets/core/modx.searchbar.js
+++ b/manager/assets/modext/widgets/core/modx.searchbar.js
@@ -138,94 +138,11 @@ Ext.extend(MODx.SearchBar, Ext.form.ComboBox, {
             ,scope: this
             ,stopEvent: false
         });
-
-        // Ext.get(document).on('keydown', function(vent) {
-        //    console.log(vent.keyCode);
-        // });
     }
 
     /**
      * Override to support opening results in new window/tab
      */
-    ,initEvents : function(){
-        Ext.form.ComboBox.superclass.initEvents.call(this);
-
-        this.keyNav = new Ext.KeyNav(this.el, {
-            "up" : function(e){
-                this.inKeyMode = true;
-                this.selectPrev();
-            },
-
-            "down" : function(e){
-                if(!this.isExpanded()){
-                    this.onTriggerClick();
-                }else{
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -351,13 +264,13 @@ Ext.extend(MODx.SearchBar, Ext.form.ComboBox, {
...
+    ,toggle: function(hide) {
         var uberbar = Ext.get( this.container.id );
-        if( uberbar.hasClass('visible') || hide ){
+        if (uberbar.hasClass('visible') || hide ) {
             this.blurBar();
-			uberbar.removeClass('visible');
+            uberbar.removeClass('visible');
         } else {
-			uberbar.addClass('visible');
+            uberbar.addClass('visible');
```
rtripault: Those are just CS changes (sorry, i'm a bit OCD)

Jako:
Did you totally remove the keyboard events?

rtripault:
No. Sorry, i might have been quick in the technical explanation...

My understanding is that the `initEvents` has been overridden to pass the even object to `onSelect` method.
`onSelect` was using `window.event`, which is not available on Firefox.

So, as a fix proposal, i removed the override, and made use of `Ext.EventObject` instead, which is a "normalized" way to get the DOM event object, no matter the browser!

Hope that makes sense.

Jako:
Ok. I am fine with that. The previous patch was not deep enough then.

pixelchutes:
Nice work @rtripault! Tested this out successfully, fixing both the regression and the original issue. LGTM!

matdave:
Thanks this was driving me crazy.

Issue #13408, opened 2017-04-03
Forgot Password in Manager allows User/Email enumeration.


Omeryl:
### Summary
The forgotten password utility on the manager allows user and email enumeration because it will announce if a user doesn't exist.

### Step to reproduce
1. Head over to /manager/
2. Try any user or email in the forgot password fields. 

### Observed behavior
If it exists it'll say it succeeds. If it fails, it'll say it failed.

### Expected behavior
It should state something like "If the user or email exists, you'll receive an email shortly." instead of disclosing that a user/email did/didn't exist. 

### Environment
Any and all environments.

pixelchutes:
Great point, @Omeryl I think this is a good recommendation for an improved, default behavior.

That said, since these error messages are managed at the Lexicon level, couldn't one choose to update the two individual lexicon keys to serve the same value, vs. differentiating (read: disclosing) as it does today? This currently gives users the control to differentiate if they so choose.

**BEFORE:**
> `login_user_err_nf_email` = User not found with that email or username. Please try again.
> `login_password_reset_act_sent` = Password Reset Activation email sent. Please check your inbox.


**AFTER:**
> `login_user_err_nf_email` = If the user or email exists, you'll receive an email shortly.
> `login_password_reset_act_sent` = If the user or email exists, you'll receive an email shortly.


![same_lexicon](https://cloud.githubusercontent.com/assets/352182/24628612/4dd69666-1874-11e7-901d-b31b2e656fc4.jpg)


![example_ambiguous](https://cloud.githubusercontent.com/assets/352182/24628574/27ef25da-1874-11e7-844b-94ef989c5a1a.jpg)

Omeryl:
@pixelchutes I believe, in this case, the default should be oblivious to the public entirely, an opt-in to insecurity vs an opt-out.

pixelchutes:
Exactly! 👍 Makes sense to me. I think easily accomplished w/ a lexicon change vs. code change, keep flexible if (for some reason) the alternative may be preferred/required.

OptimusCrime:
I think this is fixed with the merged PRs above. Feel free to correct me.

@modxbot close

PR #13409, opened 2017-04-03
Update login.inc.php #13408

mrhaw:
Fix User/Email enumeration issue #13408
Credit @pixelchutes

### What does it do?
Disable User/E-mail Enumeration

### Why is it needed?
Increases Information Security/Privacy. Troy Hunt has an awesome blog about it
» https://www.troyhunt.com/website-enumeration-insanity-how-our-personal-data-is-leaked/

### Related issue(s)/PR(s)
#13408 
Credit @pixelchutes


Omeryl:
This doesn't "fix" the issue, per se, because of crowdin.

Mark-H:
It would fix it in english, and mark other languages as needing an updated translation.

Omeryl:
I feel like the better solution is to add a new lexicon, using that, instead of waiting possibly months for the translations to catch up over multiple releases.

mrhaw:
A new lexicon? One could force update all languages files to use the english title until it has been translated...

mrhaw:
What is the verdict? Should we update the english file as recommended by Mark or 
2. also force update all other language files for these phrases
3. do nothing at this time

Omeryl:
I think, honestly, we should be adding a new phrase and deprecating the old lexicon, myself, @mrhaw

Mark-H:
I can be wrong about this, but when the source language (English) is updated, i believe that resets the language string. In the crowdin UI it will show up as untranslated, and i _think_ in the export it is also set to the English text until it is translated again. So 1 and 2 would be the same.

Omeryl:
Regardless, it seems moot having two lexicons with the same exact string, so why don't we just do it the "right way." (subjective I know) and get rid of deadweight.

pixelchutes:
I agree, introducing new would be the better approach. The naming convention `err_nf` would no longer be accurate if we simply repurposed the current lexicons vs. introducing new. Slight flexibility is lost, but for good reason.

mrhaw:
This commit is not introducing a new lexicon but takes care of User/E-mail Enumeration within existing code... If we want a new lexicon (I don't know what is involved) what does that have to do with enable a fix for existing code. I believe this is taking MODX to a higher level and it's a weakness that people can unmask  who runs a MODX website.

Omeryl:
@mrhaw I'm very aware of the issue, given I reported it to begin with. I have another one I've yet to report that would require a code change anyway, so I'm curious if the two should just go in one PR instead.

mrhaw:
I'm closing this.

Mark-H:
Could you clarify why @mrhaw? To me this seems like a good step but I have no idea what just happened to this thread. Is someone working on a different patch?

mrhaw:
@Mark-H 
I don't understand why this pull request didn't got merged and any discussion of a future lexicon or another pull request was taken place in the issue thread #13408

All it does is changing two text strings that immediately takes care of the issue. I don't understand the opposition and it quite honestly make me feel I waste time here.

Mark-H:
I'm sorry you were made to feel that way. Personally I consider lots of discussion on an issue or pull request a good thing, but sometimes it leads to "bikeshedding" over minor things. I've not merged this myself (despite involvement in the discussion) because there did not appear to be consensus on the topic with @Omeryl and @pixelchutes suggesting an alternative. 

Also if you look at the list of open pull requests you'll notice that as integrators we're a little behind, so please don't take a lack of merge as a personal or rejection of your contribution. We're aware of that. 

Let's keep this pull request open unless/until @Omeryl or @pixelchutes provide an alternative PR that everyone agrees on. If neither of them do, this PR should resolve the problem nicely.

mrhaw:
Thanks Mark! Discussion is awesome. I saw the discussion in the issue thread and were afraid it would just stay a discussion. The consensus of @Omeryl and @pixelchutes seem to be that let's not do anything as there might be use cases supporting enumeration...

> This currently gives users the control to differentiate if they so choose.

I googled Troy's article above to make the case we need this change!

Omeryl:
I think for the time being we can merge this in to get it into the next release. It'll cause a reset of translations as we discussed, and that should be sufficient while we decide if we need to add a new lexicon. 

Does the Login extra use these existing lexicons? That might be a strong support for putting the new lexicon in the 2.6 release instead so we don't break the extras lexicons, further supporting this quick fix.

pixelchutes:
I'm all also in favor of merging this PR 👍, as I do think it leaves the product better off as a result. Definitely appreciate the PR @mrhaw. I see no reason this cannot be merged ahead of the unreported, related issue @Omeryl [mentioned](https://github.com/modxcms/revolution/pull/13409#issuecomment-293397751). Whenever that happens (2.5.8, 2.6.x, etc), any related PR(s) can easily build upon this, further improving the solution.

Either way, updating lexicons for non-English languages can't be avoided. Therefore, I think ultimately a code change is ideal, since it would prevent duplication within the Lexicon, and also would correctly represent the underlying naming convention.

pixelchutes:
@Omeryl Looks like a custom lexicon: [login.user_err_nf_email](
https://github.com/modxcms/Login/blob/8f6043404dcad627cc229422ff768a89f1091c0d/core/components/login/lexicon/en/forgotpassword.inc.php#L35) is [used](https://github.com/modxcms/Login/blob/8f6043404dcad627cc229422ff768a89f1091c0d/core/components/login/controllers/web/ForgotPassword.php#L99) in Login's **forgotpassword.inc.php** lexicon file.

> `login.user_err_nf_email` = A user was not found with that email.

mrhaw:
https://github.com/modxcms/Login/pull/91

opengeek:
So, what should we do here? We have synced this change to Crowdin, but user/email enumeration is still possible in all languages except English.

PR #13418, opened 2017-04-13
[SECURITY-18] Fix stored XSS in user group name, and other potential manager XSS issues

Mark-H:
Initial XSS report affecting the user group name and various places it gets rendered was from Anti Räis via security@modx.com, ticket 18 received April 3rd.

During the investigation of that report I found that ExtJS components that define custom tpls need the htmlEncode filter on possibly untrusted content. So along with fixing the reported issues, I've done a quick search for similar issues and also patched the potential issues I could find.

Note: I've not yet tested every single usage of the other (non user group) extjs components, so I would appreciate some help testing to make sure that this doesn't inadvertently breaks something.


core/model/modx/processors/security/group/getnodes.class.php
```diff
@@ -131,7 +131,7 @@ public function prepareGroup(modUserGroup $group) {
         $c->limit(1);
         $count = $this->modx->getCount('modUserGroup', $c);
         return array(
-            'text' => $group->get('name').' ('.$group->get('id').')',
+            'text' => htmlentities($group->get('name'), ENT_QUOTES, 'UTF-8') . ' ('.$group->get('id') . ')',
```
pixelchutes: I'm not sure if this matters, but should UTF-8 be hard-coded here, or is it preferable to rely on the defined `modx_charset`?

core/model/modx/processors/security/user/get.class.php
```diff
@@ -53,7 +53,7 @@ public function getGroups() {
             if ($member->get('role') == 0) { $roleName = $this->modx->lexicon('none'); }
             $data[] = array(
                 $member->get('user_group'),
-                $member->get('user_group_name'),
+                htmlentities($member->get('user_group_name'), ENT_QUOTES, 'UTF-8'),
```
pixelchutes: See comment above ☝️

core/model/modx/processors/security/group/getnodes.class.php
Mark-H: That's a fair point. In most of my projects I tend to just hardcode UTF-8 because it takes me a lot of effort to come up with a reason why it wouldn't be UTF-8, but I'd be happy to make it read from the setting as well. I guess that kinda is the way it's done in the rest in MODX, so should be done here too.

Mark-H:
I've updated the PR to use the modx_charset setting.

PR #13423, opened 2017-04-17
[SECURITY-19] Remove htaccess from allowed file types on new installations

Mark-H:
Reported to security@modx.com by Anti Räis in ticket 19 and Tomáš Melicher in ticket 20, the ability to upload or create .htaccess files can cause code execution. In a way this is a feature, but it being enabled by default can pose a risk to users who are unaware that is possible. 

Similar to how php files are not allowed out of the box, this patch will also prevent htaccess files by default. Users that want to manage htaccess from the manager can still do so by updating the upload_files setting after installation.


rthrash:

https://community.modx.com/t/htaccess-has-a-lock-icon-on-it/4299/5

PR #13428, opened 2017-04-19
[SECURITY-20] Improve local file inclusion protections

opengeek:
### What does it do?
This commit improves the LFI protections throughout the core to remove any sequence of 2 or more `.` characters regardless of the path delimiter that precedes or follows it.

### Why is it needed?
The existing protections would not work on Windows platforms with backslash path delimiters.

### Related issue(s)/PR(s)
N/A


AgelxNash:
💩
```php
$name = './index.php';
$directory = 'public/files/.';
```
 /var/www/example.com/public/files/../index.php

Mark-H:
@AgelxNash: security@modx.com please

PR #13430, opened 2017-04-20
Do not expose full path in error message

islandsins:
### What does it do?
This commit removes the full path from an on-page error message in Manager.

### Why is it needed?
If you modify the url parameters when in Manager, you can currently produce an error message that exposes the full path of the application (eg 'Could not find action file at: /home/webuser/public_html/manager/controllers/default/system/settings0.php').  Whilst this is low risk as it only applies to already authenticated users, it would be preferable if the full path was not shown on the page, but only in the error log which is only visible to users with appropriate permissions.  

### Related issue(s)/PR(s)
None

### DIFF
--- a/core/model/modx/modmanagercontrollerdeprecated.class.php
+++ b/core/model/modx/modmanagercontrollerdeprecated.class.php
@@ -55,7 +55,7 @@ class modManagerControllerDeprecated extends modManagerController {
 
             $cbody = include $f;
         } else {
-            $cbody = 'Could not find action file at: '.$f;
+            $cbody = 'Could not find action file';
         }
 
         if (!empty($this->ruleOutput)) {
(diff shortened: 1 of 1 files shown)


pixelchutes:
When developing a custom component, this error message can be useful in helping determine a potentially missing file, leaving you a clue on where to look or what to change to help resolve.

I can appreciate the reasoning behind the proposed fix, though I wonder if a better approach may be to still log the full path in the error log, or perhaps adjust the path that is output to be relative to MODX root?

Removing entirely would omit information that could otherwise be helpful during development.

Thoughts?

islandsins:
In my tests the full path is still logged in the error log, as it is logged from the getControllerFilename function, not sure if I am missing any situation where it might not be.
Making the displayed path relative is probably ok, although personally I prefer keeping the technical details of any error in the log only.

Jako:
Maybe the controller path could be replaced by the controller url in that message …

Jako:
I have added a commit that removes only the namespace path in the output. This should be safe enough.

PR #13433, opened 2017-04-21
Remove unused path_search and url_search processors

opengeek:
### What does it do?
Removes unused processors.

### Why is it needed?
They can potentially leak system directory information.

### Related issue(s)/PR(s)
N/A


fgeek:
Have you scheduled 2.5.8 release already?

Issue #13436, opened 2017-04-23
Uberbar Not accessible via VoiceOver Rotor (Landmarks)


jpdevries:
### Summary
The uber bar is not accessible via assistive tech landmarks such as the VoiceOver rotor.

![](http://j4p.us/0J1G2p2W3u0O/Screen%20Shot%202017-04-23%20at%203.03.13%20PM.png)

### Step to reproduce
Using VoiceOver, open landmarks. Notice the search component is not present.

### Observed behavior
Can't access uberbar via the landmarks rotor.

### Expected behavior
Should be able to access uberbar via the landmarks rotor.

### Environment
MODX 2.x

### Solution
I tried to send a PR for this but the feeling between ExtJS and I is mutually. Here's the markup we have now:

```html 
<li id="modx-manager-search">
  <div class="x-form-field-wrap x-form-field-trigger-wrap" id="ext-gen274" style="width: 207px;">
    <input type="text" size="24" autocomplete="off" id="modx-uberbar" name="modx-uberbar" class="x-form-text x-form-field x-form-empty-field" style="width: 169px;">
    <div class="x-form-trigger x-form-arrow-trigger" id="ext-gen275"></div>
  </div>
</li>
```

Here's a fix:
```html 
<li id="modx-manager-search" role="search">
  <div class="x-form-field-wrap x-form-field-trigger-wrap" id="ext-gen274" style="width: 207px;">
    <input aria-label="Search" type="text" size="24" autocomplete="off" id="modx-uberbar" name="modx-uberbar" class="x-form-text x-form-field x-form-empty-field" style="width: 169px;">
    <button type="submit" aria-label="Go" class="x-form-trigger x-form-arrow-trigger" id="ext-gen275" style="margin:0"></button>
  </div>
</li>
```

So to recap
 - [x] `role="search"` on `#modx-manager-search` (or wrap the input in a actual `<form role="search">`)
 - [x] `aria-label` on the Input (not sure where the Search "label" is coming from now)
 - [x] Change the submit "button" to an actual button
 - [x] Give the submit button `aria-label` and `type="submit"`
 - [ ] Styling to remove `margin` form the submit button

* * * *
- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

rtripault:
Your turn again now :)

Ibochkarev:
Improved in #13436

PR #13437, opened 2017-04-23
Added some a11y recommendation as per #13436

rtripault:
### What does it do?

Added some a11y recommendation as per #13436

### DIFF
--- a/manager/assets/modext/widgets/core/modx.searchbar.js
+++ b/manager/assets/modext/widgets/core/modx.searchbar.js
@@ -11,12 +11,19 @@ MODx.SearchBar = function(config) {
         ,typeAhead: true
         // ,listAlign: [ 'tl-bl?', [0, 0] ] // this is default
         ,listAlign: [ 'tl-bl?', [-12, 0] ] // account for padding + border width of container (added by Ext JS)
-        // ,triggerConfig: { // handled globally for Ext.form.ComboBox via override
-        //     tag: 'span'
-        //     ,cls: 'x-form-trigger icon icon-large icon-search'
-        // }
-        // ,shadow: false // handled globally for Ext.form.ComboBox via override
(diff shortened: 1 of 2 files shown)


manager/templates/default/header.tpl
```diff
@@ -76,12 +76,12 @@
...
             </ul>
             {if $_search}
-            <div id="modx-manager-search"></div>
+            <div id="modx-manager-search" role="search"></div>
             {/if}
         </div>
     </div>
         <div id="modAB"></div>
         <div id="modx-leftbar"></div>
-		<div id="modx-action-buttons-container"></div>
```
rtripault: Those were tabs... TABS!

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -11,12 +11,19 @@ MODx.SearchBar = function(config) {
...
-        // ,triggerConfig: { // handled globally for Ext.form.ComboBox via override
-        //     tag: 'span'
-        //     ,cls: 'x-form-trigger icon icon-large icon-search'
-        // }
-        // ,shadow: false // handled globally for Ext.form.ComboBox via override
-        // ,triggerAction: 'query'
+        ,triggerConfig: {
+            tag: 'button'
+            ,type: "submit"
+            ,"aria-label": "Go"
```
rtripault: Probably needs some lexicon string here instead

Mark-H:
@jpdevries Could you check this pull request to verify it achieves what you suggested in #13436?

jpdevries:
Looks good @rtripault. I think the only thing we could do more is wrap the input and the button in an actual `<form>`. I can't remember, maybe that wasn't feasible. It looks like I didn't mention it in https://github.com/modxcms/revolution/issues/13436

Issue #13073, opened 2016-07-18
PHP Warnings like in #13041


LitvK:
### Summary

Having following errors too:
AH01071: Got error 'PHP message: PHP Warning:  session_start(): Session callback expects true/false return value in /srv/www/core/model/modx/modx.class.php on line 2270\n'
### Step to reproduce

It begins after setting variables in php.ini:
session.gc_probability = 1
session.gc_divisor = 100

In this case warning will arise every 100 pages. Perhaps if you set session.gc_divisor to 1, this warning will fall every time. It makes real troubles on AJAX scripts which returns JSON data. This warning brokes JSON data and make pages unavailable.
### Observed behavior

If you have activated garbage collector for sessions and enabled E_WARNING errorlevel in PHP, it will throw warnings about ModX callback function. The same problem was in roundcube some time ago:
https://github.com/roundcube/roundcubemail/issues/4948.
### Expected behavior

No warnings at all.
### Environment

PHP 7.0.4-7ubuntu2.1 (cli) ( NTS )
Apache/2.4.18 (Ubuntu)
MODX Revolution 2.5.0-pl

meshkov:
I have this warnings too.
`[Fri Mar 10 10:09:57.114320 2017] [:error] [pid 39551] [client 217.69.134.14:41997] PHP Warning:  session_start(): Session callback expects true/false return value in /server/public_html/core/model/modx/modx.class.php on line 2284`
Environment:
MODX Revolution 2.5.5-pl (traditional)
PHP Version 7.0.15
Apache/2.4.7
php.ini:
session.gc_divisor = 100
session.gc_probability = 1

meshkov:
This is hosting settings and I can not change them - session.gc_divisor and session.gc_probability.
In Google there are a lot of sites with this problem.
example:
![image](https://cloud.githubusercontent.com/assets/1703670/23821562/7dc75da4-0647-11e7-98cd-f8207792ed86.png)
My search query in google:
`PHP Warning: session_start(): Session callback expects true/false return value in core/model/modx/modx.class.php on line 22`

Mark-H:
Fixed in 2.6

Issue #13041, opened 2016-05-31
PHP Warnings


meshkov:
### Summary

Having the following errors:
[Tue May 31 15:30:51 2016] [error] [client xxx.xxx.xxx.xxx] PHP Warning:  session_start(): Session callback expects true/false return value in /home/x/xxx/xxx.xx/public_html/core/model/modx/modx.class.php on line 2270, referer: http://xxx.xx/manager/?a=resource/update&id=55
### Step to reproduce

I update the resources and when i try to save resource i have this errors and error in browser ERR_CONNECTION_REFUSED.
### Environment

MODX Revolution 2.5.0-pl (traditional)
PHP Version 7.0.6
Google Chrome 51.0.2704.63 m

OptimusCrime:
This was first reported in #12797.

christianseel:
Are you sure that you're using 2.5.0-pl here? I don't see this error when running MODX 2.5.0-pl with PHP 7.0.6

meshkov:
I use MODX Revolution 2.5.0-pl (traditional).

meshkov:
![2016-06-02_15-11-50](https://cloud.githubusercontent.com/assets/1703670/15744409/7ad9918a-28d4-11e6-9506-8e1f24897916.png)
After that error - i have error in browser ERR_CONNECTION_REFUSED. And have errors in logs.

OptimusCrime:
Looks like this is an error with your server or your setup. I do not have the same problems here.

Realetive:
What about repeat this actions for request this error in other browsers?

meshkov:
In mac os with google chrome i don`t have errors

OptimusCrime:
What browser produces the error?

OptimusCrime:
Closing this due to lack of feedback. It was also reported that the identical version did not produce the same error in another instance. Feel free to reopen this issue if you still experience this problem or if you can give more information about the incident.

@modxbot close

Jako:
This occurs on an 2.5.2 installation with PHP 7.0.14 on the first load of the site (when a new session is initialized). It is just a warning, but it occurs in the frontend output on that installation.

sottwell:
Is this only when logging in, or for first site load with (anonymous) session?

I'm using PHP 7.1.0 with Revo 2.5.2 on a MAMP stack with no problems.

Jako:
I have to ask my client for this. I think he has loaded the page without any login attempt at home.

[...]

PR #13443, opened 2017-04-26
Session callback expects true/false

bezumkin:
### What does it do?
It forces modSessionHandler to return boolean instead of integer.

### Why is it needed?
This will fix PHP7 warnings in logs
```
session_start(): Session callback expects true/false
```

### Related issue(s)/PR(s)
#13073 and #13041

### DIFF
--- a/core/model/modx/modsessionhandler.class.php
+++ b/core/model/modx/modsessionhandler.class.php
@@ -140,8 +140,8 @@ class modSessionHandler {
      */
     public function gc($max) {
         $maxtime= time() - $this->gcMaxLifetime;
-        $result = $this->modx->removeCollection('modSession', array("{$this->modx->escape('access')} < {$maxtime}"));
-        return $result;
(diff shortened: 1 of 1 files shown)


sottwell:
Should it return "true" even if it fails?

bezumkin:
I don`t see how it can fails and why it needed. But we can change it to
```
$result = $this->modx->removeCollection('modSession', array("{$this->modx->escape('access')} < {$maxtime}"));
return $result !== false;
```

sottwell:
Why not just convert the existing return value to a boolean? Something like
```return (bool)$result```

bezumkin:
Because removeCollection can return int(0) and it will be converted to boolean(false).

Ant it will be wrong.

meshkov:
I tested this this pull request on my sites and it solved my problems.

Mark-H:
Thank you @bezumkin, this is now fixed for in 2.6.

PR #13434, opened 2017-04-22
Fixed Fatal error.

mishanthrop:
### What does it do?
Fixes «Fatal error: Call to a member function get() on array in /.../manager/controllers/default/resource/resource.class.php on line 152»

### Why is it needed?
I upgraded from 2.5.4 to 2.5.7 and was getting page with error when I click on a resource to edit it. 

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13085
https://forums.modx.com/thread/100583/upgrade-to-2-5-1-white-screens-when-editing-resources

PR #13444, opened 2017-04-26
fixed fatal error

goldsky:
when opened a resource, manager goes blank, because $this->resource is an array
`
[Wed Apr 26 12:41:37.170177 2017] [:error] [pid 7050] [client 127.0.0.1:38750] PHP Fatal error:  Call to a member function get() on array in /var/www/html/modx-revo-addons-dev/www/manager-dev/controllers/default/resource/resource.class.php on line 152, referer: http://localhost/modx-revo-addons-dev/www/manager-dev/index.php
`

### What does it do?
Conditional checking.

### Why is it needed?
Fixes fatal error.

### Related issue(s)/PR(s)
-

### DIFF
--- a/manager/controllers/default/resource/resource.class.php
+++ b/manager/controllers/default/resource/resource.class.php
@@ -149,7 +149,7 @@ abstract class ResourceManagerController extends modManagerController {
      * @return array|bool|string
      */
     public function firePreRenderEvents() {
-        $resourceId = !empty($this->resource) ? $this->resource->get('id') : (!empty($this->scriptProperties['id']) ? $this->scriptProperties['id'] : 0);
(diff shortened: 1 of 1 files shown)


OptimusCrime:
I'd like some more information on how to reproduce this. I've never had this issue as far as I can remember.

rtripault:
Duplicate of #13434.

To give some background, the `resource` attribute could be an array because of the old TinyMCE component using `$resource->toArray` in its plugin, and we are passing the resource by reference since a few releases (we received quite a few issues on that matter)

Mark-H:
Thanks @goldsky. Included for 2.6.0.

PR #13447, opened 2017-05-02
Update unpack.class.php

SintezCode:
Check for absolute/relative path to build target path.

### What does it do?
Fix unpack error if your mediasource path is absolute


Jako:
Thanks for your addition. I think it should be solved different on base of media source values.

Mark-H:
Hmm, yeah it would be nice if this could be handled through the media sources, tho I'm wondering of unpacking will work with any non-file media source anyway?

Mark-H:
@SintezCode It looks like an unrelated commit has made it into this PR. That can be prevented by creating separate branches for each separate change that you make, so you can create PRs for each change individually.

PR #13455, opened 2017-05-22
Added ability to display a package changelog before upgrading

### LINKED ISSUE #13410: Add link to view changelog before updating a package
### Summary
This is a feature request for adding a "View changelog" button / link in the Update Package modal (the one below).

screenshot from 2017-04-05 14-39-22

### Observed behavior
Right now, you will only see the release notes _after_ you click the install button. If there's anything scary in there and you want to abort the installation, you'll end up with the package being displayed as "not installed" in the grid:

screenshot from 2017-04-05 14-47-37

Now, the package wasn't installed yet and you can go back to the currently installed version by selecting "Remove" in the grid.. But I don't know how many people are aware of that, as it's not really the behavior you'd expect when clicking a remove button.

### Desired behavior
It would be better if you can preview the changelog before deciding to install the package. Maybe you edited some core files of a package and only want to update if your fix is implemented in that release, or maybe the release comes with breaking changes that you'll need to sort out first on your local install.. Lots of reasons why you'd want to read the changelog first!

### Side note
The modal also says: Please select the version to update this package to, but it seems to always update to the latest release. If we're being really meticulous here, it should either show all updates between current and latest version, or simply say it will update to the latest version (with version number of course, and the new view changelog link).

rtripault:
### What does it do?

Adds the ability to display a package changelog before upgrading

screenshot_2017-05-22_14-38-09

### Why is it needed?

BECAUUUUUUSE!

### Related issue(s)/PR(s)

Closes #13410


core/model/modx/processors/workspace/packages/getlist.class.php
```diff
@@ -1,190 +1,190 @@
-<?php
```
rtripault: Sorry for the noise on that one...
That's just CR/LF change automatically performed by my IDE while i was playing around with this processor...

OptimusCrime:
God damnit, we should really do something about the mixed indenting and the line endings soon. This happens way too often.

Jako:
There is a .editorconfig available: https://github.com/modxcms/revolution/blob/2.x/.editorconfig

But maybe old files are not changed yet.

OptimusCrime:
No they are no. The editorconfig only applies (in most editors) when a file is opened, edited and saved.

Mark-H:
Very cool, thanks @rtripault!

manager/assets/modext/workspace/package.grid.js
```diff
@@ -489,18 +489,51 @@ Ext.extend(MODx.window.PackageUpdate,MODx.Window,{
...
+                            if (!changelog) {
+                                return;
+                            }
+                            // When the changelog link is clicked, display the changelog in a window
+                            Ext.get(changelog).on('click', function(elem) {
+                                var win = MODx.load({
+                                    xtype: 'modx-window'
+                                    ,fields: [{
+                                        xtype: 'box'
+                                        ,html: pkg.changelog
```
Jako: A replace of \n with <br> in pkg.changelog should improve the output.

PR #13462, opened 2017-06-02
Change new installs to create tables with InnoDB engine

opengeek:
### What does it do?
Creates tables using InnoDB table type in new MySQL installations.

### Why is it needed?
MyISAM is an obsolete MySQL table type/engine and we should be using InnoDB moving forward.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-dist: precise
+dist: trusty
+sudo: required
 language: php
 script:
   - phpunit --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ php:
   - nightly
   - 5.5
   - 5.4
-  - 5.3
+#  - 5.3
   - hhvm
 
 matrix:
--- a/core/model/modx/mysql/modaccess.map.inc.php
+++ b/core/model/modx/mysql/modaccess.map.inc.php
@@ -7,6 +7,10 @@ $xpdo_meta_map['modAccess']= array (
   'package' => 'modx',
   'version' => '1.1',
   'extends' => 'xPDOSimpleObject',
+  'tableMeta' => 
+  array (
+    'engine' => 'InnoDB',
+  ),
   'fields' => 
   array (
     'target' => '',
--- a/core/model/modx/mysql/modaccessaction.map.inc.php
+++ b/core/model/modx/mysql/modaccessaction.map.inc.php
@@ -8,6 +8,10 @@ $xpdo_meta_map['modAccessAction']= array (
   'version' => '1.1',
   'table' => 'access_actions',
   'extends' => 'modAccess',
+  'tableMeta' => 
+  array (
+    'engine' => 'InnoDB',
+  ),
   'fields' => 
   array (
   ),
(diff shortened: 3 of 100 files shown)


zaigham:
How can we change the existing ones? Is it complex process?

sottwell:
A big issue is that index fields can only be 190 or something chars long, while many of the existing tables use varchar(255) fields. This would probably only be a problem during installation, but that may depend on other database configuration settings. Those existing index key fields need to be adjusted - usually the 'name' fields. For example, for the original tables such as chunk and snippet 'name' fields are varchar(50) which is fine when using them as an index. The problem is with the newer Revo tables, such as the context table, use varchar(255).

opengeek:
@sottwell that is the utf8_mb4 issue you are speaking to, which will be dealt with in another pull request — InnoDB indexes can be 767 bytes I believe, which is plenty for the 3-byte utf8 charsets using `varchar(255)`.

opengeek:
@zaigham You can convert existing tables fairly easily. See https://stackoverflow.com/questions/3856435/how-to-convert-all-tables-from-myisam-into-innodb for some queries that can handle it for you.

pixelchutes:
Nice work! Will this change also cover any _future_ tables created during Extra installs?

Typically, this has been accomplished by adding the following to **core/config/config.inc.php**:

```php
$config_options = array (
  'override_table' => 'InnoDB',
);
```

I see the same override flag [in your commit](https://github.com/modxcms/revolution/pull/13462/files#diff-a70754cba47107aa25f9d6c9b181bb6eR165), I just am not sure if that persists beyond `setup/`?

opengeek:
No, it will not solve that issue. Looks like I will have to modify xPDO in order to change the default globally. Sigh…

.travis.yml
```diff
@@ -12,7 +13,7 @@ php:
   - nightly
   - 5.5
   - 5.4
-  - 5.3
+#  - 5.3
```
pixelchutes: @opengeek to confirm, is 2.6.0 officially drawing the line on php 5.4+? _Just trying to understand dropping 5.3 from the tests here..._

pixelchutes (commented):
Should we wait to fully review/test this PR until the accompanying xPDO fix (based on [this comment](https://github.com/modxcms/revolution/pull/13462#issuecomment-306547947)) is available for testing _future_ table creation (e.g. Extra installs), or do you consider that not a dependency since it's beyond "core"?

OptimusCrime:
Are the InnoDB changes backported into the xPDO repository? Both v2 and v3?

sdrenth:
What is the status of this PR, is it ready to be tested?

pbowyer (approved):
Looks good!

.travis.yml
opengeek: This was dropped because the new Trusty environments on Travis CI don't run PHP 5.3 anymore, but are needed to actually test this against InnoDB on MySQL 5.6+.

opengeek:
There is a major problem with this PR at the moment. MariaDB in some cases is reporting its version with a prefix of `5.5.5-` which forces MyISAM installation even though the actual version of MariaDB supports it. Trying to come up with a solution.

opengeek:
Ok, MariaDB issue resolved and I modified the index length for the uri field on modResource to 255 from 333, to account for utf8 (3-byte) with InnoDB's max byte length of 767. This latter issue will be moot when merging changes to support for utf8mb4, which will further limit the index length to 191.

OptimusCrime:
Can not this just adjust the length to 191 right away?

opengeek:
@OptimusCrime no, that would make no sense since this PR only addresses InnoDB, not utf8mb4.

Issue #13463, opened 2017-06-02
Help Page Newsletter Subscribe Fix


jaygilmore:
### Summary
The MODX Newsletter subscription form on the Help Page (/manager/?a=help) was using an outdated post form link and was failing.

### Step to reproduce
Visit /manager/?a=help and try to subscribe to the MODX Insider

### Observed behavior
Arrives at appilcation 404 page at Mailchimp

### Expected behavior
The newsletter subscription should work with the correct subscription form processor link in place. 

### Environment
NA

Mark-H:
Fixed per #13464

PR #13464, opened 2017-06-02
Update Subscribe Form

jaygilmore:
Newsletter Subscribe form was pointing at an old URL and old list ID
that are no longer in use. I’ve updated the email to use the correct
properties.

Issue #13463


Mark-H (commented):
I wonder if the site URL should be provided in that sign up form (merge 3)? Seems unnecessary to share that info for a newsletter signup..

jaygilmore:
@mark it was there before. I just fixed the merge field. Honestly, I don't have any issue removing it. We're not using it anyhow.

PR #13472, opened 2017-06-21
[in progress] Implement distributions into the setup (for new installs)

Mark-H:
This pull request is not done, I've opened it early to allow discussion on the topic and implementation.

### What does it do?
This adds an extra step to the setup that allows packages to be installed to get a quick site up and running. 

### Why is it needed?
At the MODX Meetup/Hackathon in Malta we sat down with people who never used MODX before to see how MODX could be made better. What are their challenges, how can we improve things with a fresh look. We split up into teams, and the team I was on decided to focus on the setup. 

One thing we discussed was that it should be easier for a new user to get a fully functional site. We've had a nicer base template for some time now, but that's still limited in showing how MODX works. So we thought that adding starter sites of sorts would be a really nice way to showcase that. 

At the same time, advanced users really don't care about starter themes. Some want the blank slate, or have their own starter sites they use to kick-start projects. But, some might still enjoy being able of quickly installing recommended packages. In the future this could be expanded with more advanced features for power users.

For now, this should be seen as a minimum viable implementation. 

It offers a selection of different site types which I labelled "distributions". These are kinda abstract, but they should offer a functional site out of the box that is properly set up (structured templates/chunks/assets). In the current iteration there's 2 distributions:

- Personal Blog (should be a chronological blog-style site, with Collections, perhaps comments (Quip or Disqus?), tagging with Tagger etc, archives?, author profiles?)
- Business (sort of standard business catalog site, home/services/team/contact us etc)

Others could be imagined and added over time (e.g. restaurant site, community site, one-page placeholder site etc). This functionality is ideal for new users that want to set up a site and see how MODX works to start tweaking things. 

The way these distributions work is that they specify a (comma separated) list of packages to install. Right now it's a (functional) proof of concept where just certain packages are used, so it doesn't actually install those sites yet, but the idea is that when we have some nice starter sites on the modx extras repo we can set those up.

The _Advanced_ distribution allows users to select packages from a list of recommended packages. See section below on core vs extras considerations. These extras are listed with a short (1 sentence) description/elevator pitch of what they do, as names alone like "Wayfinder" are not immediately clear on the function. 

#### Core vs Extras considerations

Back in evo, the core shipped with extras. They were actually included in the download. This meant the core required updates when an extra was updated, which is not ideal. 

In this PR, the packages are not included, they are just referred to by name. This makes sure it always downloads the latest version and the core does not need to be updated when an extra is.

It is still mean referring to extras from within the core. This means that when an extra goes in or out of favour over time, that the list should change. 

There's been some discussions on wether this perhaps should depend on an API so it's not "hardcoded" but I personally think using the existing contributor process and just keep it in the core.

To objectively determine what extras should or should not be listed, I think we should draft a set of guidelines on what extras can be accepted. I think this could take the form of something like "extras hosted on MODX.com with at least X downloads, which serve a specific need not covered by other more generic extras, and for which the author is not an alien". This might be something the MAB can help draft. 

This should also not be an infinitely growing list. If there's 10-15, I think that's more than enough. Perhaps the guidelines should include a hard maximum (want to add one more? Then one also needs to be removed, so you need to provide some really good reasoning), or in another way the guidelines should be made specific enough that it doesn't grow out of control. 

The page could link to MODX.com (and other extra providers like modmore and modstore) for more packages which can then be installed the normal way. 

### In progress

Still left to do before this can be considered for merging:

- [ ] Build some actual starter sites, available as packages on modx.com, that can be installed. 
- [ ] Lexicon all the text 
- [ ] Determine the list of packages to start with. There are some extras that might require some more debate (Wayfinder/getResources vs pdoResources, pthumb(of/on/up) etc), so I think it might be best to NOT have that discussion within this pull request. For starters I think we should build the starter sites, and use the bare minimum extras used to make those work.
- [ ] Design is pretty simple, might need a bit of a touch-up.

### Related issue(s)/PR(s)
N/a

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1316,7 +1316,7 @@ class modX extends xPDO {
      */
     public function getVersionData() {
         if ($this->version === null) {
-            $this->version= @ include_once MODX_CORE_PATH . "docs/version.inc.php";
+            $this->version = include MODX_CORE_PATH . "docs/version.inc.php";
         }
         return $this->version;
     }
--- a/setup/assets/css/style.css
+++ b/setup/assets/css/style.css
@@ -223,3 +223,18 @@ input, button, select {
   position: relative;
   top: -10px;
 }
+
+.distributions {
+	overflow: hidden;
+	padding-bottom: 2em;
+}
+.distribution {
+	float: left;
+	width: 25%;
+}
+.distribution label {
+	width: 100%;
+}
+.packages {
+	padding-top: 2em;
+}
--- a/setup/assets/js/sections/distribution.js
+++ b/setup/assets/js/sections/distribution.js
@@ -1,0 +1,36 @@
+Ext.onReady(function() {
+    MODx.Distributions.packages = Ext.select('.packages');
+    MODx.Distributions.packages.setVisibilityMode(Ext.Element.DISPLAY);
+    MODx.Distributions.packages.hide();
+
+    var distributions = Ext.select('.distribution input');
+    distributions.on('change', MODx.Distributions.selectDistribution);
+
+    Ext.select('#distribution').on('submit', function() {
+        Ext.select('.setup_navbar input').hide();
+    });
+});
+
+MODx.Distributions = function() {
+    return {
+        selectDistribution: function() {
+            var el = Ext.getDom(this);
+            var packages = el.getAttribute('data-packages');
+            if (packages === '*') {
+                MODx.Distributions.packages.show();
+            }
+            else {
+                MODx.Distributions.packages.hide();
+                packages = packages.split(',');
+                var inputs = Ext.select('.packages input');
+                inputs.each(function(i,v) {
+                    var el = Ext.getDom(i);
+                    el.checked = packages.indexOf(i.getAttribute('value')) !== -1;
+                })
+            }
+
+        },
+        packages: false,
+    };
+}();
+
--- a/setup/controllers/distribution.php
+++ b/setup/controllers/distribution.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * @var modInstall $install
+ * @var modInstallParser $parser
+ * @var modInstallRequest $this
+ * @package setup
+ */
+if (!empty($_POST['proceed'])) {
+    // Try to increase the time limit as installing a lot of packages may take a while
+    @set_time_limit(0);
+
+    // Install selected packages
+    $packages = array_key_exists('packages', $_POST) && is_array($_POST['packages']) ? $_POST['packages'] : array();
+
+    foreach ($packages as $package) {
+        $this->install->downloadAndInstallPackage($package, array());
+    }
+
+    $this->proceed('complete');
+}
+
+return $parser->render('distribution.tpl');
--- a/setup/controllers/install.php
+++ b/setup/controllers/install.php
@@ -9,7 +9,13 @@ $install->settings->check();
 if (!empty($_POST['proceed'])) {
     unset($_POST['proceed']);
     $install->settings->store($_POST);
-    $this->proceed('complete');
+    $installmode = $install->settings->get('installmode', modInstall::MODE_NEW);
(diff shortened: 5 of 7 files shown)


sottwell:
I would like even better to have a way to create a JSON array of extras and have them all automatically downloaded and installed. That way I could install my usuals in one step, and edit the list as necessary.

bezumkin:
Idea is great, but now we will have much more slow sites with getResources and Wayfinder. 

Do we really need to propose this old stuff to all newbies?

Mark-H:
@sottwell:

> In the future this could be expanded with more advanced features for power users.
> For now, this should be seen as a minimum viable implementation.

The idea of JSON definitions came up before, but one step at a time. Power users likely already have a way to get started, such as a custom vapor/teleport package or another process, so that is lower priority for this PR. 

@bezumkin:

>  Determine the list of packages to start with. There are some extras that might require some more debate (**Wayfinder/getResources vs pdoResources**, pthumb(of/on/up) etc), **so I think it might be best to NOT have that discussion within this pull request**. For starters I think we should build the starter sites, and use the bare minimum extras used to make those work.

The starter sites don't exist yet, so perhaps let's open a forum topic or a separate issue to discuss what those should look and work like. 

Also related:

> To objectively determine what extras should or should not be listed, I think we should draft a set of guidelines on what extras can be accepted. I think this could take the form of something like "extras hosted on MODX.com with at least X downloads, which serve a specific need not covered by other more generic extras, and for which the author is not an alien". This might be something the MAB can help draft.

I'd like the packages that are included to be based on objective guidelines, not anyone's personal preference. What's listed in the PR should be considered nothing more than a placeholder until the list of extras is determined.

bezumkin:
Thanks, Mark! Sorry for my hasty comment - I just looked in the code instead of description. My bad =(

Mark-H:
No worries, it's a looooooooong piece of text so I can imagine you jump straight into the code. :P

mrhaw:
I love this but it should be based on option to setup during install!?

Mark-H:
@mrhaw Sorry, I'm not following... what do you mean?

mrhaw:
@Mark-H Disregard my comment, even the title says we want to add this to the setup. I don't know what I was thinking. 

It would be awesome if we also could customize this. I've been using this plugin for this purpose: https://gist.github.com/jkenters/10651723

theboxer:
I'm not favour of this at all :)

@sottwell There is no need for this feature, we have teleport to pack your "default" Revo install, which can be than easily deployed.

@Mark-H @bezumkin There's no chance to agree on all packages to present to user without harming feelings of some community members. How do you want to choose between extras that solves same thing? Like getResources / pdoTools - some ppl use gr, some pdoTools and most likely not going to switch to the other one. I can see more issues coming up in the future in case this gets released which will be list of paid extras.

Mark-H:
Paid extras typically require some type of authentication, so from modmore's perspective, I have no expectation of those being included for the time being. ;)

I think that, rather than trying to build a list of extras we can all agree on, it might make more sense to focus on building two good starter sites and only use the extras that needs. That way all extras have a good reason to be there.

Mark-H:
As discussed on the MAB call, I started on a draft recommendation on how to select packages to include. It can be found here: https://github.com/modxcms/mab-recommendations/pull/25 - maybe a good idea to have discussions at https://github.com/Mark-H/mab-recommendations/issues to not clutter up the recommendation PR too much.

alroniks:
Hi @Mark-H I guess it's time to move this PR into 3.x or close and recreate it again base on 3.x. Do you agree?

opengeek:
This hasn't made any progress in over a year or gotten any feedback in almost a year. Closing.

Mark-H:
I'd be happy to revive this at some point, however first there needs to be consensus on the packages to install and where to keep that list. There have been discussions in Slack about hardcoding them in the source (as per my proposal) or exposing an API that can be changed more dynamically, for example.

From my point of view, it's up to a wider selection of the community or MAB to answer that first question, and when there's consensus there I can handle the technological implementation after that.

Issue #13461, opened 2017-05-26
Context setting listing fetches too few settings


OptimusCrime:
### Summary
The context setting listing does not list the amount of settings it should. The list is truncated too soon.

### Step to reproduce and observed behavior
We have (the default value?) setting named `default_per_page` as 20. However, in the Context Settings tab, the `Per Page` value is 30. What is even weirder is that the listed settings is only 12 although there are more settings in the list. If we increase the `Per Page` value to 50 it lists 20 settings, and 300 lists all the 27 settings.

In addition, the pagination states that there are no more pages (Page 1 of 1).

### Expected behavior
Correctly list the number of settings it is said to list.

### Environment
MODX 2.5.7

neoneddy:
I've seen this too.   Use CrossContext Settings as a work around for now.

Jako:
Maybe this is caused by the namespaces of the context settings (if you want to investigate that further).

Flowan:
I'm going to try and fix this today.

Flowan:
Confirmed bug in 2.5.x I will update my current PR. #modxbughunt

gpsietzema:
#modxbughunt #1point to @Flowan

Issue #12823, opened 2015-12-10
Missing context settings after editing name or description


tillilab:
If a user changes one of the context settings, modifying name or description, new entries are added inside the modx_lexicon_entries table; 
so the function inside **/core/model/modx/mysql/modcontextsetting.class.php** outputs multiple results due to the join conditions of the query.
This produces inconsistent **pagination** of context settings and some of them are just **not** displayed.

May be possible solution could be:
1) add inside the modcontextsetting.class.php the query option: 
$c->groupby("modContextSetting.{$xpdo->escape('key')}");
Not a real solution...since it returns the first result and usually it's not the most updated lexicon entry

2) avoid duplication of lexicon entries, updating existing ones

Mark-H:
Option 2 sounds best.

Flowan:
I'm going to try and fix this today. #modxbughunt

EasterPeanut:
#modxbughunt Tested this PR and works, no new entries in the modx_lexicon_entries table are added.

Mark-H:
Fixed in 2.5.6

PR #13493, opened 2017-07-07
Fixed page size for grids #13461 #modxbughunt

flowan:
### What does it do?
Use the correct value that is set in the system setting default_per_page for all grids.

### Why is it needed?
Fix the pagination for grids (system settings, context settings, user overview, etc).

The code `parseInt(MODx.config.default_per_page) > 30 ? parseInt(MODx.config.default_per_page) : 30` would not allow you to set a page size of 10 or 20 as `parseInt(MODx.config.default_per_page) > 30` will always return false if the value is lower then 30. So the default_per_page setting will only work of you use higher values.

### Related issue(s)/PR(s)
Related issue #13461

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.settings.js
+++ b/manager/assets/modext/widgets/core/modx.grid.settings.js
@@ -151,7 +151,7 @@ MODx.grid.SettingsGrid = function(config) {
         ,primaryKey: 'key'
         ,autosave: true
         ,save_action: 'system/settings/updatefromgrid'
-        ,pageSize: parseInt(MODx.config.default_per_page) > 30 ? parseInt(MODx.config.default_per_page) : 30
+        ,pageSize: parseInt(MODx.config.default_per_page) || 20
         ,paging: true
         ,collapseFirst: false
         ,tools: [{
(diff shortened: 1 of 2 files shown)


rtripault:
I believe the idea there what to prevent having a too high value in `default_per_page`. So i'd :-1:

Mark-H:
Hey @Flowan, could you please use some more descriptive titles on pull requests and in your commit messages? ;)

Mark-H:
Aside from @rtripault's comment, I'm not sure if this actually fixes the issue in #13461 where the total is off and not all settings are returned. Any thoughts on that @Flowan?

Flowan:
@Mark-H @rtripault Added more information to the PR.

Flowan:
@Mark-H I think the total being off had to do something with this issue #12823 and should be fixed.

gpsietzema:

theboxer:
@Flowan please check the comment in the review.

rtripault:

Flowan:
@theboxer @rtripault Reverted the change for the packages grid

Issue #13494, opened 2017-07-07
Media browser active state problem in tree


pixelive:
### Summary
Media browser active state problem in tree.

### Step to reproduce
Create a directory with no subfolders, click the directory once, then click another directory and after that switch back. You cannot select folder anymore.

### Observed behavior
You cannot select a childless directory if it is clicked once.

### Expected behavior
Selecting a childless directory should be possible.

### Environment
-

rtripault:
Confirmed on 2.x branch, with Chrome (if relevant)

Flowan:
Fixed the issue, see pr https://github.com/modxcms/revolution/pull/13496 #modxbughunt

gpsietzema:
#modxbughunt #1point to @rtripault

sdrenth:
Also confirmed on 2.5.x branch

Alroniks:
Merged, so closing.

PR #13496, opened 2017-07-07
Fix media browser active state in tree #13494 #modxbughunt

flowan:
### What does it do?
Moved the is node expanded check from beforeclick to handleclick. This was needed because if the beforeclick returns false, any other event after this would not be triggered.

### Why is it needed?
When selecting a directory the active state is not always set correctly. So when you upload a file, the file is not being uploaded in the directory you selected.

### Related issue(s)/PR(s)
Related issue #13494

### DIFF
--- a/manager/assets/modext/widgets/core/tree/modx.tree.js
+++ b/manager/assets/modext/widgets/core/tree/modx.tree.js
@@ -153,7 +153,7 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
                 node.ui.addClass('x-tree-selected');
             }
         });
-        
+
         var r = Ext.decode(resp.responseText);
         if (r.message) {
             var el = this.getTreeEl();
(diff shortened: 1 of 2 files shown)


gpsietzema:

theboxer:
@Flowan seems like this PR doesn't fix the issue on 100%.

I have two childless directories `test` and `test2`. I can switch between them without an issue after applying this fix, but, when I select directory with children (for example `core`) and then select my childless dir `test`, the active flag stays on the `core` directory. After second click on `test` it works as expected.

Flowan:
@theboxer It's because /manager/assets/modext/modx.jsgrps-min.js still contains the old version. Do you know how to rebuild modx.jsgrps-min.js? Thanks in advance

theboxer:
@Flowan Yes, I do know how to rebuild it (but it's required to not build new version as part of the PR, I do it before merge), but it's not fixing the problem. I don't use minified versions of anything in my Revo install.

gpsietzema:

rtripault:

Flowan:
@rtripault Yes :)

gpsietzema:
Just tested this an it works on 3 machines here, @rtripault @theboxer can you try again?


That's #1point for me I guess :)

rtripault:
Confirming this is working as expected (not able to reproduce your issue @theboxer).

sonicpunk:
@gpsietzema do we give Romain here another point as well?

sdrenth:
Can confirm that the fix works on 2.5.x branch

PR #13499, opened 2017-07-07
Make sure we refresh the parent (resource) node when creating the first children

### LINKED ISSUE #13492: Quick create resource does not show new resource in tree if it's the first child
### Summary
Quick creating a resource does not show the newly created resource in the tree if it's the first child of its parent.

### Step to reproduce
Quick create a resource in a resource that has no children.

### Observed behavior
The parent does not expand, so the newly created resource is not shown in the tree.

### Expected behavior
The parent should expand and the newly created resource should be shown in the tree.

### Environment
MODX 2.5.8-dev

rtripault:
### What does it do?

When creating the first children node for a resource, using the quick create window, make sure we refresh the right tree node

### Why is it needed?

See related issue

### Related issue(s)/PR(s)

Closes #13492

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.tree.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.tree.resource.js
@@ -462,7 +462,7 @@ Ext.extend(MODx.tree.Resource,MODx.tree.Tree,{
             ,listeners: {
                 'success':{
                     fn: function() {
-                        this.refreshNode(this.cm.activeNode.id, true);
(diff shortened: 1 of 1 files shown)


gpsietzema:
#modxbughunt #3points to @rtripault

rtripault:
Thanks, but shouldn't this need to be confirmed working first ? ;p

theboxer:
@rtripault this should go to 2.5.x branch, will you have a chance to submit it there? The fix is working though.

theboxer:
Merged to 2.x and cherry picked to 2.5.x

sonicpunk:

Issue #13498, opened 2017-07-07
Privacy issue: News feeds in the manager are active by default and phone home to google


mindeffects:
### Summary
See next section:

### Why is it needed?
After installing a fresh MODX and logging into the manager, google gets connected via feedburner. So before a website is even started being build it is already "in the index". This causes some privacy issues because maybe the URL of the website should not be made public or for other reason. Having the feeds turned on by default is done without asking the user and getting his OK. 

### Related issue(s)/PR(s)
none

### Step to reproduce
Install MODX locally and have a network monitor (like little snitch on macOS) running. It will report outgoing connections like "www4.l.google.com" and also to "develop.modx.com" or "api.modx.com"

### Observed behavior
When having the feeds are turned of no connection are being made. Yeah! ;-)

### Expected behavior
Don't connect any "other" servers without asking firat.

### Environment
any (except when behind a strong firewall ;-) )

gpsietzema:
#modxbughunt #1point to @mindeffects

Omeryl:
Just putting my thoughts down.

1. The URL of the website isn't "exposed," the IP of the server definitely is. But with proper vhosting they won't just see the site. 
2. We could definitely have an opt-in, that idea is cool.
3. Any other servers? Including *.modx.com? How might we do that?

OptimusCrime:
Agree with @Omeryl

OptimusCrime:
Anyone else or can this be closed?

mindeffects:
OK, can be closed. Thanks!

OptimusCrime:
@modxbot close

PR #13500, opened 2017-07-07
Turning of the feeds to gain some more privacy

mindeffects:
### What does it do?
Turning off the news feeds in the manager by default.

### Why is it needed?
After installing a fresh MODX and logging into the manager, google gets connected. So, before a website is even started being build, it is already "in the index". This causes some privacy issues because maybe the URL of the website should not be made public or for other reason. Having the feeds turned on by default is done without asking the user and getting his OK. 

### Related issue(s)/PR(s)
none

#13498 

#modxbughunt

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -549,7 +549,7 @@ $settings['feed_modx_news']->fromArray(array (
 $settings['feed_modx_news_enabled']= $xpdo->newObject('modSystemSetting');
 $settings['feed_modx_news_enabled']->fromArray(array (
   'key' => 'feed_modx_news_enabled',
-  'value' => '1',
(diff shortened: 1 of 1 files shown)


gpsietzema:
#modxbughunt #3points to @mindeffects

rtripault:
While it would indeed solve the mentioned, i doubt disabling the security feed as default is the right approach...

Mark-H:
I understand the privacy argument but am not in favor of this change... seems a bit drastic?

gpsietzema:
It would be kind of cool if this is like a questions when first loading the dashboard. Something like: "Do you want to enable the default dashboard? Yes/No". Or have it in the setup.

theboxer:
Sorry, disabling security feed as a default is a no go.

Issue #13495, opened 2017-07-07
Media browser creating directory not always refreshing the tree correctly


pixelive:
### Summary
Media browser creating directory not always refreshing the tree correctly.

### Step to reproduce
Create a new directory and leave the parent directory empty.

### Observed behavior
Not refreshing the tree at all.

### Expected behavior
Should refresh the tree.

### Environment
-

rtripault:
Confirmed on 2.x branch

Flowan:
I'm going to try and fix this today. #modxbughunt

gpsietzema:
#modxbughunt #1point to @rtripault

Alroniks:
Merged, so closing.

PR #13501, opened 2017-07-07
Fix refresh media browser tree after creating a directory #13495 #modxbughunt

flowan:
### What does it do?
Check if the parent node is empty. If its empty add a slash so the whole directory is refreshed correctly.

### Why is it needed?
After a new directory is created it is not visible. This might confuse the user that will most likely try to create the directory again.

### Related issue(s)/PR(s)
Related issue #13495

### DIFF
--- a/manager/assets/modext/widgets/system/modx.tree.directory.js
+++ b/manager/assets/modext/widgets/system/modx.tree.directory.js
@@ -508,18 +508,26 @@ Ext.extend(MODx.tree.Directory,MODx.tree.Tree,{
     ,createDirectory: function(item,e) {
         var node = this.cm && this.cm.activeNode ? this.cm.activeNode : false;
         var r = {
-            'parent': node && node.attributes.type == 'dir' ? node.attributes.pathRelative : '/'
+            ‘parent’: node && node.attributes.type == 'dir' ? node.attributes.pathRelative : '/'
             ,source: this.getSource()
         };
-        if (r.parent === '') {
-            r.parent = '/';
-        }
+
         var w = MODx.load({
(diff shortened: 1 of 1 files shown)


gpsietzema:

theboxer:
Somehow I don't understand the issue so I'm not able to check this PR. @Flowan can you please give me more detailed instructions on how to reproduce the original issue? Or maybe @rtripault who confirmed the issue?

rtripault:
* head over to the media browser (?a=media/browser)
* in the top right toolbar of the tree, hit the folder icon to create a new folder (defaulting to the root of the source) and create a new folder
* notice the tree is not refreshed and your newly created folder is not listed
* hit the refresh icon on the tree and see your folder

theboxer:
In that case this PR doesn't fix the issue :)

rtripault:

Flowan:
This fix should work. Tested this on 2 different machines. Could anyone confirm this works or not? :)

opengeek:
I just tested this one myself and it is not refreshing for me. I have to still click the refresh icon to see newly created folders in the media browser.

devlanda:
I can also confirm that the current fix does not work.

Jako:
`assets/modext/widgets/core/tree/modx.tree.js` is one of the core files that is uglified into `assets/modext/modx.jsgrps-min.js` with
https://github.com/modxcms/revolution/blob/2.x/_build/templates/default/gruntfile.js

So without running `grunt build` inside of `_build/templates/default`, the fix can't be checked in a local installation.

sdrenth:
@Flowan has committed some new code which fixes tree refreshing issues.

Issue #4581, opened 2011-05-04
Renamed Chunk Does Not Refresh Tree


johnywhy:
johnywhy created Redmine issue ID 4581

When you open an existing chunk, change the name, and click save, the Resources Tree does not refresh.

adamwintle:
@garryn @OptimusCrime I just checked and it seems that this bug is there resent, do you think it'll be an easy one to fix?

OptimusCrime:
I changed my mind, this is indeed a bug and it should be fixed.

However, I have no idea how to do this. Perhaps there are some ExtJs wizards out there that could take a look at it?

wuuti:
#modxbughunt still confirmed

Finetuned:
"I'm going to try and fix this today."

PR #13502, opened 2017-07-07
Update title in element tree after renaming resource [#4581] #modxbughunt

Finetuned:
### What does it do?
Corrected tree component lookup in  various manager elements on on successful save event.

### Why is it needed?
Element tree did not refresh after renaming chunks, snippets, templates, tvs or plugins.


gpsietzema:

gpsietzema:

Issue #13448, opened 2017-05-04
Enhance duplicate dialog


pepebe:
### Summary
If you duplicate a resource by rightclicking on it in the tree, the duplicate dialog doesn't give any feedback which resource you are going to duplicate.

![bildschirmfoto vom 2017-05-04 09-58-56](https://cloud.githubusercontent.com/assets/667315/25694752/5d9caca8-30b0-11e7-83f2-8444c27f79ae.png)

### Step to reproduce
resource -> rightclick -> context menu -> duplicate

### Observed behavior
The right clicked resources looses focus, the dialog doesn't give any clue about the resource title or id of the resource you are about to duplicate. If you are clumsy you'll duplicate a whole part of the tree including all children and so on (I just did that...)

### Expected behavior
Give some feedback about what you are about to do.

Show the pagetitle and the id on top of the dialog

![bildschirmfoto vom 2017-05-04 10-07-59](https://cloud.githubusercontent.com/assets/667315/25694996/9756b44c-30b1-11e7-85a0-d4fc652dd2a5.png)

As the available space in the head might be a concern, moving the pagetitle/id inside the box might be3 a good compromise.

### Environment
MODX all versions.

mindeffects:
+1

wuuti:
Even more important: the same applies to the delete alert. You don't have a clue if you are really deleting the right resource there...

pepebe:
@wuuti Yes, the deleted alert is also lacking in user friendliness. That being said, the whole deletion practice in MODX is terrible broken:

https://github.com/modxcms/revolution/issues/764

I really have to split this thread into different issues...

Alroniks:
Fixed by #13475.

@modxbot close

Alroniks:
#modxbughunt

gpsietzema:
#modxbughunt #1point to @Alroniks

Issue #764, opened 2013-12-28
Enhancement: Redesign how "deleted" things are handled.


pepebe:
I've started a discussion regarding this topic at the forum: http://forums.modx.com/thread/88277/redesign-how-deleted-things-are-handled. Feel free to do some mind storming there. I'll add good ideas to this thread.

I know there are extensions that make it easier to handle deleted things (VersionX, etc), but I think that some aspects of this topic have to be handled on a lower level.

**A. Add permanent deletion of resources on an individual level.**
I think it should be possible to pick a single resource and delete it permanently from the database. Right now its all or nothing and I don't think this is good solution.

Edit: This would require a new entry in the tree's context menu as well as a new button on the edit resource page.

**B. Change how the trash bin works**
I'd like to have a list of all deleted resources in a neat grid instead of a button asking me to send a number of anonymous things to the netherealms. 

Edit: Some CMP could do the same job, so that could easily be achieved mit a new extension (I think this should be a core feature). 

**C. Add delete and undelete to all elements (snippets, chunks, plugins, etc)**
Even people that don't practice wild west programming delete the wrong things from time to time. It wold be nice to have a second line of defense.

Edit: Bob Ray pointed out that this would require a new columns in a couple of tables. I don't see a problem with that, but I might me missing something.

**D. Enhance the delete modal**
The modal shouldn't just ask Are you sure you want to remove **this file/resource?**. Instead it should ask: Are you sure you want to remove **the resource Contact Details (9)?** 

Especially on the elements tab, where deletion is permanent, this would be a very important improvement.

**E. Remove the delete button from already deleted resources**
Thats only a cosmetic feature. Deleted resources still have a "delete" button next to the "duplicate" and the "view" button. Its redundant and has absolutely no meaning or function.

Note: Bob Ray pointed out, that instead of this button a "delete permanently" would be a good idea.

If you have further ideas how to improve how MODX handles deletion, please share your thoughts. If you have a solution for one of those issues, create a pull request or contact me and I'll create a pull request for you.

Regards,

pepebe

P.S. Happy Holidays to everyone!

sonicpunk:
No activity here since 2014.  Can it be closed?

OptimusCrime:
This has not gotten any attention from the core team. I belive that the functionality implemented today is tested, tried and make sense.

[...]

PR #13497, opened 2017-07-07
Enhance delete resource dialog

lawrenz1337:
### What does it do?
Added pagetitle and ID of the deleted resource to the delete dialog

### Why is it needed?
Just a continuation of #13475. The default UI for 'Delete' dialog doesn't show the name and ID and it is pretty easy to forget which resource you are trying to delete.

### Related issue(s)/PR(s)
#13475 #13448

PR #13504, opened 2017-07-07
Enhanced the delete confirm alert (fixes parts of #764, #13448) #modxbughunt

wuuti:
### What does it do?
Adds two lines to the delete confirmation dialogue which show the context the resource is in and the path to and the name of the resource. If the permission to view the ids in the tree is set, also shows the id of the resource in question. 

### Why is it needed?
The confirmation alert does not show the resource to delete, therefore it may happen that you don't see any more if you clicked on the right resource to delete.

### Related issue(s)/PR(s)
The issue was partly described and mentioned in #764 and #13448.

### DIFF
--- a/core/model/modx/processors/resource/getnodes.class.php
+++ b/core/model/modx/processors/resource/getnodes.class.php
@@ -498,6 +498,12 @@ class modResourceGetNodesProcessor extends modProcessor {
             'preview_url' => (!$resource->get('deleted')) ? $this->modx->makeUrl($resource->get('id'), $resource->get('context_key'), $sessionEnabled, 'full', array('xhtml_urls' => false)) : '',
             'page' => empty($noHref) ? '?a='.(!empty($this->permissions['edit_document']) ? 'resource/update' : 'resource/data').'&id='.$resource->id : '',
             'allowDrop' => true,
+
+            // the following are added for convenience to enable direct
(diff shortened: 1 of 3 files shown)


lawrenz1337:
I've done a much more simple solution to this in #13497 could you update your repository cause right now it will definitely cause conflicts. I also referenced it is #13448 . Sorry about that :/

gpsietzema:

#modxbughunt #1point to @lawrenz1337

wuuti:
To note: the problem was solved in parallel by @lawrenz1337 (targeted to 2.x), whereas I targeted 2.5.x. My solution is more complex (introduces a new processor which might be used elsewhere), but gives also more information in the confirmation alert (the path to the resource, the context and the id). If this is considered not necessary by the reviewers, we could drop it. Imho it is useful to see the context and the path and the id, because you may have multiple resources with the same pagetitle (e.g. home in different contexts) - making only displaying the pagetitle not distinctive enough...

lawrenz1337:
@wuuti My fix show title in a format `Delete Pagetitle (ID)` so the duplicate pagetitle is not an issue. I haven't noticed that your fix targeted 2.5.x my bad. It will be fine then but this situation in kinda awkward; I like your solution though

theboxer:
I'm closing this in favour of already merged #13497 which solves the issue.

Issue #13209, opened 2016-12-04
Blank admin dashboard if included Snoopy HTTP client is missing openssl extension for getting security feed


flowdy:
### Summary
Snoopy just exits the PHP request if it cannot find OpenSSL extension.

### Step to reproduce
Uninstall openssl extension if needed, enable security feed on dashboard.

### Observed behavior
The manager is all blank (not a bit of content) after admin log in.

### Expected behavior
An error message "HTTPS feed url cannot be fetched: openssl extension not detected" or alike should be shown at the place where the feed content would be rendered.

### Environment
Tested on MODx 2.5.{1,2}-pl on nginx with PHP-FPM.

Mark-H:
Snoopy?

Jako:
RSS Feed Fetcher: https://github.com/modxcms/revolution/search?utf8=%E2%9C%93&q=snoopy

Jako:
@flowdy As workaround you could remove those widgets from the Dashboard.

Mark-H:
... learned something new today :D

flowdy:
Yes, I have done that, thank you.

What a fun that was poking around in the sources with `die("Things work hitherto");`, thus approaching the position where they stop working. On `exit()`ing, PHP certainly has no reason whatsoever to propagate that as an error. One could argue that exit() from a function in module loaded should be always be interpreted as `die("exit() called from a module is nearly always wrong. Shame on you.")`.

OptimusCrime:
@flowdy Could we get the actual PHP error/stacktrace?

flowdy:
I do not have an error stack trace, because exit() is no die().
See <https://github.com/modxcms/revolution/blob/2.x/core/model/modx/xmlrss/extlib/snoopy.class.php#L139>.

OptimusCrime:
Ah, I am sorry. I did not understand the context of your previous comment. I am not sure how to use the snoopy library or call it from MODX. Do you know if replacing the exit with a die would solve the problem correctly? In that case I can open a PR with this fix pretty soon.

Jako:
Thats the problem.
https://github.com/modxcms/revolution/blob/fbf126690f6ce903faeaeb9f1d958211fc2203c9/core/model/modx/xmlrss/extlib/snoopy.class.php#L139

Are the widgets called directly? Maybe the content could be fetched remote.

flowdy:
[...]

Issue #2797, opened 2010-11-13
Replace MagpieRSS with SimplePIe


freelancewebdev:
freelancewebdev created Redmine issue ID 2797

I would suggest that MagpieRSS is replaced with Simplepie for the following reasons as described on the Simplepie website:

SimplePie is being actively developed.
The most recent development release of MagpieRSS was 2.0-alpha-alpha-alpha back in October 2006. The last stable release was in November 2005.
SimplePie doesn't require you to understand the XML guts.
MagpieRSS offers you ONLY a data array based on the guts.
SimplePie fully supports AND normalizes all documented versions of RSS, and all popular versions of Atom.
MagpieRSS has very limited Atom support, and doesn't do much normalizing.
SimplePie can correctly parse Tim Bray's Atom 1.0 feed.
MagpieRSS doesn't come close.
SimplePie has a well-trafficked support forum, public SVN repository, and a TON of unit tests.
Does ANYONE know what's going on with MagpieRSS?
SimplePie sanitizes the content of the feed to ensure safety against malicious feeds.
MagpieRSS doesn't implement any kind of data sanitization.
SimplePie has a less restrictive license than MagpieRSS.
MagpieRSS is licensed under the GPL which is more restrictive than SimplePie's BSD licensing.
SimplePie is contained in a single file for easier portability.
MagpieRSS comes in 4 separate files making it harder to move around without issues.
SimplePie supports the detection and embedding of audio and video podcasts including the popular Media RSS and iTunes RSS elements.
MagpieRSS has no explicit enclosure support.

amdbuilder:
This issue is still valid as 2.3.3 still references it.

hugopeek:
Looking into this today for #modxbughunt.

Learned from @goldsky that the news feeds in MODX are indeed still using Magpie. My question is: although MagpieRSS is not being actively developed anymore, does it still work without issues? Also in PHP7?

This seems to be related to #13209 also. As @Mark-H mentioned there already: maybe it's time to switch to a completely different library with MODX3.

gpsietzema:
This isn't relevant anymore t.b.h. It still works. I think we should close it until it poses problems. It still works in PHP7.

@modxbot close

sonicpunk:
#modxbughunt #1point to @gpsietzema

freelancewebdev:
Just looking at this from the perspective of the library still functioning seems a bit odd to me - what about security?

OptimusCrime:
[...]

PR #13507, opened 2017-07-07
Load feeds on dashboard with AJAX instead of on page render

Mark-H:
### What does it do?
This changes the way the news and security feeds on the dashboard are rendered. Previously, this happened during rendering of the manager (i.e. before the manager HTML is sent to the browser). Now, it happens asynchronously when the manager page has been rendered by ExtJS.

### Why is it needed?
The RSS feeds can be quite slow, especially in remote places. They also depend on an internet connection to work, which isn't always available. When the feeds are slow to load, the entire dashboard is slow to load, which is bad.

There are also other issues with the current RSS reader implementation, for example, #13209 where the entire manager is blank if the underlying RSS class can't find OpenSSL. By moving it into a separate PHP process, such issues will no longer break the entire manager just like that. I've not changed how that implementation works for now, other than putting the calls into a different place so it's more resilient. 

### Related issue(s)/PR(s)
I can't find a specific issue requesting this to be loaded this over AJAX, but #13209 and #2797 are somewhat related.

### DIFF
--- a/core/model/modx/processors/system/dashboard/widget/feed.class.php
+++ b/core/model/modx/processors/system/dashboard/widget/feed.class.php
@@ -1,0 +1,82 @@
+<?php
+
+/**
+ * Class modDashboardWidgetFeedProcessor
+ *
+ * Used to load the news and security feeds on the dashboard over AJAX. The processed feed content
+ * (i.e. HTML) is returned in object->html.
+ */
+class modDashboardWidgetFeedProcessor extends modProcessor
+{
+    /**
+     * @var modRSSParser
+     */
+    protected $rss;
+
+    public function process()
+    {
+        $feed = $this->getProperty('feed', 'news');
+        if (!in_array($feed, array('news', 'security'), true)) {
+            return $this->failure('Invalid feed type');
+        }
+
+        sleep(5);
+        $enabled = $this->modx->getOption('feed_modx_' . $feed . '_enabled', null, true);
+        if (!$enabled) {
+            return $this->failure();
+        }
+
+        $url = $this->modx->getOption('feed_modx_' . $feed);
+        // Check if the domain is valid before attempting to load it
+        $feedHost = parse_url($url, PHP_URL_HOST);
+        if ($feedHost && function_exists('checkdnsrr') && !checkdnsrr($feedHost, 'A')) {
+            return $this->failure();
+        }
+
+        return $this->loadFeed($url);
+    }
+
+    public function loadFeed($url)
+    {
+        $this->modx->loadClass('xmlrss.modRSSParser','',false,true);
+        $this->rss = new modRSSParser($this->modx);
+
+        $o = array();
+        $rss = $this->rss->parse($url);
+        if (is_object($rss)) {
+            foreach (array_keys($rss->items) as $key) {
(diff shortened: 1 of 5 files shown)


core/model/modx/processors/system/dashboard/widget/feed.class.php
```diff
@@ -0,0 +1,81 @@
...
+    public function process()
+    {
+        $feed = $this->getProperty('feed', 'news');
+        if (!in_array($feed, array('news', 'security'), true)) {
+            return $this->failure('Invalid feed type');
+        }
+
+        $enabled = $this->modx->getOption('feed_modx_' . $feed . '_enabled', null, true);
+        if (!$enabled) {
+            return $this->failure();
```
rtripault: Is this really a failure ? Seems perfectly legit  to have those disabled.

Or maybe is that because the request should have never occurred ?

core/model/modx/processors/system/dashboard/widget/feed.class.php
```diff
@@ -0,0 +1,81 @@
...
+        if ($feedHost && function_exists('checkdnsrr') && !checkdnsrr($feedHost, 'A')) {
+            return $this->failure();
+        }
+
+        return $this->loadFeed($url);
+    }
+
+    public function loadFeed($url)
+    {
+        $this->modx->loadClass('xmlrss.modRSSParser','',false,true);
```
rtripault: Why not `getService` ?

core/model/modx/processors/system/dashboard/widget/feed.class.php
rtripault: Yeah... i should have reviewed the whole diff first!

Mark-H: Yeah, this shouldn't actually trigger. Was tempted to take it out, and happy to do so if you think that's better ;)

core/model/modx/processors/system/dashboard/widget/feed.class.php
Mark-H: No particular reason. Mostly just copied the old code from the widgets into the processor, and that used it like this... getService is probably cleaner though.

core/model/modx/processors/system/dashboard/widget/feed.class.php
rtripault: No, i'm fine with it (we can never assume only core components would use it).
I was just concerned about the "failure" semantic...

Mark-H:
Updated it to use getService instead :)

rtripault:
Working nicely, many thanks!

gpsietzema:

#modxbughunt #3points to @Mark-H

christianseel:
@Mark-H This might raise potential privacy issues as described here:
https://github.com/modxcms/revolution/pull/13500 / https://github.com/modxcms/revolution/issues/13498

Sending the requests via Ajax includes data like the referrer, which can be used to identify the MODX (manager) location.
/cc @mindeffects

Mark-H:
The AJAX request is only local to MODX; it still loads the feed server-side (just in a separate process) just like before.

christianseel:
Ah gotcha! That of course if perfect! 👍

Issue #12077, opened 2014-10-21
Incorrect created timestamp for downloaded packages


pixelchutes:
If you download and install a package _(after 12pm / noon)_, the `created` timestamp in the database is not correct. The values for `updated` and `installed` are fine, however.

**Package Details:**

![package_details](https://cloud.githubusercontent.com/assets/352182/4725761/2dde8c08-595c-11e4-8b11-340a94428e77.jpg)

**Database: modx_transport_packages**

![modx_transport_packages](https://cloud.githubusercontent.com/assets/352182/4725823/a7fb16d2-595c-11e4-9771-3fe528c9c255.jpg)

pixelchutes:
**NOTE**: This particular bug is best reproduced when installing packages sometime during the p.m. *(after noon)*

For example, if you install a package at 8**am**, the Package > View Details `Uploaded On` value will appear correct.

However, if you install a package at 8**pm**, the Package > View Details `Uploaded On` value will be **incorrect**.

pixelchutes:
Just tested this on 2.5.5 and was unable to reproduce. 

Closing for #modxbughunt but planning to perform one final test at top of the hour (>= 13:00 test) 😉

pixelchutes:
Confirmed still an issue. :'(

Alroniks:
Fixed for #modxbughunt by #13509

PR #13509, opened 2017-07-07
Fixed date format for created filed of package in package provider

alroniks:
### What does it do?
It changes the date format to correct when somebody downloads package from package provider.

### Why is it needed?
Before in format string was defined `h` for hours, that means 12-hour format of hours and this value was parsed incorrectly. When time was 13:45 actually it was parsed to 01:45 am. Now it replaced by strftime as in other places where package management manipulates of dates.

### Related issue(s)/PR(s)
#12077
#modxbughunt


gpsietzema:

sonicpunk:

Issue #11954, opened 2014-09-03
[2.3] Strange focus problem on clickable elements (especially buttons) in Google Chrome


gadgetto:
Using Google Chrome there seems to be a problem while releasing the focus of a button element (and other similar elements).

**Steps to reproduce this:**

1) open Menu Manage / Users (as a sample)
2) click on the **Clear filter** button (on the button text) - the button background gets blue
3) move mouse cursor away

Normally the focus of the button should release and the background color of the button element should change back to white, but most times it stays blue until you click somewhere outside the button element.

If you don't click on the center of the button element (near border) it mostly works as expected.
This also happens on other button-like elements.

No problem in: Safari, Firefox, Opera
Couldn't test in IE...

Finetuned:
"I'm going to try and fix this today."

PR #13510, opened 2017-07-07
Unhilite the 'Clear Filter' button on mouseout across the manager [#11954] #modxbughunt

Finetuned:
### What does it do?
Added a mouseout listener to the 'Clear Filter' buttons across the manager.

### Why is it needed?
When the button is clicked,  Chrome is adding an x-btn-focus class to the dom which was not being removed on mouseout.
Other browsers were not adding this class and thus not displaying the problem.

Overrides can't be used for events so the listener had to be added to each file displaying the issue -> Ugly.

For ease of testing the affected panels are as follows:

Media: Media Sources
Manage:  Users
User: Settings
Logged In User : Messages
System Settings : System Settings
System Settings : System Events
Manager Customisation
Contexts
Dashboards : Dashboards
Dashboards : Widgets
Access Control Lists : Access Policies
Access Control Lists : Policy Templates
Lexicon Management
Namespaces
Elements:
Template : Template Variables
Template Variable : Template Access

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.settings.js
+++ b/manager/assets/modext/widgets/core/modx.grid.settings.js
@@ -73,7 +73,11 @@ MODx.grid.SettingsGrid = function(config) {
         ,cls: 'x-form-filter-clear'
         ,text: _('filter_clear')
         ,listeners: {
-            'click': {fn: this.clearFilter, scope: this}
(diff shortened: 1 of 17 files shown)


sonicpunk:
#1point to @theboxer

PR #13513, opened 2017-07-07
[Enhancement] Speedup the Feed-Dashboards

mindeffects:
### What does it do?
Speeds up the dashboard widgets for the rss feeds

### Why is it needed?
More speed and less overhead work if the feeds are turned off anyway.

### Related issue(s)/PR(s)
none

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -549,7 +549,7 @@ $settings['feed_modx_news']->fromArray(array (
 $settings['feed_modx_news_enabled']= $xpdo->newObject('modSystemSetting');
 $settings['feed_modx_news_enabled']->fromArray(array (
   'key' => 'feed_modx_news_enabled',
-  'value' => '1',
+  'value' => '0',
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'system',
@@ -567,7 +567,7 @@ $settings['feed_modx_security']->fromArray(array (
 $settings['feed_modx_security_enabled']= $xpdo->newObject('modSystemSetting');
 $settings['feed_modx_security_enabled']->fromArray(array (
   'key' => 'feed_modx_security_enabled',
-  'value' => '1',
+  'value' => '0',
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'system',
--- a/manager/controllers/default/dashboard/widget.modx-news.php
+++ b/manager/controllers/default/dashboard/widget.modx-news.php
@@ -17,17 +17,17 @@ class modDashboardWidgetNewsFeed extends modDashboardWidgetInterface {
      * @return string
      */
     public function render() {
-        $url = $this->modx->getOption('feed_modx_news');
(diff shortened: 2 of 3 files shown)


_build/data/transport.core.system_settings.php
```diff
@@ -567,7 +567,7 @@
 $settings['feed_modx_security_enabled']= $xpdo->newObject('modSystemSetting');
 $settings['feed_modx_security_enabled']->fromArray(array (
   'key' => 'feed_modx_security_enabled',
-  'value' => '1',
+  'value' => '0',
```
lawrenz1337: Why would you still leave default value as `0`. Your previous PR was closed because of this

_build/data/transport.core.system_settings.php
```diff
@@ -549,7 +549,7 @@
 $settings['feed_modx_news_enabled']= $xpdo->newObject('modSystemSetting');
 $settings['feed_modx_news_enabled']->fromArray(array (
   'key' => 'feed_modx_news_enabled',
-  'value' => '1',
+  'value' => '0',
```
theboxer: I think I rejected this change in previous PR :) Please change it back to `enabled` by default.

_build/data/transport.core.system_settings.php
theboxer: I think I rejected this change in previous PR :) Please change it back to `enabled` by default.

lawrenz1337 (changes requested):
I don't think that leaving news disabled by default is a correct decision

rtripault:
Besides the fact you are disabling the feeds as default, just wanted to point that this kind of optimization has already landed for 2.6 (see #13507). So not sure if we would want this in 2.5

_build/data/transport.core.system_settings.php
mindeffects: This is me and git only slowly getting friends. Sorry. :-(
Actually I went back from "off/0" to "on/1", so the file is as was before I made changes. I reverted.

lawrenz1337:
@rtripault I think he forgot to change the branch and had previous commit there with feeds turned off

theboxer:
@rtripault Yes, we do want them in 2.5.x, please cherry pick the change to this branch as well.

rtripault:
Well, the mentioned PR does more than just "speeding up", it loads the feeds using AJAX... i don't think that qualifies to be cherry-picked into 2.5

theboxer:
I'd say it's a bug that it doesn't load via AJAX :P

theboxer:
btw. just messing with you, leave it just there :)

rtripault:
Too late, already merged 2.x back into 2.5

theboxer:
well I hope you didn't `merge` the whole branch

rtripault:
Nah... was messing back with you... i'm not involved with 2.5 :raised_hands:

theboxer:
lol, well played :P

gpsietzema:
#modxbughunt #3points to @mindeffects

sonicpunk:

Issue #13250, opened 2017-01-12
Weird effects with nested items (cached, uncached)


pixelive:
### Summary
Weird effects with nested items. Although the use cases are a bit weird and probably give wrong results, it should not give broken results like it does now.

### Step to reproduce
I simplified the cases a bit.

**Case A**
Create a chunk called 'rowChunk' with the following content:
```
[[+rows]]
```

Create a chunk called 'equalChunk' with the following content:
```
[[!++site_url:notempty=`<p class="success">[[!++site_url]]</p>`]]
```

Put this in your template:
```
[[$rowChunk?
&rows=`[[$equalChunk]]`
]]
```

**Case B**
Create a chunk called 'rowChunk' with the following content:
```
[[+rows]]
```

Create a chunk called 'longChunk' with the following content:
```
[[!longSnippet? &value=`Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. 
Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. 
Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. 
Donec pede justo, fringilla vel, aliquet nec, vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. 
Nullam dictum felis eu pede mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus. 
Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis, feugiat a, tellus. 
Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet. Etiam ultricies nisi vel augue. Curabitur ullamcorper 
ultricies nisi. Nam eget dui. Etiam rhoncus. Maecenas tempus, tellus eget condimentum rhoncus, sem quam semper libero, sit amet adipiscing 
sem neque sed ipsum. Nam quam nunc, blandit vel, luctus pulvinar, hendrerit id, lorem. Maecenas nec odio et ante tincidunt tempus. 
Donec vitae sapien ut libero venenatis faucibus. Nullam quis ante. Etiam sit amet orci eget eros faucibus tincidunt. Duis leo. 
[...]

PR #13514, opened 2017-07-07
Fix weird effects with nested items #13250 #modxbughunt

flowan:
### What does it do?
Continue parsing tags if a nested uncached tag is parsed in a cached tag.

### Why is it needed?
Make sure all tags are processed.

### Related issue(s)/PR(s)
Related issue #13250

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -442,10 +442,10 @@ class modParser {
         $tokenOffset= 0;
         $cacheable= true;
         if ($token === '!') {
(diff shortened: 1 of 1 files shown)


opengeek:
This breaks the parser.

gpsietzema:
#modxbughunt #1point to @opengeek

flowan:
Ok, this change does not pass the tests. But the fix i submitted works. You can run the tests from https://github.com/modxcms/revolution/issues/13250 and see it works.

Issue #13459, opened 2017-05-24
failed_login_attempts setting doesn't work as you'd expect


Bryan-Lund:
### Summary
The failed_login_attempts setting in the CMS doesn't work quite as you'd expect - the number you give is not the amount of login attempts actually taken into account.

### Step to reproduce
Go into the settings in the MODx manager and update the failed_login_attempts setting (defaults to 5).

### Observed behavior
When attempting to log in once more than the amount set for the failed_login_attempts setting, you can still try to login. After checking this with different settings (3 and 2 times instead of the default of 5), I noticed that for setting it to 3, it would set your account to be blocked after the 6th attempt, setting it to 4 blocks your account after the 4th attempt (basically doubled depending on the setting you give). 

### Expected behavior
It should mark your account as blocked after the nth amount of trying to login (where n is the failed_login_attempts setting), instead of nth x 2.

### Environment
MODX version - 2.5.7
apache version 2.4.7
mysql version 15.1
browser - Chrome 58.0.3029.110 (not checked other browsers)

spectrum48k:
fixed in #13516 PR

rtripault:
PR merged, so closing.

Thanks

PR #13516, opened 2017-07-07
wrong failed login count fix #modxbughunt

spectrum48k:
### What does it do?
makes maximum login count attempts equal to system setting

### Why is it needed?
This is necessary because the user has twice as many attempts as set in the system setting.

### Related issue(s)/PR(s)
#13459

### DIFF
--- a/core/model/modx/processors/security/login.class.php
+++ b/core/model/modx/processors/security/login.class.php
@@ -240,17 +240,13 @@ class modSecurityLoginProcessor extends modProcessor {
      * Update failed login count
      */
     public function failedLogin() {
-        if (!array_key_exists('login_failed', $_SESSION)) {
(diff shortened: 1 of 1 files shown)


sonicpunk:
#modxbughunt #3points to @spectrum48k

Flowan:
Tested PR and can confirm this works! #modxbughunt

gpsietzema:
#modxbughunt #1point to @Flowan

rtripault:
Tested too and working as advertised, thanks.

@spectrum48k did you already sign the [CLA](https://develop.modx.com/contribute/cla/) ?
If so, what's your username on modx.com, if not, would you mind doing so ?

Thanks

spectrum48k:
*del*

Issue #12459, opened 2015-05-31
svg preview in file manager


Pfeifakopf:
Please add the svg support in the file manager.

natedin:
+1 would like to see this added

frischnetz:
+1 (imageTV preview for SVG is also broken)

pixelive:
I'm going to try and fix this today.

RinzeVanDerBrug:
Tested this in MODX 2.5.5.

It IS possible to add a .SVG file, but you are not able to view a thumbnail. #modxbughunt

udovichenko:
+1 Looking forward to this!

mazyone:
+1 would like to see this added

OptimusCrime:
This is an interesting issue as thumbnails are generated though phpThumb, which (as far as I know) does not handle .SVG files. SVG files are just vector data. Perhaps it could be possible to have a special connector handler for just this file type. It would be interesting to look closer at this.

frischnetz:
A first approach could be to just skip thumbnail generation for SVG-files and show the files directly.

OptimusCrime:
I am not too familiar with the SVG filetype, but how does that define image size? Can you (with HTML) display a SVG and scale it accordingly? Like defining a maximum height and/or width and everything happens by itself, or does it require more "magic" than that?

frischnetz:
It's complicated :-) There are options inside the SVG to set a size and viewbox and aspect-ratios. And you can use SVGs as imgs, as objects and inline-code and iframe and background-image. But usually it works for most files to use it just in an img-tag with a fixed size. Everything is better than no image at all.

https://css-tricks.com/scale-svg/

alroniks:
It seems that all related PRs merged and feature already released. So closing it.

PR #13517, opened 2017-07-07
Fix for displaying SVG in media browser

marcjenkins:
### What does it do?
Displays SVGs in the media browser.

### Why is it needed?
SVG's weren't being displayed as they were being processed by PHPThumb.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/12459

### DIFF
--- a/_build/templates/default/sass/_browser.scss
+++ b/_build/templates/default/sass/_browser.scss
@@ -65,6 +65,7 @@
       width: auto;
       max-width: 100%;
       height: auto;
+      max-height: 100%;
     }
   .modx-browser-thumb-wrap span {
     display: block;
--- a/_build/templates/default/sass/_browser.scss
+++ b/_build/templates/default/sass/_browser.scss
@@ -60,6 +60,7 @@
   }
     .modx-browser-thumb img {
       vertical-align: middle;
+      max-height: 80px;
     }
   .modx-browser-thumb-wrap span {
     display: block;
--- a/_build/templates/default/sass/_browser.scss
+++ b/_build/templates/default/sass/_browser.scss
@@ -60,7 +60,11 @@
   }
     .modx-browser-thumb img {
       vertical-align: middle;
-      max-height: 80px;
+    }
+    .modx-browser-thumb img[src$=".svg"] {
+      width: auto;
+      max-width: 100%;
+      height: auto;
     }
   .modx-browser-thumb-wrap span {
     display: block;
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -252,7 +252,11 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
                             'source' => $this->get('id'),
                         ));
 
-                        $image = $this->ctx->getOption('connectors_url', MODX_CONNECTORS_URL).'system/phpthumb.php?'.urldecode($imageQuery);
+                        if ($ext == 'svg') {
+                            $image = $bases['urlAbsolute'] . urldecode($url);
+                        } else {
+                            $image = $this->ctx->getOption('connectors_url', MODX_CONNECTORS_URL).'system/phpthumb.php?'.urldecode($imageQuery);
+                        }
 
                         $files[$fileName]['qtip'] = '<img src="'.$image.'" alt="'.$fileName.'" />';
 
@@ -961,7 +965,7 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
         $modAuth = $this->xpdo->user->getUserToken($this->xpdo->context->get('key'));
 
         /* get default settings */
-        $imageExtensions = $this->getOption('imageExtensions',$properties,'jpg,jpeg,png,gif');
+        $imageExtensions = $this->getOption('imageExtensions',$properties,'jpg,jpeg,png,gif,svg');
         $imageExtensions = explode(',',$imageExtensions);
         $use_multibyte = $this->ctx->getOption('use_multibyte', false);
         $encoding = $this->ctx->getOption('modx_charset', 'UTF-8');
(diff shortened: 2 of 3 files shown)


sonicpunk:
#modxbughunt #3points @marcjenkins

rtripault:
Thanks for that (not tested yet).
I believe this should go against 2.x branch instead.

Also i guess the default `imageExtensions` propertiess should be updated too in relevant media source classes

gpsietzema:
Just tested this @marcjenkins - It does show a preview now, which is awesome! But is it possible to have the SVG fit within the thumbnail area. It looks a bit weird now:


screen shot 2017-07-07 at 17 30 20

rtripault:
Just tested and we indeed see some svg preview!
However there appears to be no resizing. In my case, the svg was wider than the thumbnail width, and i only get a small portion of the svg

sonicpunk:

sonicpunk:
#modxbughunt #1point to @gpsietzema

OptimusCrime:
Wow, really nice PR! I looked at this once, and I found that handling SVG resizing was far more complex than I initially thought.

labr1005:
I would either need some help with Git or you just add this to _browser.scss

```css
.modx-browser-thumb img[src$=".svg"] {
    width: auto;
    max-width: 100%;
    height: auto;
}
```
bildschirmfoto 2017-07-08 um 12 25 16

OptimusCrime:
Oh. Is it really that simple? 😮 

@marcjenkins Could you put this code in, or do you want me to PR your fork?

rtripault:
Also, as first time contributor, would you mind signing the [CLA](https://develop.modx.com/contribute/cla/) ?

Many thanks!

labr1005:
Do we care about IE and if so, from which version on up? The problem with svg in general is, that the width and height attributes are optional. So it may look distorted in IE.

labr1005:
@marcjenkins I'd thought that you would leave max-height: 80px; in place...

labr1005:
@OptimusCrime Well, that's the best thing you can do with pure CSS.

And we're not done here, because then we'll also need to take care of TV previews for example.

bildschirmfoto 2017-07-08 um 17 03 03

The best solution for browsers would be to convert the Svg to Png for the previews. Some browsers still like pixels more than something that has "no size". I think that conversion can be done with imagick.

Another option would be add width and height attributes to the previews img tags in the manager (and then do some CSS adjustments). I'm currently looking into JP's mediamanager and how he extracts the width and height of Svgs...

Mark-H:
Not sure if converting SVGs to a bitmap format is really the best solution. SVGs have a pretty solid 97.67% support, so that's not really the problem... phpthumb just doesn't know how to handle them, so the TV previews will need to use the raw SVG source instead of a phpthumb call.

labr1005:
@Mark-H I agree. "best solution" was the wrong wording and I was talking about the thumbnails only. The problem: It's basically impossible to proportionally scale and center a svg using an img tag into a container that has a fixed width and height with pure CSS... in IE. :) The other browsers figure out what to do, of course. But I'm currently working on that.

labr1005:
Here's my current status - working also in IE as you can see - and I look forward to your opinion and have some questions.

bildschirmfoto 2017-07-09 um 20 31 22

1.) What do you think of the checked background for transparent images?

2.) Until now the thumbnails and the preview did not reflect the image's original proportion - now they would. Do you agree?

(Kinda hard to describe in not my mother language. In the example above the apple preview would have more space to the left and right and it would look the the original image is landscape and not portrait.)

3.) In modfilemediasource.class.php and mods3mediasource.class.php I find some calls to get the system settings "filemanager_image_width", "filemanager_image_height", "filemanager_thumb_width" and "filemanager_thumb_height".

But these system settings are not set on a fresh install. And looking through the repository it looks to me that these are the only files where they are tried to be read, correct?

Besides, the width and height for the thumb for example are fixed set to 100px x 80px in the CSS…

Do you know of anybody who's ever altered these settings? IMO they don't make any sense for the preview anyway. For the thumbnail, ok - could make sense.

Option 1: Get rid of the system settings.

Option 2: Make the system settings really work (for the thumbnails).

OptimusCrime:
That looks really, really good!

Mark-H:
Any objections against this PR getting merged into 2.5.x so we get SVGs in the tree, and worry about the TVs in a next one? 

@labr1005 some of your questions are probably better in a new issue... unless you are planning to send a PR with some of your improvements?

OptimusCrime:
No objections here.

labr1005:
Objections if you care for IE and Edge. But you can merge this now and I'll send a PR in a couple of days.

Mark-H:
Looking forward to your PR then @labr1005 ;)

opengeek:
Is this good to be cherry-picked into 2.x for 2.6 release?

gpsietzema:
I think so!

rtripault:
@labr1005 does your PR (#13547) applies on top of this one, or would you need to send a new PR to apply the IE/Edge fixes ?

OptimusCrime:
I can only reference what is already posted, but he said:

>Here's my current status - working also in IE as you can see - and I look forward to your opinion and have some questions.

rtripault:
Thanks.

My understanding is that the other PR satisfies the same purpose as this one, with some additions (being IE/Edge compliant) and is not something that could apply on top of this one.

So i'd be in favor of using the other PR, albeit it would be a shame not to merge this once, since Mark spent quite some time answering our concerns/feedback.

If the other PR could be used on top of this one, then no issue (or if Lars is open to submit a new PR only with his additions).

Hoping that makes sense

OptimusCrime:
I agree. Can @marcjenkins shed some light on what is left to do/not satisfied with this PR as it stand right now? I would love to see this feature in the next 2.x version of MODX.

labr1005:
I will have to make some small adjustments to my PR - but I'm very fine with that.

So if you ask me: Please merge this one first.

labr1005:
My new PR is ready for review.

opengeek:
Merged into 2.x as 9e73dc474

Issue #9751, opened 2013-03-29
Hardcoded HTTP/1.1 protocol on nginx return serives pages with wrong header


ebuster:
ebuster created Redmine issue ID 9751

When we get Error 404 (for example),  modx always return HTTP status header with HTTP/1.1 protocol (it hardcoded).
But when nginx works with backend in protocol HTTP/1.0, instead of error page from modx we get corrupted unreadable content (i think it not ungziped).
When modx return the same http version header as in request then all works fine. 

Attached patch fix this problem.

adamwintle:
@ebuster do you still have the fix for this? @rtripault Do you know if this bug is still present?

ebuster:
@adamwintle, sorry but not. Сoincidentally, I removed them a few weeks ago. 

As I remember, I'd just replaced all code like this
```
header('HTTP/1.1 503 Service Unavailable');
```
from https://github.com/modxcms/revolution/blob/2.x/index.php#L42
to something like this
```
header($_SERVER['SERVER_PROTOCOL'] . ' 503 Service Unavailable');
```
And that's it.

OptimusCrime:
This should be doable

PR #13518, opened 2017-07-07
Replace all hardcoded http versions by $_SERVER['SERVER_PROTOCOL'] va…

alroniks:
### What does it do?
It changes all headers and replaces hardcoded HTTP protocol versions to an actual version that comes from the webserver. 

### Why is it needed?
Webserver can request pages using different http protocol version (1.0, 1.1, 2.0) and backend (MODX) should answer with the same protocol version.

### Related issue(s)/PR(s)
#9751
#modxbughunt

### DIFF
--- a/connectors/index.php
+++ b/connectors/index.php
@@ -30,7 +30,7 @@ if (!defined('MODX_CORE_PATH')) {
 /* include modX class - return error on failure */
 if (!include_once(MODX_CORE_PATH . 'model/modx/modx.class.php')) {
     header("Content-Type: application/json; charset=UTF-8");
-    header('HTTP/1.1 404 Not Found');
+    header($_SERVER['SERVER_PROTOCOL'] . ' 404 Not Found');
     echo json_encode(array(
         'success' => false,
         'code' => 404,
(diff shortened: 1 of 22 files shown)


sonicpunk:

Flowan:
Tested this and can confirm this works! #modxbughunt

Mark-H:
In case anyone wondered about the security of the protocol version with escaping, looks like its reliable as enforced by the web server https://stackoverflow.com/a/6474936

gpsietzema:
#modxbughunt #1point to @Flowan

rtripault:
I did not review all the changed files, but having concerns about accessing `$_SERVER['SERVER_PROTOCOL']` when using the CLI

Mark-H:
Valid concern, but the changes seem to mostly be in request/response/fatal error handlers.. while it won't be available from the CLI, I doubt these would get called from the CLI? 

Adding an isset/array_key_exists could be an option to be safe, but as these are all setting headers that don't have meaning on the CLI anyway, I think we'll be fine without it...

Alroniks:
My main concern about inconsistency in RFC of HTML that could be unknown. In a common way, an application should answer by the same protocol version but I am not sure that it will work always and fine. For CLI yes, it doesn't matter, headers work only with webservers.

sdrenth:
@Alroniks I've found that in /core/model/phpthumb/ some http versions are still hardcoded:

* Line 538
* Line 559
* Line 645

OptimusCrime:
@sdrenth That is an external project. We are unable to rewrite the code there. It would get overwritten when we upgrade PhpThumb to a newer version. I think we have to ignore those occurrences.

sdrenth:
@OptimusCrime I understand, you may ignore my comment ;)

OptimusCrime:
Great work btw!

Issue #13511, opened 2017-07-07
Access control page view is broken #modxbughunt


lawrenz1337:
### Summary
When I pull the side slider it collapses the wrong way. Also the button is under the slider

### Step to reproduce
![](http://joxi.net/VrwnpZncKOKDVA.jpg)
Open Access control page and pull the side slider.

### Observed behavior
It collapsed to the right

### Expected behavior
It should move the slider accordingly to my mouse movements

### Environment
MODx 2.5.5

Finetuned:
"I'm going to try and fix this today."

sonicpunk:
#modxbughunt #1point

pixelchutes:
Fixed by #13555

PR #13520, opened 2017-07-07
Hide user group tree panel splitbar if center panel is hidden [#13511]

Finetuned:
### What does it do?
hide the User Group tree panel splitbar when the centre panel is hidden (no group selected or group is empty).

### Why is it needed?
This circumvents a ui bug which incrementally decreased the size of the split panel if the centre region was hidden. Each time the user clicked on the splitbar, it jumped further to the left obscuring the tree and could not be dragged out to the right.

### Related issues/PRs
#13511

### DIFF
--- a/manager/assets/modext/widgets/security/modx.panel.groups.roles.js
+++ b/manager/assets/modext/widgets/security/modx.panel.groups.roles.js
@@ -99,10 +99,6 @@ Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
                     ]
                 }]
             });
-
-
-
-
         }
         if (MODx.perm.view_role == 1) {
             tbs.push({
--- a/manager/assets/modext/widgets/security/modx.panel.groups.roles.js
+++ b/manager/assets/modext/widgets/security/modx.panel.groups.roles.js
@@ -27,10 +27,21 @@ MODx.panel.GroupsRoles = function(config) {
         })]
     });
     MODx.panel.GroupsRoles.superclass.constructor.call(this,config);
-    
-    Ext.getCmp('modx-tree-usergroup').on('expandnode', this.fixPanelHeight);
-    Ext.getCmp('modx-tree-usergroup').on('collapsenode', this.fixPanelHeight);
-    
+
+    var west, usergroupTree = Ext.getCmp('modx-tree-usergroup');
+
+    usergroupTree.on('expandnode', this.fixPanelHeight);
+    usergroupTree.on('collapsenode', this.fixPanelHeight);
+
+    usergroupTree.addListener({
+        resize : function(cmp) {
+            var centre = Ext.getCmp('modx-usergroup-users');
+            if (centre.hidden){
+                cmp.ownerCt.ownerCt.layout.west.getSplitBar().el.hide();
+            }
+        }
+    });
+
     if (MODx.perm.usergroup_user_list == 1) {
         Ext.getCmp('modx-tree-usergroup').on('click', function(node,e){
             this.getUsers(node);
@@ -69,6 +80,7 @@ Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
                             ,useSplitTips: true
                             ,monitorResize: true
                             ,width: 270
+                            ,minWidth: 270
                             ,minSize: 270
                             ,maxSize: 400
                             ,layout: 'fit'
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/security/modx.panel.groups.roles.js
```diff
@@ -146,11 +162,14 @@ Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
         var usergroup = id.replace('n_ug_', '') - 0; // typecasting
         
         var userGrid = Ext.getCmp('modx-usergroup-users');
+        var westPanel = Ext.getCmp('modx-tree-usergroup').ownerCt.ownerCt.layout.west;
```
rtripault: I guess `Ext.getCmp('modx-leftbar-tabs`) might be "safer" in case the structure changes

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
```diff
@@ -27,10 +27,21 @@ MODx.panel.GroupsRoles = function(config) {
...
+    var west, usergroupTree = Ext.getCmp('modx-tree-usergroup');
+
+    usergroupTree.on('expandnode', this.fixPanelHeight);
+    usergroupTree.on('collapsenode', this.fixPanelHeight);
+
+    usergroupTree.addListener({
+        resize : function(cmp) {
+            var centre = Ext.getCmp('modx-usergroup-users');
+            if (centre.hidden){
+                cmp.ownerCt.ownerCt.layout.west.getSplitBar().el.hide();
```
rtripault: Here also, getting `modx-leftbar-tabs` might be "safer"

sonicpunk:

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
Finetuned: Yes, it's 'modx-tree-panel-usergroup'

gpsietzema:

Mark-H:
@Finetuned: for future reference, it would be great if you could make sure that your pull request title is a bit more descriptive, and your commit messages too. That makes it easier for integrators to quickly see what a PR is about, and the history is more useful if it describes the change instead of only referring to an ID.

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
```diff
@@ -87,6 +99,10 @@ Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
                     ]
                 }]
             });
+
```
Alroniks: Not sure that this empty lines required here.

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
```diff
@@ -69,6 +80,7 @@ Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
                             ,useSplitTips: true
                             ,monitorResize: true
                             ,width: 270
+                            ,minWidth: 270
                             ,minSize: 270
                             ,maxSize: 400
```
Alroniks: I know that it was here before but could it be increased to half of width of view? Not important, just interesting.

Finetuned: It makes sense to me to have the tree at the smallest usable size. The user then expands the view as needed when they drill down into the tree. If you only have one or two levels, it makes no sense to take half the available view width as the rest would not often be utilised.

Alroniks:
And here is also an issue with i18n (on Russian button "New user group" is very wide and part of button covered by right panel). As a solution may it make sense to set a minimum width of left panel depends on button width.

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
Alroniks: Yep, it's clear. My question about `maxSize: 400` :)

Finetuned: Agreed but it feels like a separate issue to the bug fix. Maybe split this into a new issue?

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
Finetuned: I've removed these

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
Alroniks: Maybe yes, before it didn't work at all so I saw it only when tried to test this changes.

Alroniks:
Also cherry-picked to 2.6 https://github.com/modxcms/revolution/commit/9a32805a3b30c47bca87c74b089aea8cf470f003

Issue #13403, opened 2017-03-29
Error message attributed to wrong context


fountainheadllc:
### Summary
Attempted to duplicate a context into a new context that already existed and error message popped up that context already exists, but was assigned to the old key not the new key.

### Step to reproduce
Attempt to copy context to an existing alternative context, error message will reproduce in pop-up.

### Observed behavior
Error message in red displays for Old Key not New Key

### Expected behavior
Error message in red should display for New Key

### Environment
MODX 2.5.5, MODXCloud nginx and php 5.6.x
![screen shot 2017-03-29 at 9 22 21 am](https://cloud.githubusercontent.com/assets/26174817/24457137/4d45faa8-1463-11e7-88b1-a2dac54984d3.jpg)

Alroniks:
Fixed during the #modxbughunt

Alroniks:
Merged, so closing.

PR #13521, opened 2017-07-07
Fixed correct field attaching of duplication error message

alroniks:
### What does it do?
It changes allow to declare the name of the field with a new value of an object in duplication windows that allows attaching error messages to correct fields. 

### Why is it needed?
It fixes wrong error message attaching during the context duplication.

### Related issue(s)/PR(s)
#13403
#modxbughunt


sonicpunk:

opengeek:
As this changes the modObjectDuplicateProcessor, this needs to go into 2.6 as it introduces a new feature to a base class, which if utilized in an extra would not work in all 2.5.x releases.

opengeek:
Integrated into 2.x branch via cherry-picking as 083edef2a.

Issue #11269, opened 2014-04-09
if-then-else output filters break when @ is used in property text


frischnetz:
When one uses @ inside the text of a tag property inside of an if-then-else-clause, the whole statement breaks and neither the then- nor the else-part is evaluated.

Minimal example: 
Consider the two statements: (use in [[*content]] with minimal template): 

```
A[[*pagetitle:isnot=`Y`:then=`C[[$Z?&mail=`ab.de`]]E`:else=`F`]]B
A[[*pagetitle:isnot=`Y`:then=`C[[$Z?&mail=`ab@de`]]E`:else=`F`]]B
```

I sprinkled some letters for clarity of the output.

No matter what, they should produce identical output (ACEB). But in the second line the whole if-then-else-expession discarded by the renderer, so it only gives AB.

The only difference between the two lines is the @ inside the property. And it does not matter, if the whole then-statement is wrapped inside a chunk, it still breaks:

```
<p>A[[*pagetitle:isnot=`Y`:then=`[[$TEST]]`:else=`F`]]B</p>
where
[[$TEST]] = C[[$Z?&mail=`ab@de`]]E
```

This bug dates back at least till 2.2.4, that is the oldest Revo version I  tested.

A possible workaround is using the [[If]]-Snippet. I found no problems here.

opengeek:
@inreti — I tried your proposed fix but I cannot get the conditional logic to return the `else`, only the `then` part of the expression, regardless of the logic.

inreti:
Started again working on it.

netProphET:
Re-confirmed in 2.5.7. I'm going to give this a shot.

gernotebenlechner:
Checking this issue at #modxbughunt if it is still valid with the suggested code (Chunk & Snippet):

```[[*id:gt=`0`:then=`[[AddCharBug? &param=`@`]]`:else=``]]```

as well as 

```[[*id:gt=`0`:then=`[[AddCharBug? &param=`abc@xyz.com`]]`:else=``]]```

as well as 

```[[*id:gt=`0`:then=`[[AddCharBug? &param=`@EVAL`]]`:else=``]]```

just to make sure.

Result: On a fresh 2.5.8-dev install I **could NOT REPRODUCE** if-then-else output filters break when `@` is used in property text.

**Comment by gernotebenlechner** [comment]
I've used this small snippet "AddCharBug" for testing:

```
<?php
$output = '<p>'.$param.'</p>';
return $output;
```

[...]

PR #13522, opened 2017-07-07
Fixed bug 11269 with wrong parsing if at-sign inside output filter #modxbughunt

wuuti:
### What does it do?
Fixes the wrong parser behaviour if an @ is used inside output filter.

### Why is it needed?
Issues #11269 and (closed) #12159. Parsers xpdo call to escSplit works not as proposed, the string is split inside escaped parts, which causes the bug.

The split of the string to find the property set is now managed by a preg_match regex call. The benefit is, that this also replaces the second split there, which only splits PropertySet:outputFilters. Might be even quicker at all.

### Remarks
Besides the fact, that the regex imho is the more elegant solution, this seems to be a bug in the xpdo->escSplit method. The method is intended _not_ to go into escaped content, but it does here. Therefore, additionally further tests with xpdo might have to be done.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -1065,7 +1065,7 @@ abstract class modTag {
         $name = $this->get('name');
         if (strpos($name, '@') !== false) {
 
-            $tagChars = '[\w\d\-\.\/\x{a1}-\x{2af}]+';
+            $tagChars = '[\s\w\d\-\.\/\x{a1}-\x{2af}]+';
             $psChars = $tagChars;   // there might be different rules for ps names?
             preg_match('/(?<tag>'.$tagChars.')(@(?<psName>'.$psChars.'))?(\:(?<rest>.*))?/',$name, $split);
 
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -1064,10 +1064,12 @@ abstract class modTag {
         $propertySet= null;
         $name = $this->get('name');
         if (strpos($name, '@') !== false) {
+
+            $tagChars = '[\w\d\-\.\/\x{a1}-\x{2af}]+';
+            $psChars = $tagChars;   // there might be different rules for ps names?
+            preg_match('/(?<tag>'.$tagChars.')(@(?<psName>'.$psChars.'))?(\:(?<rest>.*))?/',$name, $split);
+
             $psName= '';
-
-            preg_match('/(?<tag>[\w]*)(@(?<psName>\w*))?(\:(?<rest>.*))?/',$name, $split);
-
             if (isset($split['psName'])) {
                 $name = $split['tag'];
                 $psName = $split['psName'];
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -1063,29 +1063,17 @@ abstract class modTag {
     public function getPropertySet($setName = null) {
         $propertySet= null;
         $name = $this->get('name');
-
-        if (strpos($name, '@') !== false) { // <!-- this is useless as it triggers even if an @ is inside `content`. But should be quicker and reduces unneccessary preg matches...
+        if (strpos($name, '@') !== false) {
             $psName= '';
-
-            // the escSplit seems to fail, it steps down into `@bla` parts, according to the
-            // xpdo docs it should not! so this is a bug in XPDO.
-            // nevertheless, we don't need the (slow?) escSplit here, we can work with regexes,
-            // as we are only interested in the property set!
-            $split= xPDO :: escSplit('@', $name);
 
             preg_match('/(?<tag>[\w]*)(@(?<psName>\w*))?(\:(?<rest>.*))?/',$name, $split);
 
-            // we only need to process further if we have found a property set for the tag
             if (isset($split['psName'])) {
                 $name = $split['tag'];
                 $psName = $split['psName'];
-
-                // we already have the rest, no need to further split anything
                 if (isset($split['rest'])) {
-                    // build the new name without the property set
                     $name.= ':' . $split['rest'];
                 }
-
                 $this->set('name', $name);
             }
             if (!empty($psName)) {
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -1063,17 +1063,29 @@ abstract class modTag {
     public function getPropertySet($setName = null) {
         $propertySet= null;
         $name = $this->get('name');
-        if (strpos($name, '@') !== false) {
+
+        if (strpos($name, '@') !== false) { // <!-- this is useless as it triggers even if an @ is inside `content`. But should be quicker and reduces unneccessary preg matches...
             $psName= '';
+
+            // the escSplit seems to fail, it steps down into `@bla` parts, according to the
+            // xpdo docs it should not! so this is a bug in XPDO.
+            // nevertheless, we don't need the (slow?) escSplit here, we can work with regexes,
+            // as we are only interested in the property set!
             $split= xPDO :: escSplit('@', $name);
-            if ($split && isset($split[1])) {
-                $name= $split[0];
-                $psName= $split[1];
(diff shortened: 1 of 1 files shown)


sonicpunk:

sonicpunk:

OptimusCrime:
For your information: The `@` is an _at_ sign, and `+` is an _add_  sign. Other than that, looks good :)

sdrenth:
Tested and can confirm this works!

opengeek:
I'm reviewing this and the `xPDO::escSplit()` method to see if there is a better solution.

opengeek:
FWIW, the problem in `xPDO::escSplit()` is due to the embedded escape characters. I'm really torn on adopting this solution without more exhaustive tests that prove this works as expected with all existing tag usage. The alternative is to document that you should never use embedded escape characters. For instance, the tag sequence:

```A[[*pagetitle:isnot=`Y`:then=`C[[$Z?&mail=`ab@de`]]E`:else=`F`]]B```

works fine without the embedded escape characters, e.g.:

```A[[*pagetitle:isnot=`Y`:then=`C[[$Z?&mail=ab@de]]E`:else=`F`]]B```

This tag sequence is violating recommended usage of conditionals already: the tag sequence in the `then` is *always* evaluated regardless if the condition is matched or not. But I would like to consider any dissenting opinions before making a final decision to reject or accept this change.

**Comment by pixelchutes** [comment]
I'd be interested to know if the introduction of the regex has any negative performance impact?

I realize this could be tricky to quantify / vary depending on use case, but I guess it probably depends on the frequency of `getPropertySet()` gets called. If somehow performance is _improved_, then no objections from me.

I will note that I've bumped into this particular issue before, and solved it the same way Jason is recommending (removing the double-nested backticks: ``` ` ```)

**Comment by OptimusCrime** [comment]
Regex-calls are somewhat expensive, so I guess this increases the parsing time. However, I think the increase in total is marginal, especially when you compare it against database calls.

I am all in favor of this implementation, as the other options (add exceptions to the syntax) makes everything more confusing that it already is. Also, I am confused why this seemingly utility method is placed in the ORM code, and not the MODX class itself. I searched the repository and it is used once in xPDO and twice in MODX only. It could be better to extract the method and place it in the MODX code instead. Or, dont use it at all.

Too bad we dont have a decent unit test suit that could test this implementation for us.

**Comment by wuuti** [comment]
@pixelchutes @OptimusCrime I just wrote a small (and surely silly) benchmark script, which just repeats the splitting operation with both variants a million times. Turns out that the regex variant is factor 7.5 quicker (!) than the escSplit method in this case (I only tested the example from opengeek). But to be honest, thats no wonder: a regex is not expensive per se, and compared to what is done here (split the string at the first @ before a colon), the escSplit method is like taking a sledgehammer to crack a nut (as far as I understand whats going on in that method). 

I wonder if it would be possible to generically replace the escSplit with such a simple regex...

**Comment by opengeek** [comment]
@wuuti Replacing the `xPDO::escSplit()` implementation with a solution that works in all cases would be my preference.

**Comment by OptimusCrime** [comment]
I guess that would be best. I have no idea how this method works or what it is supposed to do, so I guess I am of no help here.

**Comment by wuuti** [comment]
Who implemented that method? Anyone here ready for an archaeological research? 

I volunteer for some more work on this. Besides, this PR might be merged in the meantime, as it solves a special edge case of the use of the escSplit method where an issue occurs.

**Comment by OptimusCrime** [comment]
I agree. And I guess it was originally written by @opengeek (?)

**Comment by opengeek** [comment]
I think I have found some issues with this regex. For instance, a dash (`-`) character in the element name seems to break this. I'm researching if there is an easy way to fix the expression.

**Comment by wuuti** [comment]
@opengeek That should be quite easy to solve. The \w* just has to be replaced with [\w\-]*, possibly enhanced with all other characters being allowed for element names and for set names. 

Do we have any official restriction (also in the docs) on the allowed characters? I know there are some other issues dangling around with that (e.g. the #13381). I fear there are some other places in the code where some kind of checks for that occur (creating a new element?), and we should have a central pattern for that (constant?) which all other places consume...

**Comment by OptimusCrime** [comment]
I have kind of concluded for issue #13381 that it is a wontfix and such "hacking" is beyond the scope of what we are able to handle during parsing. I see no reason to take it into account in the regex posed above.

**Comment by opengeek** [comment]
There is regex for validation of element names in the database schema, e.g.

`/^(?!\\s)[a-zA-Z0-9\\x2d-\\x2f\\x7f-\\xff-_\\s]+(?!\\s)$/`

**Comment by wuuti** [comment]
@opengeek Thats an interesting regex. Online tools like https://www.debuggex.com do show errors for this regex.

**Comment by OptimusCrime** [comment]
Regex101 also reports errors about the ordering of the character ranges. Are we sure this regex is correct?

**Comment by wuuti** [comment]
Well, it is correct without the double backslashes for the online services: 

`/^(?!\s)[a-zA-Z0-9\x2d-\x2f\x7f-\xff-_\s]+(?!\s)$/`

The regex will also include spaces in the element name. _Is that what we want here???_ Without the spaces, simplified and internationalized it _could_ be:

`^[\w\d\-\.\/\x{a1}-\x{2af}]+$`

Which will include more letters, remove controls and remove the negative lookaheads for whitespace. This validates crude names like `abc./DEʬD3-¸_df¶` (note, it's not a comma , but U+00B8 "cedilla").  And it might be better in performance here.

**Comment by OptimusCrime** [comment]
Might be that the double backslashes are only there as it needs to escape the backslashes in strings in PHP/JS.

**Comment by wuuti** [comment]
Mh... as the pattern is noted in single quotes in the map, there is no need to escape anything but a ' or \

**Comment by OptimusCrime** [comment]
You're right. Then I don't know.

https://3v4l.org/k9RRV

**Review by bezumkin — APPROVED** [review]
Tested in production on modx.pro - no problems found

**Review by opengeek — CHANGES_REQUESTED** [review]
Unless this is going to be updated to match the regex for valid element names, this is not going to be merged.

**Comment by Jako** [comment]
@wuuti Could you somehow add that valid element names regex or help us otherwise to get this merged?

**Comment by JoshuaLuckers** [comment]
@wuuti is it possible to help us getting this merged?

**Comment by wuuti** [comment]
I have added my proposed regex for element names in the code. It matches the same names as in db validation, but also more (unicode) characters. We don't necessarily have to change the validation there, but it might be a good idea to have a central place where the allowed tag name regex is defined and then used by this code _and_ by the db validation (and possible other places). But that would be a refactoring and is imho beyond the scope of this bug fixing PR. 

Please let me know what else you expect here @JoshuaLuckers

**Comment by JoshuaLuckers** [comment]
@opengeek are the changes and arguments enough to get this merged?

**Comment by opengeek** [comment]
I don't understand why:
> It matches the same names as in db validation, but also more (unicode) characters.

I had also asked this to be worked into the escSplit method in xPDO instead of bypassing it. I'm a little hesitant to integrate this as is. This really needs some additional testing and exploration before I'll feel comfortable with it. But if there are enough confirmation from others using this change on large-scale sites with no problems, I'd be okay with moving forward as is for now.

**Comment by wuuti** [comment]
Well, imho there is no need to use escsplit method here. The code only needs to check if there is one propertyset being used for the current element. Besides the fact that the escsplit is failing here, the regex should be much more efficient for this particular use case, and its more transparent what is happening compared to the escsplit which is more like a blackbox...

**Comment by JoshuaLuckers** [comment]
Adding a Unit Test for it would be a good way to test it.

**Comment by opengeek** [comment]
This needs to be rebased or applied to the 3.x branch to be considered still. Closing PR.

**Comment by Jako** [comment]
Lately I tried to use the following template chunk part, that won't work because of this issue:

```
[[[[+resource_id:notempty=`getResources? &resources=`[[+resource_id]]` &parents=`-1` &tpl=`@INLINE Resource ID: [[+id]]``:default=``]]]]
```

I hope this is a strong example for adopting this patch into 3.x

ghost:
I found that in complex usage like this it was almost always a problem if @INLINE was used, the solution being for a chunk to be used for the tpl.

Issue #12430, opened 2015-04-30
MODX strips newline from end of static files upon saving


pixelchutes:
### Summary

When editing a static resource within MODX, it strips the newline from the end of the static file upon saving.

I believe this behavior has been around for a very long time, including in Evo. I'm sure there's a logical reasoning for the stripping that takes place, but perhaps it could allow for a single `\n` at the end if one already exists? _(for static files)_
### Step to reproduce
1. Create a PHP file that contains a newline `\n` at the very end of the file
   
   ``` php
   <?php
   // Newline test snippet
   return 'hello world';
   
   // Include a newline (\n) at the very end
   
   
   ```
2. Map it to a new Static Snippet within MODX, and save.
3. Upon refresh, you'll see the snippet loaded with trailing `\n` stripped! _(Already stripped from the source file, too)_
4. Make any change to the static file, such as adding a newline to the end again, then save.
### Observed behavior

If you now perform a `diff` of your changes, you'll see the EOF newline has been removed.

When dealing with version controlled, static files where a change is made within the MODX Manager, this infamous message accompanies any subsequent commit:

> No newline at end of file.
### Expected behavior

Ideally, MODX would not strip the EOF newline, it may be there for a reason. Many believe that having a newline `\n` at the end of a file is considered best practice / ideal.
### Environment

MODX 2.3.4-dev, NGiNX, PHP 5

zaigham:
Happens in 2.2x, 2.3x too.

Flowan:
Confirmed in 2.5.x I'm going to try and fix this today.

gpsietzema:
#modxbughunt #1point to @Flowan

Alroniks:
https://github.com/modxcms/revolution/pull/13523 merged

pixelchutes:
Re-opening since #13576 was merged to fix the regression, but now no longer solving this. Low priority.

alroniks:
The issue in 3.x as well.

PR #13523, opened 2017-07-07
Prevent new lines from being stripped in snippets and plugins #12430 #modxbughunt

flowan:
### What does it do?
Remove the trim functions when saving or loading snippets and plugins.

### Why is it needed?
When dealing with version controlled, static files are changed in modx manager and will result in unneeded changed files.

### Related issue(s)/PR(s)
Related issue #12430

### DIFF
--- a/core/model/modx/modscript.class.php
+++ b/core/model/modx/modscript.class.php
@@ -37,7 +37,7 @@ class modScript extends modElement {
      * {@inheritdoc}
      */
     public function set($k, $v= null, $vType= '') {
-        if (in_array($k, array('snippet', 'plugincode', 'content'))) {
+        if (in_array($k, array('content'))) {
             $v= trim($v);
             if (strncmp($v, '<?', 2) == 0) {
                 $v= substr($v, 2);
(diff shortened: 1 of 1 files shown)


sonicpunk:

OptimusCrime:
Can you explain why you included this line:

```
if (in_array($k, array('snippet', 'plugincode', 'content'))) {
```

?

Flowan:
@OptimusCrime I removed snippet and plugincode from the array. Because when you create a new snippet or plugin this function is called and strips the new lines.

OptimusCrime:
Ah, okey. Yeah, I see the logic now. Sorry about that. Nice work!

rtripault:
So there is no more need to check against an array, right ?

OptimusCrime:
Yeah, that was what confused me too. The line could be `$k === 'content'` now.

Alroniks:
Cherry-picked to 2.x https://github.com/modxcms/revolution/commit/1424012c1002cf04eee1422f42b6d003be70c694

pixelchutes:
Has anyone experienced double `<?php` tags at top of cached snippet when **saving a Snippet within MODX** on `2.5.8-dev` (or `2.6.x`), resulting in PHP error?

```PHP Parse error:  syntax error, unexpected '<'```

I'll file a new issue once I've confirmed related to this change, but seems like a likely culprit.

Issue #5512, opened 2011-08-12
TV input type "textarea" not high enough


mindeffects:
mindeffects created Redmine issue ID 5512

TV input type "textarea" is not high enough. RTFM says "about 15 lines", I count only about 8 lines (Chromium 15.x and FF5).

I will also write a feature request for textareas with editable height. Don't want to mix bugs and features. ;-)

devlanda:
I am working on a fix for this issue.

pyrographics:
Not sure if it would be related to this issue or not but is in same ballpark.  If you uncheck "Rich Text" on a resources settings to change a rich text box to a plain text box, a rich text TV correctly turns into a standard textarea but the size is so tiny (one short row) you can't use it and it cannot be resized.

OptimusCrime:
Perhaps worth opening a new issue for this? It does not concern the rich text editor.

ilyautkin:
PR https://github.com/modxcms/revolution/pull/13524 is closed and feature is not implemented yet.
Confirmed for MODX 2.7.0

JoshuaLuckers:
There is a `grow` property that can be set to `true`. If `true` the textarea will be the same height as its content.

PR #13524, opened 2017-07-07
Fixed issue #5512 #modxbughunt

devlanda:
### What does it do?
It makes it possible to resize the textarea (input type TV) in vertical direction. The css code adds the resize option on the right bottom edge. So you don't have to scroll when you have more than 7 lines.

### Why is it needed?
Currently you have to scroll when you have above 7 lines in a textarea. In that case you may loose the overview.

### Related issue(s)/PR(s)
TV input type "textarea" not high enough (https://github.com/modxcms/revolution/issues/5512)

### DIFF
--- a/_build/templates/default/sass/_forms.scss
+++ b/_build/templates/default/sass/_forms.scss
@@ -33,6 +33,7 @@ textarea.x-form-field,
   border-radius: $borderRadius;
   border: 1px solid $borderColor;
   position: relative;
+  resize: vertical;
   transition: border-color 0.25s;
 }
 


gpsietzema:
#modxbughunt #3points to @devlanda

OptimusCrime:
Have you tested that this actually works? I am pretty sure ExtJS and all its weird styling was the reason this feature was not added when the original issue was opened.

devlanda:
You are right! I did the test in browser. I was unaware about the weird impact of ExtJS.
This fix can be rejected.

OptimusCrime:
No problem. Just had a feeling ExtJS did not play along that nicely.

PR #13527, opened 2017-07-07
Force setting the original value of the resource template to prevent false warning when using "empty" template

### LINKED ISSUE #13483: False pending changes warning
### False pending changes warning when closing a resource
When a resources template is set to 'empty' the user receives a false 'You have changes pending; are you sure you want to cancel?' warning.

### Step to reproduce
1. Set a resources template to 'empty'
2. Save resource
3. Click close

### Observed behavior
User receives a false 'You have changes pending; are you sure you want to cancel?' warning.

### Expected behavior
No warning as resource is saved

### Environment
MODX 2.5.7, all browsers

rtripault:
### What does it do?

Make sure the original value for the template is set

### Why is it needed?

Prevent false warnings when closing the panel & using the "empty" template

### Related issue(s)/PR(s)

Closes #13483

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -40,6 +40,10 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
     ,setup: function() {
         if (!this.initialized) {
             this.getForm().setValues(this.config.record);
+            var tpl = this.getForm().findField('modx-resource-template');
+            if (tpl) {
+                tpl.originalValue = this.config.record.template;
+            }
             var pcmb = this.getForm().findField('parent-cmb');
             if (pcmb && Ext.isEmpty(this.config.record.parent_pagetitle)) {
                 pcmb.setValue('');


sonicpunk:
#modxbughunt #3points to @rtripault

OptimusCrime:
Nice! But the check you do is weak. Do you test agains the existing component? Then it perhaps should be `=== null`.

rtripault:
Sorry, not sure i understand here...

OptimusCrime:
```
var tpl = this.getForm().findField('modx-resource-template');
if (tpl) { // What does this line check? Does getForm and findField return a boolean value true/value?
    [...]
}
```

If `getField` returns either an object or null, it would be more readable with a strict comparison, like so:

```
var tpl = this.getForm().findField('modx-resource-template');
if (tpl !== null) { // We now know that we have a working object that is not null
    [...]
}
```

rtripault:
Thanks, but i feel fine with the code that way (which is similar to the existant code base).
But leaving integrator decline/request adjustments if they feel like that's needed.

Also, FWIW http://docs.sencha.com/extjs/3.4.0/source/BasicForm.html#Ext-form-BasicForm-method-findField

OptimusCrime:
I see. And sure, it works. I just prefer stricter comparison over loser ones.

Issue #12766, opened 2015-11-18
User primary_group is not set initialy


Lefthandmedia:
When adding a new user and selecting a usergroup for that user, the primary_group remains 0 on creation.
Only after adding a second usergroup the value for primary_group changes to the correct usergroupID
This behaviour is relative new. My application did not have this bug before. I think this is introduced in 2.4 but i'm not sure

this issue resembles #11139

Kookabura:
I think it doesn't connected with #11139. That's a standalone issue.
The issue occurs because primary_group isn't set up within user create processor. It's set up only within user update processor. So function setUserGroups() in user create processor should be fixed.

Lefthandmedia:
I guess you're right it is a separate issue: to recreate
- Create a new user  (give name and email address, do NOT save at this time)
- Go to permissions tab and add user to a group. When you save you can see the user ID is still set to 0 and the name to "none". thus this action has not been saved even though you clicked save
- Now when you go back to the 'general information' tab and save there. the newly chosen usergroup isn't send to the server for saving and thus does not get saved.

So the new and update action for user should also submit the data from the permissions tab

wuuti:
Reviewing this #modxbughunt

wuuti:
Can confirm this is still happening in current 2.5.x branch.

sonicpunk:
#modxbughunt #1point to @wuuti

Alroniks:
PR https://github.com/modxcms/revolution/pull/13528 merged

mrhaw:
This fix should be ported to the Login.Register snippet. ...Currently we are forced to use a postHook
*&postHooks=`primarygroup`*
example:
```
$user = $hook->getValue('register.user');
$user->set("primary_group", 1);
$user->save();
````

No matter if we create users with Register or in code, it creates this error on user deletion, if not set:
*Error - table class/name for class: modAccess*
https://forums.modx.com/thread/?thread=96248
**(Spoke to soon on this - I still see this error sometimes)**

The extra getUserProfiles Snippet also looks to primary group of a user:
https://modx.com/extras/package/getuserprofiles

Wonder why this was changed? I will reference this url back here:
https://forums.modx.com/thread/88810/what-is-the-primary-group-in-the-users-table

JoshuaLuckers:
Hi @mrhaw, thanks for the info. Is it possible to submit issues about this to the repo's of those extras?
- Login: https://github.com/modxcms/Login/issues

[...]

PR #13528, opened 2017-07-07
Fixed bug #12766, where primary user group was not set in processor #modxbughunt

wuuti:
### What does it do?
Added missing setting for primary user group in modUserCreateProcessor->setUserGroups().

### Why is it needed?
Primary group was not set when user was created directly with a group membership.

### Related issue(s)/PR(s)
Issue #12766 .

### DIFF
--- a/core/model/modx/processors/security/user/create.class.php
+++ b/core/model/modx/processors/security/user/create.class.php
@@ -86,6 +86,7 @@ class modUserCreateProcessor extends modObjectCreateProcessor {
         $memberships = array();
         $groups = $this->getProperty('groups',null);
         if ($groups !== null) {
+            $primaryGroupId = 0;
             $groups = is_array($groups) ? $groups : $this->modx->fromJSON($groups);
             $groupsAdded = array();
             $idx = 0;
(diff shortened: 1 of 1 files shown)


sonicpunk:

Alroniks:
Also backported into 2.x https://github.com/modxcms/revolution/tree/a4befced978351f5996f0547c92439c48b086ddc

mrhaw:
This fix should be ported to the Login.Register snippet. ...Currently we are forced to use a postHook
*&postHooks=`primarygroup`*
example:
```
$user = $hook->getValue('register.user');
$user->set("primary_group", 1);
$user->save();
````

No matter if we create users with Register or in code, it creates this error on user deletion, if not set:
*Error - table class/name for class: modAccess*
https://forums.modx.com/thread/?thread=96248
**(Spoke to soon on this - I still see this error sometimes)**

The extra getUserProfiles Snippet also looks to primary group of a user:
https://modx.com/extras/package/getuserprofiles

Wonder why this was changed? I will reference this url back here:
https://forums.modx.com/thread/88810/what-is-the-primary-group-in-the-users-table

PR #13530, opened 2017-07-07
Proper processing uncached elements inside cached [#13250]

bezumkin:
### What does it do?
It fixes weird parser behaviour when uncached tags called inside cached.

### Why is it needed?
When parser will see uncached element inside cached it will leave cached tag until processing of all uncached elements. So such cached element will be uncached in real.

People should not call uncached elements inside cached, so this fix will warn them about it in system log.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13250

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -449,6 +449,10 @@ class modParser {
             $cacheable= false;
             $tokenOffset++;
             $token= substr($tagName, $tokenOffset, 1);
+        } elseif (!$processUncacheable && strpos($tagPropString, '[[!') !== false) {
+            $this->modx->log(xPDO::LOG_LEVEL_ERROR, "You should not call uncached elements inside cached!\nOuter tag: {$tag[0]}\nInner tag {$innerTag}");
+            $this->_processingTag = false;
+            return $outerTag;
         }
         if ($cacheable && $token !== '+') {
             $elementOutput= $this->loadFromCache($outerTag);


sonicpunk:
#modxbughunt #3points to @bezumkin

OptimusCrime:
Wow, does this really work? In that case, AWESOME!

sdrenth:
Tested PR and can confirm that this PR fixes the bugs mentioned in issue #13250 and that the warnings are indeed added to the system logs.

opengeek:
I'm not convinced this is always an error. Should we reduce this to a warning? It seems to me there are valid use cases for calling noncacheable inner tags within a cacheable outer tag. e.g.

```
[[tag? &type=`p` &content=`[[!time]]`]]
```

bezumkin:
Of course we can reduce this to warning, but honestly, there is no difference between
```
[[tag? &type=`p` &content=`[[!time]]`]]
```
and 
```
[[!tag? &type=`p` &content=`[[!time]]`]]
```
because if there at least one uncached nested tag, parent cached tag also will work as uncached. 

Maybe somebody will think, that this cached element must work as really cached, but it will not. So the only question is should we bother developer with such reminders or not?

opengeek:
You are wrong. It does cache the results of the outer tag so it is not evaluated once the resource is cached. There is a clear difference, despite the oversimplified example.

bezumkin:
This behavior we are trying to change int this PR. Look at the code again
```php
if (!$processUncacheable && strpos($tagPropString, '[[!') !== false) {
    $this->_processingTag = false;
    return $outerTag;
}
```
If we are not processing uncached elements right now. 
And there at least one uncached tag inside current tag
Just exit

So, all cached tags will be bypassed before parser process all uncached tags inside them. And such cached tags works as uncached - that is the whole point of this PR.

opengeek:
I understand now, sorry. Confirmed it does not destroy the use case I was concerned about.

OptimusCrime:
Hurray!

hugopeek:
Ok, late to the party.. But this change does affect me.

I'm running into a lot of "You should not call uncached elements inside cached" errors now after updating all my sites to 2.6.5. I was holding off the 2.5 -> 2.6 update for a while now because I tried it on a few sites back then, got these errors, and decided to deal with this later.

My use case involves ContentBlocks. For most blocks, you can enter the HTML for the output in a Template field. Works fine, but I wanted to maintain this HTML in a static file instead, to improve portability across projects and make maintenance a little easier. So in my ContentBlocks template fields, I often call chunks like this:

```
[[$headingBasicSubtitle?
    &value=`[[+value]]`
    &level=`[[+level]]`
    &subtitle=`[[+subtitle]]`
    &alignment=`[[+alignment]]`
    &unique_idx=`[[+unique_idx]]`
]]
```

As of ContentBlocks 1.7, this use case is also supported natively:

>Field/layout templates now support @CHUNK to create chunk tags with all available properties [#461]

But if I understand correctly, this means that after the caching update, these chunks cannot contain any uncached placeholders anymore.. This would also mean that any values entered by the editor cannot contain uncached tags. In the above example, I ran into a situation where the subtitle property contained a username. This results in the following error:

```
[2018-07-23 07:29:00] (ERROR @ /home/hugo/Localhost/romanesco-nursery/core/model/modx/modparser.class.php : 452) You should not call uncached elements inside cached!
Outer tag: [[$headingBasicSubtitle?
    &value=`Your website is ready!<br>`
    &level=`h1`
    &subtitle=`What shall we do now [[!+modx.user.id:userinfo=`username`]]? `
    &alignment=`center aligned`
    &unique_idx=`1`
]]
Inner tag $headingBasicSubtitle?
    &value=`Your website is ready!<br>`
    &level=`h1`
    &subtitle=`What shall we do now [[!+modx.user.id:userinfo=`username`]]? `
    &alignment=`center aligned`
    &unique_idx=`1`
```

Using the `@CHUNK` option has the same effect. And since I'm using this pattern a lot, combined with the fact that the error message is quite elaborate, my error logs are all growing like mushrooms on a fresh pile of steamy horse poo right now. 

I'm all for this PR and making the caching more predictable, but to me this update feels a little too intrusive now. Yes, I understand that what I'm doing is "wrong", technically speaking, but if it doesn't slow down my sites too much (above example takes just 0.0023179s to parse), then why roast me this harshly for it in the error log? And secondly: why shouldn't I be free to (ab)use MODX in any way I prefer? It this works for me, then why not let me be?

So I second the motion of opengeek to reduce this to a warning. Or alternatively, maybe create a system setting instead (similar to log_snippet_not_found) to suppress the errors?

Mark-H:
A merged PR is not the place to bring up a new discussion ;)

hugopeek:
Ok.. Create a new issue then? Or continue in this one: #13671?

BenceSzalai:
Can anyone explain why it is good that a tag cannot be cached just because it has an uncached input in it? For me it seems more like a consequence of how the tag processing is implemented rather than something that is decided as the optimal way to go.

As a simple counter example, when caching of an outer tag with an uncached inner tag would be desirable is a tag that generates a list of resources and sorts them based on some user input. E.g.:

```
[[pdoResources? ... &sortby=`[[!getSortOrder]]`]]
```

In this case lets say the UI has a drop-down list and we know there are e.g. 5 different sort order they can choose from. `[[!getSortOrder]]` reads this information from GET or POST and wants to render a list of resources in that order. Caching that 5 differently sorted list is desirable, as getting the sort order from GET/POST is a cheap operation, but gathering the resources can take a lot of db and cpu power.

So in this case caching the outer tag's result is desirable even thought it has an uncached inner tag.

Have such use-cases been considered?

Afaics (i did not looked at very close) this is currently unsupported, because the parser wants to process all cached tags first and all uncached after. Obviously changing this order could break many existing implementations, so this is kind of a hard given.

What I'm thinking about is maybe another tag type, where we can tell the parser to process the given tag at the time it processes the cached tags, but do not cache it. E.g.:

```
[[pdoResources? ... &sortby=`[[¡getSortOrder]]`]]
```

(`¡` is just an example, the right symbol is to be decided...)

This could tell the parser that `getSortOrder` must be executed straight away when encountered, obtain the result, substitute it into the outer tag and than move on processing the outer tag as cached, just like if the actual output value of the `[[¡getSortOrder]]` was a simple static hardcoded value.

**So basically what I'm missing is a tag prefix that makes the tag uncached, but does not delays evaluation to after all cached tags, but performs it immediately.**

Now obviously this would put some responsibility to the developer who uses this pattern, as for example this is very beneficial when we know the inner tag produces a handful of potential results, but for example doing the same with a snippet that for example gives the actual time in miliseconds, it'd litter the cache with endless data never to be reused again.

Still I see many times that this pattern would benefit the sites I'm working on and could boost performance a lot.

I've played around a bit with the parser class to see if this can be implemented in a reasonable way, and I think it is.

However before putting much effort I'd like to hear opinions. **Is this a bad idea? Am I missing another way to achieve the same results? Would ModX team be interested in such PR?**

Another aspect of changes to the parser would be incompatibility with extras that override the current parser, like fastField that ads tag type `#`, or maybe that is now supported by the native parser as well, but anyway, some extras may become incompatible, but I'm just putting it here to not forget, but right now it is not a thing I've looked into deep enough yet.

Mark-H:
> this is currently unsupported, because the parser wants to process all cached tags first and all uncached after. Obviously changing this order could break many existing implementations, so this is kind of a hard given.

This is indeed the crux of it. 

Uncached tags are processed in a separate loop of the parser, **after the cached tags are processed and written to cache (or in a cached request, loaded from cache)**. And changing that could lead to a whole bunch of unexpected issues for all sites ever built.  

> This could tell the parser that getSortOrder must be executed straight away when encountered, obtain the result, substitute it into the outer tag and than move on processing the outer tag as cached, just like if the actual output value of the [[¡getSortOrder]] was a simple static hardcoded value.

This means that once the result is obtained and substituted, **its result will be written to the cache**, not the uncached tag itself. It's not possible to introduce what you're proposing without a major overhaul. 

The right solution here is to wrap pdoResources with an uncached snippet. Either getCache or a custom snippet. You call that snippet uncached so it can respond to the user input, but programmatically it caches each of the different result sets. 

IIRC both getPage and getCache do this automatically, including using different cache keys based on provided properties and/or $_GET params.

BenceSzalai:
> This means that once the result is obtained and substituted, **its result will be written to the cache**, not the uncached tag itself. It's not possible to introduce what you're proposing without a major overhaul.

What I mean is that when the cache is being looked up, it'll look up something like (sticking to the example) ```[[pdoResources? ... &sortby=`{"pagetitle":"DESC"}`]]```,  where `{"pagetitle":"DESC"}` is the return of the `[[¡getSortOrder]]` being executed. Next time when the `[[¡getSortOrder]]` returns something else, e.g.: `{"menutitle":"ASC"}`, i'll be substituted into the outer snippet as ```[[pdoResources? ... &sortby=`{"menutitle":"ASC"}`]]```, which if has a cache entry would be a different one, because the inputs are different.
So right, the results are implicitly written into the cache, but the point is that depending on those results it may generate and look up multiple different cache entries with different outputs.

But I get it what you are saying about using a wrapper snippet that handles the caching. I just feel like the ability to do it in simple cases by making a snippet uncached, but not delayed would be nice.

Well anyway, I might still try to make an own parser for it and see where it goes. As far as I saw the parser code it is complex, but not like incomprehensible. Probably unexpected side-effects are the culprit. Will see. For me this seems to be too obvious "nice to have" thing, every time I face this I wish there was a way to make snippet "immediate uncached" instead of "delayed uncached"...

Thanks for your answer indeed! :)

Issue #9906, opened 2013-05-04
Inextensible modUserGetListProcessor


modxbot:
leffyrant created Redmine issue ID 9906

h2. Description of the Problem

Let's say that we have created the class _customUser_ as an extension of the _modUser_ class. For its manipulation we will have to create related processors. All of the processors behave as expected and manipulate only instances of the _customUser_ class. This is easily achieved by setting the "classKey" member variable to "customUser".

However, the _customUserGetListProcessor_ cannot be configured the same way.

h2. The Cause of the Problem

The problem is located in the following methods:
# prepareQueryBeforeCount(xPDOQuery) and
# prepareQueryAfterCount(xPDOQuery)

of the _modUserGetListProcessor_.

I believe that, in both of the methods, the problem is that the string "modUser" is used, instead of the property "classKey". 

h3. prepareQueryBeforeCount(xPDOQuery)

Specifically, I think in (1) the following code:

<pre>
if (!empty($query)) {
    $c->where(array('modUser.username:LIKE' => '%'.$query.'%'));
    ...
}
</pre>

should actually be:

<pre>
if (!empty($query)) {
    $c->where(array($this->classKey . '.username:LIKE' => '%'.$query.'%'));
    ...
}
</pre>

h3. prepareQueryAfterCount(xPDOQuery)

As for (2), I think the code:

<pre>
public function prepareQueryAfterCount(xPDOQuery $c) {
    $c->select($this->modx->getSelectColumns('modUser','modUser'));
    ...
}
</pre>

should be:

<pre>
public function prepareQueryAfterCount(xPDOQuery $c) {
    $c->select($this->modx->getSelectColumns($this->classKey, $this->classKey));
    ...
}
</pre>

h3. getData()

It is worth noting that the above method initializes the _xPDOQuery_ used by the other two _prepareQuery*_ methods with the following code:

<pre>
public function getData() {
    ...
    $c = $this->modx->newQuery($this->classKey);
    ...
}
</pre>

Supposing that we have set the "classKey" property to "customUser", the above code clearly shows that the search will be about a _customUser_ and not a _modUser_.

[...]

PR #13532, opened 2017-07-07
Remove hardcoded class name from sql queries from user processors

alroniks:
### What does it do?
It replaces hardcoded modUser class name from query builders for having the ability to extends base processor classes and use with custom use them with custom user classes.

### Why is it needed?
It solves the issue with a wrong class name in case processor extended and used custom user class.

### Related issue(s)/PR(s)
#9906
#modxbughunt


sonicpunk:

OptimusCrime:
Issue with line endings, but I guess we should just go ahead and fix that by now.

Alroniks:
Yes, indeed. Seems it is just an old code. Better to fix file once that meet it again in the future.

rtripault:
In case you want a better diff here : https://github.com/modxcms/revolution/pull/13532/files?w=1

Issue #13351, opened 2017-03-03
#modxbughunt Items in file trees can only be drag dropped once


labr1005:
### Summary
Directories or files in file trees can only be drag dropped once.

### Steps to reproduce 
Files Tree
- Dir A
  - File
- Dir B
- Dir C

Drag File from Dir A to Dir B.
Drag File from Dir B to Dir C.

### Observed behavior
The complete tree reloads and File is still in Dir B.

### Expected behavior
File should be in Dir C.

### Environment
MODX 2.5.5
Apache 2.2.29
PHP 7.0.8

digitalpenguin:
I can reproduce this issue.

labr1005:
I almost had a fix, but not quite. If some ExtJs expert is interested maybe I can give you a starting point.

OptimusCrime:
@labr1005 Please link your attempt :)

labr1005:
It works, but it looks ugly and doesn't feel right to call refreshNode as this reloads the parent node.

But it sorts the child nodes alphabetically. :)

labr1005:
I'm not sure if ExtJs Tree Sorter is an option, but that's too high for me for now.
http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.tree.TreeSorter

IMO the desired behavior should be

- drop the directory/file
- expand the target directory, if it's closed
- place the directory/file in it's alphabetically correct place inside the target directory
- do the fancy drop animation on the dropped directory/file

Finetuned:
"I'm going to try and fix this today."

PR #13533, opened 2017-07-07
Fixed dragging files more than once [#13351]

Finetuned:
### What does it do?
Added missing afterSort event listener and _handleAfterDrop method.
Added a reload(true) onto the taergetNode in _handleafterDrop

### Why is it needed?
Filesystem tree would not let you move a file more than once or twice.

### Related issues
#13351

### DIFF
--- a/manager/assets/modext/widgets/system/modx.tree.directory.js
+++ b/manager/assets/modext/widgets/system/modx.tree.directory.js
@@ -68,6 +68,7 @@ MODx.tree.Directory = function(config) {
         ,'afterRemove': true
         ,'fileBrowserSelect': true
         ,'changeSource': true
+        ,'afterSort': true
     });
     this.on('click',function(n,e) {
         n.select();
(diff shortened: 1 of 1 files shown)


sonicpunk:

OptimusCrime:
Very, very nice! Could perhaps use some descriptive comments about what your implementation does somewhere, as many of these lines are somewhat unclear for me.

Finetuned:
@OptimusCrime  The existing _handleDrag() method fired an afterSort event on the success callback. This event was not present in MODx.tree.Directory. 

Looking at  the other MODx.tree.* files, the afterSort event called a _handleafterDrop method. I found and migrated the missing code from one of those classes.

So then the actual bug fix for this issue is the additional line in _handleafterDrop : targetNode.reload(); 

After moving a file, for some reason (which may be to do with extended functionality on the tree class), the dragdrop functionality in Ext is sending incorrect source and target nodes on successive events. the reload() call merely ensures a localised update.

sdrenth:
@Finetuned nice fix! I can also confirm that the fix works nicely.

Issue #13526, opened 2017-07-07
Quick resource update breaks publish date and unpublish date depending on manager_date_format and manager_time_format


labr1005:
### Summary
Quick updating a resource that has a publish date or unpublish date set breaks this data - depending on the system settings manager_date_format and manager_time_format.

### Step to reproduce
Make sure manager_time_format is **not** set to "Y-m-d" and / or manager_time_format is **not** set to "H:i:s" (default after install should be "Y-m-d" and "g:i a").

Set a publish (and / or unpublish) date for a resource. Quick update the resource.

### Observed behavior
The publish and unpublish dates should be the same as before.

### Expected behavior
The publish and unpublish dates are empty or "1970-01-01 12:00 am".

### Environment
MODX 2.5.8-dev

![ezgif com-video-to-gif 1](https://user-images.githubusercontent.com/1913250/27967458-6ce8ad36-6343-11e7-8023-c748a74c44a8.gif)

labr1005:
With manager_time_format set to "Y-m-d" and manager_time_format to "H:i:s" quick updating works fine.

Flowan:
Confirmed this is not working in 2.5.x I will have a look at this. #modxbughunt

sonicpunk:
#modxbughunt #1point to @Flowan

gpsietzema:
#modxbughunt #1point to @labr1005

labr1005:
This might be related and also fix https://github.com/modxcms/revolution/issues/11891

PR #13534, opened 2017-07-07
Fixed date format for resource quick update #13526 #modxbughunt

flowan:
### What does it do?
Skip date format check when using quick update.

### Why is it needed?
The publish and unpublish dates should be the same as before.

### Related issue(s)/PR(s)
Related issue #13526

### DIFF
--- a/core/model/modx/processors/resource/get.class.php
+++ b/core/model/modx/processors/resource/get.class.php
@@ -17,7 +17,10 @@ class modResourceGetProcessor extends modObjectGetProcessor {
     public function process() {
         $resourceArray = $this->object->toArray();
         $resourceArray['canpublish'] = $this->modx->hasPermission('publish_document');
-        $this->formatDates($resourceArray);
+        if (!$this->getProperty('skipFormatDates') ||
+            ($this->getProperty('skipFormatDates') && $this->getProperty('skipFormatDates') == 'false')) {
+            $this->formatDates($resourceArray);
+        }
         return $this->success('',$resourceArray);
     }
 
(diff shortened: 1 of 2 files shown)


sonicpunk:

pixelive:
Tested and works.

sonicpunk:
#modxbughunt #1point to @pixelive

Mark-H:
Code looks clear enough, but could you explain _how_ this fixes the issue? What did you find was the underlying problem that is fixed by not formatting the dates in the processor?

sdrenth:
@flowan The fix does work, but the real problem is the formatDates method which isn't working correctly.

The date output when manager_date_format is set to Y-m-d results in:
2017-08-30

If you read this from the strtotime docs:

> Dates in the m/d/y or d-m-y formats are disambiguated by looking at the separator between the various components: if the separator is a slash (/), then the American m/d/y is assumed; whereas if the separator is a dash (-) or a dot (.), then the European d-m-y format is assumed. If, however, the year is given in a two digit format and the separator is a dash (-, the date string is parsed as y-m-d.
To avoid potential ambiguity, it's best to use ISO 8601 (YYYY-MM-DD) dates or DateTime::createFromFormat() when possible.

So in this case the formatDates method is expecting a different notation of the date, thats why the wrong date is being returned.

Jako:
The fix looks well. But I am not sure if the 'resource/get' processor is used elsewhere in the manager  than in https://github.com/flowan/revolution/blob/dfa44ee9c4f88421782d0949e5345fc5b7d2327e/manager/assets/modext/widgets/resource/modx.tree.resource.js#L472 (a search for 'resource/get' returns only that file). So maybe all the date formatting things could be removed from the processor.

Issue #9918, opened 2013-05-12
No unique key in modTemplateVarResource


modxbot:
bezumkin2 created Redmine issue ID 9918

The table _modTemplateVarResource_ has no unique key for fields\* tmplvarid + contentid*, and I can smply add duplicate values to it trough xPDO or SQL.

We can have some values for one TV - I think, it is not normal situation.

Alroniks:
#modxbughunt work

alroniks:
Closed because related pull request is merged.

PR #13535, opened 2017-07-07
Added unique index for template variable values

alroniks:
### What does it do?
It adds unique key to index for modTemplateVarResource object

### Why is it needed?
It needs to prevent duplication of values in `*_site_tmplvar_contentvalues` table for avoiding data issues in cases when by bad script other reason in this tables can exist different data for one template and one tv.

### Related issue(s)/PR(s)
#9918
#modxbughunt

### DIFF
--- a/core/model/modx/mysql/modtemplatevarresource.map.inc.php
+++ b/core/model/modx/mysql/modtemplatevarresource.map.inc.php
@@ -79,7 +79,7 @@ $xpdo_meta_map['modTemplateVarResource']= array (
     array (
       'alias' => 'tv_cnt',
       'primary' => false,
-      'unique' => false,
+      'unique' => true,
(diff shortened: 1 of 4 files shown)


opengeek (commented):
I'm wondering if this should not be a new index so we still have an index on tmplvarid as well.

gpsietzema:

OptimusCrime:
@opengeek Not possible with both indexes?

Mark-H:
There are indices on tmplvarid, contentid, and the combined tv_cnt one that is affected by this change. The tmplvarid index is just above the shown diff.

opengeek:
Doh! Nevermind then. Please excuse my apparently poor attempts to review the work.

sdrenth:
I can confirm that the fix works!

Issue #12759, opened 2015-11-09
Context menu order doesn't preserved after page refresh


whitebyte:
If we have multiple contexts on a site and sort them in desired order by drag'n'drop, and then reload a page, contexts become sorted alphabetically and don't remember their menu order.

2.4.2

whitebyte:
Well, I just realized that sort order can be preserved with `context_tree_sort` system settings is turned on. So technically this issue may be closed. But now I'm trying to figure out, what is the purpose of this setting? Are there any use cases when we dont't want contexts to be sortable? If so, contexts should not be sortable at all when `context_tree_sort` setting is disabled and shouldn't silently restire initial order after page reload.

OptimusCrime:
I agree that this setting should be on by default. It might be because of backwards changes or something? Still voting to change the default behavior here.

sebastian-marinescu:
This should definitely be activated by default. 
Or if the setting is deactivated by default, it shouldn't allow dragging around the contexts.
Also the description of this setting **needs** to be corrected, as it states it sorts _alphanumerically_.

christianseel:
Closing because the PR has been merged. The setting is on by default in the next minor release.
Also I created #13537 to fix the setting descriptions.
#modxbughunt 

@modbot close

sonicpunk:
#modxbughunt #1point to @christianseel

PR #13537, opened 2017-07-07
Fix/Improve setting_context_tree_sort_desc

christianseel:
### What does it do?
Fixes the `context_tree_sort` setting description.

### Why is it needed?
Previously the description said that after enabling the setting context would be sorted _alphanumerically_:
> If set to Yes, Contexts will be alphanumerically sorted in the left-hand Resources tree.
which is wrong. By default they are sorted by rank (set in `context_tree_sortby`).

New description:
> If set to Yes, Contexts will be sortable in the left-hand Resources tree. You can customize the field to sort the Contexts in the \'context_tree_sortby\' setting.

### Related issue(s)/PR(s)
#12759 

#modxbughunt


sonicpunk:
#modxbughunt #3point to @christianseel

gpsietzema:
#modxbughunt #1point to @Alroniks

Issue #4014, opened 2011-03-09
modResource::joinGroup fails when numeric string is passed as an ID


jrotering:
jrotering created Redmine issue ID 4014

See forum topic here:

[[http://modxcms.com/forums/index.php/topic,62021.0.html]]

The function is written to take either a name or an ID of a resource group.  It tries to determine whether the passed argument is an ID by testing it with is_int().  This fails on numeric strings.  

I think the options are 

1) leave it as it is and focus on educating end users that if they pass an ID, they have to make sure that it is of type 'integer' and not a numeric string.
2) change the function as I suggested above - the arg will be considered an ID if it passes is_integer and contains only digits.
3) change the function so that it is 'less flexible' - make it require an array for input  "$resource->joinGroup(array('id'=>$id))" or "$resource->joinGroup(array('name'=>$name))"

Additional note - I'm not sure but this may also apply to modUser::joinGroup.  Whatever gets done should probably be consistent throughout all ::joinGroup methods.

OptimusCrime:
This will always be a problem in losely typed languages like PHP is. Solution 1 poses the restriction on the input type. I for one think having the limitation that integer = id and string = name is good enough. Solution 2 will fail in cases where groups are given names with only digits. Solution 3 would be a BC that may be a too big change to swollow.

If I was to redesign this I would not have noe method that handled multiple things, e.i. joining with id and joining with name. A much more clean approach would be to name them something like `joinGroupWithName` or something similar. However, this is too late to introduce now without creating a huge BC in the system.

Because of reasons explained above I am more in favor of having it as it currently is.

OptimusCrime:
One way to "enhancing" the core could be to create the methods `joinGroupWithName` and `joinGroupWithId` to make it easier for developers to use the API.

goldsky:
I'm going to try and fix this today. #modxbughunt

Mark-H:
Fixed per https://github.com/modxcms/revolution/commit/773e80d9f0c2e5eb1173ad844c2bcaefc8ee2c87

PR #13538, opened 2017-07-07
Add option to force joinGroup to use numeric group names as name [#4014]

goldsky:
### What does it do?
Allow API to force join group by name where the name is numeric
$resource->joinGroup('666', true);

### Why is it needed?
The current API only checks whether the value passed is integer or not.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/4014

#modxbughunt

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -747,7 +747,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
                 }
             }
         }
-     
+
         return $removed;
     }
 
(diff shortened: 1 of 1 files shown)


sonicpunk:
#modxbughunt #3points to @goldsky

rtripault:
I'd say this is a breaking change since we now need to specify when using a name while it previously try to "guess" (albeit sometimes wrongly).

:-1: here (in that current state)

goldsky:
default false, fallback to previous.

rtripault:
Oh, my bad... :ghost:

core/model/modx/modresource.class.php
```diff
@@ -1093,13 +1093,20 @@ public function duplicate(array $options = array()) {
      *
      * @access public
      * @param mixed $resourceGroupPk Either the ID, name or object of the Resource Group
+     * @param boolena $byName Force the criteria to check by name for Numeric usergroup's name
```
Mark-H: Typo in `boolean` :P

Mark-H:
Changing bases doesn't seem to have the desired effect anymore now that 2.5.x and 2.x have diverged too much.. this does seem like something that should be in 2.6 rather than 2.5.8, so perhaps will need to be cherry picked.

goldsky:
typo `boolena` is fixed :stuck_out_tongue_winking_eye:

Mark-H:
Thanks @goldsky! Cherry-picked into 2.x per https://github.com/modxcms/revolution/commit/773e80d9f0c2e5eb1173ad844c2bcaefc8ee2c87

PR #13539, opened 2017-07-07
Fix #7580. Do not add duplicate prefix for the children. #modxbughunt

wuuti:
### What does it do?
Does not add the "Duplicate of" prefix to duplicated child resources.

### Why is it needed?
Without all child resources pagetitles are changed, which then might have to get manually changed back. Can be horrible e.g. if you duplicated a big collection.

### Related issue(s)/PR(s)
Fixes #7580.

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -1075,11 +1075,12 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
                     $child->duplicate(array(
                         'duplicateChildren' => true,
                         'parent' => $newResource->get('id'),
-                        'prefixDuplicate' => $prefixDuplicate,
+                        'prefixDuplicate' => false,
                         'overrides' => !empty($options['overrides']) ? $options['overrides'] : false,
                         'publishedMode' => $publishedMode,
                         'preserve_alias' => $preserve_alias,
-                        'preserve_menuindex' => $preserve_menuindex
+                        'preserve_menuindex' => $preserve_menuindex,
+                        'preserve_pagetitle' => true
                     ));
                 }
             }


OptimusCrime:
This is a huge backwards change, so it can not target 2.5.x, actually, not even 2.x, it would need to target 3.x.

Additionally, the use of the "Duplicate of" prefix is to avoid duplicated aliases across contexts. There is no check for this here. Also, you did not reference the issue. #7580.

Mark-H:
Rather than hardcoding it to false for the children, perhaps this could be done with another option on the duplicate dialog? Should be suitable for a 2.x release then, I think.

OptimusCrime:
A checkbox on the dialog would be preferable.

Mark-H:
Thoughts @wuuti?

wuuti:
Well, checkbox seems ok for me. I can try to implement that. 

@OptimusCrime mentioned the aliases: imho that makes the original issue even more weird - we change the title (and by this the alias) with lots of text just to avoid duplicate aliases? Sounds like a dirty workaround for me, which sadly is now interpreted as standard behaviour. Wouldn't it be better to make sure that the system generates _unique aliases_ here if pagetitle leads to duplicates? (Btw: the adding of "Duplicate of ..." to the pagetitle does not prevent you from having a duplicate alias, if such a resource already exists. Which can simply happen if you duplicate a resource twice) 
I was always wondering why MODX cannot simply generate an alias suffixed with the _id_ or an incremented counter in case of a duplicate detection.

Mark-H:
Hm, regarding the duplicate ~alias~ URI issues, that shouldn't happen because the parent you're duplicating is already renamed. The checkbox would also be a little confusing, so I'm tempted to merge this as-is... would be a UX improvement for sure..

wuuti:
Mh, so whats the state of the discussion here right now? I just checked the alias behaviour in 2.6.1-dev, and the aliases here are all the same: if you duplicate resource _A_ with alias _a_ once, you end up with a resource named _Duplicate of A_ with an alias of _a_. If you duplicate _A_ again you will get another resource with name _Duplicate of A_ and and alias of _a_. So you always end up with duplicate aliases. So adding the annyoing "Duplicate of" does not help with the alias problem, but rises the problem if the need to fix all pagetitles after duplicating. Thats not really editorfriendly...

opengeek:
Is there any consensus on this issue? Does it or does it not need a checkbox added to the dialog?

JoshuaLuckers:
If a checkbox is added will it be merged into 2.7?

OptimusCrime:
I would suggest not adding it to 2.7 to prolog the waiting.

Mark-H:
We're past the 2.7 feature cut-off (that was the 13th), so: no.

azeller:
I read the whole argument and I can see that the checkbox for this seems to be of general consent. However, I think this is the wrong approach: I can't think of a single use-case where I can imagine wanting the prefix in the resource and in the URI. This is causing a lot of clutter and shouldn't even be the default behavior. I would personally recommend getting rid of this (for children) entirely.

The "duplicate of" prefix has just been introduced to create a default name to avoid URI collisions.

I think this should be default behavior to **not rename children**. I'd be happy to hear a reason why this should stay though.

JoshuaLuckers:
We also have to consider situations where `use_alias_path` is set to `No`.

azeller:
> We also have to consider situations where `use_alias_path` is set to `No`.
Definitely. However, this can be solved systematically by adding a hash or an incremental number of some sort to the URI. Adding "duplicate-of" as a prefix is the worst behavior I can think of. resource-name-1 is at least something nobody will be ashamed of when it's left over in a live site :)

global_duplicate_uri_check is also an important factor obviously. So these two should define the behavior. I don't see anybody wanting the "duplicate of" prefix. Maybe we should discuss which settings and circumstances should decide whether to name the children a certain way or just keep the names as they are.

Happy to hear other approaches on that.

opengeek:
This will need to be rebased on the latest code to still be considered. Closing the PR.

Issue #13276, opened 2017-02-06
Duplicate Context via Contexts/Context view


jaygilmore:
### Summary
It would be consistent (with other edit views) and convenient if it were possible to duplicate a context via the Contexts view `/manager/?a=context` or at the individual Context view, e.g., `/manager/?a=context/update&key=CONTEXTKEY`. 

### Step to reproduce
This is a Feature Request and follows observed/expected behaviors below.

### Observed behavior
It is only possible to duplicate a Context via the Resource Tree by right clicking on the Context name to activate the context menu and select Duplicate Context. It is not possible to do elsewhere inside the Manager. 

### Expected behavior
Retain the existing functionality to be able to duplicate a Context from the Resource Tree and:
1. Add the functionality to the Contexts view (by right clicking on the Context row to activate the context menu)
![contexts___modx](https://cloud.githubusercontent.com/assets/378199/22663947/07890218-ec85-11e6-9a68-0d74044f333c.png)

2. Add a duplicate button at the Save/Close/Help! buttons at the upper right. This would replicate the experience in Resource and Elements edit views. 
![context__web___modx](https://cloud.githubusercontent.com/assets/378199/22663953/0ca2de90-ec85-11e6-9bbb-80c7c387df9f.png)

### Environment
Revo Manager current++

joeke:
I'll look into this #modxbughunt

jaygilmore:
@Mark-H has this ready.

OptimusCrime:
@jaygilmore Really? Please reference PR.

Mark-H:
@OptimusCrime I was working on it while waiting on my flight. Had the feature pretty much ready, but didn't have wifi to push it to git, and used the last 3% of my phone battery to ask in the bug hunt slack for someone to find and comment on this issue to make sure others wouldn't also spend time on the same thing ;)

I made it home a little bit ago and just pushed up a pull request for your reviewing pleasure: #13540

OptimusCrime:
No problem. I thought jay had observed you publishing some PR. Nice work btw :)

rtripault:
Closing that one since #13540 landed in 2.x

Thanks

PR #13540, opened 2017-07-07
Allow contexts to be duplicated from the contexts grid and update panel [#13276]

Mark-H:
### What does it do?
Adds 2 new ways to duplicate a context. It was already possible to duplicate from the resource tree, but it's now also possible from the contexts grid (under System > Contexts) and when editing a context through the top bar. 

### Why is it needed?
These ways of duplicating a context are more consistent with how other types of items in MODX can be duplicated.  

### Related issue(s)/PR(s)
Requested in #13276


rtripault:
Working as advertised, thanks.

Just wondering if we should prevent/disallow duplicating the `mgr` context (as it was not listed in the tree, it was impossible to do so previously).

PR #13541, opened 2017-07-09
adding yarn.lock file

jpdevries:
yarn is [facebook’s fork of npm](https://code.facebook.com/posts/1840075619545360). 

### What does it do?
adding this file allows people to run

```bash
yarn
```

instead of
```bash
npm install
```

### Why is it needed?
It's not needed, just makes installing the dependencies a little quicker and also further ensures you have the the same versions as everyone else.

npm5 has its own way of doing this now too, so might be good to decide if we go with `yarn.lock`, `package-lock.json`, none, or both.


Alroniks:
If we are building MODX on the top version of npm, I prefer to use native `package-lock.json` instead of vendor lock from Facebook (I know that a lot of people use it already, btw).

Mark-H:
Yeah agreed with @alroniks. Yarn is nice but let's stick to one type, and as current instructions are for npm, lets go with #13542 instead.

OptimusCrime:
Does not this just allow for people to use both? As far as I've understood, the two can be used either-or?

Alroniks:
@OptimusCrime People can use both but there is core development so we have to have clear tools and instructions for contributors. Generating two different `.lock` files will produce a mess and headache for people who merge pull requests.

jpdevries:
We should probably add `yarn.lock` to `.gitignore` if it isn't already so people don't accidentally include it with PRs

PR #13542, opened 2017-07-09
adding package-lock.json

jpdevries:
### What does it do?
Adds `package-lock.json`

### Why is it needed?
It's not entirely necessary to have this under version control, but it makes running `npm install` quicker and more reliable by further ensuring people are running the same versions of stuff.

### Related issue(s)/PR(s)
#13541

### DIFF
--- a/_build/templates/default/package-lock.json
+++ b/_build/templates/default/package-lock.json
@@ -1,0 +1,3347 @@
+{
+  "name": "revolution-theme-default",
+  "version": "2.3.0",
+  "lockfileVersion": 1,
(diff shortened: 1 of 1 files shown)


OptimusCrime:
+1. This avoids edge cases where same package.json file installs different dependencies due to version selectors.

jpdevries:
Apparently npm is fixing a bug with `package-lock.json`:
https://github.com/h5bp/html5-boilerplate/pull/1966

but I think it is ok that this was merged already. Just something to be aware of.

Issue #12566, opened 2015-08-19
Bug: non-sudo users cannot change group settings


wuuti:
In 2.3.5 a non-sudo, but admin user (full admin policy) cannot change any group settings value. A "permission denied" is thrown when the user tries to change the value of a group setting. 
On the other hand he _can_ create new settings and _can_ delete existing settings...

When giving the user sudo access, everthing works fine. 

User/Context/System settings are not affected. 

Haven't tested it so far on the new 2.3.6 and 2.4.0-rc1 versions...

pixelchutes:
Hmm... what is an example of one of the group settings you are trying to change?

inreti:
Does not matter. You can create one, say "test_setting". After it is created you are not allowed to change the value any more. Same goes for overwriting common settings, like "manager_language".

Mark-H:
See #13544 for a fix.

PR #13544, opened 2017-07-10
Fix non-sudo users being unable of updating user group settings [#12566]

Mark-H:
### What does it do?
Changes the required permission from a non-existent to actually available setting. 

### Why is it needed?
As described in https://github.com/modxcms/revolution/issues/12566#issuecomment-132931163, non-sudo users were unable of saving user group setting values. They could create, but not update. 

This was due to the wrong permission being checked. Sudo users have all permissions, including non-existent ones, so they could still use the functionality.

### Related issue(s)/PR(s)
#12566


sdrenth:
@Mark-H Wouldn't the permission usergroup_edit be more logic instead of the usergroup_save permission since it is used for updating a usergroup setting?

PR #13547, opened 2017-07-12
[WIP] Fix svg preview in file manager #12459 + 2 additions

labr1005:
### What does it do?
- Adds support for Svg to the MODX Browser and the Files Tree.
- Adds checked background to transparent images in the MODX Browser.
- The preview now reflects the real proportions of the original image.

### Related issue(s)/PR(s)
#12459

### Before
bildschirmfoto 2017-07-12 um 22 26 18

### After
bildschirmfoto 2017-07-12 um 22 25 11

### Tree
bildschirmfoto 2017-07-17 um 20 39 14

In short: 1 more div per view. Sorry. And more inline styling. Sorry. See below.

Tested in IE 9+, latest Chrome, Firefox and Safari. Please also test!

Credits go to @marcjenkins for his initial PR #13517 and @sottwell for letting me use her S3 bucket. :)

Because some browsers have problems to display svgs the way they are meant to (due to that the width and height attributes for svgs are optional) the most robust solution I know of and use in every project for this purpose is an outer container (positioned relative) with either fixed width and height or no height and a padding to the bottom that reflects the height to width proportion of the image and then to position the image absolute.

Please, please let me know if you have another solution!

### DIFF
--- a/_build/templates/default/sass/_browser.scss
+++ b/_build/templates/default/sass/_browser.scss
@@ -57,10 +57,21 @@
     padding: 5px;
     text-align: center;
     width: 100px;
-  }
-    .modx-browser-thumb img {
+    .modx-browser-thumb-figure {
+      position: relative;
+      display: inline-block;
+      margin: 0 auto;
       vertical-align: middle;
+      background: url($imgPath + 'modx-theme/transparency-pattern.png') repeat left top #ccc;
+      img {
+        position: absolute;
+        left: 0;
+        top: 0;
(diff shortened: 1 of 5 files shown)


sdrenth:
Tested with default file browser and can confirm its working nicely in chrome, firefox, safari and IE 11 and IE Edge.

labr1005:
Closing this in favor of #13629.

The favored PR doesn't need the extra div per view, has no additional inline styling and a lot less Css.

Issue #7292, opened 2012-03-03
Rewording for Clarity in create/edit TV page


modxbot:
everettg_99 created Redmine issue ID 7292

_Input Option Values_

Currently reads:

<pre>
Option values for TVs with multiple selectable items, such as dropdown or tag (separate options with || ).
</pre>

Might better include fuller examples:

<pre>
Option values for TVs with multiple selectable items, such as dropdown or tag (separate options with ||, e.g. Cat||Dog or White==#000000||Black=#ffffff).
</pre>

_Default Value_

Currently reads:

<pre>
The default value this TV will have if none is specified.
</pre>

Might better read:

<pre>
The default value will be stored if the user does not specify a value.
</pre>

I just don't think it's clear who or what is doing the specifying in the bit "if none is specified".  It also might be good to let the user know that the default value is written to the database somehow, i.e. stored.

alroniks:
Solved and PR is merged.

PR #13550, opened 2017-07-14
Little improving of field captions on TV creating page

alroniks:
### What does it do?
It changes a lexicon strings for field captions on TV creating page.

### Why is it needed?
It makes field captions more readable and clear for users.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/7292


OptimusCrime:
Very nice! I always get the TV structure wrong. +10000

sdrenth:
Tested/Checked the PR.

Issue #8265, opened 2012-07-11
make User Mgmt page search on Comments field


sepiariver:
sepiariver created Redmine issue ID 8265

And/or the other fields in modx_user_attributes. "Phone number" and "Postal Code" is often asked for.

PR #13551, opened 2017-07-14
Improve search in user management

alroniks:
### What does it do?
It adds new option to search field - key for search. Query for searching by key should be like that `zip:99` or `phone:45678`. It works similar to search on github (`is:issue is:open sort:updated-asc `). Value after colon additionally wrapped by `LIKE '%%'`, that does search more simple.

### Why is it needed?
Sometimes need to find a user by any meta info that was known about this user. Not only username or email, but maybe phone, or address or any other field. Now search uses only visible fields: username, fullname and email. With the new solution, we can search by any Profile field.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/8265

### DIFF
--- a/core/model/modx/processors/security/user/getlist.class.php
+++ b/core/model/modx/processors/security/user/getlist.class.php
@@ -37,13 +37,21 @@ class modUserGetListProcessor extends modObjectGetListProcessor {
     public function prepareQueryBeforeCount(xPDOQuery $c) {
         $c->leftJoin('modUserProfile','Profile');
 
-        $query = $this->getProperty('query','');
-        if (!empty($query)) {
-            $c->where(array(
-                $this->classKey . '.username:LIKE' => '%'.$query.'%',
-                'OR:Profile.fullname:LIKE' => '%'.$query.'%',
-                'OR:Profile.email:LIKE' => '%'.$query.'%',
-            ));
+        $queryChunks = explode(':', $this->getProperty('query',''));
+        if (count($queryChunks) == 2) {
+            list($field, $query) = $queryChunks;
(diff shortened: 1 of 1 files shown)


core/model/modx/processors/security/user/getlist.class.php
```diff
@@ -37,13 +37,21 @@ public function initialize() {
...
 
-        $query = $this->getProperty('query','');
-        if (!empty($query)) {
-            $c->where(array(
-                $this->classKey . '.username:LIKE' => '%'.$query.'%',
-                'OR:Profile.fullname:LIKE' => '%'.$query.'%',
-                'OR:Profile.email:LIKE' => '%'.$query.'%',
-            ));
+        $queryChunks = explode(':', $this->getProperty('query',''));
+        if (count($queryChunks) == 2) {
```
pixelchutes: Cool idea! Identical operator `===` can be used here ^

pixelchutes (commented):
This looks pretty straightforward, great idea.

If I understand this correctly, if someone passes an invalid field (on purpose or otherwise) the resulting query/response is effectively empty, is that correct?

Alroniks:
@pixelchutes Yes, if field incorrect results will be empty. I can show the error that field is wrong but it may scary user. Not sure about that because we don't have such example in the system (i.e. such search pattern).

sdrenth:
@Alroniks Nice addition to the user management search! I've tested this PR and can say that the search is working nicely. Only comment I have is that when your searching for a non-existing profile field then no WHERE clause is added to the query and the result will return all users instead of none.

Alroniks:
@sdrenth Yes, it is. I don't want brake previous behavior because it's an unusual and new feature that has never used before.

Issue #8294, opened 2012-07-14
Filter/search on event names at plugin creation/editing


bertoost:
bertoost created Redmine issue ID 8294

It's a long list. And it will be longer and longer, where components also can use this technics.

Would be great to have searchbar at the top of the grid to search the event you're looking for.

PR #13552, opened 2017-07-14
Added filtering and search to event list during creting a new plugin

alroniks:
### What does it do?
This PR adds filtering by group and search field to top bar of the grid with event on plugin creating/updating page.

### Why is it needed?
List of events very long (180 items) and it useful when a user has the ability to filter or search events by name.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/8294


OptimusCrime:
Better diff: https://github.com/modxcms/revolution/pull/13552/files?w=1

sdrenth:
Tested this PR and can confirm that the search and filter by event groups works nicely!

Issue #13169, opened 2016-11-02
Can't install/update two packages with install options 


Jako:
### Summary
Can't install/update two packages with 'Setup Options'. The second package is never marked as installed and the first (previous) package is installed again.

### Step to reproduce
Update two packages with 'Setup Options' without reloading the workspaces manager page.

### Observed behavior
The install console repeats the install messages of the first (previous) package. The second package is not marked as installed/updated even when the installer console finishes without an error.

### Expected behavior
The second package should be installed/updated and marked as installed.

### Environment
MODX 2.5.1

Mark-H:
I've seen this behaviour from time to time, but never made the link with the installation options. I think whenever it happens, it seems to show information for the 2nd extra, however the console shows it's installing the first extra.

Jako:
It happens also during installing two packages with install options.

sottwell:
A package without install options installs fine after one with options.

At any point after installing a package with options, the previous one with options will be re-installed, even if you’ve installed other packages in between.

Alroniks:
Merged https://github.com/modxcms/revolution/pull/13557

PR #13557, opened 2017-07-19
Fixing installing two packages with setup options

Jako:
### What does it do?
Change the signature property of an existing MODx.window.SetupOptions. Otherwise the previous value will be used.

### Why is it needed?
Can't install/update two packages with 'Setup Options'. The second package is never marked as installed and the first (previous) package is installed again.

### Related issue(s)/PR(s)
#13169

### DIFF
--- a/manager/assets/modext/workspace/package.containers.js
+++ b/manager/assets/modext/workspace/package.containers.js
@@ -143,6 +143,8 @@ Ext.extend(MODx.panel.Packages,MODx.Panel,{
 				id: 'modx-window-setupoptions'
 				,signature: btn.signature || ''
 			});
+		} else {
+			this.win.signature = btn.signature || '';
 		}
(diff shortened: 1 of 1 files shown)


Alroniks:
@Jako Could you remind two any packages with setup options for checking?

OptimusCrime:
pdoTools is one I think?

sottwell:
MIGx is another one with options. Also CssSweet.

bezumkin:
@Alroniks Almost all my packages with chunks - setup options allows to update selected of them.

Jako:
FormIt, MIGX, LangRouter etc. You could even reinstall two of them, same result.

Alroniks:
Cherry-picked to 2.x https://github.com/modxcms/revolution/commit/282e751db4282abc8102a001da090859223e1627

Issue #12445, opened 2015-05-17
MySQL and utf8mb4


sottwell:
The situation with MySQL and utf8mb4 needs to be looked at. An index that is fine with MySQL's 3-byte utf8 may be too long for their utf8mb4 4-byte characters. The setup will not display warnings, and will proceed to complete the installation as if all was well, but the site will fail to load because essential tables have not actually been created.

sottwell:
So this issue is actually two issues. 

First, the Setup is not properly handling database setup errors. 

Second, fields that are being used for indices need to be made shorter, for example instead of varchar (255) they could be set to varchar (248). 

The limit of MyISAM tables for index length is 1000 bytes. utf8mb4 uses four bytes per character. 255 characters at 4 bytes each comes to 1020 bytes, while 250 characters makes it 1000 on the nose. Just to be on the safe side, 248 would be a nice comfortable number.

For example the modDbRegisterTopic table uses its 'name' field as an index, and the 'name' field is set to varchar (255), thus the table will fail to be created. Change that to varchar (248) and there's no problem.

OptimusCrime:
Kind of depends if it is a goal for Modx to support this charset. Modx could specify that they don't support it and everyone can keep using `utf8-general-ci` without problems.

I don't understand why using `utf8mb4` is a case at all. Is it better than `utf8-general-ci` or something? As far as I know, `utf8-general-ci` is the mostly used charset.

sottwell:
This bypasses the point that MySQL's utf8 is a non-standard 3 bytes per character, while the utf8 standard is 4 bytes per character. If you want to use any characters from that upper register, you're out of luck.

sottwell:
Anyway, that doesn't address the problem of the Setup not reacting to the failure of those tables to be created. It logs the errors in the setup logs, but doesn't do anything about it. You wouldn't know there was a problem until you try to log in, and can't.

exside:
This is serious...just wasted an hour finding out that this was the issue^^...setup log in core/cache/logs/ helps if you know where to look...

here's a complete dump of the setup log (2.5.0-rc1 with utf8mb4_unicode_ci charset on the table):
```
[2016-03-08 16:06:10] (ERROR @ /path/to/setup/includes/tables_create.php : 116) Could not create table `modx_access_policies`
[...]

PR #13559, opened 2017-07-24
Reduce varchar and text index prefixes for utf8mb4 support in mysql

opengeek:
### What does it do?
Reduces varchar field sizes and prefixed text indexes to 191 from anything larger.

### Why is it needed?
UTF8MB4 support in MySQL has a maximum index length and page size of 191 characters when using InnoDB.

### Related issue(s)/PR(s)
#12445

### DIFF
--- a/core/model/modx/mysql/modaccesspermission.map.inc.php
+++ b/core/model/modx/mysql/modaccesspermission.map.inc.php
@@ -30,7 +30,7 @@ $xpdo_meta_map['modAccessPermission']= array (
     'name' => 
     array (
       'dbtype' => 'varchar',
-      'precision' => '255',
+      'precision' => '191',
       'phptype' => 'string',
       'null' => false,
       'default' => '',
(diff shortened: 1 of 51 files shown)


pixelchutes:
Nice work, @opengeek! To confirm, is this strictly for new installs only, or do the changes to the map/schema files automatically get applied when updating, too?

digitalpenguin:
I'm so happy about this! :)

opengeek:
@pixelchutes If you want to migrate an existing site to InnoDB or utf8mb4 (or both), I would create a clean new install and use teleport (or a mysqldump that truncates data but does not include the table structures) to move the data.

OptimusCrime:
It would perhaps be nice if a Teleport example was given, as I have no experience with that tool what so ever, and it seem to be the recommended tool for this job.

opengeek:
There is documentation for teleport. It was also not the only solution given here. A simple mysqldump that truncates data but does not include structures will work.

PR #13560, opened 2017-07-25
Loading event log via ajax only

bezumkin:
### What does it do?
This PR will force errors log to load data only with ajax.

### Why is it needed?
Because sometimes we can get a blank screen due to characters collide. Just put this string into your `core/cache/logs/error.log`:
```
<!--]<script>
```
And it will break both error log and edit page for this file because browser can`t parse it correctly

screen
You can see on this screenshot, that tags highlight is broken.

We can use htmlspecialchars function, but it will change html tags in log. But in the other hand we already have the function to load log via ajax. Why don`t use it for initial load?

### DIFF
--- a/manager/assets/modext/sections/system/error.log.js
+++ b/manager/assets/modext/sections/system/error.log.js
@@ -33,6 +33,7 @@ MODx.page.ErrorLog = function(config) {
         }]
     });
     MODx.page.ErrorLog.superclass.constructor.call(this,config);
+    this.refreshLog();
 };
 Ext.extend(MODx.page.ErrorLog,MODx.Component,{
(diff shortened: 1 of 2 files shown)


Alroniks:
It started from 2.x, but tagged for 2.5.x. Does it correct @Jako ?

Jako:
Since it is a bugfix, I thought 2.5.x would be ok.

Alroniks:
@Jako In this case, I'll do backport from 2.x

Alroniks:
Cherry-picked into 2.5.x https://github.com/modxcms/revolution/commit/8fb82b715c071439c417eaf7432089f020b19208

PR #13563, opened 2017-07-30
Fix height of error log

bezumkin:
### What does it do?
This PR fixes height of the error log textarea

### Why is it needed?
Because after this fix #13560 textarea was always small. So I removed unnecessary afterrender action and add call of setTextareaHeight() after each successful ajax load.

### Related issue(s)/PR(s)
#13560

### DIFF
--- a/manager/assets/modext/sections/system/error.log.js
+++ b/manager/assets/modext/sections/system/error.log.js
@@ -33,6 +33,7 @@ MODx.page.ErrorLog = function(config) {
         }]
     });
     MODx.page.ErrorLog.superclass.constructor.call(this,config);
+    this.refreshLog();
 };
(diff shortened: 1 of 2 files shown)


manager/assets/modext/widgets/system/modx.panel.error.log.js
```diff
@@ -81,7 +73,8 @@ Ext.extend(MODx.panel.ErrorLog,MODx.FormPanel,{
      *
      * @param {Ext.TextArea} elem
      */
-    ,setTextareaHeight: function(elem) {
+    ,setTextareaHeight: function() {
+        var elem = Ext.getCmp('modx-error-log-content');
```
rtripault: Since this is no more a parameter, would you mind updating the JSDoc please ?

PR #13565, opened 2017-07-31
Save manager states in HTML5 sessionStorage for browser tabs

labr1005:
### What does it do?
**Additionally** save manager states in HTML5 sessionStorage for browser tabs.

### Why is it needed?
This preserves for example the active state of the "Resources/Elements/Files"-tabs when working in multiple browser tabs.

### Notes
This will need testing.

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -782,11 +782,19 @@ Ext.extend(MODx.HttpProvider, Ext.state.Provider, {
         if (!name) {
             return;
         }
+        // additionally save state in HTML5 sessionStorage for this browser tab
+        if (typeof(Storage) !== 'undefined') {
+            sessionStorage[name] = Ext.encode(value);
+        }
         this.queueChange(name, value);
     }
     ,get : function(name, defaultValue){
-        return typeof this.state[name] == "undefined" ?
-            defaultValue : this.state[name];
+        var output = typeof this.state[name] == "undefined" ? defaultValue : this.state[name];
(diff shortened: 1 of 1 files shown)


Mark-H:
You say:

> This preserves for example the active state of the "Resources/Elements/Files"-tabs when working in multiple browser tabs.

But MDN ([Source](https://developer.mozilla.org/en/docs/Web/API/Window/sessionStorage)) says:

> **Opening a page in a new tab or window will cause a new session to be initiated**, which differs from how session cookies work.

Are you trying to get different states for different tabs?

labr1005:
Yes.

When you are currently working in MODX in multiple browser tabs, changing from "Resources" to "Elements" in one browser tab will change that in every other browser tab when reloading, saving, etc. that other browser tab.

This will save and retrieve the states for every browser tab individually, so you will stay in the "area" you are working on in that specific browser tab.

labr1005:
But it's just an addon if sessionStorage is available. Everything will still be saved to core/cache/registry/state/ys/user-{USER-ID}/ for new tabs and if sessionStorage is not available.

rtripault:
Always been thinking about that. Using any browser storage should be better (performance wise) as it would prevent writing states to the filesystem (as well as a few ajax calls).

The drawback we might face is that states would only be kept in the current client browser (as opposed to the current situation where the state are restored if you use another browser.

I guess `MODx.HttpProvider` & `Ext.state.Provider` would be the places to check to implement a different provider (in case someone wants to give a try at this)

labr1005:
Closing this in favour of rtripault's approach.

Issue #12786, opened 2015-11-29
Remember state of collapsed top tab area


rainbowtiger:
When editing, especially on smaller screens (well, anything but really large screens actually), it's useful to collapse the top tabbed area to have the content area show up near the top of the window. It would be even more useful to use a cookie and have MODX remember the setting for collapsing and expanding the top tabbed area, so that a user can collapse the top tabs in order to see the content area by itself, and then move from resource to resource to edit content, maintaining the collapsed tabs without have to click the up arrow for every single resource.

jpdevries:
+1
I think we could use localStorage instead of a cookie

rainbowtiger:
That sounds better.

Mark-H:
There's already states built in to extjs, which gets persisted to the cache, might make more sense to use that to stay consistent with how things are stored.

(Though I wouldn't be opposed to moving that system to localstorage when supported)

rtripault:
Both aren't incompatible... building a local storage state provider seems doable (see https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/core/modx.js#L679).
Agreeing with Mark, using the shipped state provider is definitely the way to go, from my seat.

Issue #8769, opened 2012-09-25
Page-Tree looses state in multi-context websites


poseidon-zz:
poseidon created Redmine issue ID 8769

[I searched the tracker but didn't find this bug already filed, just a related problem (Bug #8749) seems to be filed.]

Doing several tests for a multisite/multilingual installation, I set up several contexts for different domains and languages. besides that, the "Web"-Context remains in the tree, too. "Web" is expanded by default, the other contexts aren't.

opening any resource inside a context that is not "Web", lets the manager reload and display the default context open, but the resource open for editing is hidden in the tree. It isn't shown until some changes in the resource are saved.

I think this is annoying if I had to use the tree on a daily basis.

Solution:
1) Opening a resource from the resource-tree should not change the shown view of the tree. even if resources are expanded that are not related to the recource currently edited.
2) opening a context, which was collapsed before, should show it, as if it was the default context: fully expanded. if other contexts are closed so that only one is open should be chooseable by the user. (default: allow multiple contexts open)

Long-term-solution:
I recently played around with wordpress a little: architecturally its spooky, but the backend is really nice: fully AJAXed and needs no reload for the whole page. this would be nice for odx as well, as i think it would get slightly faster. no need to redraw the page-tree, etc. (which is irritating, anyways).

Feel free to ask if something isn't clear. [Apologies for my not-so-good english.]

Poseidon

gpsietzema:
+1 in 2.5.5

Ibochkarev:
No open fix PR for this (closed https://github.com/modxcms/revolution/pull/13568 was about the state provider class, not this tree UX).

Multi-context resource tree state still rides Ext stateful tree behavior (`manager/assets/modext/widgets/resource/modx.tree.resource.js` / `MODx.tree.Tree`). Opening a resource reloads the manager and Web tends to come back expanded as the default, so the edited node stays hidden until save - same pain @manu37 and @gpsietzema (via hugopeek) described for Elements too. Nobody left a reliable repro since 2017.

I have not reproduced a multi-context manager session here.

Keep open as UI debt: preserve expanded context/path across resource open, or expand to the active resource after load. Needs a current 3.x repro before a focused PR.

PR #13568, opened 2017-08-01
Allow users to define the state provider class to use

rtripault:
### What does it do?

Added a new configuration parameter to define the class to be used as state manager/provider, as well as wrapping the state manager/provider into a dedicated method so i could be overridden in custom layouts.

### Why is it needed?

This would allow customization of the state provider, for example to use localStorage/sessionStorage instead of the current implementation.

### Related issue(s)/PR(s)

Would allow one to ship her own implementation for #13565 

Possibly related to #12786 & #8769

### DIFF
--- a/manager/assets/modext/core/modx.layout.js
+++ b/manager/assets/modext/core/modx.layout.js
@@ -22,9 +22,7 @@ MODx.Layout = function(config){
     MODx.siteId = config.auth;
     MODx.expandHelp = !Ext.isEmpty(MODx.config.inline_help);
 
-    var sp = new MODx.HttpProvider();
-    Ext.state.Manager.setProvider(sp);
-    sp.initState(MODx.defaultState);
+    this.initStateProvider();
 
     config.showTree = false;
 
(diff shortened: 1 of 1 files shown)


manager/assets/modext/core/modx.layout.js
```diff
@@ -52,13 +50,22 @@ MODx.Layout = function(config){
 };
 Ext.extend(MODx.Layout, Ext.Viewport, {
     /**
+     * Initialize the state provider to use
+     */
+    initStateProvider: function() {
+        var provider = MODx.config.stateprovider || MODx.HttpProvider,
```
rtripault: To customize the class, we would have to manually set the `MODx.config.stateprovider`

manager/assets/modext/core/modx.layout.js
```diff
@@ -52,13 +50,22 @@ MODx.Layout = function(config){
...
+            sp = new provider();
```
rtripault: We might want to make sure `provider` is an available class, else fallback to default `MODx.HttpProvider`

manager/assets/modext/core/modx.layout.js
```diff
@@ -52,13 +50,22 @@ MODx.Layout = function(config){
...
+        Ext.state.Manager.setProvider(sp);
+        sp.initState(MODx.defaultState);
```
rtripault: `initState` is bound to the `MODx.HttpProvider` not a method shipped with [`Ext.state.Provider`](http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.state.Provider), so we might want to make sure the method/function exists before calling it.

While being there, we might also want to add an option to disable the manager controllers to check for the default states (see `modManagerController::getDefaultState`)

rtripault:
Just to note that this is not really "in progress", but rather a proposal to see if there is any interest (no need to address the notes if no interest in the PR)

PR #13572, opened 2017-08-14
Fix too aggressive cache clearing when triggering an autopublish event

Mark-H:
### What does it do?

Changes the cache clear that happens automatically when a resource is automatically published, or unpublished, to be more restrictive. It was being too aggressive compared to the [standard cache clear that happens when saving a resource from the manager](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/resource/update.class.php#L741-L755).

### Why is it needed?

Previously the autopublish would clear all default cache providers, including context_settings for all contexts, system_settings, lexicons, resources etc. With this patch, it brings the cache clear in line with what happens when you save a resource, by only clearing the relevant context_settings (because the aliasMap needs to be updated when a resource is (un)published), resource, and the db cache.

The reason for this change is high-traffic sites with lots of contexts. The particular client I worked for this on has 300 (o.O) contexts with probably over 100.000 resources (I didn't count them). When a front-end request triggers an auto publish, it would literally take the site down for a couple minutes as it regenerates all context caches (looping over all resources in all contexts to generate the aliasMap). 

Because multiple concurrent requests would hit the site, they'd all get stuck trying to generate those caches causing the error log to fill up with caching errors as they can't get a file lock (as another request is already doing the same thing).

It's been tricky to test this in a simulated environment (I couldn't get enough traffic to hit my local dev server to consistently trigger the problem), but the fix has been verified on the client production site. Where previously an auto publish would cause the server load to jump from 0.4 to 70 and it would take 5-10 minutes for the site to come back online, with the fix applied there has not been a noticeable increase when a page gets (un)published, making it as quick as ever.

### Related issue(s)/PR(s)

None. 

Special thanks to Peter McAndrew at the Green Party for commissioning this fix, and allowing it to be made open source so others can benefit from it as well.

### DIFF
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -468,17 +468,45 @@ class modRequest {
     }
 
     /**
-     * Checks the current status of timed publishing events.
+     * Checks the current status of timed publishing events and automatically (un)publishes resources if needed.
      */
     public function checkPublishStatus() {
+        $partKey = $this->modx->getOption('cache_auto_publish_key', null, 'auto_publish');
+        $partHandler = $this->modx->getOption('cache_auto_publish_handler', null, $this->modx->getOption(xPDO::OPT_CACHE_HANDLER));
+        $partOptions = array(xPDO::OPT_CACHE_KEY => $partKey, xPDO::OPT_CACHE_HANDLER => $partHandler);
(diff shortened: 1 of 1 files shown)


rthrash:
👍

Issue #13573, opened 2017-08-18
Snippets and Plugins saved within MODX no longer strip the opening PHP tag


pixelchutes:
### Summary
Since the merge of #13523, Snippets (and Plugins) saved **within MODX** no longer strip the opening php tag: `<?php`

The resulting cached snippet starts throwing this error:

> NOTICE: PHP message: PHP Parse error:  syntax error, unexpected '<'

This happens due to this invalid syntax at the top:

```php
<?php
<?php
// ...
```

### Step to reproduce
1. Open a Snippet in MODX `2.5.8-dev` (or `2.6.x`)
2. Save the snippet
3. Browse to site Resource calling Snippet in question
4. See 500 error

### Observed behavior
500 error is thrown:

> NOTICE: PHP message: PHP Parse error:  syntax error, unexpected '<'

### Expected behavior
Opening PHP tag(s) should be stripped as before.

### Environment
MODX 2.5.8-dev

pixelchutes:
Fixed by #13576

PR #13576, opened 2017-08-29
Added plugins and snippets to array to strip php tags from

sdrenth:
### What does it do?
Readded plugincode and snippet to the array in modscript which handles the removal of php start and end tags to prevent 500 errors on the frontend caused by duplicate PHP tags in the code.

### Why is it needed?
These types where removed from this array in merge https://github.com/modxcms/revolution/pull/13523 and was causing 500 errors in the frontend as described in issue https://github.com/modxcms/revolution/issues/13573

### Related issue(s)/PR(s)
Related issue: https://github.com/modxcms/revolution/issues/13573
Related PR: https://github.com/modxcms/revolution/pull/13523

### DIFF
--- a/core/model/modx/modscript.class.php
+++ b/core/model/modx/modscript.class.php
@@ -37,7 +37,7 @@ class modScript extends modElement {
      * {@inheritdoc}
      */
     public function set($k, $v= null, $vType= '') {
-        if (in_array($k, array('content'))) {
+        if (in_array($k, array('snippet', 'plugincode', 'content'))) {
             $v= trim($v);
             if (strncmp($v, '<?', 2) == 0) {
                 $v= substr($v, 2);


pixelchutes:
@sdrenth Thanks for submitting this PR + fix!

I am curious, looking at the issue the original PR #13523 was trying to solve, would it be better if we simply _reverted_ the `2.5.x` and `2.x` commits? _(...and I guess any in `3.x` that were likely merged, too)?_

I only ask this because while this PR will solve the regression that was introduced, it also re-introduces the original `trim()` being bypassed for Snippets and Plugins serving as part of the solution for #12430. That's why I'm debating perhaps we revert the original commits, and instead look at another angle for solving preserving trailing newlines in PHP code managed within MODX. (One example could be checking if one exists, performing all previous trimming as before, and then restoring the EOF newline if necessary.

sdrenth:
@pixelchutes I agree that that would be a nicer solution, that way the issue https://github.com/modxcms/revolution/issues/12430 can be reopened so we can figure out an other solution to solve that issue.

I'm keeping this PR open, but can someone please revert this PR https://github.com/modxcms/revolution/pull/13523 and close this PR? And then ofcourse reopen issue https://github.com/modxcms/revolution/issues/12430?

Mark-H:
Merged into 2.5.x per https://github.com/modxcms/revolution/commit/fa1283c2e9091bf94af421da4a222bc2acc567a3
Cherry-picked out of 2.x per https://github.com/modxcms/revolution/commit/6fc6fa7bda7c1623ab8e85fbc20313d5b578de82

OptimusCrime:
This means that we effectively reverted #12430 with this PR? Just to make sure we are all on the same page here.

Issue #13488, opened 2017-07-06
Session expired after migrate


MrRoco:
### Summary
After migrating sites to new environment we keep on getting complains about people being logged out all the time. When they try to log back in they get stuck in a loop.Clearing the browser cache helps for a few days but the problem keeps on coming back.

### Step to reproduce
Migrate sites to new environments

### Observed behavior
After the migration everything works fine, after a few days users get a session expired popup, when trying to log back in, they get stuck in a loop and keep on getting the popup.

### Expected behavior
Being able to login and use the manager till logging out.

### Environment
MODX 2.3+, Modxcloud & nginx, php 5.6 & php 7.1, all

Mark-H:
Tried clearing the cookies? Are the session settings correct (path/httponly/secure)? Paths correct?

OptimusCrime:
I kept getting this error because my site was behind basic auth. Do you use it too?

gpsietzema:
Some additional data/comments-thread from a support thread:

> I haven't tested it with a brand new cloud. After reinstalling the other clouds I did like 200 logins and it didn't happen again though.

What is happening, is that HTTP_MODAUTH is empty when doing an Ajax-request to connectors/index.php

Some things I think could be a reason or related or should be checked: 
- Corrupt MODX core files (reinstall should could fix that) 
- Bad cookies (dunno how they can break or be wrong) 
- Too crowded on the server which somehow might mess up stuff 
- ACL's, but then again, they are able to login, it is just the connectors not working

I had a strange, dunno if related, thing happen: museumstadappingedam.nl (which is session-expired victim), showed a 504 page everywhere and the error.log showed max-execution-time-30secs errors everywhere. It just kept loading all over the place (frontend + manager). I replaced it with 2 day old backup and it didn't fix it. Reinstalling didn't fix it. But when I set it to PHP7.1 it worked again...

> I don't think cookies are the problem here. As for the HTTP_MODAUTH thing, I'm confused. Is that empty when the session expires or always empty? You said it didn't happen again, but that HTTP_MODAUTH is empty, so is that just an unrelated observation?

> In the cases where we got the session_expired modal, we noticed that the HTTP_MODAUTH was empty when doing ajax-requests to the connectors/index.php

> That sounds suspiciously like this might be being caused by a specific MODX extra that has a connector which is not working properly with regard to the HTTP_MODAUTH. Is there a component with a connector common to these sites that has not been updated in a while?

[...]

PR #13577, opened 2017-08-29
Bug 13488

Oetzie:
### What does it do?
This fix checks if the the 'user token session' is set or is empty (empty = 0) during the login process, if the 'user token session' is empty a new token will be set.

### Why is it needed?
Some times you get a message 'session expired' in the manager. After trying to login you get an infinity loop of messages 'Session expired'. My experience is that the 'user token session' some times gets set to 0 and that causes the 'session expired' message. The login function uses modUsers class to login, but before the user can login in the function will check if the 'user token session' is set, if yes no new token will be set. But the 'user token session' is set to 0 so no new token will be set.

Note: This fix only contains a fix the login into the manager again, not a fix why the 'user token session' gets set to 0 (thats a bug in some 3th party extra's/plugins).

### Related issue(s)/PR(s)
13488 (https://github.com/modxcms/revolution/issues/13488)

### DIFF
--- a/core/model/modx/moduser.class.php
+++ b/core/model/modx/moduser.class.php
@@ -368,7 +368,7 @@ class modUser extends modPrincipal {
             }
             $this->sessionContexts[$context]= $this->get('id');
             $_SESSION['modx.user.contextTokens']= $this->sessionContexts;
-            if (!isset($_SESSION["modx.{$context}.user.token"])) {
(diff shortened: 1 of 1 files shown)


opengeek:
this was merged, I just accidentally merged it to the wrong base branch, assuming it was against 2.5.x instead of 2.x

gpsietzema:
2.5.x was the idea ;-)

Issue #13429, opened 2017-04-19
chmod'ing a directory no longer works


Mark-H:
### Summary
Since 2.5.6 it's not possible to chmod a directory. This seems to be caused by #13352, which runs a preg_match on the result of `$this->parseMode` which runs `octdec` which transforms the value from something like `0755` to a decimal `493`, which doesn't pass that check. 

### Step to reproduce
Right click a directory in the files tree, attempt to change the chmod (folder permissions), see it fail with a generic error.

### Observed behavior
Error is thrown 

### Expected behavior
chmod should be executed

### Environment
Confirmed by 3 different people on different environments (including local installs and MODX cloud), does not appear to be server related.

Jako:
This could be solved by moving the line 212 after 216 in https://github.com/OptimusCrime/revolution/blob/f89be6e74d53aa86ce187f3bcd87db01b5ce304e/core/model/modx/modfilehandler.class.php#L212 but then the error message is wrong, when the folder can't be chmod, because of PHP/Filesystem restrictions.

Jako:
The regex validation has to be done outside of the modFileSystemResource->chmod method, i.e. in the modFileMediaSource->chmodContainer method.

Jako:
Until that #13352 should be reverted.

OptimusCrime:
@Jako Haha, why did you reference my fork?

OptimusCrime:
Oh, because of my PR. I see. I did not alter the underlying implementation, I only fixed the validation of the regex.

Jako:
But your validation uses the value of `$more`, that was changed in `$this->parseMode`. It can't be validated with the regex at that moment anymore.

PR #13580, opened 2017-08-30
Fix chmod directories and added custom error message

sdrenth:
### What does it do?
Created a new function for validating the chmod mode input value which was added in PR https://github.com/modxcms/revolution/pull/13352. Also modified chmodContainer function to call this new validate method and to return a custom error message when an invalid chmod mode was specified.

### Why is it needed?
Could not chmod a directory anymore because the preg_match validation would always fail because parseMode would modify $mode (see code below) and cause the preg_match validation to fail.

```
protected function parseMode($mode = '') {  
    return octdec($mode);  
}
````

### Related issue(s)/PR(s)
Related issue: https://github.com/modxcms/revolution/issues/13429
Related PR: https://github.com/modxcms/revolution/pull/13352


Alroniks:
It works excepting one thing. After changing a mode of folder node in the tree keeps the old value that was before. And if I want to change mode again I will see not the current value. It's not a huge issue and we agreed with @opengeek to merge it and create a new issue for the case that I described.

Issue #12465, opened 2015-06-12
Container resources are available by urls both with and without container suffix


sitex-pro:
### Summary

If we turn furls on and create a container, it will be available by both urls: site.com/container and site.com/container/ - regardless of the value of [friendly_urls_strict] setting. 
### Step to reproduce
1. Install modx 2.3.3-pl
2. rename ht.access to .htaccess
3. set [friendly_urls_strict] on
4. create container
### Observed behavior

One can access container's page both slash and without a slash at the end of url
### Expected behavior

301 redirect to the url with trailing slash should occur when accessing container's page without trailing slash
### Environment

modx 2.3.3-pl
### How to fix

file /core/model/modx/modrequest.class.php, comment out the unwanted line 332:
$identifier = "{$identifier}{$containerSuffix}";
followed by line:
$found = $this->modx->findResource("{$identifier}{$containerSuffix}");

shamarin:
On modx revo 2.5.2 it's not fixed.

shamarin:
Also main paige of site is available via index.php. It would be nice to make 301 redirect site.com/

Mark-H:
Fixed in 2.7 per #13142

Issue #13289, opened 2017-02-13
No response body when null-byte character in query string (%00)


pixelchutes:
### Summary

If passing a null-byte character (`%00`) in the URL to a MODX website, the request is immediately terminated with a `HTTP/1.1 200 OK` response, but _no_ response body.

### Step to reproduce

Browse to MODX website with null-byte character in the URL:

**Example:** https://modx.com/?%00

### Observed behavior

```c
HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/html; charset=UTF-8
Date: Mon, 13 Feb 2017 21:14:51 GMT
MC: 2Mru1ZFR/JMsNYMAR6VMFC5F3zNmMHNF7EfTrM+JyGvvVVA7Bk+CJhC8hu17ZySp
Server: nginx/1.10.0 (Ubuntu)
Transfer-Encoding: chunked
Vary: Accept-Encoding

// no response body here
```

**Example:** https://modxcloud.com/?%00

```c
HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/html; charset=UTF-8
Date: Mon, 13 Feb 2017 21:19:33 GMT
MC: RdPkGQARO+eA54v8wZoGfXPmoazL4Uc8YT/K60fe+99+jz1EZWVvKbCnpATyRlGa
Server: nginx/1.8.1
Strict-Transport-Security: max-age=31536000;
Transfer-Encoding: chunked

// no response body here
```

Testing this against the exact same server _(but bypassing MODX entirely)_ suggests that this is an internal handling issue within MODX core.

### Expected behavior

I would expect some accompanying response body if the URI was valid, and if not, then perhaps the 404 Error page, for example. Something more than just the HTTP headers.

### Environment

MODX 2.5.5, PHP 5.6.30, NGINX 1.8.* or 1.10.* or 1.11.*

Mark-H:
[Here's the cause](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L283). Seems to have been there since the first revo commit. 

Weird to just `die()` the process though, I can find some resources that say null byte injection can be a security risk, but it seems a simple str_replace on the values (perhaps in `modX::sanitize` or `modX::protect` should do the trick?

pixelchutes:
Interesting, good find @Mark-H. I did confirm this against an Evo instance, as well (same results) -- so probably originated there.

[...]

PR #13582, opened 2017-09-01
Fixing container urls and redirecting index.php to home url #12465

sdrenth:
### What does it do?
Redirects /index.php to site start URL and container resources URL's without trailing slash are now being redirected to container resource URL with trailing slash.

For example:
- https://www.example.com/index.php will be redirected to: https://www.example.com/
- When visiting https://www.example.com/container the URL will become https://www.example.com/container/

### Why is it needed?
Container resources could be visited on both url's, for example:
- https://www.example.com/container
- https://www.example.com/container/

And a user was able to visit the homepage with following URL, where it would be nicer if the user was redirected to the site start url instead of visiting the page as index.php:
- https://www.example.com/index.php

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/12465

### DIFF
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -58,12 +58,16 @@ class modRequest {
      */
     public function handleRequest() {
         /* Redirect /index.php to site start url. */
-        if (ltrim(explode('?', $_SERVER['REQUEST_URI'], 2)[0], '/') === 'index.php') {
+        $parsedUrl = parse_url($_SERVER['REQUEST_URI']);
+        $path = $parsedUrl['path'];
+        if (ltrim($path, '/') === 'index.php') {
+            $parsedQuery = $parsedUrl['query'];
+            parse_str($parsedQuery, $query);
             $this->modx->sendRedirect(
                 $this->modx->makeUrl(
                     (integer) $this->modx->getOption('site_start', null, 1),
                     null,
-                    $_GET,
+                    $query,
                     'full'
                 )
             );
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -57,6 +57,19 @@ class modRequest {
      * @return boolean True if a request is handled without interruption.
      */
     public function handleRequest() {
+        /* Redirect /index.php to site start url. */
+        if (ltrim(parse_url($_SERVER['REQUEST_URI'])['path'], '/') === 'index.php') {
+            parse_str(parse_url($_SERVER['REQUEST_URI'])['query'], $query);
+            $this->modx->sendRedirect(
+                $this->modx->makeUrl(
+                    (integer) $this->modx->getOption('site_start', null, 1),
+                    null,
+                    $query,
+                    'full'
+                )
+            );
+        }
+
         $this->loadErrorHandler();
 
         // If enabled, send the X-Powered-By header to identify this site as running MODX, per discussion in #12882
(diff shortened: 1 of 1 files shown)


Mark-H:
Aside from the test failure on PHP 5.3 (which is still supported, sorry! :P), wouldn't this be better when done as a plugin?

sdrenth:
@Mark-H What do you mean by doing it as a plugin? Making some kind of extra for it or shipping a default plugin with a fresh modx install or am I missing something? ;)

Has there been talked about upping the minimum PHP version for MODX 2.6 to 5.4? I'm asking this in relation to issue https://github.com/modxcms/revolution/issues/13289 and PR https://github.com/modxcms/revolution/pull/13581

Mark-H:
As an extra is what I meant ;)

I don't think the PHP version is planned to go up in 2.6? For the zero byte stuff we can update the minimum from 5.3.3 to 5.3.4, that's no big deal, but AFAIK there's no decision on moving to 5.4 in 2.6 yet.

sdrenth:
I think it should be in the core and should not be an extra, but I'd like to hear some more opinions about this ;)

If more people think this should be fixed in the core then I'll fix the PR to be compatible with 5.3

OptimusCrime:
I think this is a pretty cool feature. Inconsistent slashes are terrible. +1 from me.

rthrash:
I'm a fan of this also. Great job!

OptimusCrime:
Fix the test and I'll take a look at it :)

shamarin:
It should be in core by default.

sdrenth:
I've fixed the code to pass the tests

shamarin:
Great. For me it works now on php 7.1. On container resources it gives 301 redirect without slash and 302 redirect on index.php.

opengeek:
It is expected behavior with strict URL handling for containers to be redirected to the URL with the container_suffix, so that is technically a bug fix to the core if it were to depend on the setting. I'm not convinced on the index.php handling and am wondering if we could somehow separate that into a separate PR, perhaps with its own configuration setting? Those are my thoughts.

shamarin:
Why this fix is not included to 2.6.0 version and is moved to 2.7.0?

opengeek:
Because there have been some suggestions made and we are awaiting feedback from the author. It needs additional work before being accepted.

shamarin:
Maybe to include container resources bug fix in 2.6.0 version as it has long lasting high priority bug and redirect from index.php as an option to include in next release? That's my thoughts about this.

sdrenth:
@opengeek Sorry for my late reply, I could also add a system setting for redirecting the index.php and set it default to no, so that the default behaviour will not change and why cant we keep these changes in this PR instead of creating a separate PR for the index redirect?

Mark-H:
Unfortunately, this isn't ready to get merged, as it causes an infinite redirect if friendly_urls is disabled. 

> why cant we keep these changes in this PR instead of creating a separate PR for the index redirect?

It's best to keep PRs as small as possible. In this case, I can't accept any of the PR because the index.php redirect breaks non-friendly url usage. Had you sent 2 different PRs, one for the index.php and one for the container fix, I could've at least merged the container one ;)

Would you like to update the PR to fix the index.php infinite redirect @sdrenth? If not I'll go ahead and close it and wait for a new one.

Mark-H:
#13142 also fixes the container suffix handling, as does #13142.

Mark-H:
Closing as #13142 has been merged that resolves the container issue - please submit a new PR for other issues.

PR #13583, opened 2017-09-01
Readding changes from reverted PR #13044

sdrenth:
## Introduction
Readded this code because it was merged and reverted because the code could not be merged into version 2.5 because it's a breaking change. Therefore created a new PR for this so it can be merged into version 2.6.

From original PR https://github.com/modxcms/revolution/pull/13044:

### What does it do?

Call processElementTags() in the modParser always with the $maxIterations / $depth value from the parser_max_iterations system setting (default 10), instead of the "function default" of 0.

### Why is it needed?

Fixes a parser bug for special nested tags. See #13043 for steps to reproduce the bug.

### Related issue(s)/PR(s)

Directly related to the issues reported in #13043

I've tested it already on 1 production site where it fixed my issue from #13043 and didn't trigger any other issues so far.

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -428,7 +428,8 @@ class modParser {
         }
 
         /* collect any nested element tags in the innerTag and process them */
-        $this->processElementTags($outerTag, $innerTag, $processUncacheable);
+        $maxIterations= intval($this->modx->getOption('parser_max_iterations',null,10));
+        $this->processElementTags($outerTag, $innerTag, $processUncacheable, $this->modx->parser->isRemovingUnprocessed(), "[[", "]]", array (), $maxIterations);
         $this->_processingTag = true;
         $outerTag= '[[' . $innerTag . ']]';
 


Mark-H:
Assigned Jason to this as he indicated having issues with the fix when previously applied, and wanted to look at it for 2.6.

OptimusCrime:
The test needs to be edited too, as it uses the current (upstream) behavior of the parser.

opengeek:
While this does address the "bug" in question it also breaks the most basic of parser functionality. Without some additional work, this cannot be accepted into 2.x as it does break BC. Changing a functioning unit test does not get past that fact.

OptimusCrime:
@opengeek Could it be fixed by increasing the default depth? What version does the parser operate with today?

opengeek:
Changing the depth does not solve it, no. This PR changes the most basic functionality of the parser and I cannot find a way to solve the problem being addressed here without breaking this fundamental behavior.

Mark-H:
Closing as there's not currently a workable solution.

Issue #9086, opened 2012-11-09
Invalid Password Specified : Inability to set Strong passwords


modxbot:
everettg_99 created Redmine issue ID 9086

The manager is not allowing certain passwords, e.g.

<pre>
juju59)sails
</pre>

Seems the closing parenthesis is the offender.

It's important that users be able to set strong passwords.  Not sure why this error is popping up but it's occurred on 2 separate installs of 2.2.5-pl.

hackel:
No resolution to this in over 3 years?  Really?

This is quite a pretty serious bug.  Not only should there be no restrictions on the characters allowed in a password (any utf-8 character should be valid), there is currently NO indication in the UI as to which characters are forbidden.  This requires users to either just guess or hunt down this bug report.  Ridiculous.  Most of us use password managers of some kind now that generate strong passwords like e.g. `Nf#z0->]POvn}ueD7yEa<v:j\F>4uKJ@` which Modx silently rejected.

pyrographics:
I know this is an old ticket but thought this may be related so consider this a bump.  MODX 2.5 accepted a > character in the password on setup but was not able to login after using it.  I've encountered this several times.

Jako:
Since the entered password does not have to be shown again, the htmlspecialchars restriction does not count here to avoid xss.

opengeek:
I don't believe this is why the restriction is in place @Jako — I believe this is somehow related encoding certain Javascript data for ExtJS in the connector responses — I can't recall exactly where the problem is but for some reason I believe it has everything to do with https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modconnectorresponse.class.php#L223

Jako:
The reason seems terribly logical. We need a check in the installer then and some message in the password change field.

BTW: The following ExtJS combo code should be quite nice to generate a new password with char restrictions.

```
MODx.combo.RandomString = function (config) {
    config = config || {};
    Ext.applyIf(config, {
        randomLength: 30,
        randomChars: '123456789abcdefghiklmnopqrstuwxyz',
        triggerAction: 'all',
        triggerClass: 'x-form-random-trigger',
        isUpdate: config.isUpdate,
        inputType: (config.isUpdate) ? 'password' : 'text'
    });
    MODx.combo.RandomString.superclass.constructor.call(this, config);
    this.config = config;
    if (!this.config.isUpdate) {
        this.setValue(this.createRandomString(this.config.randomLength, this.config.randomChars));
    }
};
[...]

PR #13584, opened 2017-09-01
Show validation errors when setting a new user password

sdrenth:
### What does it do?
Made password related error messages visible when changing the users password and made the error message for an invalid password more descriptive. Also I've made the invalid character checks in the setup for the username and password equal to the checks which are done when creating a new user in the manager and also replaced the setups invalid password validation error message.

### Why is it needed?
- When an invalid password was provided, the error message was only visible as a tooltip, now the error message is being displayed right below the password input field.
- Made the invalid password message more descriptive to inform the user about which characters are invalid/valid to be used for a password
- Password validation where different for when creating a password via setup or via user management.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/9086


sdrenth:
I'm sorry, had a unrelated commit to this issue added in the PR. I've reverted the unrelated commit.

Mark-H:
Maybe you could squash your commits or create a new PR that doesn't have the unrelated commits?  The tests seem to have failed on your specific commit, tho maybe that had to do with the other patch.

sdrenth:
@Mark-H I've created a new PR for this https://github.com/modxcms/revolution/pull/13585

PR #13585, opened 2017-09-01
Show validation errors when setting a new user password

sdrenth:
## Introduction
Created a new PR, because PR https://github.com/modxcms/revolution/pull/13584 contained a wrong commit.

### What does it do?

Made password related error messages visible when changing the users password and made the error message for an invalid password more descriptive. Also I've made the invalid character checks in the setup for the username and password equal to the checks which are done when creating a new user in the manager and also replaced the setups invalid password validation error message.

### Why is it needed?

When an invalid password was provided, the error message was only visible as a tooltip, now the error message is being displayed right below the password input field.
Made the invalid password message more descriptive to inform the user about which characters are invalid/valid to be used for a password
Password validation where different for when creating a password via setup or via user management.

### Related issue(s)/PR(s)
PR https://github.com/modxcms/revolution/pull/13584
Issue #9086


Jako:
The error message is somehow wrong: accent characters are possible too. I would use the setup/lang/en/default.inc.php message.

But is the regex right? According to my test on https://regex101.com/r/DutwFe/1 it should be `^[^\'\x3c\x3e\(\);\x22]+$`

Mark-H:
Any thoughts on @Jako's review, @sdrenth?

sdrenth:
@Jako Sorry for my late response, I've now used the invalid password error message from setup.

The regex was false indeed, it would still allow certain characters like '{' and '}'. Your regex also missed a few characters, therefore I've updated the regex so it matches all characters as provided in the invalid password message.

bezumkin (approved):
Tested and it works.

screen shot 2018-04-23 at 11 22 24

opengeek (approved):

PR #13588, opened 2017-09-04
Add system setting to clear resource cache partially for xPDOFileCache

OptimusCrime:
### What does it do?
Added a system setting where you can enable to clear resource cache partially by the provided contexts provided to the cache refresh method.

### Why is it needed?
When updating a resource the resources cache for all contexts would always be cleared. This option makes it possible to clear the cache only for the current active context. By default the new options is set to false, so it will not change anything unless the system setting is changed.

### Related issue(s)/PR(s)
#13196
Reimplemented from #13336

### Notes
Because @sdrenth lost the original remote repository he was unable to finish the work. I therefore copied the code he wrote + the final comment he made. This code was not written by me. Also note that this is untested, as I have little knowledge about how to test this properly.

See discussions in the original PR #13336

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -285,6 +285,15 @@ $settings['cache_resource_expires']->fromArray(array (
   'area' => 'caching',
   'editedon' => null,
 ), '', true, true);
+$settings['cache_resource_clear_partial']= $xpdo->newObject('modSystemSetting');
+$settings['cache_resource_clear_partial']->fromArray(array (
+    'key' => 'cache_resource_clear_partial',
+    'value' => 0,
+    'xtype' => 'combo-boolean',
+    'namespace' => 'core',
+    'area' => 'caching',
+    'editedon' => null,
+), '', true, true);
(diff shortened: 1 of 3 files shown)


opengeek:
This needs additional work. For this to work, the resource cache partition would need to be divided into a partition per context. Otherwise, this would only ever work *if* the cache provider supported the removal of cache entries by path. Those not using cache providers that supported this would then experience unexpected behavior.

OptimusCrime:
This is only executed if the caching is xPDOFileCache.

opengeek:
Yes, that is the entire problem I'm speaking of.

Mark-H:
As there is no proper way to handle this in memcache (and probably other cache handlers), I'm okay with the way this is currently restricted to only operate on the xPDOFileCache. It's not great, but neither is having to set up lots of different partitions for each context if you use memcache, and it's a nice improvement to have the ability to only clear a partial cache. Especially as it's opt-in for people that know what they're doing.

OptimusCrime:
I agree, @Mark-H

bezumkin (approved):
Works as expected

Issue #13482, opened 2017-07-02
parser can be null in modScript class


BobRay:
### Summary
Under certain conditions, this code in the modScript.class.php file (line 81), the parser is null and causes a fatal PHP error at the second line below: 

```
 $maxIterations= intval($this->xpdo->getOption('parser_max_iterations',null,10));
 $this->xpdo->parser->processElementTags(
```

### Step to reproduce
This sometimes occurs for me when  a snippet is called with runSnippet()  from a form processing snippet (in MyComponent) and the snippet called with runSnippet() then runs code from an included file, and there are no tags in the output. 

The included code is a build.transport.php file to build an extra. The error only occurs for some extras, but I can't see any reason why. The original snippet runs fine from the command line always, for all extras, probably because it instantiates MODX.

### Solutions

There are two simple solutions:

1) 
```
 $maxIterations= intval($this->xpdo->getOption('parser_max_iterations',null,10));
 $this->xpdo->getParser();
 $this->xpdo->parser->processElementTags(
```

2)
```
 $maxIterations= intval($this->xpdo->getOption('parser_max_iterations',null,10));
 if (!isset($this-xpdo->parser)) {
     $this->xpdo->getParser();
 }
 $this->xpdo->parser->processElementTags(
```

I've tested both, but I'm not sure which one is better.

OptimusCrime:
For simplicity, related code in [modscript.class.php#L80-L90](https://github.com/modxcms/revolution/blob/ba369815b85f32702c009b505c7de35827d19105/core/model/modx/modscript.class.php#L80-L90).

What about:

```
if ($this-xpdo->parser === null) {
     $this->xpdo->getParser();
 }
```

Using isset to check if a instance variable is set is unorthodox (to me).

BobRay:
The problem is that === null will throw an E_NOTICE  error if the variable is ever undefined and the speed difference with isset() is about a millionth of a second.

Using ! isset() is perfect because it returns false if the variable is null or unset, which is exactly what we want. And it's way faster than empty().

Strangely, I edited the post above to take out the first solution after I benchmarked the if(!isset()) test at .00000016 seconds. I saved it but apparently the edit didn't take.

snoopdouglas:
Just encountered this problem as well, and (coincidentally) found the same solution (@BobRay's first one).

[...]

PR #13593, opened 2017-09-04
Add check to make sure the MODX parser is available before calling methods on it

OptimusCrime:
### What does it do?
There are calls done to `$this->xpdo->parser->processElementTags()` before we have made sure that the parser is loaded and initialized.  This PR adds a check and loads the parser if it is null (uninitialized). 

### Why is it needed?
Could result in a fatal error, as reported in the original issue.

### Related issue(s)/PR(s)
#13482

### DIFF
--- a/core/model/modx/modscript.class.php
+++ b/core/model/modx/modscript.class.php
@@ -78,6 +78,12 @@ class modScript extends modElement {
                 if ($this->_output && is_string($this->_output)) {
                     /* collect element tags in the evaluated content and process them */
                     $maxIterations= intval($this->xpdo->getOption('parser_max_iterations',null,10));
+
+                    // Make sure we have the parser available
+                    if ($this->xpdo->parser === null) {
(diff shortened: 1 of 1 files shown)


core/model/modx/modscript.class.php
```diff
@@ -78,6 +78,12 @@ public function process($properties= null, $content= null) {
                 if ($this->_output && is_string($this->_output)) {
                     /* collect element tags in the evaluated content and process them */
                     $maxIterations= intval($this->xpdo->getOption('parser_max_iterations',null,10));
+
+                    // Make sure we have the parser available
+                    if ($this->xpdo->parser === null) {
```
opengeek: This should never happen, and if it does it would have to be something setting the parser to null explicitly. `modScript::process()` calls `parent::process()` which already explicitly loads the parser service. See https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modelement.class.php#L252.

Issue #12319, opened 2015-02-01
[2.3.x] TVs not sortable by template access


whitebyte:
### Step to reproduce

Open any TV in the manager. Go to "Available for template" tab. Try to sort by "Access" column (the last one).
### Observed behavior

The whole grid becomes empty (all rows become invisible)
### Expected behavior

Rows should be sorted in a regular way.
### Environment

2.3.2-pl, 2.3.3-pl traditional, PHP 5.5.18

Issue #7896, opened 2012-05-16
Attempting to sort the Template Variable Access tab fails on Description and Access fields


wshawn:
wshawn created Redmine issue ID 7896

The description is simply ignored, as it appears to be disabled.

Attempting to sort on Access completely blanks the form.

wshawn:
Still exists in 2.4.0 GO to the Template access tab and attempt to sort on the description or access fields.
The Access column will blank the display area, while the description column does nothing.

Reclicking the name column repopulates the page.

alroniks:
PR is merged and not reproducible on 2.7.0-dev, so closing.

PR #13597, opened 2017-09-05
Fixes TV > Tpl Access sorting bug

davidpede:
### What does it do?
Fixes v old bug by setting remoteSort to false in the 'TV>Template Access' grid

### Why is it needed?
When sorting by the Access column the processor returns no results and xpdo logs an unknown column error. This because the access value is set manually in the processor after the db query as it is not stored in db table: https://github.com/modxcms/revolution/blob/5c59d5bc4f2ffd77454d2a42a1d1046ba2f7d808/core/model/modx/processors/element/tv/template/getlist.class.php#L101

### Related issue(s)/PR(s)
Fixes: #12319, #7896


Jako:
Instead of disabling remote sort, the access column should not be sortable in https://github.com/davidpede/revolution/blob/ed8038bfdc75d840d2d624a430a665b725748b88/manager/assets/modext/widgets/element/modx.grid.tv.template.js#L15

Jako:
But before you disable sortable of the access column: How expensive would it be to create a sortable `template_access` column in the query and reference it in the grid?

davidpede:
@Jako How does last commit look for a quick solution? It seem to work as `tmplvarid` is either the current tv id or null for that query

core/model/modx/processors/element/tv/template/getlist.class.php
```diff
@@ -27,6 +27,11 @@ public function initialize() {
             'dir' => 'ASC',
             'tv' => false,
         ));
+        if ($this->getProperty('sort') == 'access') {
+          $this->setProperties(array(
```
Jako, line 31: $this->setProperty('sort', 'tmplvarid');

Jako (commented):
Should be almost suitable. But since `tmplvarid` could be an integer, the sorting depends not only on true/false but on the integer value.

You could add a select query with 
```
$c->select(array('access' => 'IF(`TemplateVarTemplates.tmplvarid`,true,false)'));
```
in the getData method, but I don't know wether this is only MySQL syntax.

Mark-H:
Sounds this is still pending improvements, @davidpede?

davidpede:
@Mark-H Yes leave this for now, I'll look at jako's suggestions asap

gpsietzema:
@davidpede MODX 2.7 is coming soon, do you see any chance to change this PR on short term?

davidpede:
Can look at this next week if soon enough? @gpsietzema

gpsietzema:
That would work if it is early next week @davidpede - Jason is planning on merging everything in this and next week. Believe me, it's a sh*tload of work ;-)

davidpede:
@gpsietzema @Jako I completely screwed this branch so have re-opened here #13893

Issue #9685, opened 2013-03-21
Ability to filter / sort TVs that have been assigned to a template


studiorepublic:
chris@studiorepublic.com created Redmine issue ID 9685

When working on a site with lots of TVs it's hard to quickly tell which TVs have been assigned. It would be great to have a filter to just show assigned/unassigned TVs, or alternatively be able to sort on the access column

whitebyte:
Right now this can be easily archived with modDevTools extra, so should this really be a part of the core?

Jako:
Sorting by 'assigned' column possible with the PR of @davidpede.

PR #13598, opened 2017-09-05
Sort TVs that have been assigned to a template

davidpede:
### What does it do?
Enables sorting on the access column in tpl > tv's grid.

### Why is it needed?
Allows you to easily show which tv's are assigned to a template. Very useful on large sites with 100s of tv's.

### Related issue(s)/PR(s)
Addresses: #9685


OptimusCrime:
Is the new ordering saved, or is it returned back to the initial ordering when you refresh the site?

Jako:
Cherry picked into 2.x

Issue #13457, opened 2017-05-24
[Bug] manager->lexicon not show new language


FerX:
### Summary
Language management does not display all translation languages.

### Step to reproduce
I installed the "SimpleSearch" extra package.
In manager-> lexicon select namespace "sisea"
Added new translation for a language that is not present. (Italian in my example)
Translation works but does not show the new language in the list of languages.

### Observed behavior
The getLanguageList method in core/model/modx/modlexicon.class.php create list of languages supported by that namespace read only on disk and not in the database.

### Environment
MODX 2.5.7, apache, mysql.

gernotebenlechner:
Checking if this issue is valid at #modxbughunt with a fresh 2.5.8-dev install. 

New language for the SimpleSearch extra in core/components/simplesearch/lexicon/ shows up immediately in Manager --> Settings --> Lexicons --> namespace "sisea" within a MAMP instance.

BTW: for a remote server (2.5.7, Apache, MySQL) it shows up immediately after upload in Manager --> Settings --> Lexicons --> namespace "sisea", too.

Result: I **can NOT confirm this issue**.

gpsietzema:
@modxbot close

#modxbughunt #1point to @gernotebenlechner

FerX:
I apologize but I did not explain it well.
**The new language is added from the panel, and then inserted into DB.**
The **getLanguageList** method in core/model/modx/modlexicon.class.php create list of languages supported by that namespace **read only on disk** and not in the database.
If I then insert a new language from the panel (and not on disk) does not appear.

Mark-H:
Thanks for the additional details!

wuuti:
If this was already integrated into 2.6.0, this issue can be closed, right? @OptimusCrime

OptimusCrime:
I think so, yes. 

@modxbot close

PR #13599, opened 2017-09-05
Fetch Lexicon languages and topics from DB as well as file system

OptimusCrime:
### What does it do?
Implement two additional queries. These attempts to fetch topics and languages for lexicon that are not present in the file system. This means that the lexicon entries are added with the manager interface.

### Why is it needed?
Solves problems that people have when they add lexicon entries for extras (or the core) with namespaces and/or languages that are not present in the lexicon files.

### Test case

Download Simplesearch and install the extra. Go to the manager interface. Add a new Lexicon entry:

- Name: foo
- Namespace: sisea
- Topic: context
- Language: zh
- Value: lorem

Update the page. If you select namespace sisea in the dropdown at the top of the grid, then the context topic will not be present. Similar, if click the dropdown for language, then zh will not be present. This PR fixes that.

Note that the system correctly fetches the actual entries for topics and languages that are only stored in the database, it is only the dropdowns that are not populated correctly.

### Related issue(s)/PR(s)
#13457

### DIFF
--- a/core/model/modx/modlexicon.class.php
+++ b/core/model/modx/modlexicon.class.php
@@ -336,6 +336,23 @@ class modLexicon {
                 }
             }
         }
+
+        $c = $this->modx->newQuery('modLexiconEntry');
+        $c->where(array(
+            'namespace' => $namespace,
+            'topic:NOT IN' => $topics,
+        ));
+        $c->select(array('topic'));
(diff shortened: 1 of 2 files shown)


opengeek:
Cherry-picked into 2.x via 321179b78.

Issue #13411, opened 2017-04-06
Generic error when renaming a User Group to something preexisting


pixelchutes:
### Summary
When attempting to rename an existing User Group to something that already exists, a generic error occurs and the rename does not take place.

### Step to reproduce
1. Go to Access Control Lists
2. Right-click User Group you'd like to rename and select "Update User Group"
3. Enter the new name **using a User Group already defined** (on "General Information" tab), then click <kbd>Save</kbd>
4. See error prompt (below)

### Observed behavior
The following error is received:
> An error occurred while trying to save the user group.

![error](https://cloud.githubusercontent.com/assets/352182/24768216/0899709e-1abf-11e7-8107-54d42db52ec9.jpg)

### Expected behavior
A more helpful error would be one indicating the duplicate issue. Otherwise, it's not immediately clear what the actual error is, and appears to be some unknown, technical problem.

### Environment
MODX 2.5.6, macOS 10.12, Chrome 57

OptimusCrime:
Sent a PR for this issue.

Jako:
PR was merged.

PR #13600, opened 2017-09-05
Return better error message if group name already exists

OptimusCrime:
### What does it do?
Returns a better error message if user attempts to rename an existing user group to a name that already exists in the system.

### Why is it needed?
To make it less confusing for the users.

### Related issue(s)/PR(s)
#13411


Jako:
Cherry picked into 2.x

PR #13601, opened 2017-09-05
Make sure we dont store the appended context key in the context name

OptimusCrime:
### What does it do?
You can edit the name of contexts in the context grid (manager -> wheel -> contexts). Here you can double click to edit the context name. The context name is rendered on the format: `context name (context key)`. If you change the name and save, it will be saved as `context name (context key) (context key)` because the value of `(context key)` is concatenated to the original context name.

This PR solves this problem by stripping away `(context key)` at the end of the context name string. This is an alternative to @achterbahn solution, which removed the endfix with the context key. His code is available [here](https://github.com/achterbahn/revolution/commit/9e7f57a416b384cbd679914734bc3524c171dfe5#comments), but as far as I can tell was never PR'ed to this repository.

### Why is it needed?
Fixes the issue described above.

### Related issue(s)/PR(s)
#13371

### DIFF
--- a/core/model/modx/processors/context/updatefromgrid.class.php
+++ b/core/model/modx/processors/context/updatefromgrid.class.php
@@ -8,7 +8,7 @@
 class modContextUpdateFromGridProcessor extends modProcessor {
     /** @var modContext $context */
     public $context;
-    
+
     public function checkPermissions() {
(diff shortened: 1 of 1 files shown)


opengeek:
I believe #13372 from @achterbahn is the preferable solution.

PR #13344, opened 2017-03-03
Fix positioning of TVs on the first tab #13318 #modxbughunt

pepebe:
Fixes positioning of tvs on the first tab and moves the resetTV link to the right side of the tv label.

### What does it do?
Remove padding left on tv formfields and move the resettv link to the right side of the tv label.

The details are described here: https://github.com/modxcms/revolution/issues/13318

### Why is it needed?
TVs moved to the first tab have a padding left thats not in line with other form fields. The padding was necessary so the reset TV link could be positioned to the left. Moving the link to the right side of the label and setting padding-left of the formfield to 0 solves the problem.  

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13318

Issue #13323, opened 2017-03-03
#modxbughunt - Wrong width of textarea and text TVs 


pepebe:
### Summary
If you are moving textfields and textareas to 

modx-resource-main-left
or
modx-resource-main-right

they are slightly to large on the right side, breaking the bounderies of their surrounding columns.

![screenshot 2017-03-03 12 08 30](https://cloud.githubusercontent.com/assets/667315/23548945/263607fe-000a-11e7-872b-c63ebe0edcef.png)

Width of these elements is set by css (99%). On the other hand textfields and textareas as for example pagetitle and description have a custom width set by js.

Setting box-sizing to border-box as well as adding some dedicated styles for line-height, etc would solve this problem.

![screenshot 2017-03-03 12 02 08](https://cloud.githubusercontent.com/assets/667315/23548741/451ee704-0009-11e7-95e4-ad76514be292.png)

A problem might be support for older browsers:

![screenshot 2017-03-03 11 40 11](https://cloud.githubusercontent.com/assets/667315/23548119/5479904e-0006-11e7-9c10-28ec77f76d42.png)
 

### Step to reproduce
How to reproduce the issue, including custom code if needed.

### Observed behavior
How it behaved after following steps above.

### Expected behavior
How it should behave after following steps above.

### Environment
MODX version, apache/nginx and version, mysql version, browser, etc. Any relevant information.

pepebe:
````
    div#modx-resource-main-columns .modx-tv input[type="text"],
    div#modx-resource-main-columns .modx-tv textarea {
        box-sizing: border-box;
        width: 100%!important;
        min-width: 200px;
    }
    div#modx-resource-main-columns .modx-tv input[type="text"] {
        line-height: 20px;
        height: auto;
    }    
       
    div#modx-resource-main-columns .modx-tv input[type="text"] {
        line-height: 20px;
        height: auto;
    }       
    div#modx-resource-main-columns .modx-tv .x-superboxselect-input-field {
        min-width: 200px;
    }   
````

rthrash:
at ~98% browser compatibility, I think `box-sizing` is safe enough

OptimusCrime:
@pepebe You submitted a PR to your own repository. You need to create a PR to this (modxcms/revolution).

Ruslan-Aleev:
Related with - https://github.com/modxcms/revolution/issues/11856

PR #13603, opened 2017-09-05
Fix wrong width of textarea and text TVs when moved

OptimusCrime:
### What does it do?
If you are moving textfields and textareas to `modx-resource-main-left` or `modx-resource-main-right`, they are slightly too wide on the right side, breaking the bounderies of their surrounding columns.

Please note that this is untested! I copied the code from @pepebe 's original PR.

### Test case
Please see original issue #13323

### Why is it needed?
Fixes issue with styling.

### Related issue(s)/PR(s)
Issue ##13323
PR #13344
Local PR [pepebe/revolution/#2](https://github.com/pepebe/revolution/pull/2)

### DIFF
--- a/_build/templates/default/sass/_forms.scss
+++ b/_build/templates/default/sass/_forms.scss
@@ -148,10 +148,6 @@ input::-moz-focus-inner {
       line-height: 20px;
       height: auto;
     }
-    .modx-tv input[type="text"] {
-      line-height: 20px;
-      height: auto;
-    }
     .modx-tv .x-superboxselect-input-field {
       min-width: 200px;
     }
--- a/_build/templates/default/sass/_forms.scss
+++ b/_build/templates/default/sass/_forms.scss
@@ -138,6 +138,25 @@ input::-moz-focus-inner {
     padding: 0 0 0 0px !important;
   }
 
+  #modx-resource-main-columns {
+    .modx-tv textarea {
+      box-sizing: border-box;
+      width: 100%!important;
+      min-width: 200px;
+    }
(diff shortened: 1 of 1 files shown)


Jako:
This PR should be enhanced, since the width and the box-sizing rule is applied only to textarea template variables. Other rules seems to be missing (input etc.).

OptimusCrime:
@Jako Care to elaborate how you tested this? I can take it for a spin and check if some of the other classes needs to be adjusted too.

Jako:
I just have read the scss code. There is no width adjustment for a text TV. Only a textarea TV has a `width: 100%!important;`, which should do the job.

OptimusCrime:
I'll try to test this properly. I am still a bit unsure how this works.

Mark-H:
This habit of yours to submit PRs that other people made, without knowing how to test it, is interesting @OptimusCrime ;)

Mark-H:
This does not fix the problem reported in #13323. The built CSS resolves to this:

```
  .x-form-item #modx-resource-main-columns .modx-tv textarea {
    box-sizing: border-box;
    width: 100% !important;
    min-width: 200px; }
  .x-form-item #modx-resource-main-columns .modx-tv input[type="text"] {
    line-height: 20px;
    height: auto; }
  .x-form-item #modx-resource-main-columns .modx-tv .x-superboxselect-input-field {
    min-width: 200px; }```

while the `#modx-resource-main-columns` should be at the front of the definition. 

Would you like to correct this @OptimusCrime or should the PR be closed?

Issue #12684, opened 2015-09-29
Add sanity check in Widget Controller


BobRay:
Above line 61 of the manager\controllers\default\system\dashboards\update.class.php file, there's no sanity check on the $placement->getOne('Widget') call. If the placement exists and the widget doesn't (how this can happen is another issue), line 61 throws a PHP fatal error. 

```
        $placement->getOne('Widget');
        if ($placement->Widget->get('lexicon') != 'core:dashboards') {
            $this->modx->lexicon->load($placement->Widget->get('lexicon'));
        }
```

I'm not sure what should happen if the widget doesn't exist -- maybe this:

```
  if (! $placement->getOne('Widget')) {
       $placement->remove();
       continue;
   }
```

Ibochkarev:
@Alroniks @Jako fix was merged #13604 .

Please close.

PR #13604, opened 2017-09-05
Make sure Widget exists before calling methods on it

OptimusCrime:
### What does it do?
Add a "sanity" check that the widget from a placement is not null before running methods on it.

### Why is it needed?
Avoid fatal error.

### Related issue(s)/PR(s)
#12684

### DIFF
--- a/manager/controllers/default/system/dashboards/update.class.php
+++ b/manager/controllers/default/system/dashboards/update.class.php
@@ -58,6 +58,11 @@ class SystemDashboardsUpdateManagerController extends modManagerController {
         /** @var modDashboardWidgetPlacement $placement */
         foreach ($placements as $placement) {
             $placement->getOne('Widget');
+
+            if (!($placement->Widget instanceof modDashboardWidget)) {
+                continue;
+            }
+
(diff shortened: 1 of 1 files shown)


manager/controllers/default/system/dashboards/update.class.php
```diff
@@ -58,6 +58,11 @@ public function getWidgets() {
         /** @var modDashboardWidgetPlacement $placement */
         foreach ($placements as $placement) {
             $placement->getOne('Widget');
+
+            if ($placement->Widget == null) {
```
Alroniks: `if (is_null($placement))` looks shorter and more clear and safe, or need to use `===`

opengeek: I believe this should make sure it is an instance of modDashboardWidget instead of checking if not null.

Jako:
@OptimusCrime Do you want to add the change requested by @opengeek?

OptimusCrime:
@Jako Done! Please note that this functionality has not been tested, as I was unsuccessful in trying to recreate a failing scenario here.

Issue #13612, opened 2017-09-19
2.5.8 - title of buttongroup-xtype unvisible, because of line-height:0


Bruno17:
### Summary
the line-height of 
```
.x-toolbar td, .x-toolbar span, .x-toolbar input, .x-toolbar div, .x-toolbar select, .x-toolbar label
```
has changed from 1 to 0, which results in unvisible titles in buttongroups.

### Step to reproduce
See filter- or actions- buttongroups for example at above MIGX-grids.

### Observed behavior
titles of buttongroups are unvisible

### Expected behavior
buttongroup - titles should be visible.

### Environment
MODX 2.5.8

OptimusCrime:
Can you post a image illustrating this? I have never used MIGX and I am not sure how those button groups looks. Does this result in problems in the MODX core or only extras?

Bruno17:
![image](https://user-images.githubusercontent.com/148223/30596032-453f47c8-9d53-11e7-9990-79ebba9159ce.png)
![image](https://user-images.githubusercontent.com/148223/30596101-7bfac1ca-9d53-11e7-819f-c4b0be85dd59.png)

Bruno17:
Not sure, if they are used anywhere else. This is just xtype:buttongroup and I think the title should be visible by default, if used anywhere.

Alroniks:
It seems that it was merged, so I am closing it. Feel free to reopen.

PR #13543, opened 2017-07-10
Secondary button height fixes

davidpede:
### What does it do?
SASS changes to fix some visual bugs in Google Chrome where some secondary buttons are a different height to their form inputs.

### Why is it needed?
chrome_2017-07-10_10-30-26
chrome_2017-07-10_10-32-19

Tested in: Chrome 59, FF 54 & IE11

PR #13613, opened 2017-09-19
Make buttongroup title visible again

Jako:
### What does it do?
Make buttongroup title visible again

### Why is it needed?
The line height was of the x-toolbar elements were reduced in #13543. That change makes the buttongroup title invisible.

### Related issue(s)/PR(s)
#13612


opengeek:
Cherry-picked into 2.x as f29206979.

Issue #13619, opened 2017-09-27
Create composer.json


pixelive:
### What does it do ?
Turns MODX into [Composer](https://getcomposer.org/) library. No autoloading by namespaces, no core split into separate packages - just making MODX available to be installed via Composer, nothing more and nothing less.

### Why is it needed ?
It gives MODX an opportunity to be installed via the most popular dependency manager for PHP. It wont be on [packagist](https://packagist.org) by default but we can at least require MODX with Composer using [repositories](https://getcomposer.org/doc/05-repositories.md). This won't have any disadvantages. 

### Related issue(s)/PR(s)
Thanks to @PatchRanger issue #12422.

And pull request #13620.

OptimusCrime:
https://github.com/modxcms/revolution/pull/12422#issuecomment-172581023

pixelive:
@OptimusCrime I know... but what is the disadvantage of doing it? It can improve our workflow a lot and it doesn't have any disadvantages for MODX. If you have good arguments please tell me, instead of linking to a comment which has no arguments at all....

OptimusCrime:
Other than the fact that the MODX people would have to manage the Packagist package, nothing I guess. 

However, I am not really sure how it would work when the Revolution code does not adhere to any autoloading what so ever? Can this create problems with composer?

Mark-H:
I'm not sold on the why bit yet. It doesn't allow MODX to be installed as modx has a separate setup process, at most it allows to download the MODX files, which people would then need to manually build (as if installing from git) before they can run setup. 

Is that really an improved workflow?

pixelive:
It is not a requirement to be on Packagist we can still require MODX using other composer methods.

I think that is a matter of configuration. They managed to do it with WordPress, for example: https://github.com/roots/bedrock

pixelive:
@Mark-H why [manual](https://getcomposer.org/doc/articles/scripts.md)?

PatchRanger:
@Mark-H To extend @pixelive comment: even more - from MODX-styled scripts they will turn into declarative composer-scripts, which are much more standard.

> MODX to be installed as modx has a separate setup process

The whole setup process could (and in my view - should) be turned into composer scripts. Ideally `composer install` should be enough to build entire system from scratch - without additional manual handling or custom actions.

Mark-H:
@patchranger I agree, that would be amazing, but that's something for 3.x, which has already started using a composer file and autoloading for xPDO, and not 2.x.

[...]

PR #13620, opened 2017-09-27
Turns MODX into Composer library

johanmolen:
### What does it do ?
Turns MODX into [Composer](https://getcomposer.org/) library. No autoloading by namespaces, no core split into separate packages - just making MODX available to be installed via Composer, nothing more and nothing less.

### Why is it needed ?
It gives MODX an opportunity to be installed via the most popular dependency manager for PHP. It wont be on [packagist](https://packagist.org) by default but we can at least require MODX with Composer using [repositories](https://getcomposer.org/doc/05-repositories.md). This won't have any disadvantages. 

### Related issue(s)/PR(s)
Issue #13619

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,3 @@ _build/templates/default/sass/neat
 _build/templates/default/sass/font-awesome
 *.css.map
 *.js.map
-
-#ignore composer files
-composer.json
-composer.lock
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,27 @@
+{
+  "name": "modx/revolution",
+  "description": "Content Management System and Application Framework.",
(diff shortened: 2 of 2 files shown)


Mark-H:
Johan, you've already contributed often enough to know better than to get rid of the default PR text to replace it with a single link to an issue, with a non descriptive title and commit message as bonus...

pixelive:
@Mark-H changed.

bezumkin:
I think we can safely close this, because 3.x is already [uses with composer](https://github.com/modxcms/revolution/blob/3.x/composer.json).

Mark-H:
Closing in favour of #13784

Issue #13046, opened 2016-06-04
[Enhancement] cleanAlias on base of context settings


Jako:
### Summary

The modX cleanAlias method should get the friendly_alias_translit setting on base of the context of the resource.
### Observed behavior

At the moment the translit table could only be set globally for the whole site (according to some tests and after inspecting the code).
### Expected behavior

It should be set on context base.

Flowan:
I'm going to try and fix this today. #modxbughunt

OptimusCrime:
PR merged. Closing.

@modxbot close

PR #13622, opened 2017-10-01
cleanAlias on base of context settings

Jako:
### What does it do?
Add the `friendly_alias_translit` context/system setting as option to the filterPathSegment call.

### Why is it needed?
Before this patch, the cleanAlias method called the filterPathSegment method without the `friendly_alias_translit` context/system setting. That way always the system setting was used. 

This patch will add a possible `friendly_alias_translit` context setting to the filterPathSegment method call. Since requesting a context setting will default to the system setting, if no context setting is available and the `friendly_alias_translit` context setting could be overridden by the options of the cleanAlias method, there occurs no BC with the patch.

### Related issue(s)/PR(s)
#13046

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -612,10 +612,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
      */
     public function cleanAlias($alias, array $options = array()) {
         if ($this->xpdo instanceof modX && $ctx = $this->xpdo->getContext($this->get('context_key'))) {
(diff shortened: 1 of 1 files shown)


OptimusCrime:
Nice work, but the tests are broken.

Jako:
I have noticed that too. Fixed now.

opengeek:
Should this not just merge all context settings and pass those as options so all of the various options could favor context-specific overrides?

Jako:
You are right. I will add that to the PR.

Jako:
Done and tested well here.

PR #13623, opened 2017-10-02
Performance improvements lexicon js [recreated from #12643]

wuuti:
### What does it do ?

This PR attempts to improve the performance and caching logic for the lang.js files created by connector. It also fixes an issue with etags causing never a 304 modified being sent. 

It is quite unlogical to regenerate the dictionary (=lexicon entries collection) for every manager page request according to _some_ topics maybe needed. As a result, nearly every page generates its own dictionary. Why don't we send the _whole_ dictionary for a language and the system at once to the browser? It is send only once, after that the browser does not need to reload it regardless on which page you are. 
In this PR the dictionary is generated by getting all namespaces and all topics inside, then fetching the whole lexicon for all topics. The result is cached (currently in cache partition 'lexicon-js'), so the etag computation can be done efficiently. 
### Why is it needed ?

Speeds up manager experience and fixes bug with etag-header. The requests to connectors/lang.js.php get a lot faster and re-requests get the correct 304 not modified.
### Open issues/enhancements
- purge cache partition on cache refresh and lexicon changes for that language. 
- (update) the cache partition also needs to be refreshed when an extra is installed/updated

Original PR was #12643, recreated due to diverged branches.


rthrash:
This is a great improvement. Huge fan of leveraging browser cache for things like this and simplifying how it works. Well done.

Mark-H:
I'm not sure if it's something wrong with my testing, but this doesn't seem to be making much use of browser caching just yet. Only when loading the same page in the manager twice, it seems to take it from memory cache. 

My guess is that because lexicon topics are still being appended to the lang.js.php file, the browser is not bothering to check if it matches what's in the cache. 

@wuuti Any chance you could review and see if you're experiencing the same issue with this?

wuuti:
checking that....

opengeek (changes requested):
I'd love to actually review this work, but it seems a merge commit has added 459 changed files to this PR. You need to remove any merges from this and rebase your work on top of the latest 2.x branch if this is to be considered for inclusion.

opengeek:
I got no response here. Closing.

PR #13649, opened 2017-10-20
Hide empty tvs tabs on resource

SintezCode:
### What does it do?
Do not add emty TVs category tab on resource and hide TV-tab if is empty

### Why is it needed?
If you move TV by form custumization or by plugins, some TV categories or tab with TVs may be empty after.
BEFORE:
image 001
AFTER:
image 002


Jako:
Merged into 2.x

Issue #12818, opened 2015-12-08
Missing remove old packages button


Jako:
### Summary

To clean up the space of an installation, a 'Remove Old Packages' button is missing.
### Step to reproduce

In Package Mananagement the details of a package could be shown with a list of all previous installed package versions. The entries in that grid could be removed one by one. There is a button missing for 'Remove Old Packages' that removes all old package versions.
### Expected behavior

The Button should remove all old package entries in the database and all old package files on the installation for that package.

Mark-H:
Ready for 2.7

PR #13654, opened 2017-10-21
Purge old packages button in packages management

Jako:
### What does it do?
Add a 'Purge Old Packages' button to the Package Management manager page.

### Why is it needed?
In Package Mananagement the details of a package could be shown with a list of all previous installed package versions. The entries in that grid could be removed one by one. There is a button missing for 'Purge Old Packages' that removes all old package versions.

The processor could be restricted to one package name by an option. A 'Purge Old Versions' button is available in the package details window.

### Related issue(s)/PR(s)
#12818


rthrash:
Awesome!

sottwell:
Maybe a checkbox and bulk operation on all checked items instead of willy-nilly removing all previous versions?

Jako:
Nice idea, but a lot more work than my addition.

Alroniks:
Purge all old versions may be dangerous and yes, I'd prefer to select items and delete only selected. Nevertheless, it's a good start.

Jako:
A trash is a trash and previous installed versions are previous installed versions. Why should a special previous version be retained? 

Maybe there could be an input to remove all but the last two versions of each package. And of course: there should be a button in each package details window, to purge only the previous versions of this package.

Alroniks:
@Jako It was that I mean. Sometimes you need at least keep the previous version to have an ability to rollback changes. But removing all except current and -1 version is also not so useful. Checkboxes more complex to implement but it adds more freedom and covers more use cases.

Omeryl:
@alroniks What’s preventing the user from downloading the old version again? If I recall, we don’t delete old versions from the official provider automatically.

OptimusCrime:
@Omeryl The API only returns the last version. The REST API would need to be upgraded.

Omeryl:
@optimuscrime yeah, but we’re not doing this on upgrade. If you really need to downgrade, you can pull up the old package version and just download it yourself. The normal workflow I see would be:
1) Upgrade package
2) Test your site
3) Purge old versions

Alroniks:
I agree with that we should have the ability to remove old versions but I am advocating variant and implementation when a user has more ways that one (real freedom) to do that.

rthrash:
I would suggest starting with a simple purge all but current … you're not forced to purge anything and 99% of users won't notice this anyway. If there is demand down the road it can be expanded to save the N-last versions, but it sure would get rid of a lot of cruft in the interim.

Jako:
Now the old versions of one package could be purged with another button.

Mark-H:
Works nicely @Jako! Thanks!

PR #13660, opened 2017-10-30
Exclude resource alias for child

labr1005:
### What does it do?
This includes and extends the PR #13395 by @OptimusCrime.

> ### What does it do?
> 
> Implement functionality to hide a alias in the path of children resources.
> 
> Includes:
> 
> This PR makes sure to add the new field in the map files of both MSSQL and MySQL. I've also added scripts for each of them to be added during an upgrade. This is test and works as expected.
> Adds a new checkbox under Settings in the box at the right side, below Searchable and above Freeze URI.
> Adds a simple lexicon entry with a wording similar to that in Evo. Appears when hoving the checkbox like the other settings in this area.
> It works like this. Say we have the current relationship of resources Foo -> Bar, which has aliases foo and Bar which has its alias to bar. The default value of alias_visible is 1, so it does not change the current behavior of any sites without alterations. If both Foo and Bar has this setting to 1 (as is the default), it would produce the URI as before foo/ and foo/bar. Below are the other configurations and their URI values:
> 
> Foo alias_visible = 0. Bar alias_visible = 1
> Foo URI = foo/. Bar URI = bar.
> 
> Foo alias_visible = 1. Bar alias_visible = 0
> Foo URI = foo/. Bar URI = foo/bar.
> 
> Foo alias_visible = 0. Bar alias_visible = 0
> Foo URI = foo/. Bar URI = bar.
> 
> As we see, the URIs ignore the PARENT alias if alias_visible is 0, but will always use its own alias regardless of the alias_visible value. This is because we always require an alias for any resource.
> 
> Note: This PR is currently missing tests. I'd love to implement some tests to make sure this works like it should.
> 
> ### Why is it needed?
> 
> Could be useful in certain situations. See issue referenced below. Seem like a common request. The feature was present in Evo, but for some reason removed since then. I see no reason this should not be included in the core. The functionality to go around this via an extra/plugin would be rather painful to implement.

### Related issue(s)/PR(s)
#11153
#13395


OptimusCrime:
Nice. This is fine by me. I was waiting to include some unit tests in my PR. I did not find the time to include this. However, I am working on simplifying and improving the test suit, which in turn will make it easier to include proper tests for this. If someone can test the functionality and validate that it works, then I think that should be good enough for now.

Bruno17:
Anyone knows when and where this file
setup/includes/upgrades/mysql/2.6.0-pl.php
 is called in the setup - process?

I've added this upgrades-files and was trying to rerun the setup of 2.6.0-dev
Seems this script doesn't run in this situation.

OptimusCrime:
Most likely because the Git version is not tagged with the correct release version. I did not investigate just how this process is executed, but it is a part of the setup process.

labr1005:
Hm, worked for me without any problems. Took the 2.x branch and ran setup. Switched to this feature branch and reran setup.

> Successfully upgraded table for class modResource
> Added new `alias_visible` column to ``modx_site_content``.

OptimusCrime:
Closed my original PR #13395 in favor of this. Note that my original PR contains some questions and answers concerning the implementation. Check this first you have any questions.

weranda:
Hi, guys.
Maybe you know.
Will be included this changes in the new version MODX (2.6.x)? Are there any chances?
I don't understand what's going on here)

Mark-H:
For future reference (not sure if this was you @labr1005 or @OptimusCrime) - there's no need to change all the lexicon files. CrowdIn will add the changes to those automatically.

Mark-H:
It may have taken close to 4 years since Bert's original request in #11153, but at last this has been incorporated 😬 

@Bruno17 Perhaps you were already on 2.6.0 (or higher) causing the upgrades to not run. It looks at the stored version number (`settings_version` system setting) and the name of the file in the database-driver-specific directory. I've updated the files to 2.7.0 to get them to run in my testing as well.

PR #13662, opened 2017-11-01
Add context filter in tree resource

tillpaid:
### What does it do?
Adds context filter in the resource tree

### Why is it needed?
When I have a lot of contexts on my site, I spend a lot of time searching for the right context. It is not confortable

### Related issue(s)/PR(s)
No

### DIFF
--- a/core/model/modx/processors/resource/getnodes.class.php
+++ b/core/model/modx/processors/resource/getnodes.class.php
@@ -130,9 +130,9 @@ class modResourceGetNodesProcessor extends modProcessor {
         $this->itemClass= 'modContext';
         $c= $this->modx->newQuery($this->itemClass, array('key:!=' => 'mgr'));
         if (!empty($this->getProperty('contextFilter'))) {
-            $c->where([
+            $c->where(array(
                 'name:LIKE' => "%{$this->getProperty('contextFilter', '')}%",
-            ]);
+            ));
         }
         if (!empty($this->defaultRootId)) {
             $c->where(array(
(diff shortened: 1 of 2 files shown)


core/model/modx/processors/resource/getnodes.class.php
```diff
@@ -128,7 +128,7 @@ public function getRootNode() {
      */
     public function getContextQuery() {
         $this->itemClass= 'modContext';
-        $c= $this->modx->newQuery($this->itemClass, array('key:!=' => 'mgr'));
+        $c= $this->modx->newQuery($this->itemClass, array('key:!=' => 'mgr', 'name:LIKE' => "%{$this->getProperty('contextFilter', '')}%"));
```
lawrenz1337: I think it would be better to separate this from the base query. Having ```LIKE %%``` on every query in no property sent seems not optimal.

lawrenz1337 (changes requested):

lawrenz1337 (commented):
The code seems fine but the looks of the input need some work and love

zaigham:
Could this be done as an Extra?

tillpaid:
this feature changes the system files, I do not think that the extra is a good idea

Jako:
Ext.override the MODx.tree.Resource function to extend the functionality before it is displayed. That should be an option to build that thing as an extra. The script needed for this could be added with a plugin during OnManagerPageInit.

It makes no sense to have the filter input always displayed and it seems to be a quite special need, that you are covering.

Jako:
@tillpaid I think you should do this in an installable extra.

Issue #13664, opened 2017-11-03
bug on several of our sites after upgrade 


antfrolov:
1. https://yadi.sk/i/1Iwn8OGJ3PLxov 
2. without quotes in the name - ок https://yadi.sk/i/iedSYa1j3PLy3d 
3. with quotation marks in the name - bug https://yadi.sk/i/PhXQKRob3PLyHF

on several of our sites

OptimusCrime:
I can not get the images to show here. Could you upload them directly to GitHub instead of using external sources?

antfrolov:
![2017-11-02_18-55-22](https://user-images.githubusercontent.com/33340307/32364222-790fbdb6-c095-11e7-8d16-5f334e50481b.png)
![2017-11-02_18-57-07](https://user-images.githubusercontent.com/33340307/32364225-7bdffe3e-c095-11e7-867f-c2dfffd65136.png)
![2017-11-02_18-57-54](https://user-images.githubusercontent.com/33340307/32364227-7e095eda-c095-11e7-90fd-86d964cf7e1b.png)

Mark-H:
An empty Manager page like that suggests a JavaScript error. Could you
check for that in the browser developer tools?

Op 3 nov. 2017 8:50 a.m. schreef "antfrolov" <notifications@github.com>:

antfrolov:
Uncaught SyntaxError: Unexpected identifier
?a=resource/update&id=1241:387 Uncaught TypeError: Cannot read property 'getForm' of undefined
    at ?a=resource/update&id=1241:387
    at ext-all.js:21
    at b (ext-all.js:21)
?a=resource/update&id=1241:427 Uncaught TypeError: Cannot read property 'getForm' of undefined
    at ?a=resource/update&id=1241:427
    at ext-all.js:21
    at b (ext-all.js:21)
?a=resource/update&id=1241:467 Uncaught TypeError: Cannot read property 'getForm' of undefined
    at ?a=resource/update&id=1241:467
    at ext-all.js:21
    at b (ext-all.js:21)
?a=resource/update&id=1241:528 Uncaught TypeError: Cannot read property 'getForm' of undefined
    at ?a=resource/update&id=1241:528
    at ext-all.js:21
    at b (ext-all.js:21)
?a=resource/update&id=1241:587 Uncaught TypeError: Cannot read property 'getForm' of undefined
    at ?a=resource/update&id=1241:587
    at ext-all.js:21
    at b (ext-all.js:21)
[...]

PR #13665, opened 2017-11-03
htmlspecialchars pagetitle for delete resource dialog UI

labr1005:
### What does it do?
htmlspecialchars ~~JSON encodes~~ the pagetitle for delete resource dialog UI.

### Why is it needed?
2.6 introduced an advanced delete resource dialog that shows the pagetitle of the resource to be deleted. When the pagetitle contains quotes, the update resource manager page would not load due to a js syntax error.

### Related issue(s)/PR(s)
#13664 
#13497

### DIFF
--- a/manager/controllers/default/resource/update.class.php
+++ b/manager/controllers/default/resource/update.class.php
@@ -42,7 +42,7 @@ class ResourceUpdateManagerController extends ResourceManagerController {
             MODx.load({
                 xtype: "modx-page-resource-update"
                 ,resource: "'.$this->resource->get('id').'"
-                ,pagetitle: "'.$this->modx->toJSON($this->resource->get('pagetitle')).'"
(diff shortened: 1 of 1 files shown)


Jako:
JSON encode is somehow wrong. htmlspecialchars should be enough.

labr1005:
I think #13497 could also be partly reverted/refactored, as all the needed information should already be available in https://github.com/modxcms/revolution/blob/d0934af616bea39abb0ecc75abaf87fe706c2e46/manager/controllers/default/resource/update.class.php#L46
But I don't have the time for this today and this seems to be really urgent.

Jako:
Could you check my comments on https://github.com/modxcms/revolution/pull/13497/files ?

Jako:
Closing in favour of #13669

Issue #13670, opened 2017-11-03
Media browser wrong image size


Pfeifakopf:
### Summary
The media browser shows the wrong image size. Because of that, image+ think all images are to small.

### Step to reproduce
Just uploaded the images via media browser upload. 

### Observed behavior
You can see the small image-sizes. My image was shown with an width of 800px instead of the real 2600px. I downloaded the image from the server. This image has still the right size.

### Environment
modx 2.6.0
PHP 7.0.25 fastcgi
Ubuntu 14.04 
![image](https://user-images.githubusercontent.com/3351449/32391714-5ff7f24c-c0d3-11e7-8781-e4fb097ab0a3.png)

labr1005:
Confirmed.
Introduced in #13629.
I'm working on a fix.

labr1005:
@Pfeifakopf Please check the PR #13672 - it should fix the issue.

labr1005:
@Pfeifakopf Did you have some time to check the recommended PR and did it fix your problems?

Pfeifakopf:
@labr1005 Changes are done, works all fine now.

Jako:
Merged into 2.6.x

PR #13629, opened 2017-10-07
Extends fix for displaying SVG in media browser

labr1005:
### What does it do?
This extends the PR made by @marcjenkins with these features:
- SVG previews for the Files Tree
- SVG support for IE 9+ and Edge (by adding width and height attributes to the img tags for all previews and some CSS)
- image previews are now created proportionally to the original image respecting the options "filemanager_image/thumb_width" and "filemanager_image/thumb_height" as max width and max height
- checked background for transparent images in the MODX Browser
- previews of S3 images are now processed with phpthumb
- file size info for S3

Tested in IE 9+, latest Chrome, Firefox and Safari.

### Related issue(s)/PR(s)
#12459
#13517

### Before
bildschirmfoto 2017-07-12 um 22 26 18

### After
bildschirmfoto 2017-07-12 um 22 25 11

### Tree
bildschirmfoto 2017-07-17 um 20 39 14

PR #13672, opened 2017-11-04
Revert behavior of image_width and image_height for media source images

labr1005:
### What does it do?
Prior to 2.6, the returned values of "image_width" and "image_height" were the size of the original image. 2.6 returns the size of the with phpthumb processed (and therefor often downscaled) image.

This reverts the behavior of "image_width" and "image_height" to the original state.

### Why is it needed?
This PR does not affect MODX directly but fixes problems with extras that rely on and process these values - like ImagePlus.

### Related issue(s)/PR(s)
Introduced in #13629
Reported in #13670


core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -1134,11 +1134,9 @@ public function getObjectsInContainer($path) {
                     'id' => $bases['urlAbsoluteWithPath'].$fileName,
                     'name' => $fileName,
                     'cls' => 'icon-'.$fileExtension,
-                    'file_width' => $size[0],
-                    'file_height' => $size[1],
```
Mark-H: Could you explain why you're getting rid of these array items?

labr1005: I introduced them with my PR #13629. So they occurred first in 2.6.

My idea was to have separate variables for a) the width and height of the preview image in the MODX media browser (image_width and image_height) and b) for the display of the original image's size (file_width and file_height) displayed below the preview image.

But in my aforementioned PR image_width and image_height is the size of the with phpthumb processed (and therefor often downscaled) image.

My error was that I didn't take in account that other extras like ImagePlus rely on and process image_width and image_height.

So I basically only reverted image_width and image_height to the original state and therefor file_width and file_height are obsolet.

labr1005: Or in short: I changed a variables inital purpose and this reverts it.

labr1005: @Mark-H Did this answer your question?

If no, anything else that I can clear up?

If yes, may I suggest considering this for 2.6.1 or the next patch? It's confirmed here #13670 and since I'm also responsible for the PR #13629 I can secure that there will be no problems with the new Media Browser features introduced with 2.6.

PR #13677, opened 2017-11-06
Improve changelog display

Jako:
### What does it do?
~~Replace newlines with br tags in the displayed changelog.~~ Display the changelog with the style `white-space: pre`. Better than adding br tags.

### Why is it needed?
Sometimes the changelog will contain only text and the output in the modx-window expects html. The css change should improve the readability of the changelog output.

### Related issue(s)/PR(s)
#13455


Jako:
It seems to be provider specific. My package provider uses the changelog.txt in the transport file, which is a text file. The MODX package provider uses the html code of the MODX extras frontend.

PR #13684, opened 2017-11-09
Allow callback if nothing is selected

matdave:
Would not fire browser callback if cancel hit or nothing was selected. Fixes TinyMCE RTE Issue 10 https://github.com/modxcms/tinymce-rte/issues/10

### What does it do?
Adds callback support to the Cancel button and the OK button when nothing is selected.

### Why is it needed?
If browser is iframed, there is no way to close it.

### Related issue(s)/PR(s)
https://github.com/modxcms/tinymce-rte/issues/10

### DIFF
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -1485,8 +1485,8 @@ Ext.extend(MODx.browser.RTE,Ext.Viewport,{
         var callback = this.config.onSelect || this.onSelectHandler;
         var lookup = this.view.lookup;
         var scope = this.config.scope;
-        if(selNode && callback) {
-            data = lookup[selNode.id];
+        if(callback) {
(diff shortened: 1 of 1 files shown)


OptimusCrime:
I think I fixed the same error in the old TinyMCE package, here: https://github.com/modxcms/TinyMCE/pull/15

matdave:
@OptimusCrime I just tested, and the old TinyMCE has the same error. Close button does nothing. The only difference is the Old TinyMCE has an X to override closing it.

OptimusCrime:
Odd. I also made this PR https://github.com/modxcms/revolution/pull/12238 to this repository. I'll take a look at it tomorrow. Perhaps they are related or something got reverted somehow.

Anyways, great work!

matdave:
Yeah, that change was in the old browser.  It must have got missed when they updated it.

PR #13699, opened 2017-11-17
Add new output modifier "filterPathSegment"

labr1005:
### What does it do?
This outputs the "slug" or "friendly URL" version of the input. It uses the same function that is used by MODX to build friendly URLs.

```
[[+foo:filterPathSegment]]
```
"Test 1 ? oA u" -> "test-1-oa-u"

As it uses the core functionality, it also takes the translit system settings in account.

E.g. when the extra "translit" is installed with the setting "friendly_alias_translit" set to "german":

"Test 1 ? öÄ ß" -> "test-1-oeae-ss"

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -657,6 +657,9 @@ class modOutputFilter {
                             }
                             $output = implode($delimiter, $return_values);
                             break;
+                        case 'filterPathSegment':
+                            $output = $this->modx->filterPathSegment($output);
+                            break;
 
                         /* Default, custom modifier (run snippet with modifier name) */
                         default:
@@ -698,7 +701,7 @@ class modOutputFilter {
         try {
             $m_con = ($conditional !== '') ? @eval("return (" . $conditional . ");") : false;
             $m_con = intval($m_con);
-            
+
             // If negate is true, we want the value of $m_con to be false
             if (!$negate) {
                 if ($m_con) {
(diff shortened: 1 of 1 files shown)


labr1005:
I'm not sure about the name "filterPathSegment".

Alternatives could be "alias" or "furl", but with those one might think that a real URL is returned?

Maybe "translit"?

Mark-H:
`slug` maybe?

There's also the question of wether or not we want to add many more output filters or not. I personally think there's value in this one as it's a core function and not a standard php function or whatever that's being used.

labr1005:
I like "slug".

And also agree to the rest.

OptimusCrime:
`slug` is very generic, and this output filter could cause problems if anyone already have a snippet called `slug` in their system. Worth a heads up.

labr1005:
Ah. I really thought the order would be the other way round and that a custom snippet would be used favorably over an output modifier when they have the same name.

:thinking:

OptimusCrime:
@labr1005 I am not 100% sure about this though. I just thought that was how it worked.

Jako:
The filter should respect the translit settings of the system/context (if it does not already).

labr1005:
@OptimusCrime You're right. It's a switch statement and the default is to use a custom snippet (if found). So a modifier with that name will be prefered.

https://github.com/modxcms/revolution/blob/2.x/core/model/modx/filters/modoutputfilter.class.php#L661

labr1005:
@jako It does.

Jako:
Then I am fine with that enhancement.

Mark-H:
Thanks @labr1005 - merged into 2.7. Also added to the input/output filter docs.

PR #13700, opened 2017-11-17
Adding system setting "automatic_template_assignment" to fix #5980

### LINKED ISSUE #5980: default_template setting not used for New (child) Resource

matdave:
### What does it do?
I've reworked my previous pull #12705, by sanitizing the default template function, and added a system setting to disable the behavior of inheriting parent template. 

### Why is it needed?
Sometimes you need a special template for a parent page, but want all the siblings to have a standard template.  Without being able to disable inheriting the parent template, every time you create a child it will pick the parent's template instead of the default template. 

### Related issue(s)/PR(s)
Reworking pull request #12705 and closes issue #5980, which was accidentally closed as the pull request was never merged. 

### UPDATE:
I've reworked this to work like Evo and changed it to "automatic_template_assignment" with three options: system, parent, sibling

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -114,6 +114,15 @@ $settings['automatic_alias']->fromArray(array (
   'area' => 'furls',
   'editedon' => null,
 ), '', true, true);
+$settings['automatic_template_assignment']= $xpdo->newObject('modSystemSetting');
+$settings['automatic_template_assignment']->fromArray(array (
+    'key' => 'automatic_template_assignment',
+    'value' => 'parent',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'site',
+    'editedon' => null,
+), '', true, true);
 $settings['base_help_url']= $xpdo->newObject('modSystemSetting');
 $settings['base_help_url']->fromArray(array (
   'key' => 'base_help_url',
@@ -808,15 +817,6 @@ $settings['hidemenu_default']->fromArray(array (
   'namespace' => 'core',
   'area' => 'site',
   'editedon' => null,
-), '', true, true);
-$settings['inherit_parent_template']= $xpdo->newObject('modSystemSetting');
-$settings['inherit_parent_template']->fromArray(array (
-    'key' => 'inherit_parent_template',
-    'value' => 1,
-    'xtype' => 'combo-boolean',
-    'namespace' => 'core',
-    'area' => 'site',
-    'editedon' => null,
 ), '', true, true);
 $settings['inline_help']= $xpdo->newObject('modSystemSetting');
 $settings['inline_help']->fromArray(array (
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -809,6 +809,15 @@ $settings['hidemenu_default']->fromArray(array (
   'area' => 'site',
   'editedon' => null,
 ), '', true, true);
+$settings['inherit_parent_template']= $xpdo->newObject('modSystemSetting');
+$settings['inherit_parent_template']->fromArray(array (
+    'key' => 'inherit_parent_template',
+    'value' => 1,
+    'xtype' => 'combo-boolean',
+    'namespace' => 'core',
+    'area' => 'site',
+    'editedon' => null,
+), '', true, true);
 $settings['inline_help']= $xpdo->newObject('modSystemSetting');
 $settings['inline_help']->fromArray(array (
   'key' => 'inline_help',
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -119,6 +119,9 @@ $_lang['setting_allow_multiple_emails_desc'] = 'If enabled, Users may share the 
 $_lang['setting_automatic_alias'] = 'Automatically generate alias';
 $_lang['setting_automatic_alias_desc'] = 'Select \'Yes\' to have the system automatically generate an alias based on the Resource\'s page title when saving.';
 
+$_lang['setting_automatic_template_assignment'] = 'Automatic Template Assignment';
+$_lang['setting_automatic_template_assignment'] = 'Choose how templates are assigned to new Resources on creation. Options include: system (default template from system settings), parent (inherits the parent template), or sibling (inherits the most used sibling template)';
+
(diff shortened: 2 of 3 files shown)


bezumkin:
For example, category could have template "Category" but children must be with "Child" template.

I believe, we need another setting - `type_of_default` or something like this. And combo with options:
- default (from `default_template` setting)
- parent 
- nearby (from last children of parent)
- another variants, that could be added later

matdave:
@bezumkin, that would be nice, but could possibly break current installations based on how it is implemented.  I think it would be even better to have an option to define default child template when creating a parent template. This could be easily achieved by just adding the variable &template=XX to the "create document" action in the tree. 

This pull request is aimed at solving the immediate issue without breaking current installs.

bezumkin:
> This pull request is aimed at solving the immediate issue without breaking current installs.

Okay. I just worried about hundreds of already existing system settings and there will be more.

rainbowtiger:
Why not just use a plugin? I made one that adds a TV so I can select whatever child template I want for whatever parent resource I want.

labr1005:
Why not use the Form Customization?

sottwell:
Or use Collections.

matdave:
Heck, I could just use Evo.

FerX:
I think there are many utilities like this that should be implemented,
es. restrict template selection based on parent, show / hide children, ...

The solution for not having too many extras would be to **add all the options to Collections**, which I consider a good basis for making many customizations.

matdave:
@labr1005  That is assuming you know what all the possible parent IDs will be prior to handing a site over to a client.  What if they add a new section?

matdave:
Just putting it out there, but I think what I am going to do is make it more like Evo, rather than calling it `inherit_parent_template` I'm going to rework it to be `automatic_template_assignment` with the options for 'System', 'Parent', and 'Sibling' like it used to be.
image
 Anyone object to that?

OptimusCrime:
Sounds like a really good idea, but I am confused as to how this would work integrated with form customization?

rainbowtiger:
matdave, That sounds great to me.

matdave:
@OptimusCrime the performant method would be to just grab one randomly. Since no resource is created yet ( and thus no "nearest neighbor" ) I could also set it up to grab either to first or last based on menu index.  A less performant method would be to grab all of them and check which template is used the most. My code right now (prior to committing) just grab a random item. Which would you say is the most ideal?

matdave:
Actually, I could group by template and order by count. that would be pretty fast and pick the highest one. that would be really fast and probably the best result.

matdave:
Commit 54adf44 adds the new logic that makes it function like Evo, allowing "sibling" as an optional source to inherit templates from.

Mark-H:
How does this relate to #13330?

matdave:
@Mark-H  It would not affect that process at all. If the script properties define the template already, it goes with that. This is for how to handle template selection where there isn't a defined template.

Mark-H:
Thanks for checking/clarifying @matdave :)

OptimusCrime:
@matdave 
>Actually, I could group by template and order by count. and pick the highest one. that would be really fast and probably the best result.

Behavior like this needs to be predictable. If this option only makes sense if you check the template for each child and count them, it will cause more confusion than good. I suggest using the closest sibling. That way we always know what template will be used without too much detective work needed.

matdave:
@OptimusCrime how do you pick the "closest sibling" of a resource before it exists?

OptimusCrime:
@matdave new resources are (as far as I know) added to the end of the list of existing children. That means that the "closest" sibling in this scenario would be the one with the highest menu index.

And if there are no siblings, it can not chose one either way.

matdave:
@OptimusCrime Personally I see that as an annoyance.  I think using the majority template of a specific parent would lead to better results. Say you have a section and you put a "contact page" template at the bottom, but everything else uses an inside page template.  Soley going off the last item would make it so that every new child would default to the "contact page" template, even though the majority of pages in that section use an inside page template.

sottwell:
Easy, put the contact form in a chunk and use a conditional like "[[[[*id:is=42:then=`$formChunk`]]]]

sottwell:
Personally, I always missed the options for defining the default template, system default, parent, or sibling.

matdave:
The contact page template was just an example, it could be a dynamic staff listing or anything that we want the client to have the ability to add using a special defined template without requiring additional programming.  Why should the 1 bottom variant derail the default selection when there is an established (and likely to be used) majority for the rest of the pages? Just because someone adds a page doesn't mean it is going to stay in the bottom of the list after they save it, so why assume it is?

OptimusCrime:
@matdave My five cents is that these options needs to be intuitive and what they do needs to be "logical". The user should know beforehand what their resource is going to get as their template. If you have to count the siblings and find the most common template, there is still a manual and confusing process involved i the process.

labr1005:
The most logical, IMO, would be the possibility to use a(ny) parent field as constraint in the manager customization.

Mark-H:
There are so many different ideas on how this should work (and lately, various different PRs too), that I'm starting to think only uncontested ideas should go into the core, and for everything else, use an extra. Different extras could be made that do different things (majority of sibling, last sibling for example) which allow everyone to use the system that matches their project. 

We also need to figure out the priority of all the different ways we have (/will have) now to set templates. Probably worth a completely new page in the docs that goes over the different options and when they are useful.

rthrash:
Uncontested ideas in Open Source Software … is that possible? :P 

Currently, new children inherit the template of the Parent Container.  Upgrades would maintain this behavior. I really can't think of many sites without different templates under a parent—landing pages with articles or products under them, e.g.

For new installs, I'd suggest defaulting to the sibling with the highest Resource ID under a parent. If no siblings exist, yet, fall back to the parent Template.

matdave:
commit dccdf1f implements @rthrash. The only thing missing is a different system setting for New vs. Upgrade. It just defaults to 'parent' in either case. Not sure how to distinguish the difference.

PR #13704, opened 2017-11-22
Fix files tree and media browser phpthumb bug

labr1005:
### What does it do?
The main fix is to handle "Division by zero" errors when getimagesize() fails.

This PR fixes issues with the files tree and media browser **related to image files** where
- the media browser's grid would only show the message "No files match the specified filter." although there are files in the selected directory.
- hovering an image file in the files tree would show an empty preview.
- choosing a different directory in the media browser would show the results of a previous chosen directory in the grid.
- ...

Additionally it logs an error in the MODX error log when this problem occurs:
```
[2017-11-22 20:02:45] (ERROR @ /core/model/modx/sources/modfilemediasource.class.php : 1132) Thumbnail could not be created for file: /absolute/path/to/assets/uploads/images/broken.jpg
```
### Step to reproduce
- When a directory includes an image that's not an image (rename test.txt to text.jpg).
- When a directory includes a broken/corrupt image (server crash, incomplete file transfer).
- When the media source's setting "imageExtensions" includes an extension that can't be thumbnailed.

### Why is it needed?
- Show the default "No preview available" in the grid for "images" that can't be thumbnailed.
- No preview in the files tree for "images" that can't be thumbnailed.

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -222,6 +222,7 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
 
                         $modAuth = $this->xpdo->user->getUserToken($this->xpdo->context->get('key'));
 
+                        $preview = 1;
                         $imageWidth = $this->ctx->getOption('filemanager_image_width', 400);
                         $imageHeight = $this->ctx->getOption('filemanager_image_height', 300);
                         $thumbnailType = $this->getOption('thumbnailType', $properties, 'png');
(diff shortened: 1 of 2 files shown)


core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -222,6 +222,7 @@ public function getContainerList($path) {
 
                         $modAuth = $this->xpdo->user->getUserToken($this->xpdo->context->get('key'));
 
+                        $preview = 1;
```
Mark-H: Any particular reason `$preview` here is an integer and not a boolean? It seems to be used as a boolean further along the code..

devlanda:
I have tested the solution for the file modfilemediasource.class.php.

This solved the problem that an image could not be selected and that the thumbnails were not generated (caused by corrupted files, in a particular case they were empty). It also displayed which files could not be processed by phpthumb.

Nice contribution!

core/model/modx/sources/modfilemediasource.class.php
labr1005: No. Changed.

PR #13706, opened 2017-11-25
[3.x] Load PhpThumb via Composer instead of including the source files

OptimusCrime:
### What does it do?
Instead of including the source files in the MODX code, we load the dependecy using Composer.

_Technical explanation coming_

I have verified that this code, without any BC worked within the manager (by hoving a file in the file tree and using the file browse). pThumb also works without any alterations.

### Why is it needed?
To couple the code from other dependencies.

### Related issue(s)/PR(s)
[MAB-01](https://github.com/modxcms/mab-recommendations/blob/master/01-replace-3rd-party-libraries-with-composer-dependencies.md)


bezumkin:
Tested and found no errors.

Great job!

PR #13707, opened 2017-11-28
2.6.x

HMoellendorf:
### What does it do?
Solved a bug when trying editing pages with " in pagetitle

### Why is it needed?
Pages with " in title weren't be able to edit.


labr1005:
Already fixed with #13669.

PR #13714, opened 2017-12-06
Resolve unlink() error

cjholowatyj:
### What does it do?
Wrapped the unlink($file) call in an if statement to confirm that $file still exists.

### Why is it needed?
Resolves unlink() errors that occur when concurrent requests for a cache file results in one unlink() call being successful while other unlink() calls result in a PHP error being thrown.

### Related issue(s)/PR(s)
None to my knowledge


opengeek:
This needs to be submitted to the xpdo repository.

Issue #13733, opened 2018-01-03
Inconsistent usage of class keys for similar actions on resources


matdave:
### Summary
When viewing the manager log, it appears that there is some inconsistency when using class keys for certain actions.  Sometimes it falls on the overall modResource class, and other times it uses the specific class, e.g. modDocument or modWeblink.

### Step to reproduce
Publish and Unpublish a resource, then check the manager actions log. 

### Observed behavior
See how those two actions are logged.  Publishing uses the modDocument class key, and unpublishing uses the modResource class key.

### Expected behavior
It would be nice if all actions specific to a resource used the modResource class.

### Environment
MODX 2.6.1

OptimusCrime:
I could be wrong, but I think this is due to inheritance and where the method is called. If publishing is doing within the `modDocument` class, and it calls something like `get_class($this)` it will say `modDocument`. However, if the method for unpublishing is so similar, that it can be in the (shared) parent class `modResource` then the log entry `get_class($this)` will say `modResource`. 

Due to the way inheritance and OOP work, there is no easy way to solve this without yanking out the method from the shared parent class into the individual classes instead.

Have you looked at the code and identified the cause?

matdave:
I have not had time to really dive into this. I guess I could retract my previous statement and say that it would make more sense if everything used the object class that it was defined as, as all actions of the parent class would be available with the addition of any extended processes on the defined class. 

I'll see if I can find some more time over the next couple weeks to document why the parent class gets used the majority of the time, while the extended class only shows up on occasion.

OptimusCrime:
Yeah, my gut is that some of the code is so general it can be used across all inherited classes, while other things are implemented specific.

PR #13734, opened 2018-01-03
Resource Action Log

matdave:
### What does it do?
Adds a resource-specific manager log to the changes tab of the Resource Overview page.

### Why is it needed?
Show more data on the changes tab than just the latest static information. Easier to point blame on a specific resource.

### Related issue(s)/PR(s)
This doesn't close any issues, but it did expose an issue with inconsistent classKey usage on resources #13733.  It works for the majority of pages until that issue is resolved. I could also probably add a class key selector for filtering through data if we feel that is necessary.  Another option would be rewriting the getList class to support multiple class keys.

### DIFF
--- a/core/model/modx/processors/system/log/getlist.class.php
+++ b/core/model/modx/processors/system/log/getlist.class.php
@@ -59,7 +59,7 @@ class modSystemLogGetListProcessor extends modProcessor {
      */
     public function getData() {
         $actionType = $this->getProperty('actionType');
-        $classKey = $this->getProperty('classKey');
+        $classKey = $this->explodeAndClean($this->getProperty('classKey'));
         $item = $this->getProperty('item');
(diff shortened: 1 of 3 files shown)


matdave:
Okay, so I modified it a bit in 2214b32 to have the manager log getList processor accept comma-separated class keys. Now, it sends whatever the assigned class key is along with modResource to generate a complete list. 🎉

Mark-H:
Thanks @matdave - that's a nice addition for 2.7.

Mark-H:
The PR didn't autoclose as I merged this into 2.x, rather than the 2.6.x you targetted in your PR. It was merged in https://github.com/modxcms/revolution/commit/b63c1cd4002956f0e45dc4374a64a161c4332032.

Issue #13753, opened 2018-02-08
Feature: Set childtemplate automatically based on template property


sdrenth:
### Summary
I'd like a new feature in the core that can automatically set the correct template when creating a new resource based on the parent template. I think the best option would be to add a template property called childTemplate which MODX will listen to and preselect the childTemplate when creating a new resource.

To explain this, please see example below:
- Home: template Home (Template ID: 1)
- News overview: template News overview (Template ID: 2)
-- News detail: template News detail (Template ID: 3)
- Contact: template Contact

We add the template property childTemplate to the News overview template and set the value to 3 (The ID of the news detail template).

Now when creating a new child resource for the news overview resource, the template will be preselected to the News detail template.

Any thought about this feature/solution?

Mark-H:
Related:

#13700

sdrenth:
I'm going to try and fix this today #modxbughunt

opengeek:
#13700 was added for inclusion in 2.7.

sdrenth:
@opengeek Can this issue be reopened? Because my solution is setting the default template by using the template property and in my opinion this would work perfectly in addition to #13700. Although the fix is quite similar, there is a difference.

Issue #11529, opened 2014-06-17
[2.3] Feature request - "add" button to every resource in tree


eighthday:
The add button (below) is great on the context, but would be really useful if it was also present next to every resource, so you can create a new child without right clicking. 

![screen shot 2014-06-17 at 12 33 29](https://cloud.githubusercontent.com/assets/4104767/3299497/4384c63c-f613-11e3-857a-195615cd74ea.png)

Perhaps the button should only be show on mouseover?

inreti:
+1 for that. and also: have a setting (per resource? template?) which allows to choose which templates the user is allowed to add as new child. So we do not "Add child _resource" and choose templates afterwards, but have something like "Add blog article", "Add Foto-Gallery", "Add basic page" according to the actually allowed templates _there_

alroniks:
PR is merged, so closing this issue.

PR #13754, opened 2018-02-09
Set default template based on template property childTemplate

sdrenth:
### What does it do?
Rewritten shortif containing multiple shortifs to if statement and added a check if template property childTemplate is set, then the default template is set to this value.

### Why is it needed?
This PR makes it possible to control the default templates that is used for childresources of a resource with a specific template. 

### How to use
Simply add the template property **childTemplate** to a template and set it to the desired child template id. Now you can create a child resource which will automatically have the template set to the configured childTemplate.

### Related issue(s)/PR(s)
Issue #13753

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -188,7 +188,23 @@ class ResourceCreateManagerController extends ResourceManagerController {
      * @return int
      */
     public function getDefaultTemplate() {
-        $defaultTemplate = (isset($this->scriptProperties['template']) ? $this->scriptProperties['template'] : (!empty($this->parent->id) ? $this->parent->get('template') : $this->context->getOption('default_template', 0, $this->modx->_userConfig)));
+        $defaultTemplate = $this->context->getOption('default_template', 0, $this->modx->_userConfig);
+        if (isset($this->scriptProperties['template'])) {
+            $defaultTemplate = $this->scriptProperties['template'];
+        } elseif (!empty($this->parent->id)) {
+            $defaultTemplate = $this->parent->get('template');
+
+            $parentTplObj = $this->modx->getObject('modTemplate', $this->parent->get('template'));
+            if ($parentTplObj) {
+                $props = $parentTplObj->getProperties();
+
+                $childTpl = !empty($props) ? $this->modx->fromJSON($props['childTemplate']) : null;
+                if ($childTpl) {
+                    $defaultTemplate = $childTpl;
+                }
+            }
+        }
+
         $userGroups = $this->modx->user->getUserGroups();
         $c = $this->modx->newQuery('modActionDom');
         $c->innerJoin('modFormCustomizationSet','FCSet');


manager/controllers/default/resource/create.class.php
```diff
@@ -188,7 +188,23 @@ public function process(array $scriptProperties = array()) {
...
+        if (isset($this->scriptProperties['template'])) {
+            $defaultTemplate = $this->scriptProperties['template'];
+        } elseif (!empty($this->parent->id)) {
+            $defaultTemplate = $this->parent->get('template');
+
+            $parentTplObj = $this->modx->getObject('modTemplate', $this->parent->get('template'));
+            if ($parentTplObj) {
+                $props = $parentTplObj->getProperties();
+
+                $childTpl = !empty($props) ? $this->modx->fromJSON($props['childTemplate']) : null;
```
Mark-H: Is `$this->modx->fromJSON($props['childTemplate'])` right? Shouldn't it be `$this->modx->fromJSON($props)`?

Mark-H:
Similar idea (but different implementation) #13700

manager/controllers/default/resource/create.class.php
sdrenth: @Mark-H It returns the childTemplate if it is set, so it's a quicker way to retrieve the setting directly

sdrenth:
@Mark-H Regarding #13700, I think it could coexist perfectly next to this PR, if the childTemplate is not set then the logic of PR #13700 could be handled.

wuuti:
If that is implemented it would be nice if the context menu on the tree would show something like "__Create > 'ChildTemplateName'__" instead of just "__Create > Document__".

Related to https://github.com/modxcms/revolution/issues/11529.

sdrenth:
@wuuti Although its a great idea I don't have the time to add that feature to this PR. It would be nice if someone could add that feature in a seperate PR.

Would like to see this PR reviewed and merged into the core :)

opengeek:
If you can rebase this on top of the latest changes and force push it to update this PR, that would be awesome, because they are conflicting with #13700.

opengeek (changes requested):
If you can rebase this on top of the latest changes and force push it to update this PR, that would be awesome, because they are conflicting with #13700.

manager/controllers/default/resource/create.class.php
alroniks: @sdrenth Actually no. If `$props['childTemplate']` contains string or integer it works because it is a valid json entity. But there are may happen errors when `childTemplate` not found in `$props`.
Safe variant would be:
```
$props = $parentTplObj->getProperties();

$childTpl = !empty($props) && isset($props['childTemplate']) 
    ? $props['childTemplate'] 
    : null;
```
Also, additionally childTemplate value can be forced to integer by `intval($props['childTemplate'])`

alroniks (changes requested):

alroniks:
Hi @sdrenth Do you have time to rebase and fix conflicts in this PR in this week. Otherwise, we will have to decline this PR because we limited by time of 2.7 release.

sdrenth:
@Alroniks unfortunately i wont have time to fix it this week.

alroniks:
@sdrenth Thanks for working on this improvement, Sander! Unfortunately, we're wrapping up MODX 2.7 very soon as I've mentioned before and won't be able of accepting this pull request for now. If possible, please send a new pull request targeting the 3.x branch to get child templates detecting included in MODX 3 instead. Thank you!

Issue #13611, opened 2017-09-19
switching language in process of setup


meshkov:
### Summary
Setup switching language during installation or updating modx

### Step to reproduce
Install or update modx with Russian language. During the installation, the Russian language is switched to English language.
on page /setup/index.php?action=install - Ru
on page /setup/index.php?action=complete - Eng

### Observed behavior
![jamwsmk3mm](https://user-images.githubusercontent.com/1703670/30584986-174132e2-9d34-11e7-98d3-e4c6507fdce7.gif)

### Expected behavior
Does not switch language during installation or updating modx.

### Environment
MODX 2.5.7 and above.

sdrenth:
@meshkov I'm going to try and fix this today #modxbughunt

OptimusCrime:
@modxbot close

PR #13755, opened 2018-02-09
Fixing complete step during setup would always be in EN language

sdrenth:
### What does it do?
It sets the language setting again after clear installation setting so the complete step is shown in correct language.

### Why is it needed?
During complete step in setup the settings are cleared, which would also clear the language setting. Thats why the complete step would always be shown in English.

### Related issue(s)/PR(s)
Issue #13611

### DIFF
--- a/setup/controllers/language.php
+++ b/setup/controllers/language.php
@@ -14,6 +14,7 @@ if (!empty($_POST['proceed'])) {
         $language = $_REQUEST['language'];
     }
 
+    $cookiePath = preg_replace('#[/\\\\]$#', '', dirname(dirname($_SERVER['REQUEST_URI'])));
     setcookie('modx_setup_language', $language, 0, $cookiePath . '/');
 
     unset($_POST['proceed']);
--- a/setup/controllers/language.php
+++ b/setup/controllers/language.php
@@ -7,12 +7,15 @@
  */
 /* parse language selection */
 if (!empty($_POST['proceed'])) {
-    $language= 'en';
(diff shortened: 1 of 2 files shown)


gpsietzema:
I'll go test this now at the bug hunt.

gpsietzema:
@sdrenth I tested your fix, but still end up with the default en-thankyou page:

screen shot 2018-02-09 at 10 58 56

What I did:
- New install within your branch
- Clear caches (browser/core) 
- Run setup in incognito
- Choose 'nl' at the first-setup page
- End up at an English thankyou-page

Mark-H:
Also confirmed this doesn't fix the problem in my testing.

sdrenth:
@gpsietzema @Mark-H After some more testing it sometimes did and sometimes didn't work. 

I already found the cause of the problem and will update the PR. The problem was with an incorrect check which led to a language cookie being set incorrectly.

sdrenth:
I've updated the PR. Thanks for testing guys!

gpsietzema:
Will test :)

gpsietzema:
I can confirm it works now. Awesome work! 👍

setup/controllers/language.php
```diff
@@ -7,12 +7,15 @@
  */
 /* parse language selection */
 if (!empty($_POST['proceed'])) {
-    $language= 'en';
-    $cookiePath = preg_replace('#[/\\\\]$#', '', dirname(dirname($_SERVER['REQUEST_URI'])));
```
Mark-H: The cookie path is still being used below on line 17 (15), which is now undefined.

sdrenth:
@Mark-H Thanks! I've fixed that with the latest commit

Issue #13637, opened 2017-10-13
Remove maxWidth from west/east splitbar


pepebe:
### Summary
According to the extjs docs, the max-Width for the splitbar is 2000px. Setting a max-width is optional. 

https://github.com/modxcms/revolution/blob/31e663b05930373e4ffd5eb60ba1ce4c8f7f4980/manager/assets/modext/core/modx.layout.js#L144

sets this value to 800.

### Step to reproduce
Move the splitbar to the right.

### Observed behavior
You can't move the splitbar beyond the 800px limit.

### Expected behavior
I can see no apparent need to do so. Removing this property would give the user the ability to decide whats appropriate. 

If there are use cases where this would destroy something, then there should at least be a system setting that controls that behaviour.

### Reasons why this is important for me.
The western region can be used for more things than displaying resource, elements and files. 

Examples are alternative views for the file browser, a preview function , etc.

Removing unnecessary restrictions will make it easier to work with this region. 

### Environment
Latest MODx version and all versions before.

Mark-H:
I've seen your slack messages for what you're trying to do, and I'm not against making that better :D

Finetuned:
modxbughunt#3

Mark-H:
Fixed in #13756

Issue #13309, opened 2017-02-24
Help pages shows old documentation-link


gpsietzema:
**PLEASE NOTE - THIS ISSUE IS FOR CREATING A CONTRIBUTORS VIDEO - PLEASE DO NOT REMOVE/CLOSE/TOUCH IT**

### Summary
?a=help shows rtfm.modx.com link, should be docs.modx.com

### Step to reproduce
Go to the help page and click 'Documentation'.

### Observed behavior
It goes to http://rtfm.modx.com/, which redirects you to https://docs.modx.com/

### Expected behavior
Go straight to https://docs.modx.com/

### Environment
MODX 2.5.5

gpsietzema:
Test comment with #modxbughunt

Mark-H:
Will you post a test fix too? #modxbughunt :trollface:

mrhaw:
This will need to be updated in all lexicons: 
https://github.com/modxcms/revolution/search?utf8=%E2%9C%93&q=rtfm&type=Code

- core/lexicon/en/about.inc.php 

» https://github.com/mrhaw/revolution/commit/91f6981855c386f246dbd19ef82ec111877952cf

OptimusCrime:
@mrhaw As far as I can tell you did not open a PR to this?

mrhaw:
Here it is https://github.com/mrhaw/revolution/pull/1

OptimusCrime:
@mrhaw Ah. You opened a PR in your own repository. You have to open one in this one (modxcms) :)

OptimusCrime:
@modxbot close

I think?

PR #13756, opened 2018-02-09
Addressed  issue #13637 #modxbughunt

Finetuned:
### What does it do?
Removed the maxSize restriction on the west region allowing the splitbar to be moved further than 800px. As @pepebe says, this will allow the west region to be used for new functionality.

### Why is it needed?
To allow the splitbar to be made wider.

### Related issue(s)/PR(s)
#13637

### DIFF
--- a/manager/assets/modext/core/modx.layout.js
+++ b/manager/assets/modext/core/modx.layout.js
@@ -141,7 +141,6 @@ Ext.extend(MODx.Layout, Ext.Viewport, {
             ,split: true
             ,width: 310
             ,minSize: 288
-            //,maxSize: 800
             ,autoScroll: true
             ,unstyled: true
             ,collapseMode: 'mini'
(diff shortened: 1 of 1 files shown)


Mark-H:
@Finetuned Could you double check the issue number you refered to? #13309 isn't about a maxSize restriction.

manager/assets/modext/core/modx.layout.js
```diff
@@ -141,7 +141,7 @@ Ext.extend(MODx.Layout, Ext.Viewport, {
             ,split: true
             ,width: 310
             ,minSize: 288
-            ,maxSize: 800
+            //,maxSize: 800
```
Mark-H: Could you please remove it instead of commenting out? Makes the code cleaner.

Mark-H:
Thanks @Finetuned!

PR #13366, opened 2017-03-03
You cant delete/unpublish the start_page anymore by the checkboxes on update resource page #10218 #modxbughunt

Oetzie:
### What does it do?
You cant delete/unpublish the start_page anymore by the checkboxes on update resource page

### Why is it needed?
Describe the issue you are solving.

### Related issue(s)/PR(s)
#10218

PR #13760, opened 2018-02-09
You cant delete/unpublish the start_page anymore by the checkboxes on update resource page #10218 #modxbughunt

Oetzie:
### What does it do?
This fix checks the site_start, error_page and site_unavailable_page context settings to prevent that a user can delete/unpublish the site_start, error_page and site_unavailable_page page.

### Why is it needed?
The resource delete, update and unpublished processors did not check for the context settings site_start, error_page and site_unavailable_page. So you could delete the site_start, error_page and site_unavailable_page pages.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/10218 and https://github.com/modxcms/revolution/pull/13366


gpsietzema:
Will test this now at the bug hunt.

gpsietzema:
It partly works.

In my test environment, I have set the following **system settings**:
- site_start: 3
- error_page: 4
- site_unavailable_page: 5

I tested removing/unpublishing Resource ID3/4/5 I get your new notices, which is good.

After that, I made these three context-settings in my web-context:
- site_start: 6
- error_page: 7
- site_unavailable_page: 8

If I try to remove/unpublish Resource ID 6/7/8 I get your messages again.

However, when I try to remove/unpublish Resource ID 3/4/5, I still get that message, so it is still remembering the System Settings. Shouldn't the Context-settings overrule the system settings? I should be able to remove/unpublish Resource ID 3/4/5 in this example I think.

Oetzie:
@gpsietzema there is a check that checks if the current resource ID is the same as the context setting OR the system setting. Do you think its better to check if context setting is set then only check the context setting otherwise check the system setting. So not check both settings, agree?

gpsietzema:
I agree :)

Mark-H:
I disagree.

When you have a single context, it doesn't really make sense to configure it on the context level, so your test @gpsietzema does not really reflect how it would be used. MODX may still use the system settings if you have multiple contexts, and don't define the settings for web for example. What you're defining now doesn't make much sense - why would you set system settings to one thing, and context settings to another, with a single context?

Based on your description of the test, I think it's behaving as it should. It checks both context and system settings.

Oetzie:
@Mark-H @gpsietzema I made a fix, the fix checks if the current context has a context setting 'site_start', 'error_page' or 'site_unavailable_page'. If the context setting does not exists or is empty the default system setting will be checked. It works in a multi and single context site.

gpsietzema:
@Mark-H Those system settings are already there, by default. So if we're overruling with context-settings the system settings should be ignored. Especially in multi-context environments. So I'm guessing we're having the same opinion and @Oetzie latest changes fix that :)

Mark-H:
As long as a resource is mentioned in the system setting, it should not be deletable/unpublishable. That seems like a different opinion than not checking the system settings at all if a context setting is set?

Oetzie:
@gpsietzema @Mark-H when you both come with the same solution I can fix that :-)

gpsietzema:
@Oetzie we discussed it, @Mark-H wins :) So we have to check both 👍  So your initial PR was right.

Oetzie:
@gpsietzema changed back to the first commit ;-)

core/model/modx/processors/resource/unpublish.class.php
```diff
@@ -60,15 +71,6 @@ public function process() {
         return $this->success('',$this->resource->get(array('id')));
     }
 
-    /**
```
opengeek, line 63: Removing this public API function is a BC break. This cannot be merged in this current state until 3.x.

opengeek (changes requested):

gpsietzema:
@opengeek @Oetzie we could target 3.x then... ?

core/model/modx/processors/resource/unpublish.class.php
opengeek: The point of my review was to suggest the public API method `isSiteStart()` should not be removed at all. This change is not necessary and should not be done, even in 3.x without first releasing it with a deprecated mark. But again, I don't see removing this method to be necessary in the first place. Just don't use it for the purpose it was used prior to the other changes this PR makes. This SHOULD be integrated into 2.x for 2.7 release and can be if we can make that change.

Mark-H:
If the `isSiteStart` method is restored this can be merged into 2.x - for now this is unfortunately not suitable and has been waiting for an update for a while. Please reopen when that's back :)

Issue #13750, opened 2018-02-06
The wrong mediasource is selected when updating a static element


Oetzie:
### Summary
The wrong mediasource is selected when updating a static element.

### Step to reproduce
Make an static element with a custom mediasource and save the static element. After saving and refreshing the window the wrong mediasource (default mediasource) is selected for the element.

### Observed behavior
The mediasource of the 'static_file' is set by a listener on  the 'mediasource' field. But when you update a static element you don't change the mediasource so the listener wont be fired. 

### Expected behavior
Add the following line to the config of the 'static_file' field for each element panel (plugins, chunks, templates, snippets). With this line the mediasource will be set to the record or the default mediasource on the page load and not only with a listener on the 'mediasource' field.

```,source: config.record.source != null ? config.record.source : MODx.config.default_media_source```

This can be fix with this pull request: https://github.com/modxcms/revolution/pull/13646

### Environment
All MODX versions

Oetzie:
I'm going to try and fix this today #modxbughunt

Mark-H:
Fixed per #13762 in 2.6.2

PR #13762, opened 2018-02-09
The wrong mediasource is selected when updating a static element #13750 #modxbughunt

Oetzie:
### What does it do?
A default mediasource will be set on the 'static_file' field. If the current element has a mediasource setting this setting will be used otherwise the 'default_media_source' system setting will be used.

### Why is it needed?
The wrong mediasource is selected when updating a static element.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13750


Mark-H:
Thanks @Oetzie!

Issue #13752, opened 2018-02-07
System setting to change the error log path and error log name


IamNeo1:
I am looking for an system setting to change the default error log path and the name.
I am logging to a different path with an error_handling snippet and i want to see the log at the modx error log page. Currently it is empty, because it expects the defaut path an name ".../cache/logs/error.log"

It is similar to this request #12352

OptimusCrime:
Please use the issue template.

christianseel:
I'm working on this right now…

PR #13751, opened 2018-02-07
System setting to change the error log path and error log name

IamNeo1:
I am looking for an system setting to change the default error log path and the name.
I am logging to a differnt path with a error_handling snippet and want to see the log at the modx error log page.

it is like this PR #12352

PR #12352, opened 2015-02-08
add system setting for error.log which defaults to 'logs/' and uses this configurable path for logging

gerrit1986:
worked on this at modx ccc 2015. 
i hope it does not break anything. 
any feedback is appreciated.
https://github.com/modx-ccc-2015/whishlist/issues/39

PR #13763, opened 2018-02-09
Make error.log location customizable #13752

christianseel:
### What does it do?
Adds 2 new system settings for the MODX error log to specify the filename (`error_log_filename`) and filepath (`error_log_filepath`).

Note: Changing the error log filename or path was already possible with `setLogTarget`, however this is not really convenient, so we added those system settings. The system settings internally updated the log target.

The manager error log processors are updated to also finally support the log target and show the correct log when the target has been customized.

### Why is it needed?
This makes it possible to fully customize the location of the MODX error log. Now one could move the error log outside of the MODX directory and name it modx.log easily with the system settings. Also the error log view in the manager now works correctly with custom locations.

### Related issue(s)/PR(s)
Issue: #13752
Failed PRs: #12352 by gerrit1986 and #13751 by IamNeo1

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2110,7 +2110,7 @@ $settings['error_log_filename']->fromArray(array (
     'value' => 'error.log',
     'xtype' => 'textfield',
     'namespace' => 'core',
-    'area' => 'site',
+    'area' => 'system',
     'editedon' => null,
 ), '', true, true);
 $settings['error_log_path']= $xpdo->newObject('modSystemSetting');
@@ -2119,7 +2119,7 @@ $settings['error_log_path']->fromArray(array (
     'value' => '',
     'xtype' => 'textfield',
     'namespace' => 'core',
-    'area' => 'site',
+    'area' => 'system',
     'editedon' => null,
 ), '', true, true);
 return $settings;
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2104,4 +2104,22 @@ $settings['log_snippet_not_found']->fromArray(array (
     'area' => 'site',
     'editedon' => null,
 ), '', true, true);
+$settings['error_log_filename']= $xpdo->newObject('modSystemSetting');
+$settings['error_log_filename']->fromArray(array (
+    'key' => 'error_log_filename',
+    'value' => 'error.log',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'site',
+    'editedon' => null,
+), '', true, true);
+$settings['error_log_path']= $xpdo->newObject('modSystemSetting');
+$settings['error_log_path']->fromArray(array (
+    'key' => 'error_log_path',
+    'value' => '',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'site',
+    'editedon' => null,
+), '', true, true);
 return $settings;
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2113,9 +2113,9 @@ $settings['error_log_filename']->fromArray(array (
     'area' => 'system',
     'editedon' => null,
 ), '', true, true);
-$settings['error_log_path']= $xpdo->newObject('modSystemSetting');
-$settings['error_log_path']->fromArray(array (
-    'key' => 'error_log_path',
+$settings['error_log_filepath']= $xpdo->newObject('modSystemSetting');
+$settings['error_log_filepath']->fromArray(array (
+    'key' => 'error_log_filepath',
     'value' => '',
     'xtype' => 'textfield',
     'namespace' => 'core',
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -524,6 +524,18 @@ class modX extends xPDO {
             $this->getCacheManager();
             $this->getConfig();
             $this->_initContext($contextKey, false, $options);
+            $logTarget = $this->getLogTarget();
+            if ($logTarget === 'FILE') {
+                $options = array();
+                $filename = $this->getOption('error_log_filename', $options, '');
+                if (!empty($filename)) $options['filename'] = $filename;
+                $filepath = $this->getOption('error_log_filepath', $options, '');
+                if (!empty($filepath)) $options['filepath'] = $filepath;
+                $this->setLogTarget(array(
+                    'target' => 'FILE',
+                    'options' => $options
+                ));
+            }
             $this->_loadExtensionPackages($options);
             $this->_initSession($options);
             $this->_initErrorHandler($options);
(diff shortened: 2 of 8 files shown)


core/xpdo/xpdo.class.php
```diff
@@ -2076,8 +2076,10 @@ protected function _log($level, $msg, $target= '', $def= '', $file= '', $line= '
         $content= @ob_get_contents();
         @ob_end_clean();
         if ($target=='FILE' && $this->getCacheManager()) {
-            $filename = isset($targetOptions['filename']) ? $targetOptions['filename'] : 'error.log';
-            $filepath = isset($targetOptions['filepath']) ? $targetOptions['filepath'] : $this->getCachePath() . xPDOCacheManager::LOG_DIR;
+            $defaultFilename = !empty($this->getOption('error_log_filename')) ? $this->getOption('error_log_filename') : 'error.log';
```
Mark-H: This is failing on PHP 5.4 with error:
> Fatal error: Can't use method return value in write context in /home/travis/build/modxcms/revolution/core/xpdo/xpdo.class.php on line 2079


The problem is the `empty` method being called on the result of `$this->getOption()`. Old PHP doesn't seem to like that.

Instead of using these ternary operators, you can use the following:

`$defaultFilename = $this-getOption('error_log_filename', null, 'error.log', true);`

The last `true` indicates an empty() check will be run on the value. Looks a bit cleaner, too. 

This comment applies to the other files, too.

Mark-H (changes requested):
Aside from the failure on PHP 5.4, this also makes changes to xPDO. These changes have to be submitted to the xPDO repo at https://github.com/modxcms/xpdo and once integrated there, can be integrated into MODX.

christianseel:
@Mark-H PR has been updated with your feedback. Thanks!

XPDO PR created at: https://github.com/modxcms/xpdo/pull/131

Mark-H (dismissed):
Looks good and can be considered for merging when the xPDO PR is merged.

christianseel:
After feedback from @opengeek, the PR is now updated to make use of the `xpdo::logTarget` which already supported changing the filename and filepath.

The PR now still adds those 2 new settings which are then set as log target.

The manager processors for the error log have been updated to also use the log target, so they keep showing the correct error file.

_build/data/transport.core.system_settings.php
```diff
@@ -2104,4 +2104,22 @@
...
+$settings['error_log_filepath']= $xpdo->newObject('modSystemSetting');
+$settings['error_log_filepath']->fromArray(array (
+    'key' => 'error_log_filepath',
+    'value' => '',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'system',
+    'editedon' => null,
+), '', true, true);
 return $settings;
```
bezumkin, line 2125: What about creating new settings on system upgrade?

core/model/modx/modx.class.php
```diff
@@ -524,6 +524,18 @@ public function initialize($contextKey= 'web', $options = null) {
...
             $this->getConfig();
             $this->_initContext($contextKey, false, $options);
+            $logTarget = $this->getLogTarget();
+            if ($logTarget === 'FILE') {
+                $options = array();
+                $filename = $this->getOption('error_log_filename', $options, '');
+                if (!empty($filename)) $options['filename'] = $filename;
+                $filepath = $this->getOption('error_log_filepath', $options, '');
+                if (!empty($filepath)) $options['filepath'] = $filepath;
+                $this->setLogTarget(array(
```
bezumkin: I think there should be some protection against wrong settings, like
```php
$options['filepath'] = rtrim($filepath, '/') . '/';
```

bezumkin (changes requested):

_build/data/transport.core.system_settings.php
bezumkin: Checked - they are created on upgrade, it is ok.

gpsietzema:
@christianseel What do you think of @bezumkin 's suggestions?

christianseel:
@gpsietzema @bezumkin Now updated! Sorry for the huge delay…

core/model/modx/modx.class.php
```diff
@@ -524,6 +524,18 @@ public function initialize($contextKey= 'web', $options = null) {
             $this->getCacheManager();
             $this->getConfig();
             $this->_initContext($contextKey, false, $options);
+            $logTarget = $this->getLogTarget();
+            if ($logTarget === 'FILE') {
+                $options = array();
+                $filename = $this->getOption('error_log_filename', $options, '');
+                if (!empty($filename)) $options['filename'] = $filename;
```
alroniks, line 531: Not a strictly required but it's better to follow PSR rules and use `if` with {} to avoid unexpected behavior in the future somehow.

alroniks:
I've tested several cases with different filenames and file places. All works fine when path inside MODX installation directory. 

I have MODX in /var/www/html/public/ and I tried save log files in /var/www/html/logs (directory already exists) but it failed. Not any file was created.

*NOTE* Maybe it doesn't work only on my environment, need to check additionally. Getting of data works. Inside working directory all works fine. BC is preserved.

Testing matrix:

|| Write | Clear | Refresh | Download |
|-|-|-|-|-|
| Original state | [x] | [x] | [x] | [x] |
| Another file name | [x] | [x] | [x] | [x] |
| Another file path | [x] | [x] | [x] | [x] |
| Another file name and path | [x] | [x] | [x] | [x] |
| Outside MODX installation | [-] | [-] | [-] | [-] |

core/model/modx/processors/system/errorlog/download.class.php
```diff
@@ -10,16 +10,22 @@ public function checkPermissions() {
...
+        $filename = $this->modx->getOption('filename', $logTarget['options'], 'error.log', true);
+        $filepath = $this->modx->getOption('filepath', $logTarget['options'], $this->modx->getCachePath() . xPDOCacheManager::LOG_DIR, true);
+        $file = rtrim($filepath, '/').'/'.$filename;
+        if (!file_exists($file)) {
             return $this->failure();
         }
         header('Content-Type: application/force-download');
-        header('Content-Length: ' . filesize($f));
+        header('Content-Length: ' . filesize($file));
         header('Content-Disposition: attachment; filename="error.'.time().'.log');
```
alroniks, line 25: @christianseel How do you think, does it make sense to use here name form system settings also?

alroniks (approved):
Described functions work as expected except the case when log file outside working directory (file permissions issue on my env).

alroniks:
@bezumkin Please, have a look and approve if you requests were fixed.
@opengeek Please, have a look too. And is it enough such summary of testing?

bezumkin (approved):

christianseel:
I‘d bet that the log outside of the working dir is something like a base_dir restrictions (that was the name of that annoying permission thingy, right?). Especially if the folder already existed and is writable, nothing should prevent MODX from writing the log. I think we’re fine if everything works inside the working directory.

Issue #13739, opened 2018-01-23
No fall back to context key if context name isn't specified.


bennyb:
### Summary
A number of sites where i've upgraded from older version don't contain a context name, so nothing is displayed in the context drop down.  Pre 2.6.1 both the context name and key were displayed so if no name was added the key would still display but in 2.6.1 only the name is displayed.

### Step to reproduce
No value is set for the context name.
Manager >> Access Control Lists >> Update a user group>> Permissions >> Context Access >> Context drop down.
This is the same for any combobox that lists the context.

### Observed behavior
Combobox doesn't display a name.

### Expected behavior
Should fall back to displaying the key
### Environment
MODX version 2.6.1

Oetzie:
I'm going to try and fix this today #modxbughunt

alroniks:
PR is merged, so closing this one.

PR #13765, opened 2018-02-09
No fall back to context key if context name isn't specified. #13739 #modxbughunt

Oetzie:
### What does it do?
Just a fallback on the key in de getlist processor.

### Why is it needed?
When a context has no name (from updates old MODX to new MODX) an empty row will show in the context combos.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13739

### DIFF
--- a/core/model/modx/processors/context/getlist.class.php
+++ b/core/model/modx/processors/context/getlist.class.php
@@ -60,6 +60,7 @@ class modContextGetListProcessor extends modObjectGetListProcessor {
     public function prepareRow(xPDOObject $object) {
         $contextArray = $object->toArray();
         $contextArray['perm'] = array();
+        $contextArray['name'] = $object->get('name') != '' ? $object->get('name') : $object->get('key');
         if ($this->canCreate) {
             $contextArray['perm'][] = 'pnew';
         }


superspace:
@Oetzie I tested the issue by deleting the context name and can reproduce it. With your PR the dropdown is populated with the key as fallback value. Maybe it's a little bit irritating to see the fallback value on the contexts-manager-page in the column "Name". I think this should be empty.

opengeek:
I'm trying to figure out why the key was removed from this list processor. I think it would be better to put the key back into the list display.

Oetzie:
@opengeek I am not sure what you trying to say. But the name field is added and thats now leading in the back-end for displaying the context in the tree and combo boxes and stuff because it's cleaner then the key. But when the name is empty you don't see the context in combo boxes, so now the name has a fallback on the key... So it is fixed.....

opengeek:
I'm saying this should not have been switched to just display the name. It would have been preferable to include both so if no name was provided it would still be identifiable, but without arbitrarily putting the key in the `name` field, as @superspace indicated in his comment does not make sense in the Contexts manager page. In fact, it is misleading.

opengeek:
In other words, I believe this should be fixed by changing the display in the combo elements, not in this list processor.

core/model/modx/processors/context/getlist.class.php
```diff
@@ -60,6 +60,7 @@ public function prepareQueryBeforeCount(xPDOQuery $c) {
     public function prepareRow(xPDOObject $object) {
         $contextArray = $object->toArray();
         $contextArray['perm'] = array();
+        $contextArray['name'] = $object->get('name') != '' ? $object->get('name') : $object->get('key');
```
opengeek, line 63: I'd prefer if this was done in the combo element rather than in the list processor, as this affects other UI elements that use this processor which could confuse users if the `name` field which is empty now shows the same value as the `key`.

opengeek (changes requested):

core/model/modx/processors/context/getlist.class.php
Jako: Since it is a bit difficult to change the displayField in a combo during rendering, you should add a new array key there (i.e. `display_name`) and fill it with those values:

```
$contextArray['display_name'] = $object->get('name') != '' ? $object->get('name') : $object->get('key');
```

Afterwards you could refer that key in the combo (change `name` to `display_name`):
https://github.com/modxcms/revolution/blob/05d23c0478fb7cf26a5921ebd9a31d2c7fbfccfc/manager/assets/modext/widgets/core/modx.combo.js#L336-L338

Oetzie: @opengeek I do not agree with you? So if you have 3 different combo boxes (maybe people make a own combo box instead of the context combo box) you will still have empty rows.

The getlist processor is the source of the 'data' so you need to change it in the source, not on 10 different places in 10 different UI elements... In a normal situation the name is set, if not it has a fallback on the key and the key will be the name of the context... The name of a context should never be empty...

Jako: It is never a good idea to use the default field key and change the value. That key/value is posted back i.e. during inline editing and could lead to unwanted values in that table. It could be also posted back from a MODx.window.

This was the reason, why it was removed in the processor. See https://github.com/modxcms/revolution/issues/13371 and https://github.com/modxcms/revolution/commit/62a3d33348661cb46836b20f4cc69c132ad6cd0a

There is only one `modx-combo-context` xtype, so you don't have to change it in 10 different places.

Oetzie: @Jako thats is a different issue, it ADDS something to a name instead of a fallback, and yes then you will get unwanted values... The name of a context should be required, and a fallback for the name is needed because a MODx update.

But I won't fix this in ExtJS because thats not the good solution.

Jako: It is not a very different issue. Both change the original value, which should not happen during a getlist (even if the new value 'should' be right).

Issue #12706, opened 2015-10-09
Silent Fail on Login where User doesn't have mgr Context access.


jaygilmore:
Uers who are in User Groups that do not have access permissions for the manager (`mgr`) context who attempt to login to the MODX Manager will see the login page reload and silently fail. It appears that upon submit, the manager page reloads, a 301 redirects to the manager with no trailing slash eg: `manager` and then 302 to `manager/`. 

This should not fail silently, should it?

bezumkin:
I will try to fix it

OptimusCrime:
Is this fixed then?

gpsietzema:
Jup 👍 🎉

bezumkin:
I think Optimus asked about to close this issue. Lets try this magic!

@modxbot close

PR #13767, opened 2018-02-09
Fix silent fail on login

bezumkin:
### What does it do?
Return error if user has no permission for using manager

### Why is it needed?
If user successfully login to manager, but has no permission to use backend, login pagewill  just reload without any errors. That can be very confusing.

### Related issue(s)/PR(s)
#12706

### DIFF
--- a/core/model/modx/processors/security/login.class.php
+++ b/core/model/modx/processors/security/login.class.php
@@ -374,11 +374,18 @@ class modSecurityLoginProcessor extends modProcessor {
      */
     public function afterLogin() {
         $this->addSessionContexts();
+        if ($this->loginContext == 'mgr') {
+            $this->modx->user = null;
+            $this->modx->getUser('mgr', true);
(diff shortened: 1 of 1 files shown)


superspace:
I'm testing your patch and got the following result:

1. After a login with a user without permissions to use the backend i get the message "Login failed".
2. I enter the username and password of a user with all required permissions and hit the login button – but the only thing happening is a page reload. No message, no successful login. 
3. I repeat the first step and i get again the message "Login failed".
4. I try another (incorrect) password and hit the login button – the only thing happening is a page reload without a message
#modxbughunt

bezumkin:
If you can't login to manager you will see an error, but the next attempt will just reload window. It is because user is actually logged it but can't see backend, so manager controller will check it and logout him on the next request. That is why we have a page reload instead of new attempt to login.

I made a new commit, please, check it.

superspace:
I checked it again and now it behaves as i would expect it.

PR #13768, opened 2018-02-09
Fix incorrect view_url after changing the resource alias #13740

### LINKED ISSUE #13740: Changing Alias before View Issue
### Summary

If you edit any resource, change an alias, save, then click the view button the link loads the old alias instead of the new one.  You can do it over and over, MODX always loads the alias from before the change is made and saved.

### Step to reproduce

1. Go to a saved resource
2. Edit the alias field, save, then view resource
3. When viewing the resource, the page loads fine
4. Go back to the resource, edit the alias field, save and view the resource again
5. This results in a page not found until you refresh the page (or double save) in the manager

### Observed behavior

Incorrect (old) URL opened.

### Expected behavior

Updated URL opened.

### Environment

MODX 2.6.1 and before

jdaehne:
### What does it do?
makeUrl() was loading the cached aliasMap. It reloads the Content after saving the resource to receive the updated aliasMap with the new set alias.

### Why is it needed?
After changing the alias and saving the resource, the "publish button" showed/opened the old URL (preview_url).

### Related issue(s)/PR(s)
#13740

### DIFF
--- a/core/model/modx/processors/resource/update.class.php
+++ b/core/model/modx/processors/resource/update.class.php
@@ -726,6 +726,7 @@ class modResourceUpdateProcessor extends modObjectUpdateProcessor {
         }
         $returnArray['class_key'] = $this->object->get('class_key');
         $this->workingContext->prepare(false);
+        $this->modx->reloadContext($this->workingContext->key);
         $returnArray['preview_url'] = '';
         if (!$this->object->get('deleted')) {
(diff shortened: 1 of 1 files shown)


superspace:
I tested several combinations of published/unpublished resources, duplicating, editing  etc. and the "View"-button always opended the correct URL.
#modxbughunt

Mark-H:
@jdaehne Please use a better title next time ;)

Mark-H:
I'm happy this is finally fixed - thanks @jdaehne!

Issue #13626, opened 2017-10-04
Elements (Chunks, Snippets, etc.) Check if there is a more recent update


pepebe:
### Summary
It happens from time to time, especially if you are working on multiple screens. For some reason you have two open tabs displaying the same chunk. They might be on the same window or in different windows, it doesn't really matter. 

You work in the first tab for two hours, save your work and close it. Then you treat yourself to a cup of strong coffee and after returning to your desk, you somehow save the old chunk over your hard work. Sometimes you might notice it at once and you do something about it, sometimes you don't notice and do nothing.

###  Current workaround 
Install versionx or some other kind ov versioning solution.

### Expected behavior
An element should know its last update timestamp BEFORE saving the current version, compare that with the current status in the database and if there is a more recent version already present ask for confirmation.

mindeffects:
Yes, please! But this mechanism is already working for resources so there shouldn't be a need your another solution. All the stuff we need is already on board of MODX. (In case you are using static files, well, this is your "local" problem. ;-) )

OptimusCrime:
The system with locks for resources is very complex. I think another (easier) solution could be used for elements. When the data is fetched from the processor, we compile a checksum (md5 or something) which is used to verify the element state. Upon saving the element, we send the old checksum, re-calculate the current element state (stored in the database) and see if they match. If they differ, then the element state has changed since we fetched the data from the processor. This can prompt the user with a warning, giving the option to cancel saving, or overwrite the changes. I think this should be pretty doable. The problem is how to calculate the cheksum. Just the content/code/markup should be easy enough, but should we include the description too? Property sets?

mindeffects:
Why not a simple check-in-out? If the element is fetched for editing within the manager it is checked in (usename&|timestamp is stored). If the element gets saved and(!) closed it will get checked out. If a second attempt opens the element meanwhile, it'll get a warning and save is disabled.
Maybe to simple?

OptimusCrime:
This will require new fields in all the element tables. This is something that has been discussed for some time, so this might be a good idea. However, this approach will give false positives if you save the element multiple times (without doing any actual changes). Also, checking for user might be error prone, if executed via a processor. So it would have to look at the timestamp only.

rthrash:
Resource locking in Revo is mediocre at best. They expire after an arbitrary time, which I think the default is 5 minutes. Further, I am not certain that closing expires the lock either. Locks should be a lot simpler and much less magic—they're not reliable IMO.

[...]

PR #13770, opened 2018-02-09
Added editedon column for elements and prevent overwriting outdated elements

sdrenth:
### What does it do?
Added editedon column for all elements and on update the editedon is being updated. Also added a check on editedon field to see make sure you are not saving an outdated element overwriting previous changes to the element. If this is the case, an error is shown and the element is not being saved.

The confirmation window has not yet been implementen. But I also question if that would even be desirable behaviour, because then you don't know what values you'll be overwriting. So I think when you receive this error you should manually open a new tab, check the changes in the current element and the changes you wanted to save and then manually merge these changes?

### Why is it needed?
It helps to prevent overwriting changes when working on the same element in for example multiple tabs (As described in issue #13626).

### Related issue(s)/PR(s)
Issue #13626

### NOTE
I've already added the migration scripts, but these are untested and I don't really know how to properly test these. Is this tested when creating a new version of MODX?


Mark-H:
Please use a better title @sdrenth ;)

sdrenth:
@Mark-H haha you're quick, I was still finishing my PR  👍

OptimusCrime:
This is awesome. I am currently working on a extra that does this, but having it in the core is much better.

opengeek (changes requested):
This cannot be accepted as is. All changes to map files *must* be made through the schemas.

bezumkin:
Closing this in favour of #13857

PR #13771, opened 2018-02-09
Bug parser returning unattended message on comments

AmaZili:
### What does it do?
Modified parser in order to properly proccess placeholder with conditional output modifier returning a - sign to be treated as a comment

### Why is it needed?
```
[[[[+majcache:gt=`0`:then=`AC_SupprCache`:else=`-`]]]]
```
generate a messages in the log file like
```
(ERROR @ /var/www/vhosts/yourdomain.tlds/core/model/modx/modparser.class.php : 540) Could not find snippet with name .
(ERROR @ /var/www/vhosts/yourdomain.tld/core/model/modx/modparser.class.php : 540) Could not find snippet with name -.
```

### Related issue(s)/PR(s)
Issue discussed with MarkH during the bughunt#3

### DIFF
--- a/core/model/modx/modparser.class.php
+++ b/core/model/modx/modparser.class.php
@@ -421,7 +421,6 @@ class modParser {
 
         $outerTag= $tag[0];
         $innerTag= $tag[1];
-
         /* Avoid all processing for comment tags, e.g. [[- comments here]] */
         if (substr($innerTag, 0, 1) === '-') {
             return "";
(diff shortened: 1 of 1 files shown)


julienstuder:
#modxbughunt 

@AmaZili @Mark-H  bug is fixed, no more error message in the logs.

Mark-H:
As I've been involved in writing this patch I'll let someone else take care of the integration. This change affects two different use cases:

- Tags like ````[[[[+foo:gt=`0`:then=`$foo`]]]]```` where there is no `else` statement (adding of the `elseif`
- Tags like ````[[[[+foo:gt=`0`:then=`$foo`:else=`- do nothing`]]]]```` where the result of the inner tag is a comment, which was getting processed as if it was a snippet instead of a comment (adding of the `case '-'`)

Issue #13746, opened 2018-01-29
Duplicating a resource does not refresh the context aliasMap


Mark-H:
### Summary
First reported in #13375 (with an incorrect fix), it appears that duplicating a resource in the manager does not refresh the aliasMap automatically. This is causing unnecessary errors from getting logged.

### Step to reproduce

1. First, set the log_level to 3. 
2. Duplicate a resource. 
3. Check the error log to see the errors mentioned below.
4. Refreshing the resource tree causes the errors to get repeated
5. Clear the MODX cache through Manage > Clear Cache
6. Refresh the resource tree to see the errors no longer appear

### Observed behavior
These errors are getting logged for each request:

````
[2018-01-29 22:26:29] (WARN @ /Users/mhamstra/Sites/revolution.local/core/model/modx/modcontext.class.php : 244) `35` was requested but no alias was located.
[2018-01-29 22:26:29] (INFO in modContext::makeUrl @ /Users/mhamstra/Sites/revolution.local/core/model/modx/modcontext.class.php : 321) Resource with id 35 was not found in context web
````

### Expected behavior
Duplicating a resource should cause an update to the contexts' aliasMap, so that the errors do not get logged.

### Environment

Confirmed on MODX 2.6.1/2.x latest.

opengeek:
Those are not errors. That's a warning and info. This needs to use refresh or the PR should be closed.

opengeek:
Or whatever is triggering those warnings should be addressed directly.

PR #13772, opened 2018-02-09
Delete context cache on resource duplicate

bezumkin:
### What does it do?
Delete context cache after resource duplicate

### Why is it needed?
To get rid of errors about not existing resource in alias map

### Related issue(s)/PR(s)
#13746

### DIFF
--- a/core/model/modx/processors/resource/duplicate.class.php
+++ b/core/model/modx/processors/resource/duplicate.class.php
@@ -64,6 +64,7 @@ class modResourceDuplicateProcessor extends modProcessor {
 
         $this->fireDuplicateEvent();
         $this->logManagerAction();
+        $this->clearCache();
 
         return $this->success('', array ('id' => $this->newResource->get('id')));
     }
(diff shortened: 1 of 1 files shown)


AmaZili:
Tested successfully.
No more error in the log file with error_level set to 3

opengeek:
The context cache should never be deleted. It must be refreshed and replaced so the burden of regeneration of the cache is not put onto the front-end requests which can cause problems in high-traffic environments. Could you refactor this to refresh rather than delete it?

bezumkin:
> Could you refactor this to refresh rather than delete it?

No, I can't.

I tried to use refresh the half of day and it wasn't work for me. For some reason it only worked when I added `sleep(3);` after context refresh. At least 3 seconds, not less.

I tried this
```
$this->modx->cacheManager->refresh(array(
    'db' => array(),
    'auto_publish' => array('contexts' => array($this->newResource->get('context_key'))),
    'context_settings' => array('contexts' => array($this->newResource->get('context_key'))),
    'resource' => array('contexts' => array($this->newResource->get('context_key'))),
));
```
and this
```
$this->modx->cacheManager->generateContext($this->newResource->get('context_key'), array(
    'cache_alias_map' => $this->modx->getOption('cache_alias_map'),
));
```
and even this
```
if ($context = $this->newResource->getOne('Context')) {
    $context->prepare(true, array(
         'cache_alias_map' => $this->modx->getOption('cache_alias_map'),
    ));
}
```
No luck without `sleep(3);`. So, the way with deletion of the whole file is the only way that I found to get rid from error.

gpsietzema:
Any ideas @opengeek ?

opengeek:
I think it is better to live with the log messages (these are NOT errors) immediately following the refresh, which should not appear again. Deleting the context cache is not an acceptable solution.

opengeek:
Why was this closed without further discussion?

bezumkin:
I agree with you.

gpsietzema:
@modxbot reopen

gpsietzema:
@bezumkin @opengeek I guess you guys have to re-open it then ;-)

bezumkin:
@gpsietzema Why? Nothing changed.

I agree with Jason that it is just warning and we don't need to delete the cache of a whole context.

gpsietzema:
Oh, then I misunderstood.

PR #13774, opened 2018-02-10
Implemented quick create buttons on tree nodes for documents.

wuuti:
### What does it do?
Solves the feature request in #11529 by adding a plus-icon to every document node in the resource tree. The plus is only shown when you hover over the tree node (though not polluting a bigger tree with hundreds of plusses), and a special tooltip is shown when you hover the icon then.

screen shot 2018-02-10 at 18 58 03

### Why is it needed?
It quickens up the creation of new documents at the places you want. You don't need the context menu any more for this. 

### Related issue(s)/PR(s)
#11529
If #13754 is implemented, it will be quite easy to e.g. replace the tooltip "Create new document" with "Create new _ChildTemplateName_" to make it even easier for editors.


wuuti:
One little thing missing: the parent tree node needs to be expanded after a successful resource creation. But as I am not an extjs expert, I will have to take a deeper look in the code.

wuuti:
Ah, and one more: the tooltip is right now hardcoded and not using lexicon :-(

wuuti:
Fixed that lexicon thingy.

OptimusCrime:
I really like this feature, but the hovering thing does a little against the more modern "mobile first" principles, as this is impossible to simulate on mobile devices. Perhaps something we should address in the total overhaul of the manager some time in the future.

wuuti:
Well, that improves only the _current_ manager. I am aware of mobile problems here, but see also no reason to think about that as longs as the whole rest of the manager is not mobile friendly at all. For now, this will improve the "editing-speed" of many editors (myself as well) a lot and fits quite perfect in the current look-and-feel.

Mark-H:
Many menus needing right clicks is probably worse on mobile. I agree with @OptimusCrime that when a new manager is being made that it should be taken into account, but that doesn't have to stop improvements like this from making it into the core now.

manager/assets/modext/widgets/core/tree/modx.tree.js
```diff
@@ -145,13 +145,50 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
...
                 node.ui.addClass('x-tree-selected');
             }
+
+            var elId = node.ui.elNode.id + '_tools';
+            var el = document.createElement('div');
+            el.id = elId;
+            el.className = 'modx-tree-node-btn-create';
+
+            // for now we only add a create icon to documents
+            if (node.attributes.classKey == "modDocument") {
```
Mark-H: I've not loaded the PR locally yet, but it looks like this may be preventing it from adding the icon on weblinks or custom resource types (collections, moregallery, etc)? Perhaps this onLoad method should call a new method on the tree like `this.prepareChildNode(node)` which the resource tree object can extend to add this button?

manager/assets/modext/widgets/core/tree/modx.tree.js
```diff
@@ -145,13 +145,50 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
...
+                node.ui.elNode.appendChild(el);
+
+                var btn = MODx.load({
+                    xtype: 'modx-button',
+                    text: '',
+                    scope: this,
+                    tooltip: new Ext.ToolTip({
+                        // TODO use dictionary here
```
Mark-H: I think you resolved this todo

manager/assets/modext/widgets/core/tree/modx.tree.js
```diff
@@ -145,13 +145,50 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
...
+                        ,target: this
+                    }),
+                    node: node,
+                    handler: function (btn,evt) {
+                        evt.stopPropagation(evt);
+                        node.getOwnerTree().handleDirectCreateClick(node);
+                    },
+                    iconCls: 'icon-plus-square',
+                    renderTo: elId,
+                    listeners: {
```
Mark-H: Have you tried, instead of the `tooltip` + `listeners`, to add a `description`? I think that'd normally handle the showing/hiding automatically, so might save some code..

Mark-H (changes requested):
I've left some comments with minor tweaks that would be useful.

_build/templates/default/sass/_tree.scss
```diff
@@ -376,6 +376,37 @@
...
+    opacity:0;
+    transition: opacity 0.4s ease-in;
+
+    .x-btn {
+      color: $darkGray !important;
+      opacity:    0.4 !important;
+      transition: opacity 0.2s, color 0.2s;
+      transition-timing: ease-in-out;
+
+      &:hover {
```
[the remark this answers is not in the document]
wuuti: Will do that.

manager/assets/modext/widgets/core/tree/modx.tree.js
wuuti: Well, that was on purpose. I did not see a realistic use case where e.g. one would add a child to a weblink. Chris already mentioned Collections to be a "problematic" case. I will see what I can do here with exts-fu - probably coming back to you for help with that.

manager/assets/modext/widgets/core/tree/modx.tree.js
wuuti: Right.

manager/assets/modext/widgets/core/tree/modx.tree.js
wuuti: To be honest, that code was a copy of the context btn code. That should be changed there as well then. I will try that.

manager/assets/modext/widgets/core/tree/modx.tree.js
Mark-H: On some sites I've seen a parent weblink pointing to the first child (although there's other ways to go about that too), so I don't think it should be making too many judgements on where it can/cannot work.

manager/assets/modext/widgets/core/tree/modx.tree.js
Mark-H: Fair enough! ;)

manager/assets/modext/widgets/core/tree/modx.tree.js
wuuti: Yeah, doing this right now. Due to performance reasons I also thought about adding the buttons only on hover of the tree nodes (and not in advance for all nodes), but I don't now currently (thanks to the wonderful easily to understand extjs) where to start that.

manager/assets/modext/widgets/core/tree/modx.tree.js
wuuti: Using a description instead of tooltip+listeners does not show up anything. Any idea?

wuuti:
@Mark-H Changes were made btw

gpsietzema:
Looks like this can be merged now?

guyinpv:
I have a couple issues.

On smaller screens, the popup comes up out of view. It opens to the right of the icon and is off screen.

The icon is positioned absolutely which, I would think, could interfere with any additional icons that might be placed here in the future. Flex positioning or even floats would probably be better than absolute in this case.

The index.css file is changed but it looks to be only a whole bunch of prefix directives are added. This might have been done accidentally by the author's code editor or something, are all those prefixes necessary to dump into the css file for this PR?

Two classes called "undefined" are added in the DOM. I'm not sure if this is normal or an ExtJS thing or what. SS:
undefined classes

As a feature request, it would be consistent to have this also on elements. To be able to add a new chunk or snippet to a category or as a sister element. Maybe that can be added later, or for completeness, do this on elements too within this PR.

Cheers

gpsietzema:
@wuuti Any chance you find some time on short term to fix @Mark-H 's suggestions?

I did some testing on safari/chrome/firefox on both small/normal/large resolutions and found no issues. The only thing which can be annoying is large titles and an undefined title in the hover:


screen shot 2018-05-01 at 11 16 38

alroniks:
It is a good approach but I suggest moving this feature to the 3.x branch with a new design because merged now it will add a headache for integrators when they will do resolving conflicts and fixing issues with design in the future.
@opengeek @gpsietzema

SnowCreative:
Regarding automatic opening of the folder when creating new resources, I'm against it. All that automatic opening and closing drives me crazy. If I open a folder, I want it to stay open until I close it, and if I close a folder I want it to stay closed until I open it.

Mark-H (approved):
While I appreciate the sentiment of not making the 3.x merge harder than it should be - this looks like a fine improvement for 2.x.

Jako:
@guyinpv Those undefined classes seems to be caused by ExtJS after rendering the node button. It would be quite time consuming to get the code that adds those additional classes.

@gpsietzema The undefined text in the popup is caused by a not refreshed lexicon cache.

Otherwise the PR is working fine. There could be an issue with Collections/Selections, since it shows the Button there too. But that part has to be solved in Collections by disabling the 'pnew_modDocument' class for that node.

Jako (approved):

alroniks (approved):

Issue #3772, opened 2011-02-08
Form customization for BOTH update and create


ncrossland:
ncrossland created Redmine issue ID 3772

It would be useful if there was the ability to create a rule for BOTH the update and creation of resources - it's very rare I'd want to do something different, and would remove a lot of duplication of effort in setting up every rule twice.

neoneddy:
+1 on this.

gpsietzema:
+1 from me as-well.

There's an important thing though: can there be situations where you want different rules for create and update? I can't think of any right now. But if there are, this isn't a good idea.

Any ideas?

pixelive:
+1

OptimusCrime:
Please take a look at answer given earlier:

>splittingred submitted:
>While nice in theory, this does not work pragmatically - constraints are impossible to apply without separated create/update rules.
>If I set a constraint of 'template' => '4', one would assume that means don't do FC rules except on Resources with Template of 4.
>However, on a 'Create Resource' page, no new resource has a template of 4 (unless done via a default_template setting or another FC rule). So, currently, FC gets around this by having Constraints apply to the parent of the newly made Resource. This could not be done without separation of update/create.

I think is good enough reason to close the request, unless someone with good knowledge in the implemenation can tell if doing a workaround could be possible without redoing too much of the code. I have never looked at how FC is implemented "under the hood", so I can't tell.

gpsietzema:
That's not a reason not to do it. If a rule has a constraint for template==4 and you're creating a resource where template==4 doesn't apply, it just doesn't work. That's that.

There is a possibility to build in something like template inheritance. We use a plugin for this: when creating new resources it checks a system setting for template inheritance, based on the parent template.

This could be something for a new MODX release I guess. We'll how we can implement this.

isaacniebeling:
I'm with @gpsietzema on this one... If it doesn't apply, it doesn't apply, but keeping update and create rules in sync has had me tearing out my hair on more than one occasion. As long as FC is (one of) the last bits of JS to fire on resource load, it should be able to cover just about any constraint the same on Create as on Update.

rainbowtiger:
[...]

PR #13775, opened 2018-02-11
Add wildcard support to form customization actions (i.e., support resource/create and resource/update in single set)

Mark-H:
### What does it do?
Changes form customization to support actions like `resource/*`, which will then apply to any controller that starts with resource/.  Primarily this will be used by people to manage FC rules that apply to create and update, instead of just one of those.

FC was architected to be useful for a lot more than just resources, so this method does not hardcode the resource bits anywhere, instead opting for a bit more flexible wildcard approach. 

When editing a wildcard action, FC will look at the update action to determine the default fields/tabs/etc. This could have also been the create, but it is assumed that update actions may have more fields than create in certain (hypothetical) situations, so update seemed like the better idea. 

### Why is it needed?
To make it easier to manage rules that affect both create and update.

### Related issue(s)/PR(s)
#3772


rainbowtiger:
In 2.6.2, update.class.php is very different, and in a different location.

Issue #12625, opened 2015-09-02
File-type TV and prepended ./ to path


sottwell:
Revo 2.4.0

When selecting a file from the media browser when using a file-type TV it will prepend a ./ to the path. This can be prevented by first clicking on the Source label in the left tree view in the browser, and corrected by removing the leading ./ from the resulting path in the TV field. That's kind of hard to explain to the users, though.

sottwell:
Dupe of #11830?

Bruno17:
seems this could be fixed by adding one line 
```
$dir = str_replace('./','/',$dir);
```
to core/model/modx/processors/browser/directory/getfiles.class.php

```
    public function process() {
        if (!$this->getSource()) {
            return $this->failure($this->modx->lexicon('permission_denied'));
        }
        $allowedFileTypes = $this->getProperty('allowedFileTypes');
        if (empty($allowedFileTypes)) {
            // Prevent overriding media source configuration
            unset($this->properties['allowedFileTypes']);
        }
        $this->source->setRequestProperties($this->getProperties());
        $this->source->initialize();
        if (!$this->source->checkPolicy('list')) {
            return $this->failure($this->modx->lexicon('permission_denied'));
        }

        $dir = $this->getProperty('dir');
        $dir = preg_replace('/[\.]{2,}/', '', htmlspecialchars($dir));
        $dir = str_replace('./','/',$dir);//<-this line seems to fix this issue
        if ($dir === 'root') {
            $dir = '';
        }
        $list = $this->source->getObjectsInContainer($dir);
        return $this->outputArray($list);
    }
```

Ibochkarev:
This has been fixed earlier. You can close.
MODX 2.7.1

![2019-02-17_15-38-22](https://user-images.githubusercontent.com/2138260/52911154-3b3ac280-32ca-11e9-9f70-ea52c75f3097.png)

Issue #11830, opened 2014-07-27
Media Source ./ to image path


sottwell:
Using a simple Media Source with path and URL set to relative 'assets/images/'

When first selecting an image, the path is fine - image1.jpg and until the resource is saved, the thumbnail is also fine. But when the image is saved, it loses the thumbnail.

If you change the image, it adds ./ to the path - ./image2.jpg

phpthumb is failing to find the image:

[2014-07-27 00:39:09](ERROR @ /connectors/system/phpthumb.php) phpThumb was unable to generate a thumbnail for: /paas/c1327/www/core/cache/phpthumb/jonline.sottwell.modxcloud.com__src9ca6fcfedb26085c674b36d08d11264b_par5df3dad6621628fd102555d4b8000f6e_dat0.jpeg
[2014-07-27 00:39:09](ERROR @ /connectors/system/phpthumb.php) Error outputting thumbnail:
OutputThumbnail() failed because !is_resource($this->gdimg_output) in file "phpthumb.class.php" on line 572

JayCarney:
I've noticed an odd behaviour, I'm having this issue on 2.4.3pl.
Selecting an image from the root of the media source does prepend `./`, but If I change to a folder within the media source, select an image, change back to the root and select the image the `./` isn't prepended. Not sure if this information helps

Fi1osof:
+1

Bruno17:
I think this is related to https://github.com/modxcms/revolution/issues/12625
where I had posted a possible fix. Would this fix it for you, @Fi1osof ?

Fi1osof:
@Bruno17 it shoud not be $dir = str_replace('./','/',$dir);
Need $dir = str_replace('./','',$dir);

Bruno17:
what happens with `$dir = str_replace('./','/',$dir);`
did you try it?

sottwell:
I would presume it would replace the “./” with “/” which is not at all what is wanted. At least the "./" indicates "this directory", and not "root directory".

Fi1osof:
Yes, this should not be the root directory. It will be concatenated with media-source root dir.

sottwell:
If you click on the folder in the left tree before selecting the file, it doesn't prepend the "./" to the file path. I never was able to find where that prepending took place.

Ibochkarev:
Thanks for your issue. This was fixed in #13778

PR #13778, opened 2018-02-14
Prevent adding ./ to filepath when in root of a mediasource [12625]

Bruno17:
### What does it do?
Only returns the dirname in getOpenTo, when not in root, otherwise returns an empty string

### Why is it needed?
Solves the issue with prefixed ./ in image or file - TVs, when in Root.
Doesn't happen the first time, but as soon as the Manager - Page is reloaded and a image or file-TV is saved again. Then the openTo was ./ and this was prefixed to the filepath

### Related issue(s)/PR(s)
Related to
https://github.com/modxcms/revolution/issues/12625
and
https://github.com/modxcms/revolution/issues/11830


bezumkin (approved):
Yes, it is already was fixed in new media sources for 3.x branch

PR #13781, opened 2018-02-16
[3.x] Load PhpMailer via Composer instead of including the source files

bezumkin:
### What does it do?
Instead of including the source files in the MODX code, we load the dependecy using Composer.

Tested by installing QuickEmail extra and sending messages with and without errors. Also tested managers "forgot login" function. Everything works fine. 

### Why is it needed?
To couple the code from other dependencies.

### Related issue(s)/PR(s)
[MAB-01](https://github.com/modxcms/mab-recommendations/blob/master/01-replace-3rd-party-libraries-with-composer-dependencies.md)


OptimusCrime:
Awesome! Nice work :)

PR #13783, opened 2018-02-17
"Forgot password" for new manager login

bezumkin:
### What does it do?
- Removed system settings "manager_language" and "manager_lang_attribute"
- System settings "emailsubject" and "forgot_login_email" was moved to lexicons.
- Language of the manager is detected automatically and saved to session of user.
- User can change preferred language on login screen and it will work for "reset password emails" and the whole manager after login
- No more password in reset emails. User should confirm reset by link and specify new password manually.
- Removed ability to generate password in manager and send it to user with email.
- Added Russian and Netherlands lexicons for new functions.

### Why is it needed?
To prevent sending passwords in email

### Related issue(s)/PR(s)
N/A


bezumkin:
Closed because we need to add something more from MODX3 plan.

I will send another PR with all improvements later.

PR #13784, opened 2018-02-20
Ability to create MODX project with composer

bezumkin:
### What does it do?
Add composer.json file to default branch so we could submit this repository to Packagist.

### Why is it needed?
With this file we will be able to create new projects pretty simple:
```
composer create-project modx/revolution new_site --stability=dev
```

### Related issue(s)/PR(s)
#13620, #13619

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,67 @@
+{
+  "name": "modx/revolution",
+  "description": "MODX Revolution is a Content Management System",
+  "type": "project",
+  "keywords": [
+    "content",
+    "CMS",
+    "routing",
+    "templating"
+  ],
+  "homepage": "http://modx.com/",
+  "license": "GPL-2.0+",
+  "authors": [
+    {
+      "name": "Jason Coward",
+      "email": "jason@opengeek.com",
+      "homepage": "http://jasoncoward.com/",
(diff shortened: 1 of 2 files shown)


Mark-H:
The discussion in #13619 was mostly around answering why MODX should become composer installable; the proposed solution in #13620 failed to really improve the way you can download/install MODX. 

This pull request, however, makes it possible to actually get to the setup in one command, so I'm more positive about this one :)

One thing to note is that it targets 2.x as that is the default branch for packagist to use. It would need to be pushed up to packagist by probably Jason or John.

composer.json
```diff
@@ -0,0 +1,67 @@
...
+    "irc": "irc://irc.freenode.org/modx",
+    "issues": "https://github.com/modxcms/revolution/issues/",
+    "source": "https://github.com/modxcms/revolution/"
+  },
+  "config": {
+    "vendor-dir": "core/vendor"
+  },
+  "require": {
+    "php": ">=5.6",
+    "xpdo/xpdo": "^3.0@dev",
```
Mark-H: As the PR targets 2.x, it should not include the 3.x dependencies (in both the require and autoload section)

PR #13785, opened 2018-02-20
Also remove second nocompress checkbox from the setup (re #12857)

Mark-H:
### What does it do?
After merging #12857, I noticed there was another checkbox for it left. 

### Why is it needed?
Nocompress is no longer needed since the manager no longer uses dynamic compression (v2.5). This option should be removed from 3.0.

### Related issue(s)/PR(s)
#12857 and further issues mentioned there.

### DIFF
--- a/setup/templates/options.tpl
+++ b/setup/templates/options.tpl
@@ -97,17 +97,6 @@
 {/if}
 <table class="options">
 <tbody>
-{if $installmode NE 0}
-<tr>
-    <th>
-        <label>
-            <input type="checkbox" name="nocompress" id="nocompress" value="1" />
-            {$_lang.options_nocompress}
-        </label>
-    </th>
-    <td>{$_lang.options_nocompress_note}</td>
-</tr>
-{/if}
 <tr>
     <th style="padding-top: 2em;">
         <label>


alroniks (approved):

Jako (approved):

gpsietzema:
Works perfectly.

Before:
screen shot 2018-04-04 at 14 55 07


After:
screen shot 2018-04-04 at 14 55 25



Shouldn't these system settings be removed as-well though?

screen shot 2018-04-04 at 14 57 03

Mark-H:
The `compress_css` and `compress_js` settings still have a function. You can disable those to use the uncompressed files (which is useful in development). The other two could also be removed, yes, so a separate PR for that would be useful.

alroniks:
`manager_js_zlib_output_compression` used in `manager/min/index.php`. Does manager/min still used in the system?

Mark-H:
`manager/min/` was removed in #12778 - looks like @opengeek did a 2.x->3.x merge recently that reintroduced it...

alroniks:
Created the issue about removing mentioned system settings https://github.com/modxcms/revolution/issues/13858

PR #13786, opened 2018-02-20
Forgot password for new login screen

bezumkin:
### What does it do?

- Removed system settings "manager_language" and "manager_lang_attribute".
- System settings "emailsubject" and "forgot_login_email" was moved to lexicons.
- Language of the manager is detected automatically and saved to session of user.
- User can change preferred language on login screen and it will work for "reset password emails" and the whole manager after login
- No more password in reset emails. User should confirm reset by link and specify new password manually.
- Removed ability to change password in manager and send it to user by email.
- Added Russian and Netherlands lexicons for new functions.
- Added pelago/emogrifier to improve HTML emails
- New default HTML template for manager emails
- Ability to notify user about creating an account for him
- Remove sending password from user create/update processors
- Added support of backgrounds to login screen

### Why is it needed?
To provide proper "forget password" workflow
and HTML emails
and other good things

### Related issue(s)/PR(s)
N/a


Mark-H:
We'll need to make sure that the old (no longer used) settings are removed on upgrade, or that those setting values aren't used. Right now it's falling back to older versions of the email on an upgrade, which prevents the functionality from being used.

````emailsubject
forgot_login_email
manager_lang_attribute
manager_language
websignupemail_message
````

bezumkin:
screen shot 2018-03-16 at 13 23 02
screen shot 2018-03-16 at 13 23 08
screen shot 2018-03-16 at 13 23 26

Mark-H:
🎉

PR #13789, opened 2018-02-22
Delete a few no-longer-needed images

### LINKED ISSUE #13788: default favicon.ico can overwrite custom one with git/composer installs
### Summary
The git installation includes an ancient favicon.ico in web root that could overwrite an existing 
custom one.

### Step to reproduce
Install via git or composer into a directory with a custom favicon present.

### Observed behavior
image

### Expected behavior
image … or whatever custom favicon.ico file that actually existed previously.

The default content can have a basic favicon and assets/ and the Manager could get a more robust set in the manager path as well. Suggested graphics found in the [favicons.zip](https://github.com/modxcms/revolution/files/1746140/favicons.zip) archive.

### Environment
~~A. MODX installs from Downloads on MODX.com
B. MODX installs from third-party installers like Bitnami~~
C. Hipster-compliant installs via git/composer
~~D. All of the above
E. None of the above~~

rthrash:
### What does it do?

Removes no-longer-needed images.

### Why is it needed?

The favicon cannot be changed in a git installation if it is in the repository as discussed in #13788. The credits/*.gif files are no longer needed as they’re not referenced anywhere in the code base.

### Related issue(s)/PR(s)

Resolves #13788


Mark-H:
I've edited your PR description to follow the standard template.

PR #13790, opened 2018-02-22
CLI install script to use after composer create-project

bezumkin:
### What does it do?
Add ability to install MODX from CLI with interactive dialog, instead of edit `config.xml`. If there is already `config.xml` exists, script will suggest to load values from it.

Also you can always override any value by CLI arguments to file. This script will generate `config.xml` and start default MODX CLI install.

#### Examples:
default run with interactive questions
```
php www/setup/cli-install.php
```
specify some values with arguments
```
php www/setup/cli-install.php --database=s12973 --database_user=s12973
```
specify mode
```
php www/setup/cli-install.php --mode=upgrade --language=ru --remove_setup_directory=0
```

### Why is it needed?
It allows to install MODX in interactive mode without opening web browser.
Also you can make installation fully automatically along with `composer`. For example:
```
composer create-project modx/revolution ./www 2.x-dev --keep-vcs &&
rm -rf ./www/.git && 
php ./www/setup/cli-install.php --database_server=127.0.0.1 --database=s12973 --database_user=s12973 --database_password=o0D9VRNdulnJ --table_prefix=modx_ --http_host=s12973.h2.modhost.pro --language=en --cmsadmin=admin --cmspassword=adminpass --cmsadminemail=admin@s12973.h2.modhost.pro --core_path=/home/s12973/www/core/ --context_mgr_path=/home/s12973/www/manager/ --context_mgr_url=/manager/ --context_connectors_path=/home/s12973/www/connectors/ --context_connectors_url=/connectors/ --context_web_path=/home/s12973/www/
```

### Related issue(s)/PR(s)
#13784

### DIFF
--- a/setup/cli-install.php
+++ b/setup/cli-install.php
@@ -1,0 +1,172 @@
+<?php
+
+/**
+ * CLI install script for MODX
+ *
+ * Run and answer the questions. Also you can pre-define values:
+ * php www/setup/install.php --database=my_db --database_user=db_user_12 --language=ru
+ */
+
+if (PHP_SAPI != 'cli') {
+    exit('This file should be run through the command line interface!');
+}
+$path = dirname(__FILE__) . '/';
+$config = $path . 'config.xml';
+$mode = 'new';
+
+$variables = [
+    'database_type' => 'mysql',
+    'database_server' => [
+        'prompt' => 'Database server',
+        'default' => '127.0.0.1',
+    ],
+    'database' => [
+        'prompt' => 'Database',
+        'default' => '',
+    ],
+    'database_user' => [
+        'prompt' => 'Database user',
+        'default' => '',
+    ],
+    'database_password' => [
+        'prompt' => 'Database password',
+        'default' => '',
+    ],
+    'database_connection_charset' => 'utf8',
+    'database_charset' => 'utf8',
+    'database_collation' => 'utf8_general_ci',
+    'table_prefix' => [
+        'prompt' => 'Table prefix',
+        'default' => 'modx_',
+    ],
+    'https_port' => 443,
+    'http_host' => [
+        'prompt' => 'Http host',
+        'default' => '',
+    ],
+    'cache_disabled' => 0,
+    'inplace' => 0,
+    'unpacked' => 0,
+    'language' => [
+        'prompt' => 'Language',
+        'default' => 'en',
+    ],
+    'cmsadmin' => [
+        'prompt' => 'Admin login',
+        'default' => 'admin',
+    ],
+    'cmspassword' => [
+        'prompt' => 'Admin password',
+        'default' => '',
+    ],
+    'cmsadminemail' => [
+        'prompt' => 'Admin email',
+        'default' => '',
+    ],
+    'core_path' => [
+        'prompt' => 'Path to core',
+        'default' => dirname($path) . '/core/',
+    ],
+    'context_mgr_path' => [
+        'prompt' => 'Path to manager',
+        'default' => dirname($path) . '/manager/',
+    ],
+    'context_mgr_url' => [
+        'prompt' => 'Manager url',
+        'default' => '/manager/',
+    ],
+    'context_connectors_path' => [
+        'prompt' => 'Path to connectors',
+        'default' => dirname($path) . '/connectors/',
+    ],
+    'context_connectors_url' => [
+        'prompt' => 'Url to connectors',
+        'default' => '/connectors/',
+    ],
+    'context_web_path' => dirname($path) . '/',
+    'context_web_url' => '/',
+    'remove_setup_directory' => 1,
+];
+
+// Parse CLI arguments
+$args = array_slice($argv, 1);
+$cli_variables = [];
+foreach ($args as $arg) {
+    $tmp = array_map('trim', explode('=', $arg));
+    if (count($tmp) === 2) {
+        $k = ltrim($tmp[0], '-');
+        if (isset($variables[$k])) {
+            $cli_variables[$k] = $tmp[1];
+        } elseif ($k == 'mode' && in_array($tmp[1], ['new', 'upgrade', 'advanced'])) {
+            $mode = $tmp[1];
+        }
+    }
+}
+
+// If it is upgrade - parse old variables
+$old_variables = [];
+if (file_exists($config)) {
+    $use = ($mode == 'upgrade')
+        ? 'Y'
+        : readline("There is an old config found at {$config}. Do you want to load this values? (Y): ");
+    if (empty($use) || strtolower($use) == 'y') {
+        $tmp = simplexml_load_file($config);
+        $tmp = @json_decode(json_encode($tmp), true);
+        if (is_array($tmp)) {
+            $old_variables = $tmp;
+        }
+    }
+}
+
+// Merge everything
+$variables = array_merge($variables, $old_variables, $cli_variables);
+
+// Process config
+$data = [];
+foreach ($variables as $key => $params) {
+    if (!is_array($params)) {
+        $data[$key] = $params;
+    } elseif (!empty($params['prompt'])) {
+        $prompt = $params['prompt'];
+        $res = '';
+        if (!empty($params['default'])) {
+            $prompt .= " ({$params['default']})";
+            if (!$res = readline($prompt . ': ')) {
+                $res = $params['default'];
+            }
+        } else {
+            while (!$res = readline($prompt . ': ')) {
+                echo $prompt . " is required!\n";
+            }
+            readline_add_history(trim($res));
+        }
+        $data[$key] = trim($res);
+    }
+    if (strpos($key, '_path') !== false || strpos($key, '_url') !== false) {
+        $data[$key] = preg_replace('#/+#', '/', ('/' . trim($data[$key], '/') . '/'));
+    }
+}
+
+// Check directories
+$dirs = [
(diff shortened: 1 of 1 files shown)


alroniks:
Does it make sense to add some library for handling CLI commands and params for better UX? I mean create clean and readable usage, --help command with a description of all params etc.

bezumkin:
My idea to run this script as interactive "question -> answer" format, ability to override values from arguments is just a bonus.

And I really don't want to bring here another dependency only for such simple script. Let's listen to another opinions.

Mark-H:
[xPDO 3 uses Symfony/Console](https://github.com/modxcms/xpdo/tree/3.x/src/xPDO/Console), so in 3.x we could potentially use that without introducing an extra dependency. For 2.x, which does not use composer for dependencies at all yet, I think the approach in this PR is better.

jgulledge19:
This looks cool! I have been working a lot on Blend: https://github.com/LippertComponents/Blend/tree/dev which I changed over to Symfony/Console last week on the dev branch. I am looking to do a PR to completely replace the setup/upgrade process in MODX to use migrations for 3.0. To allow CLI and the traditional web installation approach. The idea with migrations is that every install/update has both and up method and an equal down method to handle everything for the DB changes. Also with this approach every PR for MODX that has any DB effects would need a related migration file with valid up and down methods. I believe it will ease complex PRs and new release builds. If anyone would like to help that would be appreciated!

setup/cli-install.php
```diff
@@ -0,0 +1,172 @@
...
+}
+$path = dirname(__FILE__) . '/';
+$config = $path . 'config.xml';
+$mode = 'new';
+
+$variables = [
+    'database_type' => 'mysql',
+    'database_server' => [
+        'prompt' => 'Database server',
+        'default' => '127.0.0.1',
```
Mark-H: Wouldn't `localhost` be a more appropriate default here? I seem to recall the regular setup defaults to that.

setup/cli-install.php
```diff
@@ -0,0 +1,172 @@
...
+// Check directories
+$dirs = [
+    'core_path' => '/core/',
+    'context_mgr_path' => '/manager/',
+    'context_connectors_path' => '/connectors/',
+];
+foreach ($dirs as $key => $value) {
+    $target = dirname($path) . $value;
+    if (file_exists($target) && $data[$key] != $target) {
+        shell_exec("rm -rf {$data[$key]}");
```
Mark-H: Could there be a case where someone (accidentally) manages to put value `/` into `$data[$key]`? Maybe if `$value` starts with a space? 

`escapeshellarg` sounds like a good idea for any values that are inserted into `shell_exec` (this line, line 160, line 172)

setup/cli-install.php
```diff
@@ -0,0 +1,172 @@
...
+    ],
+    'cache_disabled' => 0,
+    'inplace' => 0,
+    'unpacked' => 0,
+    'language' => [
+        'prompt' => 'Language',
+        'default' => 'en',
+    ],
+    'cmsadmin' => [
+        'prompt' => 'Admin login',
```
Mark-H: Might be nice if we could use the lexicons for this to make sure that the fields are labeled the same in the UI and the CLI helper. For example, this one is "Administrator username" https://github.com/modxcms/revolution/blob/2.x/setup/lang/en/default.inc.php#L41

A simple implementation could just include the lexicon file at the top, and use `$_lang['foo_bar']` in the prompt definitions.

Bonus points for a `--language=nl` on the cli to also ask for values in the localised language (looks like that would already affect the installed language), but I'd consider that a nice to have, not required for this to be accepted.

setup/cli-install.php
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+}
+
+// If it is upgrade - parse old variables
+$old_variables = [];
+if (file_exists($config)) {
+    $use = ($mode == 'upgrade')
+        ? 'Y'
+        : readline("There is an old config found at {$config}. Do you want to load this values? (Y): ");
```
Mark-H: This is nitpicking, but I'd suggest `existing config file` instead of `old config` - old implies it's out of date. 

It looks like this will grab only from a config.xml in the root - not from a config.inc.php if it exists. Not sure if it has to, but what do you think about that for upgrades? Or is this not meant to be used with upgrades?

setup/cli-install.php
bezumkin: localhost points to `/var/run/mysql.sock`, but in chroot environment on hosting it inaccessible. But `127.0.0.1` points to tcp port, that opened in MySQL by default and works in chroot.

So, that is why it is by default in my script.

setup/cli-install.php
bezumkin: It is not for upgrade, it for continue current install after abort. And it would get file from /setup/ directory

To upgrade you need specify `--mode=upgrade` and if there will be no `config.xml`, just answer again to questions. Maybe we can load existing config, really.

setup/cli-install.php
bezumkin: Good point, here we can shoot in the leg.

setup/cli-install.php
bezumkin: Interesting idea

setup/cli-install.php
AmaZili: Do'nt know if it is your case, but we are using servers with plesk having multiple php version, and need to define the path to php command as well, depending on the version used by the modx installation
Current example of command line installation /upgrade  :  /opt/plesk/php/7.1/bin/php /var/www/vhosts/domain.tld/httpdocs/setup/index.php --installmode=upgrade-advanced --core_path=/var/www/vhosts/domain.tld/core/ --config=domain.tld.config.xml;

bezumkin: You mean, that if just type `php` in console it will answer `command not found`? There are always should be default `php`.

Please, check it in your environment.

Jako: At least one big german shared hoster that has php 4.4.3 installed as default `php` in the console. For whatever reason.

Mark-H: That's not something this PR has to address. There are different ways to make the desired binary available as `php`.

OptimusCrime: It should be safe to assume the `php` in `$PATH` is a correct PHP version that is above the minimal requirement. This PR should not have to deal with any such edge cases.

AmaZili: Vasilly, There is a php, but it is a php5.6, old stuff, I prefer to install all websites with the most performance wyse version of it...

AmaZili: @OC, not sure the PATH would be something I will consider as granted... In that case, it is not valid in ssh command line, gives you php old version

OptimusCrime: Most applications assume this. I have never seen any command line tool that goes into great detail to explain `PATH` and covering weird edge cases. That is up to the developer to fix the environment. See [composer](https://getcomposer.org/doc/00-intro.md#locally) for instance, their installation instructions are just

```
php composer-setup.php --install-dir=bin --filename=composer 
```

That YOUR `php` executable gives you incorrect version is not our fault, it is your configuration that is incorrect. On my server `php` gives me the latest version I have installed.

So yeah, in the world of command line tools, you would assume that the `PATH` and related settings/configurations are correct.

Additionally, there is no way to cover _all_ the edge cases in the world for this scenario.

bezumkin: Solved the problem differently - index.php is now simply included inside the script

AmaZili:
May I add, that we usually change the name and path to the assets directory in our config xml file, I heard others are doing so as well, so it may be a good idea to have the same kind of treatment as connectors.

bezumkin:
@AmaZili And you can do it with script - it will ask you about your paths.

setup/cli-install.php
AmaZili: Optimus, This is not our php version, but as Jakko said, there are lost of sutiation were php with the valid version is not active. Plesk users and there are thousands will tell you it's the case.
Regarding composer, we can do /opt/plesk/php/7.1/bin/php composer-setup.php --install-dir=bin --filename=composer and its done because from what I know, it does not call another php command inside .
The issue arise when the php command is called inside a command file and thus must be configurable in some way for such users.

bezumkin: @AmaZili problem is already solved https://github.com/modxcms/revolution/pull/13790/commits/12db2266379a01e7fc46e096e10d14337b80ea55#diff-bdb40b842e5d2c02e8822780192e0d72R183

AmaZili:
@bezumkin that's fine, thanks
Other may ask about it later :-)

setup/cli-install.php
```diff
@@ -0,0 +1,183 @@
...
+    'context_mgr_path' => '/manager/',
+    'context_connectors_path' => '/connectors/',
+];
+foreach ($dirs as $key => $value) {
+    $target = dirname($path) . $value;
+    if (file_exists($target) && $data[$key] != $target && strpos($data[$key], dirname($path)) === 0) {
+        shell_exec('rm -rf ' . escapeshellarg($data[$key]));
+        shell_exec('mv ' . escapeshellarg($target) . ' ' . escapeshellarg($data[$key]));
+    }
+}
```
opengeek: I'm not sure this is the best idea and is not compatible across all supported platforms. Would it not be better to use native PHP features to handle this?

bezumkin: @opengeek You are right, I`m totally forgot about Windows.

Will try to rewrite it tomorrow.

Issue #13792, opened 2018-02-23
500 error on dashboard due to issue in getonline.class.php (3.x)


gpsietzema:
### Summary
Issue in the who's online widget on 3.x

### Step to reproduce
Install 3.x from Github, go to the dashboard. Watch your console and PHP error log.

### Observed behavior
Console shows a 500 error 

Error log shows this:
```[error] 3027#3027: *410 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught TypeError: Argument 1 passed to modUserWhoIsOnlineProcessor::prepareQueryBeforeCount() must be an instance of xPDOQuery, instance of xPDO\Om\mysql\xPDOQuery given, called in /var/www/vhosts/modx3/core/model/modx/modprocessor.class.php on line 556 and defined in /var/www/vhosts/modx3/core/model/modx/processors/security/user/getonline.class.php:13
Stack trace:
#0 /var/www/vhosts/modx3/core/model/modx/modprocessor.class.php(556): modUserWhoIsOnlineProcessor->prepareQueryBeforeCount(Object(xPDO\Om\mysql\xPDOQuery))
#1 /var/www/vhosts/modx3/core/model/modx/processors/security/user/getonline.class.php(36): modObjectGetListProcessor->getData()
#2 /var/www/vhosts/modx3/core/model/modx/modprocessor.class.php(176): modUserWhoIsOnlineProcessor->process()
#3 /var/www/vhosts/modx3/core/model/modx/modx.class.php(1736): modProcessor->run()
#4 /var/www/vhosts/modx3/core/model/modx/modconnectorresponse.class.php(140): modX->runProcessor('security/user/g...', Array, Array)
#5 /var/ww" while reading response header from upstream, client: 10.0.2.2, server: modx3.gauke, request: "POST /connectors/index.php HTTP/1.1", upstream: "fastcgi://unix:/var/run/php/php7.1-fpm.sock:", host: "modx3.gauke", referrer: "http://modx3.gauke/manager/?"
```

### Expected behavior
No 500 error :)

### Environment
MODX 3.x from Github
NGINX, PHP-FPM 7.1, MariaDB (mysql  Ver 15.1 Distrib 10.1.20-MariaDB), chrome 64

bezumkin:
@modxbot close

PR #13793, opened 2018-02-23
Fixed fatal error in "who is online" processor

bezumkin:
### What does it do?
Fixes  "who is online" processor

### Why is it needed?
To fix fatal error in processor

### Related issue(s)/PR(s)
#13792


gpsietzema:
Just tested it. Works like a charm 👍

PR #13594, opened 2017-09-04
Better handling of reverse HTTPS proxies

OptimusCrime:
### Note

[Better diff here](https://github.com/modxcms/revolution/pull/13594/files?w=1).

### What does it do?
Does a better job at attempting to identify HTTPS requests before MODX decides if it should serve over HTTP or HTTPS.

### Why is it needed?
As reported in the original issue #13259, the current checks have problems with handling reverse proxies for some providers (such as Cloudflare).

### Details
Note that I took great care negating the original expression:

```
defined('PHP_SAPI') && (PHP_SAPI == "cli" || PHP_SAPI == "embed")
```

Into:

```
!defined('PHP_SAPI') || (PHP_SAPI != "cli" && PHP_SAPI != "embed")
```

I have actually verified that these are opposites of each other using a [online truth table generator](http://web.stanford.edu/class/cs103/tools/truth-table-tool/).

The first expression would be `a && (b || c)` while the latter is `~a || (~b && ~c)`. These are opposites of each other.

_Also note that I have been unable to test this as I have no providers that set these server variables/serve requests this way._

### Related issue(s)/PR(s)
#13259

PR #13794, opened 2018-02-24
Make less strict https status check

kudashevs:
### What does it do?
It make more flexible check if the script was requested through the HTTPS protocol or not. Due to [PHP documentation](http://php.net/manual/en/reserved.variables.server.php) the $_SERVER['HTTPS'] value can be set to non empty value, not only 'on' value.

### Why is it needed?
I cannot get HTTPS connection on one of the client projects with MODX Revolution 2.6 on-board. When I investigating this problem, i found that hosting's web server set $_SERVER['HTTPS'] variable to 1 and MODX expected 'on' value. That's why HTTPS connection didn't worked.

### Related issue(s)/PR(s)
Not related.

### DIFF
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -49,7 +49,7 @@ if (!defined('MODX_BASE_PATH')) {
 if(defined('PHP_SAPI') && (PHP_SAPI == "cli" || PHP_SAPI == "embed")) {
     $isSecureRequest = false;
 } else {
-    $isSecureRequest = ((isset ($_SERVER['HTTPS']) && strtolower($_SERVER['HTTPS']) == 'on') || $_SERVER['SERVER_PORT'] == $https_port);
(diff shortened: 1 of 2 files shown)


OptimusCrime:
You might want to check out my PR here: https://github.com/modxcms/revolution/pull/13594

kudashevs:
@OptimusCrime thanks for information, but my problem not related with PR #13594.

In my case provider's web-server set $_SERVER['HTTPS'] value to 1. When i asked they to change this value to 'on' (as MODX expects, let's see the changed code: strtolower($_SERVER['HTTPS']) == 'on')) technical support answered me, that this variable can get any value, not only 'on'. I went to CGI specification, but value of 'HTTPS' is not described there, then i went to [PHP documentation](http://php.net/manual/en/reserved.variables.server.php) and found that:

'HTTPS' Set to a non-empty value if the script was queried through the HTTPS protocol. 
Note: Note that when using ISAPI with IIS, the value will be off if the request was not made through the HTTPS protocol.

I think it is reasonable to remade the HTTPS check to get all values, as it is described in documentation, not only 'on' as it works right now.

Issue #13795, opened 2018-02-25
Message system sends group messages to wrong or nonexistent users


BobRay:
### Summary
If you select the 'usergroup' option in the Manager's message System, the messages are sent to the wrong recipients. 

### Step to reproduce
Send a test message to a user group.

### Observed behavior
Some (or all) messages don't arrive. Some messages get sent to non-group members or members that don't exist.

### Issue

The problem is that the security/message/create processor uses the ID of the record in the modx_member_groups table as the user ID instead of the 'member' field.

### The fix (tested)

It's a simple change to one line of code (line 135 in my version) in the processor:

core/model/modx/processors/security/message/create.class.php 

    `public function getRecipients($users, $primaryKey = 'id') {
            /** @var xPDOSimpleObject $user */
            $recipients = array();
            foreach ($users as $user) {
               if ($user->get($primaryKey) == $this->modx->user->get('id')) {
                      continue;
                }
               -  $recipients[] = $user->get('id');
               + $recipients[] = $user->get($primaryKey);
            }
            return $recipients;
     }
`

The bug is sometimes masked by the fact that the ID of the member_group_,member record may be the same as the user ID.

Let me know if you want a PR.

OptimusCrime:
We always want PRs, otherwise someone else will need to redo the change you just did :)

However, I might add that we are currently debating if we should remove the messages as a part of the core in the upcoming 3.x release and instead make it a separate extra. Reason being that it is a complex system and not very much used. Might be worth considering here. Ping @Alroniks

alroniks:
Fixed in this PR https://github.com/modxcms/revolution/pull/13390 and works as expected. Closing.

PR #13796, opened 2018-02-26
Fix messages being sent to wrong or nonexistent users when sending to a user group

BobRay:
Fixes bug with Message Create processor using the wrong recipient ID when processing usergroup messages.

### What does it do?
Modified the method to get the 'member' field of the modUserGroupMember object rather than the 'id' field.

### Why is it needed?
Messages to user groups are sent to wrong or non-existent users.

### Related issue(s)/PR(s)
Issue #13795

### DIFF
--- a/core/model/modx/processors/security/message/create.class.php
+++ b/core/model/modx/processors/security/message/create.class.php
@@ -131,7 +131,7 @@ class modMessageCreateProcessor extends modObjectProcessor {
             if ($user->get($primaryKey) == $this->modx->user->get('id')) {
                 continue;
             }
-            $recipients[] = $user->get('id');
+            $recipients[] = $user->get($primaryKey);
         }
         return $recipients;
     }


Mark-H:
Thanks @BobRay - could you perhaps explain the reasoning behind how this change fixes the reported bug? Not sure if I'm immediately seeing it...

BobRay:
Sure, it's confusing because of the variable names. Both $users and $primaryKey are misleading names. 

    `public function getRecipients($users, $primaryKey = 'id') {
            /** @var xPDOSimpleObject $user */
            $recipients = array();
            foreach ($users as $user) {
               if ($user->get($primaryKey) == $this->modx->user->get('id')) {
                     continue;
               }
               -  $recipients[] = $user->get('id');
               + $recipients[] = $user->get($primaryKey);
            }
           return $recipients;
     }

The purpose of the function is to assemble an array of user IDs to send a message to.

When sending to a usergroup, the $users variable holds, not users, but modUserGroupMember objects. In that case $primaryKey is set to 'member', which is the field that holds the member's user ID. 

Hardcoding 'id' in that line adds the ID of the modUserGroupMember object (not the user) to the $recipients array, so the message is stored in the DB, but the intended recipient can't see it because it has the wrong User ID.

BTW, a sanity check to make sure the recipient exists would be a good idea too.  Otherwise, over time you could amass lots of messages that can't be removed cluttering up the DB.

Mark-H:
Thanks Bob!

Mark-H:
I've marked this for the 2.6.2 milestone as it's a clear bug fix, but as the PR targets 2.x, it will probably need to be cherry picked.

BobRay:
Sorry, I can do another one for the 2.6.x branch if that won't mess things up.

opengeek:
Cherry-picked into 2.6.x branch as commit 66589f76c.

OptimusCrime:
Can this be closed, @opengeek ?

opengeek:
No; it will get marked as merged when this gets pushed to 2.x later today.

OptimusCrime:
Ah, okey, sorry. I thought you merged 2.6.x into 2.x

opengeek:
Or perhaps it won't get marked as merged. Sigh. Anyway, it was merged. Closing.

PR #13798, opened 2018-02-28
Solved a typo in modRest

JoshuaLuckers:
vernose -> verbose


JoshuaLuckers:
I'm not sure if we should add something to add "backwards compatibility" for this typo.

Mark-H:
I don't think we need special BC handling here. The worst that could happen is that people finally get the verbose output that they specifically requested (but didn't work).

Issue #12695, opened 2015-10-03
Plugin properties not saving since v2.4.0


Oliver-Meyer:
When I edit a plugin property and click on the save-button I get the message that the saving was successful, but the little red arrow next to the value stays and the property won't be changed.

Same goes for new properties. You can add a new property and save it - again you'll get the message that the saving was successful, but the new property is not added. If you leave the page and then open it again, the property is missing.

sottwell:
I've been editing and creating plugin properties in 2.4.0 and 2.4.1 with no problems. Is this an existing plugin? Can you save changes to the plugin without doing anything to its properties?

Oliver-Meyer:
I just re-checked it on two fresh installations (both v2.4.1, using PHP 5.5.29). I can edit & save the plugins themselves without problems, but I'm not able to edit / add & save properties - saving doesn't work. No errors in the server logs.

whitebyte:
May this be related to #12580?

Oliver-Meyer:
I don't think so, because I can not even save a property. And I have absolutely no problems with chunks or snippets.

sottwell:
Just installed 2.4.1 fresh on localhost. Can save properties for newly created or installed plugins. Can also do so on several MODX Cloud installations. Both are PHP 5.6-something.

sottwell:
Check your permissions, make sure all your Administrator permissions are checked in the policy.

whitebyte:
@Oliver-Meyer you should alse check you mod_security settings. This is  a known source of such issues.

Oliver-Meyer:
I just logged in into a MODX-Installation with v2.3.5 - I could add / edit plugin properties without any problem. Then I updated this installation to v2.4.1 - and now I'm not longer able to add or edit plugin properties.

So now I have two fresh installations with v2.4.1, one installation (on a different server) that was installed with v2.4.0 and then upgraded to v2.4.1 and one installation that was installed with v2.3.2 and then upgraded to v2.3.3 -> v2.3.5 -> v2.4.1.

I'm not able to add or edit plugin properties on any of these installations now (saving doesn't work / no server response). These installations are on different servers with different php versions and without mod_security. I tried with Google Chrome and Firefox. I checked all permissions - OK.

This must be a bug - caused by whatever.

Please let me know if someone wants access to an installation on my server to test/verify.

whitebyte:
[...]

PR #12614, opened 2015-08-28
Bugfix for #12580

Fi1osof:
https://github.com/modxcms/revolution/issues/12580

Issue #12580, opened 2015-08-23
Revo 2.4 Property set with the lowest id drops after saving any chunk or snippet


manicko:
I found that after saving any element (chunk, plugin or snippet) one of my property sets became empty. It is not deleted but only drops all parameters. I was trying to find out any dependencies and the only thing is that it is the Property set with minimal id. 
If i create the new Property set with same parameters it will not drop and if i delete the  Property set with the lowest id the next Property will drop.

bezumkin:
Quick fix by Fi1osof - https://gist.github.com/Fi1osof/ff3ea018841b1bb1f99b

Fi1osof:
@bezumkin Without discussion will not understand what need to edit. https://modx.pro/development/6312-how-not-to-extend-modx-processors/

bezumkin:
@Fi1osof As i understood we can just [replace this processor](https://github.com/modxcms/revolution/blob/e5727d94938269205e3fac84b8cb15c7eaf21b62/core/model/modx/processors/element/propertyset/updatefromelement.class.php) with your quickfix and it will work?

For proper PR need to add various lexicons and comments on english.

Fi1osof:
@bezumkin yes.

manicko:
thanks for  bug fix it works like a charm!

Fi1osof:
@manicko you are welcome! :)

mrhaw:
Nice! :+1:

mrhaw:
Google translated to english https://gist.github.com/mrhaw/97a664ad46e614899f1d

Fi1osof:
@mrhaw  Thanks!

Sended PR https://github.com/modxcms/revolution/pull/12614

PR #13799, opened 2018-03-01
Only save properties modified from the default for an element

opengeek:
In the Property Set UI, a "bug fix" incorporated in 2.4.2 broke property sets causing them to save all properties from an element to the set, instead of only those modified from the Default property values for an element. This should restore that behavior but still prevent the reported bug which lead to this change.

This reverts commit bd61e77bfba648563f7bf0faeaab6efb2cc599dc introduced via PR #12614 intended to address a bug reported in #12580.

### What does it do?
Reverted the changes which prevented comparing default properties to the values being sent to the processor.

### Why is it needed?
Restoring intended behavior to property sets (i.e. saving only properties to the set with values modified from the default specified for the element they originated from).

### Related issue(s)/PR(s)
#12580
#12614
#12695

### DIFF
--- a/core/model/modx/processors/element/propertyset/updatefromelement.class.php
+++ b/core/model/modx/processors/element/propertyset/updatefromelement.class.php
@@ -1,5 +1,6 @@
 <?php
-include_once dirname(__FILE__) . '/update.class.php';
+include_once dirname(__FILE__).'/update.class.php';
+
 /**
  * Saves a property set
  *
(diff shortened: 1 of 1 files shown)


Mark-H:
Rather than blankly reverting a bugfix, any chance there's a way to fix the issue you're seeing without also reintroducing the previous issue? In slack you mentioned something about it being hard to fix in both the property sets panels and the elements' properties tab, but if this were to get merged, someone else might want to reintroduce the other fix to address that issue again...

opengeek:
This was not blankly reverting anything. I can't reproduce the original issue at all. Feel free to test since no one did originally, obviously.

theboxer (approved):

Issue #13738, opened 2018-01-18
Resource Description field does not fit new Google Description 320 character limit


ATypicalGiraffe:
### Summary
Google has changed the maximum characters displayed to 320 characters in search result listings. MODX currently has a 255 character limit in the Description field.

### Step to reproduce
View page description returned in Google search results.

### Observed behavior
MODX Resource Description field will not accept updated Google character limit.

### Expected behavior
The default Description field should be expected to accomodate full SEO Description character limit.

### Environment
All versions - this is caused by external standards being changed.

digitalpenguin:
Actually as of 2.6 the max length is 191 characters due to UTF8mb4 support.
As varchar can't be longer than 191 characters, I assume if this is changed it would need to be TEXT type.

sottwell:
No, not all varchar fields, only those used as an index. In all cases here as far as I know, those are "name" fields, not "value" or "content" fields.

sottwell:
All of the elements that came from Evo had short name fields, like chunks and templates, have always been something like 50 chars. It's the new tables with Revo that had everything set to 255 chars by default. I suppose it seemed like a good idea at the time...

OptimusCrime:
>Values in VARCHAR columns are variable-length strings. The length can be specified as a value from 0 to 255 before MySQL 5.0.3, and 0 to 65,535 in 5.0.3 and later versions. The effective maximum length of a VARCHAR in MySQL 5.0.3 and later is subject to the maximum row size (65,535 bytes, which is shared among all columns) and the character set used.

UTF8mb4 uses 4 bytes so that means the highest length is rounded down to 65,535 / 4 = 16,383. This upper limit means that you easily can specify an upper limit of 320 * 4 = 1,280 without any problems.

**However**, the minimum requirement for MODX MySQL is version 4.1.20. 

This means that unless we change the field to be TEXT type, there is no way of limiting this without bumping the MySQL minimum requirement to 5.x.

Mark-H:
As description is not an index by itself (it is part of an index group `content_ft_idx`, but that's fulltext, so not affected by the utf8mb4/innodb limits), I see no problems with slightly bumping the description field to give people that bit of extra room. 

Alternatively, the `introtext` field is `text` so has no practical limit.

gpsietzema:
I'd love to see the description-field changed to text as well then.

dimitrihilverda:
+1 for me

alroniks:
PR is merged, so closing.

PR #13802, opened 2018-03-02
Change modResource.description column to text

bezumkin:
### What does it do?
Changes the type of modResource.description from varchar(255) to text
screen shot 2018-03-02 at 16 57 43

### Why is it needed?
To use more letters in description and work better with crawlers

### Related issue(s)/PR(s)
#13738

### DIFF
--- a/core/model/modx/mysql/modresource.map.inc.php
+++ b/core/model/modx/mysql/modresource.map.inc.php
@@ -99,8 +99,7 @@ $xpdo_meta_map['modResource']= array (
     ),
     'description' =>
     array (
-      'dbtype' => 'varchar',
-      'precision' => '191',
+      'dbtype' => 'text',
       'phptype' => 'string',
       'null' => false,
       'default' => '',
--- a/core/model/modx/sqlsrv/modresource.map.inc.php
+++ b/core/model/modx/sqlsrv/modresource.map.inc.php
@@ -96,7 +96,7 @@ $xpdo_meta_map['modResource']= array (
     'description' =>
     array (
       'dbtype' => 'nvarchar',
-      'precision' => '255',
+      'precision' => 'max',
       'phptype' => 'string',
       'null' => false,
       'default' => '',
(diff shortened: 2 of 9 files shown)


ghost:
Why not just increase the size of the varchar? Varchar is stored inline, text is stored as a pointer to the separate file the text is stored in. Varchar is faster for smaller, more frequently accessed content.

ghost:
Hm. A bit more research brought up the point that a size-limited varchar works until it doesn't. Today we may only want to increase from 255 chars to whatever limit the crawlers use now, until (like Twitter) they raise that limit. And exactly how InnoDB tables handle text appears to vary depending on the size of the text.

JoshuaLuckers:
What about changing/removing the textarea maxLength in [manager/assets/modext/widgets/resource/modx.panel.resource.js](https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/resource/modx.panel.resource.js#L589)?

bezumkin:
@JoshuaLuckers Good point! I will change it too and test better. Thank you!

alroniks:
@bezumkin Could you also fix lexicons to avoid doubles "``" in messages? Table name already quoted in `getTableName` method, so in lexicon strings, it should be just [[+table]] without `.

Also would be good to fix such issues in the whole `upgrade.inc.file` but not required.

bezumkin:
@Alroniks I think it should be the separate PR to fix this bug in Installer.

alroniks:
@bezumkin Could you fix only your string to avoid issues during the merging in future? I will prepare a fix for 2.6.1 (because it is a bug fix), but your changes do not exist there.

opengeek:
This should have been changed to varchar(350), IMO. Changing description to a field that takes up to 4Gb of size is a little extreme. We already have introtext for longer descriptions. Also, mediumtext would have been more appropriate than (long)text.

bezumkin:
@opengeek I have no problem with changing my PR, but MODX [have official server requirements](https://docs.modx.com/revolution/2.x/getting-started/server-requirements)
> MySQL Database Requirements: 4.1.20 or newer

And varchar longer then 255 available [only from version 5.0.3](http://ftp.nchu.edu.tw/MySQL/doc/refman/5.0/en/char.html)
> The effective maximum length of a VARCHAR in MySQL 5.0.3 and later is subject to the maximum row size (65,535 bytes, which is shared among all columns) and the character set used.

opengeek:
We don't need to support the older MySQL any longer.

bezumkin:
Glad to read it. 

Maybe somebody will write it in official documentation?

opengeek:
I will update it.

Issue #13633, opened 2017-10-10
Media Browser getting slow


pixelive:
### Summary
Media Browser is getting slow when you have a lot of files and directories.

### Step to reproduce
Add a lot of files and directories on your filesystem.

### Environment
Tried multiple environments but not a noticeable difference.

Does someone have a quick fix for this or some optimization tips?

OptimusCrime:
@pixelive Could you please be more specific so it is easier for others to recreate? How many files and directories do you have? What does slow mean? What browser do you use? Do you see any difference when using other browsers?

pixelive:
@OptimusCrime 

I think the problem is getting caused once retrieving the files, some folders have around ~400 files mostly images. The directories do not have that much impact I think (on the speed).

It sometimes takes around 10 seconds to see the first results within the media browser. 

We tried multiple browser, upgrading the server (which is pretty strong at the moment). The only thing which is untested at the moment is using PHP7.

It probably has to do with the connector call or thumbnails, but I was wondering if somebody had some quick fix for it.

OptimusCrime:
Thank you for more information. I suspected it was concerning images. I believe this is the culprit as well. I am not 100% sure, but I think the thumbnailing is happening for all images in the directory for each file in the directory. As far as I know, there is no easy way to get around this. One fix would be to be sure never to have that many files in a single directory. This is something you should do either way, as many files in a single directory may cause slow responses for various reasons.

Jako:
Having direct access to the thumbnail images without using a connector would be the solution. Maybe a deferred loading of viewport only thumbnails could avoid booting up modx inside the connector 400 times in such folders.

Jako:
See the $imageQuery generation visible in #13704 for modifying the thumbnail generation. The thumbs could be done with runSnippet pThumb + options or more general by a system event that calls pThumb or different installed thumbnailer or a direct phpThumb call + a caching solution.

rainbowtiger:
Thumbnail generation is definitely the problem here. Switch to list view and all is well. My server load goes sky high whenever viewing a directory with even a modest number of images in it. PLUS, the thumbnails don't actually seem to be cached on the server anywhere; they seem to be pushed to the browser and cached there. When I use one browser and wait for all the thumbnails to get generated, then switch to a different browser, then thumbnails get regenerated all over again. Is this actually what's going on here? Thumbnails should be cached on the server.

JoshuaLuckers:
[...]

PR #13805, opened 2018-03-05
Media manager optimizations

bezumkin:
### What does it do?
Enabled modx_browser_tree_hide_files system setting by default
Enabled caching of images from phpthumb connector by browser
Lazy load of images in media browser

### Why is it needed?
To make media manager work faster and make less requests to server. 

Tested on directory with more than 2700 images (click to see gif)
[](https://file.modx.pro/files/c/0/9/c09336f9fd12106c40d65087cc3491f9.gif)

### Related issue(s)/PR(s)
#13633

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1100,7 +1100,7 @@ $settings['manager_week_start']->fromArray(array (
 $settings['modx_browser_tree_hide_files']= $xpdo->newObject('modSystemSetting');
 $settings['modx_browser_tree_hide_files']->fromArray(array (
   'key' => 'modx_browser_tree_hide_files',
-  'value' => false,
+  'value' => true,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'manager',
--- a/connectors/system/phpthumb.php
+++ b/connectors/system/phpthumb.php
@@ -3,6 +3,8 @@
  * @var modX $modx
  * @package modx
  */
+
+session_cache_limiter('public');
 define('MODX_CONNECTOR_INCLUDED', 1);
 require_once dirname(__DIR__).'/index.php';
 $_SERVER['HTTP_MODAUTH'] = $modx->user->getUserToken($modx->context->get('key'));
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -472,7 +472,7 @@ $_lang['setting_modRequest.class'] = 'Request Handler Class';
 $_lang['setting_modRequest.class_desc'] = '';
 
 $_lang['setting_modx_browser_tree_hide_files'] = 'Media Browser Tree Hide Files';
-$_lang['setting_modx_browser_tree_hide_files_desc'] = 'If true the files inside folders are not displayed in the Media Browser source tree. Defaults to false.';
+$_lang['setting_modx_browser_tree_hide_files_desc'] = 'If true the files inside folders are not displayed in the Media Browser source tree.';
 
 $_lang['setting_modx_browser_tree_hide_tooltips'] = 'Media Browser Tree Hide Tooltips';
 $_lang['setting_modx_browser_tree_hide_tooltips_desc'] = 'If true, no image preview tooltips are shown when hovering over a file in the Media Browser tree. Defaults to true.';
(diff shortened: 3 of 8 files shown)


Jako:
Why didn't you move the thumbnail generation to the getlist processor? It could be initially slow during creating all thumbnails for one folder, but afterwards they are cached.

bezumkin:
@Jako Thumbnails are generated only when images loading in manager, because they have `src` that points to `phpthumb` connector. 
Lazy load solves exactly this problem - you will not generate any thumbnails while you can't see them.

And we can't move it to processor, because it depends from media source. Remote sources will not use phpthumb, I think.

Jako:
Yes, the way you have changed the thumbnail display code, the files use lazy (deferred) load. Thats better than the previous system for sure.

But that way each image still has to go through the `phpthumb` connector and each loaded image triggers a php handled request.

If the (media source) code could be changed to handle the thumbnail generation inside of the getlist connector, the images could be loaded without triggering php and accessing the image directly from a thumbnail folder inside of the webroot.

bezumkin:
What do you mean by _thumbnail generation inside of the getlist connector_?

To make actual thumbnails and store them in cache folder right in the moment of files listing? And no matter, will user scroll to them in manager, or not?

Jako:
The thumbnails could be generated and saved after this line https://github.com/modxcms/revolution/blob/73d6e0911db89b9b6d88b4f2e9210097d0f70459/core/model/modx/sources/modfilemediasource.class.php#L263 in the getContainerList method. Instead of a phpthumb connector call, the thumbnail should be generated and cached directly and a path/filename should be returned.

If a similar caching system as the system of pThumb is used, the thumbnails could be cached by the browser. As far as I see, the results of the phpthumb connector are not cached by the browser at the moment. Every click on a different folder in the MODX Resource Browser creates a new request for each thumbnail. And the thumbnail is not loaded from the browser cache.

The thumbnails have to be handled during uploading a new file or deleting/renaming an existing file.

Jako:
The deferred load in your patch is a good start to improve the speed. Browser cacheable thumbails could improve this significant.

bezumkin:
>  Instead of a phpthumb connector call, the thumbnail should be generated and cached directly and a path/filename should be returned.

Phpthumb processor already [caches thumbnails](https://github.com/modxcms/revolution/blob/d6fb64035ee2af9dab1d526d310b344447f6e0c3/core/model/modx/processors/system/phpthumb.class.php#L52-L55).

> As far as I see, the results of the phpthumb connector are not cached by the browser at the moment.

That exactly what this PR is fixes. You can see it on GIF file, that I attached in first post (size - "from disk cache" in DevTools).

Jako:
Ok, I did not noticed that session_cache_limiter change in the connector.

But now there could be a problem, if the image is changed on the server in another session.

Jako:
That could be solved, if the file filemtime is sent to the phpthumb connector.

bezumkin:
> But now there could be a problem, if the image is changed on the server in another session.

If you will upload new image with old name - browser will show you cached image even on frontend. So, it is expected behaviour.

labr1005:
> If you will upload new image with old name - browser will show you cached image even on frontend. So, it is expected behaviour.

Hm, but we're not talking about frontend here but about a media manager in the backend. When you replace an image on your local drive you (hopefully) see a thumbnail of the new image and not the old. Editors should always see the latest version in the backend. The presentation in the frontend is something else IMO.

Jako:
> So, it is expected behaviour.

Not in a CMS backend, that could be used by multiple users. It would be strange, if the full image is showing the current image and the thumbnail is different. It would be quite easy to add the result of filemtime (or better `$file->getMTime()`) as a property to the connector calls in https://github.com/modxcms/revolution/blob/73d6e0911db89b9b6d88b4f2e9210097d0f70459/core/model/modx/sources/modfilemediasource.class.php#L1109-L1130

Jako:
Thanks! It looks good now.

opengeek:
Does this apply to the file tree as well?

bezumkin:
@opengeek Yes.

This will enable caching of images in every place that uses phpthumb connector.

opengeek:
I just noticed after trying it out that thumbnails still show on hover in the file tree but not in the tree in the media browser. I'm not sure if that is even part of the intended behavior change, as I do not understand what `modx_browser_tree_hide_files` is supposed to do.

bezumkin:
`modx_browser_tree_hide_files` hides files when you in media manager and leaves only folders. It is very useful, because otherwise there will be loaded thousands of images right in the tree and it could be veeeery slow.

It is only for media manager tree and not affects usual files tree.

Issue #13806, opened 2018-03-06
Disable 'sudo' field for non-sudo users


ghost:
### Summary
Prevent non-sudo users from elevating themselves or other to 'sudo'.

### Step to reproduce
Edit any user, by any user, the 'sudo' checkbox is available.

### Observed behavior
A user can elevate himself or another user to 'sudo'.

### Expected behavior
A non-sudo user should not even see the sudo checkbox, and most certainly should not be able to elevate himself or another user to 'sudo'.

### Environment
Any

![screen shot 2018-03-06 at 20 47 08](https://user-images.githubusercontent.com/1073780/37051709-e16e2ab4-217f-11e8-9db6-dd71f63fe93a.png)

Bruno17:
but, by default, no one is a Sudo User, right?
So, no one would be able to set any Sudo User.

I think, it has to be a permission, if you can set Sudo Users

ghost:
The installed default user is Sudo.

Bruno17:
you are right, sorry
But I still think, there should be a permission 'set_sudo' or something like that

ghost:
This is intended as step one of tightening user management. Only sudo can do suso, step 2 only Administrator can do Administrator, step three otherwise user managers can only deal with users in the same groups as the current user. Or something along those lines. But one step at a time.

akimsullec:
It can be managed by plugin.

labr1005:
@akimsullec How?

akimsullec:
```
$en = $modx->event->name;
switch ($en) {
    case 'OnUserBeforeAddToGroup':
    case 'OnUserBeforeRemoveFromGroup':
    case 'OnUserBeforeSave':
        $user_current = (int)$modx->user->get('id');
        $user_managed = $user->toArray();
        if ($user_current !== 1 && (int)$user_managed['id'] === 1){
            $modx->false();
        }
        if($user_current !== 1 && (int)$user_managed['id'] !== 1) {
            if ($en === 'OnUserBeforeAddToGroup'){
                $user_group = $usergroup->toArray();
                if ( (int)$user_group['id'] === 1 || (int)$user_group['id'] === 4 ) {
                    $modx->false();
                }
            }
            if ($en === 'OnUserBeforeSave'){
                if ((int)$user_managed['sudo'] === 1){
                    $modx->false();
                }
            }
        }
        break;
    case 'OnUserDuplicate':    
    case 'OnUserBeforeRemove':
[...]

PR #13807, opened 2018-03-06
Add ACL "set_sudo" to prevent non-sudo users from elevating themselves or others to "sudo"

labr1005:
### What does it do?
- adds a new permission "set_sudo" ("To make any User sudo"), enabled for Access Policy "Administrator" by default
- shows/hides the checkbox "Sudo User" on the user create/update page
- checks for "set_sudo" permission on create/update user

### Why is it needed?
To prevent non-sudo users from elevating themselves or others to 'sudo'.

### Related issue(s)/PR(s)
#13806


ghost:
I have a problem with this particular way of handling the sudo issue. Should non-Sudo users _ever_ be able to manipulate the Sudo property of any other users? What is the use case of such an ability? ACL permissions are for optional abilities, aren't there some actions that should never be allowed? For example, unpublishing or deleting the site_start page comes to mind.

labr1005:
With this ACL in place a non-sudo user can't change any user's sudo status - including their own status.

Without an ACL, if you disable sudo for yourself, how could you ever enable it again for yourself? You would become a non-sudo user without the permission to change the status for any user...

Bruno17:
+1 for this solution (didn't test it, yet)
It should be possible to give set_sudo - permission to groups at will.

Mark-H:
This approach looks good as a way to limit users from assigning themselves sudo without permission (hah!), and passes my code review 👍

Mark-H (dismissed):

opengeek (changes requested):
Can you please remove this merge from the pull request? You should never merge the base branch back into a pull request branch. If you want to update the pull request, use `git rebase upstream/2.x` (assuming the remote for the `modxcms/revolution` repository is named `upstream`) and do a forced push of the branch to your fork to update the pull request.

labr1005:
@opengeek Done. Can you please have a look and sorry for this. My purpose was to include the `"events":true` from #13830 which is now missing again in my PR... https://github.com/modxcms/revolution/pull/13807/files#diff-65f2d46431c4fd7a2cdb261551bf9fe7

opengeek:
@labr1005 if it is dependent on something in another PR it should probably be added to that PR. If that PR is already merged, just rebase on top of the latest code and force push, as I explained.

labr1005:
@opengeek I did a rebase and force push, but that long one-liner made trouble. Fixed now.

opengeek (dismissed):

opengeek:
Somehow, after merging this and running upgrade, my Administrator policy has 0 of 175 permissions. Investigating…

labr1005:
I confirm and will look into this.

labr1005:
Found it.
`...,"events":true,export_static":true,...`
`...,"events":true,"export_static":true,...`

opengeek:
Thanks for taking a look @labr1005 — I'll test again shortly and get this merged.

opengeek:
Hmm, I'm not sure this is enough to be a practical solution to the problem. If I can edit users, I can put myself in the Administrators group and in this way I can essentially just give myself access to edit sudo on any user.

labr1005:
You're right. But IMO that's a different problem that you show up here - the missing possibility to define for users/groups what groups they're allowed to set when creating/editing users.

bezumkin (approved):
Works as expected.

Of course, if you have enough permissions you can just enable `set_sudo` permission to you group and make yourself sudo user. 
But, anyway, it is better than nothing.

PR #13809, opened 2018-03-07
Enable permission "events" for Access Policy "Administrator" at MODX install

labr1005:
### What does it do?
This enables the permission "events" for Access Policy "Administrator" at MODX install.

### Why is it needed?
Currently the permission "events" for Access Policy "Administrator" is disabled by default at MODX install.

bildschirmfoto 2018-03-07 um 20 34 43


Mark-H (approved):

Mark-H:
While the PR targets 2.x, as it's a (very useful) bug fix I think it would be best to add to 2.6.x.

opengeek:
I've pushed a new PR #13830 targeting `2.6.x` with a cherry-pick from this PR.

Issue #13711, opened 2017-12-04
An Open angle bracket < in custom field  within Manager Form crashes the save process


donShakespeare:
An open angle bracket `<` in custom field (from MODx.onDocFormRender or however) within Manager Form crashes the save process

### To Reproduce
1. Please insert custom field using this tut
 https://docs.modx.com/revolution/2.x/case-studies-and-tutorials/adding-custom-fields-to-manager-forms
2. Then alter the content of said field to have an open angle bracket.
3. Then save the resource

### Quick Reproduce
1. Open a native MODX resource in the Manager
2. Run code in browser's JS dev console
3. Save the resource
```
 var modxForm = document.getElementsByTagName('form')[0];
  var newTextarea = document.createElement("textarea");
  newTextarea.setAttribute('id', 'mynewTextArea');
  newTextarea.setAttribute('name', 'mynewTextArea');
  newTextarea.value='<THIS__WILL__CRASH'; //crashes when you save the resource
  modxForm.appendChild(newTextarea);
```

donShakespeare:
My current solution is

**JS**
```javascript
document.getElementById("myEl").value = content.replace(/(?:<)/g, "&lt++;");
 //snuff out all opening angle brackets with custom entity
```

**PHP**
```php
$content = str_replace("&lt++;", "<", $content);
//remove your custom entity before saving resource
```

gadgetto:
I can confirm this!

Finetuned:
After looking into this in modxbughunt#3, the issue is due to Ext.data.Connection truncating the json which arrives intact from modconnectorresponseclass->outputContent(). 

So it is already invalid _before_ it arrives into utilities.js -> Ext.override(Ext.form.Action.Submit.... handleResponse() where the error is raised.

The ExtJS docs for Ext.data.Connection state that: 

> Characters which are significant to an HTML parser must be sent as HTML entities...

So the content from custom fields must be processed before it gets sent to the server.

OptimusCrime:
Nice detective work. Is there any way to solve this issue, or is the issue in Extjs itself?

Finetuned:
It's in Ext. It creates a iframe and drops the response text into the body. I am thinking to override the class to add a <textarea> into the iframe. The response would be put into that and not be parsed as html.

OptimusCrime:
Sounds tricky. Perhaps this is impossible to deal with properly until we can get rid of extjs or bump its version?

Finetuned:
[...]

PR #13810, opened 2018-03-08
Bug 13711 - An Open angle bracket < in custom field within Manager Form crashes the save process

Finetuned:
### What does it do?
This is a two-part bugfix:

1. Encloses json output from the modConnectorResponse class in a <textarea> tag
2. Overrides the handleResponse function in Ext.data.Connection to remove the <textarea> tag from the response body after it has been processed by doFormUpload function in Ext.data.Connection

### Why is it needed?
A single angle bracket in a custom field in the Manager crashes the save process. This is due to Ext.data.Connection using a fake ajax response (?!):  Json returned from the server is dumped into the body of a hidden iframe. The browser truncates the json at the unescaped angle bracket triggering a json decode exception down the line.

The textarea encapsulation is already implemented in Ext.data.connection in doFormUpload.
[See the function cb for details.](http://docs.sencha.com/extjs/3.4.0/source/Connection.html#Ext-data-Connection)

### Related issue(s)/PR(s)
As listed in [13711](https://github.com/modxcms/revolution/issues/13711)

### DIFF
--- a/core/model/modx/modconnectorresponse.class.php
+++ b/core/model/modx/modconnectorresponse.class.php
@@ -179,7 +179,7 @@ class modConnectorResponse extends modResponse {
             if (!empty($_GET['callback'])) {
                 $json = $modx->stripTags($_GET['callback']) . '(' . $json . ')';
             }
-            die('<textarea>'. $json .'</textarea>');
+            die($json);
         } else {
             @session_write_close();
             die($this->body);
--- a/core/model/modx/modconnectorresponse.class.php
+++ b/core/model/modx/modconnectorresponse.class.php
@@ -179,7 +179,7 @@ class modConnectorResponse extends modResponse {
             if (!empty($_GET['callback'])) {
                 $json = $modx->stripTags($_GET['callback']) . '(' . $json . ')';
             }
-            die($json);
+            die('<textarea>'. $json .'</textarea>');
         } else {
             @session_write_close();
             die($this->body);
--- a/manager/assets/modext/util/utilities.js
+++ b/manager/assets/modext/util/utilities.js
@@ -319,6 +319,26 @@ Ext.override(Ext.form.Action.Submit,{
     }
 });
 
+Ext.override(Ext.data.Connection, {
+    handleResponse: function(j) {
+        this.transId = false;
+        var k = j.argument.options;
+        j.argument = k ? k.argument : null;
+        // trim out the enclosing textarea tag (the long way to avoid any tags in the response body) if exists
+        j.responseText = j.responseText.replace('<textarea>', '');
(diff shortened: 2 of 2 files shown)


core/model/modx/modconnectorresponse.class.php
```diff
@@ -179,7 +179,7 @@ public function outputContent(array $options = array()) {
             if (!empty($_GET['callback'])) {
                 $json = $modx->stripTags($_GET['callback']) . '(' . $json . ')';
             }
-            die($json);
+            die('<textarea>'. $json .'</textarea>');
```
Mark-H: I remember you talking about a hack-ish approach, but I didn't quite see this one coming 😬This is going to break interaction with connectors from non-extjs sources, which I know people are doing, so I think we'll have to go back to the drawing board on this one...

manager/assets/modext/util/utilities.js
```diff
@@ -319,6 +319,26 @@ Ext.override(Ext.form.Action.Submit,{
     }
 });
 
+Ext.override(Ext.data.Connection, {
+    handleResponse: function(j) {
+        this.transId = false;
+        var k = j.argument.options;
+        j.argument = k ? k.argument : null;
+        // trim out the enclosing textarea tag (the long way to avoid any tags in the response body) if exists
+        j.responseText = j.responseText.replace('<textarea>', '');
```
Mark-H: I suspect this would also remove textarea's from within the content (say, someone having a form markup in a chunks' content). 

Is there a different way to handle `j.responseText` here to fix the issue, which doesn't require the `<textarea>` markup to be returned from the processor? Maybe that can be dynamically added in this handleResponse method or elsewhere to fix the issue?

Mark-H (changes requested):
Please see the code comments - this will break too much to merge in its current state I'm afraid :(

core/model/modx/modconnectorresponse.class.php
Finetuned: Yes, I thought this might have far-reaching consequences. This was working back from Ext.data.Connection which provides for the textarea addition

manager/assets/modext/util/utilities.js
Finetuned: Noted. I added this in to restore 'normal behaviour ' after the enclosing above. I could be more specific about the stripping but it doesn't resolve the first problem. Back to square one then : s

Mark-H:
I may have found a better fix for this, see https://github.com/modxcms/revolution/issues/13711#issuecomment-401204379

Mark-H:
Closing as this would introduce non-acceptable breakage. The other proposal might work better but has not yet been turned into a PR.

PR #13811, opened 2018-03-09
Update media sources

bezumkin:
I tried to add a new media source after implementing Flysystem and run into some troubles with new "visibility" feature.

I was a little carried away, when I tried to fix it, sorry.

### What does it do?
- Refactored modMediaSource
- Refactored modS3MediaSource
- Static elements now works via media sources API
- Increased speed of work due to optimization of requests
- More detailed manager log records about work with files
- Media sources are no longer require modFileHandler
- Added modFtpMedia source (because its adapter is installed by default anyway)
- Replaced "chmod" files to FlySystem "visibility". Each media source can support it for files and directories separately.
- Removed processor "directory/browser/chmod"
- Refactored "browser/file/unpack" processor.
- "Unpack" menu item displayed only for local media sources.
- Added new ExtJS xtype "modx-combo-visibility"
- Added ability to return and display error on ExtJS tree load.
- English lexicons "file" and "source" was edited, according to changes.
- Fixed every founded error.

- Refactored all browser processors.
- Added support of characters like `@#$%^&*()` in filenames.

screen shot 2018-04-13 at 09 56 44

### Why is it needed?
To fix errors in current media source implementation and provide an ability to add new types of sources.

### Related issue(s)/PR(s)
#13709


core/model/modx/processors/browser/file/download.class.php
```diff
@@ -34,9 +35,22 @@ public function getObjectUrl() {
...
+            /** @var League\Flysystem\Filesystem $filesystem */
+            $filesystem = $this->source->getFilesystem();
+            if ($data = $filesystem->getMetadata($this->getProperty('file'))) {
+                $name = array_pop(explode(DIRECTORY_SEPARATOR, $data['path']));
+                header('Content-type: ' . $data['mimetype']);
+                header('Content-Disposition: attachment; filename=' . $name);
+                header('Content-Length: ' . $data['size']);
 
-        $fileobj->download();
+                exit($filesystem->read($data['path']));
```
Mark-H: To prevent E_WARNING notifications, before `exit()` add `@session_write_close();` to close the session.

core/model/modx/sources/modmediasource.class.php
```diff
@@ -452,129 +520,164 @@ public function getObjectsInContainer($path)
...
 
         if (!$this->filesystem->has($path)) {
-            $this->addError('file',$this->xpdo->lexicon('file_err_nf'));
+            $this->addError('file', $this->xpdo->lexicon('file_err_nf'));
         }
 
         try {
             /** @var League\Flysystem\File $file */
             $file = $this->filesystem->get($path);
+        } catch (\xPDO\Exception\Container\NotFoundException $e) {
```
Mark-H: is this the right exception type here?

core/model/modx/sources/modmediasource.class.php
```diff
@@ -811,45 +956,38 @@ public function renameObject($oldPath, $newName)
...
 
         /* check to see if the new resource already exists */
         if ($this->filesystem->has($newPath)) {
             $this->addError('name', sprintf($this->xpdo->lexicon('file_err_ae'), $newName));
+
             return false;
         }
 
         if (!$this->checkFileType($newName)) {
-            // @TODO what should the error message be?
```
Mark-H: was this todo resolved? Looks like it returns no error message now at all

core/model/modx/sources/mods3mediasource.class.php
```diff
@@ -47,181 +41,298 @@ public function initialize() {
...
 
+
+    /**
+     * @return array
+     */
+    public function getListDirContextMenu()
+    {
+        $menu = parent::getListDirContextMenu();
+        foreach ($menu as $k => $v) {
+            if ($v['handler'] == 'this.renameDirectory') {
```
Mark-H: Interesting approach to disallowing the rename on s3 directories - assuming you checked if there were a cleaner way to prevent that option?

Mark-H (commented):
Some minor comments - the exception one is probably most important to check. Seeing those xPDO exceptions in several places, while I think you may've meant to use the Flysystem FileNotFoundException?

Other than that, passes my code review. Nice work!

core/model/modx/sources/mods3mediasource.class.php
bezumkin: Another option is to do not use the parent method and generate the whole menu in S3 class. I think, it is better to just remove one item.

core/model/modx/sources/modmediasource.class.php
bezumkin: Yes - the error message is set in method `checkFileType`.

core/model/modx/sources/modmediasource.class.php
bezumkin: Of course, not! Thank you for your attentiveness.

bezumkin:
@Mark-H I found a vulnerability in static elements workflow. You can create any file on disk, even its extension is banned in system settings. It is because modElement internal methods hardcoded to work only with local file system and don`t use media sources.

Now it is fixed too.

gpsietzema:
@Mark-H @opengeek I just had a discussion with @bezumkin about the following: you can't have support for Amazon AWS S3 version 2 AND version 3. They require different libs.

That shouldn't really be a problem, because most sites/applications use just one bucket or one location, meaning one S3 version. A few weeks ago though, we decided to have the S3 adapter shipped with MODX3 by default (version 3).

There are multiple possible solutions:
1. Remove S3 from the core and have two extras available: AWS S3 V2 + AWS S3 V3 adapter. Install the one you need.
2. Leave version 3 in there and have an extra for AWS S3 V2, which removes V3 from the core on installation.

Solution 2 sounds really bad and unnatural. Solution 1 means a MODX2 > MODX3 upgrade needs some manual work when using S3.

Our vote goes to solution 1. What about you guys?

Jako:
I prefer removing S3 from the core and two installable extras.

gpsietzema:
So, @Mark-H MarkH just brought up the brilliant idea that the V3 adapter might be backwards compatible ;-) Just tested that and he's completely right. So no reason to worry about this anymore. Thanks mark :D

sepiariver:
Some tests performed so far: https://docs.google.com/document/d/1boaa3mI02yulpYPfnorI0H-lUKM27EDq1ocsV0jd62Q/edit?usp=sharing

(Will update in-situ)

bezumkin:
@sepiariver Thanks, great work!

Extract ZIP can work only on local media sources, because it uses php `xPDOZip` class, that know nothing about remote sources.

Edit of files depends on mime type, that media source will send to website - it should be `text/anything`. I think it is much better than [read first 512 bytes of each file](https://github.com/modxcms/revolution/blob/c0da88f733c4de74128ea2c207a8866f92eab20e/core/model/modx/sources/modmediasource.class.php#L1964-L1984) to understand, can we edit it or not.

About Tree State Persistence - I think that it was made for good (not by me), because we have a lot of complains about heavy trees that trying to expand on every page load and you need to wait for them.
I don`t think, that it is the part of this media sources PR.

sepiariver:
It’s you who is doing great work @bezumkin :) but thanks for the kind words. Understood re: editing remote files—those test results can be ignored, as with Tree persistence.

Will continue to test more on local Filesystem. One thing I think I missed: should the media source output the remote URL? Not sure I saw that...

bezumkin:
>  should the media source output the remote URL?

No, we have no this function, but I can add it!

sepiariver:
That one might be useful, because if a TV is created to use a remote media source, when the value is rendered (say inside an img tag) the browser will need the remote base URL in order to include the file.

Thanks again for everything you’re doing @bezumkin !!!!

sepiariver:
Updated with local filesystem tests: https://docs.google.com/document/d/1boaa3mI02yulpYPfnorI0H-lUKM27EDq1ocsV0jd62Q/edit#

TODO: MS test permissions

bezumkin:
@sepiariver Upload of PNG in allowed by default. Here is test of unalowed extension
screen shot 2018-03-17 at 08 51 07
screen shot 2018-03-17 at 08 51 41

sepiariver:
Thanks @bezumkin there is the system_setting for allowed file types, and those tests pass, however in the media source there is another config setting for allowed file types—it seems to only affect what is viewable in the media source. You can upload whatever is allowed in the system_setting, even if media source config setting exists and does not include the current file type.

Maybe this is expected behavior? It’s quite possible I misunderstand the goal/usage of the media source config setting...?

bezumkin:
@sepiariver As I see, it is about display files, not upload
screen shot 2018-03-17 at 16 10 37

And I didn`t test it, btw.

sepiariver:
Ok if that’s what the setting is for, then it works perfectly :)

bezumkin:
@sepiariver Great!
Looks like everything is works okay, and if we will find something later - will fix it later.

Big thank you for checking my PR, it is very important for me.

sepiariver:
No problem! Note: still need to test when different permissions assigned to different users for MS’s. I will do that this weekend.

One thing: I think the TV is still rendering without MS baseUrl. Maybe I setup the TV incorrectly: I’ll double check. In case it’s not rendering I think it’s still an important functionality...

Overall this implementation is fantastic, @bezumkin you’ve done a great thing here.

bezumkin:
> One thing: I think the TV is still rendering without MS baseUrl. Maybe I setup the TV incorrectly: I’ll double check. In case it’s not rendering I think it’s still an important functionality...

Today I fixed it and send addition to this PR. You should see the changes on our test site.

sepiariver:
Updated with some permissions testing: https://docs.google.com/document/d/1boaa3mI02yulpYPfnorI0H-lUKM27EDq1ocsV0jd62Q/edit# 

Can't really continue testing because it seems as soon as a permission rule is assigned the Administrator Group no longer gets access to it. (Please see items 3 and 5 in "Permissions" section of Google Doc)

sepiariver:
Regarding the baseUrl prefix on TV render, I don't think I see it there...? view-source:http://s13174.sterc.modhost.pro/index.php?id=1

bezumkin:
@sepiariver Fixed [url of TVs on frontend.](http://s13174.sterc.modhost.pro/index.php?id=1)

Permissions for objects are the part of xPDO, not media sources. When you add access to media source at least for one group - all other group need to have permission too, or they will not see it at all.

Once admin have permission to "Test local"
screen shot 2018-03-18 at 08 17 41

Content editor can`t access this MS anymore
screen shot 2018-03-18 at 08 21 28

So, it will work anywhere on xPDO level - you can`t just load media source object without proper permissions.

sepiariver:
Thanks @bezumkin the TV output looks perfect!

Regarding permissions you’re right of course it’s handled elsewhere. However I suspect something is going wrong. In current Revo version, you can give Admins MS Admin Access, and then also give Content Editor MS User Access.

I wasn’t able to accomplish this, because as soon as I assigned MS Admin Access the MS disappeared from the combo-box to set permissions on it for another group.

Are you able to reproduce this?

bezumkin:
I don`t see any problem. As an admin I see this media source and can set it for anybody
screen shot 2018-03-18 at 16 59 01

Anyway, it have nothing to do with media sources, is handled only by xPDO permissions system.

sepiariver:
Maybe it’s some issue with my user. I’ll try some ways around it. If it persists I’ll make another Issue when this PR gets merged. Maybe it’s a side effect of some other commit.

Thanks again!!

Mark-H (dismissed):

gpsietzema:
Everything is working, but I have some feedback which we should get in 3.x. There are two options now @bezumkin 
1. Merge this and create new issues for the things I found
2. Add my issues here, change the PR and then merge it

What do you prefer?

bezumkin:
@gpsietzema I prefer 2.

gpsietzema:
* The visibility-context-menu-item doesn't work on the file-tree, media tree or browser window. To replicate: got to "Media" -> "Mediasources" -> Select S3 media source -> Rightclick a random file in the Tree, click 'Visibility' -> Nothing happens, no error, nothing. Tested on default media source + a S3 V3 bucket (Frankfurt).
* The dialog window for creating a new media source is too small. The error message of the 'Source type' error message is only visible for 50%.
* FTP mediasource: Image previews only work if you set the 'url' field. The description lexicon for this field is currently "If this FTP is connected to website, you can enter its external http address.". It should be changed to: "If this FTP is has a public URL, you can enter its public http-address here. This will also enable image previews in the media browser."
* The full image path is shown instead of the filename on the thumb-tile and in the right sidebar. See image 'fullpath-3xfilesystem.png'

fullpath-3xfilesystem

* When moving around files, the open directory structure is forgotten. Example: Create 'folder1', in that folder create 'folder2'. Now upload a file in 'folder2'. Move that file to 'folder1' or even the root of your media source. The whole tree collapses, while the expected result is the tree to stay open. It not only collapses the tree to the level you dropped the file in, but also reloads the whole thing. This isn't a very smooth transition.
* modx_browser_tree_hide_files isn't working anymore

gpsietzema:
All my things were fixed (where possible) and tested. Would be nice to now have this merged into 3, so we can continue working on 3, with more and more features merged into the 3 branch. If we find more issues, we can create new issues for this.

The above has been tested by me and YJ.

opengeek (dismissed):
In the future, please do not commit the modx.jsgrps-min.js file into the PR, but do note that it requires a grunt build. I will work on merging it today and handle the conflicts on the minified file.

bezumkin:
@opengeek I do this only because MODX uses minified version by default and when people test my PR - they see old javascript and notice me about problems.

opengeek:
@bezumkin I knew you were going to say that. 😅 I would love to see us get rid of the minified stuff altogether, but not sure if that is practical. Something we need to explore moving forward.

bezumkin:
I updated PR with refactored browser processors to avoid such problems https://github.com/modxcms/revolution/issues/13855

Mark-H (dismissed):
I found one small issue with this PR. Dragging a file from one source into the *root* of another one (by dropping it onto the source name) throws an error that no path is specified. 

As the moving between sources is a new feature, this doesn't have to block merging, so I am going to approve this, but it should be logged as a new issue if we decide to merge first.

It's missing some lexicons while testing, but I'm 99% sure that's my environment acting up as they do appear to be in place properly in the code.

bezumkin:
@Mark-H Fixed move file from one source to the root of another.

Please, approve this PR again!

arjen-t:
I reviewed the code and it looks to me as an solid rewrite, no weird bugs appeared on mine local machine, thumbs up for the rewrite! I pretty sure a bug slipt into the rewrite. 

It can be found in the method checkFileType in class modmediasource on line 1717;

https://github.com/bezumkin/revolution/blob/07a8b6827be5e25050f4e453dab7cf60b53f23cc/core/model/modx/sources/modmediasource.class.php#L1717

It's checking if 'upload_images' options exists but instead the 'upload_files' options are added to the array.

I noticed this 'bug' also exists in the current 2.* repo.

p.s. why not allowing each Media Source has his own upload limit instead of an general upload limit? So basically treat this option the same way as allowed file types. This gives a more convient control for each media source to control the bandwith of the file size (maybe this is not the right place to discuss)

Mark-H (dismissed):

gpsietzema (dismissed):

bezumkin:
Sorry @gpsietzema & @Mark-H, I rebased this PR because @opengeek told me that.

Hope, now it is okay.

Mark-H (dismissed):

gpsietzema (approved):

Mark-H (approved):
4th time I've approved this PR, if you want it to get merged can you please stop force pushing/rebasing/adding/whatever you keep doing to it now? :P

bezumkin:
@Mark-H Sorry, but it is really big PR. I will just merge it myself, finally.

Mark-H:
I guess now that we use approvals it's okay to self-merge, but "it's a really big PR" is a terrible reason to just mash that merge button. _Especially_ if it's big.

bezumkin:
@Mark-H I meant, that because it is big - I spent a lot of time to make it ready to merge. That is why here was a lot pushing/rebasing/adding/whatever.

But now everything is done, that is why I pressed this button.

Issue #13697, opened 2017-11-16
Remove flash (.swf) based copy to clipboard


OptimusCrime:
### Summary
MODX is shipped with a [Flash file](https://github.com/modxcms/revolution/tree/2.x/manager/assets/modext). This is used to set up the old way of doing ["Copy to clipboard"](https://github.com/modxcms/revolution/blob/fbf126690f6ce903faeaeb9f1d958211fc2203c9/manager/assets/modext/util/utilities.js#L361-L388). This functionality was in MODX before, but has since been removed. I see no traces of this functionality anywhere but [lexicon entries and changelogs](https://github.com/modxcms/revolution/search?p=1&q=Clipboard&type=&utf8=%E2%9C%93). I think we can remove this and get rid of the Flash file once and for all.

Edit: Actually, MODX is shipped with [more Flash files](https://github.com/modxcms/revolution/tree/2.x/manager/assets/fileapi). I do not know if these are in use. This issue only concerns the clipboard one.

### Environment
Current version c2466999ddc527186d7855330bb5562901d23001

rthrash:
Some possibly useful things to help evaluate:

Quick search everywhere the fileAPI is referenced: https://github.com/modxcms/revolution/search?utf8=%E2%9C%93&q=fileapi&type=

How to banish the flash bits on the fileAPI itself and only use HTML5 methods: github.com/mailru/FileAPI

(Speaking of which looks like we're more than a few versions behind on that dependency…)

OptimusCrime:
Thank you for doing your research @rthrash :) Nice to have written down for later. Worth pointing out that this issue only concerns the other swf file though.

JoshuaLuckers:
This issue may be closed.

PR #13814, opened 2018-03-14
Remove flash (.swf) based copy to clipboard (#13697)

JoshuaLuckers:
### What does it do?
This pull request deletes a flash file, removes a Core JS function and lexicon entries.

### Why is it needed?
The "copy to clipboard" function isn't being used anymore.

### Related issue(s)/PR(s)
#13697


Mark-H:
Thanks!

Issue #13816, opened 2018-03-15
Problem with system settings


salamanderer:
### Summary
If user set "cache_system_settings" = 0 (No), system setting didn't apply automatically. Even if he change it back to "cache_system_settings" = 1 (Yes)!
So user definitely must perform two actions: 1)set "cache_system_settings" = 1 (Yes), after that - 2)clear the folder "/core/cache/" manually.

### Step to reproduce
Set "cache_system_settings" = 0, change "site_name" for the visible effect, refresh the page. 
Then set "cache_system_settings" = 1, change "site_name" for the visible effect, refresh the page. 

### Observed behavior
The site name changed in visible system settings.
The site name didn't apply in both cases.

### Expected behavior
The site name changed in visible system settings.
The site name applied in both cases.

### Environment
Any i think, (my version of Modx is MODX Revolution 2.6.1-pl)

JoshuaLuckers:
I'm closing this because `cache_system_settings` has been removed as a setting in system settings for new installations.

PR #13817, opened 2018-03-15
Clean the system settings cache after saving system settings

JoshuaLuckers:
### What does it do?
It makes sure we clean the system settings cache after saving them. 

### Why is it needed?
It solves an issue where the cache has to be deleted manually when turning on/off cache_system_settings. 

### Related issue(s)/PR(s)
#13816

### DIFF
--- a/core/model/modx/modcachemanager.class.php
+++ b/core/model/modx/modcachemanager.class.php
@@ -557,7 +557,9 @@ class modCacheManager extends xPDOCacheManager {
                     $results['auto_publish'] = $this->autoPublish($partOptions);
                     break;
                 case 'system_settings':
-                    $results['system_settings'] = ($this->generateConfig($partOptions) ? true : false);
+                    if ($results['system_settings'] = $this->clean($partOptions)) {
+                        $results['system_settings'] = ($this->generateConfig($partOptions) ? true : false);
+                    }
                     break;
                 case 'context_settings':
                     if (array_key_exists('contexts', $partOptions)) {


Mark-H:
generateConfig should be taking care of that already (plus I don't recognize the issue you've described though I haven't tried to reproduce it yet) - maybe this warrants more research into why the generate method isn't doing what it should for you before applying a workaround like this?

JoshuaLuckers:
I’m not behind my computer at this moment. But if I recall correctly it’s because getCollection() in generateConfig retrieves items from the cache. So when you change the cache system settings from on to off the getCollection() will still fetch the cached results because the cache file with the old value is being used.

Mark-H:
getCollection would only (potentially) retrieve from cache if cache_db is enabled as far as I know, which is off by default... My feeling is that this PR is jumping to the wrong conclusion about the root cause for a potentially isolated problem, but I'd be happy to be proven wrong! :)

The potential problem I see with calling `clean` before `regenerateConfig` is that on high traffic sites, visitors might hit an empty settings cache, which MODX will try to regenerate, even though another process is already doing the same. That's bad for performance and may open up weird timing-related issues, double writes, or file locking issues preventing the process from completing...

JoshuaLuckers:
Point taken. I will investigate the issue a bit more.

opengeek:
Yes, the system settings and context settings cache should always be replaced and never removed. If this is a cache_db issue, I would recommend not using cache_db and/or investigate the problem from that perspective, i.e. if a fix should be proposed for this it needs to solve the cache_db problem.

Issue #13609, opened 2017-09-14
Loading wrong mediasource on create page


MrRoco:
### Summary
When creating a new resource in a context where a default media source is defined in the context settings different from the default media source system setting it loads the system setting, which results in not working images.

When updating the correct media source is loaded but image preview doesn't work anymore.

### Step to reproduce
- Create multiple mediasources with different paths  1 =  /uploads/1, 2 = uploads/2 etc 
- 1is default media source in system setting. 
- Create context and set default media_source 2
- Create new page in created context
- Add images in tv's
- Save & reload page
- Preview in TV's do not work + image path is incorrect

### Observed behavior
When adding image through tv the wrong media source is selected.

### Expected behavior
Load the correct media source from context setting in which the page is being created

### Environment
Modx 2.5.6, nginx

MrRoco:
The fault seems to be in the following file: core/model/modx/processors/element/tv/renders/mgr/input/image.class.php line 15,
on page create if fetches the default context (mostly web) and the media_source belonging tot that.
When people in different user groups with different media source try to create a page in a non default context, they won't be able to select an image.

johanmolen:
We just reproduced this issue in MODX 2.5.7-pl.

Mark-H:
Fixed in 2.6.2 per #13819

PR #13818, opened 2018-03-16
Save resource given from controller to proper render of TVs

bezumkin:
### What does it do?
If modElement::renderInput() method will get resource object, it should be loaded as current resource so any TVs can check it properties and act correctly

### Why is it needed?
It fixes weird issues when you create a new resource in another context but TVs uses context "web".

### Related issue(s)/PR(s)
#13609


gpsietzema:
@sdrenth @pixelive Try this ;)

PR #13819, opened 2018-03-16
Fix for #13609

bezumkin:
### What does it do?
If modElement::renderInput() method will get resource object, it should be loaded as current resource so any TVs can check it properties and act correctly

### Why is it needed?
It fixes weird issues when you create a new resource in another context but TVs uses context "web".

### Related issue(s)/PR(s)
#13609

_Recreated this PR because of wrong target branch_


Mark-H:
Thanks!

PR #13820, opened 2018-03-17
Fixed display of help block on login screen

bezumkin:
### What does it do?
Added css class to display help block on login screen

### Why is it needed?
Because I forgot it in last PR -(

### Related issue(s)/PR(s)
#13786


Mark-H (approved):
Tested and indeed brings back the help panel when enabled.

gpsietzema:
Works as expected:
screen shot 2018-04-04 at 12 52 08

alroniks (approved):

Issue #13815, opened 2018-03-15
Fixed URI too long when using "use_alias_path".


Yoshix:
### Summary
Vague error message when saving a resource *"Please correct the errors in your form before submitting."*, without having the relevant input being highlighted.

### Step to reproduce
- use `use_alias_path`
- have a nested resource tree, with resources having very long aliases.
- try to save the child resources when the URI becomes longer than 191 characters
- an error message *"Please correct the errors in your form before submitting."* will be shown, but to actually know what the error is, one has to check the *"Freeze URI"* checkbox. The text-input will then be highlighted as invalid.

### Expected behavior
Not having an upper limit on the auto-generated URI. Especially with `friendly_alias_max_length = 0`.
Or at least having the hidden text-input be shown automatically with a clear error message.

### Notes
Setting the `modx_site_content.uri = NULL` will allow to save the resource once, though for the second time the URI will have been regenerated and the above behavior occurs. The same is true, when using *"Refresh URIs"*.

### Environment
MODX Revolution 2.6.1-pl

Mark-H:
That's a really long URI!

The 191 character limit is the most that can be supported on the different database platforms we try to support for an index. As the URI is an important field to have indexed, and 191 seems to me like it should be enough for 99.9% of the urls, I don't think I'd be in favor of increasing the limit.

Improving the error handling of course is always a good thing.

Yoshix:
@Mark-H Thank you for your reply!

That was my guess as well, and until now I didn't have any similar problems with previous projects.

But stumbling over this, a few things seem very counter-intuitive to me:

- The wording for `friendly_alias_max_length` implies that there is no upper limit, but in fact a limit on the database-layer is imposed (`varchar(191)`)
- The error message for the pagetitle and alias reads: `The maximum length for this field is 255` but entering anything above 191 chars will be silently cut off.
- And (though I understand the reasons) having an index on `modx_site_content.uri` with a max-length while the field itself is simply `text` seems inconsistent and problematic.

Mark-H:
> And (though I understand the reasons) having an index on modx_site_content.uri with a max-length while the field itself is simply text seems inconsistent and problematic.

That indeed looks like a problem! Never noticed that before, I just assumed the field itself was a varchar(191). 😬 

I've sent a fix for the first two bullet points in #13824 and may look at a PR for the schema this week as well but that requires a bit more time than I have now to test. Of course, if anyone else has time, feel free to beat me to it.

[...]

PR #13824, opened 2018-03-20
Fix incorrect description and validation for aliases/URIs per #13815

Mark-H:
### What does it do?
Updates the lexicon description to indicate the limit on aliases isn't technically unlimited, but maximum 191 characters. Also updates validation on the uri field to properly use the same length.

### Why is it needed?
Inconsisten behaviour/error messages as discussed in #13815.

### Related issue(s)/PR(s)
#13815


opengeek (changes requested):
While each resource `alias` is limited to 191 chars, the `uri` field is *not* limited to 191 chars. It is not a varchar type, but a text type in the database.

Mark-H:
@opengeek Per my comment at https://github.com/modxcms/revolution/issues/13815 that should be fixed in a different PR which I've been planning to send, but didn't make time for yet. This just makes sure the error handling is correct and more predictable.

opengeek:
That's not correct though. It can be much longer than 191 characters, regardless of the index prefix.

opengeek:
It just means it will not be sorted properly if multiple uri's go beyond that length (and match the first 191 characters).

PR #13829, opened 2018-03-22
Replace usages of each() to avoid deprecated warnings in PHP 7.2

opengeek:
### What does it do?
Replaces usages of each() in the core (as well as in xPDO and Smarty dependencies).

### Why is it needed?
The `each()` function is deprecated in PHP 7.2 and causes warnings when encountered.

### Related issue(s)/PR(s)
modxcms/xpdo#132
modxcms/xpdo#134


OptimusCrime:
Have you accidentally turned on some code transformation in your IDE?

Why is `$foo[$bar]['baz']` changed to `$foo[ $bar ][ 'baz' ]`? There also seem to be a lot of other things in this PR.

opengeek:
That is from the Smarty update @OptimusCrime — they apparently reformated their entire codebase. And updating Smarty has broken a unit test as well. Loud Sigh.

theboxer (approved):

alroniks (approved):

PR #13834, opened 2018-03-24
[WIP/PoC] YAML config file

OptimusCrime:
### What does it do?
This work-in-progress Proof of Concept uses a YAML config file, instead of the regular _config.inc.php_ file. 

We use the _symphony/yaml_ library for parsing the file. Note that the file type is `.yaml.php`. This was ment to make the file unreadable from the internet, but I realize now that that will not work because we have no opening `<?php`  tag in the file. Not sure how to handle this.

An example of a working config file looks like this:

```
modx:
  db:
    type: mysql
    server: db
    user: revolution
    password: revolution
    connection_charset: utf8
    database: revolution
    table_prefix: modx_
    dns: mysql:host=db;dbname=revolution;charset=utf8
    config_options:
    driver_options:
  last_install_time: 1511612195
  id: modx5a195f23229026.07399157
  https_port: 443
  uuid: 55e54c57-9018-44cd-ba9a-5bfcc0b3bec0
  url_scheme: automatic
  http_host: 192.168.99.100:8000
  disable_cache: false
  paths:
    base: /var/www/html/
    core: /var/www/html/core/
    processors: /var/www/html/core/model/modx/processors/
    connectors: /var/www/html/connectors/
    manager: /var/www/html/manager/
    assets: /var/www/html/assets/
  urls:
    base: /
    connectors: /connectors/
    manager: /manager/
    assets: /assets/
  log_levels:
    fatal: 0
    error: 1
    warn: 2
    info: 3
    debug: 4
```

Note that there is still a lot work that needs to be done on this (installation, upgrade, migration from 2.x etc). I am also considering moving the YAML logic out of the MODX class to avoid bloating it even more.

This PR works if you install the composer depdency and place the config content above in the file `core/config/config.yaml.php` (on traditional installations).

Also note that this PoC is backwards compatible with the old `config.inc.php` type config files.

### Why is it needed?
The benefit of using a YAML config file is that we can avoid some global variables that can cause problems with other systems (we experienced this with phpBB). YAML files are also more readable and easier to edit. This leads to less chance of bugs or problems. We can also mitigate away from having logic in the config files. This should be placed in the actual MODX code, not resolved in the config itself. There is still some work left to do here. The plan is to slim the config down, meaning remove redundant settings.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,8 @@
         "phpmailer/phpmailer": "^6.0",
         "smarty/smarty": "^3.1",
         "james-heinrich/phpthumb": "^1.7",
-        "pelago/emogrifier": "^2.0"
+        "pelago/emogrifier": "^2.0",
+        "symfony/yaml": "^4.0"
     },
     "autoload": {
         "classmap": ["core/xpdo/"]
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -8,6 +8,9 @@
  * file that was distributed with this source code.
  */
 
+use Symfony\Component\Yaml\ParseException;
+use Symfony\Component\Yaml\Yaml;
+
 /**
  * This is the main file to include in your scripts to use MODX.
  *
@@ -460,6 +463,10 @@ class modX extends xPDO {
         }
     }
 
+    private static function configFileExists($configPath, $suffix) {
+        return file_exists($configPath . MODX_CONFIG_KEY . '.' . $suffix);
+    }
+
     /**
      * Load the modX configuration when creating an instance of modX.
      *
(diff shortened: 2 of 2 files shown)


garryn:
I'm not against YAML but I know of a few use cases where this could fail (based on 2.x usage) that could cause issues (specifically around config_options) ... I'd like to see how they could be addressed in this scenario.

bezumkin:
What global variables are could cause problems? I never run in this problem.

I believe that we need to have very serious benefits to do such big change.

OptimusCrime:
@bezumkin the global variables called `$database_name` might crash if you try to integrate MODX with another system. We ran into this issue because of an identical variable in phpBB, but this is just one example.

From Slack just this morning:

>Thanks Mark! -- you set me on the right path.  I looked at config.inc.php no   less than 5 times after I edited it to deal with the new environment, _except_ for where I left off the ~mysite portion the connectors, manager, base and assets urls.     Time to go see the optometrist.

I believe that the config files are made way too complicated as they are structured today. I'll attempt to do another round with this. And again, the plan was to make the two configs interchangeable, which means that if you already have the "old" type of config, then it will work just like it does today. It will not overwrite any of your changes.

bezumkin:
If we will have ability to use old style config - no problem. 

But we can`t require PHP 7.1.3, minimum version now is 5.6.

garryn:
Is 5.6 going to be dependency on Revo 3.0? By the time we get this version out, 5.6 is pretty much (if not) completely EOL ... I'm almost erring on saying PHP7.1 should be minimum - anyone got thoughts?

bezumkin:
@garryn https://w3techs.com/technologies/details/pl-php/all/all

garryn:
@bezumkin PHP 5.6 is EOL for everything by end of this year, we have a decision we can make now to be a leader or a lagger (shedding 5.6 support could be so lucrative down the line) ... just saying, eventually other hosts, and therefore, sites will have to shift at some point

bezumkin:
@garryn No problem, but we need a real reason to increase **minimum** PHP version.

Not because we are very modern and cool (and of course, we are!), but because we really need to force our users to upgrade.

I hope there are libraries to work with YAML on PHP 5.6, for example - https://packagist.org/packages/jms/serializer

Mark-H:
I agree with @bezumkin - increasing the minimum version should be because of a technological need that offers lots of real value.  

After dropping support for 5.3 and 5.4 last year, modmore users predominantly use PHP 5.6 now, and 7.0/7.1/7.2 combined is around 35%. Here's what that looks like in a pie chart (versions ordered clockwise; e.g. orange is 5.5 and light-blue-grey-ish is 7.0):
schermafbeelding 2018-03-26 om 13 52 43

If you'd want to [strictly follow the PHP EOL dates](http://php.net/supported-versions.php), we'd need to actually bump to 7.1+ in December, which is only about 15% of current users. That's way too many people that will need to get their servers upgraded or remain stuck on an old version. 

We can _recommend_ 7.1 but support back to 5.6 of course. Could even include a configuration check that alerts people if they're running below 5.6 to help boost adoption of 7.x.

OptimusCrime:
I know minimum version is 5.6, I did not have the proper requirements in my package.json file. Anyways, it will not be necessary to do any changes to the packages when I remove the yaml library from the code.

mrhaw:
PHP 7 is awesome on Linux hosting but sucks on Windows. There are bugs like this that needs to be fixed: https://bugs.php.net/bug.php?id=74929

alroniks:
Personally, I don't like it. I prefer dotenv way for tweaking current instance. The configuration itself can be improved but without moving to another format of the file.

OptimusCrime:
I agree that yml is was not the best option, but I honestly belive that returning an array, without any logic in the config file, is an improvement.

I will try to look at this as soon as I get the time. My schedule is full most days.

Mark-H:
Do you want to send a new PR when you have a different proposal @OptimusCrime?

OptimusCrime:
Yes. I've rewritten the code now. Opening a new PR.

PR #13843, opened 2018-03-29
Prevent line-wrap in error log

MaartenW:
Makes for easier scanning in the log file.

### What does it do?
Added CSS style to error log textarea

### Why is it needed?
Quickly scanning through the error log in the manager is cumbersome because of line-wraping


bezumkin (approved):

alroniks:
Makes sense for technical part of the system. Logs not actual for plain users, but very useful for developers and indeed to search in logs that looks the same as in the console it's a good idea.

alroniks (approved):

alroniks:
🙏 @opengeek @Mark-H

Mark-H (approved):

PR #13844, opened 2018-03-30
Update dashboards

bezumkin:
### What does it do?
- Ability to customize dashboard: move, resize, change order and size of widgets.
- Improved look and feel
- 2 new widgets: Buttons and Updates
- Dashboard and widgets use so minimum of ExtJS as possible
- Added new column "customizable" to modDashboard
- Added 2 new columns to modDashboardWidget: "properties" and "permission"
- Added 2 new columns to modDashboardWidgetPlacement: "user" and "size"
- Added new aggregate to modUser: "DashboardWidgets"
- Default dashboard and widgets will not be overwritten on upgrade anymore
- Recently edited resources widget now shows all users and requires "view_document" permission

### Why is it needed?
Dashboard of a user can be custimizable, or not. If it is, user can add new widgets, remove old, resize, change size and order by drag-n-drop.
If Dashboard is not customizable, it will be static as admin specified it to be. User will not be able to do anything, as it was before.

Widgets now have JSON properties similar to modUserProfile extended field, that could be used by developers for any puprose. For example, you can change all links in texts in default Buttons widget.
Also widgets can now require any permission to install them on users Dashboard.

### Related issue(s)/PR(s)
N/A


wax100:
When "dashboards" permission is not active, User can delete, change size and move widget. But after page reload we see initial Dashboard.

gpsietzema:
Can we please have this merged as well? This has been tested and we need a 3.x build to continue further testing.

alroniks (approved):

gpsietzema (approved):

wax100:
When U have no errors in **"Configuration check"** widget, it is not shown on dashboard.
When U try to drag and drop Blocks on dashboard, U can get sort, that U do not expected.
It happens because for example, widget **"Configuration check"** has rank 3. 
Second widget has rank 5.
Third widget has rank 6.
When I try to swap Second widget and Third widget in data, sended to server, I get this :
**from 4 
to 5**
And processor save Second widget with rank 5. 
After reload pafe I get initial sorting.
Because rank != position.

rthrash:

https://community.modx.com/t/modx-3-moddashboardwidgetplacement-questions/5136/2

Issue #13846, opened 2018-03-31
Remove type="text/javascript"


ezersky:
### Summary
The validator.w3.org shows an error when connecting scripts using $modx->regClientScript or $modx->regClientStartupScript . 

After removing type="text/javascript" from strings:

- core/model/modx/modmanagercontroller.class.php
- core/model/modx/modmanagercontrollerdeprecated.class.php
- core/model/modx/modx.class.php

we have successful page validation by the validator.w3.org

JoshuaLuckers:
Relevant discussion: https://github.com/modxcms/revolution/pull/13850

Ibochkarev:
HTML5 no longer requires an explicit type statement. Can we delete this?

```html
<!-- HTML4 -->
<script type="text/javascript" src="javascript.js"></script>

<!-- HTML5 -->
<script src="javascript.js"></script>
```

alroniks:
Closed because it merged with https://github.com/modxcms/revolution/pull/15533

PR #13847, opened 2018-03-31
Remove unnecessary type="text/javascript"

Jako:
### What does it do?
Remove an obsolete type="text/javascript" attribute from script tags.

### Why is it needed?
The attribute is not needed anymore in HTML5. See the [answers](https://stackoverflow.com/questions/5265202/do-you-need-text-javascript-specified-in-your-script-tags) on StackOverflow. It should be safe with modern browsers and IE7+.

### Related issue(s)/PR(s)
#13846


alroniks:
Hi @Jako It looks like that your branch is stale. Could you sync PR with 2.x and resolve conflict?

alroniks (dismissed):

Jako:
I have created another PR with the travis tests fixed. Maybe the `type="text/css"` should be removed too with another PR.

PR #13850, opened 2018-04-04
Text javascript

Jako:
### What does it do?
Remove an obsolete type="text/javascript" attribute from script tags.

### Why is it needed?
The attribute is not needed anymore in HTML5. See the [answers](https://stackoverflow.com/questions/5265202/do-you-need-text-javascript-specified-in-your-script-tags) on StackOverflow. It should be safe with modern browsers and IE7+.

### Related issue(s)/PR(s)
#13846

### DIFF
--- a/_build/test/Tests/Model/Filters/modOutputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modOutputFilterTest.php
@@ -774,7 +774,7 @@ goes here'),
         $this->tag->set('name','utp:jsToHead=`'.($plainText ? 1 : 0).'`');
         $this->tag->process();
         if ($addTag) {
-            $value = '<script type="text/javascript" src="'.$value.'"></script>';
+            $value = '<script src="'.$value.'"></script>';
         }
         $this->assertContains($value,$this->modx->sjscripts);
         unset($this->modx->sjscripts[$value]);
@@ -785,7 +785,7 @@ goes here'),
     public function providerJsToHead() {
         return array(
             array('assets/js/hscript.js',true,false),
-            array('<script type="text/javascript" src="assets/js/hscript2.js"></script>',false,false),
+            array('<script src="assets/js/hscript2.js"></script>',false,false),
             array('assets/js/hscript3.js',false,true),
         );
     }
--- a/_build/test/Tests/Model/Filters/modOutputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modOutputFilterTest.php
@@ -745,7 +745,7 @@ goes here'),
         $this->tag->set('name','utp:jsToBottom=`'.($plainText ? 1 : 0).'`');
         $this->tag->process();
         if ($addTag) {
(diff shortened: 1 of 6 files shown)


Jako:
There are two changes in the tests in this PR, since the `type="text/javascript" ` appearance is checked in _build/test/Tests/Model/Filters/modOutputFilterTest.php

opengeek:
If you fixed the tests, why are they still failing?

opengeek:
1) modOutputFilterTest::testJsToHead with data set #0 ('assets/js/hscript.js', true, false)
    Failed asserting that an array contains '<script type="text/javascript" src="assets/js/hscript.js"></script>'.
    /home/travis/build/modxcms/revolution/_build/test/Tests/Model/Filters/modOutputFilterTest.php:779

I think you just missed one.

opengeek:
I have a couple of concerns. First, you modified smarty files. These should not be modified directly. Second, while I have no concern with removing this from the manager UI, you are changing the entire way registering scripts works in the MODX core API, which means that this will change behavior for existing sites/extras that use the `regClient*Script()` methods. While we can assume this is fine for the manager UI, I'm not convinced this is a good idea to modify for front-ends that might need to support older browsers for one reason or another. I'd like some feedback on these concerns.

opengeek (commented):
I have a couple of concerns. First, you modified smarty files. These should not be modified directly. Second, while I have no concern with removing this from the manager UI, you are changing the entire way registering scripts works in the MODX core API, which means that this will change behavior for existing sites/extras that use the regClient*Script() methods. While we can assume this is fine for the manager UI, I'm not convinced this is a good idea to modify for front-ends that might need to support older browsers for one reason or another. I'd like some feedback on these concerns.

Jako:
I will remove the smarty changes.

The PR fixes a validating issue in #13846 which is a frontend issue with regClient\*Script() methods. Maybe we need a doctype context/system setting, that switches the output of `type="text/javascript"` in the regClient\*Script() methods on for XHTML and off for all other doctypes.

opengeek:
Maybe. I'd honestly like to see those methods deprecated and removed for 3.x.

opengeek:
Did not mean to close this, sorry.

gpsietzema:
Also did a test and still see a lot of text/javascripts floating around...

gpsietzema:
@opengeek @Jako what do you want to do with this?

> Maybe. I'd honestly like to see those methods deprecated and removed for 3.x.

Are you talking about the manager text/javascript mentions or something else?

bezumkin (changes requested):
I believe we should have a system setting to enable this new behaviour in 2.7 or target it to 3.x only.

There are a lot of legacy corporate websites and we can't just change script tags because of w3c validator don't like it.

OptimusCrime:
Do we really need a system setting for this? We already have a million system settings. For 3.x we are allowed to break stuff. Making MODX do HTML5 seem like a logical step in the right direction.

Jako:
This PR was closed last year.

Issue #10201, opened 2013-09-15
Support markdown in package manager readme files


exside:
exside created Redmine issue ID 10201

As most projects are hosted on github it would be very convenient if we could just copy and paste the content of the readme.md file to the core/components/packagename/docs/readme.txt file and have it displayed as nicely as on github. This would save time for many of us I think =).

I suck at extjs but maybe this would be helpful: http://www.sencha.com/forum/showthread.php?254825-ExtJS-and-Markdown

alroniks:
It should be easy to implement it in 3.0 using any markdown library from packagist via composer.

alroniks:
Closing as PR is merged to 3.x

PR #13853, opened 2018-04-09
Add support of Markdown in package attributes

bezumkin:
### What does it do?
This adds erusev/parsedown as a dependency and enable markdown in package attributes.

### Why is it needed?
To make MODX better

### Related issue(s)/PR(s)
#10201

screen shot 2018-04-09 at 11 26 42
screen shot 2018-04-09 at 11 26 49


core/model/modx/processors/workspace/packages/getattribute.class.php
```diff
@@ -39,8 +39,12 @@ public function initialize() {
     public function process() {
         $attributes = array();
         $attributesToGet = explode(',',$this->getProperty('attributes',''));
+        $Parsedown = new Parsedown();
         foreach ($attributesToGet as $attribute) {
-            $attributes[$attribute] = $this->transport->getAttribute($attribute);
+            $data = $this->transport->getAttribute($attribute);
+            $attributes[$attribute] = in_array($attribute, ['changelog', 'license', 'readme'])
+                ? $Parsedown->text(htmlentities($data, ENT_COMPAT, 'UTF-8'))
```
alroniks: I would add such bit mask `ENT_COMPAT | ENT_HTML5`

core/model/modx/processors/workspace/packages/getattribute.class.php
```diff
@@ -39,8 +39,12 @@ public function initialize() {
     public function process() {
         $attributes = array();
         $attributesToGet = explode(',',$this->getProperty('attributes',''));
+        $Parsedown = new Parsedown();
```
alroniks, line 42: As we use the external tool to improve readability we also can add a little bit more security. 
`$parsedown->setSafeMode(true);` can help to avoid XSS and other bad things, because packages from repository checked but they still unsafe user input.

alroniks (changes requested):

core/model/modx/processors/workspace/packages/getattribute.class.php
bezumkin: Markdown don't work after `ENT_HTML5`
screen shot 2018-04-11 at 11 52 47

I can use `ENT_HTML401`

alroniks: MODX 3 requires php 5.6+. In this case it can be simplified to just htmlentities() because `ENT_COMPAT | ENT_HTML401` is default value. And since 5.6 encoding param use value of conf option `default_charset` by default. And option set as "UTF-8" by default.

More details by links:
http://php.net/manual/en/function.htmlentities.php
http://php.net/manual/en/ini.core.php#ini.default-charset

bezumkin: Updated

alroniks (approved):

core/model/modx/processors/workspace/packages/getattribute.class.php
Jako: I think there can't be any security issues like XSS since the text is encoded with htmlentities (there are no html tags visible after that, since <> are encoded). Or does Parsedown replace those entities inside?

opengeek (commented):
So if you submit packages with markdown, how are these going to look in package management for older releases of MODX?

bezumkin:
@opengeek Almost all of them already use markdown, and it is okey right now. For example [here is the Gallery](https://github.com/modxcms/Gallery/blob/develop/core/components/gallery/docs/changelog.txt)

opengeek (approved):

Issue #13641, opened 2017-10-16
Feature Request: Add Created and Modified Columns to Database Site Tables


playcock:
### Summary
Database table columns to add: Created, Modified

### Step to reproduce
Creating/Saving a Resource or Element in the Manager

### Expected behavior
I would like to see created and a modified columns on the following database tables when creating or updating a Resource or Element from within the Manager, please. 

Both columns would be the timestamp type. 
The created column's Default would be CURRENT_TIMESTAMP
The modified column would contain the date and time whenever the Resource or Element was modified.
 
This would help me track when a resource or element is created and/or modified.

modx_site_content
modx_site_htmlsnippets
modx_site_plugins
modx_site_plugin_events
modx_site_snippets
modx_site_templates
modx_site_tmplvars
modx_site_tmplvar_access
modx_site_tmplvar_contentvalues
modx_site_tmplvar_templates

### Environment
MODX Revolution 2.5+

playcock:
What I am looking for is a readable date.

Mark-H:
For tracking modifications, the manager log is useful to look at. However I'd also like to see more created/edited timestamps and user IDs for accountability. 

That does beg the question of where those fields would be exposed to the user, otherwise they may have little value. Also what constitutes an edit - should interactions via xPDO cause the edited fields to be updated, or only though the manager?

OptimusCrime:
I agree with updated timestamp, created timestamp, created by user and updated by user for snippets, templates, chunks and plugins. But I fail to see the use of this for the rest. Like @Mark-H already said, the manager log can be used to some extent for this.

isaacniebeling:
`modx_site_content` already has timestamps... I don't think the others do (though `modx_site_tmplvar_contentvalues` could likely be assumed to be the same as `modx_site_content`). Probably related to #13626...

I'd say _elements_ make a lot of sense, but `modx_site_tmplvar_access` and `modx_site_tmplvar_templates` change along with `modx_site_tmplvars`, `modx_site_plugin_events` changes with `modx_site_plugins`... you _could_ manually change those via xpdo, I suppose, but... that seems like a super-duper-edgy-edge-case

playcock:
I just like a peace of mind when pushing data from those tables from sandboxes to dev to staging to production with multiple people working on the web sites. The people working on the modx web sites could give me a list of resources, templates, snippets, and chunks that they have worked on, but we're people and make mistakes. 

[...]

Issue #2027, opened 2008-09-27
Add created/modified timestamps and authorstamps to all elements


modxbot:
rethrash created Redmine issue ID 2027

The following elements should have created/modified timestamps and datestamps. This would provide a way to audit when the last changes occurred and by whom for tracking down issues and as a starting point for a simple versioning system. Would also mirror the content field. chunks
snippets
TVs
plugins
modules
manager users
web users

adamwintle:
I don't think this is needed in the core, VersionX can handle it, plus you can look in the Manager Log to see whats been changed - perhaps a better Manager Log UI (wither better filters, etc)

However if this is implemented; would this record every modified timestamp or just the latest change? 

Should this be bumped up to 2.6.0 milestone?

sottwell:
The problem with depending on the manager activity log is that anyone can clear it. There needs to be a way to save it to file before it will be dependable.

adamwintle:
@sottwell That is a good point.  Recently I was thinking that it could be useful if the Manager Log was a .log file on the filesystem instead of in the database.  It could be in a JSON or YAML formate.

This way the Manager Log could be backed up to a secure and more permeant location (alongside other server logs) using a CRON job or other task runner.

Currently when I am running a content audit on a large site I rely on VersionX and the Manager Logs, but both of these can quite easily be manipulated.  

If modified timestamp is added to all elements in the core it should be handled as a security feature with additional ACL/privileges integrations.

OptimusCrime:
Closing as a duplicate of #13641, although this issue is _way_ older, the other one contains more information. Clearly this has been a requested feature for some time now.

OptimusCrime:
@modxbot close

PR #13857, opened 2018-04-10
Add created and edited columns to elements

bezumkin:
### What does it do?
- Added 4 new columns (createdby, createdon, editedby, editedon) to all modElement descendants: to modChunk, modPlugin, modSnippet, modTemplate, modTemplateVar

- Prevent overwrite of elements by checking createdon value before save

### Why is it needed?
To protect users from overwriting elements. Also now we would be able to get actual creators and editors of elements with MODX API.

### Related issue(s)/PR(s)
#13770 #13626 #13641 #2027

screen shot 2018-04-10 at 11 15 48

This PR changes javascript files, so please, don`t forget to disable `compress_js` system setting before testing.

### DIFF
--- a/core/model/modx/mysql/modelement.map.inc.php
+++ b/core/model/modx/mysql/modelement.map.inc.php
@@ -74,17 +74,65 @@ $xpdo_meta_map['modElement']= array (
       'default' => 0,
     ),
   ),
-  'indexes' =>
+  'indexes' => 
   array (
-    'editedon' =>
+    'createdby' => 
+    array (
+      'alias' => 'createdby',
+      'primary' => false,
+      'unique' => false,
+      'type' => 'BTREE',
+      'columns' => 
+      array (
+        'createdby' => 
+        array (
+          'length' => '',
+          'collation' => 'A',
+          'null' => false,
+        ),
+      ),
+    ),
+    'createdon' => 
+    array (
+      'alias' => 'createdon',
+      'primary' => false,
+      'unique' => false,
+      'type' => 'BTREE',
+      'columns' => 
+      array (
+        'createdon' => 
+        array (
+          'length' => '',
+          'collation' => 'A',
+          'null' => false,
+        ),
+      ),
+    ),
+    'editedby' => 
+    array (
+      'alias' => 'editedby',
+      'primary' => false,
+      'unique' => false,
+      'type' => 'BTREE',
+      'columns' => 
+      array (
+        'editedby' => 
+        array (
+          'length' => '',
+          'collation' => 'A',
+          'null' => false,
+        ),
+      ),
+    ),
+    'editedon' => 
     array (
       'alias' => 'editedon',
       'primary' => false,
       'unique' => false,
       'type' => 'BTREE',
-      'columns' =>
+      'columns' => 
       array (
-        'editedon' =>
+        'editedon' => 
         array (
           'length' => '',
           'collation' => 'A',
--- a/core/model/modx/mysql/modelement.map.inc.php
+++ b/core/model/modx/mysql/modelement.map.inc.php
@@ -14,11 +14,47 @@ $xpdo_meta_map['modElement']= array (
   ),
   'fields' => 
   array (
+    'createdby' => 0,
+    'createdon' => 0,
+    'editedby' => 0,
+    'editedon' => 0,
     'source' => 0,
     'property_preprocess' => 0,
   ),
   'fieldMeta' => 
   array (
+    'createdby' => 
+    array (
+      'dbtype' => 'int',
+      'precision' => '10',
+      'phptype' => 'integer',
+      'null' => false,
+      'default' => 0,
+    ),
+    'createdon' => 
+    array (
+      'dbtype' => 'int',
+      'precision' => '20',
+      'phptype' => 'timestamp',
+      'null' => false,
+      'default' => 0,
+    ),
+    'editedby' => 
+    array (
+      'dbtype' => 'int',
+      'precision' => '10',
+      'phptype' => 'integer',
+      'null' => false,
+      'default' => 0,
+    ),
+    'editedon' => 
+    array (
+      'dbtype' => 'int',
+      'precision' => '20',
+      'phptype' => 'timestamp',
+      'null' => false,
+      'default' => 0,
+    ),
     'source' => 
     array (
       'dbtype' => 'int',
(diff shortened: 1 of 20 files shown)


gpsietzema:
Note to self and anyone else who has issues with missing fields "[2] => Unknown column 'modTemplate.createdby' in 'field list'": do not forget to execute setup after checking out.

I'll now do real testing :-)

gpsietzema:
Successfully tested with templates, chunks, snippets, plugins and template variables. Everything works as expected and I didn't see any issues in either the php/modx error log.

Tested on PHP 7.2 and 5.6.

gpsietzema (dismissed):
Successfully tested with templates, chunks, snippets, plugins and template variables. Everything works as expected and I didn't see any issues in either the php/modx error log.

Tested on PHP 7.2 and 5.6.

core/model/modx/mysql/modelement.map.inc.php
```diff
@@ -38,6 +74,25 @@
       'default' => 0,
     ),
   ),
+  'indexes' =>
```
alroniks: I think it makes sense to add indexes for fields `createdby` and `editedby` for further use in widgets and extras when elements should be selected by author or editor.

bezumkin: I think, right now it have no sense, because it is not used anywhere.

Otherwise we could index everything just in case.

alroniks: No need index all fields but for these fields this is obvious. As it works for resources already. 
Implementing it later will require additional work for reviewing, merging and releasing a new version that takes time and resources.

bezumkin: Okay, no problem.

alroniks (dismissed):

gpsietzema (dismissed):

opengeek (dismissed):
NEVER merge the base branch back into your PRs. This needs to be corrected. If you want to update your PR based on changes to the base branch since you submitted, you MUST rebase on top of the base and force push the new commits. Otherwise, we create a conflict nightmare.

bezumkin:
@opengeek Fixed

bezumkin:
@Alroniks & @gpsietzema Sorry, I rebased this PR to make it easier to merge.

Please, approve it again.

gpsietzema (dismissed):

bezumkin:
@opengeek Can you check this PR again, please?

opengeek:
If someone can do it before I have the chance, we need to know how these changes affect static elements. Currently, if a static element's source file is changed, the content is compared to the database and updated in the database if it does not match. This comparison happens every time it is rendered, which is anything but ideal, but I would imagine that this could cause some unexpected behavior now in conjunction with the changes in this PR.

bezumkin:
@opengeek When content of element is updated from static files, it will save new `editedon` date.

So, if you are editing a chunk in manager and somebody uploaded new static file for it, and then render it on frontend - you will not be able to save your changes and overwrite this chunk content.

opengeek:
@bezumkin this is not about updating in the manager. Every time a static element is executed/processed and not cacheable or not yet cached, the content is compared directly and if the static source is different from the current database content, the database is updated. I see no diff in this PR which touches that code.

bezumkin:
@opengeek did you seen this todays commit? https://github.com/modxcms/revolution/commit/ecef22cd1df91c5ce56a5b4b7fff0ecc7d87ea65

opengeek:
No, sorry, I did not see there was a new commit.

alroniks:
I've tested and fixed one minor issue (`modScript` should be skipped because it does not have a table in the database and when it used it produces various errors in installing/upgrading report) and it seems that it works except one case that related to your comment @opengeek:
> Currently, if a static element's source file is changed, the content is compared to the database and updated in the database if it does not match. This comparison happens every time it is rendered, which is anything but ideal, but I would imagine that this could cause some unexpected behavior now in conjunction with the changes in this PR.

It does not update `editedon` field in the case when the content was changed in the file and came to the database from the file. But, code that checks if a code was changed works fine.
For example, I've created the static chunk, open the page with the chunk in the manager, then updated file directly and update page where chunk used. On the page, I saw new content of the chunk, but when I tried to change the chunk via manager (without refreshing the page with opened chunk) I've got the error that chunk has a new version.

So if it is acceptable, it can be merged I guess.

alroniks (approved):

opengeek:
modScript represents modSnippet and modPlugin, so I'm confused what you mean about skipping those.

opengeek:
Also, it would be preferable for any updates to the records to get the updated timestamp before we merge this.

alroniks:
> modScript represents modSnippet and modPlugin, so I'm confused what you mean about skipping those.

@opengeek Indeed, but it gets data like a tree with modScript included into the list, but it cannot add new fields to this class directly because we don't have real data with this type in the datatabe, only nested modSnippet and modChunk. It produces errors in the report list after upgrade/install finished.

alroniks:
> Also, it would be preferable for any updates to the records to get the updated timestamp before we merge this.

I'll have a look and send fix for it.

opengeek:
It should work fine with the fields added in modScript. If not, something is wrong. I'll have to dig into this today and figure it out.

alroniks:
Such errors in the report:
```
Error updating table for class modScript
Array
(
[0] => 42S02
[1] => 1146
[2] => Table 'modx.modx_site_script' doesn't exist
)
```

ilyautkin:
PR is tested. Standard elements and static elements tracks their changes. The message is displayed where it is needed.

ilyautkin:
Bug found on work with static elements.

1. Create static element (chunk or snippet). Close manager tab (window).
2. Open frontend - field `editedon` will not changed in database
3. Change element in IDE (not in manager)
4. Open frontend - changes will appeared, but field `editedon` will NOT changed
5. Open element in manager. Do not save - only open. Field `editedon` WILL changed

Mark-H:
@bezumkin Any thoughts on the issue reported by @ilyautkin?

bezumkin:
@Mark-H There are a lot of weird stuff.

0. Everything is works fine with clear cache.

1. When an element is in cache - it is always `new`, because of [modParser::getElement method](https://github.com/modxcms/revolution/blob/21fe0868cd04f706b6f27d1f11e105348848331e/core/model/modx/modparser.class.php#L568-L583), so [this condition](https://github.com/modxcms/revolution/pull/13857/files#diff-05daf9f3d1cbef3a884c85b66f20ac82L408) is always false.

2. If we will add `$element->_new = false;` to `getElement`, it could't be saved because of related `modSource`. If we will add `$source->_new = false;` then it will be saved each time, because new content will not be in cache of resource (I don't know why and really don't want to dig deeper in cache system).

So, I see here only 2 options:
1. Leave it as is and hope that people with static elements will resolve their editing conflicts in VCS.
2. Make raw SQL query to update `editedon` in table, but it looks a little dirty.

My choice is number 1.

Ibochkarev:
@bezumkin Hello! Please continue to work on this fix. Thank

JoshuaLuckers:
Thins is a more then welcome improvement. I do wonder if it's really a good idea to prevent users from saving when MODX thinks there is newer content in the database.

Mark-H:
Past the cut-off for alpha1 so pushing forward. Perhaps someone else wants to take the lead on getting this updated for inclusion in 3.0.

Issue #13858, opened 2018-04-10
Remove obsolete system settings about compressing 


alroniks:
### Summary
Need to clean up system settings and remove `compress_js_max_files` and `manager_js_zlib_output_compression` as they not used in the system. 

### Step to reproduce
n/a

### Observed behavior
System settings exist in the system, has translates string but never used.

### Expected behavior
Unused system settings should be removed including their lexicons.

### Environment
3.x

Lelated discussion https://github.com/modxcms/revolution/pull/13785

alroniks:
Closing issue as PR is merged.

PR #13859, opened 2018-04-10
Remove obsolete system settings

alroniks:
### What does it do?
Removes obsolete system settings.

### Why is it needed?
They not used in the system anymore and are messing up the code.

### Related issue(s)/PR(s)
#13858

### DIFF
--- a/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
+++ b/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
@@ -3,18 +3,21 @@
  * Remove obsolete system settings about compression
  */
 
-/** @var modSystemSetting $compress_js_max_files */
-$compress_js_max_files = $modx->getObject('modSystemSetting', [
-    'key' => 'compress_js_max_files'
-]);
-if ($compress_js_max_files) {
-    $compress_js_max_files->remove();
-}
+$settings = ['compress_js_max_files', 'manager_js_zlib_output_compression'];
 
-/** @var modSystemSetting $manager_js_zlib_output_compression */
-$manager_js_zlib_output_compression = $modx->getObject('modSystemSetting', [
-    'key' => 'manager_js_zlib_output_compression'
-]);
-if ($manager_js_zlib_output_compression) {
-    $manager_js_zlib_output_compression->remove();
+foreach ($settings as $key) {
+    /** @var modSystemSetting $setting */
+    $setting = $modx->getObject('modSystemSetting', ['key' => $key]);
+    if ($setting instanceof modSystemSetting) {
+        if ($setting->remove()) {
+            $this->runner->addResult(modInstallRunner::RESULT_SUCCESS,
+                '<p class="ok">' . $key . ' System Setting removed.</p>');
+        } else {
+            $this->runner->addResult(modInstallRunner::RESULT_FAILURE,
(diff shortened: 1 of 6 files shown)


Mark-H:
They should probably also be removed in an upgrade script in the setup - just removing them from the build is not enough to make them actually go away.

bezumkin (changes requested):
I think we also need to delete these settings when updating the site.

alroniks:
Good catch @Mark-H. I will update the pr.

alroniks:
@bezumkin @Mark-H PR is updated. Please, have a look.

bezumkin (dismissed):

opengeek (dismissed):

alroniks:
@opengeek @bezumkin Excuse me for troubling but I updated upgrading script to avoid silence mode. Now it says what was removed and errors in the case when system settings cannot be removed.

setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
```diff
@@ -0,0 +1,28 @@
...
+
+foreach ($settings as $key) {
+    /** @var modSystemSetting $setting */
+    $setting = $modx->getObject('modSystemSetting', ['key' => $key]);
+    if ($setting instanceof modSystemSetting) {
+        if ($setting->remove()) {
+            $this->runner->addResult(modInstallRunner::RESULT_SUCCESS,
+                sprintf($messageTemplate, 'ok', $this->install->lexicon('system_setting_cleanup_success', ['key' => $key])));
+        } else {
+            $this->runner->addResult(modInstallRunner::RESULT_FAILURE,
```
opengeek: I don't think we want this be a failure if it doesn't remove a system setting. Do we?

opengeek: I think warning there, and nothing if the setting doesn't exist. A warning if it has already been removed is unnecessary in my opinion.

alroniks: Yes, make sense. I updated this part.

gpsietzema (commented):
I can't seem to get this to work on PHP7.2/Mariadb 10.2.14 (mysql 5.5.5).

I do get the message in the setup that the 2 settings have been removed, but they are still there.

Another thing: shouldn't compress_js also be removed. And what about compress_css, we are keeping that one I think?

alroniks:
@gpsietzema Did you try to rebuild core package after checkout? Because it removes settings during the upgrade but if core package is not rebuilt it sets up these settings from the package again.

gpsietzema (approved):
@Alroniks You are right, rebuilding the core did the job!

Still wondering about compress_js settings. Shouldn't it be removed as well?

bezumkin:
@gpsietzema `compress_js` is still in use, because we have [minified version](https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/modx.jsgrps-min.js)

gpsietzema:
@bezumkin but it was removed from the installer somehow... or is that totally unrelated?

bezumkin:
@gpsietzema https://github.com/modxcms/revolution/pull/13785#issuecomment-380115763 - looks like we use minified versions by default, but can disable it for development.

So, this settings don`t need in setup.

Mark-H (approved):

PR #13861, opened 2018-04-11
Fix installation of transport packages with setup options

Mark-H:
### What does it do?

Make sure setup options are sent to the processor when installing a package.

### Why is it needed?

In #13813 I removed the `va` variable from being sent to the processor as that contained an extjs config object which was tripping over mod_security rules. But I failed to notice that setup options from a package were also being sent in this variable, so in 2.6.2 setup options were no longer being sent to the server, breaking installation (or preventing easy set-up) of some packages.

To remove the ambiguity of the `va` variable I've updated the setup options install method to send the same variables to install, with the `options` as a new third parameter instead. That way it can be treated properly. I've also added extensive comments for the next poor sap to explore into the depths of the package installation process.

### Related issue(s)/PR(s)

#13813

### DIFF
--- a/manager/assets/modext/workspace/package.containers.js
+++ b/manager/assets/modext/workspace/package.containers.js
@@ -81,26 +81,41 @@ Ext.extend(MODx.panel.Packages,MODx.Panel,{
         Ext.each(this.buttons, function(btn){ Ext.getCmp(btn.id).hide(); });
     }
 
-	,install: function(va){
+    /**
+	 *
+     * @param va ExtJS instance of the button that was clicked
+     * @param event The Ext.EventObjectImpl from clicking the button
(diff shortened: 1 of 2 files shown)


alroniks:
Do you remember an example of such packages? I had own but removed setup options from it a few weeks ago.

Mark-H:
modmore has a few: Commerce, ContentBlocks, ClientConfig (v2), SimpleCart. Public ones I can think of on the spot are FormIt (newsletter signup - might be hard to test if it works), MIGX (menu location).

ghost:
CssSweet - although that's only acknowledgement of a notice of breaking changes in versions if updating.

Jako (approved):

alroniks (approved):

PR #13863, opened 2018-04-11
Fix the incorrect sorting in Dashboard widgets, when empty widget is …

wax100:
### What does it do?
Show empty widget as empty hidden div, when output of widget is empty

### Why is it needed?
Because current sorting od widgets work incorrect, because of div position != widget rank

### Related issue(s)/PR(s)
[#13844](https://github.com/modxcms/revolution/pull/13844)

### DIFF
--- a/core/model/modx/moddashboardwidget.class.php
+++ b/core/model/modx/moddashboardwidget.class.php
@@ -3,9 +3,6 @@
  * @package modx
  * @subpackage mysql
  */
-use xPDO\Om\xPDOSimpleObject;
-use xPDO\xPDO;
-
 /**
  * Abstraction of a Dashboard Widget, which can be placed on Dashboards for welcome screen customization.
  *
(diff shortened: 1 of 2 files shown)


core/model/modx/moddashboardwidget.class.php
```diff
@@ -311,26 +313,14 @@ public function getFileChunk($tpl,array $placeholders = array()) {
...
-            if (is_array($scriptProperties)) {
-                extract($scriptProperties, EXTR_SKIP);
-            }
-
-            return eval($content);
-        };
-
-        return $closure([
+    public function renderAsSnippet($content = '') {
+        if (empty($content)) $content = $this->widget->get('content');
```
alroniks, line 319: Please, do not use short syntax in `if` statements (PSR requires brackets). Also, it may produce elusive issues in the future.

core/model/modx/moddashboardwidget.class.php
```diff
@@ -311,26 +313,14 @@ public function getFileChunk($tpl,array $placeholders = array()) {
...
+        $content = str_replace(array('<?php','?>'),'',$content);
```
alroniks, line 320: PSR is not strictly required in MODX but I guess better to follow the common standard of PHP development. Arguments should be split by space symbol.

core/model/modx/moddashboardwidget.class.php
```diff
@@ -311,26 +313,14 @@ public function getFileChunk($tpl,array $placeholders = array()) {
...
+        $closure = create_function('$scriptProperties','global $modx;if (is_array($scriptProperties)) {extract($scriptProperties, EXTR_SKIP);}'.$content);
```
alroniks, line 321: Why here `$scriptProperties` variable in quotes? This code will fail. Variable dereferencing is possible only in double-quoted strings (ex. "$var").
Second issue - `create_function` is deprecated since 7.2 (http://php.net/manual/en/function.create-function.php).

alroniks (changes requested):

bezumkin:
Closing in favour of #13864

PR #13864, opened 2018-04-11
New config_check processor

bezumkin:
### What does it do?
- Refactored config_check file to processor
- Removed unneeded comments in the head of widgets

### Why is it needed?
It fixes possible bug with sorting of widgets when there are no problems with config_check and widget not showing.

### Related issue(s)/PR(s)
#13863 

screen shot 2018-04-11 at 15 28 24
screen shot 2018-04-11 at 15 29 15


core/model/modx/processors/system/config_check.class.php
```diff
@@ -0,0 +1,232 @@
...
+        ]);
+        if ($tags) {
+            $this->warnings[$this->modx->lexicon('configcheck_allowtagsinpost_context_enabled')] =
+                $this->modx->lexicon('configcheck_allowtagsinpost_context_enabled_msg');
+        }
+    }
+
+}
+
+return 'modSystemConfigCheck';
```
alroniks, line 232: We can use `modSystemConfigCheck::class` since 5.6 instead of strings. PHPStorm may show error when any mistake was done in the name accidentally (very rare case but it may happen).

bezumkin: Nice!

opengeek (approved):

alroniks (approved):

PR #13865, opened 2018-04-11
[3.x] Remove @EVAL binding from TVs to prevent arbitrary code execution by admins

sepiariver:
### What does it do?
Take advantage of major version to remove dangerous feature.

### Why is it needed?
The `@EVAL` binding allows any User with Manager access, that has permissions to CREATE or UPDATE Template Variables, to execute arbitrary php code, yet creating or updating Template Variables is an action that non-admin, non-developer users may commonly need to undertake.

### Related issue(s)/PR(s)
Branch 3.x-tveval

### DIFF
--- a/setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
+++ b/setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * Common upgrade script to clean up deprecated system settings for TV eval feature.
+ *
+ * @var modX
+ *
+ * @package setup
+ */
+
+$deprecatedSetting = $modx->getObject('modSystemSetting', array(
+    'key' => 'allow_tv_eval'
+));
+if ($deprecatedSetting instanceof modSystemSetting) {
+    if ($deprecatedSetting->remove()) {
+        $this->addResult(modInstallRunner::RESULT_SUCCESS,'<p class="ok">allow_tv_eval System Setting removed.</p>');
+    } else {
+        $this->addResult(modInstallRunner::RESULT_FAILURE,'<p class="notok">allow_tv_eval System Setting could not be removed.</p>');
+    }
+} else {
+    $this->runner->addResult(modInstallRunner::RESULT_WARNING,'<p class="warning">allow_tv_eval System Setting was not found.</p>');
+}
--- a/setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
+++ b/setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
@@ -6,16 +6,22 @@
  *
  * @package setup
  */
+$settings = [
+    'allow_tv_eval'
+];
 
-$deprecatedSetting = $modx->getObject('modSystemSetting', array(
-    'key' => 'allow_tv_eval'
-));
-if ($deprecatedSetting instanceof modSystemSetting) {
-    if ($deprecatedSetting->remove()) {
-        $this->addResult(modInstallRunner::RESULT_SUCCESS,'<p class="ok">allow_tv_eval System Setting removed.</p>');
-    } else {
-        $this->addResult(modInstallRunner::RESULT_FAILURE,'<p class="notok">allow_tv_eval System Setting could not be removed.</p>');
+$messageTemplate = '<p class="%s">%s</p>';
+
+foreach ($settings as $key) {
+    /** @var modSystemSetting $setting */
+    $setting = $modx->getObject('modSystemSetting', ['key' => $key]);
+    if ($setting instanceof modSystemSetting) {
+        if ($setting->remove()) {
+            $this->runner->addResult(modInstallRunner::RESULT_SUCCESS,
+                sprintf($messageTemplate, 'ok', $this->install->lexicon('system_setting_cleanup_success', ['key' => $key])));
+        } else {
+            $this->runner->addResult(modInstallRunner::RESULT_WARNING,
+                sprintf($messageTemplate, 'warning', $this->install->lexicon('system_setting_cleanup_failure', ['key' => $key])));
+        }
     }
-} else {
-    $this->runner->addResult(modInstallRunner::RESULT_WARNING,'<p class="warning">allow_tv_eval System Setting was not found.</p>');
 }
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -22,9 +22,10 @@ modext.js
 .settings/
 nbproject/
 
-# ignore php and txt files beginning with test in root directory
+# ignore dir, php and txt files beginning with test in root directory
 /test*.php
 /test*.txt
+/test
```
opengeek: I don't necessarily mind this being added to the .gitignore file, but it does not belong in this PR with unrelated changes.

opengeek (changes requested):
An update script to remove the obsolete setting would be appropriate to include with this PR.

opengeek:
I think this is a grand idea.

.gitignore
sepiariver: BAh. Sorry...snuck in there.

sepiariver:
Questions @opengeek 

1. What are best practices for including an update script as you suggest? `_build/resolvers/`?
2. Is it preferred to squash commits in PRs like this?

opengeek:
1. Upgrade scripts are in `setup/includes/upgrades/` by version.
2. You can squash now that you introduced another commit to remove the changes in the first commit, then `git push --force` to your branch. Another approach would have been to rebase the branch (`git rebase -i HEAD~2` would have allowed you to remove the first commit from your branch interactively), removing the first commit and then doing the same `git push --force`. Both approaches are fine.

OptimusCrime:
Just curious here, with the removal of `@EVAL` how are we supposed to do similar functionality in 3.x in a "safe" way?

sepiariver:
@OptimusCrime one option is to use the @CHUNK binding, which can in turn call a Snippet. In this scenario, a User may be granted permission to edit TVs, but not to edit Snippets, and they would only be able to execute Snippets already installed/prepared.

Alternatively the @SELECT binding provides a very flexible way to fetch input options for TVs. It does grant access to read any value in the database, but at least you cannot execute arbitrary code with it.

Note: _implementation_ of @EVAL in a given site is not required to exploit it—the mere existence of the feature in core, exposes a substantial vulnerability, albeit scoped to manager users.

setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @package setup
+ */
+
+$deprecatedSetting = $modx->getObject('modSystemSetting', array(
+    'key' => 'allow_tv_eval'
+));
+if ($deprecatedSetting instanceof modSystemSetting) {
+    if ($deprecatedSetting->remove()) {
+        $this->addResult(modInstallRunner::RESULT_SUCCESS,'<p class="ok">allow_tv_eval System Setting removed.</p>');
```
alroniks: Here should be `$this->runner->addResult` otherwise it produces errors about an undefined method of the class.
But I would prefer also refactor a bit `modInstallVersion` class to have method `getRunner` there.

setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
```diff
@@ -0,0 +1,21 @@
...
+    } else {
+        $this->addResult(modInstallRunner::RESULT_FAILURE,'<p class="notok">allow_tv_eval System Setting could not be removed.</p>');
```
alroniks: The same as above.

setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
```diff
@@ -0,0 +1,21 @@
...
+    }
+} else {
+    $this->runner->addResult(modInstallRunner::RESULT_WARNING,'<p class="warning">allow_tv_eval System Setting was not found.</p>');
```
Alroniks: The same as above.

alroniks (changes requested):

setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
alroniks, line 1: My PR is not merged yet but after I would suggest using one common upgrading script for removing obsolete system settings that never used. Example here https://github.com/modxcms/revolution/pull/13859/files#diff-ce54080b4829628803023b4d30b21700

sepiariver: Hi @Alroniks so would you suggest adding setting keys to your script once it’s merged therefore removing the script from this PR entirely? Or duplicate the script in both PRs?

opengeek: No, it should not use one common upgrade script. Dependencies between PRs is not a good idea. If one needs to be reverted, then that makes the process more complicated.

alroniks: It will violate the DRY principles but make sense according to maintenance. But at least script in this PR should follow your recommendations that you left in my PR.

opengeek: Agreed.

sepiariver:
> Merging can be performed automatically once the requested changes are addressed.

I think all changes are addressed now @opengeek @Alroniks

alroniks (dismissed):

Mark-H (dismissed):
LGTM!

alroniks:
We needed that @opengeek change his opinion. :)

opengeek (dismissed):

alroniks:
@sepiariver Could you add `3.0.0-remove-tv-eval-system-setting.php` file into `3.0.0-pl.php` file in both `mysql` and `sqlsrv` folders, please? Code is OK, but upgrading does not remove system setting.

sepiariver:
OK @Alroniks that's done but in doing so, some conflicts emerged, which I tried to resolve, and now there's a Merge branch 3.x commit back into this branch. Not sure if we're ok with this... or how you want to handle it?

opengeek:
No, you need to remove the merge commit.

opengeek:
In general, don't concern yourself with conflicts unless they are on a much larger scale or a reviewer asks you to rebase your work on top of the latest branch.

sepiariver:
Thanks @opengeek I removed the merge commit.

Mark-H (approved):

alroniks (approved):

alroniks:
Actually, it checked many times. Last change that was required is to add missed scripts to db driver specific updater files. Also, there are conflicts, but as it was discussed before it may be fixed during the merge. So I am going to merge it right now.

PR #13869, opened 2018-04-13
Add template icon for resources in serch results.

SintezCode:
### What does it do?
Add template icon for resources in uberbar serch results.

### Why is it needed?
If we have some resources with same names, but different templates, in search results we see its icons to distinguish them.

### Related issue(s)/PR(s)
result


core/model/modx/processors/search/search.class.php
```diff
@@ -159,20 +159,22 @@ public function searchResources()
         }
 
         $c = $this->modx->newQuery('modResource');
+        $c->leftJoin('modTemplate','modTemplate','modResource.template=modTemplate.id');
+        $c->select($this->modx->getSelectColumns('modResource','modResource').", modTemplate.icon as icon");
```
Mark-H: This is quite confusing to read - please just split it up into two `$c->select` lines.

core/model/modx/processors/search/search.class.php
```diff
@@ -186,6 +188,7 @@ public function searchResources()
                 'type' => $type,
                 'class' => $record->get('class_key'),
                 'type_label' => $typeLabel,
+                'icon' => $record->get('icon')?(str_replace('icon-','',$record->get('icon'))):false,
```
Mark-H: Can be slightly simplified to 

````
'icon' => $record->get('icon') ? str_replace('icon-', '', $record->get('icon')) : false,
````

or perhaps put above the `$this->results[] = array(` where it only has to get the icon into a variable once, and then manipulates that?

Mark-H (changes requested):
Few minor tweaks. 

Also note that no more 2.6.x releases are planned, so this would likely need to be merged into 2.x instead.

Issue #13729, opened 2017-12-22
modx revo 2.6.1  Parse error: syntax error, unexpected '<', expecting end of file in...\core\cache\includes\elements\modplugin\4.include.cache.php on line 25


ALexeyP0708:
modx revo 2.6.1  Parse error: syntax error, unexpected '<', expecting end of file in...\core\cache\includes\elements\modplugin\4.include.cache.php on line 25

generates script in file .\core\cache\includes\elements\modplugin\4.include.cache.php
```

<?php
/** @var modX $modx */
switch ($modx->event->name) {

    case 'OnMODXInit':
        $fqn = $modx->getOption('pdoTools.class', null, 'pdotools.pdotools', true);
        $path = $modx->getOption('pdotools_class_path', null, MODX_CORE_PATH . 'components/pdotools/model/', true);
        $modx->loadClass($fqn, $path, false, true);

        $fqn = $modx->getOption('pdoFetch.class', null, 'pdotools.pdofetch', true);
        $path = $modx->getOption('pdofetch_class_path', null, MODX_CORE_PATH . 'components/pdotools/model/', true);
        $modx->loadClass($fqn, $path, false, true);
        break;

    case 'OnSiteRefresh':
        /** @var pdoTools $pdoTools */
        if ($pdoTools = $modx->getService('pdoTools')) {
            if ($pdoTools->clearFileCache()) {
                $modx->log(modX::LOG_LEVEL_INFO, $modx->lexicon('refresh_default') . ': pdoTools');
            }
        }
        break;
}
return;
<?php
/** @var modX $modx */
switch ($modx->event->name) {

    case 'OnMODXInit':
        $fqn = $modx->getOption('pdoTools.class', null, 'pdotools.pdotools', true);
        $path = $modx->getOption('pdotools_class_path', null, MODX_CORE_PATH . 'components/pdotools/model/', true);
        $modx->loadClass($fqn, $path, false, true);

        $fqn = $modx->getOption('pdoFetch.class', null, 'pdotools.pdofetch', true);
        $path = $modx->getOption('pdofetch_class_path', null, MODX_CORE_PATH . 'components/pdotools/model/', true);
        $modx->loadClass($fqn, $path, false, true);
        break;

    case 'OnSiteRefresh':
        /** @var pdoTools $pdoTools */
        if ($pdoTools = $modx->getService('pdoTools')) {
            if ($pdoTools->clearFileCache()) {
[...]

Issue #11199, opened 2014-03-11
Modx/xPDO cache manager - Append Vs Write


morsolutions:
Unsure if this should go to Modx Revolution support or the xPDO repository.. but heres the issue...

We have a Modx setup on a Zend Server stack on a Windows Machine running as an intranet setup for a large corporation. The intranet / Modx site gets a reasonable amount of traffic. 

From a "clear" cache, when multiple users visit the site, the cache breaks. We presume similar to the rackspace issue.. However we can't confirm this as we're not rackspace users! Our error logs show that the cache is wrote to twice, where we get an "Unexpected <" on cache files...

We found changing the fmode in /core/xpdo/xpdocachemanager.class.php to w fixes the issue from happening...

``` php
//original line 222:
$fmode = (strlen($mode) > 1 && in_array($mode[1], array('b', 't'))) ? "a{$mode[1]}" : 'a';
//fix
$fmode = "w";
```
- This also however breaks the Modx Log, as the log needs to be append, not truncated and recreated.. Small price to pay for now.

The part we can't figure out is why is Modx / xPDO always defaulting to using append? The whole cache manager appears to just queue up users to be able to write to a cache file...?

It is also worth pointing out, we have over 100+ Modx sides in production, and we only experience this on this one site with this one setup on a Windows Server with a Zend Server WAMP stack... All of our other sites run on a CentOS CPanel or HSphere stack.

Mark-H:
I wonder if this could be somewhat of a solution for the flock problem that seems to be cropping up again.

ttyiav:
Hello,
I would like to share my solution of the mentioned problem, despite the fact that the original post was dated 2014. The initial setup is similar: using of multithreaded Windows webserver with MODX and use_flock=FALSE. It leads to race condition and cache file corruption.

By design, MODX writes to cache (in writeFile) in append mode regardless of _$mode_ parameter value. It is stated that this is because the append mode doesn't truncate file on open (and before obtaining file lock).

PHP documentation, in turn, states that 
> If you have opened the file in append ("a" or "a+") mode, any data you write to the file will always be appended, regardless of the file position. 
Source: http://php.net/manual/en/function.rewind.php

Therefore, the successful cache file write will occur **only if both _fseek()_ and _ftruncate()_** calls at https://github.com/modxcms/revolution/blob/2.x/core/xpdo/cache/xpdocachemanager.class.php#L246 succeed. Both results must be checked because _ftruncate()_ does't change file position.

[...]

PR #13874, opened 2018-04-21
Issue #13729

VoiD2008:
### What does it do?
Changes xpdo::cacheManager::writeFile  fopen mode 

### Why is it needed?
Sometimes when cache manager trying to write file twice at the same time script content appends creating exception <?php ...code... <?php ...code... (not closed "<?php" tag but opened twice)

xpdo::cacheManager::writeFile has hardcoded "A" file open mode when using flag "B" or "T" on call
`$fmode = (strlen($mode) > 1 && in_array($mode[1], array('b', 't'))) ? "a{$mode[1]}" : 'a';`

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13729

### DIFF
--- a/core/model/modx/modscript.class.php
+++ b/core/model/modx/modscript.class.php
@@ -159,7 +159,7 @@ class modScript extends modElement {
                 if ($folderMode) $options['new_folder_permissions'] = $folderMode;
                 $fileMode = $this->xpdo->getOption('new_cache_file_permissions', null, false);
                 if ($fileMode) $options['new_file_permissions'] = $fileMode;
-                $result = $this->xpdo->cacheManager->writeFile($includeFilename, "<?php\n" . $script, 'wb' , $options);
+                $result = $this->xpdo->cacheManager->writeFile($includeFilename, "<?php\n" . $script, 'w' , $options);
             }
         }
         if ($result !== false) {


VoiD2008:
patch didnt help
issue not fixed

alexeyp0708:
@VoiD2008 
Your patch can affect the rest of the cache for which has methods for write  in end  cache.
I already pointed out the problem [https://github.com/modxcms/revolution/issues/13729#issuecomment-375453763](url). If you want to make corrections, you need (example code):
`if ($locked) {
fclose ($file);
$file = fopen ($filename, 'wb');
 //fseek ($file, 0); //delete line
// ftruncate($file, 0); //delete line`

VoiD2008:
No, my patch affects only /core/model/modx/modscript.class.php
and cache for snippets and plugins (they caching one per file).
But my patch is useless. Bug with duplicate caching was still there even after patching. So I closed this request. Only file lock option helping with this issue. Or maybe changing xPDO behavior with file writing. Maybe using temporary files. But I'm not ready for changing xPDO source code yet.;)

alexeyp0708:
on the fingers I will explain how to safe lead the changes.
create a copy of the class named xPDOCacheManager2. Make changes to it.
create a copy of the class named xPDOFileCache2 extends xPDOCache
replace in this class all text xPDOCacheManage on xPDOCacheManage2.
Register in the manager settings the cache driver xPDOFileCache2

alexeyp0708:
Or just replace the cache driver from the available ones. you will have to work with the settings of third-party keshev, but it's worth it.

VoiD2008:
yep, but I tryed to fix the issue, not to find some other ways;) 
Also this issue happens only on windows OS in my case, just on dev, so finding workaround was just a skill training.

alexeyp0708:
https://github.com/modxcms/revolution/issues/11199

PR #13875, opened 2018-04-23
MODX3 redesign

bezumkin:
### What does it do?
- Refactored resource controllers
- Added simple tests for resource controllers
- Redesigned resource pages in manager
- 2 new checkboxes: show_in_tree and hide_children_in_tree
- New styles for windows, tables, tabs and checkboxes

### Why is it needed?
To release MODX 3

### Related issue(s)/PR(s)
N/A

**Warning! Disable `compress_js` and `compress_css` system settings while test this PR!**

screen shot 2018-04-23 at 10 55 38

screen shot 2018-04-23 at 10 55 55 screen shot 2018-04-23 at 10 55 41

screen shot 2018-04-23 at 10 56 10

screen shot 2018-04-23 at 10 57 14

screen shot 2018-04-23 at 10 57 19


joeke:
I'm not sure if replacing all the checkboxes with switches is the way to go. From a UI/UX perspective a switch is an instant on/off action, and it's something the user expects as well (read this http://uxmovement.com/buttons/when-to-use-a-switch-or-checkbox/ for example).
Also, are the radiobuttons replaced with these switches as well? If so, this would cause confusion, because checkboxes and radio buttons will look the same.
So maybe we could keep checkboxes in the appropriate places, but redesign them? For instance like in Google's Material design: https://material.io/guidelines/components/selection-controls.html
@gpsietzema @bezumkin

bezumkin:
@joeke This way we should remove all switches, because we have no forms in current manager, without submission.

I'm really sure, that now resources page looks better, than with checkboxes (that was already styled, btw). I can leave new switches only at resources page, but I believe, there will be weird feeling on all other. We will loose consistency.

Maybe someone should speak about it before approve the design.
resource-edit-content
resource-edit-settings

wax100:
Will U also change checkboxes here?
screenshot_7
And restyle combo-box-boolean here? 
screenshot_6

joeke:
@bezumkin Yes I do like the switches as well and they look nice, but in some places they just feel a little weird. I see no problem in having switches in some places (for example the resource edit screen) and checkboxes in other places, for example when using multiselect / bulk select in grids.. a checkbox would make more sense to me there.

Another question: how is the radio button styling gonna be? for example here:
screen shot 2018-04-24 at 11 01 09

bezumkin:
@joeke 
> Another question: how is the radio button styling gonna be?

You can just test this PR and look yourself =)

joeke:
@bezumkin Haha yes you're right :).. will do that

joeke:
I did some testing, and saw that radio buttons indeed look exactly the same as the checkboxes. I think this will be confusing for a lot of people, as you cannot see beforehand if it's a radio or a checkbox.
Check this screencast to see what I mean:
http://recordit.co/qeBtb11Zeo
As you see only by clicking on the items you know if you can only select one or multiple. Also people could wonder perhaps if there's something broken because when you click a second radio the first one gets de-selected, which is how a radiobutton works, but very confusing if you think you clicked a checkbox.

rthrash:
I definitely wouldn't try to substitute toggles for long lists or radio buttons. However, the toggles do work well and look great on the Resources overview page. To address the issue of no "ajax saving" it may make sense if the toggles are changed to put a warning stating so above them if they change.

RinzeVanDerBrug:
Nice work, Vasily! Few notes:
- You're using three periods in the _‘More’_ button. Conventionally these dots are vertically centered. Would recommend to use this icon instead of the periods: https://fontawesome.com/icons/ellipsis-h?style=solid
- For consistency, I would try to move the icons of the publishing dates to the left, just like in the buttons (or vice versa) (made this mistake in the design as well).
- The switches seem a little too big (in size), in my opinion.
- About the panels in the sidebar: did you intentionally left out the title section including the show/hide dropdown (which was in the design)?
- Some labels of input fields have colons and some don't. This should be consistent. I'd suggest to leave them out. 

About the discussion around the switches/checkboxes and radio buttons: @joeke makes a fair point: there certainly should be visual distinction between radio buttons and switches/checkboxes. 

However, these three UI elements can co-exist without being inconsistent. We only need some logic. So let’s say we use checkboxes for long lists in which you can select multiple items at once (like the one from Yana (@wax100) her screenshot), radio buttons for lists in which you can select only one item at once (like in @joeke’s screenshot) and switches for independent on/off settings (like yes/no system settings and published/unpublished resources).

I also heard some community members suggesting to move the alias/longtitle etc. to the sidebar as well, so the content field moves up. I will make a redesign in which this point is considered.

bezumkin:
@RinzeVanDerBrug 

1, 2, 3 - Fixed
screen shot 2018-04-30 at 09 24 07

4 - Fixed too
screen shot 2018-04-30 at 10 03 30
screen shot 2018-04-30 at 10 03 42

5 - Now all labels have colons
screen shot 2018-04-30 at 10 04 50

bezumkin:
screen shot 2018-05-01 at 15 14 43
screen shot 2018-05-01 at 15 15 15
screen shot 2018-05-01 at 15 16 48

Ibochkarev:
Nice!!!

RinzeVanDerBrug:
In addition to the switches/checkbox/radio buttons discussion:
artboard
Guidelines are from the Google Material Design guide. 

You could use these: 
https://materializecss.com/radio-buttons.html
https://materializecss.com/checkboxes.html

OptimusCrime:
@bezumkin How are these menues opened? On click or on hover?

bezumkin:
After all this discussion about checkboxes, switches and radio buttons I decided to change checkboxes only at resource page as was in approved design by Rinze.

All other pages should stay untouched before we will understand how we want to change them.

JoshuaLuckers:
My findings:

My first impressions: it looks great, good work all!

- The left-bar trigger might cause confusion at first. It looks like a "one step back arrow". It also misses a label to explain what kind of button it is.
- The left-bar should be context aware. When I navigate to "Manage --> Users" I'm in the "Manage" or "Users" context. I would expect the left-bar to have other actions related to "Manager" and/or "Users" to be visible here. It now looks like I'm still in the "Content" context.
- It might be common to use the logo to go back to "home". But for some users it might be unclear how they navigate back to the dashboard.
- I can't scroll in the dashboard (on my laptop not all elements are in the view)
- The top and bottom elements in the dashboard need some top and bottom margin.
- Please add back "hover to collapse" in the left-bar. Or make it optional. 
- When using the Media Browser the active directory should be marked in the tree. It's unclear where I am now unless I look at the info in the status bar. Before the redesign the first file in the directory was marked as "active" and also visible in the tree.
- When editing a resource, the tab "Settings" would make more sense if it was named "Properties". In my opinion "settings" would be stuff like "visible in the menu, template etc". Properties would be things like what's now in the settings tab.
- When a field is invalid when creating a resource, like when the title is missing or to long) it only shows the red border around the input. The error message is missing.

JoshuaLuckers:
It would be good if the active tab in de left-bar was bold. 
screen shot 2018-06-03 at 09 07 25

Mark-H (approved):
I'm going to approve this not because it's perfect, but because it allows the project team and others to submit fixes and improvements to it. Nothing beats actually being able of installing and working with it to find how to improve it, and some of the test remarks already posted in this PR would be good to address following merge.

alroniks (approved):
I agree with Mark, and approving it to allow further moving.

alroniks:
It is able to merge but as it is a huge change I guess @opengeek should make a final decision. 
cc @gpsietzema

PR #13877, opened 2018-04-23
Make the resource class a little closer to SOLID principles

sergant210:
### What does it do?
Move resource parsing functionality from the modResponse class to the modResource class. 

### Why is it needed?
1. A little closer to SOLID principles. Response class does not need to know about how the resource class must be parsed. Derivatives of the resource class can have their own implementation of parsing.
2. Also it can be useful for ajax request:
```
$resource = $modx->getObject('modResource', $id);
$content = $resource->parseContent();
```

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -432,6 +432,19 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
     }
 
     /**
+     * Prepare the resource for output.
+     */
+    public function prepare()
+    {
+        # 1. Parse cacheable elements if exist.
+        $this->process();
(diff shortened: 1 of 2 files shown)


Mark-H:
We'd probably want this to target 3.x instead of 2.x. I do like this idea :)

alroniks:
@Mark-H We had a discussion with Jason, all changes that do not break BC (this case) should be targeted to 2.x.

Mark-H (approved):

alroniks (approved):

opengeek:
I really didn't want this change merged, especially for 2.x. I'm not convinced this does not break BC.

PR #13884, opened 2018-05-01
Add Russian  translation to new Dashboard

wax100:
### Why is it needed?
Add Russian lexicon to new Dashboards


Mark-H:
Hey Yana 👋 Thanks for this, but translations should be submitted through CrowdIn http://translate.modx.com/

It looks like we need to push the new 3.0 strings to crowdin though. @theboxer or @opengeek, is that something you can do so people can start submitting translations there?

opengeek:
@Mark-H that will happen soon, but we need to get 2.7 wrapped and merged up before we do that.

alroniks:
I am closing this PR because it's already discussed. Question is known and plans are clear for all.

Issue #13885, opened 2018-05-01
[SECURITY] Upgrade PHPMailer (at least) to v5.2.22


jarda-svoboda:
"An issue was discovered in PHPMailer before 5.2.22. PHPMailer's msgHTML method applies transformations to an HTML document to make it usable as an email message body. One of the transformations is to convert relative image URLs into attachments using a script-provided base directory. If no base directory is provided, it resolves to /, meaning that relative image URLs get treated as absolute local file paths and added as attachments. To form a remote vulnerability, the msgHTML method must be called, passed an unfiltered, user-supplied HTML document, and must not set a base directory."

https://github.com/PHPMailer/PHPMailer/releases/tag/v5.2.22
https://nvd.nist.gov/vuln/detail/CVE-2017-5223

alroniks:
PR is merged and release even published, so I am closing this thicket.

PR #13886, opened 2018-05-01
Update phpmailer from 5.2.21 to 5.2.26 to fix various security issues [#13885]

Mark-H:
### What does it do?
Upgraded the included version of phpmailer from 5.2.21 (last updated in #13229) to the latest in the 5.2.x series, specifically 5.2.26.

### Why is it needed?
There have been a couple of security fixes in phpmailer since our last update. From https://github.com/PHPMailer/PHPMailer/releases:

- 5.2.26: Minor security backport from 6.0 - set Debugoutput in constructor according to SAPI in use, avoiding potential XSS in default debug output. Thanks to Bankde Eakasit for spotting it.
- 5.2.24: ~SECURITY Fix XSS vulnerability in one of the code examples, CVE-2017-11503. The code_generator.phps example did not filter user input prior to output. This file is distributed with a .phps extension, so it it not normally executable unless it is explicitly renamed, so it is safe by default. There was also an undisclosed potential XSS vulnerability in the default exception handler (unused by default). Patches for both issues kindly provided by Patrick Monnerat of the Fedora Project.~ doesn't apply to us as we don't ship examples
- 5.2.22: SECURITY Fix CVE-2017-5223, local file disclosure vulnerability if content passed to msgHTML() is sourced from unfiltered user input. Reported by Yongxiang Li of Asiasecurity. The fix for this means that calls to msgHTML() without a $basedir will not import images with relative URLs, and relative URLs containing .. will be ignored.

This upgrades to the latest in thet 5.2 series. We should probably migrate to 6.x but that's more of a MODX 3 job.

Note that I am not aware of any exploits for this specific to MODX, but probably best to stay a little up to date... good thing MODX3 uses composer ;) 

### Related issue(s)/PR(s)
Notified via #13885
Previous upgrade #13229


opengeek (approved):
Looks good. Merging…

PR #13887, opened 2018-05-01
Fix some minor XSS issues reported in security ticket 69

Mark-H:
### What does it do?
Applies htmlspecialchars to some processors which were exposing 3 stored XSS issues. 

### Why is it needed?
Tighten security a bit. 

None of these XSS issues should have any far-reaching consequences. They all require manager access for starters. Also

- The one in the recently edited dashboard widget would only be useful to XSS yourself, as the resource list only shows resources you personally edited. Perhaps a problem when sharing user accounts. 
- Access to edit the top menu (actions) is needed for number 2. That's an advanced permission that only full admins should have.
- Adding/editing setting permission is needed for number 3. Also, that should already be limited to full admin users.

### Related issue(s)/PR(s)
Security inbox ticket number 69. Reported by Aman Sahey on April 25th.

### DIFF
--- a/core/model/modx/processors/security/user/getrecentlyeditedresources.class.php
+++ b/core/model/modx/processors/security/user/getrecentlyeditedresources.class.php
@@ -72,7 +72,7 @@ class modUserGetRecentlyEditedResourcesProcessor extends modObjectGetListProcess
         if (!$object->checkPolicy('view')) return array();
 
         $resourceArray = $object->get(array('id','pagetitle','description','published','deleted', 'context_key'));
-        $resourceArray['pagetitle'] = htmlspecialchars($resourceArray['pagetitle'], ENT_QUOTES, 'UTF-8');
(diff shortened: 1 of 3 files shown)


Jako:
Maybe you should use `$modx->getOption('modx_charset', $options, 'UTF-8')`

opengeek (changes requested):
I concur with @Jako in that you should be using `$modx->getOption('modx_charset', $options, 'UTF-8')` rather than hardcoding UTF-8.

Omeryl:
Just a minor note, looks like we're going to need to go back and fix this up in a number of other places: https://github.com/modxcms/revolution/blob/4e82a449ca7e152e6bad6507d4b705539af1ae0f/core/model/modx/processors/security/group/getnodes.class.php#L134

opengeek (approved):

Issue #13841, opened 2018-03-29
Could not load class: xPDOTransport from transport.xpdotransport


gpsietzema:
### Summary
When installing any extra in the 3.x branch, you get this error.

### Step to reproduce
Go to Extras > Installer -> install an extra

### Observed behavior
Extra installs fine, but you get this error mentioned in red: Could not load class: xPDOTransport from transport.xpdotransport

### Expected behavior
No error message.

### Environment
MODX on 3.x branch, nginx, php 7.1

BobRay:
I'm seeing this as well.

bezumkin:
@gpsietzema We can merge #13888 to current 3.x code - it will fix this error.

alroniks:
It seems PR was merged, so it's a reason to close this ticket.

PR #13888, opened 2018-05-02
Improve compatibility with extras

bezumkin:
### What does it do?
- Fixed "undefined constant XPDO_PHP_VERSION" on extras install
- Fixed "Could not load class: xPDOTransport" on extras install
- Fixed modX::getControllerPath method signature

### Why is it needed?
To improve compatibility with old extras

### Related issue(s)/PR(s)
#13841

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -24,7 +24,7 @@ interface modResourceInterface {
      * @param xPDO $modx A reference to the modX object
      * @return string The absolute path to the controller for this Resource class
      */
-    public static function getControllerPath(&$modx);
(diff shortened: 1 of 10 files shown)


Mark-H (dismissed):
Doesn't resolve every issue I'm seeing in my extras (which I didn't expect it to!), but the getControllerPath signature change does fix a very hard-to-work-around issue for custom resources.

bezumkin:
Closed due to favour of #13900

bezumkin:
Reopened in case #13900 will not be accepted.

Mark-H (approved):

PR #13892, opened 2018-05-04
Escaping ORDER BY fields.

zeves095:
since MySQL word RANK (R) added in 8.0.2 (reserved). This field is used in modCategory for sorting ASC in manager.
without that escaping you get false result for query.

### What does it do?
i just make xpdo->escape for field when generating sql order by part.

### Why is it needed?
With Mysql >=8.0.2  core object modCategory have field name "rank" - this is reserved keyword. And in manager on resource TV tab i can'nt see category vertical tabs.

### DIFF
--- a/core/xpdo/om/mysql/xpdoquery.class.php
+++ b/core/xpdo/om/mysql/xpdoquery.class.php
@@ -136,11 +136,11 @@ class xPDOQuery_mysql extends xPDOQuery {
         if ($command == 'SELECT' && !empty ($this->query['sortby'])) {
             $sortby= reset($this->query['sortby']);
             $sql.= 'ORDER BY ';
-            $sql.= $sortby['column'];
+            $sql.= $this->xpdo->escape($sortby['column']);
             if ($sortby['direction']) $sql.= ' ' . $sortby['direction'];
             while ($sortby= next($this->query['sortby'])) {
                 $sql.= ', ';
-                $sql.= $sortby['column'];
+                $sql.= $this->xpdo->escape($sortby['column']);
                 if ($sortby['direction']) $sql.= ' ' . $sortby['direction'];
             }
             $sql.= ' ';


ghost:
Dashboard widgets also use a "rank" field.

ghost:
There are several other tables that use a `rank` field, some use that field as an index.

zeves095:
This is my first commit to you. What should i change|add also ?

@sottwell > There are all ok for select columns - they are escaped.  This problem is only for sorting  fields (for grouping - not yet fixed - but it's same in same file)

Jako:
Could you please open an issue or PR on https://github.com/modxcms/xpdo (since it is an xPDO issue).

opengeek:
This is NOT an xPDO issue. This is a MODX core issue because fields named `rank` are now matching a new reserved word in MySQL 8.

opengeek:
However, with that said, this cannot be solved in the xPDO layer. You need to make sure those fields are escaped in the MODX layer or the fields need to be renamed.

Jako:
Ok. It is not an xPDO issue, but the PR targeted an xPDO file. I create an issue for escaping rank in the MODX code.

jeanclaudereiss:
I am sorry to intervene so late, but I think that this an architecturally wrong decision. I  might not have all the necessary information, but here are my arguments, based on the information I have.

- xPDO is an ORM (or as its web site says, an ORB). As such, one of its objectives should be to shield the programmers from the differences between, and the specifics of each of the database platforms and versions supported by the ORM.
- With the currently proposed approach, xPDO transfers to the client applications (e.g. MODX) the responsibility of knowing if the field names (and also the table names) used in the application correspond to reserved words in one of the database platforms and versions supported by xPDO. 
- In addition, the developers of the application would then have to constantly monitor developments in each of the database platforms supported by xPDO in order to make sure that changes in the database platforms do not generate new errors in the application.

In my opinion, this is the wrong approach. Each database platform's query language supports the escaping of reserved words used as column or table names. xPDO should either escape reserved words for supported database platforms, or (simpler and probably better) routinely escape all names used in queries.

In MySQL for example, you can write a query like this to avoid problems with reserved words: SELECT * FROM \`modx_dashboard_widget\` ORDER BY \`name\`
In PostgreSQL and MS SQL Server this would be SELECT * FROM "modx_dashboard_widget" ORDER BY "name". This also applies for MySQL when ANSI SQL mode is enabled.

We see from this example that, in order to be correct, if we need to escape the names in the application, we need to introduce database specific code in the application itself. This code belongs clearly in the database layer.

Additionally, the list of reserved words in database systems is fairly long and sometimes surprising, which complicates the matter. There are also quite a number of variations between database systems. You find reserved words such as "year", "action", "avg". For example, "year" is reserved in MySQL and MS SQL, but not in PostgreSQL. Keeping on top of this at the application level is just an impossible task. So therefore the solution of routinely escaping all keywords at the ORM/ORB level is in my opinion a much better solution.

Unless there is something I have completely overseen

zeves095:
it can be done on xpdo layer: https://github.com/modxcms/xpdo/pull/138#issuecomment-386628724

opengeek:
@Jako This is not a "currently proposed approach." This has been the behavior of the code since it was developed. And this has been discussed many times including when some core misuse exposed vulnerabilities.

@zeves095 As I asked in the xPDO PR, do you have documentation or supporting evidence on your proposed solution? How does that ensure the value is not an expression? If we can get some thorough unit tests covering a wide range of expressions into a PR for this, I would be willing to include a solution that we can reasonably ensure does not break existing code that expects the values not to be escaped.

PR #13893, opened 2018-05-05
Fixes TV > tpl access sorting

davidpede:
### What does it do?
Fixes v old bug by setting remoteSort to false in the 'TV>Template Access' grid

### Why is it needed?
When sorting by the Access column the processor returns no results and xpdo logs an unknown column error. This because the access value is set manually in the processor after the db query.

### Related issue(s)/PR(s)
Fixes: #12319, #7896 New pr version of #13597

Tested on: MODX 2.6.4-dev, PHP 7.0.27 & 10.1.30-MariaDB

### DIFF
--- a/core/model/modx/processors/element/tv/template/getlist.class.php
+++ b/core/model/modx/processors/element/tv/template/getlist.class.php
@@ -69,19 +69,20 @@ class modElementTvTemplateGetList extends modProcessor {
                 'Category.id' => $category
             ));
         }
-        
+
         $data['total'] = $this->modx->getCount('modTemplate',$c);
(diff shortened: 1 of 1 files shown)


davidpede:
I've moved setting the access value to the getData() method as suggested by Jako in #13597. However when setting it to true/false based on tmplvarid existing, every checkbox in the grid is checked.

Seems setting access to either tmplvarid or empty is sufficient for the grid.

Jako:
Did you fix the always checked issue?

davidpede:
Yes they are checked correctly in this pr

Issue #13890, opened 2018-05-03
[3.x] Flash swf file success check in installation summary shows as warning


Mark-H:
### Summary
That the flash file isn't there is a good thing, shouldn't be styled as warning.

![schermafbeelding 2018-05-03 om 12 09 20](https://user-images.githubusercontent.com/312944/39571239-29b1d7bc-4ecb-11e8-9d0b-2774dfe21b56.jpg)

I think this was introduced in #13814

### Step to reproduce
Install 3.x branch from git which has that file removed already

### Observed behavior
Styled as warning.

### Expected behavior
Styled as success

### Environment
Revision 2f762359b163674cd6e845a52d6de72111547cb3

alroniks:
Issue fixed in 3.x.

PR #13897, opened 2018-05-10
If the copy to clipboard flash file is missing it's a good thing and …

JoshuaLuckers:
…we should inform the user with a success message telling that the file has already been removed.

### What does it do?
Changed the result from type "warning" to type "success" and changed the message for the install summary.

### Why is it needed?
That the flash file isn't there is a good thing, shouldn't be styled as warning. 

### Related issue(s)/PR(s)
Issue #13890


alroniks (approved):

Mark-H (approved):
Thanks for addressing this :)

alroniks:
It looks pretty easy and doesn't change behavior, only fixes text and type of the message. So it can be merged safely.

PR #13908, opened 2018-05-19
Unit Test fixes

JoshuaLuckers:
### What does it do?
Made some Unit Tests run again and it also solves some issues.

### Why is it needed?
Tests not running is a waste and might give false information about changes (if a test doen't run, it also doesn't check if changes break something).

### Related issue(s)/PR(s)
None that I'm aware of.


alroniks:
I think that makes sense fix it in 2.x branch. Release of 2.6.x isn't in our plans.

opengeek:
Actually, this one is fine in 2.6.x; we are releasing 2.6.4. Just leave this as is.

opengeek (approved):

Issue #12996, opened 2016-05-03
"password_generated_length" & "password_min_length" not working


syberknight:
xpost @ https://forums.modx.com/thread/100064/forgot-password-length-too-small#dis-post-541129

i've discovered (after getting all the complaints & having to fix manually all the people who's having this problem) that the "Forgot Password" option auto-generates an 8 character password. but it never works to sign-in with because i have all my validation set to 10 characters.

i would prefer not to have to drop my password registration/login validation to 8 characters, but if there's no solution, i realize i'll have to.

so i went hunting first & found in the System Settings > Core listing a couple parameters that seem to me should fix this to my preference... but they aren't working.
they are...
Password Auto-Generated Length "password_generated_length" which defaults to "8" (https://rtfm.modx.com/revolution/2.x/administering-your-site/settings/system-settings/password_generated_length)
Minimum Password Length "password_min_length" which also defaults to "8" (https://rtfm.modx.com/revolution/2.x/administering-your-site/settings/system-settings/password_min_length)

so i changed them to 11 & 10 respectively. but the forgot password emails STILL are only sending 8 character passwords.

am i missing something?
thanks!!!

fyi: using Revolution 2.4.3-pl & the Login extra 1.9.2-pl.

syberknight:
anybody? nobody?

Mark-H:
- Manager password resets: Looks like a potential bug in `manager/controllers/default/security/login.class.php`. Line 226 generates the new password through modUser->generatePassword which accepts a $length parameter, but that isn't passed. The default password length in that case would be 10 though, not 8. 
- Login package password resets: components/login/controllers/web/ForgotPassword.php calls generatePassword without parameters, which has a standard length of 8. Does not check the proper setting at all. Sounds like a bug to report at https://github.com/modxcms/Login/issues

syberknight:
thanks Mark-H. 
fyi, sottwell & BobRay came up with a workaround for now @ https://forums.modx.com/thread/100064/forgot-password-length-too-small#dis-post-541202

alroniks:
Issue fixed in a general but there are also some new issues opened with more detailed cases according to respecting system settings values. Closing it now. Feel free to reopen if needed.

PR #13909, opened 2018-05-19
modUser - generatePasssword improvements.

JoshuaLuckers:
### What does it do?
Removed a duplicate generatePassword and make it use "password_generated_length"

### Why is it needed?
Duplicate code can cause problems down the road.
This also solves an issue where the system setting "password_generated_length" didn't work properly in all cases.

### Related issue(s)/PR(s)
It solves issue #12996

### DIFF
--- a/_build/test/Tests/Model/Security/modUserTest.php
+++ b/_build/test/Tests/Model/Security/modUserTest.php
@@ -105,6 +105,19 @@ class modUserTest extends MODxTestCase {
         );
     }
 
+    public function testGeneratePasswordWithPasswordLengthOption()
+    {
+        $password = $this->user->generatePassword();
+        $this->assertNotEmpty($password);
+        $this->assertNotEquals(12, strlen($password));
+
+        $this->modx->setOption('password_generated_length',12);
+        $anotherPassword = $this->user->generatePassword();
+
+        $this->assertNotEmpty($anotherPassword);
+        $this->assertEquals(12, strlen($anotherPassword));
+    }
+
     /**
      * Ensure passwordMatches works
      * @return void
--- a/_build/test/Tests/Model/Security/modUserTest.php
+++ b/_build/test/Tests/Model/Security/modUserTest.php
@@ -113,9 +113,13 @@ class modUserTest extends MODxTestCase {
 
         $this->modx->setOption('password_generated_length',12);
         $anotherPassword = $this->user->generatePassword();
-
         $this->assertNotEmpty($anotherPassword);
         $this->assertEquals(12, strlen($anotherPassword));
+
+        $this->modx->setOption('password_generated_length','');
+        $yetAnotherPassword = $this->user->generatePassword();
+        $this->assertNotEmpty($yetAnotherPassword);
+        $this->assertEquals(10, strlen($yetAnotherPassword));
     }
 
     /**
--- a/core/model/modx/moduser.class.php
+++ b/core/model/modx/moduser.class.php
@@ -788,11 +788,8 @@ class modUser extends modPrincipal {
      * @return string The newly generated password
      */
     public function generatePassword($length = null,array $options = array()) {
-        if ($length === null || $length === '') {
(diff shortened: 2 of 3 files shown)


core/model/modx/moduser.class.php
```diff
@@ -787,7 +787,10 @@ public function removeLocks(array $options = array()) {
      * @param array $options
      * @return string The newly generated password
      */
-    public function generatePassword($length = 10,array $options = array()) {
+    public function generatePassword($length = null,array $options = array()) {
+        if ($length === null) {
+            $length = $this->getOption('password_generated_length',null,10);
```
alroniks: @JoshuaLuckers Just a comment, not requirement. But as we are fixing this part maybe we should use here length that set in system settings by default to keep consistency? In system settings, it set as 8 symbols.

_build/test/Tests/Model/Security/modUserTest.php
```diff
@@ -105,6 +105,19 @@ public function providerGeneratePassword() {
         );
     }
 
+    public function testGeneratePasswordWithPasswordLengthOption()
```
Mark-H, line 108: More tests! ❤️

core/model/modx/moduser.class.php
alroniks: Another issue that I found it when to set `password_generated_length` as an empty string (not null, but '') generator will return the empty password. As I remember getOption does not check empty strings and return it as is. And we need to check it additionally.
At least we have inconsistency between `getOption` in `xPDOObject` and in `xPDO` classes. The second one has fourth parameter `$skipEmpty = false` that allows properly check empty strings when it set in true. 
Details here https://github.com/modxcms/revolution/blob/cdbbb8f611ce2b7c6e6460895b297ed5cdf54c75/core/xpdo/xpdo.class.php#L686 and https://github.com/modxcms/revolution/blob/31fdf1a342844f9efc854e9d6f7c18815e28c09e/core/xpdo/om/xpdoobject.class.php#L697

alroniks (changes requested):

_build/test/Tests/Model/Security/modUserTest.php
```diff
@@ -105,6 +105,19 @@ public function providerGeneratePassword() {
...
+    {
+        $password = $this->user->generatePassword();
+        $this->assertNotEmpty($password);
+        $this->assertNotEquals(12, strlen($password));
+
+        $this->modx->setOption('password_generated_length',12);
```
alroniks, line 114: Also it possible to set `password_generated_length` as an empty string and it also should be checked in tests.

core/model/modx/moduser.class.php
JoshuaLuckers: Thanks! I decided to use 10 because the current method signature also uses 10. I would suggest increasing the default system settings to 10 instead.

_build/test/Tests/Model/Security/modUserTest.php
JoshuaLuckers: I will add it to the tests! Thanks.

core/model/modx/moduser.class.php
```diff
@@ -787,7 +787,13 @@ public function removeLocks(array $options = array()) {
      * @param array $options
      * @return string The newly generated password
      */
-    public function generatePassword($length = 10,array $options = array()) {
+    public function generatePassword($length = null,array $options = array()) {
+        if ($length === null || $length === '') {
```
alroniks: After discussion with Jason.
Better to use such variant:
```
if ($length === null) {
    $length = $this->xpdo->getOption('password_generated_length', null, 10, true);
}
```
And return default value (10) to function signature.

alroniks (changes requested):

alroniks (approved):
My requirements fixed, functions work as expected.
Fallback when system setting set as empty string works.

alroniks:
@opengeek @Mark-H Looks like it is ready for merge.

Mark-H (approved):

PR #13911, opened 2018-05-21
Fixes a permissions lexicon typo

JoshuaLuckers:
### What does it do?
Fixes a typo in the dutch permissions lexicon "tree_show_resource_ids".

### Why is it needed?
It solves a typo.

### Related issue(s)/PR(s)
None


alroniks:
We follow the convention that all translates should be provided only via translation platform to avoid issues with accidentally lost translations. Please, fix it here https://crowdin.com/project/modx-revolution and all translated strings will be added to the code once during the making release.
I'm closing this PR for now.

Issue #13910, opened 2018-05-21
Add packages required for developing to composer.json


JoshuaLuckers:
### Summary
When you install MODX through composer packages for development are not installed. To help new contributors set up a development environment it would be helpful if the development packages were included in composer.json.

### Step to reproduce
Download the MODX source via Git, composer etc.

### Observed behavior
composer.json has a list of required packages but it's missing the packages required for development (like the correct phpunit version).

### Expected behavior
composer.json should include require-dev packages.

OptimusCrime:
Which other packages does this concern other than PHPUnit?

JoshuaLuckers:
At the moment PHPUnit seems to be the only one (it has to be a specific version to get it work with the current MODX unit test code).

pbowyer:
@modxbot close

PR #13913, opened 2018-05-26
Added phpunit/phpunit ^5 as development requirement.

JoshuaLuckers:
### What does it do?
It marks phpunit/phpunit ^5 as development requirement in composer.json

### Why is it needed?
To help new contributors set up a development environment it would be helpful if the development packages were included in composer.json. 
In this case it's only phpunit.

### Related issue(s)/PR(s)
#13910

### DIFF
--- a/_build/test/MODxTestCase.php
+++ b/_build/test/MODxTestCase.php
@@ -26,7 +26,7 @@
  *
  * @package modx-test
  */
-abstract class MODxTestCase extends PHPUnit_Framework_TestCase {
+abstract class MODxTestCase extends \PHPUnit\Framework\TestCase {
     /**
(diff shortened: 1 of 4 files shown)


alroniks (dismissed):
It works perfectly.

opengeek (changes requested):
This actually does not work for PHP 7.2. See the xPDO composer.json for an example of how to specify the dependency to work on 5.6 through 7.2.

```    "require-dev": {
        "phpunit/phpunit": "~5.7|~6.5"
    },
```

JoshuaLuckers (commented):
Thanks! I've made the necessary changes.

opengeek (approved):

PR #13914, opened 2018-05-26
Add ID number to manager pages (resources and elements)

### LINKED ISSUE #11096: Add ID number to manager pages (resources and elements)

JoshuaLuckers:
### What does it do?
It makes the ID of the object on various manager pages visible after the title/name.
This improves usability of the current manager design.

### Why is it needed?
It makes it easier to find the ID of a certain object (like the ID of a page). More discussion about this can be found in issue #11096.

MODX 3 is right around the corner, but it would still be very helpful to have this in a 2.* release. 

### Related issue(s)/PR(s)
Fixes #11096

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -54,7 +54,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             if (!Ext.isEmpty(this.config.record.pagetitle)) {
                 var title = Ext.util.Format.stripTags(this.config.record.pagetitle);
                 title = Ext.util.Format.htmlEncode(title);
-                Ext.getCmp('modx-resource-header').getEl().update('<h2>'+title+'</h2>');
+                Ext.getCmp('modx-resource-header').getEl().update('<h2>'+title+' <small>('+this.config.record.id+')</small></h2>');
             }
             // initial check to enable realtime alias
             if (Ext.isEmpty(this.config.record.alias)) {
@@ -547,6 +547,9 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                     var titlePrefix = MODx.request.a == 'resource/create' ? _('new_document') : _('document');
                     var title = Ext.util.Format.stripTags(f.getValue());
                     title = Ext.util.Format.htmlEncode(title);
+                    if (MODx.request.a !== 'resource/create') {
+                        title = title+ '<small>('+this.config.record.id+')</small>';
+                    }
                     Ext.getCmp('modx-resource-header').getEl().update('<h2>'+title+'</h2>');
 
                     // check some system settings before doing real time alias transliteration
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -548,7 +548,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                     var title = Ext.util.Format.stripTags(f.getValue());
                     title = Ext.util.Format.htmlEncode(title);
                     if (MODx.request.a !== 'resource/create') {
-                        title = title+ '<small>('+this.config.record.id+')</small>';
+                        title = title+ ' <small>('+this.config.record.id+')</small>';
                     }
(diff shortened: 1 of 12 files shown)


JoshuaLuckers:
I didn't check and/or add permissions as suggested by [wuuti](https://github.com/modxcms/revolution/issues/11096#issuecomment-285280918) in issue #11096

Mark-H (changes requested):
Per the discussion about this in #11096, this should be configurable. 

Could be a setting or the mentioned permissions. I think the permissions would be best as those already toggle similar behavior. Perhaps the permissions are already available to check with `MODx.perm.tree_show_element_ids`, or they would need to be checked for in the controller first. 

Thanks for working on this - always good to see an issue from 2014 (or before!) being addressed!

JoshuaLuckers:
Thanks @Mark-H. I'm not sure if the permission is useful (since a user can still see it in the URL but that would be worth it's own issue/discussion). For now I will check the permissions like you suggested.

I love cleaning and fixing some old bugs. It's the small things that improve the product.

alroniks:
System settings now overloaded enough in MODX. Permissions also not a simple thing in MODX but according to show/hide something it better place to describe behavior. I would review existing system settings somehow to move some of them to permissions area.

JoshuaLuckers:
The ID's are now only visible when the user has "tree_show_element_ids" permission.

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -54,6 +54,9 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             if (!Ext.isEmpty(this.config.record.pagetitle)) {
                 var title = Ext.util.Format.stripTags(this.config.record.pagetitle);
                 title = Ext.util.Format.htmlEncode(title);
+                if (MODx.perm.tree_show_element_ids === 1) {
```
alroniks: For resources need to use proper permissions according to resources instead of elements. This one `tree_show_resource_ids`

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -547,6 +550,9 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                     var titlePrefix = MODx.request.a == 'resource/create' ? _('new_document') : _('document');
                     var title = Ext.util.Format.stripTags(f.getValue());
                     title = Ext.util.Format.htmlEncode(title);
+                    if (MODx.request.a !== 'resource/create' && MODx.perm.tree_show_element_ids === 1) {
```
alroniks: tree_show_resource_ids

manager/controllers/default/resource/update.class.php
```diff
@@ -57,6 +57,7 @@ public function loadCustomCssJs() {
                 ,mode: "update"
             });
         });
+        MODx.perm.tree_show_element_ids = '.($this->modx->hasPermission('tree_show_element_ids') ? 1 : 0).';
```
alroniks: tree_show_resource_ids

alroniks (changes requested):
All looks good except that for resources need to use more suitable permission.

manager/assets/modext/widgets/resource/modx.panel.resource.js
JoshuaLuckers: I changed it, thanks for reviewing!

alroniks (approved):

alroniks:
I've tested this PR. All works fine, ids showing/hiding depends on permissions, resources and elements uses proper permissions.
Also, I found an issue with undefined on template pages but I've already fixed (see the last commit).
So, ready for merging I guess.
@opengeek @Mark-H

Mark-H (approved):

Issue #13899, opened 2018-05-12
3.x Regex validation for text TVs not working correctly


BobRay:
### Summary
Some simple regular expressions seem to work, but others don't.

### Step to reproduce
Create a text-type TV
In the Regular Expression input, put any of these expressions to validate a 3-digit number (with or without enclosing delimiters):

```
^\d\d\d$
\d\d\d
^\d{3}$
\d{3}
```
Edit the TV in a resource.

### Observed behavior
Anything you type appears to be valid.

Strangely, the following regular expression works as expected:

`hello|goodbye`

The input must contain hello or goodbye.

I couldn't find the code that actually validates the input.

### Expected behavior
Invalid message unless input is a 3-digit number

### Environment
MODX 3.0.0 Windows 10. PHP 7.1

OptimusCrime:
What browser did you use? Can you also try this expression: `^[0-9]{3}$`?

BobRay:
Results are the same in both Chrome and Firefox.  Your expression works, so apparently the regex engine we're using is not PCRE compliant. I see that \d is not supported by POSIX (or GNU grep), though it's a pretty standard usage in my experience. Validating a phone number gets pretty verbose without it and it's certainly recognized by any recent version of JavaScript. 

Is this an ExtJS issue?

If we're stuck, I guess this would be an option: [convert PCRE to Regexp](https://www.npmjs.com/package/pcre-to-regexp)

OptimusCrime:
I am not sure why it does not work. `\d` should work fine in javascript as far as I am aware. As long as Extjs uses the in-browser regex engine (and not a drop-in replacement) it should work out of the box. We will have to dig a bit deeper on this one.

JoshuaLuckers:
I found the problem. When using `\d` all values are invalid. Except when using an actual d.
This means the `\` is being ignored/removed.

When I use `\\d` it works like expected.

JoshuaLuckers:
The [MDN documentation on RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) states the following: 
> When using the constructor function, the normal string escape rules (preceding special characters with \ when included in a string) are necessary. For example, the following are equivalent:
```javascript
var re = /\w+/;
var re = new RegExp('\\w+');
```

BobRay:
I had a similar problem  with a regex search function in a processor I was writing that It used the MySQL regex engine. 

This odd-looking function solved it:

```
[...]

PR #13916, opened 2018-05-28
Fix Regex validation for TV's of the input type "Text" 

JoshuaLuckers:
### What does it do?
Create the RegExp object using the literal notation since our regular expression will remain constant. 

### Why is it needed?
This solves an issue with the Regular Expression Validator used for "Text" Template Variables.
When creating the RegExp object using the constructor function, the normal string escape rules (preceding special characters with \ when included in a string) are necessary. This might be unclear to users. 

More info about creating RegExp objects using literal notation or constructor can be found at the [MDN documentation on RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) 

### Related issue(s)/PR(s)
#13899

### DIFF
--- a/manager/templates/default/element/tv/renders/input/textbox.tpl
+++ b/manager/templates/default/element/tv/renders/input/textbox.tpl
@@ -19,7 +19,7 @@ Ext.onReady(function() {
         ,allowBlank: {if $params.allowBlank == 1 || $params.allowBlank == 'true'}true{else}false{/if}
         {if $params.minLength|default},minLength: {$params.minLength|default}{/if}
         {if $params.maxLength|default},maxLength: {$params.maxLength|default}{/if}
-        {if $params.regex|default},regex: new RegExp('{$params.regex|default}'){/if}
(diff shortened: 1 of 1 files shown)


Jako:
Nice find!

Jako:
But this could be a BC, because the original bug could be bypassed with backslashes in the validation string. Or am I wrong with that?

JoshuaLuckers:
You are right, maybe it’s better if I target 3.0 instead of 2.x?

Maybe I should open another issue about how to use the Regular Expression Validator in "Text" TV's since it might be unclear and needs more explanation.

Mark-H (approved):
LGTM!

Thanks for taking the time to include a proper description of the problem and what it fixes.

Jako (approved):

alroniks:
This PR is approved and simple, but I've tested it and results in screenshots below:
RegEx configuration:
template variable_ test _ modx revolution
Valid value:
editing_ home _ modx revolution
Invalid value and error message:
editing_ home _ modx revolution 1

I think it is pretty enough to be sure that it works now as expected. Going to merge it.

PR #13924, opened 2018-06-10
Load all mgr permissions at once to MODx.perm

### LINKED ISSUE #12211: Load all mgr permissions at once to MODx.perm
The same is done with settings which are loaded to MODx.config in core/model/modx/processors/system/config.js.php (time is about 0.06s in my tests).

I suggest to load all distinct permissions to MODX.perm in this config.js.php file. Something like:

``` php
$c = $modx->newQuery('modAccessPermission');
$c->select('name');
$c->distinct();
$perms = $modx->getIterator('modAccessPermission', $c);
$values = array();
foreach ($perms as $perm) {
    $name = $perm->get('name');
    $values[$name] = $modx->hasPermission($name);
}
$o .= 'MODx.perm = ' . $modx->toJSON($values) . ';';
```

My tests show ~0.04s for that.
### Why is it needed ?

Now permissions in JS manager code are scattered in about 40 controllers added not only to MODx.perm but in MODx object itself and on different Ext.Panel objects. Most common code is something like

``` php
MODx.perm.unlock_element_properties = "'.($this->modx->hasPermission('unlock_element_properties') ? 1 : 0).'";
```

Extra developers must write this time by time.
With this addition all simple manager permissions are concentrated in one place that yields more clean code. All additional complex permissions still will be added as above.
Also the array for MODx.perm can be cached by user id or in $_SESSION.
# area-core #enhancement

JoshuaLuckers:
### What does it do?
Load all mgr permissions at once to MODx.perm and removes related JS code scattered around. 

### Why is it needed?
Manager permissions are concentrated in one place. This means that developers do not have to do this manually every time they want to check for a permission using JavaScript.

### Related issue(s)/PR(s)
Fixes #12211

### DIFF
--- a/manager/assets/modext/widgets/security/modx.panel.groups.roles.js
+++ b/manager/assets/modext/widgets/security/modx.panel.groups.roles.js
@@ -41,7 +41,7 @@ MODx.panel.GroupsRoles = function(config) {
         }
     });
 
-    if (MODx.perm.usergroup_user_list == 1) {
+    if (MODx.perm.usergroup_user_list) {
         Ext.getCmp('modx-tree-usergroup').on('click', function(node,e){
             this.getUsers(node);
         }, this);
@@ -52,7 +52,7 @@ MODx.panel.GroupsRoles = function(config) {
 Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
     getPageTabs: function(config) {
         var tbs = [];
-        if (MODx.perm.usergroup_view == 1) {
+        if (MODx.perm.usergroup_view1) {
             tbs.push({
                 title: _('user_groups') + ' & ' + _('users')
                 ,autoHeight: true
@@ -98,7 +98,7 @@ Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
                 }]
             });
         }
-        if (MODx.perm.view_role == 1) {
+        if (MODx.perm.view_role) {
             tbs.push({
                 title: _('roles')
                 ,autoHeight: true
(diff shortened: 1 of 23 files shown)


JoshuaLuckers:
This should target 3.x, however since the 3.x branch is a bit to far behind I decided to target 2.x.
When the 3.x branch is more up to date I will retarget this PR.

Mark-H (dismissed):
Nice one!

manager/assets/modext/widgets/security/modx.panel.groups.roles.js
```diff
@@ -52,7 +52,7 @@ MODx.panel.GroupsRoles = function(config) {
 Ext.extend(MODx.panel.GroupsRoles,MODx.FormPanel,{
     getPageTabs: function(config) {
         var tbs = [];
-        if (MODx.perm.usergroup_view == 1) {
+        if (MODx.perm.usergroup_view1) {
```
alroniks: Looks like here a typo after removing `==1`

alroniks (changes requested):

JoshuaLuckers:
@Alroniks should There is a conflicting file. Should I fix this or is this something an integrator should do?

alroniks:
@JoshuaLuckers Diffs looks easy but there are a lot of stings in diff. It would be nice to do rebase and fix conflict from your side to be sure that we don't miss something.

JoshuaLuckers:
@Alroniks done! 👍

alroniks (approved):

alroniks:
@Mark-H Could you reapprove it, please? There was a minor issue with the typo that fixed and also code was rebased to avoid conflicts. Nothing more.

JoshuaLuckers:
@Mark-H ping

Mark-H (approved):
Pong!

alroniks:
It's already approved but I am wondering shouldn't it be targeted to 2.x because no any BC is broken?
@Mark-H @opengeek

JoshuaLuckers:
I decided to target 3.x because it would be a good opportunity for extra developers when they are testing and/or modifying their extra's to be compatibly with version 3.x to also clean up their code if they manually add those permissions as described in issue #12211.

alroniks:
Yes, make sense according to extras. Merging.

PR #13926, opened 2018-06-11
Implement native PHP password hashing

Mark-H:
### What does it do?

Adds a new modHash instance called modNative that implements native password hashing using password_hash and password_verify. These functions were introduced in PHP 5.5 as a way to make strong password hashing easy, and adaptable for the future.

Full information about the rationale behind those functions can be found in the RFC at https://wiki.php.net/rfc/password_hash

This change also introduces a new method on modHash called verify, and updates the modUser->passwordMatches method to delegate responsibility of the actual verification to the modHash implementation. This is to prevent against time-based attacks when comparing hashes, also detailed in the linked RFC. The current MD5 and modPBKDF2 hash implementations use the verify() method of modHash to preserve existing behaviour there.

The modNative hasher does not use salts the same way modMD5/modPBKDF2 do, as the hash itself includes a salt (automatically and securely generated by password_hash) which is present in the returned hash for database persistence.

### Why is it needed?

We're currently relying on modPBKDF2 for password hashing. While to the best of my knowledge that is still considered safe (and it also uses salts), we need to keep up with crypto to remain secure. 

These native functions are future proof as we rely on the PHP default algorithm, which can be changed in future versions. Currently that's bcrypt, but [Argon2i was added in PHP 7.2](http://php.net/manual/en/function.password-hash.php) so that may be switched out in the future.

### Next steps

- To make the core actually uses this new hasher, we need to make modNative the default hash_class for users, per #13927. 
- We also need to think about a migration path for existing user accounts. See the request for comments in #13928.
- As password_hash may return longer hashes in the future, PHP recommends expecting hashes of up to 255 characters even though it will only return hashes of 60 characters in the forseeable future. See #13929.

### DIFF
--- a/core/model/modx/hashing/modnative.class.php
+++ b/core/model/modx/hashing/modnative.class.php
@@ -5,6 +5,10 @@
  * @subpackage hashing
  */
 
+if (!function_exists('password_hash')) {
+    require MODX_CORE_PATH . 'model/lib/password.php';
+}
+
 /**
  * A PHP native password_hash/password_verify implementation of modHash.
  *
--- a/core/model/modx/hashing/modnative.class.php
+++ b/core/model/modx/hashing/modnative.class.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * This file contains a modHash implementation of PHP's native password_hash/password_verify handling.
+ * @package modx
+ * @subpackage hashing
+ */
+
+/**
+ * A PHP native password_hash/password_verify implementation of modHash.
+ *
+ * {@inheritdoc}
+ *
+ * @package modx
+ * @subpackage hashing
+ */
+class modNative extends modHash {
+    /**
+     * Generates the hash for the provided string using PHP's password_hash function.
+     *
+     * @param string $string A string to generate a secure hash from.
+     * @param array $options An array of options to be passed to the hash implementation.
+     * @return mixed The hash result or false on failure.
+     */
+    public function hash($string, array $options = array()) {
+
+        return password_hash($string, PASSWORD_DEFAULT);
+    }
+
+    /**
+     * Verifies with PHP's native password_verify function that the provided hash in $expected matches the
+     * raw (unhashed) $string.
+     *
+     * @param string $string
+     * @param string $expected
+     * @param array $options
+     * @return bool
+     */
+    public function verify($string, $expected, array $options = array())
+    {
+        return password_verify($string, $expected);
+    }
+}
--- a/_build/test/Tests/Model/Hashing/modNativeHashTest.php
+++ b/_build/test/Tests/Model/Hashing/modNativeHashTest.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * MODX Revolution
+ *
+ * Copyright 2006-2014 by MODX, LLC.
+ * All rights reserved.
+ *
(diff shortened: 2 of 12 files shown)


opengeek (dismissed):
Great contribution, Mark.

Omeryl:
I have a concern:

> These native functions are future proof as we rely on the PHP default algorithm, which can be changed in future versions. Currently that's bcrypt, but Argon2i was added in PHP 7.2 so that may be switched out in the future.

Will the PHP functions be smart enough to compare old hashes with bcrypt if they do switch it out for Argon2i in the future?

opengeek:
@Omeryl the native password hashing stores the algorithm as part of the hash, so that is irrelevant.

Omeryl:
@opengeek That's all I needed to know, and looking at the docs: `This function is safe against timing attacks.`

So.. super plus from me.

core/model/modx/hashing/modnative.class.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * This file contains a modHash implementation of PHP's native password_hash/password_verify handling.
+ * @package modx
+ * @subpackage hashing
+ */
+
+/**
+ * A PBKDF2 implementation of modHash.
```
pbowyer: This comment needs updating. Otherwise looks good!

Mark-H: Good catch!

Mark-H:
Sorry @opengeek - had to fix the comment. Super important stuff.

OptimusCrime (dismissed):

opengeek:
Integrators: do not merge this — I am going to pull this in for 2.7

Jako:
@opengeek Will you work on this or should we set the milestone back to 3.x?

opengeek:
I'll work on it this week.

Mark-H:
I've updated the PR to be based on latest 2.x instead of 3.x to make merging a little easier.

Mark-H:
Tests are failing (E_FATAL) because password_hash isn't available on PHP 5.4. Any thoughts on the best way to deal with that? Could wrap it in a function_exist but that's code smell - and the tests are still going to fail on 5.4.

alroniks:
@Mark-H Maybe mark a test as skipped from the test itself?

OptimusCrime:
Can you just remove the travis tests for PHP 5.4? If that is not going to be supported either way. Remove the tests, and set the requirements in composer.json. Might need to rebase this (or other) PRs though.

Mark-H:
Given 2.x officially still supports 5.3, I'm not sure removing 5.4 from the tests is a great idea... this may be why I started on it for 3.x in the first place 😬

OptimusCrime:
Well, with this PR, it can not still support 5.3, which is kind of the point with these tests :P 

But there is a polyfill for the missing functions, I think? We could consider including that.

Edit: Oh, the support for this was introduced in 5.5, I thought it was 5.4. Hm, that is a problem then. Perhaps targetting 3.0 will make more sense?

JoshuaLuckers:
Target 3.x so we can finally start to remove support for older php versions.

alroniks:
Seems it's a time when code requires bumping supporting PHP version. Go to 5.5 as minimal required!

PR #13930, opened 2018-06-11
Adds a RewriteRule to the example htaccess

### LINKED ISSUE #13925: Add a RewriteRule for blocking hidden directories to the example htaccess
### Summary
Block hidden directories (except for .well-known) by adding a rewrite rule to the example htaccess.
```
RewriteRule "/\.|^\.(?!well-known/)" - [F]
```

### Step to reproduce
Download/install MODX via Git.

### Observed behavior
Dot directories, also known as "hidden directories" like .git are present at the file system and might be exposed to the public.

### Expected behavior
Hidden directories should not be exposed to the public.

### Environment
MODX 2.\*, MODX 3.0

JoshuaLuckers:
### What does it do?
Adds a RewriteRule to hide dot directories from the public.

### Why is it needed?
To make things a bit more secure.

### Related issue(s)/PR(s)
Fixes #13925

### DIFF
--- a/ht.access
+++ b/ht.access
@@ -13,6 +13,11 @@ RewriteBase /
 
 
 
+# Prevent dot directories (hidden directories like .git) to be exposed to the public
+# Except for the .well-known directory used by LetsEncrypt a.o
+RewriteRule "/\.|^\.(?!well-known/)" - [F]
+
+
 # Rewrite www.domain.com -> domain.com -- used with SEO Strict URLs plugin
 #RewriteCond %{HTTP_HOST} .
 #RewriteCond %{HTTP_HOST} ^www.(.*)$ [NC]


Jako:
IMHO it makes more sense not to deploy those folders to the host.

JoshuaLuckers:
I agree but better be safe then sorry.

alroniks:
I think it does not something dangerous for usual websites but makes the website a bit more secured. So I vote for this.

Jako:
Maybe it should be disabled by default.

Mark-H:
It's in an example file that people need to copy to use, so, it's indeed disabled by default ;)

Jako:
Yep, but the rule itself not. Since it is a BC it should either be disabled or targeted to 3.x.

JoshuaLuckers:
When updating an existing install the .htaccess file (if any is present) will not be overwritten. It's a manual operation.

Mark-H (approved):
Before this is merged someone the integrator should do a double check to make sure it behaves as expected (mostly, does `.well-known/foo.txt` remain accessible, but is `.git/foo.text` locked down), but from a code review point of view I'm happy to give this my stamp of approval. 

Thanks Joshua!

JoshuaLuckers:
@Mark-H I did a double check myself and it seems to work fine. 
@Alroniks if you have time, will you also took a look at this? :)

alroniks:
@JoshuaLuckers Need to find Apache anywhere at first :) Everywhere nginx used.

OptimusCrime:
@Alroniks If you have Docker, then my [docker setup](https://github.com/OptimusCrime/modx-development-docker) might help you.

alroniks (approved):
Tested with Apache web server, all works fine. The content of .well-known folder ignored by rules, so scripts that requires this folder will work.

PR #13931, opened 2018-06-14
Prefer ampersand replacement of the the translit class

Jako:
### What does it do?
Move the ampersand replacement after running the translit service

### Why is it needed?
Before an ampersand replacement in the translit service was not possible

### Related issue(s)/PR(s)
None known


Mark-H:
Relevant somewhat to #13932 as well.

Mark-H (approved):

alroniks (approved):

PR #13932, opened 2018-06-14
Add iconv_ascii transliteration

Jako:
### What does it do?
Add a new iconv_ascii transliteration

### Why is it needed?
The current iconv transliteration does only a transliteration to the same charset. That way i.e. UTF-8 chars were not transliterated to ASCII chars. Transliterating them to ASCII would change i.e. ä to ae with a de_DE.utf8 locale setting.

### Related issue(s)/PR(s)
None known


Mark-H (approved):
Thank you!

alroniks (approved):

PR #13934, opened 2018-06-25
Design updates

bezumkin:
### What does it do?
- Added new system setting "manager_logo"
- Design updates

### Why is it needed?
To release MODX 3

### Related issue(s)/PR(s)
#13875

Don't forget to disable system setting `compress_css` before test this PR!

screen1


ghost:
Sweet!

manager/assets/modext/core/modx.layout.js
```diff
@@ -287,6 +288,19 @@ Ext.extend(MODx.Layout, Ext.Viewport, {
...
+
+                        var header = Ext.get('modx-leftbar-header');
+                        if (header) {
+                            var html = '';
+                            if (MODx.config.manager_logo !== '' && MODx.config.manager_logo !== undefined) {
+                                html += '<img src="' + MODx.config.manager_logo + '">';
+                            }
+                            html += '<span>' + MODx.config.site_name + '</span>';
+                            header.dom.innerHTML = html;
+                            header.on('click', function() {
```
Mark-H: Can we just make this a plain old link so it's accessible and things like ctrl+click work out of the box?

Mark-H (changes requested):
Link should be a link, as commented.

bezumkin:
@Mark-H problem with js link solved

Mark-H (approved):

Mark-H:
Deleting it is not really the same as turning it into a proper link, but I guess that does do away with the a11y concerns of it not being the right semantic element too ¯\_(ツ)_/¯

alroniks (approved):

alroniks (approved):

Issue #13946, opened 2018-07-03
MODX3: Errors in log all displaying on one line. (Firefox)


muzzwood:
### Summary
All errors are displaying on a single line in the error log.

### Step to reproduce
View the error log page when errors have been logged.

### Observed behavior
There are no new lines / line breaks

### Expected behavior
Each error should be on it's own line.

### Environment
MODX 3.0 alpha. PHP 7.1, Firefox 60

gadgetto:
Hmm, can't reproduce this. In my environment it's fine. But I got the MODX 3 version from @bezumkin which seem a bit ahead of the official one.

muzzwood:
Hmmm ok. This is the zip download with dependencies included.

gadgetto:
I'm testing this version: https://cloud.2head.ru/s/iRbxnxcNwXHTzc6 which is the full build with all dependencies included.

muzzwood:
Actually I just tried with Chrome and it works fine. It seems to be a Firefox specific issue.

muzzwood:
This is Firefox 60 running on Ubuntu 18.04
If you're on a different platform, could you try with Firefox and see if it's platform specific?

muzzwood:
Removing `white-space:nowrap` fixes the issue. 
However it also changes the behaviour in Chrome so that each line wrapped rather than overflow.

shamarin:
Confirm, in Firefox log error is displaying in one line.

jonleverrier:
Confirmed too, that in FF the error log is displaying on one line.

muzzwood:
Thanks for confirming, will attempt a PR.

muzzwood:
Closing due to PR  #13966.

muzzwood:
Re-opening issue until PR is merged.

Jako:
I don't think that this is a pure CSS issue. That nowrap attribute makes sense because locating a new log line is easier then.

Please locate why Firefox does not wrap after one log line.

alroniks:
PR is merged, so closing.

PR #13966, opened 2018-07-04
Fixed error log line wrapping problem in Firefox.

muzzwood:
### What does it do?
In latest 3.x when using Firefox, all errors were appearing on a single line.

### Why is it needed?
Each error should have its own line.

### Related issue(s)/PR(s)
This is related to issue #13946

Please check in at least Firefox and Chrome to confirm it's fixed appropriately.

### DIFF
--- a/manager/assets/modext/widgets/system/modx.panel.error.log.js
+++ b/manager/assets/modext/widgets/system/modx.panel.error.log.js
@@ -34,7 +34,7 @@ MODx.panel.ErrorLog = function(config) {
                     ,grow: true
                     ,anchor: '100%'
                     ,hidden: config.record.tooLarge ? true : false
-                    ,style: 'white-space: nowrap; overflow: auto;'
+                    ,style: 'overflow: auto;'
                 },{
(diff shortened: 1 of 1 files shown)


Jako (changes requested):
I don't think that this is a pure CSS issue. That nowrap attribute makes sense because locating a new log line is easier then.

Please locate why Firefox does not wrap after one log line.

muzzwood:
@jako So you're saying you want each line to overflow horizontally?

muzzwood:
PR update: I've replaced `white-space:nowrap;` with `white-space:pre;` which also fixes the issue but allows each line to extend beyond the width of the textarea without wrapping. Tested in both Firefox and Chrome.

Jako:
@digitalpenguin Yes, it was introduced in https://github.com/modxcms/revolution/commit/0d51102639fb2c531c1aac0cfb158c60b78cb22d#diff-b3350acaa6824ba055033cb0fb57a2df

Jako (approved):

Mark-H (approved):

PR #13980, opened 2018-07-07
Preventing deleting any file in registry by mistake and filtering paths in topics

alroniks:
### What does it do?
It fixes some issues with path traversal in the cache manager and determining file extensions for deleting.

### Why is it needed?
It is a security fix, let's discuss details personally in Slack.

### Related issue(s)/PR(s)
none

### DIFF
--- a/core/model/modx/registry/modfileregister.class.php
+++ b/core/model/modx/registry/modfileregister.class.php
@@ -64,7 +64,10 @@ class modFileRegister extends modRegister {
      */
     public function clear($topic)
     {
-        $topicDirectory = $this->directory . ltrim(basename($topic), '/');
+        /** @var modFileHandler $fileHandler */
+        $fileHandler = $this->modx->getService('fileHandler', 'modFileHandler');
+
+        $topicDirectory = $this->directory . ltrim($fileHandler->sanitizePath($topic), '/');
 
         return $this->modx->cacheManager->deleteTree(
             realpath($topicDirectory),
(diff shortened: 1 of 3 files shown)


alroniks:
@opengeek @Mark-H @Jako 
I've updated PR. Please, have a look asap, because it is a security issue.

Mark-H (approved):

alroniks:
@OptimusCrime Hi, could you have a look?

core/model/modx/registry/modfileregister.class.php
```diff
@@ -263,4 +272,14 @@ public function send($topic, $message, array $options = array()) {
...
     }
+
+    /**
+     * Sanitize the specified path
+     *
+     * @param string $path The path to clean
+     * @return string The sanitized path
+     */
+    protected function sanitizePath($path) {
+        return preg_replace(array("/\.*[\/|\\\]/i", "/[\/|\\\]+/i"), array('/', '/'), $path);
```
OptimusCrime, line 283: Any reason the two expressions can not be merged to `/\.*[\/|\\]+/i`? It should match both `../`, `..////` and `////`. You might also want to consider an expression such as `(?:\.*[\/|\\]+)+`. This matches `../../....//////` too.

OptimusCrime:
Looks fine to me. The expression can be improved/combined I think. I do not have the security issue available, so I do not know just what the attack vector is.

opengeek (dismissed):
In my opinion, all files in the specified registry topic should be removed, so I find the extensions part of this extraneous. I'm going to submit a security fix that addresses only the path traversal issue. We can discuss the extensions issue later. Regardless, the change to _build/transport.core.php should not be in here.

Issue #13976, opened 2018-07-05
MODX 3: Login - no css style for .is-success on forgotten password


jonleverrier:
### Summary
A non styled success message displays if you enter a user that exists in the system.

![modx_login_sucess](https://user-images.githubusercontent.com/2373940/42332161-a076ca64-8077-11e8-917c-2bbe7d01a92f.gif)

### Step to reproduce
1. Go to the forgotten password tool on the login
2. Type a user that exists in the system, 
3. After pressing "Send Activation Email" a non styled success message displays.

### Observed behavior
A non styled success message displays.

### Expected behavior
For a styled success message to display. Although if they are colour coded - their is a minor security issue; somebody might be able to tell that a user does exist (green success message) or does not exist (red success message) in the system. Maybe develop a neutral message colour (although they could still check the generated html for `.is-success` vs `.is-error`.

### Environment
MODX3 Alpha v 3.0.0-dev
From https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99

OptimusCrime:
Please reopen this

alroniks:
Closing as PR is merged.

Issue #13950, opened 2018-07-03
MODX 3: Login - No hover style on login button


jonleverrier:
### Summary
I appreciate this is a small detail...

When you hover over the Login button, no hover style is triggered.

### Step to reproduce
Hover over the Login button.

### Observed behavior
No hover style is triggered.

### Expected behavior
For consistency, a hover style should be triggered like the other green buttons used within the manager (as an example, like the save button on a resource page).

### Environment
MODX3 Alpha v 3.0.0-dev
From https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99

Button hover
![screen shot 2018-07-03 at 11 10 59](https://user-images.githubusercontent.com/2373940/42210643-cc64c47c-7eb1-11e8-8518-0c94af426b9c.png)

Button
![screen shot 2018-07-03 at 11 10 56](https://user-images.githubusercontent.com/2373940/42210644-cc82915a-7eb1-11e8-9621-0341fedc83f1.png)

Login button on hover
![screen shot 2018-07-03 at 11 10 31](https://user-images.githubusercontent.com/2373940/42210646-cca18a2e-7eb1-11e8-83ee-483abf313757.png)

meshkov:
Why are you closing?

alroniks:
Closing as PR is merged.

PR #13984, opened 2018-07-09
Fix hover style on login button, forget password button

Ibochkarev:
### What does it do?
Added css style for hover state the login button in login.css

### Why is it needed?
Add behavior rules when hovering over a login button
It's a fix for issue #13976 #13950 

### Related issue(s)/PR(s)
#13976 #13950


OptimusCrime:
I think you should edit and commit the scss files, not the css ones.

Ibochkarev:
Hi.

Tell me, please, where to get the files scss. I will correct them

jonleverrier:
@Ibochkarev - @Mark-H says they're based in `_build/templates/default`

alroniks:
@Ibochkarev Hi. This file https://github.com/modxcms/revolution/blob/3.x/_build/templates/default/sass/login.scss

_build/templates/default/sass/login.scss
```diff
@@ -217,6 +217,17 @@ input[type="checkbox"] {
   }
 }
 
+
+#modx-login-btn {
+  border: 1px solid transparent;
+  &:hover {
```
Mark-H: Maybe worth turning this into `&:hover, &:focus {` to also cover keyboard users? Or is there already an outline shown for those?

Ibochkarev: Ok, im fix

Mark-H (dismissed):
Thanks!

_build/templates/default/sass/login.scss
```diff
@@ -217,6 +217,21 @@ input[type="checkbox"] {
   }
 }
 
+#modx-login-btn {
+  border: 1px solid transparent;
+  &:hover {
+    background-color: $primary1;
```
alroniks: It seems there should be $primary2 color.

alroniks (changes requested):
Also, here still an error styles instead success message in the password reset form.

Ibochkarev:
@Mark-H @Alroniks  Add Add button#modx-fl-btn hover state 

Example:
Fred settings

alroniks (dismissed):

OptimusCrime:
You seem to have included code related to the slack badge icon in the README.md file.

Ibochkarev:
@OptimusCrime @Alroniks please test PR.

_build/templates/default/sass/login.scss
```diff
@@ -217,6 +217,36 @@ input[type="checkbox"] {
...
+  &:focus{
+    box-shadow: none;
+  }
+}
+
+button#modx-fl-btn {
+  border: 1px solid transparent;
+  &:hover {
+    background-color: $white;
+    color: $primary1;
```
alroniks: Buttons look the same, but here color $primary1 when in login button $primary2. Why? I guess it should be the same for both buttons.

_build/templates/default/sass/login.scss
```diff
@@ -217,6 +217,36 @@ input[type="checkbox"] {
...
+  &:hover {
+    background-color: $white;
+    color: $primary2;
+    box-shadow: none;
+    border: 1px solid $primary2;
+    transition: color .2s ease-out,
+    background-color .2s ease-out,
+    border .2s ease-out;
+  }
+  &:focus{
```
alroniks: Mark means that styles for the focused element should be the same as for the hover state. It needed for people who use a keyboard to navigate across the page instead of a mouse. So it should be:
```
&:hover, &:focus {
    ... styles
}
```

_build/templates/default/sass/login.scss
```diff
@@ -217,6 +217,36 @@ input[type="checkbox"] {
   }
 }
 
+#modx-login-btn {
```
alroniks: CSS rules seem absolutely identical for both buttons, so here unnecessary duplication of the code.
Should be:
```
#modx-login-btn,
#modx-fl-btn {
 ... styles here
}
```

_build/templates/default/sass/login.scss
```diff
@@ -217,6 +217,36 @@ input[type="checkbox"] {
...
+    transition: color .2s ease-out,
+    background-color .2s ease-out,
+    border .2s ease-out;
+  }
+  &:focus{
+    box-shadow: none;
+  }
+}
+
+button#modx-fl-btn {
```
alroniks: Here already selector by ID, no need to add the listing of all buttons before finding a proper element with the specified identifier. And should be combined with the first rule.

alroniks (changes requested):

alroniks:
Ops.

Ibochkarev:
Thats ok. ;)

Issue #13982, opened 2018-07-09
Package update fails if Update button was pressed twice


ant-tar:
### Summary
Package update fails if  Update button was pressed twice

### Step to reproduce
It's a bit curious and I believe not everyone can be so tricky :) but I had wide tree section so popup had a bit another positioning..I've tried to repeat that with standard width for left part and this is unlikely, but possible. So I wanted to close the popup but missed the goal and pressed the Update button again. After that as you can see from below some strange popup cycling happened, you can't update any package correctly until you reload the page.

### Observed behavior
Here is animated GIF how it was, hope it's clear:

![button_popup_issue](https://user-images.githubusercontent.com/2815640/42443068-262d0eca-8375-11e8-8430-7ffeba3161b6.gif)

### Expected behavior
The update package process should continue correctly.

### Environment
MODX 2.6.4-pl on MODXCloud

JoshuaLuckers:
Thanks for taking time to report this issue. The issue is clear and I'm able to reproduce it.

PR #13985, opened 2018-07-10
Fix package update window not working if update button was pressed twice

JoshuaLuckers:
### What does it do?
Don't force the package update window to open to prevent issues when the update button has been clicked multiple times.

### Why is it needed?
It's a fix for issue #13982 

### Related issue(s)/PR(s)
issue #13982

### DIFF
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -361,6 +361,9 @@ Ext.extend(MODx.grid.Package,MODx.grid.Grid,{
 
 	/* Search for a package update - only for installed package */
     ,update: function(btn,e) {
+        if (this.windows['modx-window-package-update']) {
+            this.windows['modx-window-package-update'].destroy();
+        }
         MODx.Ajax.request({
             url: this.config.url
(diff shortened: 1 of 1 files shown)


alroniks (dismissed):

Jako (dismissed):

ant-tar:
Thanks @JoshuaLuckers for this fix. Yes we don't have cycling now, but after double click I can't call update popup again at all, even after reverted installation try. Please check animated gif below: 

popup_issue_new

JoshuaLuckers:
Thanks @himurovich for the feedback. ~I will try to see if I can fix it this weekend.~

I hope it's fixed now.

JoshuaLuckers:
@Jako Good suggestion. In this case the window was always forced to open (the window didn't close so it could not be destroyed). I will check if your suggestion is a solution for this problem.

JoshuaLuckers:
@Jako I tried both solutions but it does not solve the problem described in issue #13982. The current changes do seem to fix the issue even if it might not be the best solution.

Jako (approved):

PR #13986, opened 2018-07-10
Fix modx.class.php's sendError() always has "503 Error" as the headline

### LINKED ISSUE #12379: modx.class.php's sendError() always has "503 Error" as the headline
Whenever some core error page gets generated it shows a "503 Error" headline even if it is really a 404 error.

JoshuaLuckers:
### What does it do?
When the $options['error_header'] value is set for $modx->sendError() the page "core/error/unavailable.include.php" should not override this header.

### Why is it needed?
To fix a 3 year old issue.

### Related issue(s)/PR(s)
Fixes #12379

### DIFF
--- a/core/error/unavailable.include.php
+++ b/core/error/unavailable.include.php
@@ -1,5 +1,7 @@
 <?php
-header($_SERVER['SERVER_PROTOCOL'] . ' 503 Service Unavailable');
+if (isset($errorHeader) === false) {
+    header($_SERVER['SERVER_PROTOCOL'] . ' 503 Service Unavailable');
+}
 ?>
 <html>
 <head>
--- a/core/error/unavailable.include.php
+++ b/core/error/unavailable.include.php
@@ -1,5 +1,5 @@
 <?php
-if (isset($errorHeader) === false) {
+if (!isset($errorHeader)) {
     header($_SERVER['SERVER_PROTOCOL'] . ' 503 Service Unavailable');
 }
 ?>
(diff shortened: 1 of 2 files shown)


alroniks:
I think it should be targeted to 2.x because it does not look something that's breaking backward compatibility.

core/error/unavailable.include.php
```diff
@@ -1,5 +1,7 @@
 <?php
-header($_SERVER['SERVER_PROTOCOL'] . ' 503 Service Unavailable');
+if (isset($errorHeader) === false) {
```
OptimusCrime: [isset](http://php.net/isset) returns a `bool`, so there is no need for the strict comparison to `false`. `if (!isset($errorHeader))` should be enough.

Mark-H (dismissed):

Jako (dismissed):

core/error/unavailable.include.php
JoshuaLuckers: Good point. I have to be more consistent with those things.

@Jako @Mark-H Sorry guys, but can you re-approve this PR?

Jako (approved):

OptimusCrime (approved):

opengeek:
Technically, the unavailable file should not be called unless it is a 503. Perhaps we need to introduce a default error type file instead? Any thoughts?

alroniks (approved):
I've experimented with this PR and it works as expected, but the user should set correct own header if he wants to get it works. 
For example: setting in system settings `404 Not found` doesn't work but `HTTP/1.0 404 Not found` works.

JoshuaLuckers:
The templates for the error types "fatal" and "unavailable" are now identical (only the message is different). Both are using an error code in the 5xx group.

Maybe a template named "error500.include.php" can be used for those type of errors.
Additionally there could be a default "error.include.php" template (with a "friendlier" theme) and a "error400.include.php" etc.

alroniks:
I approved the changes, but I guess it can be improved more. Now we have two files `fatal.include.php` and `unavailable.include.php` that looks a bit messy in the core folder. Maybe it makes sense to think more on it, as Jason asked.

JoshuaLuckers:
I totally agree but it might be better if we can discuss this in a new issue.

Mark-H (approved):

PR #13987, opened 2018-07-10
Login forgotten password improvements

JoshuaLuckers:
### What does it do?
If a password reset is done for a non-existing user it also shows the success message. 
The style for is-success is also being added.

### Why is it needed?
It's a fix for issue #13976

### Related issue(s)/PR(s)
#13976

### DIFF
--- a/manager/controllers/default/security/login.class.php
+++ b/manager/controllers/default/security/login.class.php
@@ -259,11 +259,11 @@ class SecurityLoginManagerController extends modManagerController {
                 $this->modx->log(modX::LOG_LEVEL_ERROR,$err);
                 $this->setPlaceholder('error_message',$err);
             } else {
-                $this->setPlaceholder('error_message',$this->modx->lexicon('login_password_reset_act_sent'));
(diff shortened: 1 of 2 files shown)


Mark-H:
This might cause a conflict with part of #13984 depending on what is merged in what order. This PR seems to fix the problem where the response might indicate if a user existed or not (which would be seen as a potential security issue) while both address missing styling for success messages.

JoshuaLuckers:
Didn't see that PR. I will remove my changes for the style change.

opengeek:
Why is this not targeting 2.x?

JoshuaLuckers:
@opengeek solved

alroniks (approved):

alroniks (changes requested):
It seems that you should bring from 3.x code from the template, because now message disappears at all because in curent template not user placeholder `success_message`.
It may cause conflicts after but it shouldn't bring new bugs into 2.x I guess.

JoshuaLuckers:
Good catch @Alroniks! I was too rushed and did not check it myself. Thanks.

Mark-H (approved):

alroniks (approved):

PR #13992, opened 2018-07-13
Added smart detecting of used plugins for expanding rows in grids

### LINKED ISSUE #13989: CollapseAll doesn't work in grids when RowExpander plugin used
### Summary
When the grid uses custom row expander attempts to expand or collapse all rows triggers an error that collapseAll is not a function (i.e. method not found).

### Step to reproduce
Set up the grid
```
this.exp = new Ext.ux.grid.RowExpander({
    tpl: new Ext.Template(nlInfos)
    ,enableCaching: false
});
```
and call after anywhere method `collapseAll()` from grid. For example, in any listener:
```
listeners: {
    'check': function(cbx,checked){
        this.collapseAll();
        this.toggleAutoRefresh(cbx,checked);
    }
    ,scope:this
}
```

### Observed behavior
Now it raises an error `Uncaught TypeError: this.exp.collapseAll is not a function`

### Expected behavior
It should work as before, collapsing/expanding all rows.

### Environment
2.6.4, I guess 3.x also affected, but I don't check yet.

The reason why it happens here - https://github.com/modxcms/revolution/commit/26583acc454bf85f0f28d53005fa8012984b7a98
Plugin RowExpander was updated and in updated code expandAll, collapseAll method removed.

Detailed description on the forum - https://forums.modx.com/thread/104033/extjs-problem-with-grid-rowexpander-plugin-collapseall-doesn-039-t-work#dis-post-559499

### LINKED ISSUE #13991: ExtJS RowExpander plugin: expanded rows looses it's row body on grid refresh (with enableCaching: false)
### Summary
We have a MODX grid with rowExpander plugin enabled and `enableCaching: false`. If row is expanded and the grid is refreshed, the expanded row seems to collapse but the [+] sign stays at [-]. You then have to click the [-] and then the [+] to re-expand the row.

This problem is also related to: #13989

### Step to reproduce
Setup a MODX grid and create a rowExpander:

```
    // Content of the expanded grid row
    var nlInfos = [
        '<table id="gon-nlinfo-{id}" class="gon-expinfos">',
            '<tr>',
                '<td class="gon-expinfos-key">'+_('goodnews.id')+'</td><td class="gon-expinfos-val">{id}</td>',
                '<td class="gon-expinfos-key">'+_('goodnews.newsletter_createdon')+'</td><td class="gon-expinfos-val">{createdon_formatted}</td>',
                '<td class="gon-expinfos-key">'+_('goodnews.newsletter_publishedon')+'</td><td class="gon-expinfos-val">{publishedon_formatted}</td>',
                '<td class="gon-expinfos-key">'+_('goodnews.newsletter_sender')+'</td><td class="gon-expinfos-val">{sentby_username}</td>',
                '<td class="gon-expinfos-key">'+_('goodnews.newsletter_scheduled')+'</td><td class="gon-expinfos-val gon-scheduled">{pub_date_formatted}</td>',
            '</tr>',
        '</table>'
        ].join('\n');

this.exp = new Ext.ux.grid.RowExpander({
    tpl: new Ext.Template(nlInfos)
    ,enableCaching: false
});
```

**Please note: if you set `enableCaching: true` it works as expected - except that the fields in the expanded row aren't updated!**

Bind it to the MODX grid:

```
        ...
        ,plugins: [this.exp]
        ,columns: [
        this.exp
        ,{
            header: _('goodnews.id')
            ,dataIndex: 'id'
            ,hidden: true
        ...
```

### Observed behavior
If row is expanded and the grid is refreshed, the expanded row seems to collapse.

### Expected behavior
The row should stay expanded on grid refresh and the expanded content should be refreshed too.

### Environment
MODX version: 2.6.4-pl
Browser: tested in Safari and Chrome

alroniks:
### What does it do?
It solves some issues around row expander in grids.

### Why is it needed?
It solves some issues around row expander in grids. 
One issue - after upgrading RowExpander component methods collapseAll and expandAll were lost. This PR fixes this.
The second issue that this PR solves - the hardcoded name of the expander (`thix.exp`) and inability to use own name for the plugin.

Result of work:

expandcollapseall

### Related issue(s)/PR(s)
Fixes #13991 
Fixes #13989

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -488,6 +488,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
         }
         return Ext.encode(rs);
     }
+
     ,encode: function() {
         var p = this.getStore().getRange();
         var rs = {};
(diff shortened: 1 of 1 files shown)


alroniks:
Hi @gadgetto 
Please, have a look. It should fix your issues (only after release of course)

gadgetto:
Just tested with your fixed code and I can confirm - it works! Cool!

JoshuaLuckers:
I see some code duplication. Is it possible to move the duplicate code to a function?

alroniks:
@JoshuaLuckers I also don't like it, but there are two different type of grids and it seems it is impossible to extend one from another, at least it is not easy and requires more massive refactoring that can cause new issues with code that use this functionality. But I am open to suggestions.

JoshuaLuckers:
Ah, that makes sense. I don't have any better solution at this moment. 
I was just wondering why, thanks!

alroniks:
@Mark-H @jako @OptimusCrime @opengeek What do you think? (Sorry for mentions and annoying)

OptimusCrime:
This looks fine by me. Nice fix!

Mark-H (approved):

Jako (approved):

PR #13993, opened 2018-07-14
[ReUp] Update the copyright year in various files

### LINKED ISSUE #13872: Demo content copyright date incorrect
### Summary
The demo content created with a new installation contains the copyright year 2016. This should be updated to the correct year, or better, made dynamic to always show the current year.

### Step to reproduce
Install MODX Revolution.

### Observed behavior
See incorrect copyright year on the default resource on the front end.

### Expected behavior
One would expect to see the current year in the copyright notice.

### Environment
All new MODX Revolution installations where the demo content is applied to the default single resource that is created.

JoshuaLuckers:
### What does it do?
Updates the copyright year to 2018 in various files where the copyright year is (still) present

### Why is it needed?
With a new major release it's a good idea to also update the copyright year.

### Related issue(s)/PR(s)
Fixes #13872

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 MODX Revolution
 
-Copyright 2006-2016 by MODX, LLC.
+Copyright 2006-2018 by MODX, LLC.
 
 All rights reserved.
 
--- a/_build/docs/themes/modx/template.xml
+++ b/_build/docs/themes/modx/template.xml
@@ -2,7 +2,7 @@
 <template>
   <author>MODX CMF</author>
   <version>1.0</version>
-  <copyright>2011-2014 MODX, LLC</copyright>
+  <copyright>2011-2018 MODX, LLC</copyright>
   <transformations>
     <transformation query="copy" writer="FileIo" source="ajax_search.php" artifact="ajax_search.php"/>
     <transformation query="copy" writer="FileIo" source="js" artifact="js"/>
--- a/_build/templates/default/sass/_xtheme-modx.scss
+++ b/_build/templates/default/sass/_xtheme-modx.scss
@@ -1,6 +1,6 @@
 /**
  * xtheme-modx
- * Copyright (c) 2009-2014, MODX, LLC
+ * Copyright (c) 2009-2018, MODX, LLC
  *
  * xtheme-modx is licensed under the terms of the GNU Open Source GPL 2.0
  * or later license.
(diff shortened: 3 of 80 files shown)


Mark-H (dismissed):

opengeek (commented):
I actually would like to see the dates removed from all the files except the LICENSE.md file and change the headers to simply refer to the LICENSE.md file. Continually having to update these dates is disruptive to identifying important diffs.

JoshuaLuckers:
Good point. I will change the headers if that’s okay with you.

opengeek:
Let's agree on a format.

opengeek:
For PHP files:

```
/*
 * This file is part of the MODX Revolution package.
 *
 * Copyright (c) MODX, LLC
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */
```

opengeek:
Also, these needs to go into 2.x.

JoshuaLuckers:
Fine for me! Will do it tomorrow.

JoshuaLuckers:
I've added the header with the copyright and license information to most of the files where it was missing. 
For 2.x I will open another PR (and if it's ok only replacing the existing headers with the "new" header instead of also adding it to all the other files).

opengeek:
Well, the reason I said it needed to be done on 2.x is that we are going to have to merge 2.x into 3.x and if you commit this directly to 3.x AND 2.x it is going to be a conflict nightmare.

JoshuaLuckers:
Good point. Didn’t think about that 😅. I will make the changes for 2.x in the upcoming days. When it’s merged into 3.x I will check if there are any files without the header and add them.

alroniks (approved):

alroniks:
I think it can be safely merged. Not code or functionality touched, only comments. And your @opengeek requirements followed and implemented.

OptimusCrime:
By the way, the GitHub standard is to have the file named LICENSE, not LICENSE.md. This will display the license type in the repo. Perhaps worth changing. The link would need to be changed, but that should not be an issue.

JoshuaLuckers:
@OptimusCrime both are fine: [3. In the file name field, type LICENSE or LICENSE.md (with all caps).](https://help.github.com/articles/adding-a-license-to-a-repository/)

OptimusCrime:
Oh, sorry, I did not know. However, it does not look like GitHub has picked up on the license in this repo yet. It is not shown in the information header, but perhaps that is because our license is altered?

JoshuaLuckers:
It's indeed not listed because the format of the LICENSE file is different from the original license text. (it does link to the full license).

PR #13994, opened 2018-07-16
Correct display of permissions in access settings for a group of users

### LINKED ISSUE #7316: Inactive permissions are saved in access policy
gadamiak created Redmine issue ID 7316

Default access policies contain only active permissions, ie. those which are set to @true@. However, custom policies contain all permissions, where active are set to @true@ and inactive are set to @false@. While MODX permissions are cumulative, ie. they sum up what is allowed and don't apply what is not allowed, saving @false@ values is incorrect.

More, when reviewing or re-applying access policy all permissions, including those inactive with @false@ value are displayed, which is misleading.

The fix would be to ensure only active permissions are saved.

### LINKED ISSUE #7974: Policy grid and update policy window show not enabled permissions
gadamiak created Redmine issue ID 7974

This happens on the policy grid when policy is expanded by clicking a 'plus' icon and in update policy window. All permissions from a policy template are listed despite being enabled or not. However when adding a policy only enabled permissions are displayed as expected. See screens.

This makes reviewing permissions for a user group impossible.

alroniks:
### What does it do?
It fixes some presenting issues on the screen of configuring access for user groups.

### Why is it needed?
It needed for admins who configures access for other users in the system and it should decrease the number of mistakes because before the fix they saw wrong data.

### Related issue(s)/PR(s)
Fixes #7974 
Fixes #7316

### DIFF
--- a/core/model/modx/processors/security/access/usergroup/category/getlist.class.php
+++ b/core/model/modx/processors/security/access/usergroup/category/getlist.class.php
@@ -88,10 +88,11 @@
         $data = $this->modx->fromJSON($data);
         if (!empty($data)) {
             $permissions = array();
-            foreach ($data as $perm => $v) {
-                $permissions[] = $perm;
+            foreach ($data as $permission => $enabled) {
+                if (!$enabled) { continue; }
+                $permissions[] = $permission;
             }
-            $objectArray['permissions'] = implode(', ',$permissions);
+            $objectArray['permissions'] = implode(', ', $permissions);
         }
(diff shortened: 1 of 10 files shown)


Mark-H (approved):

alroniks:
@jako @OptimusCrime @opengeek What do you think? (Sorry for mentions and annoying)

opengeek:
Looks like a solid bug fix. Haven't had time to review the details.

Issue #6410, opened 2011-12-18
Allow the Resource/Element trees to be filtered by category


wshawn:
wshawn created Redmine issue ID 6410

This would make life easier on large projects.  In essence allow the display to only reflect a single category or user selectable categories.

alroniks:
I guess this PR solves the described issue (not fully, but at least idea is the same)
https://github.com/modxcms/revolution/pull/13997

Ruslan-Aleev:
@Alroniks Probably we can close this issue.

PR #13997, opened 2018-07-16
Ability to see all elements under selected category in categories tree

### LINKED ISSUE #2899: Combined view of all elements and dividet view for context
theuros created Redmine issue ID 2899
1. I'd like to have an option to see all elements from one category like in modx evolution. 
   With big pages and lot of elements becomes difficult to find them. If you have a lot of chunks wich depends on  snippets .. it's a nightmare to find them one or another for editing ... 
2. If you create a multisite and you  use a lot speciffic elements for each context you get too many of them. It will be nice to have a possibility for elements to define for wich context are  (one contex,more context or general for all) ... and of course to have an ability to display only elements of speciffic context.

### LINKED ISSUE #3934: Display of categories
en8scl created Redmine issue ID 3934

In the Elements window on the left in the manager

The different categories are correctly shown at the bottom of the list
It also correctly shows the number of items in that category

However it does not allow you to expand any category to see the individual items in that category

alroniks:
### What does it do?
It adds an ability to see nested elements under categories in categories tree.

### Why is it needed?
Sometimes it difficult to walk across all elements and needed to see all elements under the specific category to choose one.

See details in the animated gif.

categories

### Related issue(s)/PR(s)
Fixes #3934 
Fixes #2899
#6410
And maybe others.

### DIFF
--- a/core/model/modx/processors/element/getnodes.class.php
+++ b/core/model/modx/processors/element/getnodes.class.php
@@ -16,6 +16,7 @@ class modElementGetNodesProcessor extends modProcessor {
         'plugin' => 'modPlugin',
         'category' => 'modCategory',
     );
+
     public $actionMap = array();
 
     public function checkPermissions() {
@@ -53,11 +54,9 @@ class modElementGetNodesProcessor extends modProcessor {
                 break;
         }
 
-        if ($this->getProperty('stringLiterals',false)) {
-            return $this->modx->toJSON($nodes);
-        } else {
-            return $this->toJSON($nodes);
-        }
+        return $this->getProperty('stringLiterals', false)
+            ? $this->modx->toJSON($nodes)
+            : $this->toJSON($nodes);
     }
 
     public function getActions() {
@@ -77,7 +76,6 @@ class modElementGetNodesProcessor extends modProcessor {
         /* split the array */
         return explode('_',$id);
     }
-
 
     /**
      * Default icons for element types
@@ -222,7 +220,12 @@ class modElementGetNodesProcessor extends modProcessor {
         return $nodes;
     }
 
-    public function getCategoryNodes(array $map) {
+    /**
+     * @param array $map
+     * @return array
+     */
+    public function getCategoryNodes(array $map)
+    {
         if (!empty($map[1])) {
             /* if grabbing subcategories */
             $c = $this->modx->newQuery('modCategory');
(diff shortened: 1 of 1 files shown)


opengeek:
I don't understand the purpose of this at all. Can you please explain better what this does?

alroniks:
@opengeek We have categories under a type of elements and also have a list of categories below. People are awaiting that under this categories they can see elements but actually it is not a thuth. So they can be confused (and they was a lot of time, based on my conversations with customers and users before). So it brings an ability to see elements using category as the start point.

opengeek:
I have no idea what you just said. I'm so confused.

garryn:
Jason, it’s to have the various elements load in the Categories tree as well as under their respective trees. Currently, Categories only expand to show if there are sub-categories.

opengeek:
I figured it out with some help from others in community chat. It is definitely Monday. Sigh.

Ruslan-Aleev:
@Alroniks Hello, Ivan, is it planned to add icons (instead of empty icons) for all elements (chunks, snippets, TV, etc.) that are listed in your gif-example?
It would be much easier to navigate in the tree of elements.

alroniks:
@Ruslan-Aleev In the elements itself, it is unnecessary to add icons because of elements already under sections. I mean snippets in snippets section plugins in plugin sections. In case of categories, when in one category can be different types of elements it makes sense and I've added them, for a code it uses icons for files and in case element is static - special icons for mark element as static.

Ruslan-Aleev:
@Alroniks Yes, there are no questions with the category.
But if there many elements in the tree - the icon of the elements will be useful (an empty icon is already displayed).
Usability increases, though not much :) In any case, thanks.

alroniks:
@Ruslan-Aleev But what icons to use? Set a new field for each item with a specific icon too complicated. It makes sense for templates when it shows resources with the applied template but elements it requires to add a new field in DB and I am sure a small number of people will use it.

Ruslan-Aleev:
@Alroniks No, do not create a field in the DB for the icon, but use the standard icons (icons of the headings category elements, like in your gif).
If the tree of elements is long, it will be easier to understand by the tree which elements are now on the left, it will not be necessary to search for the main category of elements (chunks, snippets, etc.).

Ruslan-Aleev:
@Alroniks On the example of chunks: I immediately see that I am correcting the chunk, and I will not be confused when switching tabs of the browser, that to the left is the tree of chunks. Or i will not search for TV in this list of chunks, but with empty icons I have to look for the parent category of elements to understand that this is not TV list.

chunks_example

alroniks:
@Ruslan-Aleev I see. Could you create an issue and describe your idea there? It should be discussed more and not within the scope of this pull request.

Ruslan-Aleev:
@Alroniks Ok

Mark-H (approved):

PR #13866, opened 2018-04-12
[3.x] Update access policy description to include warnings

sepiariver:
### What does it do?
Updated (English) default lexicon entries for Access Policy Permissions to include security warnings.

### Why is it needed?
Seemingly innocuous permissions such as `edit_tv` can expose passwords and API credentials. While providing Manager access is, in and of itself, a risky endeavour, it's made more risky by the lack of warning or indication when a permission is critical for security.

### Related issue(s)/PR(s)
3.x accesspolicydesc branch

PR #13999, opened 2018-07-17
Update access policy description to include warnings (copy of 13866)

alroniks:
### What does it do?
Updated (English) default lexicon entries for Access Policy Permissions to include security warnings.

### Why is it needed?
Seemingly innocuous permissions such as edit_tv can expose passwords and API credentials. While providing Manager access is, in and of itself, a risky endeavour, it's made more risky by the lack of warning or indication when a permission is critical for security.

### Related issue(s)/PR(s)
It is updated and rebased code from this PR #13866


alroniks:
As a CSS was changed in original PR, I've regenerated theme to update changed CSS file by the proper way from SCSS.

Mark-H (dismissed):

alroniks:
It's recreated PR (that means that it's already discussed before and can be merged), but as I am an author of this one I cannot approve my own PR. So, please, give some kudos! :) @Jako @OptimusCrime

OptimusCrime:
This looks fine by me. Not sure if the `index-min.css` file should be commited though.

alroniks:
It's copy of original PR so as is. By the way it will be resolved during the merging

Jako:
PRs should be submitted without any css to avoid those merge conflicts.

alroniks:
@Jako Usually yes and I know it. But as I said, it's recreated with copied commits and my last commit just only fixed previous conflicts. So I would say that is an edge case and only one such specific PR.

alroniks:
@Jako The problem was also in that fact that changes were made in one commit. So I could not just revert commit. I don't think that we should fight here about rules, especially when all know them, but resolve it by fixing conflict in css files during the merge.

ilyautkin:
PR is tested. Warnings appeared in the _Access Control Lists_:

1yntpv

1. Perhaps we should highlight the notifications in bold? Like this:

1yntuw

2. These descriptions are also displayed on the **policy template** editing page. Should _white-space_ rule be added to `#modx-grid-template-permissions`?

Jako:
The warning strings could be separated from the description strings. That way they could be easier colored red, formatted bold etc.

Jako:
Maybe such a construct is possible:

```
$_lang['perm.edit_context_desc'] = 'To edit any Contexts. <strong class="red">[[%perm.edit_context_warning]]</strong>';
$_lang['perm.edit_context_warning'] = 'WARNING: exposes Context Settings which can override security-related System Settings, such as uploadable file types and API credentials.';
```

alroniks:
> Maybe such a construct is possible:

I tried to implement but it seems that it requires a lot of refactoring existing code and the final result not as good as we would like. The parsing lexicons tags inside lexicons strings look like a big overhead. 

So I would like to leave it as is. Maybe only change the style of warning to bold. Red is too much.

cc @Jako

Mark-H (dismissed):
Looks good!

alroniks:
@jako Could you have a look?

JoshuaLuckers (changes requested):
It would make sense to have the warning between strong tags like how it's done for the lexicon `setting_access_category_enabled_desc`.

alroniks:
@Mark-H @JoshuaLuckers I've updated PR.

JoshuaLuckers (approved):

opengeek (approved):

PR #14002, opened 2018-07-19
Fixed handling $depth variable during recursive calls in getTree method

### LINKED ISSUE #4973: getTree does not get all Container values
mts7 created Redmine issue ID 4973

When using getTree (as referenced in http://rtfm.modx.com/display/revolution20/modX.getTree ), even with specifying a valid depth, not all containers are displayed in the tree.  For instance, I have a page that has a parent line with 5 parents, though when I execute getTree with the first parent value, I only get children down to 3 more values.  When testing the resource IDs closer to the original page, getTree works sometimes, and not others.

### LINKED ISSUE #8430: getTree from root doesn't get nested resources
ben_omycode created Redmine issue ID 8430

Hi,

I tried to get the complete resource tree from root using :

<pre>
$modx->getTree(0, 10);
</pre>

but I only get the level 1 resources.

However, getTree successfully returns the tree of a specific resource which id different of 0.

The problem is the same with : 

<pre>
$modx->getChildIds(0, 10);
</pre>

alroniks:
### What does it do?
It fixes issues with a recursive call of the `getTree` (and `getChildIds`) method. Before a fix depth doesn't work at all, showing the full list of nested elements and also there was an issue when each next section of elements has less value of depth than previous. 
It happened because was used the wrong way to decrease the value of $depth. $depth-- decreases the value of variable itself and between iterations (in foreach) it becomes less than should be. But for passing into the recursive call we should use decreased value. So `$depth - 1` solves the issue.

See examples below:
Before a fix:
```
// $tree = $modx->getTree(0, 3); print_r($tree)
Array(
    [1] => 1
    [3] => Array(
        [2] => Array(
            [4] => Array(
                [5] => Array(
                    [6] => Array(
                        [7] => Array(
                            [8] => Array(
                                [9] => Array(
                                    [10] => Array(
                                        [11] => Array(
                                            [12] => Array(
                                                [13] => 13
                                            )
                                        )
                                    )
                                )
                            )
                        )
                    )
                )
            )
        )
    )

    [14] => Array(
        [15] => Array(
            [16] => Array(
                [17] => Array(
                    [18] => Array(
                        [19] => Array(
                            [20] => Array(
                                [21] => Array(
                                    [22] => Array(
                                        [23] => Array(
                                            [24] => Array(
                                                [25] => 25
                                            )
                                        )
                                    )
                                )
                            )
                        )
                    )
                )
            )
        )
    )
    [26] => 26
)
```

After fix:
```
// $tree = $modx->getTree(0, 3); print_r($tree)
Array(
    [1] => 1
    [3] => Array(
        [2] => Array(
            [4] => Array(
                [5] => 5
            )
        )
    )
    [14] => Array(
        [15] => Array(
            [16] => Array(
                [17] => 17
            )
        )
    )
    [26] => Array(
        [27] => Array(
            [28] => Array(
                [29] => 29
            )
        )
    )
)
```

### Why is it needed?
It solves an issue when during the `getTree` call user gets not a full list of nested items. 

### Related issue(s)/PR(s)
Fixes #4973
Fixed #8430

### DIFF
--- a/_build/test/Tests/Model/modXTest.php
+++ b/_build/test/Tests/Model/modXTest.php
@@ -29,11 +29,71 @@
  * @group Model
  * @group modX
  */
-class modXTest extends MODxTestCase {
+class modXTest extends MODxTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        /*
+         * This map following the next pattern:
+         *  1 Mainpage
+         *  2 Services
+         *  └──3 Group of services
+         *     └──4 Service
+         *        └──5 SubService
+         *  6 Catalog
+         *  └──7 Category
+         *     └──8 SubCategory
+         *        └──9 SubCategory
+         *           └──10 SubCategory
+         *              └──11 SubCategory
+         *                 └──12 SubCategory
+         *                    └──13 SubCategory
+         *                       └──14 SubCategory
+         *                          └──15 SubCategory
+         *                             └──16 SubCategory
+         *                                └──17 SubCategory
+         *                                   └──18 SubCategory
+         *                                      └──19 SubCategory
+         */
+        $this->modx->resourceMap = array(
+            0 => array(1, 2, 6),
+            2 => array(3),
+            3 => array(4),
+            4 => array(5),
+            6 => array(7),
+            7 => array(8),
+            8 => array(9),
+            9 => array(10),
+            10 => array(11),
+            11 => array(12),
+            12 => array(13),
+            13 => array(14),
+            14 => array(15),
+            15 => array(16),
+            16 => array(17),
+            17 => array(18),
+            18 => array(19)
+        );
+
+        $ctx = new stdClass();
+        $ctx->resourceMap = array(
+            21 => array(22),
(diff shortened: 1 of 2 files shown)


opengeek (dismissed):
Looks good. Nice catch!

JoshuaLuckers:
It would be nice if there was a Unit Test for this!

alroniks:
@JoshuaLuckers I agreed. Will try to add.

Mark-H (dismissed):

Jako (dismissed):

alroniks:
I am working on unit tests. But if we need to merge it to release 2.7 faster, I promise to send a new PR with tests a bit later. FYI @opengeek

opengeek:
I would prefer the tests get added before merging.

alroniks:
@opengeek @Mark-H Unit tests added.

Mark-H (approved):

Jako (approved):

JoshuaLuckers:
Amazing job!

Issue #13988, opened 2018-07-10
MODX 3 - UX/UI/Functionality Changes


kolbykruger:
### MODX 3 Changes

Below are some of the UX/UI/Functionality changes MODX3 needs. 

### UX / UI
**Files**

- Format file information into two columns (File Name and Path in the left; File Size, Last Accessed and Last Modified in the right). Reason: This shrinks the top space being used, allowing more room for the file's content to increase in size (especially on small screens).

**Chunks**

- Hide 'Tip' notification at the top. Not necessary unless it's the first chunk created. Reason: Allows more room for the chunk.

**Template Variables**

- Hide 'Tip' notification at the top. Not necessary unless it's the first chunk created. Reason: Allows more room for the TV.

**Snippets**

- Hide 'Tip' notification at the top. Not necessary unless it's the first chunk created. _Reason: Allows more room for the Snippets._

**Plugins**

- Hide 'Tip' notification at the top. Not necessary unless it's the first chunk created. _Reason: Allows more room for the Plugins._

**Categories for all Elements**

- Remove pagination. _Reason: A HUGE usability improvement to be able to scroll through all of the categories. The pagination is clunky and unnecessary._

**Toasts**

- Toasts (like the 'save successful' message) should display on the bottom right of the window. These should be a solid background color, with a maximum width of no larger than 300px. _Reason: More modern feel, better location that doesn't block functionality._

**Media Browser**

- More padding around the media browser items would allow for a more modern feel.
- Items have too many borders. Remove border wrapping the individual image, possibly make an off-white background to differentiate items

**Dashboard**

- Action buttons should stretch to full width (25%) - mentioned in a previous issue here:
- Resize and delete buttons need to appear/hide via a system setting
- Close the gap between the tray an the content area
- Scrollbars on windows are a hinderance to how the sidebar functions.
- Recently edited resources 'Resource's title' font size is off. It should be the same as the ID and then ellipsis if too long
- Recently edited resources buttons error out on click

**Edit Account**

- Move to two column layout
- Profile image should display in the right column (clipped by a circle for consistency)

**Manage -> Users**

- Include profile picture

### Functionality
**Extras**

- Most Popular -> Change to most popular for 20XX year. Overall most popular is too arbitrary.

**Site name**

- The site name needs to appear somewhere on the sidebar/tray. I would suggest adding it on the tray above the tabs (Resource/Elements/Files) as this will be open most of the time

### Navigation
**Content**
[...]

Issue #13942, opened 2018-07-02
Proposed Sidebar Menu Rework & More


kolbykruger:
With the state of the new Alpha PR for MODX3, I find the sidebar arrow to be a UX nightmare (no label, could be a back button, etc...)

My proposed change would rework how the sidebar and it's navigational items work.

I suggest that the 'Content' navigation item on click, opens the tray including 'Resources, Elements, Files, and Trash"  (in place of the arrow) while it's child elements move under the 'Manage' navigation item. I believe that 'Preview Site' should be it's own top level navigation item, and the others get recategorized under the others where they fit in (mostly under 'Manage'). 

With this structure change, clicking the other top level navigation items like 'Media,' 'Extras,' and 'Manage' would open up a similar tray to the 'Content' tray containing their children, while closing the other trays. This includes an active state added to the top level navigation items for open visibility. See included screenshot for a mockup. Should probably match the widths of the menu's to the 'Resources, Elements, Files, and Trash" tray as well.

![modx3_sidebar_content](https://user-images.githubusercontent.com/10890005/42174054-115bc6ca-7def-11e8-86b0-8b0f2e9c7636.png)
![modx3_sidebar_manage](https://user-images.githubusercontent.com/10890005/42174055-117124c0-7def-11e8-9c7c-15773f1a8b37.png)

This is probably controversial, however I can already see the support tickets pouring in from clients over the arrow.

While we're on the subject, the 'Success' toast popup should be moved/restyled to match the modern feel of MODX3. I propose it be moved to the bottom right, colored to be green (no transparency) and the width should be much smaller. The current one covers up the action buttons, is barely visible, and it's location is not ideal as you have to wait for it to disappear to click any of the buttons. Example below:

![modx3_save](https://user-images.githubusercontent.com/10890005/42174584-dab22428-7df0-11e8-847a-ecdf87953390.png)

But I do like where this redesign is going. With a few changes, it could be really _fluid_.

jonleverrier:
I like the idea of moving the status message to the bottom - it gets in the way of the "close" button positioned at the top.

muzzwood:
I love the look of the menus sliding out from the side at full height. 
I'm not a designer but to keep an obvious difference from the resource tree, the expanded menu could perhaps also be a darker colour. 
Maybe even similar to the original resource tree colouring before it was changed to grey?

PR #14003, opened 2018-07-19
[3.x] [ReUp] Changed status message appearance and position

jonleverrier:
### What does it do?
This PR changes the default appearance and position of the status notification (the one you receive after saving a resource).

modx3_message_save2

screen shot 2018-07-19 at 14 32 47

Based on feedback from Slack, this PR also adds some css utility styles so you can position the status message in different positions on the screen. These are `.modx-status-msg.has-position-center-center`, `.modx-status-msg.has-position-center-top`, and `.modx-status-msg.has-position-right-top`. 

If one day, somebody decides to create a system setting for the position, they can hook into these css utility classes.

**Screen shot of `.modx-status-msg.has-position-right-top`**
screen shot 2018-07-19 at 14 26 41
**Screen shot of `.modx-status-msg.has-position-center-top`**
screen shot 2018-07-19 at 14 26 16
**Screen shot of `.modx-status-msg.has-position-center-center`**
screen shot 2018-07-19 at 14 25 35

### Why is it needed?
In 2.x the status message was almost invisible. In MODX 3 we have lots of new styles - styling up the status message makes the product feel more consistent. 

It also moves the status message away from the action buttons at the top, which are unusable whilst the message is in display.

### Related issue(s)/PR(s)
Mentioned in #13988, #13942

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1028,14 +1028,13 @@ a.x-grid-link:hover, a.x-grid-link:focus{ text-decoration: underline; }
   padding-left: 24px !important;
 }
 
-
 /* status messages */
 .modx-status-msg {
-  background: $green;
+  background: $statusMessageBg;
   border-radius: $borderRadius;
   box-sizing: border-box;
   bottom: 20px;
-  color: $white;
+  color: $statusMessageText;
   padding: 15px 15px 15px 65px;
   position: fixed;
   right: 15px;
@@ -1045,9 +1044,9 @@ a.x-grid-link:hover, a.x-grid-link:focus{ text-decoration: underline; }
       position: relative;
   }
   &:after {
-      background: $white;
+      background: $statusMessageIconBg;
       border-radius: 50%;
-      color: $green;
+      color: $statusMessageIconText;
       content: "\f00c";
       display: inline-block;
       font-family: FontAwesome;
@@ -1067,7 +1066,7 @@ a.x-grid-link:hover, a.x-grid-link:focus{ text-decoration: underline; }
       font-size: 14px;
   }
   h3 {
-     color: $white;
+     color: $statusMessageText;
      margin: 0;
   }
 
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1030,6 +1030,9 @@ a.x-grid-link:hover, a.x-grid-link:focus{ text-decoration: underline; }
 
 /* status messages */
 .modx-status-msg {
+    @include media($mobile) {
+        max-width: 100%;
+    }
   background: $statusMessageBg;
   border-radius: $borderRadius;
   box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


JoshuaLuckers (changes requested):
It might be a good idea to define the variables in `_colors-and-vars.scss` like `$modx-status-msg-bg: $white`.

It’s makes it easier to customize if those variables are in one file.

jonleverrier:
@JoshuaLuckers The background colour is being defined from the default success colour, which i think is `$green`. You're saying it's worth creating a variable just for the status message background colour?

JoshuaLuckers:
All color variables should be defined in `_colors-and-vars.scss`. The background color was just a little example.

jonleverrier:
Understood. The status message colour (text, icon, background) is using variables from `_colors-and-vars.scss`. If you think I should create new variables specific to the status message, let me know and i'll get this changed!

JoshuaLuckers:
I do not know what the core team prefers. It's my personal preference to have variables for specific components to be in a "variables" file.
For example, I want to the text to be black and no border radius for this specific component. I now have to modify the "core" component file and increase the risk of my custom modifications to be overwritten when updating MODX. If those specific variables where in the "variables" file I could create a custom variables file "overruling" the "core" variables and thus reducing the risk of my changes being overwritten. 

TLTR: It makes it easier to theme core components.

jonleverrier:
@JoshuaLuckers added specific sass variables for the status message

jonleverrier:
@Mark-H @Alroniks - is this likely to get merged into 3.x or does it require more discussion? Is there anything more I can do?

OptimusCrime:
This looks fine by me.

alroniks:
I like this solution but have not check yet how it works, so cannot approve right now. But will try today.

Mark-H (dismissed):

_build/templates/default/sass/index.scss
```diff
@@ -1030,18 +1030,89 @@ a.x-grid-link:hover, a.x-grid-link:focus{ text-decoration: underline; }
...
+  background: $statusMessageBg;
+  border-radius: $borderRadius;
+  box-sizing: border-box;
+  bottom: 20px;
+  color: $statusMessageText;
+  padding: 15px 15px 15px 65px;
   position: fixed;
-  top: 10px;
+  right: 15px;
   width: 25%;
```
alroniks, line 1045: It seems 25% too much. On wide screens, it looks strange. I guess it can be solved by adding max-width property.

jonleverrier: happy to add a max-width to the message. I guess the original thinking behind the 25% width, is that is takes up 1 quarter of your screen, regardless of how wide.

alroniks: Yes, of course. But on too wide screens it looks not so awesome as expected.

jonleverrier:
Should be good to go now @Alroniks @Mark-H

jonleverrier:
should be good to go now @Alroniks @Mark-H !

alroniks (approved):

Mark-H (approved):

PR #14009, opened 2018-07-20
Translated descriptions for policies, policy templates and policy template groups

### LINKED ISSUE #11360: Replace hardcoded policy template descriptions with lexicon entries
It seems the descriptions for the policy templates (like AdministratorTemplate) under Access Controls > Policy Templates are hardcoded during installation in _build/data/transport.core.accesspolicytemplates.php, so changing the language has no effect on them as they are always displayed in english.

alroniks:
### What does it do?
It replaces hardcoded strings in descriptions by keys and moves texts into lexicon files + some work to get it worked after moving.

Fixed several additional issues, found during the works:
- Fixed template group for ContextTemplate policy template
- Fixed list of core groups, templates, and policies, unified code for protecting them from deleting.
- Drastically simplified data-files that used for generating modx core package, also moving names to class constants to make the code more reliable
- Removed ability in-row editing of descriptions of policies and templates to avoid uncertainties when lexicon keys are used. In the full form descriptions updated and clear, but in the table, it is difficult to explain to the user what is going on.
- Added ability to change template group of created Policy Template. Remastered editing form.

Policy Template_ AdministratorTemplate _ MODX Revolution

### Why is it needed?
It adds an ability to see descriptions of access policy templates and their groups on the user's native language (if translations exist). 

### Related issue(s)/PR(s)
Fixes #11360


alroniks:
I found out that policies also has not translations so I gonna fix them too.

Mark-H (dismissed):
Moar lexicons! Nice!

alroniks:
Too early @Mark-H :)
Need to fix processors and components for translatable names of policies.
It's mean that it shouldn't be merged for now (note for integrators). I will update the PR soon.

alroniks:
Tomorrow is 13th and it seems that I don't have time to finish this PR in time. Also, it will include additional big work for translators but we a limited in time. So I suggest moving to 3.x. I marked it for 3.x, later I will change the base of PR to 3.x and fix conflicts.

JoshuaLuckers:
@Alroniks are you still working on this?

alroniks:
@JoshuaLuckers Need to move it to 3.x and add some more translatuons but yes, I am working.

Mark-H:
Past the cut-off for new features and breaking changes in alpha1, so pushing forward. Maybe someone else could take the lead on getting this updated for 3.0.

alroniks:
@Mark-H @opengeek @JoshuaLuckers I finished, please, check.

CrazyBoy49z:
@Alroniks thanks

Ibochkarev:
@JoshuaLuckers please, change status - pr-state/review-needed

JoshuaLuckers:
@Alroniks would it be possible to fix the merge conflicts?

CrazyBoy49z:
> @Alroniks would it be possible to fix the merge conflicts?

Yes he fixed conflicts

alroniks:
@JoshuaLuckers I've fixed conflicts and did rebase as well.

alroniks:
@JoshuaLuckers @Jako @Mark-H @opengeek Does exist any chance that it will be reviewed?

JoshuaLuckers:
It appears there are some new conflicts that must be resolved. I'll review your PR tomorrow.

alroniks:
@JoshuaLuckers It is not a problem to fix these conflicts. But yes, it is annoying to fix them every time. That's why I am asking to check it and if it fits needs and requirements, merge it and go further.

JoshuaLuckers:
> @JoshuaLuckers It is not a problem to fix these conflicts. But yes, it is annoying to fix them every time. That's why I am asking to check it and if it fits needs and requirements, merge it and go further.

You are totally right! I'll check it out tomorrow!

_build/data/transport.core.accesspolicies.php
```diff
@@ -9,136 +9,54 @@
...
+    $data = [];
+    foreach ($permissions as $key => $permission) {
+        if (is_string($key) && is_bool($permission)) {
+            $data[$key] = $permission;
+        } else {
+            $data[$permission] = true;
+        }
+    }
+    return json_encode($data);
+}
```
JoshuaLuckers, line 22: Wouldn't it make sense to have this functionality available to everyone who interacts with `modAccessPolicy::class`? Maybe it would be useful to do it automatically when setting the `data`? Or maybe when saving? Just some thoughts!

_build/data/transport.core.accesspolicies.php
```diff
@@ -9,136 +9,54 @@
...
+    modAccessPolicy::POLICY_CONTEXT => ['load', 'list', 'view', 'save', 'remove', 'copy', 'view_unpublished'],
+    modAccessPolicy::POLICY_HIDDEN_NAMESPACE => ['load' => false, 'list' => false, 'view' => true],
+];
```
JoshuaLuckers, line 37: I'm curious as to why the core permissions are defined here but the policy names as constants in the object?
Furthermore I would go for a multiline formatting of the arrays to make it, in my opinion, more readable.

_build/data/transport.core.accesspolicytemplategroups.php
```diff
@@ -7,56 +7,24 @@
...
-    'description' => 'All Media Source-based policy templates.',
-]);
+foreach (modAccessPolicyTemplateGroup::getCoreGroups() as $index => $groupName) {
+    $group = $xpdo->newObject(modAccessPolicyTemplateGroup::class);
+    $group->fromArray(
+        [
+            'id' => $index + 1,
+            'name' => $groupName,
+            'description' => sprintf('policy_template_group_%s_desc', strtolower($groupName)),
+        ]
```
JoshuaLuckers, line 24: Isn't the `id` a primary key here that would require the parameter `setPrimaryKeys` set to `true`?
I'm just wondering what makes this different from the `modAccessPolicy` objects we create.

core/src/Revolution/Processors/Security/Access/Policy/GetList.php
```diff
@@ -118,15 +118,16 @@ public function beforeIteration(array $list)
...
-        $core = ['Resource', 'Object', 'Administrator', 'Load Only', 'Load, List and View'];
-        $policyArray = $object->toArray();
-        $permissions = [];
-        $cls = 'pedit';
-        if (!in_array($object->get('name'), $core)) {
-            $cls .= ' premove';
+        $policy = $object->toArray();
+
+        $classes = ['pedit'];
+        if (!in_array($object->get('name'), modAccessPolicy::getCorePolicies(), true)) {
```
JoshuaLuckers: I'm considering to add a `isCorePolicy` method to the `modAccessPolicy*` models, what do you think?

setup/includes/upgrades/common/3.0-policy-description.php
```diff
@@ -0,0 +1,19 @@
...
+$policies = $modx->getIterator(
+    modAccessPolicy::class,
+    ['name:IN' => modAccessPolicy::getCorePolicies()]
+);
+
+foreach ($policies as $policy) {
+    $policy->set('description', sprintf(
+        'policy_%s_desc',
+        str_replace([',', ' '], ['', '_'], strtolower($policy->get('name')))
+    ));
```
JoshuaLuckers, line 17: What if I changed the description of one of the core policies to something custom? Would it be possible to check if the description equals the old "default" and only then update it?

The same question applies to the two other upgrade scripts.

core/lexicon/en/policy.inc.php
```diff
@@ -29,7 +30,7 @@
 $_lang['policy_data'] = 'Policy Data';
 $_lang['policy_desc'] = 'Access policies are generic policies that restrict or enable certain actions with MODX.';
 $_lang['policy_desc_name'] = 'The name of the Access Policy';
-$_lang['policy_desc_description'] = 'Optional. A short description of the Access Policy';
+$_lang['policy_desc_description'] = 'Optional. A short description of the Access Policy. Also you might use lexicon keys here.';
```
JoshuaLuckers, line 33: Can we use all available lexicon keys or are they limited to the policy topic?

core/lexicon/en/policy.inc.php
```diff
@@ -55,10 +56,11 @@
 $_lang['policy_remove_multiple_confirm'] = 'Are you sure you want to remove these Access Policies? This is irreversible.';
 $_lang['policy_update'] = 'Update Policy';
 $_lang['policy_template'] = 'Policy Template';
-$_lang['policy_template.desc'] = 'A Policy Template defines which Permissions will show up in the Permissions grid when editing a specific Policy. You can add or remove specific Permissions from this template below. Note that removing a Permission from a Template will remove it from any Policies that use this Template.';
+$_lang['policy_template_desc'] = 'A Policy Template defines which Permissions will show up in the Permissions grid when editing a specific Policy. You can add or remove specific Permissions from this template below. Note that removing a Permission from a Template will remove it from any Policies that use this Template.';
```
JoshuaLuckers, line 59: If you're wondering (just like me 😅) what the change is: The `.` is replaced in the lexicon key!

core/src/Revolution/modAccessPolicy.php
```diff
@@ -25,6 +25,42 @@
...
+            self::POLICY_OBJECT,
+            self::POLICY_ELEMENT ,
+            self::POLICY_CONTENT_EDITOR,
+            self::POLICY_MEDIA_SOURCE_ADMIN,
+            self::POLICY_MEDIA_SOURCE_USER,
+            self::POLICY_DEVELOPER,
+            self::POLICY_CONTEXT,
+            self::POLICY_HIDDEN_NAMESPACE,
+        ];
+    }
```
JoshuaLuckers, line 62: Since the core polices don't change, have you considered using a constant for the list of core policies instead of a static method?

JoshuaLuckers (commented):
Great work! I left some general feedback and questions, nothing show stopping!

core/src/Revolution/Processors/Security/Access/Policy/GetList.php
alroniks: Yes, it makes sense.

core/lexicon/en/policy.inc.php
alroniks: It is done just to simplify logic in updater scripts. And a side effect, make lexicon keys more consistent. :)

JoshuaLuckers: It's a good change, I just had a hard time spotting it haha.

_build/data/transport.core.accesspolicies.php
alroniks: I agree, that it may be useful, but in this case, need to make this method as static to have the ability to use it in such scripts like this, that looks not good for a class that represents objects.
Move it to a separated helper also not looks good, so I would leave it as is before somebody will really need such functions in another place.

JoshuaLuckers: Fair enough!

_build/data/transport.core.accesspolicytemplategroups.php
alroniks: My solution based on the previous script (https://github.com/modxcms/revolution/blob/3.x/_build/data/transport.core.accesspolicytemplategroups.php), and this index guaranties that order will be always the same. Maybe I can simplify it more, I don't remember exactly already why I chose this solution.

core/lexicon/en/policy.inc.php
alroniks: Based on language topics defined in controllers, so I would say that yes, mostly related to access control, but need to check because the manager loads a lot of topics and all of them will be cached.

core/src/Revolution/modAccessPolicy.php
alroniks: I prefer method because it looks better in code, where it used, but yes, it is possible to move them to constant.
And another reason, method easier to override if needed (but I don't see any example when it can be required). :)

_build/data/transport.core.accesspolicies.php
alroniks: Because it is the only initial setup, after installation this list can be changed by users at any time, and having the default list does not makes sense. And in constants I stored the names, here I reuse these constant to better readability of code. The problem that we store all these permissions and policies in the database, and we can not fix it in the class.

alroniks:
@JoshuaLuckers I have move ideas about improvements, but I prefer to stop and finish with it at first, to do not make a pull request too big (it is already pretty big).

JoshuaLuckers:
If you have time to resolve the conflicts I'll test it today or tomorrow and if everything works I'll approve it!

setup/includes/upgrades/common/3.0-policy-description.php
alroniks: It is pretty safe, because now in the database we store names as-is. And in constants, we also stored such values. So, if you changed something in the database, the following query just will not find such entries because it does not equal the requested values.
```
$policies = $modx->getIterator(
    modAccessPolicy::class,
    ['name:IN' => modAccessPolicy::getCorePolicies()]
);
```

alroniks:
@JoshuaLuckers Conflicts are resolved, as well I added some improvements with methods for checking core items.

alroniks:
Besides conflicts (honestly, I already tired fix them every time), @opengeek, do you have anything to say against? 
@JoshuaLuckers Did you have chance to test it? Are all work as expected?

opengeek:
The only issue I have with this is it needs to be squashed into 1 commit and force pushed to the PR. In addition, you should never merge changes from upstream into a PR branch. If you need to resolve conflicts that show up after you submit the PR, you can rebase the changes on top of the latest from upstream.

opengeek:
Actually, I see now that you are editing the translated files directly in this PR. I'm not sure that won't completely break the integration with CrowdIn.

Jako:
Crowdin detects renamed/changed lexicon keys, if the text itself does not changes much. So the other language changes should be removed.

alroniks:
> Actually, I see now that you are editing the translated files directly in this PR. I'm not sure that won't completely break the integration with CrowdIn.

It shouldn't because only keys updated, not the strings.

opengeek:
> > Actually, I see now that you are editing the translated files directly in this PR. I'm not sure that won't completely break the integration with CrowdIn.
> 
> It shouldn't because only keys updated, not the strings.

It will. You need to remove the translated files from the commit as the changed key will be synced to CrowdIn, CrowdIn will recognize the key changed and will automatically update the translations accordingly.

alroniks:
@opengeek I squashed all into one commit.

alroniks:
@opengeek I forgot to say, but I also removed all changes from locales except English. Please, check and if all fine, let's merge it.

Ibochkarev:
@opengeek Can we check and accept this PR?

core/src/Revolution/Processors/Security/Access/Policy/GetList.php
```diff
@@ -137,16 +134,35 @@ public function prepareRow(xPDOObject $object)
...
+    }
+
+    /**
+     * @param xPDOObject|modAccessPolicy $object
+     *
+     * @return string
+     */
+    protected function prepareRowClasses(xPDOObject $object)
+    {
+        if ($object->isCorePolicy($object->get('name'))) {
```
Mark-H: Shouldn't this be a `if not isCorePolicy` check?

manager/assets/modext/widgets/security/modx.grid.access.policy.js
```diff
@@ -71,7 +71,10 @@ MODx.grid.AccessPolicy = function(config) {
             header: _('description')
             ,dataIndex: 'description'
             ,width: 375
-            ,editor: { xtype: 'textarea' }
+            ,renderer: function(value, metaData, record) {
+                return record['data']['description_trans'];
```
Mark-H: By defining a custom renderer, the standard renderer to encode html is no longer applied. Do we consider the translation lexicon safe or user input? Also repeated below in the policy template grid.

Mark-H:
This is looking preeeeettty darn good 👌

core/src/Revolution/Processors/Security/Access/Policy/GetList.php
alroniks: Yep, you are right. I fixed it. Also in templates as well.

manager/assets/modext/widgets/security/modx.grid.access.policy.js
alroniks: Added encoding to avoid potential XSS.

alroniks:
@Mark-H @opengeek @JoshuaLuckers I fixed issues and improved some other things, as well I fixed conflicts. It would be nice if you guys will find time to check this one. I know, it is a big pull request, but still atomic and brings only one new feature, which requires a lot of work though.

Mark-H (approved):
This is one of those changesets that needs extra testing during the merge (i.e. does the build work, altho I see no reason why it wouldn't) but 👍

Ibochkarev (approved):
LGTM 👍 

Thanks for the hard work! Need to fix conflicts and merged it

PR #14012, opened 2018-07-24
[ReUp] Default media source type

### LINKED ISSUE #13918: Creating a new Media Source starts with an error message
### Summary
When creating a new Media Source the dialog instantly gives an error on the "Source Type" field.

### Step to reproduce
Create a new Media Source

### Observed behavior
It starts with an error because there is no default "Source Type" selected.

### Expected behavior
It should not give an error immediately. Give the error when the Media Source is being saved or add an option to set a default source type. Preferable defaulting to "sources.modFileMediaSource".

### Environment
MODX 2.6.*, 3

JoshuaLuckers:
### What does it do?
It adds a new system setting to set the Default Media Source Type. It also selects the default media source type automatically when creating a new Media Source. 

### Why is it needed?
It fixes issue #13918

### Related issue(s)/PR(s)
Fixes #13918


setup/includes/upgrades/common/3.0.0-media-soure-type-system-setting.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+$messageTemplate = '<p class="%s">%s</p>';
+
+$defaultMediaSourceTypeSystemSetting = $modx->getObject('modSystemSetting', array(
+    'key' => 'default_media_source_type',
+));
+if ($defaultMediaSourceTypeSystemSetting === null) {
+    $defaultMediaSourceTypeSystemSetting = $modx->newObject('modSystemSetting');
+    $defaultMediaSourceTypeSystemSetting->set('key','default_media_source_type');
```
alroniks: This code can be simplified using `fromArray()` method. It will become more readable.

alroniks:
Hi @JoshuaLuckers Did you test it with context/usergroup/user settings?

alroniks:
And as it does not brake BC it should be targeted into 2.x.

JoshuaLuckers:
> Did you test it with context/usergroup/user settings?

No, I will test it in the upcoming days.

> And as it does not brake BC it should be targeted into 2.x.

Since it's a "new" system setting with new lexicons I think it makes more sense to introduce it in 3.x.

alroniks:
By SemVer 2.7 also is opened for new features and it's an easier to avoid conflicts during the merging in the future when some functionality crossing.

JoshuaLuckers:
Fair enough! I will also retarget the PR.

Mark-H:
@JoshuaLuckers I don't think you need the upgrade script you added in that last commit - when installing, it will already create new settings from the source in _build/data/.

JoshuaLuckers:
Even when upgrading?

Mark-H:
Yeah. [All settings are put into the transport package](https://github.com/modxcms/revolution/blob/2.x/_build/transport.core.php#L258-L267). Because `UPDATE_OBJECT` is set to false it doesn't update anything that's already there (identified by the key), but it will create stuff that isn't already there.

JoshuaLuckers:
Thanks @Mark-H. I will remove the upgrade script.

Mark-H (dismissed):
Excellent!

alroniks (approved):
It works, that's I would like to say. 👍

setup/lang/en/upgrades.inc.php
```diff
@@ -44,3 +44,5 @@
 $_lang['iso_country_code_converted'] = 'Successfully converted user profile country names to ISO codes.';
 $_lang['legacy_cleanup_complete'] = 'Legacy file clean up complete.';
 $_lang['legacy_cleanup_count'] = 'Removed [[+files]] file(s) and [[+folders]] folder(s).';
+$_lang['media_source_type_system_setting_success'] = 'System Setting `Default Media Source Type` added.';
```
alroniks: Looks like this lexicon strings no needed anymore too.

_build/data/transport.core.system_settings.php
```diff
@@ -447,6 +447,15 @@
   'area' => 'manager',
   'editedon' => null,
 ), '', true, true);
+$settings['default_media_source_type']= $xpdo->newObject('modSystemSetting');
```
alroniks, line 450: And also lexicons for this new system setting are missed.

alroniks (changes requested):
Need to add lexicons for new system setting.

alroniks:
Hi @JoshuaLuckers Please, have a look at my comments. Lexicons definitely should be added.

JoshuaLuckers:
When I have some time I will do so.

alroniks (approved):

alroniks:
@Mark-H Could you have a look again? The missed code was added, now it looks fine.

Mark-H (approved):

PR #14018, opened 2018-07-28
[Package Manager] Re-Download missing packages

exside:
Was surprised that the package manager is not able to re-download packages in case the zip files / extracted folders are missing for an installed package (for whatever reason, in my particular case did I want to save space for backups and did not include the packages folder...). When trying to re-install a package with missing files it would proceed to the installation (of course without metadata displaying), then when hitting "Continue" show the modal and say could not download the package etc., but it never actually tried to re-download...

### What does it do?
This PR changes that so if the zip file is not present, it would actually try to re-download it. There is an edge case where the package is not available anymore on the provider (for example I had an old version of the "console" extra that was not available anymore) and the provider would return an error code (e.g. 404 in that case), but the old code would actually "package" the response HTML or JSON code into a .zip file and then of course the installation would fail with "could not unpack package"...This PR also solves that issue by handling error codes from the provider.

Although not used very often, I did rewrite large parts of the _getByFsockopen() method as it needed some love.

### Why is it needed?
See above.


exside:
If anybody knows a better way to get a packages download link than (without making an unnecessary remote request)

```php
$location = $this->get('metadata')[array_keys(array_filter($this->get('metadata'), function($item) { if ($item['name'] === 'location') { return $item; } }))[0]];
```

please let me know, happy to change that monster...

And of course it would be nice if the frontend would inform the user about whats happening before he hits "Continue/Install" and then gets the error, but not sure how to add that...

OptimusCrime:
I do not know how the lookup is structured in `$this->get('metadata')`, but at the very least have the method on a seperate line:

```
$location = $this->get('metadata')[array_keys(array_filter($this->get('metadata'), function($item) { 
    if ($item['name'] === 'location') { 
        return $item; 
    }
}))[0]];
```

I would split this expression into a method instead and do each step seperate. Here the code is a bit too confusing to just read out. You also seem to use `array_filter` and then just use the first element in the resulting array. This could be split into a method that finds the correct `$item` or returns null, which, in turn, should execute the remote fetch instead.

exside:
@OptimusCrime added your multiline suggestion, I'd like to avoid creating new methods for something "that trivial", as far as I can see if a package is installed it will always have the location in the metadata, so there should be no need for a remote request. The problem is, that the metadata is horribly structured to say the least, its an array with numeric keys, so the location was at index 28 in my case, but I don't want to rely on that index to be always the same, therefore the `array_filter()` which of course will always only return one item. Maybe we can take the multiliner a bit apart without creating a new method:

```php
// loop trough metadata arrays until location is found
$locationarr = array_filter($this->get('metadata'), function($item) { 
    if ($item['name'] === 'location') { 
        return $item; 
    }
});
// determine the index at which the location metadata was found
$locationkey = array_keys($locationarr)[0];
// get the location metadata
$location = $this->get('metadata')[$locationkey];
```

will update like that for now.

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -362,8 +362,26 @@ public function transferPackage($sourceFile, $targetDir) {
...
             $productVersion = $this->xpdo->version['code_name'].'-'.$this->xpdo->version['full_version'];
-
-            $source= $this->get('service_url') . $sourceFile.(strpos($sourceFile,'?') !== false ? '&' : '?').'revolution_version='.$productVersion;
+            
+            /* make sure the package is downloaded, if not attempt re-download */
+            if (!file_exists($targetDir . $sourceFile)) {
+                /* not very pretty way of getting the download URL, but otherwise we have to make a 
+                 * remote request to $this->Provider->info($this->signature) to get the URL.
+                 * loop trough metadata arrays until location is found */
+                $locationarr = array_filter($this->get('metadata'), function($item) {
```
JoshuaLuckers: To maintain readability I suggest using the camelCase naming convention for variable names.

JoshuaLuckers: To maintain readability I suggest using the camelCase naming convention for variable names.

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -362,8 +362,26 @@ public function transferPackage($sourceFile, $targetDir) {
...
+                /* not very pretty way of getting the download URL, but otherwise we have to make a 
+                 * remote request to $this->Provider->info($this->signature) to get the URL.
+                 * loop trough metadata arrays until location is found */
+                $locationarr = array_filter($this->get('metadata'), function($item) { 
+                    if ($item['name'] === 'location') { 
+                        return $item; 
+                    }
+                });
+                /* determine the index at which the location metadata was found */
+                $locationkey = array_keys($locationarr)[0];
```
JoshuaLuckers: What happens if the array is empty after applying the filter?

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -432,7 +455,7 @@ public function transferPackage($sourceFile, $targetDir) {
                     $transferred= $cacheManager->writeFile($target, $content);
                 }
             } else {
-                $this->xpdo->log(xPDO::LOG_LEVEL_ERROR,'MODX could not download the file. You must enable allow_url_fopen, cURL or fsockopen to use remote transport packaging.');
+                $this->xpdo->log(xPDO::LOG_LEVEL_ERROR, 'MODX could not download the file. ' . (isset($status) ? 'Server responded with status code ' . $status . '.' : 'You must enable allow_url_fopen, cURL or fsockopen to use remote transport packaging.'));
```
JoshuaLuckers, line 501: To maintain readability it might be better to use `sprintf`.

JoshuaLuckers:
Nice work! Just to be sure, did you [sign the Contributor License Agreement](https://docs.modx.com/community/contribute/becoming-a-contributor)?

core/model/modx/transport/modtransportpackage.class.php
exside: It would die... =)... question is, would this ever happen? Answer, probably...so maybe I use `reset()` to get the first element and check later if it's not empty...you right, should have taken care of that!

core/model/modx/transport/modtransportpackage.class.php
exside: Can you help me out with that?

exside:
@JoshuaLuckers @OptimusCrime 

Separated the metadata code into a separate, more generic method `getMetadata()`, like this a package's metadata can be accessed more easily in general, any other field as well, in case its needed.

```php
    /**
     * Get metadata for a package in a more usable format
     *
     * @return array an array of metadata accessible by metadata name field
     */
    public function getMetadata() {
        $metadata = array_reduce($this->get('metadata'), function ($result, $item) {
            $key = $item['name'];
            unset($item['name']);
            
            $result[$key] = $item;
            
            return $result;
        }, array());
        
        return $metadata;
    }
```

That method will return an array in the form of:

```
Array
(
    [id] => Array
        (
            [attributes] => Array
                (
                )

            [text] => 4d556cf0b2b083396d000eeb
            [children] => Array
                (
                )

        ),
...
    [location] => Array
        (
            [attributes] => Array
                (
                )

            [text] => http://modx.com/extras/download/?id=4d556cf1b2b083396d000eec
            [children] => Array
                (
                )

        )
...
)
```

that reduces the code within `transportPackage()` to:

```php
            /* make sure the package is downloaded, if not attempt re-download */
            if (strpos($sourceFile, '//') === false && !file_exists($targetDir . $sourceFile)) {
                /* get the package metadata */
                $metadata = $this->getMetadata();

                if (!empty($metadata) && $metadata['location']) {
                    /* assign remote download URL */
                    $source = $metadata['location']['text'];
                }
            }
```

Need some help with the sprintf() thingy, not sure how you would do that optimization.

@JoshuaLuckers signed the contributor agreement years ago =)

OptimusCrime:
Not sure if we use sprintf that much in the core. I usually just see string concatenation like you have already done.

core/model/modx/transport/modtransportpackage.class.php
JoshuaLuckers: Just leave it as it is for now, looks good!

exside:
Another update, did the direct return as suggested but also found some strange edge cases with an installation I have:

1. it failed because there was no `location` attribute on the first level, only in the nested array of the `file` attribute --> therefore I had to implement recursion into `getMetadata()`, so it also parses the nested arrays and falls back to `$metadata['file']['children']['location']['text']` (yeah, my fingers hurt after typing that =D), the method got a bit uglier but also more versatile/complete. It's worth mentioning that I used `use`, but as the min. PHP requirements for MODX 2.6+ is PHP 5.3+, that should be fine right?

2. For some strange reason there are packages where a nicely formatted, one-dimensional attributes array is returned from `$this->get('metadata)` (newest MIGX and pThumb for example), had to deal with that as well...

There is an edge case I couldn't solve: Missing dependencies (like resizer for pThumb) are not properly re-downloaded, I had to delete the package via manager first, then re-install pThumb to make it work...don't know why^^, not familiar enough with the dependency feature of the package manager.

Jako:
Just one thing. Could this PR target 2.x?

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -236,6 +236,40 @@ public function getTransport($state = -1) {
...
+     * @param array $metadata optional input array to be processed, defaults to package metadata
+     * @param string $keyfield the inner array element pointing to the value for key substitution
+     * @return array an array of metadata accessible by $keyfield
+     */
+    public function getMetadata($metadata = array(), $keyfield = 'name') {
+        if (empty($metadata)) {
+            $metadata = $this->get('metadata');
+        }
+        
+        if (is_array($metadata[0])) {
```
JoshuaLuckers: Might be better to do this:
```
if (!is_array($metadata[0]) {
    return $metadata;
}
```
instead of a "huge" `if { } else { }` block.

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -661,33 +722,45 @@ protected function _getByFsockopen($url) {
...
+            $out = implode("\r\n", array(
+                "GET $path " . $_SERVER['SERVER_PROTOCOL'],
+                "Host: $host",
+                "User-Agent: Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.8.0.3) Gecko/20060426 Firefox/1.5.0.3",
+                "Accept: */*",
+                "Accept-Language: en-us,en;q=0.5",
+                "Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7",
+                "Keep-Alive: 300",
+                "Connection: keep-alive",
+                "Referer: http://$host",
```
JoshuaLuckers: I don't think it has any influence on how the code works but just to be consistent it would be nice if the referer would use the same protocol (http or https) as the "host".

exside:
@Jako just changed the base to 2.x, is that the proper way of targeting PRs? It seems it kind of messes up the PR a bit, there are now 10 changed files (probably because the PR was based on 2.6.5) instead of just one...do I need to resolve that by re-doing the PR based on the 2.x branch?

@JoshuaLuckers changed the way of returning metadata according to your suggestion and added the proper URL scheme to the referrer, guess that's the right way:

`"Referer: " . MODX_URL_SCHEME . $host,`

?

JoshuaLuckers:
Looks fine to me! After you changes the target you have some conflicts that need to be solved. Best way to fix it is to create a new branch with your changes. Revert the current branch to the latest commit from 2.x and cherry pick your changes on to the current branch and force push them.

Jako:
I am not sure, if it is the case. It should be possible to redownload every package, not only missing ones.

opengeek:
> Looks fine to me! After you changes the target you have some conflicts that need to be solved. Best way to fix it is to create a new branch with your changes. Revert the current branch to the latest commit from 2.x and cherry pick your changes on to the current branch and force push them.

@JoshuaLuckers — Don't fear the rebase! ;)

exside:
@Jako A general Re-download feature would be another story, this is simply about fixing a broken "Re-install" functionality in case there is no zip-file present in the packages directory (or the unpacked folder is empty)... 

regarding the branch mess, was that really necessary (to change the target branch to 2.x)? Switched back to the original 2.6.x for now...

JoshuaLuckers:
@exside you have to target `2.x`. As mentioned by @opengeek a rebase might be the solution to solve the "branch mess".

I believe it can be fixed using the following command: `git rebase --onto 2.x 2.6.x patch-1`

Jako:
@exside Re-download is not much more work, if I am not wrong. The processor just has to delete (or better rename) the current transport package download & the unpacked folder before downloading it again (and rename it back, when it is not downloadable).

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -661,33 +722,45 @@ protected function _getByFsockopen($url) {
...
+                "Keep-Alive: 300",
+                "Connection: keep-alive",
+                "Referer: " . MODX_URL_SCHEME . $host,
+                "\r\n"
+            ));
+            fwrite($fp, $out);
+            
+            $status = explode(' ', fgets($fp, 13))[1];
+            
+            if (strpos($status, '4') !== false || strpos($status, '5') !== false) {
```
Jako: `(strpos($status, '4') === 0 || strpos($status, '5') === 0)`

Otherwise a xx4 or xx5 would be detected too.

Jako (changes requested):

Ibochkarev:
@exside ping..

JoshuaLuckers:
If we can help you let us know @exside 👍

Jako (dismissed):

Ibochkarev:
@Jako Still need to rebase on 3.x

exside:
@Jako didn't really 

> dismissed Jako’s stale review via 6353263 6 minutes ago

but probably the rebase/merge-conflict commit broke that review, right? Did my best to rebase on 3.x, hope you guys can work with that...! Feel free to squash all the bs-commits below, the rebase messed with me...

core/model/modx/transport/modtransportpackage.class.php
```diff
@@ -216,9 +216,9 @@ public function getTransport($state = -1) {
                     }
                     if ($transferred) {
                         if ($state < 0) {
-                            /* if directory is missing but zip exists, and DB state value is incorrect, fix here */
+                            /* if directory is missing or empty but zip exists, and DB state value is incorrect, fix here */
                             $targetDir = basename($sourceFile, '.transport.zip');
-                            $state = is_dir($packageDir.$targetDir) ? $this->get('state') : xPDOTransport::STATE_PACKED;
+                            $state = (is_dir($packageDir.$targetDir) && count(glob($packageDir.$targetDir.'/*')) !== 0) ? $this->get('state') : xPDOTransport::STATE_PACKED;
```
Jako: Some whitespace is missing between the dots.

Jako (changes requested):

core/model/modx/transport/modtransportpackage.class.php
exside: @Jako you mean here `$packageDir.$targetDir`? E.g. should be `$packageDir . $targetDir` etc. ?

JoshuaLuckers:
```suggestion
                            $packageTargetDirIsNotEmpty = (count(glob($packageDir . $targetDir . '/*')) !== 0)
                            $state = (is_dir($packageDir . $targetDir) && $packageTargetDirIsNotEmpty) ? $this->get('state') : xPDOTransport::STATE_PACKED;
```

JoshuaLuckers (changes requested):

JoshuaLuckers:
@exside if you have time please accept my suggestion and merge conflict. If you need any help please let us know. 

Thanks.

Mark-H:
Past the cut-off for new features in alpha1, so pushing further.

Ibochkarev:
@exside please accept @JoshuaLuckers  suggestion and merge conflict.

Jako (approved):

Jako:
I have commited the changes of @JoshuaLuckers here.

opengeek:
@cla-bot check

JoshuaLuckers:
@Mark-H @opengeek any idea what's the best way to resolve the conflict(s) in this PR? Do we need to manually apply the changes to the refactored file?

Mark-H:
Typically you'd need to do a rebase + force push. If that's not feasible or complex, manually applying changes In a new branch is sometimes easier.

opengeek:
There are 31 commits in this PR. It needs to be reapplied manually on top of latest. It seems someone merged other work into this PR branch.

OptimusCrime:
Can this PR be closed or merged? It keeps appearing in my "Pull requests" overview, and I can not see any way of removing it myself. Thanks.

JoshuaLuckers:
@OptimusCrime you can unsubscribe from the notifications (at least on desktop) in the sidebar (almost at the bottom of the sidebar there is "Notifications").

Ibochkarev:
## Summary

Fixes the Package Manager so it can **re-download** packages when the zip file or extracted folder is missing (e.g. after excluding the packages folder from backups). Previously, re-install would fail without attempting a new download. Also handles HTTP error responses (4xx/5xx) so HTML/JSON error pages are not saved as .zip files.

**Main changes:**
- **Re-download:** When the zip is missing, use metadata `location` (or `file.children.location`) to get the download URL and fetch the package.
- **Empty directory:** Treat an empty unpacked directory as packed so the package is re-unpacked.
- **HTTP errors:** For cURL and fsockopen, detect 4xx/5xx and avoid saving error responses as zip.
- **getMetadata():** New method to convert metadata into a keyed array for easier access.
- **Referer:** Use `MODX_URL_SCHEME . $host` instead of hardcoded `http://`.
- **Update mode:** Skip re-download when `$sourceFile` already contains `//` (URL).
## Suggestions

### 1. Syntax error — missing semicolon

In the diff, line:

```php
$packageTargetDirIsNotEmpty = (count(glob($packageDir . $targetDir . '/*')) !== 0)
$state = ...
```

is missing a semicolon after the first statement. It should be:

```php
$packageTargetDirIsNotEmpty = (count(glob($packageDir . $targetDir . '/*')) !== 0);
$state = ...
```

### 2. Rebase required — 31 commits, merge conflicts

The PR has 31 commits and includes unrelated changes from branch switching. @opengeek: "It needs to be reapplied manually on top of latest." The 3.x file is `core/src/Revolution/Transport/modTransportPackage.php` (namespace `MODX\Revolution\Transport`). All logic should be ported there.

### 3. getMetadata() — typo and edge cases

- **Typo:** "recurisvely" → "recursively".
- **Index check:** `$metadata[0]` may not exist; use `!empty($metadata) && isset($metadata[0]) && !is_array($metadata[0])` or similar to avoid notices.
- **Nested structure:** The fallback `$metadata['file']['children']['location']['text']` assumes a specific structure. Document supported metadata formats or add defensive checks.

### 4. transferPackage — $source may be unset

When re-downloading, if metadata has neither `location` nor `file.children.location`, `$source` is never set. The code then falls through to the `else` branch which uses `$this->get('service_url') . $sourceFile...` — but that branch is only reached when `strpos($sourceFile, '//') !== false` or `file_exists(...)`. So when re-downloading and metadata is incomplete, we never assign `$source` and the subsequent cURL/fopen/fsockopen calls use an undefined variable.

**Recommendation:** Add an explicit check and early return or log when metadata does not contain a usable location:

```php
if (empty($metadata) || empty($metadata['location']) && empty($metadata['file']['children']['location'])) {
    $this->xpdo->log(xPDO::LOG_LEVEL_ERROR, '...');
    return false;
}
```

### 5. allow_url_fopen — no HTTP status check

The PR adds status checks for cURL and fsockopen, but `file_get_contents($source)` / `fread()` does not expose HTTP status. A 404 could still be saved as content. Consider using `get_headers()` first or wrapping in a stream context to detect redirects to error pages. Alternatively, document that allow_url_fopen users may still get error pages saved when the server returns 4xx/5xx with a body.

### 6. _getByFsockopen — status parsing

```php
$status = explode(' ', fgets($fp, 13))[1];
```

- `fgets($fp, 13)` limits to 12 chars (e.g. `HTTP/1.1 404`). Status codes can be 3 digits; ensure the buffer is large enough (e.g. 16).
- `$_SERVER['SERVER_PROTOCOL']` may be unset in CLI; add a fallback (e.g. `'HTTP/1.0'`).

### 7. Missing dependencies (author note)

The author noted: "Missing dependencies (like resizer for pThumb) are not properly re-downloaded." Consider tracking as a follow-up issue.

## Assessment

### Strengths

- Fixes the core issue: re-install works when zip or extracted folder is missing.
- Correctly handles HTTP errors so error pages are not saved as zip.
- getMetadata() improves metadata access and can be reused.
- Empty directory check treats empty unpacked dir as packed.
- Referer uses MODX_URL_SCHEME for consistency.
- Update mode is preserved: re-download is skipped when source is already a URL.

### Concerns

- Syntax error (missing semicolon).
- PR needs rebase/reapply onto 3.x.
- Possible undefined `$source` when metadata lacks location.
- allow_url_fopen path does not check HTTP status.
- fsockopen status parsing may fail for long status lines.

## Verdict

**Approve with changes** — after:

1. Fix the missing semicolon.
2. Rebase/reapply onto latest 3.x.
3. Add a guard when metadata has no usable location.
4. Fix `fgets` buffer size and `$_SERVER['SERVER_PROTOCOL']` fallback.
5. Fix the "recurisvely" typo.

The feature is valuable and the approach is sound. Jako has approved; the remaining work is mainly rebase and small fixes.

Issue #14013, opened 2018-07-25
Two PDF files return different right click options


ant-tar:
### Summary
Firstly I've noticed this issue here, https://forums.modx.com/thread/104115/file-source-permission-options-displaying-different-options#dis-post-559910 after tried myself and got the same result - there are different menu options for different PDF files.

### Step to reproduce
You should have some PDF files, open File manager tab, go through all of them and bring up the context menu by right click, compare with the previous results  This works only for some files, not for all(don't know why and how....). 

### Observed behavior
Two different context menu options, please  see screenshots. Almost all PDF files provide extended menu with ("Edit", "Quick update")

![edit file_ lunchmenu-2017-yarmouth-stellarton pdf google chrome 2018-07-25 17 25 51](https://user-images.githubusercontent.com/2815640/43207264-5bd21fda-9030-11e8-902b-802996a8c6a1.png)

and **only one** returns correct shorter menu:

![edit file_ lunchmenu-2017-yarmouth-stellarton pdf - google chrome 2018-07-25 17 27 35](https://user-images.githubusercontent.com/2815640/43207262-5ba039b6-9030-11e8-99d7-930f75a1f164.png)

### Expected behavior
As for me the same options set should be displayed for all PDF's - "Rename", "Download", "Delete" only.
Like it works for other types, f.e. for .xls.

### Environment
MODX 2.6.5 on MODXCloud

ghost:
Possibly a different media source with the same paths and urls but different permissions?

ant-tar:
Hi Susan,
well, the only default media source is there. And as far as I can see the same user added these files

SnowCreative:
Any difference in file size?

SnowCreative:
Can you use FTP and check the server permissions for these files?

smashstack-kris:
@sottwell thanks for responding.  I'll have one of our team members review the scenarios being posted above and respond shortly.

JoshuaLuckers:
I can confirm this bug. It's fixed in MODX 3.0. I think this should also be solved in 2.x.

alroniks:
@JoshuaLuckers Do you know details about fix in 3.x?

JoshuaLuckers:
@Alroniks A lot has changed in 3.x (see PR #13709).  I have a good idea where to look to fix the issue here (a better way to check if we can edit the file).

alroniks:
Closing as PR is merged, but feel free to reopen if the issue is still there.

PR #14027, opened 2018-08-07
Improves the isBinary check

JoshuaLuckers:
### What does it do?
It improves the isBinary check by checking the mime type of the file using the "finfo" class. 

### Why is it needed?
In most cases the mime type of a file is a good indicator if it's a binary file or not.

### Related issue(s)/PR(s)
Fixes issue #14013

### DIFF
--- a/core/model/modx/modfilehandler.class.php
+++ b/core/model/modx/modfilehandler.class.php
@@ -151,13 +151,8 @@ class modFileHandler {
 
         if (class_exists('\finfo')) {
             $finfo = new \finfo(FILEINFO_MIME);
-            $fileMimeType = $finfo->file($file);
 
-            if (substr($fileMimeType, 0, 4) !== 'text') {
-                return true;
-            }
-
-            return false;
+            return substr($finfo->file($file), 0, 4) !== 'text';
         }
 
         $fh = @fopen($file, 'r');
--- a/core/model/modx/modfilehandler.class.php
+++ b/core/model/modx/modfilehandler.class.php
@@ -145,15 +145,26 @@ class modFileHandler {
      * @return boolean True if a binary file.
      */
     public function isBinary($file) {
-        if (file_exists($file)) {
-            if (!is_file($file)) return false;
-            $fh = @fopen($file, 'r');
-            $blk = @fread($fh, 512);
-            @fclose($fh);
-            @clearstatcache();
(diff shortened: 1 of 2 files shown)


core/model/modx/modfilehandler.class.php
```diff
@@ -145,15 +145,26 @@ public function getDirectoryFromFile($fileName) {
...
-            @fclose($fh);
-            @clearstatcache();
-            return (substr_count($blk, "^ -~" /*. "^\r\n"*/) / 512 > 0.3) || (substr_count($blk, "\x00") > 0) ? false : true;
+        if (!file_exists($file) || !is_file($file)) {
+            return false;
+        }
+
+        if (class_exists('\finfo')) {
+            $finfo = new \finfo(FILEINFO_MIME);
+            $fileMimeType = $finfo->file($file);
```
OptimusCrime: Alternative:

```
return substr($finfo->file($file), 0, 4) !== 'text';
```

This also eliminates the rest of the lines inside the if.

bezumkin:
What is the point of improvements for 2.x version that [are already made in 3.x](https://github.com/modxcms/revolution/blob/0f6f73049357c14e4d50a30dedc3e19c4994736e/core/model/modx/sources/modmediasource.class.php#L2245-L2256)?

JoshuaLuckers:
@bezumkin it's a small improvement for a bug that should be fixed in the 2.x release. 
If the improvements in 3.x are non breaking changes they should also be in 2.x (in my opinion).

Mark-H (approved):
Blimey, I don't even wanna know how that substr_count stuff is supposed to work, so if we can bypass that with finfo, awesome!

OptimusCrime:
I feel like that last line deserves a line (or two, or more) with comments that explains just what they do. It would be very nice for people reading the code afterwards.

Edit: Oh, that line was already there. Sorry.

JoshuaLuckers:
It's also a function that can be deprecated. As far as I can see the function isn't being used anymore in the 3.x branch.

ghost:
Then how are binary resource types streamed?

OptimusCrime:
Perhaps handled by flysystem?

JoshuaLuckers:
Indeed.

Jako (approved):

Jako:
Checked. Works fine.

Issue #13849, opened 2018-04-03
Font Awesome Icons for MODX3 [Feature]


kolbykruger:
It would be awesome if MODX moved over to the new Font Awesome icon library, instead of using the old 4.7 library. Is there any plan for this?

alroniks:
3.x version upgraded (with some minor issues) to FA 5.6.3 - free version. Further upgrades should be easier. So as issue solved I am closing it.

PR #14030, opened 2018-08-07
FontAwesome 5.2.0 and updated versions of frameworks

alroniks:
### What does it do?
This PR provides updated tools for building default MODX manager theme to avoid various warnings messages during the build and also provides updated icons font [FontAwesome](https://fontawesome.com/) to the latest free version - 5.2.0 (**1295** icons against only **675** in 4.7)

### Why is it needed?
This PR solves issues with outdated packages that were used for building the default theme, also allows using new classes `.fa .fab .fas .far` for icons in the manager.

### Related issue(s)/PR(s)
#13849

P.S. I'm not sure in targetting this changes to 2.x, but it can be discussed. And also testing strongly needed.


JoshuaLuckers:
I would target 3.x. Especially since new classes for icons are being used.

JoshuaLuckers:
~~Maybe also include the v4 shims. The v4 shims makes it possible to use old version 4 icon names, aliases, and syntax while mapping version 5 icons to current references.~~

alroniks:
@JoshuaLuckers I already have done it. v4-shims just include `variables`and `shims` files.

~~But there is still an issue with brands. Brands use another font family and all seems valid but the browser does not download font file with icons of brands.~~

alroniks:
Closed favor to #14031.

PR #14032, opened 2018-08-09
[ReUp] Enhancement for dashboard jumping around on page load

### LINKED ISSUE #14019: Content Jumping when loading Manager page
### Summary
It's a cosmetic issue, but the Widget panel is loaded before the resource tree. The Widget panel loads on the left, then jumps suddenly to the right when the tree loads.  This is especially pronounced if any of the widgets take time (especially if the cache has been cleared or there are many resources).

Maybe it would be possible to load the tree ahead of the Widget panel, or load a placeholder for the tree to reserve space for it ahead of the widgets.

### Step to reproduce
Clear the cache when not on the dashboard, then click on the MODX icon.

### Observed behavior
Content jumps to the right

### Expected behavior
Content is stable.

### Environment
Any recent version of MODX

jonleverrier:
### What does it do?
This PR adds a default CSS `left` value to `#modx-content`. When the javascript value is loaded, this default CSS value is overwritten and ignored. It is also ignored for mobile/tablets.

### Why is it needed?
It stops the dashboard from jumping around on page load, as seen below. It makes the loading experience seem more solid. It makes the perceived loading time feel faster.

43406496-a25bf8be-941c-11e8-903c-81f8908aebc6

After the fix has been applied:

modx_dashboard_glitch

Having said that, if a user sets the tree wider than the default width and visits the dashboard again, a small amount of jumping will re-appear because the `left` value is set to `310px`.

Up to the integrators to decide if this is a fix, or if this PR creates another problem, or if this issue simply can't be fixed at this moment in time.
 
### Related issue(s)/PR(s)
Fixes #14019

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -146,6 +146,12 @@ hr {
   position: absolute;
 }
 
+/* give #modx-content an initial left value to prevent the panel from
+jumping around */
+#modx-content {
+    left: 310px;
+}
+
 .modx-form p {
   padding-bottom: 10px;
 }


Mark-H (approved):
I'm okay with the caveat of the jump not going away entirely if people changed the width of their sidebar. It's a solid and simple step in the right direction.

Jako (approved):

PR #13900, opened 2018-05-13
MODX 3 refactoring

bezumkin:
### What does it do?
- PSR-4 autoloading
- Core model updated to xPDO 3.0
- Refactored processors
- Removed modXMLRPCResource and modJSONRPCResource (and also their Response classes)
- Replaced MagpieRSS to SimplePie
- Core files formatted to PSR-2
- Updated build, setup, model generator, tests, etc. Everything needed to make it work.

### Why is it needed?
To make MODX modern again.

### Related issue(s)/PR(s)
#11272 

### How to test
1. `composer create-project modx/revolution www 3.x-dev` where `www` is the directory to install.
When it will ask `Do you want to remove the existing VCS (.git, .svn..) history? [Y,n]?` answer **n**.
2. Fetch this PR: `cd www && git fetch origin pull/13900/head:3x-refactoring`.
3. Checkout to the new branch: `git checkout 3x-refactoring`
4. Update composer: `composer update`
5. Build core files: `php ./_build/transport.core.php`
6. Open your website in browser and run Setup.
7. Login to manager and disable system setting `compress_js`. Clear browser cache and after this you will see no more errors about wrong processors paths.

### Compatibility
Sorry, but after moving MODX to namespace, we will need to update all complicated extras to this modern codebase. 

Of course, there are some compatibility workarounds like `MODX::loadClass()` method and automatic aliases to old class names, but there can't be 100% support of old codebase.

For now it was tested with:
- [x] Ace
- [x] pdoTools
- [x] Theme.Bootstrap
- [x] MinifyX
- [ ] ms2Gallery - installs, but can't work with `modMediaSource` to generate previews
- [ ] miniShop2 - installs, but can't extend `modResource`
- [ ] mSearch2 - installs, but cant't index resources.
- [ ] Office - installs and can authentificate users, but can't register them, because of extending `modUserCreateProcessor`

I think that `modResource` extending packages will require more work (Articles, Collections, Tickets, miniShop2, etc). Because now they should extend `MODX\modResource`, but not `modResource` and it will require to change their model to version 3.0
 
### Whats next?
I didn't finished yet, but now I want to show and discuss this changes. Then I want to move further:
- Remove all `getService()` calls from the core
- Rewrite all `getObject('modResource')` to `getObject('MODX\modResource')` because it will not require to get correct class name with `MODX::loadClass()`.
- ? 
- Update all necessary extras.

PR #14033, opened 2018-08-09
Remove unused and outdated xmlrpc and jsonrpc classes

alroniks:
### What does it do?
It removes outdated classes that rarely on never used last time. In this case - `xmlrps` and `jsonrpc` staff.

### Why is it needed?
Because an original refactoring pull request is too big to review it will help to bring fixes a bit faster splitting refactoring into small pieces.

### Related issue(s)/PR(s)
#13900


OptimusCrime:
This is awesome! Nice work :) I like the diff size:

+0 lines
-7,654 lines

:D

Mark-H (approved):
Excellent!

Mark-H (changes requested):
There's actually even more we can remove here that should probably be removed at the same time. `modXMLRPCResource` and `modJSONRPCResource` resource classes + their mysql/sqlsrv variants + the references to it in the schemas. That's where those responses classes were used.

alroniks:
@Mark-H I added 9 new lines (I had to :/) and removed yet 133 :)

Mark-H (approved):
Awesome!

ilyautkin:
PR is tested. Everything works as usual. Nothing was broken.

alroniks:
Hi @opengeek Could we delete this files? Yes, I am asking for approve :)

opengeek (approved):

PR #14034, opened 2018-08-09
Fix/remove custom xmlrss

alroniks:
### What does it do?
It removes custom classes for parsing RSS feeds in dashboard widgets by replacing them SimplePie library (composer depps).

### Why is it needed?
Because an original refactoring pull request is too big to review it will help to bring fixes a bit faster splitting refactoring into small pieces.

### Related issue(s)/PR(s)
#13900


composer.json
```diff
@@ -51,10 +51,14 @@
...
         "erusev/parsedown": "^1.7",
-        "pelago/emogrifier": "^2.0"
+        "pelago/emogrifier": "^2.0",
+        "simplepie/simplepie": "^1.5"
     },
     "autoload": {
-        "classmap": ["core/xpdo/"]
+      "classmap": [
+        "core/xpdo/",
+        "core/model/modx"
```
Mark-H: is this necessary here?

Mark-H:
schermafbeelding 2018-08-09 om 17 10 22

❤️

composer.json
alroniks: I've experimented with external core and found out that MODX classes not added to autoload configuration of the composer. So maybe this PR not the best place for this change, but it should be in MODX 3 to get MODX core classes available via autoload.php.

core/model/modx/processors/system/dashboard/widget/feed.class.php
```diff
@@ -37,33 +32,50 @@ public function process()
...
+            $output[] = $this->getFileChunk('dashboard/rssitem.tpl', [
+                'title' => $item->get_title(),
+                'description' => $item->get_description(),
+                'link' => $item->get_permalink(),
+                'pubdate' => $item->get_local_date()
+            ]);
         }
-        return $this->success('', array('html' => implode("\n",$o)));
+
+        return $this->success('', array('html' => implode(PHP_EOL, $output)));
```
JoshuaLuckers: Maybe use the short array syntax `[]`  here as well?

Mark-H (dismissed):
Looks good, thanks @Alroniks. Suggestion from @JoshuaLuckers is valid and would be welcome but I'd be happy to see this merged without as well. After 2.7 when 2.x has been merged into 3.x I'd love to throw `phpcs` and `phpcbf` against the entire codebase to get all different code styles cleaned up. :D

alroniks:
@Mark-H I've removed also mentioning of xmlrss from the build configuration and replaced array by short syntax.

Mark-H (approved):

alroniks:
Hi @opengeek Could we delete this files? Yes, I am asking for approve :)

opengeek (approved):
Please don't just change syntax (array syntax in this case) or formatting when making specific other changes like this in the future. We can apply those is atomic commits to avoid unnecessary conflicts.

ilyautkin:
PR is tested. Feed widgets works after `composer update`. No bugs found.

PR #14036, opened 2018-08-09
Array based config file

OptimusCrime:
### What does it do?
The spiritual successor of PR #13834.

This PR provides a new, cleaner, and easier-to-work-with config file for MODX Revolution. 

I have always thought that the current config file is far too complex. It contains both constants, and a lot of logic. It is difficult to find what you are looking for, and it is easy to make mistakes. Some variables derived from the setup process is even defined twice. With this PR, we instead return the complete config array, and define the constants in the modx class, which I think makes much more sense.

Example of a working config file, produced from the setup process:

```php
<?php
/**
 *  MODX Configuration file
 */

return [
    'last_install_time' => 1533854383,
    'id' => 'modx5b6cc2af7f1be3.27746164',
    'https_port' => '443',
    'uuid' => '157b9732-ea87-47f8-b447-8c4ee24893fb',
    'url_scheme' => 'automatic',
    'http_host' => '192.168.99.100:8000',
    'disable_cache' => false,
    'db' => [
        'type' => 'mysql',
        'server' => 'db',
        'user' => 'revolution',
        'password' => 'revolution',
        'connection_charset' => 'utf8',
        'database' => 'revolution',
        'table_prefix' => 'modx99_',
        'dns' => 'mysql:host=db;dbname=revolution;charset=utf8',
        'config_options' => array (
),
        'driver_options' => array (
),
    ],
    'paths' => [
        'base' => '/var/www/html/',
        'core' => '/var/www/html/core/',
        'processors' => '/var/www/html/core/model/modx/processors/',
        'connectors' => '/var/www/html/connectors/',
        'manager' => '/var/www/html/manager/',
        'assets' => '/var/www/html/assets/',
    ],
    'urls' => [
        'base' => '/',
        'connectors' => '/connectors/',
        'manager' => '/manager/',
        'assets' => '/assets/',
    ],
    'log_levels' => [
        'fatal' => 0,
        'error' => 1,
        'warn' => 2,
        'info' => 3,
        'debug' => 4,
    ]
];
```

No additional changes has been made, other than changing the format of the config template file, as well as reimplementing the logic that was earlier found there.

Note that this needs testing, but I have successfully used the code in this PR on a test site, and it seem to function as it should.

### DIFF
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -2,90 +2,45 @@
 /**
  *  MODX Configuration file
  */
-$database_type = '{database_type}';
-$database_server = '{database_server}';
-$database_user = '{database_user}';
-$database_password = '{database_password}';
-$database_connection_charset = '{database_connection_charset}';
-$dbase = '{dbase}';
-$table_prefix = '{table_prefix}';
-$database_dsn = '{database_dsn}';
-$config_options = {config_options};
-$driver_options = {driver_options};
 
-$lastInstallTime = {last_install_time};
-
-$site_id = '{site_id}';
-$site_sessionname = '{site_sessionname}';
-$https_port = '{https_port}';
-$uuid = '{uuid}';
-
-if (!defined('MODX_CORE_PATH')) {
-    $modx_core_path= '{core_path}';
-    define('MODX_CORE_PATH', $modx_core_path);
-}
-if (!defined('MODX_PROCESSORS_PATH')) {
-    $modx_processors_path= '{processors_path}';
-    define('MODX_PROCESSORS_PATH', $modx_processors_path);
-}
-if (!defined('MODX_CONNECTORS_PATH')) {
-    $modx_connectors_path= '{connectors_path}';
-    $modx_connectors_url= '{connectors_url}';
-    define('MODX_CONNECTORS_PATH', $modx_connectors_path);
-    define('MODX_CONNECTORS_URL', $modx_connectors_url);
-}
-if (!defined('MODX_MANAGER_PATH')) {
-    $modx_manager_path= '{mgr_path}';
-    $modx_manager_url= '{mgr_url}';
-    define('MODX_MANAGER_PATH', $modx_manager_path);
-    define('MODX_MANAGER_URL', $modx_manager_url);
-}
-if (!defined('MODX_BASE_PATH')) {
-    $modx_base_path= '{web_path}';
-    $modx_base_url= '{web_url}';
-    define('MODX_BASE_PATH', $modx_base_path);
-    define('MODX_BASE_URL', $modx_base_url);
-}
-if(defined('PHP_SAPI') && (PHP_SAPI == "cli" || PHP_SAPI == "embed")) {
-    $isSecureRequest = false;
-} else {
-    $isSecureRequest = ((isset($_SERVER['HTTPS']) && !empty($_SERVER['HTTPS']) && strtolower($_SERVER['HTTPS']) !== 'off') || $_SERVER['SERVER_PORT'] == $https_port);
-}
-if (!defined('MODX_URL_SCHEME')) {
-    $url_scheme=  $isSecureRequest ? 'https://' : 'http://';
-    define('MODX_URL_SCHEME', $url_scheme);
-}
-if (!defined('MODX_HTTP_HOST')) {
-    if(defined('PHP_SAPI') && (PHP_SAPI == "cli" || PHP_SAPI == "embed")) {
-        $http_host='{http_host}';
-        define('MODX_HTTP_HOST', $http_host);
-    } else {
-        $http_host= array_key_exists('HTTP_HOST', $_SERVER) ? htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES) : '{http_host}';
-        if ($_SERVER['SERVER_PORT'] != 80) {
-            $http_host= str_replace(':' . $_SERVER['SERVER_PORT'], '', $http_host); // remove port from HTTP_HOST
-        }
-        $http_host .= ($_SERVER['SERVER_PORT'] == 80 || $isSecureRequest) ? '' : ':' . $_SERVER['SERVER_PORT'];
-        define('MODX_HTTP_HOST', $http_host);
-    }
-}
-if (!defined('MODX_SITE_URL')) {
-    $site_url= $url_scheme . $http_host . MODX_BASE_URL;
-    define('MODX_SITE_URL', $site_url);
-}
-if (!defined('MODX_ASSETS_PATH')) {
-    $modx_assets_path= '{assets_path}';
-    $modx_assets_url= '{assets_url}';
-    define('MODX_ASSETS_PATH', $modx_assets_path);
-    define('MODX_ASSETS_URL', $modx_assets_url);
-}
-if (!defined('MODX_LOG_LEVEL_FATAL')) {
-    define('MODX_LOG_LEVEL_FATAL', 0);
-    define('MODX_LOG_LEVEL_ERROR', 1);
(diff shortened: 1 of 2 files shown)


alroniks:
Now adding the `.env` seems very easy. :)

OptimusCrime:
@Alroniks There are many additional things that can be done with this PR, but I have kept it as minimal as possible to begin with.

opengeek (changes requested):
This destroys the ability for MODX to determine the `http_host` automatically, among some other details that should not be set directly in the config file without the logic to determine the values based on values from the server and request.

OptimusCrime:
@opengeek Why is that? I have reimplemented that logic [here](https://github.com/modxcms/revolution/pull/14036/files#diff-7a5d5e3dc0a43d8bb090f75fbfbc31b0R706). The config value is the same derived from the setup process and placed in the config file, [here](https://github.com/modxcms/revolution/pull/14036/files#diff-7a5d5e3dc0a43d8bb090f75fbfbc31b0R706) on line 60 and 63.

opengeek:
> @opengeek Why is that? I have reimplemented that logic here. The config value is the same derived from the setup process and placed in the config file, here on line 60 and 63.

@OptimusCrime this is not the same. You are hardcoding the http_host in the config file and then setting the MODX_HTTP_HOST constant based on that value. There is no calculation of `$config['http_host']` so it will never be calculated at run-time. Calculating it only at setup is not going to be sufficient.

OptimusCrime:
I could update the code to set the value of `$config['http_host']` be determined by the checks done in  [modX::resolveArrayConfigRequestSettings#706](https://github.com/modxcms/revolution/pull/14036/files#diff-7a5d5e3dc0a43d8bb090f75fbfbc31b0R706). Would that improve the situation?

The whole point here is to separate the raw config values from logic.

opengeek:
> I could update the code to set the value of $config['http_host'] be determined by the checks done in modX::resolveArrayConfigRequestSettings#706. Would that improve the situation?

> The whole point here is to separate the raw config values from logic.

This is not the only configuration setting that currently depends on logic being executed instead of values being hardcoded into the config file. There are several that critical MODX features depend on. So my whole point is that removing this logic at this point destroys some functionality that is essential to the various ways people use the product.

OptimusCrime:
Hm, that is a shame. I was really hoping we could improve upon the current config file, which I think is a huge mess. I guess this is impossible then.

Feel free to close this PR, otherwise, I'll do it myself.

Mark-H:
@opengeek From what I can tell, the logic that was previously in the config file, is still being executed with this PR, but [it's in the modX class instead of the config file](https://github.com/modxcms/revolution/pull/14036/files#diff-7a5d5e3dc0a43d8bb090f75fbfbc31b0R693). What's hardcoded in the config file are the defaults, just like it was in the old config file. 

The biggest breaking change I see is that the global variables (e.g. `$database_server`) will no longer be filled, but that's not necessarily a bad thing so long as we document that change.

I'm inclined to say this is a massive step forward in how we manage our configuration (like @OptimusCrime says - splitting up the configuration and logic), so if you could be a bit more specific about the dynamic features you see missing in this PR, then we can figure out a path forward.

opengeek:
@Mark-H it is not the same. It is only executed in this PR for the constant MODX_HTTP_HOST, and that is based on whatever value is hardcoded in the config file. This IS a major breaking change. Anyone who relies on routing based on the supplied HTTP_HOST header value will be dead in the water. We should be figuring out how to change the configuration into DI configuration. Then any of the variables here could be dynamically configured with a callable.

opengeek:
FWIW, the current logic is, dynamically calculate all of the variables associated with defining the URL for MODX (url_scheme, http_host, base_url, site_url) and to override that, hardcode it in system or context settings.

opengeek:
The only time a "default" value is ever considered in the current config file is when no HTTP_HOST header is provided, i.e. through CLI usage.

Mark-H:
`MODX_URL_SCHEME`, `MODX_HTTP_HOST` and `MODX_SITE_URL` **are** being dynamically generated in `resolveArrayConfigRequestSettings` if the `$_SERVER['HTTP_HOST']` is provided. It's quite literally the same code that used to be in the config file itself, so I don't see a reason for the behaviour to be different from before there. Line 694-722 in modx.class.php is line 49-74 of the old config file. `MODX_URL_SCHEME` on line 700, `MODX_HTTP_HOST` on line 716, `MODX_SITE_URL` on 721.  

The `MODX_BASE_URL` does not seem to be dynamically calculated in the old situation (at least not in the config file itself, old line 47), and also isn't in this PR, so while that should be consistent, that could be worth addressing.

What I'm unsure off based on the PR alone, and maybe that's what you're worried about too, is how this affects `$modx->getOption('http_host')`, versus accessing the constant `MODX_HTTP_HOST`. I'm not sure how that is currently being populated and if changes to that might be required.

opengeek:
Ugh, no it's not. The setting itself must be dynamic, not the constants, which always fallback to the originally calculated value in the current approach.

opengeek:
Ignore the MODX_ constants, those do not matter.

opengeek:
In a current multi-context site, it detects the HTTP_HOST from the header, sets the value to be used in system settings and then sets the MODX_ constants. You are correct that base_url is not dynamically calculated, but site_url is based on the other dynamic values + base_url.

OptimusCrime:
@Mark-H I guess he means that `$this->config['http_host']` from the modX object needs to be adjusted with the output from the checks as well.

I'll look into this to see if it is possible to retrofit the values back into the original array.

Mark-H:
I had a chat with Jason yesterday and he was under the impression the array in the config file was directly being put into $modx->config (like $config_options would be in the current situation), hardcoding those references. But that's not the case, and I've confirmed this PR works as expected with those values being dynamically populated ($modx->config as well, from the constants, in getConfig).

So the remaining question is if this is the approach we want to go with. Jason suggested wanting to have this use a container callable. And also that the values in the config file should get written into $modx->config to replace the current $config_options. 

Personally I think this PR, in its current state, would be a good step up towards a more sensible way of defining, and make it easier to do those other things, but it might need some more discussions about the architecture we want to go with in the future.

OptimusCrime:
I thought Jason was correct. How does defining `define('MODX_SITE_URL', $site_url);` result in it ending up in the config array? I have to take a closer look at this.

Regardless, if we decide to go with a container callable, we should still be able to use this format at the basis for the configuration file. This is the way Slim does it, for example.

So, I would agree, wither way, this is an improvement over the current config file.

opengeek:
> I thought Jason was correct. How does defining define('MODX_SITE_URL', $site_url); result in it ending up in the config array? I have to take a closer look at this.

@OptimusCrime because your actual `$config` variable does not become part of the actual `$modx->config` (i.e. settings loaded from system, config, & user), the `loadConfig()` method uses the define to set the value in system settings.

Overall, this is indeed an improvement, I just would like values set here to actually end up in the `$modx->config`, and the ones that have no business being in it (such as the global variables) should simply be taken out and replaced by a more appropriate approach. For now, that could just be embedded in a callable. With this preliminary work, I can take a stab at revising it to what I'm thinking of shortly. Just let me try and wrap up this `pgsql` work I'm in the middle of.

OptimusCrime:
No problem. I'll wait. Let me know :)

Mark-H:
I'm going to close this PR. It's a great exploration of a different way of handling the configuration but is going to need some more work and collaboration to make it into the core. Hopefully, the seeds have been planted and we'll revisit this soon :)

OptimusCrime:
Hmm, okey. Is there anything I can do? I mean, the PR and everything was working 100% with installs and upgrades.

Mark-H:
Work with Jason on exploring the idea further :)

Issue #14024, opened 2018-08-02
Invalid relative media source path shows root path contents


opengeek:
### Summary
If you specify an invalid relative path for a Filesystem media source, it displays the contents of the server root folder.

### Step to reproduce
Specify basePath as a non-existent relative path with the basePathRelative option set to Yes.

### Observed behavior
The media source renders the contents of the root folder of the server.

### Expected behavior
It should render nothing since the path is invalid and display some kind of error.

### Environment
MODX 2.x (latest), nginx 1.15.0, MariaDB 10.3.7, Chrome, MacOS 10.13.6

Jako:
Ah, I saw this lately on a foreign installation, but I did not realized, that it could be a bug.

JoshuaLuckers:
I tried to fix this.. but I really can't figure out why it happens..

Mark-H:
I think there's a `realpath()` somewhere that is evaluating to false or null because of the invalid path, resulting in it looking at the root instead.

SnowCreative:
Yes, this has always bothered me, and my clients get confused (and can potentially do bad stuff) when the contents of the root folder show up.

mrhaw:
Question, if we fix this now, how do I go about showing root path for my personal media sources? :)

JoshuaLuckers:
@mrhaw that's possible by setting `basePath` to `/` and `basePathRelative` to `no`.

mrhaw:
Sweet!! Thanks @JoshuaLuckers

Jako:
#14038 solves that somehow better. But thanks for the PR.

Jako:
Sorry, I have accidentally closed the issue and not the PR.

Jako:
Closing this, since #14038 is merged.

PR #14038, opened 2018-08-10
Fix for "Invalid relative media source path shows root path contents"

JoshuaLuckers:
### What does it do?
An invalid relative media source path would result in '/' being set as path.
With this PR there is a check if realpath returns an actual path instead of false.

### Why is it needed?
Quick solution for issue #14024 

### Related issue(s)/PR(s)
#14024 

This also needs to be fixed in 3.x

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -47,7 +47,10 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
         $bases['path'] = $properties['basePath']['value'];
         $bases['pathIsRelative'] = false;
         if (!empty($properties['basePathRelative']['value'])) {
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
This is a quick solution. There are various improvements that can be made.
I also found some other bugs. I will fill new issues for those at a later moment.

Mark-H:
What is `$bases['pathAbsolute']` set to if `$realpath === false`?

Mark-H:
The failure on travis was due to travis, not the change. Rebuild marked is as green now.

JoshuaLuckers:
> What is `$bases['pathAbsolute']` set to if `$realpath === false`?

I believe it's set to nothing (iirc elsewhere it checks if `$bases['pathAbsolute']` is empty. If yes it returns an empty array.

Jako:
Checked. Works fine.

Jako (approved):

OptimusCrime:
Does this PR also close #14024? @Jako

Jako:
Yes it did.

PR #14039, opened 2018-08-11
Do not give TVs with output type as HTML an ID if no ID is sat

### LINKED ISSUE #10012: TVs with Output Type set to HTML Tag and a blank ID field still get an ID
eerne created Redmine issue ID 10012

TVs with Output Type set to HTML Tag and a blank ID field still get an ID.

http://bugs.modx.com/issues/1589

OptimusCrime:
### What does it do?
If a template variable has their output type sat to HTML, and not ID is specified for that output type, it should not produce an ID in the rendered markup. The old behavior was to use the template variable name.

### Why is it needed?
It is inconherent and wrong behavior. 

### Related issue(s)/PR(s)
Fixes #10012

### DIFF
--- a/core/model/modx/processors/element/tv/renders/web/output/htmltag.class.php
+++ b/core/model/modx/processors/element/tv/renders/web/output/htmltag.class.php
@@ -16,7 +16,7 @@ class modTemplateVarOutputRenderHtmlTag extends modTemplateVarOutputRender {
             $tagvalue = is_array($value[$i]) ? implode(' ', $value[$i]) : $value[$i];
             if (!$tagvalue) continue;
 
-            $domId = $tagid ? $tagid : $id;
+            $domId = $tagid ? $tagid : '';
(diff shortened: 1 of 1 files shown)


Mark-H (approved):

alroniks (approved):
Yes, it works and should be merged.
In case the same TV is used twice on the page it will generate an invalid HTML output. PR fixes that.

alroniks:
Hm, I missed that it's targeted to 3.x instead of 2.x. 
@OptimusCrime Could you try to change base branch for this PR, please? It does not break BC, so should be targeted to 2.x.

OptimusCrime:
@Alroniks yes it does. If people rely on using the TV name as the HTML tag ID, then this will break that behavior. I belive that this would be very uncommon, but I still think 3.x is more correct than 2.x for this PR.

However, if you still disagree, then I can change the target.

alroniks:
Yes, make sense. So merging.

Issue #7293, opened 2012-03-03
TV Input Option Values should be hidden or inactive when not relevant


modxbot:
everettg_99 created Redmine issue ID 7293

Some TVs do not make any use of the _Input Option Values_; for those TVs, that text area should be hidden or visually deactivated so the user is clear that the field is not relevant.

E.g. what does the "Input Option Values" do for Text or Number fields?

PR #14040, opened 2018-08-11
Hide Input Option Values for TV types where this is useless

OptimusCrime:
### What does it do?
The box called Input Option Values are displayed for all types of TVs. This makes no sense for a bunch of them, and hiding the box would most likely be a good idea. This PR attempts to do this.

The approach is not very good, but I found no better ways of doing while attempting to solve this. The approach is to hide the box in the renderes, and display it when a new TV type is selected. This way, we do not hide the box for third party or custom TV types.

### Why is it needed?
Reasons stated above.

### Related issue(s)/PR(s)
#7293


Jako (approved):
Checked, works fine.

We should remove an existing value in those inputs during saving the template variable. But since this would be a BC, this should be prepared for MODX3.

Issue #13883, opened 2018-04-29
Undo removal of system setting "manager_language"


labr1005:
### Summary
#13786 (amongst other things) removed the possibility to manually set the manager (and login screen) language globally, for usergroups or specific users.

IMO the part that changed the manager_language handling should be reverted respectively it should be possible to define which mode to use - auto detect or manual.

Some more opinions on this from Slack:

> I like it better to specify. The Norwegian language is terrible, and we usually just force them to use english.

> also if you're doing SSO into the manager you may not see a login screen, if anyone does that sort of thing

> yeah, it definitely needs to be manageable, not just automated or set from login, people aren't going to want to switch it manually on every login if the auto selection is not what they wanted

> Yes please have it configurable on top of auto-detect.

### Observed behavior
It's not possible to manually set the manager (and login screen) language globally, for usergroups or specific users.

### Expected behavior
It should be possible to manually set the manager (and login screen) language globally, for usergroups or specific users.

### Environment
MODX 3

Mark-H:
> I like it better to specify. The Norwegian language is terrible, and we usually just force them to use english.

One translation being terrible is not a very good reason to take away a very useful autodetection feature. If anything, that is an argument for people to spend time on their translation to ensure it is usable and helps their local geographic manage their sites in their native language. 

> also if you're doing SSO into the manager you may not see a login screen, if anyone does that sort of thing

SSO integrations can set the language in the session they create. This could be hardcoded, fed in from a third party system that knows what language people want to use or have the same autodetect as the standard login. 

> yeah, it definitely needs to be manageable, not just automated or set from login, people aren't going to want to switch it manually on every login if the auto selection is not what they wanted

I believe the chosen language is stored in a cookie, not just the session, so it doesn't revert to an auto selection every single time; only when the user has not yet chosen one in which case it will look at browser language. But that's worth testing to be sure.

> Yes please have it configurable on top of auto-detect.

How would configuration and auto detect work together? If the configuration overrides the auto detect, that means auto detect is useless. If auto detect overrides configuration, that means configuration is useless. From a core perspective, I don't see how you can have both. 


In my opinion, 90% of the arguments against the changes to how the language is handled in the manager boils down to change aversion. People are used to having a setting, and feel like they're losing control, while in reality that control is only handed down from the admin to the users themselves who can now easily change their language themselves. And by automatically detecting the browser language, people don't even have to think about it in most cases: it just works!

[...]

PR #14046, opened 2018-08-14
Ability to switch manager language on the fly in the manager

alroniks:
### What does it do?
It adds an ability to switch the language of the manager on the fly in the manager by the user.
Also improved list of native names of languages, now it removed from translation and stored as a static array. 

### Why is it needed?
After refactoring, login pages system setting manager_language was removed and not it impossible to switch the manager language being logged in the manager. You have to log out, select proper language and login back. WTF?
Also, native names are used once on the login page, but being placed to lexicons they create a real hell for translators on Crowdin, especially when they are native and shouldn't be translated. 

### Related issue(s)/PR(s)
None

P.S. Original PR was massively updated by new code so I had to change descriptions and title.

How it works now (video):

langswitcher

Mobile view

langswitchermobile


Jako:
I would prefer a more general solution with a `language_name` key in the core lexicon, a system setting that could be filled with a comma separated list of possible login languages (that defaults to all languages the core is translated to) and a cache file where all possible language_names are collected in.

alroniks:
I don't like the idea of adding yet one system setting for the thing that pretty generic. The current list is not so big to cut it more. But moving names to more common place may make sense. And I think caching static array of 31 elements absolutely useless.

Jako:
If each language contains an `language_name` key, the caching should be done. Otherwise the login manager controller has to load all (i.e. core) language files each time and create an array of `language_name` values of each existing language.

The modLexicon::NATIVE_LANGUAGES array could maybe created somehow during the build process.

alroniks:
In this case, NATIVE_LANGUAGES is a static cache of languages names. So it works as you are describing, but implementation a bit different. And now manager login gets it once, no any parsing files. Menus, where lexicons used, cached already.

Jako:
It executes the lexicon 31 times in https://github.com/modxcms/revolution/pull/14046/files#diff-beeddbdf99549badfa0fdae61172b0b5R93 and for each lexicon one additional cached lexicon file is loaded. One Lexicon is loaded by default depending on the login language. The other ones don't have to be loaded, if the result of the getLanguageMenu is cached.

alroniks:
But it requests lexicons from one file that loaded after login. See this file https://github.com/modxcms/revolution/pull/14046/files#diff-db3d3b4ad891730dac246e89bceb9129. It used to build language menu with native names and provide translations of this language to the language that user selected on the login page. So it works fine without additional requests.
When user select language before entering, it sets manager_language, manager loaded with all lexicons for this language. So reloading of lexicons happens only when the user switched the language.
`$this->xpdo->lexicon('language_' . $code)` - it means only get translation by key, not load files for each language.

Jako:
I just want to get rid of the modLexicon::NATIVE_LANGUAGES array and make that automatic generated by existing languages. One core lexicon of one language should contain a `language_name` key that is used to build a cached combined array, that is used to prefill the language switch dropdown. I don't like the idea of having language strings not managed by a translation system even if it is a quite short list.

jonleverrier:
I thought perhaps the language options might have been better placed in the `User` folder, as it's relevant and specific to the `User`. It also declutters the side bar, and something less to hide with ACL's.

Why was the manager language per user system setting removed ?

alroniks:
> Why was the manager language per user system setting removed?

See this discussion https://github.com/modxcms/revolution/issues/13883

SnowCreative:
Yes, under the User menu makes more sense.

Jako:
Thanks, looks good!

Jako:
The translations have to be added to Crowdin after the PR containing the english translation is merged. 

Other languages than english are not updated automatically at Crowdin. Maybe @theboxer has to be contacted to do that, because most translators don't have write access to all languages.

alroniks:
> The translations have to be added to Crowdin after the PR containing the English translation is merged.

It's why I added native languages predefined in their files. Otherwise, it will look strange for people who will want to use it. Also, translation takes time but the system should look properly regardless of our process. So I would say that that we have to accept such a compromise.

alroniks:
@Jako I finished work. Also, I updated description by new videos of work.
@Mark-H Have a look also, please.

Jako:
Works fine. The only thing issue is the height of the language switcher caused by the two line output in the flyout.

alroniks:
> The only thing issue is the height of the language switcher caused by the two line output in the flyout.

There is overflow-y: auto and this list is scrollable. It was impossible when the item was in the top level menu, but now it works fine.

Jako:
It could be changed to one line like in this screenshot:

Screenshot

The spans just need `display: inline-block; width: auto; margin-left: 6px`

alroniks:
But scrolling via overflow should be left for laptop screens with small height. 
I would prefer to leave current styles to get the menu in one style.

Jako:
The one line menu still has a scroll bar in a browser on 1280x1024, but the height of the flyout is reduced (less scrolling needed).

Mark-H (approved):
I honestly don't really care if it's single or double line - this is a massive improvement for people that need different manager languages in 3.x, so I'm going to approve it.

We can continue to iterate on it later if it is proven to be desirable to do it differently.

alroniks:
Hi @Jako Do you agree with Mark or question about height is still actual?

Jako:
I am fine with one or two lines in that menu, so get this merged.

alroniks:
@Jako I need your approval in this case :relaxed:

Jako (approved):

PR #14047, opened 2018-08-15
Actual versions of PHP in Travis

alroniks:
### What does it do?
It removes hhvm as unsupported from Travis builds. As well it adds actual 7.2 version of PHP.

### Why is it needed?
HHVM build always failed so it makes sense just remove it to do not load Travis by useless tasks. Also, the current actual and latest version of PHP is 7.2 and we need to be sure that MODX works on the actual version of PHP as well.

### Related issue(s)/PR(s)
None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,19 +7,17 @@ script:
 
 # Set php versions
 php:
+  - 7.2
   - 7.1
   - 7.0
   - 5.6
   - nightly
   - 5.5
   - 5.4
-#  - 5.3
-  - hhvm
 
 matrix:
   allow_failures:
     - php: nightly
-    - php: hhvm
 
 # database credentials
 mysql:


JoshuaLuckers (changes requested):
PHP 5.4 and 5.5 are end of life and no longer supported by PHP. The MODX server requirements also recommend PHP 5.6 or higher (PHP 5.5 as minimum. However I don't think we should support End of Life versions of PHP).

alroniks:
@JoshuaLuckers It was discussed many times, while MODX can work on old versions PHP, we will support it. But if some new functions will require more modern version, we'll drop support. So now config based on composer file, that requires 5.4 as a minimum. 
@Mark-H has nice charts with the actual state of using PHP versions across MODX installations and it says that still a lot of people use old versions.

JoshuaLuckers:
I remember the discussion, good point!

JoshuaLuckers (approved):

Jako (approved):

Mark-H:
@opengeek Were you looking at this the other day?

opengeek:
@Mark-H wrote:
> @opengeek Were you looking at this the other day?

No. This looks good to me though.

opengeek (approved):

PR #14050, opened 2018-08-17
Add a reloadContext option to modResourceUpdateProcessor

gvozdb:
Added context reload check for bulk actions on resources, for example for bulk import of goods.

### What does it do?
Reducing execution time of the processor.

### Why is it needed?
При массовой выгрузке товаров у нас экономится время.

### Related issue(s)/PR(s)

### DIFF
--- a/core/model/modx/processors/resource/update.class.php
+++ b/core/model/modx/processors/resource/update.class.php
@@ -735,7 +735,11 @@ class modResourceUpdateProcessor extends modObjectUpdateProcessor {
         }
         $returnArray['class_key'] = $this->object->get('class_key');
         $this->workingContext->prepare(false);
-        $this->modx->reloadContext($this->workingContext->key);
+        
+        if ($this->getProperty('_reloadContext', true)) {
+            $this->modx->reloadContext($this->workingContext->key);
+        }
+        
         $returnArray['preview_url'] = '';
         if (!$this->object->get('deleted')) {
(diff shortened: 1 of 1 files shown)


alroniks:
@gvozdb From where property `_reloadContext ` is coming?

gvozdb:
@Alroniks From parameters send to processor.

```
$modx->runProcessor('resource/update', [
    ...
    '_reloadContext' => false,
]);
```

alroniks:
@gvozdb In this case, why this parameter starts from underscore (_)? If you want, we can discuss it more on Russian in Telegram, because current solution unclear.

gvozdb:
@Alroniks Yes, please! https://t.me/pavelgvozdb

Mark-H:
@Alroniks Did you have a chat about this?

alroniks:
@Mark-H Yes, a bit. But it still unclear to me. It should increase performance in several specific cases but I disagreed with the current implementation. Need time to check cases provided by Pavel.

Jako:
I think this special case could be better solved with an own processor class that extends modResourceUpdateProcessor and overrides just the cleanup method.

```
$modx->runProcessor('resource/update', [
    ...
    ], [
        'processors_path' => 'whateverpath'
    ]
]);
```

Mark-H:
This probably speeds up import scripts. ImportX also uses the core processor in the same way as @gvozdb posted, I believe. Dropping the underscore would make sense if we want it to be seen as a supported option, but that's also nitpicking.

Mark-H (approved):

JoshuaLuckers (approved):

PR #14053, opened 2018-08-18
Merge 2.6.x into 2.x

JoshuaLuckers:
### What does it do?
Merge missing changes from 2.6.x into 2.x

### Why is it needed?
The 2.x branch shouldn't be behind the 2.6.x branch.

### DIFF
--- a/_build/build.sample.properties
+++ b/_build/build.sample.properties
@@ -11,7 +11,7 @@ git.command = git
 #project.name.fs = modx
 
 # Override to set the version and release strings
-#modx.core.version = 2.6.4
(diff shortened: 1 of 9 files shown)


Mark-H:
I don't think we typically do PRs for this sort of merge.

JoshuaLuckers:
Fair enough, you can close it if you want. What's the normal procedure when 2.x is behind 2.6.x?

opengeek:
It is behind because the security patches last month were cherry-picked from 2.x to 2.6.x. There is no need to merge. We generally merge upstream whenever completing a merge session to a specific branch.

Issue #13412, opened 2017-04-07
Media browser filename containing "&" issues. 


BigBlockStudios:
### Summary
Users can upload but rename or delete files containing linux metacharacters (&) in the file name.  only tested the ampersand (&).  
Trying to rename the file results in an error: The webserver does not have the necessary permissions to update the directory.
Try to delete the file gives error: Please specify a valid directory.: /vhome/accordfinancial.com/httpdocs/assets/AAA&AAA.pdf

file had to be manually removed VIA the shell 

### Step to reproduce
Upload and try to manipulate a file with  "&" in the name.

### Environment
Modx Version 2.5.4 running on centOS 6xx
PHP Version 5.4.45

BigBlockStudios:
just tried updating a different site & uploading/managing files with metacharacters  in it - same problem exists in 2.5.6

mrhaw:
Adding this as clues how I would tackle this
Extjs filter:
http://stackoverflow.com/questions/36061698/extjs-how-to-escape-special-characters-in-post-and-get-xml-data/36071288#36071288

MODX util files
https://github.com/modxcms/revolution/tree/fbf126690f6ce903faeaeb9f1d958211fc2203c9/manager/assets/modext/util

mindeffects:
It's a different thing but having "#" or "+" inside the filename break the image preview thumbnail. But renaming is possible.

OptimusCrime:
I am pretty sure a # in the name will break it no matter what, as the anchor tag has a special meaning in HTTP.

meshkov:
Issue fixed in 3.x.

alroniks:
MODX 3 now use flysystem library for dealing with files and it seems that the issue is already fixed (as @meshkov mentioned). I am closing it but feel free to reopen if the issue still actual.

meshkov:
Tested it in MODX 3.0 in TV with type image and it is break the image preview in TV.
TV with this images does not work on pages.
![7kcndmuer6](https://user-images.githubusercontent.com/1703670/41839834-e98bcdca-786c-11e8-824d-47c644db0145.gif)

alroniks:
Hi @meshkov Does it mean that we should reopen the issue?

meshkov:
Hi @Alroniks yes, but it problem with TV, not with media browser. Maybe new issue?

alroniks:
TV utilize media browser so it is related. Reopened.

PR #14054, opened 2018-08-21
Delete files in Media Browser with '&' in the file name

pbowyer:
### What does it do?
In November 2016 extra protection was added to [prevent local file inclusion/traversal/manipulation](https://github.com/modxcms/revolution/commit/d3df889703f712e71eb0cdca4f9b316731f51143#diff-ea84051473ef39b6641bf6f3d3f33e5a). One side-effect was that files uploaded with '&' in the file name could no longer be deleted.

With MODX 2.6.5 it is still the case that files uploaded do not have their filename encoded with `htmlspecialchars()`, but requests to delete a file do.

This PR reverses that so the treatment of file names is symmetrical, and a file that is uploaded can be removed.

### Steps to reproduce
1. Log into the MODX manager and go to `Media > Media Browser`.
2. Upload an image with `&` in the filename. Mine is `Test - B&W.jpg`
3. Confirm it's uploaded, you can view it etc
4. Right click and choose `Delete file`
5. Without the patch, you will get a popup saying _Please specify a valid directory.: /paas/cXXX/www/uploads/Test - B&W.jpg_. The file will not be removed
6. With the patch, the file will be deleted

### Security considerations
I am not a security professional. But I don't see this making the system less secure.

Not sanitising filenames on upload is a potential issue; to be able to delete files we need to have the same behaviour when deleting. Even if new filenames were sanitised in the same way, files previously uploaded would need to be deleted.

### Related issue(s)/PR(s)
#13412

### DIFF
--- a/core/model/modx/processors/browser/file/remove.class.php
+++ b/core/model/modx/processors/browser/file/remove.class.php
@@ -33,7 +33,7 @@ class modBrowserFileRemoveProcessor extends modProcessor {
         if (empty($file)) {
             return $this->modx->error->failure($this->modx->lexicon('file_err_ns'));
         }
-        $file = preg_replace('/[\.]{2,}/', '', htmlspecialchars($file));
+        $file = preg_replace('/[\.]{2,}/', '', $file);
 
         $loaded = $this->getSource();
         if (!($this->source instanceof modMediaSource)) {


Jako (approved):
Checked. Working fine.

Jako:
I don't see a security issue either. The filename exists without entity encoded chars and directory traversal is still prohibited.

PR #14056, opened 2018-08-22
Sanitize parameters in the RestServiceRequest class

arjen-t:
Added missing sanitize logic for sanitizing the parameters in the modRestServiceRequest class.

### What does it do?
It sanitize the parameters in the modRestServiceRequest class.
The sanitizing is based on the modX sanitizePatterns.

### Why is it needed?
To avoid potential  XSS and other attacks

### Related issue(s)/PR(s)
-

### DIFF
--- a/core/model/modx/rest/modrestservice.class.php
+++ b/core/model/modx/rest/modrestservice.class.php
@@ -373,8 +373,7 @@ class modRestServiceRequest {
 			default:
 				break;
 		}
-
-        if ($this->getProperty('sanitize', false) {
+        if ($this->service->getOption('sanitize', false)) {
             $this->sanitizeRequest();
         }
     }
--- a/core/model/modx/rest/modrestservice.class.php
+++ b/core/model/modx/rest/modrestservice.class.php
@@ -271,6 +271,8 @@ class modRestServiceRequest {
     public $headers = array();
     /** @var array $parameters The request parameters on the request */
     public $parameters = array();
+    /** @var bool $sanitize Sanitize the request parameters */
+    public $sanitize = false;
 
     /**
      * @param modRestService $service A reference to the modRestService instance
@@ -374,7 +376,9 @@ class modRestServiceRequest {
 				break;
 		}
 
-		$this->sanitizeRequest();
+        if ($this->sanitize) {
+            $this->sanitizeRequest();
+        }
     }
 
     /**
--- a/core/model/modx/rest/modrestservice.class.php
+++ b/core/model/modx/rest/modrestservice.class.php
@@ -271,8 +271,6 @@ class modRestServiceRequest {
     public $headers = array();
     /** @var array $parameters The request parameters on the request */
     public $parameters = array();
-    /** @var bool $sanitize Sanitize the request parameters */
-    public $sanitize = false;
 
     /**
      * @param modRestService $service A reference to the modRestService instance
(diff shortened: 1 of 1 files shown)


Jako:
Seems ok for me, but it needs an option to switch it off by a property.

arjen-t:
@Jako thanks for your feedback. Maybe it's a better option to synchronize the behaviour the same way as it is in the core. This [method](https://github.com/modxcms/revolution/blob/9c833bc6f3850243d5c171103b7df7d5fd3a4259/core/model/modx/modrequest.class.php#L365 ) is using the system setting for switching on/off for sanitising the POST params, by default the rest of the incoming request is sanitised. 
I'm more convinst to adapt this logic, so modx behaviour is consistent and no suprises/assumptions can occure.

I hope that @Mark-H  or @opengeek can give their feedback on this one?

Mark-H:
This seems useful, but also changes behavior in potentially unpredictable ways. At the very least it should have a boolean config option passed to the modRestService config, and I'd be inclined to suggest that to be off by default for backwards compatibility reasons. (Alternatively, add the config option defaulting to true, and target 3.x, so it can be documented as a breaking change.)

arjen-t:
@Mark-H  Thanks for your feddback/input. The reason why I implemented the sanitising part had to do with one motivation, security.
I noticed that the entire request handling in the REST implemention is missing this. 
The first time I was using the REST functionality mine assumption was that the entire request had been sanitised by the request handler via the modx core.
I did some local testing and notice that the entire request by-pass this core functionality, which means no sanitising.
This feels odd to me. Maybe I'm missing some history on this topic which I'm not aware off. 
But it doesn't feel right to call this a opt-in instead of an opt-out (refering to your 1.6.X backwards compatibilty)

As for your understanding the entire request at this current moment is not been sanitised. So this can cause funny stuff if the developer is not aware of this and accept the input as it is.

Let me know what is your point of view about this. I personal feel much more comfortable with the approach, better safe then sorry, if it comes to security.

Cheers

Mark-H:
I completely agree with the principle of making security measures the default - but this is a backwards compatibility break which we need to keep in mind, too. 

If I'm reading it correctly, the proposed change will remove any MODX tag, script tag, and `&#entities;` from any request to an API built with the modRestService. Let's say someone built an API to manage their templates through an app (mxManager?), this change will filter out the MODX tags that are essential to that API. 

What might be a good middle ground is to add it, defaulting to off, in 2.7. That gives you the ability to start opting-in to the more secure behavior ASAP. Then for 3.0 we change it to being enabled by default. That planned change can be documented right away, so people building (/maintaining) their API now can decide if it needs the sanitisation or not.

Mark-H (changes requested):
Needs to handle backwards compatibility for inclusion in 2.x, as discussed.

Jako:
It is now optional.

Jako (dismissed):

Mark-H (changes requested):
I don't see how this allows it to be enabled without subclassing `modRestServiceRequest` - ideally this would be a property one can provide to a `modRestService` instance that the request checks in the `__construct`, or something like that.

Jako:
It is now configurable by an option.

Since sanitizing runs very early (before the controller is instantiated), the sanitize option can't be set by the controller. It has to be specified before initializing the modRestService.

Mark-H (approved):
Thanks! :)

arjen-t:
@Jako thanks for your input (sorry I respond so late, lack of time)

opengeek (approved):

alroniks:
As this PR was marked for 3.x I've merged changes into 3.x instead of 2.x, however base branch still 2.x. By this reason, it can not be closed automatically, so I am closing this one by hands.

PR #14057, opened 2018-08-22
Removing obsolete contentType field from resource

alroniks:
### What does it do?
It removes usage of `contentType` field that came as a legacy from MODX Evolution but still exists in the system with fallbacks. 

### Why is it needed?
As we are working on MODX 3, that means changing major version, all legacy should be removed but withing respecting to keeping BC where it is possible.

### Related issue(s)/PR(s)
discussions in chats


JoshuaLuckers:
It would be a good idea to mark the field as deprecated in 2.x.

cjholowatyj:
Is this actually legacy though? I use the contentType all the time to develop MODx-generated CSS, JS, and JSON files... what is obsolete about it? (Maybe I'm missing something)

opengeek:
@cjholowatyj You should be using the `content_type`, which links to the `modContentType` object (which is represented by the table `modx_content_type`) and is available on a modResource via the `ContentType` related object.

cjholowatyj:
@opengeek ok I stand corrected 😊 it appears I have been setting both fields in my code this whole time!

👍

core/model/modx/modrequest.class.php
```diff
@@ -197,11 +197,6 @@ public function getResource($method, $identifier, array $options = array()) {
                 $resource->fromArray($cachedResource['resource'], '', true, true, true);
                 $resource->_content = $cachedResource['resource']['_content'];
                 $resource->_isForward = $isForward;
-                if (isset($cachedResource['contentType'])) {
```
opengeek, line 200: I don't think this is right. The cached resource is using this to restore the object.

core/model/modx/modresource.class.php
```diff
@@ -556,58 +556,16 @@ public function getTemplateVars() {
...
-            case 'contentType' :
-                if ($v !== $this->get('contentType')) {
-                    if ($contentType= $this->xpdo->getObject('modContentType', array ('mime_type' => $v))) {
-                        if ($contentType->get('mime_type') != $this->get('contentType')) {
-                            $this->addOne($contentType, 'ContentType');
-                        }
-                    }
-                }
-                break;
-            case 'content_type' :
```
OptimusCrime, line 577: I think we need this?

Mark-H: @OptimusCrime Can you be more specific about why?

pbowyer:
This is used in Articles https://github.com/modxcms/Articles/blob/develop/core/components/articles/model/articles/articlescontainer.class.php#L232

JoshuaLuckers:
I will change it in Articles if needed

core/model/modx/modrequest.class.php
alroniks: Yes, in a case when it was an old cache and need to add the proper modContentType object to `content_type` field. I don't think that anywhere exists so old site with so old cache where `content_type` field doesn't exist.

core/model/modx/modresource.class.php
alroniks: Nope. It used for setting a legacy `contentType` field. As now it won't be stored in the cache, not necessary to handle this additionally.

alroniks:
Yes, better to fix it in extras against keeping old code in the system. Entropy in the core should be reduced.

pbowyer:
From inspecting 800 of the 805 extras in the directory (which 5 are missing is unknown), it's also used:
* pdfParser (pdfparser.plugin.php line 38)
* bloodline (bloodline.plugin.php lines 166, 170, 179)
* lingua (linguarequest.class.php line 161)
* Advsearch (writeindex.php, line 155)
* Articles (listed above)
* Statcache (StatCache.php, line 81)
* Statiker (statikerbuilder.class.php, line 177)
* xhtmlbeautify (xhtmlbeautify.class.php, line 62)
* frontpage (frontpage.class.php, line 1020)

And possibly:
* Handyman (view.class.php line 96)
* Shopkeeper (both copies of context_switch.php line 250)
* tinymcewrapperdiy (tinymcewrapperdiy.plugin.php, lines 661, 693)
* virtunewsletter (virtunewsletter.class.php, line 1711)

core/model/modx/modrequest.class.php
opengeek: It's not old cache. This is how the new ContentType objects are loaded from cache. Without it, the ContentType object is not available from what I can see.

opengeek: I think this needs to be refactored so that the ContentType is loaded from cache in a different way?

Mark-H: @Alroniks It looks like you took out the contentType out of the cache in modCacheManager->generateResource, and because of that also removed it from here - any particular reason for that? I suspect that could be a (small) performance improvement to keep it in.

Mark-H (changes requested):
See https://github.com/modxcms/revolution/pull/14057#discussion_r230518082

JoshuaLuckers:
@Alroniks do you have time to finish this PR?

alroniks:
Hi @JoshuaLuckers
Not sure that it can be fixed in this PR, it seems need start from scratch. I will try to find some time on two next weeks.

Mark-H:
Past the cut-off for breaking changes in alpha1, so pushing further.

Ibochkarev:
Are there any progress on this PR?

Mark-H:
Superseded by #15822

Issue #14059, opened 2018-08-24
Middle mouse button does not work in the tree of elements (resources and files)


Ruslan-Aleev:
This problem has been discussed earlier:
https://github.com/modxcms/revolution/issues/12374
https://github.com/modxcms/revolution/pull/13103

But the **mouse does not work** in the browser, although **"ctrl + click" - works**.

**In Windos / Linux OS the middle button click event does not hold the proper information in the onclick event, but it does in onmouseup.**

p.s. Without the middle mouse button - as if the finger was cut off, especially if it's standard browser behavior :)

MODX 2.6.5
Google Chrome 68.0.3440.106
Firefox 61.0.2
Probably other browsers

JoshuaLuckers:
Relevant information can be found in the comments of PR #14061: https://github.com/modxcms/revolution/pull/14061#issuecomment-416001521

Ruslan-Aleev:
@JoshuaLuckers There is information about this bug?
The question is important :), periodically messages turn up on the network about it, Windows users are many times more than the rest of the OS, and without the middle mouse button the interaction with MODX is getting worse.

JoshuaLuckers:
I am not able to reproduce the issue. The combination of your OS and/or mouse (might be the driver) are not sending the right event. Since I'm unable to reproduce it I can not fix it myself.

Have you tried it with another mouse on another machine?

Ruslan-Aleev:
@JoshuaLuckers Yes, I tested on different devices and OS, I found out that on Windows (I tested on 7 and 10), on Linux (Mint) the wheel does not work at all (the current window opens in a new tab).
It only works on Mac OS and does not work completely, for example, the "+" icon next to the category, for example, chunks, also does not open a new window.

JoshuaLuckers:
Thanks for testing. Did you test with the same mouse or with different ones?

Ruslan-Aleev:
Tested on different mouse, computers and OS, and several people helped in the test with own mouse :)

JoshuaLuckers:
Are you available on Slack? I want you to test something if that is possible.

Ruslan-Aleev:
@JoshuaLuckers Yes, I am in slack - Ruslan Aleev, including https://modxcommunity.slack.com

JoshuaLuckers:
Fixed by PR #14243

PR #14061, opened 2018-08-25
Allow a middle mouse click to open links in a new tab/window 

JoshuaLuckers:
### What does it do?
It detects if the middle mouse key is pressed and makes the browser handle the way the link should be opened (new tab/window)

### Related issue(s)/PR(s)
#14059

### DIFF
--- a/manager/assets/modext/core/modx.layout.js
+++ b/manager/assets/modext/core/modx.layout.js
@@ -396,10 +396,15 @@ MODx.LayoutMgr = function() {
             var url = parts.join('&');
             if (MODx.fireEvent('beforeLoadPage', url)) {
                 var e = window.event;
-                if (e && (e.button == 1 || e.ctrlKey == 1 || e.metaKey == 1 || e.shiftKey == 1)) {
-                    // Keyboard key pressed, let the browser handle the way it should be opened (new tab/window)
+
+                var middleMouseButtonClick = (e && (e.button === 4 || e.which === 2));
+                var keyboardKeyPressed = (e && (e.button === 1 || e.ctrlKey === 1 || e.metaKey === 1 || e.shiftKey === 1));
+                if (middleMouseButtonClick || keyboardKeyPressed) {
+                    // Middle mouse button click or keyboard key pressed,
+                    // let the browser handle the way it should be opened (new tab/window)
                     return window.open(url);
                 }
+
                 location.href = url;
             }
             return false;


Ruslan-Aleev:
@JoshuaLuckers I tested on MODX 2.6.5 - the problem still exists.
Google Chrome 68.0.3440.106, Windows 7.

Jako:
@Ruslan-Aleev Are you sure, that the changed script code is loaded in your browser? Maybe you have to run the things described in _build/templates/default/README.md

Ruslan-Aleev:
@Jako I'm probably doing something wrong, that's how I checked:
1) Installed a clean MODX 2.6.5 standard distribution and replaced the file manager/assets/modext/core/modx.layout.js with corrected (clean caсhe in manager and browser), checked mouse button - the problem still exist.
2) I deployed through "git clone" 2.7.0-dev with this PR, checked the modx.layout.js file for edits - all the edits were in place. Again tested mouse - the problem still exist :)

Does it work for you?

JoshuaLuckers:
Disable JS compression in system settings. If it still doesn’t work let me known where you tried it.

Ruslan-Aleev:
@JoshuaLuckers Does not work, turned off compression js.
Test in MODX 2.6.5, 2.7.0-dev, Google Chrome 68.0.3440.106, Windows 7.
I can send access from the site to the email.

JoshuaLuckers:
Thanks for testing.

Can you paste the following code into the dev console of your browser and paste the result it returns when middle clicking?

```
document.body.onclick = function (e) {
  if (e) {
    console.log([e.which, e.button]);
  }
}
```

Ruslan-Aleev:
When the middle (or right) button is pressed, nothing is written to the console, if press the left button, it is written - (2) [1, 0].

mouse_info

Also noticed the mention of "scroll-blocking mousewheel" from Google Chrome in console, here is the link https://www.chromestatus.com/feature/5745543795965952

OptimusCrime:
What OS are you on? And does the middle click work on "regular" links?

Ruslan-Aleev:
@OptimusCrime I'm using Windows 7 Home Edition.
On "regular" links the middle button works, for example, in the top menu MODX manager (Content | Media | Extras ...) there are no problems.

OptimusCrime:
That is really odd, the code posted above should be able to detect the button click. Here is the MND example: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#Example could you try it?

Paste it in the console:

```
var whichButton = function (e) {
    // Handle different event models
    var e = e || window.event;
    var btnCode;

    if ('object' === typeof e) {
        btnCode = e.button;

        switch (btnCode) {
            case 0:
                console.log('Left button clicked.');
            break;

            case 1:
                console.log('Middle button clicked.');
            break;

            case 2:
                console.log('Right button clicked.');
            break;

            default:
                console.log('Unexpected code: ' + btnCode);
        }
    }
}
```

And then:

```
document.body.onclick = whichButton;
```

If that does not work, could you also try:

```
document.body.onmouseup = whichButton;
```

Ruslan-Aleev:
@OptimusCrime The script works with **document.body.onclick = whichButton;** like this:

- Click on the left mouse button - console writes "Left button clicked.";
- Click on the middle mouse button - nothing is written to console;
- Click on the right mouse button - nothing is written to console.

With **document.body.onmouseup = whichButton;** works fine:
- Click on the left mouse button - console writes "Left button clicked.";
- Click on the middle mouse button - console writes "Middle button clicked.";
- Click on the right mouse button - console writes "Right button clicked.".

OptimusCrime:
Yeah, that was what I thought. The event does not hold the proper information in the onclick event, but it does in onmouseup. Could we somehow change this in extjs, or does that break everything?

Jako:
First it should checked on touch devices, to be quite sure, that is does not break clicking there.

JoshuaLuckers:
@Ruslan-Aleev does the "middle mouse click" do anything at all? For example: if you middle mouse click the "create new resource" button, does anything happen?

Ruslan-Aleev:
@JoshuaLuckers No, the middle mouse button does not react at all, only in the menu where there are physical links, for example, in top menu.
p.s. In my opinion this is happening in all versions of Windows, a little later I'll try on Windows 10 and with a different mouse.

JoshuaLuckers:
Good to know. To solve the issue you are experiencing a more complex change is needed.

This PR does not solve the issue I marked as related but it should be accepted nonetheless.

In the issue you reported the additional info we have now should be added. Then a new PR can be created for the more complex changes.

Ruslan-Aleev:
@JoshuaLuckers OK, thank you for the information

Ruslan-Aleev:
@JoshuaLuckers Windows 10 has this problem too

Mark-H (approved):

alroniks (approved):

Ruslan-Aleev:
@OptimusCrime Can you create a PR with the necessary edits to make the middle button work? I will test it on different devices.
Middle button are very necessary for work :)

OptimusCrime:
The problem is that I am not entirely sure how to do that. Does there exist an issue for that? Could you reference me in it?

Ruslan-Aleev:
@OptimusCrime For some reason, I did not receive a notification from github, I did not see your messages.
Here I described the problem - https://github.com/modxcms/revolution/issues/14242
and here is the solution for the elements in the tree - https://github.com/modxcms/revolution/pull/14243

PR #14066, opened 2018-08-28
Fix uninstall errors

theboxer:
### What does it do?
Fixes not passing property to a lexicon string & an error when uninstall fails.

### Why is it needed?
Currently if package uninstall fails, current messages are shown in the console:
```
Recoverable error: Object of class xPDOObjectVehicle could not be converted to string
Error uninstalling package with signature: Object
Error removing package with signature: [[+signature]]
```


MrRoco:
Testing it on the #bughunt

MrRoco:
Tested & confirmed working #bughunt 

Tested it by changing the write permissions of the zip.

PR #14069, opened 2018-08-31
Applied unused before configuration options with charset and collation during database creation

alroniks:
### What does it do?
It passes `$containerOptions` variables with proper charset and collation to the `createSourceContainer` method that creates a database during the installation.

### Why is it needed?
I have often been faced with it a long time, using Gitify (and other tools for CLI installation) when after installing the MODX, all non-latin strings become "???" because the wrong charset was applied to the database and tables (the default charset, that was configured for MySQL server, usually `latin`). It was happening even if charset and collation were defined in the configuration (they are predefined as urtf8 by default, but were ignored). 
This changes should fix the wrong behavior for new installations.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/setup/includes/tables_create.php
+++ b/setup/includes/tables_create.php
@@ -100,7 +100,7 @@ if (!$connected) {
     $dsnArray= xPDO :: parseDSN($modx->getOption('dsn'));
     $containerOptions['charset']= $install->settings->get('database_charset', 'utf8');
     $containerOptions['collation']= $install->settings->get('database_collation', 'utf8_general_ci');
-    $created= $modx->manager->createSourceContainer($dsnArray, $modx->config['username'], $modx->config['password']);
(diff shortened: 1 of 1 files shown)


Mark-H:
If this does what I think it does, I'm going to buy a couple of beers for this. But I'll need to pull it down to test and can't review it from the source alone unfortunately.

JoshuaLuckers:
Since it's a rather small file, would you mind to reformat all the code in the file?

alroniks:
Yes, @Mark-H, it does what it does :) I had a headache with it using Gitify and finally found the solution.
@JoshuaLuckers I heard about plans to feed the code to an automatic formatter in the future, so I would leave it as is to do not useless work.

alroniks:
Hi @opengeek 
Could you have a look? It's a stupid issue and very simple to check.

Jako (approved):

Issue #14075, opened 2018-09-06
Double-click of image in Media Browser goes to blank page


guyinpv:
### Summary
Double-click  image in Media Browser opens blank page.

### Step to reproduce
Open Media Browser and go to a folder with JPGs in it. Double-click on a JPG image (could be other images formats too). Multiple people confirmed this behavior already on Slack.

### Observed behavior
Blank page comes up. And further, hitting the back button does not take you back to Media Browser, but back to whatever page you had open before Media Browser. So this really just knocks you out of browsing media entirely.

### Expected behavior
I'm assuming the best behavior here would be a popup preview window. There already exists a popup preview for images, the double-click action should open that as well.

### Environment
Using 2.6.5-pl on  Apache server. Tested on a very clean install.

JoshuaLuckers:
I can confirm this issue.

Jako:
> I'm assuming the best behavior here would be a popup preview window. There already exists a popup preview for images, the double-click action should open that as well.

What kind of popup do you mean? There is a sidebar with a preview.

jonleverrier:
For reference, this is the sidebar @Jako is talking about;

![screen shot 2018-09-07 at 13 52 19](https://user-images.githubusercontent.com/2373940/45217737-778edf00-b2a5-11e8-94cb-179e1fe7afe4.png)

guyinpv:
@jonleverrier In the screenshot you posted, you can click on the image in the upper-right of the sidebar and it will show a popup image preview at full size of the image.

I'm thinking a double-click would just open the same popup.

Further, that popup shows the full image but has scroll bars, maybe a slightly improved image preview popup is in order, like a full size lightbox for example.

Double-click needs some kind of behavior. I mean, it is actually doing something now, it goes to this blank page. Maybe in the past this had a different behavior? Why the blank page?

Regardless, using the same popup as the sidebar is an easy-button fix perhaps.

Jako:
I'm a bit confused. First you are talking about double clicking on an image in the media browser and now you are talking about a popup issue when clicking on an image in the media browser sidebar.

I could confirm the double clicking issue in in the browser and tried to fix that. 

I don't think that a double click should open a popup. If you use that browser i.e. in a template variable, double clicking a file will fill the TV value and close the window. If a similar browser does a totally different thing on a double click (instead of nothing), this feels like bad UX for me.

Please open another issue for the sidebar image popup.

guyinpv:
[...]

PR #14076, opened 2018-09-07
Fixing Double-click of image in Media Browser goes to blank page

Jako:
### What does it do?
Make sure that MODx.Media.onSelect code is not executed by removing the code.

### Why is it needed?
MODx.Media.onSelect has hidden `this.config.animEl || null` before, which means that MODx.Media is hidden accidentally. That way a double-click of an image in the Media Browser goes to blank page.

### Related issue(s)/PR(s)
#14075

### DIFF
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -1117,6 +1117,7 @@ Ext.extend(MODx.Media, Ext.Container, {
     }
 
     ,onSelect: function(data) {
+        return false;
         // @todo make sure this is never used
         console.log('MODx.Media#onSelect', data);
         var selNode = this.view.getSelectedNodes()[0];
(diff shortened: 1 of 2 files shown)


jonleverrier:
I've tested this PR and it works as described.

JoshuaLuckers (commented):
Can the unreachable code be removed?

Jako:
I really don’t know where it was used, so yes.

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -1117,6 +1117,7 @@ Ext.extend(MODx.Media, Ext.Container, {
     }
 
     ,onSelect: function(data) {
+        return false;
         // @todo make sure this is never used
         console.log('MODx.Media#onSelect', data);
```
alroniks: I guess console.log call should be definitely removed.

alroniks (changes requested):

Jako:
I have removed all not needed 'make sure this is never used' code and the logging in those methods. The only part where a double click in the media browser leads to an action still works (doubleclick a file during editing a file selecting template variable or a file insert in the RTE). I did not find another place where the double click is used.

gpsietzema:
Works as expected. Tested in latest versions of firefox/chrome/safari and edge.

Issue #11666, opened 2014-07-10
Feature request: Copy complete file path from Media Browser


pyrographics:
The contextual menu on the media browser should have the feature to copy the complete path to the file to your clipboard so you can paste it anywhere in the manager you need it that does not allow drag and drop.  Currently users have to guess the path to the file which may be hidden if using a pre-defined media source.  The media source base path (from site root) should also be shown somewhere so users better understand where the files are stored.

alroniks:
PR is merged, closing as solved.

PR #14078, opened 2018-09-09
[ReUp] Copy file/folder path to clipboard

JoshuaLuckers:
### What does it do?
It adds an option to copy the path to a file/folder in the MODX file tree and media browser.

### Why is it needed?
To make life easier for manager users.

### Related issue(s)/PR(s)
Issue #11666

### DIFF
--- a/core/lexicon/en/file.inc.php
+++ b/core/lexicon/en/file.inc.php
@@ -15,6 +15,7 @@ $_lang['file_delete_folder'] = 'Delete Folder';
 $_lang['file_download'] = 'Download File';
 $_lang['file_download_file'] = 'Download File';
 $_lang['file_download_unzip'] = 'Unzip File';
+$_lang['file_copy_path'] = 'Copy File Path';
 $_lang['file_edit'] = 'Edit File';
 $_lang['file_err_ae'] = 'File %s already exists';
 $_lang['file_err_chmod'] = 'An unknown error occurred while trying to chmod the target.';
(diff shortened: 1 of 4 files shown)


jonleverrier:
Just tested this PR @JoshuaLuckers. `Copy File Path` works fine, but I can't seem to `Copy Folder Path`. When I click on that option, nothing happens after click, and the path isn't saved to my clipboard.

There is also an error in my console:

```
Uncaught ReferenceError: dummRelativePathInput is not defined
    at MODx.tree.Directory.copyRelativePath (modx.tree.directory.js:667)
    at h.Event.fire (ext-all.js:21)
    at S.fireEvent (ext-all.js:21)
    at S.onClick (ext-all.js:21)
    at S.onClick (ext-all.js:21)
    at HTMLUListElement.I (ext-all.js:21)
```
I tested in Chrome Version 68.0.3440.106 and Firefox

jonleverrier:
Also, probably out of scope, but it would be great to make the clipboard icon at the bottom of the media browser, copy to clipboard when it's clicked on.

screen shot 2018-09-10 at 13 10 28

JoshuaLuckers:
@jonleverrier where did you try the `Copy Folder Path`? In the tree or media browser?

jonleverrier:
@JoshuaLuckers I tried both ways - same issue for me

JoshuaLuckers:
@jonleverrier The  clipboard icon is actually a pseudo element and "clicking it" triggers focus on the input element. If the input element is focussed it automatically selects the text. It might be a better idea to change this in MODX 3 instead of 2.x.

jonleverrier:
@JoshuaLuckers I've tested the latest update. 

`Copy Folder Path` and `Copy File Path` work perfectly!

Mark-H (approved):
Nice!

alroniks (approved):

Issue #13964, opened 2018-07-03
MODX 3: Dashboard - Buttons not full width. Bug or by design?


jonleverrier:
### Summary
The new dashboard buttons don't appear to be full width, which looks odd when comparing this block of content to the other full width content on the same page. I appreciate this may be by design.

![screen shot 2018-07-03 at 15 31 08](https://user-images.githubusercontent.com/2373940/42222739-4812a2aa-7ed6-11e8-9275-4307ec125bf8.png)

### Step to reproduce
Login to the manager and view the dashboard buttons.

### Observed behavior
Notice how the buttons are not full width compared to the rest of the content on the page.

### Expected behavior
For the buttons to go full width, like the working demo below:

![modx3_dashboard_buttonbug](https://user-images.githubusercontent.com/2373940/42223101-3cceb900-7ed7-11e8-8763-970e824c43b4.gif)

### Fix

```
.dashboard-buttons {
    display: flex;
    /* flex-wrap: wrap; */
    justify-content: center;
    width: calc(100% + 1rem);
    margin: -1rem 0 0 -1rem;
    flex-wrap: nowrap;
}
```
```
.dashboard-buttons .dashboard-button {
    display: flex;
    align-items: center;
    background-color: #FFF;
    border-radius: 3px;
    box-shadow: 0 0 5px 0 rgba(0, 0, 0, 0.1);
    margin: 1rem 0 0 1rem;
    padding: 20px;
    text-decoration: none;
    color: #53595F;
    /* max-width: 185px; */
    justify-content: flex-end;
    flex-grow: 1;
}
```
```
.dashboard-buttons .dashboard-button-wrapper {
    flex-grow: 1;
    margin-left: 10px;
}
```

### Environment
MODX3 Alpha v 3.0.0-dev
From https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99

muzzwood:
I remember seeing a mock-up of the dashboard buttons looking differently (much better) and had just assumed this was a temporary solution to build the functionality while waiting for the final design to be applied.
Anyone know?

jonleverrier:
Can anybody confirm if this needs fixing. Perhaps @gpsietzema or @Mark-H . If so, i'll send in a PR.

alroniks:
I would like to fix it too. Maybe it's by design, but in this case it's a bad design. 
[...]

PR #14080, opened 2018-09-10
[3.x] [ReUp] Make dashboard buttons go full width

jonleverrier:
### What does it do?
Makes the dashboard buttons go full width and fill all horizontal space. When the largest size is selected, it makes the buttons _go large_.

screen shot 2018-09-11 at 09 38 54

### Why is it needed?
To make MODX great again.

This is a quick fix. Some native flexbox kicks in, in a row of 6 buttons, on smaller viewports, the last 2 items might be split 50/50, whilst the top row will be split 4 across.

This can be fixed, but it gets complicated rather quickly;
1. How many buttons are there? Just 4 or unlimited?
2. Size of viewport with tree
3. Size of viewport with no tree
4. Stacked button vs 2 column button and at what viewport?

Feedback welcome.

### Related issue(s)/PR(s)
This fixes https://github.com/modxcms/revolution/issues/13964

### DIFF
--- a/_build/templates/default/sass/_dashboard.scss
+++ b/_build/templates/default/sass/_dashboard.scss
@@ -249,7 +249,7 @@
   margin: -1rem 0 0 -1rem;
   .dashboard-button {
     display: flex;
-    align-items: center;
+    align-items: flex-start;
     background-color: $white;
     border-radius: $borderRadius;
     box-shadow: $boxShadow;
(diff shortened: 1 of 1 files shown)


jonleverrier:
There are 7 sizes... .quarter, .one-third, .half, .two-thirds, .three-quarters, .full and .double. 

Perhaps the first 3 sizes could be 1 across, the following 2 could be 2 across, and the final 2, 4 across ?

Mark-H (approved):
This was one of my nitpicks with the 3.x dashboard. Thanks for fixing it :)

alroniks (approved):

Issue #14084, opened 2018-09-11
Incorrect password behavior and message issue


minagerges:
### Summary
When failed login is attempted N number of times, the account gets blocked as it was blocked by an admin, and displayed message is "login_blocked_admin" instead of "login_blocked_too_many_attempts"

### Step to reproduce
1- Enter credential several times, and check received message is incorrect.
2- Notice which message received for block
3- Check DB table [modx_user_attributes] fields: [blocked] which should be set upon admin blocking a user and [failedlogincount] is 0 (just misleading)

### Observed behavior
How it behaved after following steps above.
- The logic used in "checkIsBlocked" function in "/var/www/dispatch_Portal/core/model/modx/processors/security/login.class.php" is unusual
- User attribute "failedlogincount" is set to 0
- Only indication to user and admin that the user was blocked due to several invalid credentials login attempts is a guess based on the value in "blockeduntil" and "blocked", which should mean that an admin blocked the user for a period of time.
- A successful login does not reset "failedlogincount" 

### Expected behavior
How it should behave after following steps above.
- Correct message to be displayed to user
- "failedlogincount" value should remain as an indicator of the block cause
- Block should be based on "failedlogincount" and "blockeduntil", while "blocked" should be reserved for admin blocking only
- the lexicon message "login_blocked_admin" should not indicate (manager) as same lexicon is returned for other contexts.
- After the specified block period passes, the counter should be reset allowing additional N number of failed logins, prior to blocking again for same period of time.
- A successful login should reset "failedlogincount"

### Environment
MODX version, apache/nginx and version, mysql version, browser, etc. Any relevant information.

Revolution 2.6.5-pl
Apache 2.4
MariaDB

**Proposed fix below**, provided while keeping in mind minimal changes, to avoid introducing any new issues in such critical class.
**N.B:** Code has been tested for all mentioned criteria.
1- Correct status represented in DB (failedlogincount + blockeduntil) only
2- Reset "failedlogincount" upon providing correct password
3- Block period end resets "failedlogincount"
4- Correct messages are returned based on block status and cause

_../core/lexicon/en/login.inc.php_
`$_lang['login_blocked_admin'] = 'You have been blocked by an administrator.';`

_../core/model/modx/processors/security/login.class.php_ 

```
     /**
     * Check if user is not active or blocked
     * @return bool|null|string
     */
    public function checkIsBlocked() {
        if (!$this->user->get('active')) {
            return $this->modx->lexicon('login_user_inactive');
        }

[...]

PR #14085, opened 2018-09-11
Fix incorrect password behavior and message issue #14084

minagerges:
### What does it do?
Change the checks and the order of checks in checkIsBlocked method.

### Why is it needed?
When failed login is attempted N number of times, the account gets blocked as it was blocked by an admin, and displayed message is "login_blocked_admin" instead of "login_blocked_too_many_attempts"

### Related issue(s)/PR(s)
#14084

### DIFF
--- a/core/model/modx/processors/security/login.class.php
+++ b/core/model/modx/processors/security/login.class.php
@@ -145,6 +145,7 @@ class modSecurityLoginProcessor extends modProcessor {
         if ($profile->get('blockeduntil') < time()) {
             if ($profile->get('blockeduntil') > 0) {
                 $profile->set('failedlogincount', 0);
+                $flc = 0;
             }
             $profile->set('blockeduntil', 0);
             $profile->save();
@@ -152,6 +153,7 @@ class modSecurityLoginProcessor extends modProcessor {
         if ($this->user->passwordMatches($this->givenPassword) && $profile->get('blockeduntil') < time()) {
             $profile->set('failedlogincount', 0);
             $profile->save();
+            $flc = 0;
         }
         else {
             $flc++;
--- a/core/model/modx/processors/security/login.class.php
+++ b/core/model/modx/processors/security/login.class.php
@@ -163,7 +163,7 @@ class modSecurityLoginProcessor extends modProcessor {
         }
         
         /* Validate block state */
-        if ($profile->get('failedlogincount') >= $this->modx->getOption('failed_login_attempts') &&
+        if ($flc >= $this->modx->getOption('failed_login_attempts') &&
             $profile->get('blockeduntil') > time()) {
             return $this->modx->lexicon('login_blocked_too_many_attempts');
         }
--- a/core/model/modx/processors/security/login.class.php
+++ b/core/model/modx/processors/security/login.class.php
@@ -148,7 +148,7 @@ class modSecurityLoginProcessor extends modProcessor {
             $profile->set('blockeduntil', 0);
             $profile->save();
         }
-        if ($this->user->passwordMatches($this->givenPassword)) {
+        if ($this->user->passwordMatches($this->givenPassword) && $profile->get('blockeduntil') < time()) {
             $profile->set('failedlogincount', 0);
             $profile->save();
         }
--- a/core/model/modx/processors/security/login.class.php
+++ b/core/model/modx/processors/security/login.class.php
@@ -140,22 +140,28 @@ class modSecurityLoginProcessor extends modProcessor {
         /** @var modUserProfile $profile */
         $profile = $this->user->Profile;
 
-        if ($profile->get('failed_logins') >= $this->modx->getOption('failed_login_attempts') &&
-            $profile->get('blockeduntil') > time()) {
-            return $this->modx->lexicon('login_blocked_too_many_attempts');
-        }
-
-        if ($profile->get('failedlogincount') >= $this->modx->getOption('failed_login_attempts')) {
-            $profile->set('failedlogincount', 0);
-            $profile->set('blocked', 1);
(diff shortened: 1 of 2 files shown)


jonleverrier:
I've tested the process aspect of this PR;

1. After 5 failed login attempts I get the following error message "You have been blocked due to too many failed login attempts." vs "You have been blocked by an administrator."
2. The user is blocked by the value in `blocked by` and the `blocked checkbox` remains unticked vs the value in `blocked by` and the `blocked checkbox` ticked

So functionality works as described.

core/model/modx/processors/security/login.class.php
```diff
@@ -140,23 +140,29 @@ public function checkIsBlocked() {
...
+        /* Update block state */
+        if ($profile->get('blockeduntil') < time()) {
+            if ($profile->get('blockeduntil') > 0) {
+                $profile->set('failedlogincount', 0);
+            }
+            $profile->set('blockeduntil', 0);
             $profile->save();
         }
-        if ($profile->get('blockeduntil') != 0 && $profile->get('blockeduntil') < time()) {
+        if ($this->user->passwordMatches($this->givenPassword)) {
```
Mark-H: This seems a bit odd to be added to a checkIsBlocked method - is it necessary to check the password here again?

minagerges:
The whole class is odd, and should br rewritten!
I explained everything in the issue, and the lines after the if statement
is clear, a successful password resets the counter, as per the common
behavior.
In the wrong place? Well yes, also explained that before!!!
Keeping changes to minimal, for a sensitive class.

I didn't have time to rewrite the whole class. May be in 2 weeks, after I'm
back from my vacation.

JoshuaLuckers:
A rewrite might be good for 3.x

sonicpunk:
So will this need to be rewritten or is it ok the way it is to merge?

Jako:
It should be ok for the moment, but it has to be checked.

sonicpunk:
@Jako I just installed it and tested the use-case and got the desired text output:
screen shot 2018-10-13 at 12 25 36

core/model/modx/processors/security/login.class.php
opengeek: More than odd; would this not destroy the ability to use external authentication via plugins?

minagerges: No, it will not, it only resets "failedlogincount" and has nothing to do with the evaluation of blocked state. 
A successful login should reset "failedlogincount". Using an external authentication plugin will still pass, yet "failedlogincount" will be reset time based on next login.

opengeek: With an external authentication plugin, the password will not be stored locally. So this does not seem like a valid solution. I will need to dig into and test this with a system using external authentication, but I can almost guarantee this will not work.

JoshuaLuckers: Wouldn't it make more sense to reset the `failedlogincount` in `afterLogin()` instead of doing so in `checkIsBlocked()`?

minagerges: @JoshuaLuckers, A successful login should reset `failedlogincount` **before** `checkIsBlocked()`, so it would bypass the `failedlogincount` block status _(as the user should not be blocked anymore based on the successful login)_.
@opengeek 's assumption is **only** valid if a user is already blocked because of `failedlogincount`, then successfully logs in via an external authentication plugin _(excluding Login extra)_. Only in such case, as I replied earlier, it will not reset the count and will be time based. Thus, it's a specific use case, based on the assumption of that (authentication plugin) incrementing  `failedlogincount` manually, as it's only incremented via the password validation (refer to `checkPassword()` logic).

Mark-H (dismissed):

core/model/modx/processors/security/login.class.php
JoshuaLuckers: I just ran some tests. If I try to login a few times with a wrong password I get the `login_blocked_too_many_attempts` message as expected.

If I login a few more times with a wrong password the message stays the same.
But when I login after `login_blocked_too_many_attempts` with the correct password I can't login because I'm blocked (as expected). 

When I try to login again with a wrong/correct password the message I receive is different (`login_blocked_error`). This happens because after the password matches `failedlogincount` is reset to 0.

An attacker can use this behaviour to retrieve a users password.

JoshuaLuckers (changes requested):

core/model/modx/processors/security/login.class.php
JoshuaLuckers: @minagerges do you have time to check my comment?

Ibochkarev:
@minagerges is it possible to help us getting this merged?

minagerges (commented):
**Regression**: Correct password during block period changes returned message. 
https://github.com/modxcms/revolution/pull/14085#discussion_r243739086

**Change**: Correct password resets `failedlogincount` only after `blockeduntil` is past

core/model/modx/processors/security/login.class.php
minagerges: @JoshuaLuckers please test based on PR commit https://github.com/modxcms/revolution/pull/14085/commits/3183308eb861149f43747c09367f7863c61a1f8e

minagerges: @JoshuaLuckers please test against https://github.com/modxcms/revolution/pull/14085/commits/6f07f88772767c126c0e07231edba9a8a2c4e8b3 as latest change.

Fixed blocked occurring on `failedlogincount` + 1

JoshuaLuckers (approved):

Mark-H (approved):

PR #14098, opened 2018-09-30
Fixed depth parameter in the modX::sanitize

AgelxNash:
### What does it do?
Using the max_input_nesting_level option for max depth argument

### Why is it needed?
Security reason

### DIFF
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -364,15 +364,20 @@ class modRequest {
      */
     public function sanitizeRequest() {
         $modxtags = array_values($this->modx->sanitizePatterns);
-        modX :: sanitize($_GET, $modxtags);
+
+        $depth = (int)ini_get('max_input_nesting_level');
+        $depth = ($depth <= 0) ? 99 : $depth + 1;
+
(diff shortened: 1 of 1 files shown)


ghost:
This value is based on PHP configuration settings?

I'm interested because I've had problems with hosting where this was set too low, and some third-party libraries like TinyMCE would fail. I was able to get the hosting to increase their depth setting from 4 to 6, and then all was well.

AgelxNash:
[The default value is 64](https://secure.php.net/manual/en/info.configuration.php#ini.max-input-nesting-level)
If the value is greater than 100, then this request bypasses filtering
```
data[][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][][]=[[snippet]]
```
More info: https://modx.pro/security/16484 :ru:

Jako:
I'll check this today #modxbughunt

Jako (approved):
Checked, works fine. Thanks!

PR #14099, opened 2018-09-30
Built in replace braces method for MODX tags sanitization

AgelxNash:
Security reason

### DIFF
--- a/_build/test/Tests/Cases/Modx/ReplaceBracesTest.php
+++ b/_build/test/Tests/Cases/Modx/ReplaceBracesTest.php
@@ -1,0 +1,110 @@
+<?php
+
+class ReplaceBracesTest extends \PHPUnit\Framework\TestCase
+{
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        include_once dirname(dirname(dirname(dirname(dirname(__DIR__))))) . '/core/model/modx/modx.class.php';
+    }
+
+    public function testEmptyString()
+    {
+        $this->assertEquals(
+            '',
+            modX::replaceBraces('')
+        );
+
+        $this->assertEquals(
+            '',
+            modX::replaceBraces(new stdClass)
+        );
+
+        $this->assertEquals(
+            '',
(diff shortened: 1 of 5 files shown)


Mark-H:
Could you please provide details of security issue to security@modx.com so we know what/how to properly test this?

AgelxNash:
It just a helper method. This method should be used by the developers of snippets.

MrRoco:
Tested & works fine

Jako:
This helper method makes the replacement recursive for array values. Thats nice.

But the naming of the method is somehow strange, since it does not only replace braces.

Mark-H:
I'm happy with adding this helper, but would appreciate @opengeek weighing in.

opengeek:
I agree about the method name, but this otherwise seems solid. Do we have any alternative ideas? I'd be happy to refactor it on merge to address that issue.

Jako:
Maybe it should be called `replaceArrayValues` or `recursiveReplaceArrayValues`

Mark-H (approved):
Prime example of bikeshedding, but perhaps `sanitizeTags` or `escapeMODXTags` would be more descriptive. Thanks @AgelxNash.

opengeek:
I'm sorry, but coming to agreement on a method name that is added to the main class of the software is not exactly bikeshedding. If it goes on for 10 messages, I would agree, but I was just asking for a few suggestions as I could not think of a good name at the time I reviewed this.

alroniks:
I agree with Jason, current name not the best option. A public method that should be widely used should have a more intuitive name.
My vote for a `escapeMODXTags` or more generic `escapeReservedSymbols`

Jako:
Since it is used without escaping MODX tags in https://github.com/modxcms/revolution/pull/14099/files#diff-b6cd1dec11656d0ae875f625ee8494ef the method could be simply named `escapeTags`

AgelxNash:
Just ```modx::replacements()```

opengeek:
This PR is superseded by #14156.

Issue #14107, opened 2018-10-10
Error Page Fallback


matdave:
### Summary
Not having a defined error page causes a site to collapse

### Step to reproduce
Set the error page to a non resource match value

### Observed behavior
Site times out on 404 of simple things, e.g. favicon.ico

### Expected behavior
If error page undefined or id doesn't exist should 404 and return site_start 

### Related Issues
#13659

JoshuaLuckers:
Thanks for reporting this issue.

I'm able to reproduce it with an error message:
> Fatal error: Uncaught Error: Maximum function nesting level of '256' reached, aborting! in /path/to/modx/core/xpdo/om/xpdoquery.class.php on line 843

There is an infinite redirect between `$modX->sendErrorPage( )` and `$modX->sendForward( )` when the error_page points to a non-existing resource.

jonleverrier:
I can reproduce this issue too. 

However, if I have missing assets like a favicon, the page still loads. It only times out if I load a page that doesn't exist (because the `error_page` doesn't exist).

Jako:
The fix was merged.

PR #14108, opened 2018-10-10
Prevent an infinite loop in sendErrorPage()

JoshuaLuckers:
### What does it do?
It makes it possible to skip the `sendErrorPage` if the resource does not exist.

### Why is it needed?
To prevent an infinite loop if the defined `error_page` in system settings is a non-existent resource.

### Related issue(s)/PR(s)
#14107

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1114,6 +1114,7 @@ class modX extends xPDO {
      *
      * @param integer $id The resource identifier.
      * @param string $options An array of options for the process.
+     * @param boolean $sendErrorPage Whether we should skip the sendErrorPage if the resource does not exist.
      */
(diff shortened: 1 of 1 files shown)


Jako (commented):
`elseif` instead of `else if`

jonleverrier:
I've tested this PR @JoshuaLuckers.

When the `error_page` is set to an ID that doesn't exist, it returns a 404 (page not found). It no longer produces an error 500 / timeout. 

Just to confirm, it doesn't redirect to `start_id`.

Jako:
I'll check this today #modxbughunt

Jako:
Checked, works fine. Thanks!

Jako (approved):

alroniks (approved):

Mark-H (approved):

Issue #14109, opened 2018-10-13
Setup password length error message wrong


labr1005:
### Summary
e6e1d504 changed the minimum password length during setup from 6 to 8. The error message still displays `Your password must be at least 6 characters long.`;

### Step to reproduce
Setup MODX. Choose a password shorter than 8 chars.

https://github.com/modxcms/revolution/blob/2.x/setup/lang/en/default.inc.php#L189

### Environment
MODX 2.7.0-dev

labr1005:
https://github.com/modxcms/revolution/blob/2.x/setup/lang/en/default.inc.php#L189

PR #14110, opened 2018-10-13
Fix for the password length error message during setup 

JoshuaLuckers:
### What does it do?
Changes the error message you get when the password is to short (when creating an account during setup)

### Why is it needed?
To fix issue #14109

### Related issue(s)/PR(s)
#14109


JoshuaLuckers:
Do I have to change the lexicons in other languages as well or is this done via Crowdin?

Jako:
That has be done on Crowdin.

setup/lang/en/default.inc.php
```diff
@@ -186,7 +186,7 @@
 $_lang['password_err_invchars'] = 'Your password may not contain any invalid characters, such as /, \\, &apos;, &quot;, (, ) or {}.';
 $_lang['password_err_nomatch'] = 'Does not match password';
 $_lang['password_err_ns'] = 'Password is empty';
-$_lang['password_err_short'] = 'Your password must be at least 6 characters long.';
+$_lang['password_err_short'] = 'Your password must be at least 8 characters long.';
```
Jako, line 189: Maybe the 8 could be replaced by a placeholder [[+length]] and set that value in the lexicon call.

Jako (changes requested):

Jako:
Closing this in favour of #14112

PR #14111, opened 2018-10-13
Fixing bug #14024

MrRoco:
### What does it do?
If you specify an invalid relative path for a Filesystem media source, it does not display the contents of the server root folder.

### Why is it needed?
If you specify an invalid relative path for a Filesystem media source, it displays the contents of the server root folder. #14024

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14024


Jako:
#14038 solves that somehow better. But thanks for the PR.

PR #14112, opened 2018-10-13
Fix for the password length error message during setup

Jako:
### What does it do?
Changes the error message you get when the password is to short (when creating an account during setup)

### Why is it needed?
To fix issue #14109 (improved version of #14110)

### Related issue(s)/PR(s)
#14109, #14110


MrRoco:
Tested & confirmed to be working :)

Issue #13267, opened 2017-01-27
ExtJS: modx-combo-user doesn't render username but user ID instead when user is preselected


gadgetto:
### Summary
Using an extjs modx-combo-user and trying to preselect a user when combo is loaded doesn't render the username but it's ID instead.

### Step to reproduce
```
{
    xtype: 'modx-combo-user'
    ,fieldLabel: _('resource_createdby')
    ,description: '<b>[[*createdby]]</b><br />'+_('resource_createdby_help')
    ,name: 'created_by'
    ,hiddenName: 'createdby'
    ,id: 'modx-resource-createdby'
    ,anchor: '100%'
    ,value: config.record.createdby || MODx.user.id
}
```

### Observed behavior
Doesn't render the username of preselected user but it's ID instead.

### Expected behavior
Should render username of the preselected user.

### Environment
MODX version: 2.5.4-pl

*Can be fixed by adding the following lines to:*

https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/security/user/getlist.class.php#L35

```
$id = $this->getProperty('id');
if (!empty($id)) {
    $c->where(array('id' => $id));
}
```

Jako:
It looks like all modx-combo variants should have this issue.

gadgetto:
The template-combo works as expected.

Jako:
Yes, but it fetches the whole templates with limit:0 for some reason. Seems that the store is filtering the pagination.

sdrenth:
@gadgetto Is this issue fixed already? I can't seem to reproduce this issue in branch 2.x.

gadgetto:
You need a big user list > 20 users to reproduce this. I still can reproduce this.

sdrenth:
@gadgetto Ahh I see, the problem occurs when you have a listbox which has pagination. If you have a value preselected which is being retrieved on page 1 of the listbox options all is fine. However if you have a preselected value which is not being retrieved yet (so its on page 2 or further on in the listbox) the name is not yet retrieved so it is displaying the value instead.

I'm thinking, maybe a render value method should run, which performs an ajax request to retrieve the name for the selected user?

Jako:
It could be fixed with the code from @gadgetto at the start of this thread.

alroniks:
Solved. PR is merged.

PR #14113, opened 2018-10-13
Fixing user combo with preselected user ids #13267

Jako:
### What does it do?
Adding a where clause to the query, if an id property is requested. 

### Why is it needed?
Fixing showing an id instead of an username, if the selected user id is not returned in the list of users using the default limit.

It seems that most of the getlist processors have the same issue. But this will be fixed in a different PR.

### Related issue(s)/PR(s)
#13267


Mark-H (approved):
Yes please.

Mark-H:
On second thought - shouldn't the where clause (and the ones in #14116) be in the `prepareQueryBeforeCount` method instead of `prepareQueryAfterCount`?

Jako:
I have seen that solution first in another processor on `prepareQueryAfterCount`. It would make no difference, since it filters just the shown ids. This way, the total count contains all records. But the count is not used.

Mark-H:
Fair enough. 👍

alroniks (approved):

Jako:
Just for the records: the valuefield query needs to be in the prepareQueryAfterCount, otherwise the combo won't get a pagination.

PR #14116, opened 2018-10-14
Fixing getlist processors for combos that have a pagesize 

Jako:
### What does it do?
Add a filtering query (in most cases by id) to the getlist processors of combos that have a pagesize set.

### Why is it needed?
Those combos show the following issue. If the current value of the combo is not returned in the valuefield value with the list of records limited by the pagesize, the combo displays the valuefield instead of the displayfield. I.e. the site has more than 20 contexts, and a modx-combo-context value is not in the first 20 contexts, the combo shows only the context key and not the context name.

### Related issue(s)/PR(s)
#14113

### DIFF
--- a/core/model/modx/processors/context/getlist.class.php
+++ b/core/model/modx/processors/context/getlist.class.php
@@ -69,9 +69,9 @@ class modContextGetListProcessor extends modObjectGetListProcessor {
         }
         return $c;
     }
-    
+
     /**
-     * {@inheritDoc}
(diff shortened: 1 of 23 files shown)


JoshuaLuckers (changes requested):
It would be nice to have PHPDoc comments for the "new" methods.

Mark-H (approved):

alroniks (approved):

Issue #14101, opened 2018-10-01
Dashboard Widget Double Size


pyrographics:
### Summary
The Double Dashboard Widget Size is No Different from the Full Size even though the documentation says it should be twice the rows of a full widget.

### Step to reproduce
Create simple HTML dashboard widget with Double selected for size and placeholder content.  Create another HTML dashboard widget with Full selected for size and placeholder content.
Add both widgets to the top of your dashboard.
View dashboard

### Observed behavior
Minuscule difference in row height.

![screen shot 2018-10-01 at 1 44 37 pm](https://user-images.githubusercontent.com/274572/46308946-96208700-c580-11e8-8310-1cb39c5eeab4.png)

### Expected behavior
Double size should be double row height.

### Environment
MODX 2.6.5, MODX Cloud, PHP 7.2

pyrographics:
The manager template css file index.css has conflicting code in it.  The .dashboard-block-double class has a minimum height of 400px but the .dashboard-block .body class has a maximum height of 300px which applies to all blocks including the double block.  Commenting out the 300px max height allows the double block to be larger.

<img width="350" alt="screen shot 2018-10-05 at 4 48 52 pm" src="https://user-images.githubusercontent.com/274572/46561875-2e8b7400-c8bf-11e8-8ef8-b2322768ee28.png">

pyrographics:
I think a more appropriate way to deal with it would be to add this to the CSS:

`.dashboard-block-double .body {
   max-height: 600px;
}`

PR #14117, opened 2018-10-15
Improved display of "double" height dashboard widget 

jonleverrier:
### What does it do?
Dashboard widgets come in 3 heights; half, full and double. When double was selected, it was not actually double in size. For more info see https://github.com/modxcms/revolution/issues/14101.

On `.dashboard-block-double`, the min-height was increased to 600px so it looks double in size. A `.body` selector was also added to `.dashboard-block-double` so that it overrides the default `.dashboard-block` max-height.

### Why is it needed?
Better UX.

### Related issue(s)/PR(s)
Fixes https://github.com/modxcms/revolution/issues/14101

### DIFF
--- a/_build/templates/default/sass/_dashboard.scss
+++ b/_build/templates/default/sass/_dashboard.scss
@@ -75,8 +75,12 @@
 .dashboard-block-double {
   clear: both;
   height: auto;
-  min-height: 400px;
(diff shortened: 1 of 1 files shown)


Jako:
Works fine, thanks!

PR #14118, opened 2018-10-16
Wrong translation for user State field

ant-tar:
### What does it do?
Better translation variant for the field.

### Why is it needed?
In Russian connotation it will be better to sound like "Регион / область"
P.S. if you let me know how to add this to CrowdIn I'll be happy to add it, I've added there many times ago..but can't see how exactly this is possible. Thanks!

### Related issue(s)/PR(s)
No such ones as for me.


Jako:
Could you please change/add this translation on Crowdin 
https://crowdin.com/project/modx-revolution

You have to login on crowdin and follow this link https://crowdin.com/translate/modx-revolution/449/en-ru to edit the user.inc.php strings of Revolution 2.5

alroniks:
Please, add translations via Crowdin, as @Jako said.

PR #14119, opened 2018-10-17
Make path to MODX error handler class configurable

labr1005:
### What does it do?
This introduces a new config setting "error_handler_path" to make the path to a custom MODX error handler configurable.

### Why is it needed?
When using custom MODX error handlers they currently have to be placed in `core/model/modx/*` since a path is not configurable.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -2426,9 +2426,9 @@ class modX extends xPDO {
         if ($this->errorHandler == null || !is_object($this->errorHandler)) {
             if ($ehClass = $this->getOption('error_handler_class', $options, 'modErrorHandler', true)) {
                 $ehPath = $this->getOption('error_handler_path', $options, '', true);
-                if ($ehClass= $this->loadClass($ehClass, $ehPath, false, true)) {
-                    if ($this->errorHandler= new $ehClass($this)) {
-                        $result= set_error_handler(array ($this->errorHandler, 'handleError'), $this->getOption('error_handler_types', $options, error_reporting(), true));
+                if ($ehClass = $this->loadClass($ehClass, $ehPath, false, true)) {
+                    if ($this->errorHandler = new $ehClass($this)) {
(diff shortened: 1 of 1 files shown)


core/model/modx/modx.class.php
```diff
@@ -2425,7 +2425,8 @@ protected function _initCulture($options = null) {
     protected function _initErrorHandler($options = null) {
         if ($this->errorHandler == null || !is_object($this->errorHandler)) {
             if ($ehClass = $this->getOption('error_handler_class', $options, 'modErrorHandler', true)) {
-                if ($ehClass= $this->loadClass($ehClass, '', false, true)) {
+                $ehPath = $this->getOption('error_handler_path', $options, '', true);
+                if ($ehClass= $this->loadClass($ehClass, $ehPath, false, true)) {
```
JoshuaLuckers:
```suggestion
                if ($ehClass = $this->loadClass($ehClass, $ehPath, false, true)) {
```

christianseel:
I tested this PR locally with MODX 2.6.5 successfully. Didn't found any issues with default `modErrorHandler` and also with a custom error handler it works fine.

Used this as a custom handler:
```
<?php
require_once MODX_CORE_PATH.'model/modx/error/moderrorhandler.class.php';

class sedadigitalErrorHandler extends modErrorHandler {
    function __construct(modX &$modx, array $stack = array()) {
        $modx->setLogTarget(array(
            'target' => 'FILE',
            'options' => [
                'filename' => 'modx.log',
                'filepath' => MODX_CORE_PATH.'logs/'
            ]
        ));
        parent::__construct($modx, $stack);
    }
}
```

JoshuaLuckers (approved):

Jako (approved):

christianseel:
Note: This approach is failing when using `$modx->switchContext()` as it's used by most routing/context plugins (like XRouting). The MODX core function switchContext is calling _initContext which is overriding the logTarget from the system/context-settings.

See https://github.com/modxcms/revolution/pull/14195 for a suggested solution.

Issue #13879, opened 2018-04-26
When updating multiple extras, setup options window title shows the previously updated extra


muzzwood:
### Summary
While updating extras, I noticed if two extras with setup options were updated consecutively, the second extra's setup options window has the title of the previous extra's setup options window. The actual contents of the setup options window was correct though.
I noticed that if I refreshed the page between updating extras, this did not occur.

### Step to reproduce
Try updating multiple extras with a setup options window consecutively.

### Observed behavior
The setup options window did not have the correct title. It was the title of the previously updated extra.

### Expected behavior
The setup options window for each extra should have its own title.

### Environment
MODX 2.6.3 PHP 7.1 Apache MySQL

ghost:
I thought that was fixed several releases ago. https://github.com/modxcms/revolution/pull/13557

Jako:
It is just the title of that window that does not change.

`this.win.title` has to be modified in the else branch there, if I am right.

muzzwood:
Yeah this is just the title. That older bug would try to reinstall the previous extra.

Jako:
This is some special issue caused by some javascript, that runs i.e. in STERC extras. The title is not changing by default. But it can be fixed.

alroniks:
Closed as related PR is merged.

PR #14120, opened 2018-10-18
Restore the setup options window title, if two packages are installed…

Jako:
### What does it do?
Restore the setup options window title, if two packages are installed with setup options.

### Why is it needed?
Some extras change the setup options title by javascript. It will not be changed back if the window is opened again.

### Related issue(s)/PR(s)
#13879

### DIFF
--- a/manager/assets/modext/workspace/package.containers.js
+++ b/manager/assets/modext/workspace/package.containers.js
@@ -159,6 +159,7 @@ Ext.extend(MODx.panel.Packages,MODx.Panel,{
 			});
 		} else {
 			this.win.signature = btn.signature || '';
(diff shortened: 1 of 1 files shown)


Jako:
This happens after installing i.e. SEOPro and another extra afterwards caused by these lines: https://github.com/Sterc/SEOPro/blob/4d2cc38c0367b3523c24fa3ffd2c2cb972530a16/_build/setup.options.php#L46-L47

Mark-H (approved):

alroniks (approved):

Jako:
This should be reverted, since the title text can't be changed anymore.

PR #14124, opened 2018-10-29
Fixed the name of system setting about allowing or not eval in TV

alroniks:
### What does it do?
It fixes the naming of system setting to be true.

### Why is it needed?
Currently, the name says that this setting will disable the ability to run eval if a value is true, but it is not.
In description placed correct variant - if a value is no, no any code will be handled, the value will be return as a plain text.
So correct variant should be "Enable eval in TV bindings". 

### Related issue(s)/PR(s)
IDK

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -95,8 +95,8 @@ $_lang['setting_allow_manager_login_forgot_password_desc'] = 'Setting this to "N
 $_lang['setting_allow_tags_in_post'] = 'Allow Tags in POST';
 $_lang['setting_allow_tags_in_post_desc'] = 'If false, all POST variables will be stripped of HTML script tags, numeric entities, and MODX tags. MODX recommends to leave this set to false for Contexts other than mgr, where it is set to true by default.';
 
-$_lang['setting_allow_tv_eval'] = 'Disable eval in TV binding';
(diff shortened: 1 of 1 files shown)


SnowCreative:
I'm not following. Current setting is called "allow_tv_eval" and description says "Select this option to enable or disable eval in TV binding. If this option is set to no, the code/value will just be handled as regular text." What's wrong with any of that? It says that "Yes" means that eval is allowed.

opengeek:
@rainbowtiger The lexicon entry for `allow_tv_eval` was "Disable eval in TV binding". This commit changes it to say "Enable eval in TV bindings" to properly clarify the intent of the setting.

opengeek (approved):

alroniks:
@rainbowtiger Description is OK, yes, but the title is not.
Now it is "Disable eval in TV binding" but should be "Enable eval in TV binding".

In examples:
Disable eval in TV binding? -> no (but actually eval disabled, because allow_tv_eval == false) wtf?
Disable eval in TV binding? -> yes (but actually eval enabled, because  allow_tv_eval == true) double wtf?

Jako (approved):

SnowCreative:
> @rainbowtiger The lexicon entry for `allow_tv_eval` was "Disable eval in TV binding". This commit changes it to say "Enable eval in TV bindings" to properly clarify the intent of the setting.

Ah, I see where that is now. Makes sense!

PR #14127, opened 2018-11-03
Add $modx->deprecated method to log usage of deprecated methods

Mark-H:
### What does it do?
This PR adds a new method to the modX object, `deprecated`, which is to be used to log usage of deprecated methods across the core. Each unique message (deprecated method + caller) is only logged once per request. A setting is added that allows disabling the logging. 

I've also done a search for `@deprecated` docblocks, and added appropriate messages in various locations. Some code was deprecated way back in 2.1, so I look forward to seeing that removed in 3.0. 

### Why is it needed?
As we get ready to cleanup some code with 3.0, it's important users can be sure their site can be safely upgraded. Old or custom extras may use functionality that is no longer supported and will be removed in 3.0, which will require manual intervention. 

The `deprecated` method will allow a site to function normally with the deprecated methods still available in 2.7, but thanks to the logging the site owner will have more information on what should be taken care of first. 

### Related issue(s)/PR(s)
#13199, which mentions some deprecated methods for 3.0.

### DIFF
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -526,6 +526,12 @@ class modRequest {
      * @return array
      */
     public function getAllActionIDs($namespace = '') {
+        // This method is deprecated, but also still in use by the core in
+        // core/model/modx/processors/system/config.js.php, which runs on each
+        // manager page, so leaving the deprecated notice commented out until
+        // that is resolved.
+        // $this->modx->deprecated('2.3.0', 'Support for modAction has been replaced with routing based on a namespace and action name.', 'modAction support');
+
         $c = array();
         if (!empty($namespace)) $c['namespace'] = $namespace;
         $actions = $this->modx->getCollection('modAction',$c);
--- a/core/model/modx/modrequest.class.php
+++ b/core/model/modx/modrequest.class.php
@@ -526,12 +526,6 @@ class modRequest {
      * @return array
      */
     public function getAllActionIDs($namespace = '') {
-        // This method is deprecated, but also still in use by the core in
-        // core/model/modx/processors/system/config.js.php, which runs on each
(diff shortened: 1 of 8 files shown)


core/model/modx/modrequest.class.php
```diff
@@ -526,6 +526,12 @@ public function checkPublishStatus() {
      * @return array
      */
     public function getAllActionIDs($namespace = '') {
+        // This method is deprecated, but also still in use by the core in
+        // core/model/modx/processors/system/config.js.php, which runs on each
+        // manager page, so leaving the deprecated notice commented out until
+        // that is resolved.
+        // $this->modx->deprecated('2.3.0', 'Support for modAction has been replaced with routing based on a namespace and action name.', 'modAction support');
```
Jako: Should that be executed?

Jako (dismissed):
I don't have a clue, why this line is commented out.

core/model/modx/modrequest.class.php
Mark-H: That's literally explained in the lines above it. 

It's a deprecated method (per the docblock), that should not be used, but every single manager request currently triggers it.

JoshuaLuckers: Add a @todo to the comment so it’s easier to detect.

Jako: Sorry. I only have seen the commented out line and forgot to read the comment.

Mark-H: I've removed all of the comments. The method is already tagged with `@deprecated`. When `modAction` is removed from 3.x this method should also be removed.

opengeek (approved):
This may not be the ideal solution, but it will work for now and provides an abstraction for better solutions moving forward.

PR #14128, opened 2018-11-06
Improved messages on plugin editor page

### LINKED ISSUE #12900: Function/feature terminology mismatch on Plugins
Plugins - using the shortcut menu (right-click) the user has the option of Activating or Deactivating the plugin whereas on the Create/edit plugin page the user has the option to check Plugin Disabled or not. Need to standardize terminology if they're the same function.

alroniks:
### What does it do?
It changes messages about disabling/enabling plugins and also fixed bug with proper determining the value of `inline_help` system setting. Additionally, it improves descriptions of checkboxes on the page.

### Why is it needed?
It is standardizing terminology a bit (at least on the one page) and also fixes some bugs that improves UX.

### Related issue(s)/PR(s)
Fixes #12900


core/lexicon/en/plugin.inc.php
```diff
@@ -37,7 +38,7 @@
 $_lang['plugin_event_msg'] = 'Select the events that you would like this plugin to listen to.';
 $_lang['plugin_event_plugin_remove_confirm'] = 'Are you sure you want to remove this plugin from this event?';
 $_lang['plugin_event_update'] = 'Update Plugin Event';
-$_lang['plugin_lock'] = 'Lock plugin for editing';
+$_lang['plugin_lock'] = 'Locked for editing plugin';
```
philipwhiuk: 'Plugin locked for editing' would read better. This is grammatically unclear.

The meaning has changed from an action to a state.

alroniks: Thank you for notice. And yes, checkbox should describe state instead an action. It's why I changed it.

Jako (approved):

alroniks:
@Mark-H Your time ;)

Mark-H (approved):

Mark-H:
Looks good :)

PR #14129, opened 2018-11-07
Fix and normalize combo-boolean tv option values

Jako:
### What does it do?
Fix and normalize combo-boolean values in template variable input options. Fix a bug in the number template variable type in a non english manager. 

### Why is it needed?
If a number template variable was created by i.e. a package, the allowDecimals and allowNegative option shows a 'true' or 'false' as the combo box value, instead of a 'Yes' or 'No'. In other languages than english the saved template variable options of a number template variable were saved with the language string because of a missing hiddenName option.

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/templates/default/element/tv/renders/inputproperties/autotag.tpl
+++ b/manager/templates/default/element/tv/renders/inputproperties/autotag.tpl
@@ -23,8 +23,8 @@ MODx.load({
         ,name: 'inopt_allowBlank'
         ,hiddenName: 'inopt_allowBlank'
         ,id: 'inopt_allowBlank{/literal}{$tv|default}{literal}'
-        ,value: params['allowBlank'] == 0 || params['allowBlank'] == 'false' ? false : true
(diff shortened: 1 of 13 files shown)


opengeek:
Seems fine to me. I'm assuming this has no BC implications @Jako ?

opengeek (dismissed):

Jako:
There should be no BC with this change, since it saved 'true'/'false' before. 

The only BC is a real bug because of a missing hiddenName in allowNegative and allowDecimals. Because of this bug, the options could only be enabled in an english manager. A german 'Nein' is not equal to 'No' and the tv options are set on the resource edit page with this code: https://github.com/modxcms/revolution/blob/2.x/manager/templates/default/element/tv/renders/input/number.tpl#L20-L21

alroniks (approved):
Looks good for me

Mark-H (approved):

PR #14133, opened 2018-11-08
Remove outdated classes for converting evo tags to revo

alroniks:
### What does it do?
It removes outdated and deprecated in 2.7 classes `modParser095`, `modTranslate095` and `modTranslator`.

### Why is it needed?
It used for migration Evo sites to Revo, but as Evo does not on a support any more officially and it is an independent project, it does not make sense to keep this functionality in the third version.

### Related issue(s)/PR(s)
#14131


Mark-H (approved):

JoshuaLuckers (changes requested):
Is it possible to solve the "merge conflicts"? :)

alroniks:
@JoshuaLuckers This files should be deleted at all so absolutely no make sense to resolve conflicts. It can be done during the merging process. I can handle this if you will approve it.

JoshuaLuckers (approved):

PR #14134, opened 2018-11-08
Example of PSR autoloading of classes instead of $modx->loadClass()

alroniks:
### What does it do?
It fixes a small part of the core (hashing) to use PSR standards in the formatting and also for loading classes.

### Why is it needed?
As MODX tries to use a modern approach core should follow PSR standards and use most useful abilities that already implemented in other tools. Composer autoloading works very well and it can replace own loadClass method and it will allow use namespaces in classes and not only.

Note: It's a just example of how whole core can be refactored.

### Related issue(s)/PR(s)
N/A


opengeek (commented):
This is not PSR autoloading, this is composer classmap generation.

alroniks:
Yes, of course. It's a composer generation.  Sometimes difficult to explain thoughts in my head. :)

OptimusCrime:
Would it make more sense to wait until 2.7 is finished, and the 2.7 branch is merged into 3.0, and then start moving classes out of /core/model/modx into a suitable composer autoloaded directory? Like /core/src or something? That way, this rollout could happen gradually, and it would use autoloading to its full potential.

opengeek:
@OptimusCrime yes, that is the plan… I have some ideas on how to compromise on some of the previous objections to the existing PR that I'm going to try out as soon as 2.7 is out the door.

alroniks:
@OptimusCrime Of course. This PR just an example and not only about the class map but mostly about that it is possible to do huge changes by small pieces like this. Against that big pull request about refactoring where broken everything and readable nothing.

alroniks:
@opengeek @Mark-H Could we move it forward? Do exist any objections?

Mark-H:
I think Jason is planning some more work on autoloading and some under-the-hood refactoring soon. 

Would be nice to at some point do a massive code style cleanup, the PSR code style can probably be automatically done with `phpcs`/`phpcbf` to the entire code base at once. (Obviously, that breaks mergeability with older branches.. fun)

opengeek:
I don't think we need to leave this open. If you feel differently, please reopen and we can discuss specific points.

PR #14136, opened 2018-11-12
Add some deprecated logging

Jako:
### What does it do?
Add $modx/$xpdo->deprecated calls.

### Why is it needed?
Log deprecated methods/classes

### Related issue(s)/PR(s)
#13199

### DIFF
--- a/core/model/modx/processors/system/clearcache.class.php
+++ b/core/model/modx/processors/system/clearcache.class.php
@@ -143,6 +143,7 @@ class modSystemClearCacheProcessor extends modProcessor {
     public function clearByPaths() {
         $pathResults = array();
         /* deprecated: use a dedicated cache partition rather than specifying paths */
+        $this->modx->deprecated('2.1.4', 'Use a dedicated cache partition rather than specifying paths.', 'modSystemClearCacheProcessor clearByPaths');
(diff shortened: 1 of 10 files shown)


core/model/modx/processors/system/clearcache.class.php
```diff
@@ -143,6 +143,7 @@ public function getPartitions() {
     public function clearByPaths() {
         $pathResults = array();
         /* deprecated: use a dedicated cache partition rather than specifying paths */
+        $this->modx->deprecated('2.1.4', 'Use a dedicated cache partition rather than specifying paths.', 'modSystemClearCacheProcessor clearByPaths');
```
Mark-H: This method is called on each request, better to place this inside the `if (!empty($paths))` below so it only logs when the processor is called with paths to clear.

Mark-H (changes requested):

core/model/modx/processors/system/clearcache.class.php
Jako: This is fixed.

Mark-H (approved):
Thanks :)

alroniks (approved):

PR #14137, opened 2018-11-13
Skip target validation when MODX tags detected in the content field

alroniks:
### What does it do?
It skips the target validation when MODX tags detected in the content field.

### Why is it needed?
Original PR broke BC and should be fixed.

### Related issue(s)/PR(s)
#14041

### DIFF
--- a/core/model/modx/processors/resource/create.class.php
+++ b/core/model/modx/processors/resource/create.class.php
@@ -648,7 +648,10 @@ class modResourceCreateProcessor extends modObjectCreateProcessor {
     public function checkSymLinkTarget() {
         $target = $this->getProperty('content', null);
 
-        if ($target === null || $target === '') {
+        $matches = array();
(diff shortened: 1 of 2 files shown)


opengeek:
No, you can't pre-process it. It has to be processed at runtime, as I said.

opengeek (changes requested):
You cannot pre-process the target before validation. The values may not be available in the mgr context and could rely on runtime information.

opengeek:
I have some errands this afternoon, but I will try to get this solved when I get back. We need to simply skip validation if the target contains MODX tags.

alroniks:
@opengeek Rebased the whole PR. Now it detects tags and skips validation.

Jako:
A weblink could redirect to foreign hosts, so checking for an integer would be enough in that part.

But the current solution is suitable.

Jako (approved):

Issue #14138, opened 2018-11-14
Trash: right click "Purge Resource"


jonleverrier:
### Summary
When you click on the trash can, and right click over a resource to "purge resource", you get the following error message "Resource not specified."

### Step to reproduce
1. Create some test resources
2. Delete those test resources
3. Click on the trash can icon
4. Right click on a particular resource and select "Purge resource"
5. Click "yes" button to confirm

### Observed behavior
Error is reported "Resource not specified."

![bug1](https://user-images.githubusercontent.com/2373940/48489044-6670ca00-e822-11e8-8b05-1d8979c38c2b.gif)

### Expected behavior
No error

### Environment
MODX 2.7.0-dev built from github

alroniks:
Closing as related PR is merged.

PR #14141, opened 2018-11-16
Fixing single purge resource

Jako:
### What does it do?
Set the ids property instead of the id property for the resource/trash/purge processor.

### Why is it needed?
The wrong property was submitted

### Related issue(s)/PR(s)
#14138

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.grid.trash.js
+++ b/manager/assets/modext/widgets/resource/modx.grid.trash.js
@@ -200,7 +200,7 @@ Ext.extend(MODx.grid.Trash, MODx.grid.Grid, {
             url: this.config.url,
             params: {
                 action: 'resource/trash/purge',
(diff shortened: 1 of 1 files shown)


alroniks (approved):
Good catch! It was one of those things that I wanted to refactor. Now I remember :)

jonleverrier:
Tested and works fine

Mark-H (approved):
👍

JoshuaLuckers:
I was wondering if the property needs to be changed here as well: https://github.com/modxcms/revolution/blob/83827c4db49754cc1e9b0ef59e4f1b3a53426da7/manager/assets/modext/widgets/resource/modx.grid.trash.js#L234-L236

Jako:
It is not necessary there, since the resource/undelete processor is used there.

PR #14148, opened 2018-11-22
Fix prefix for lexicon key of log_snippet_not_found system setting

alroniks:
### What does it do?
It adds prefix `setting_` to keys in lexicons files to get translations to work for system setting with key `log_snippet_not_found`.

### Why is it needed?
Author of an original PR added lexicon key without prefix, that required for strings that will be used for describing system settings.

### Related issue(s)/PR(s)
#12984


Jako (approved):

Mark-H (approved):

Jako:
Next time we just have to change the english key. Everything else seems to be managed by Crowdin.

Issue #14048, opened 2018-08-16
Force composer to require PHP extensions that needed for proper work of MODX


alroniks:
### Summary
I guess subject explain the issue enough. Five cents additionally. We already have documented requirements to environments for running MODX but also It would be cool to have these requirements in the comoser.json file.

### Observed behavior
Now, you can try to download MODX and setup it but if some extensions not installed it can cause issues during the using the system later. For example, problems with image thumbs when `php-gd` wasn't installed before.

### Expected behavior
As we have an awesome tool - Composer, it should notify the user that the environment is not ready for proper work of the software that he or she tries to use.

### Environment
all

Jako:
That leads to the following question: Will the installable package of MODX3 work without using composer? I hope so, since a lot of shared host installations can't run composer.

alroniks:
It will work, of course, but if somebody will use composer-way of installation, it will show the found problems.

OptimusCrime:
@Jako all hosts can run composer, but it is a lot more difficult for some (those without cli for example).

However, I was under the impression that MODX itself would ship without the composer.json file? Who would this work for then? People that clone the repository?

alroniks:
Any automation tools that can setup MODX via composer, people who like composer-way of installation, and I hope all other people, who have more modern environments. Of course, it won't be must-have tool for all users, but... Who knows.

JoshuaLuckers:
People who are going to use `composer create-project modx/revolution`

Jako:
Could those composer php extension requirements somehow be used to automatically generate an installation check for an 'composer-less' installation?

alroniks:
@Jako Potentially yes, there is simple JSON file and can be parsed by any part of a code.

PR #14150, opened 2018-11-23
Add required and suggested PHP extensions to composer.json

JoshuaLuckers:
### What does it do?
This adds required and suggested PHP extensions to composer.json.

### Why is it needed?
We want to check and suggest the system requirements when installing MODX via composer.

### Related issue(s)/PR(s)
Fixes issue #14048


Mark-H (dismissed):

alroniks (dismissed):

opengeek (changes requested):
I don't think ext-imagick is required for MODX is it? Is this a new requirement? Because I've not installed this extension in quite some time and am now blocked from installing. :P

alroniks:
Jason is right. Imagick can be replaced (and it does) by php-gd. It works better but it is not installed everywhere so we can not lock MODX to use this extension. I suggest moving it to recommended section @JoshuaLuckers

JoshuaLuckers:
It’s listed as one of the [server requirements in the docs](
https://docs.modx.com/revolution/2.x/getting-started/server-requirement).

I will do the suggested changes later today. 👍

alroniks:
But php-gd is really required for thumbnails and imagick is missed in recommended section.

JoshuaLuckers:
> But php-gd is really required for thumbnails and imagick is missed in recommended section.

My misunderstanding!

alroniks:
Not sure, but maybe it does not make sense to install imagic in travis.

alroniks (approved):

alroniks:
Now it looks better. @opengeek

Mark-H (approved):
Not sure if this can be merged with a `changes requested` review from the other code owner.. but LGTM.

alroniks:
@Mark-H We have to wait for approval from @opengeek in this case.

JoshuaLuckers:
@opengeek I made the suggested changes.

opengeek (approved):

PR #14154, opened 2018-11-26
Fix Call to undefined function URLschemeDefaultPort()

smanu85:
### What does it do?
Add two missing self:: to fix a Fatal error: Uncaught Error: Call to undefined function URLschemeDefaultPort() in \core\model\phpthumb\phpthumb.functions.php:748

### Why is it needed?
phpthumb connector is unable to generate thumbnails in the manager (media browser and/or template vars preview) due to a fatal error.


Jako:
That file is part of [phpThumb](https://github.com/JamesHeinrich/phpThumb). Could you please create the PR there. If it is merged, please tell us to update our dependency from there.

smanu85:
@Jako Just saw that this has been already fixed in the original repository starting with version 1.7.15-201807091031
Fixed in: JamesHeinrich/phpThumb@1fc28a6acf1ffa5ad60a57ee1304fd580892c990 

MODX 2.7.0-dev phpthumb is an older version: 1.7.15-201806071234.

PR #14156, opened 2018-11-26
Add recursive method to replace reserved characters

opengeek:
### What does it do?
Adds a static method to the modX class to allow recursive replacement of reserved characters.

### Why is it needed?
Security implications.

### Related issue(s)/PR(s)
This is a #14099 with the method renamed from modX::replaceBraces() to modX::replaceReserved().


alroniks (approved):

JoshuaLuckers (changes requested):
Please add the copyright header. I also see some inconsistencies in calling the function:
`modX :: replaceReservered` and `modX::replaceReservered`

PR #14157, opened 2018-11-27
split head and body tags to confuse bad people

goldsky:
### What does it do?
It just splits the `</head>` and `</body>` end tags.

### Why is it needed?
Some hackers grep the `</head>` tag to inject new code in the core.

### Related issue(s)/PR(s)
We had some core files modified as such.
The code was javascript, like: `<script> var _0xsdfa = ... </script></head>`

### DIFF
--- a/core/model/modx/modresponse.class.php
+++ b/core/model/modx/modresponse.class.php
@@ -64,15 +64,15 @@ class modResponse {
             /*FIXME: only do this for HTML content ?*/
             if (strpos($this->contentType->get('mime_type'), 'text/html') !== false) {
                 $this->modx->invokeEvent('OnBeforeRegisterClientScripts');
-                /* Insert Startup jscripts & CSS scripts into template - template must have a </head> tag */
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Thanks for taking time to make MODX better by submitting this PR.

I don't think this is the best solution for this problem. It's better to find the cause of the hack and solve that instead.

goldsky:
In my case, MODX is installed in subdomain, alongside with Wordpress in
main domain.

What could we do if the hacking vector is not our own?

Mark-H:
Can you try isolating the sites on the server so a WP issue cannot result in MODX being compromised? 

I'm with Joshua.

goldsky:
Unfortunately, we are not the main site's developer.

While my solution might not be the best answer, but this issue will keep
making MODX vulnerable as a collateral damage.

JoshuaLuckers:
We appreciate the input! Sadly this is only a temporary solution and starts a cat/mouse game with the attacker. Once the attacker is able to modify files it’s easy to inject malicious code elsewhere (your site might still be infected). Or he/she simply adds this pattern to his toolkit.

PR #14161, opened 2018-11-28
Update config.js.php processor to be class based

Mark-H:
### What does it do?
In #14160 I made the deprecated message a bit more useful for identifying offending processors, which immediately pointed to the config.js.php which is called on each manager request as loading the config into a JS object. So, I wrapped it into a modProcessor class to resolve that, and added a migration to remove the old processor file when running the 2.7.1 setup. 

### Why is it needed?
The deprecated log message on each manager request will not only get annoying fast, but as it's not very useful until #14160, this particular instance may encourage people to disable the deprecated logging preventing them from seeing other deprecated usages that need to be fixed. 

### Related issue(s)/PR(s)
#14160


setup/includes/upgrades/common/2.7.1-cleanup-script.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * Common upgrade script for 2.5 to clean up legacy files for download distributions.
```
JoshuaLuckers: I don’t think this is the right description for this file 😋

JoshuaLuckers (changes requested):

Jako (approved):

opengeek (approved):

PR #14162, opened 2018-11-28
Improve deprecated notice for "Old modTemplateVar getRender input"

Mark-H:
### What does it do?
Aside from changing the wording and adding context to make it a bit clearer what's the problem, this also moves the deprecated notice into the conditional. 

Before, the message would get logged a lot for each first TV of a certain type due to the way the foreach handles registered render methods. Now it only gets logged when a deprecated flat file is used.

### Why is it needed?
Make sure only relevant deprecated notices get logged and with appropriate context.

### Related issue(s)/PR(s)
#14136

### DIFF
--- a/core/model/modx/modtemplatevar.class.php
+++ b/core/model/modx/modtemplatevar.class.php
@@ -425,7 +425,6 @@ class modTemplateVar extends modElement {
             $output = $render->render($value,$params);
         } else {
             $deprecatedClassName = $method == 'input' ? 'modTemplateVarInputRenderDeprecated' : 'modTemplateVarOutputRenderDeprecated';
-            $this->xpdo->deprecated('2.2.0', '', 'Old modTemplateVar getRender ' . $method . 'method');
             $render = new $deprecatedClassName($this);
 
             foreach ($paths as $path) {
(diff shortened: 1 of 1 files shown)


core/model/modx/modtemplatevar.class.php
```diff
@@ -443,6 +442,7 @@ public function getRender($params,$value,array $paths,$method,$resourceId = 0,$t
                 /* 2.1< backwards compat */
                 $renderFile = $path.$type.'.php';
                 if (file_exists($renderFile)) {
+                    $this->xpdo->deprecated('2.2.0', '', '<2.2 style template variable flat render file ' . $renderFile . ', for TV ' . $this->get('name'));
```
Jako: Don't include the deprecated since version twice and add a message, who has to fix that issue:
```
$this->xpdo->deprecated('2.2.0', '', 'Using old style template variable flat render file ' . $renderFile . ', for TV ' . $this->get('name') . '. Please contact the author of the template variable type to change this code part. );
```

Jako (changes requested):

core/model/modx/modtemplatevar.class.php
Mark-H: Do we need to include "Please contact the author of blabla" in all the deprecation notices? Seems unnecessary to me.

Mark-H: Are you happy with the updated message @Jako?

Jako (approved):

alroniks (approved):

alroniks:
Need your help with approval @opengeek

opengeek (approved):

PR #14178, opened 2018-12-03
Some code inspection fixes

JoshuaLuckers:
### What does it do?
It fixes some issues/suggestions after running the code inspection.

### Why is it needed?
To improve the quality of the code

### Related issue(s)/PR(s)
None that I'm aware of.

### DIFF
--- a/manager/controllers/default/context/update.class.php
+++ b/manager/controllers/default/context/update.class.php
@@ -86,7 +86,7 @@ class ContextUpdateManagerController extends modManagerController {
         }
         /* prepare context data for display */
         if (!$this->context->prepare()) {
-            return $this->failure($this->modx->lexicon('context_err_load_data'), $this->context->toArray());
+            return $this->failure(sprintf($this->modx->lexicon('context_err_load_data'), $this->context->toArray()));
         }
 
         /* invoke OnContextFormPrerender event */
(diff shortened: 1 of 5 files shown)


manager/controllers/default/context/update.class.php
```diff
@@ -86,7 +86,7 @@ public function process(array $scriptProperties = array()) {
         }
         /* prepare context data for display */
         if (!$this->context->prepare()) {
-            return $this->failure($this->modx->lexicon('context_err_load_data'), $this->context->toArray());
+            return $this->failure(sprintf($this->modx->lexicon('context_err_load_data'), $this->context->toArray()));
```
Mark-H: This doesn't seem right - either `$this->context->toArray()` should be the second param to `$this->modx->lexicon` if it's meant to add information to the error message, or the context should be returned in a different way if it's expected to be in the failure response... my guess is it should be the first.

Mark-H (changes requested):

manager/controllers/default/context/update.class.php
JoshuaLuckers: I think you are right. Will check it tomorrow.

JoshuaLuckers: Do I also need to change the lexicon when I pass `$this->context->toArray()` as second parameter in `$this->modx->lexicon`?

Mark-H: I'm not sure what's supposed to happen in that line, but if the lexicon string uses [[+placeholders]], no. Perhaps it's not using the context in which case it could also be removed.

Mark-H (approved):

alroniks (approved):

PR #14194, opened 2018-12-11
Adds modifiers to get information from filepath: dirname, basename, fi…

kudashevs:
### What does it do?
Adds group of modifiers to work with placeholders contain filepath. This modifiers allow:

- dirname - get directory name only (without filename)
- basename - get file basename (filename with extension)
- filename - get filename (filename without extension)
- extension - get file extension

In fact this is a wrapper for PHP's pathinfo() function.

### Why is it needed?
Using modifiers more convenient and simple way to work with placeholders contain filepath.

### Related issue(s)/PR(s)
This PR don't related to any issue. This is proposal only.

### DIFF
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -595,16 +595,16 @@ class modOutputFilter {
                             break;
 
                         case 'dirname':
-                            $output = pathinfo($output)['dirname'];
+                            $output = pathinfo($output, PATHINFO_DIRNAME);
                             break;
                         case 'basename':
-                            $output = pathinfo($output)['basename'];
+                            $output = pathinfo($output, PATHINFO_BASENAME);
                             break;
                         case 'filename':
-                            $output = pathinfo($output)['filename'];
+                            $output = pathinfo($output, PATHINFO_FILENAME);
                             break;
                         case 'extension':
-                            $output = pathinfo($output)['extension'];
+                            $output = pathinfo($output, PATHINFO_EXTENSION);
                             break;
 
                         case 'toPlaceholder':
--- a/core/model/modx/filters/modoutputfilter.class.php
+++ b/core/model/modx/filters/modoutputfilter.class.php
@@ -594,6 +594,19 @@ class modOutputFilter {
                             $output = urldecode($output);
                             break;
 
+                        case 'dirname':
+                            $output = pathinfo($output)['dirname'];
+                            break;
+                        case 'basename':
+                            $output = pathinfo($output)['basename'];
+                            break;
+                        case 'filename':
+                            $output = pathinfo($output)['filename'];
+                            break;
+                        case 'extension':
+                            $output = pathinfo($output)['extension'];
+                            break;
+
                         case 'toPlaceholder':
                             $this->modx->toPlaceholder($m_val,$output);
                             $output = '';


JoshuaLuckers (changes requested):
Thanks for taking time to create this PR!

I have two requests that might improve the chance of getting this into the core:
1. Please elaborate why this needs to be in the core (instead of being provided by an extra).
2. It's better to target this PR against 3.x

kudashevs:
@JoshuaLuckers thank you very much for respond. About your suggestions:

1. Any extra, in our case additional snippet, takes extra place in components block in manager, makes additional db query, takes time to add to project. It's is not critical, but i don't see any reason not to include this simple modifiers (at least basename and extension) in modifiers code base. 
2. The current default on github is 2.x.

P.S. Right now i make a project related to file catalog and faced with problem of manipulating filepaths in TVs and placeholders. And there is no convenient way to do it in chunks or templates. That's why i spend little time to make MODX a little better. From my point of view :)

JoshuaLuckers (dismissed):

core/model/modx/filters/modoutputfilter.class.php
```diff
@@ -594,6 +594,19 @@ public function filter(&$element) {
...
+                            $output = pathinfo($output)['dirname'];
+                            break;
+                        case 'basename':
+                            $output = pathinfo($output)['basename'];
+                            break;
+                        case 'filename':
+                            $output = pathinfo($output)['filename'];
+                            break;
+                        case 'extension':
+                            $output = pathinfo($output)['extension'];
```
alroniks: In manual said that if a file does not have an extension value won't be returned. It should be additionally checked to avoid warnings or notices.
http://php.net/manual/en/function.pathinfo.php
Also, please read the note about files that starts with dot (`.`). Filename also can be empty.

alroniks (changes requested):
Hello. Thank you for your impact. ❤️ 

But I have a few requirements before we will go ahead.
See my comment about code and please, target the PR to 3.x branch. Current branch left as a base while we are collecting bugs and fixes for them for 2.7.0 version. But all new feature should come to 3.x as we are not going to release 2.8.

core/model/modx/filters/modoutputfilter.class.php
kudashevs: Hi, @Alroniks. Thanks a lot for you time.

About warning and notices. The pathinfo() function doesn't produce any warning or notices. Even if empty value will be given, this function will return empty "basename" and empty "filename". I think this behavior will not change, since function has not changed since version 5.2.

I can make check for empty output and return something: "basename not found" or "filename empty". But i think more reasonable return same as the function returns, because modifiers are only a wrappers for this function.

Any suggestions or idea?

kudashevs (commented):
Hi, @Alroniks. Thanks a lot for code review. Now i get an idea, how to improve a little this PR.

About warning and notices i wrote in [review discussion](https://github.com/modxcms/revolution/pull/14194/files/eed96c1acb2491201bdc1e678ef0e32a7aa78e52#r240814546).

Any suggestions or idea? I will move PR to 3.x branch ASAP.

core/model/modx/filters/modoutputfilter.class.php
alroniks: It is not correct.
```
php > $path = 'filename';
php > print_r(pathinfo($path));
Array
(
    [dirname] => .
    [basename] => filename
    [filename] => filename
)
```
Here `extension` key does not exist in the array but you try get array item with this key
```
$output = pathinfo($output)['extension'];
```
PHP will throw Notice that index not found.

kudashevs: Sorry for my inattention (i have tested pathinfo function and forgot about array dereferencing).

I update this PR with [new code](https://github.com/modxcms/revolution/pull/14194/files), which completely lies on pathinfo() and now it works without any warnings.

alroniks:
I am closing this as a new one created.

PR #14195, opened 2018-12-11
Recognize error_log_file* settings when switching contexts

christianseel:
### What does it do?
This PR moves the logic for handling the `error_log_filename` and `error_log_filepath` system-settings into the `_initContext` method instead of the modx `initialize` method. (`_initContext` is also called in `initialize`, so no breaking change)

### Why is it needed?
Previously those settings worked, but only after the modx initialization and before any potential context switching. For multi-context projects, where you use a context switching / routing plugin (e.g. XRouting), those plugins normally use `switchContext()` which calls `_initContext` which uses
```
$this->setLogTarget($this->getOption('log_target', $options, 'FILE', true));
```
So here the previous logTarget gets overridden by any context/systemsetting or the default `FILE`. The potentially targetoptions filename/filepath get lost. So we end up writing to the default error.log after any context switch.

Moving the logic in the `_initContext` function solves this easily and keeps existing functionality.

### Related issue(s)/PR(s)
This is a follow up of https://github.com/modxcms/revolution/pull/13763
Same goal as https://github.com/modxcms/revolution/pull/14119

#### Regarding #14119 
#14119 was a recent try to also move the error log easily, however, this strategy has the exact same issue as discussed here. So I'd promote those 2 systemsettings (which can now also be used as context settings) instead when developers want to move the modx error log.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -554,18 +554,6 @@ class modX extends xPDO {
             $this->getCacheManager();
             $this->getConfig();
             $this->_initContext($contextKey, false, $options);
-            $logTarget = $this->getLogTarget();
-            if ($logTarget === 'FILE') {
-                $options = array();
-                $filename = $this->getOption('error_log_filename', $options, '');
-                if (!empty($filename)) $options['filename'] = $filename;
-                $filepath = $this->getOption('error_log_filepath', $options, '');
-                if (!empty($filepath)) $options['filepath'] = rtrim($filepath, '/') . '/';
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

Mark-H (approved):
Seems like a solid improvement and fix for error log not being recognised after switching contexts.

Jako (approved):

christianseel:
Thanks everyone for the fast merge! 😍

alroniks:
You are welcome! :) It's our work.

PR #14196, opened 2018-12-11
Use a secure and not spammable example hostname

Jako:
### What does it do?
Change the example hostname to a valid not spammable hostname.

### Why is it needed?
The previous hostname redirects to a malware (or worse) host.

### Related issue(s)/PR(s)
https://forums.modx.com/thread/104706/virus-link-in-standard-htaccess

### DIFF
--- a/ht.access
+++ b/ht.access
@@ -18,12 +18,12 @@ RewriteBase /
 RewriteRule "/\.|^\.(?!well-known/)" - [F]
 
 
-# Rewrite www.domain.com -> domain.com -- used with SEO Strict URLs plugin
+# Rewrite www.example.com -> example.com -- used with SEO Strict URLs plugin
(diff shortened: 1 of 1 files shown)


ht.access
```diff
@@ -36,15 +36,15 @@ RewriteRule "/\.|^\.(?!well-known/)" - [F]
 # Rewrite secure requests properly to prevent SSL cert warnings, e.g. prevent 
 # https://www.domain.com when your cert only allows https://secure.domain.com
```
JoshuaLuckers: To me it would make more sense if we also use the `example.com` domain here.

Jako: Done.

JoshuaLuckers (approved):

alroniks (approved):

Mark-H (approved):
👍

PR #14202, opened 2018-12-17
Log only flat file processor getRender usage

Jako:
### What does it do?
Log only flat file processor modTemplateVar getRender usage. At the moment all (not registered) class based processors are logged deprecated too.

### Why is it needed?
Remove invalid deprecated logging.

### Related issue(s)/PR(s)
#14136


Mark-H:
Dupe of https://github.com/modxcms/revolution/pull/14162

Issue #14188, opened 2018-12-07
MAJOR PROBLEM with use of static files in 2.7.0


SnowCreative:
### Summary
It's no longer possible to duplicate an existing template, snippet, etc. when these use static files. An error pops up saying there is already a template, snippet, etc. that uses the particular static file name. 

### Step to reproduce
Create a template that uses a static file. Then, open that template and click the "Duplicate" button. Nothing happens. No warning, nothing. Right-click on the template in the tree and select duplicate, and the usual popup appears, but the Static File field is outlined in red, and the error below says the file already exists. It's impossible to save the new template without changing the name of the file.

### Expected behavior
I often create templates that share a common static file, because I use the template to control what TVs appear in the layout for different types of content. My base template is set up to be shared so that I don't end up with a slew of templates that all have the same code in them. The alert is fine, but it should be a warning and not prevent people from creating elements that share the same static file. 

Furthermore, for things like snippets (where you wouldn't actually want them to share a common file), the Duplicate button should bring up the same popup with the option to specify a new file, instead of just doing nothing at all.

### Environment
MODX 2.7.0.

JoshuaLuckers:
Does it work when you create a new template that links to a static file that is already used by another template?

JoshuaLuckers:
This PR is related: 
https://github.com/modxcms/revolution/pull/14135

SnowCreative:
Strangely, yes, I can create a new template and specify an existing static file.

Jako:
Did you force reload the script files?

timx:
Hi, is this fixed? I can't duplicate any existing static template, chunks, etc .

Get error: A static file already exists within the specified path.

Using MODX Revolution 2.7.2-pl

timx:
Hi, is this fixed? I can't duplicate any existing static template, chunks, etc .

Get error: A static file already exists within the specified path.

Using MODX Revolution 2.7.2-pl

timx:
Has anyone had any joy with this?

JoshuaLuckers:
@timx a temporary workaround is to change the name of the static file to something unique in the dialog that pops up when duplicating a static element.

timx:
Thanks @JoshuaLuckers , yeah I've done that cheers, it just creates another static file though, which I don't need.

smg6511:
[...]

PR #14135, opened 2018-11-09
Enable automating static elements workflow rebased

sdrenth:
# Note: This is a rebased PR for https://github.com/modxcms/revolution/pull/13646

# What does it do?
Added settings to allow for an automated static elements workflow and added the handling of creating/removing automated static files.

## Features:

- Added System settings allowing an automated static elements workflow for each element type (Templates, Template Variables, Chunks, Snippets and Plugins)
- Added System setting for setting a default mediasource for your static elements
- Added System setting for setting a default category for your static elements
- Added System setting for setting a static elements basepath
- If set, will automatically create a static file where the static filename is based on the name of the element, stripping it from special characters, removing spaces and making it lowercased.
- When removing/deleting a static element, it will remove the static file (if not used by another -element) and will also remove the parent directory automatically for you if the directory is empty.
- Added static file path to duplicate element window

# Why is it needed?
We use static elements in our workflow and by automating the creation of static elements it will save a lot of manual work and it also keeps the static elements free of clutter by removing unused static element files.

PR #14203, opened 2018-12-17
Fix duplicating static templates, snippets etc. with an empty category

Jako:
### What does it do?
Add the category to the duplicated record.

### Why is it needed?
Fix the duplicate method for static templates, snippets etc. without a category (when static_elements_automate... is disabled) because the category is needed afterwards. 

### Related issue(s)/PR(s)
#14188, #14135

### DIFF
--- a/manager/assets/modext/sections/element/chunk/update.js
+++ b/manager/assets/modext/sections/element/chunk/update.js
@@ -57,6 +57,7 @@ Ext.extend(MODx.page.UpdateChunk,MODx.Component, {
             ,source: this.record.source
             ,static: this.record.static
             ,static_file: this.record.static_file
+            ,category: this.record.category
         };
         var w = MODx.load({
             xtype: 'modx-window-element-duplicate'
--- a/manager/assets/modext/sections/element/plugin/update.js
+++ b/manager/assets/modext/sections/element/plugin/update.js
@@ -57,6 +57,7 @@ Ext.extend(MODx.page.UpdatePlugin,MODx.Component, {
             ,source: this.record.source
             ,static: this.record.static
             ,static_file: this.record.static_file
+            ,category: this.record.category
         };
         var w = MODx.load({
             xtype: 'modx-window-element-duplicate'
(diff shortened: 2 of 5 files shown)


JoshuaLuckers (approved):

ilyautkin:
I checked duplicating static elements on current 2.x and 3.x code - it processes correctly, category for new object sets normally.

How to test your case, @Jako? What I need to do for see the bug?

Maybe this PR not actual now?

Jako:
This issue occurs only on the duplicate button topright in a static element (template, snippet etc.).

ilyautkin:
Yes I checked it too.

Jako:
And only with disabled static_elements_automate… option.

ilyautkin:
Yes, option is disabled:
screen-2019-01-24_15 13 48

Jako:
The issue occurs only when a static chunk has no category. Then a JS error will occur.

Jako:
The issue could be checked this way:
screencast 2019-01-26 22-00-59

ilyautkin:
Finally I was able to reproduce the bug. I have overlooked that element should not have a category. I confirm this bug. PR tested, it fixes this bug. Everything works as expected now.

ilyautkin (approved):

Jako:
I should have written the click path to reproduce the issue in the PR, since the original issue did not contain this.

alroniks (approved):
All works fine. 👍

alroniks:
I've tested described cases and applied changes fix the issue. It can be safely merged. cc @Mark-H

Mark-H (approved):

timx:
Hi, is this fixed? I can't duplicate any existing static template, chunks, etc .

Get error: A static file already exists within the specified path.

Using MODX Revolution 2.7.2-pl

Jako:
The 'A static file already exists within the specified path.' is a different issue.

timx:
Thanks @Jako , I commented on another issue, but this comment got pulled in to this one automatically, not by me! This is the other issue, that @Mark-H  reopened in Jan:  https://github.com/modxcms/revolution/issues/14188

Issue #14206, opened 2018-12-18
Unused static files disappear from server after updating new static file


valokammi:
When I upload new static chunk file and point to it in MODX unused chunk file disappears from server.

### Step to reproduce
Create  new static file and upload to server. Point chunk to new static file, old unused file disappears from server.

### Expected behavior
Old file should stay in server.

### Environment
2.7.0-PHP Version 7.0.33, linux

jonleverrier:
I can reproduce the issue.

Just to re-clarify the steps to reproduce: 
1) Upload 2 files. Lets say `chunk_test1.tpl` and `chunk_test2.tpl`
2) Create a static chunk and point it towards `chunk_test1.tpl` then save.
3) Now point the static chunk from `chunk_test1.tpl` to `chunk_test2.tpl` then save.

`chunk_test1.tpl` is missing from the file system

Jako:
Does this happen with or without enabling `static_elements_automate_chunks`?

valokammi:
Setting is "no".

valokammi:
Same happens when changing static template too.

JoshuaLuckers:
PR #14135 introduced this behaviour:
> When removing/deleting a static element, it will remove the static file (if not used by another -element) and will also remove the parent directory automatically for you if the directory is empty.

Jako:
That behaviour should only be executed, when the automatic static workflow is enabled. So this has to be fixed. @sdrenth Could you look for this?

sdrenth:
@Jako @tassu I've added a PR that fixes this issue

JoshuaLuckers:
Fixed by PR #14208 and PR #14228

PR #14208, opened 2018-12-21
Bug fix where static files are removing if automated static files are turned off

sdrenth:
This fixes the issue where old static files where removed even if the automate static files system setting was turned off for that element type.

### What does it do?
I've added a system setting check that checks if static files are automated for the current element type.

### Related issue(s)/PR(s)
Issue: https://github.com/modxcms/revolution/issues/14206
PR that created the issue: https://github.com/modxcms/revolution/pull/14135

### DIFF
--- a/core/model/modx/modelement.class.php
+++ b/core/model/modx/modelement.class.php
@@ -185,7 +185,7 @@ class modElement extends modAccessibleSimpleObject {
         }
 
         /* Removing old static file when succesfull saved and oldPath has been set. */
-        if ($saved && $oldPath) {
+        if ($saved && $oldPath && $this->isStaticFilesAutomated()) {
             if (@unlink($oldPath)) {
                 $pathinfo = pathinfo($oldPath);
                 $this->cleanupStaticFileDirectories($pathinfo['dirname']);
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
What happens if you select the wrong file by accident (like index.php), save it and change it to the correct file afterwards? 

Is it possible to have those files move to a "trash folder" instead of deleting them from the filesystem immediately?

sdrenth:
@JoshuaLuckers I don't think a trash folder is necessary. If you set up your static file structure correctly the source does not have access to files such as index.php

core/model/modx/modelement.class.php
```diff
@@ -195,6 +195,36 @@ public function save($cacheFlag = null) {
...
+                break;
+            case 'modSnippet':
+                $type = 'snippets';
+                break;
+            case 'modPlugin':
+                $type = 'plugins';
+                break;
+        }
+
+        return (bool) $this->xpdo->getOption('static_elements_automate_' . $type, null, false);
```
alroniks: What is `type` is an empty string? In this case, all times will be called wrong setting. Make sense fetch setting value only when the type is defined.

core/model/modx/modelement.class.php
```diff
@@ -195,6 +195,36 @@ public function save($cacheFlag = null) {
...
+    /**
+     * Determine if static files should be automated for current element class.
+     *
+     * @return bool
+     */
+    protected function isStaticFilesAutomated()
+    {
+        $type = '';
+
+        switch ($this->_class) {
```
alroniks: I suggest replacing this long switch by a simple map implemented by the array.
```
$map = [
    'modTemplate' => 'templates',
    ...
];
```

And after just check if the key exists in the array and fetch proper element to get a system setting.

alroniks (changes requested):

core/model/modx/modelement.class.php
sdrenth: @Alroniks Thank you for the review. I've adjusted the code.

alroniks (approved):

tolanych:
I tested this PR (reproduced from https://github.com/modxcms/revolution/issues/14206#issuecomment-448231556) for setting `static_elements_automate_chunks`.

When this settings as Yes - previous static file removed from file system.
When this setting as No - previous static file stay.

JoshuaLuckers (changes requested):
I tested the following with `static_elements_automate_*` turned off:
- If I change the static file for an element the old file is not deleted (good).
- If I delete an element that uses a static file the static file is deleted (not good).

Jako:
@JoshuaLuckers Your issue should be fixed with #14228. But this PR covers another part.

Jako (approved):

JoshuaLuckers (approved):

opengeek (approved):

PR #14209, opened 2018-12-22
[ReUp] "Use current alias in alias path" in Form Customization sets

### LINKED ISSUE #14193: Missing "Use current alias in alias path" setting in Form Customization
Not finding "Use current alias in alias path" setting in Form Customization? Available fields in modx-page-settings-right-box-left section are: isfolder, searchable, richtext, uri_override and uri.

### Environment
2.7.0

JoshuaLuckers:
### What does it do?
It adds the field `alias_visible` in the schema for modActionField objects.

### Why is it needed?
In 2.7.0 a new resource option "Use current alias in alias path" to allow hiding resources from the URI was introduced. This option was missing in Form Customization.

### Related issue(s)/PR(s)
Fixes #14193

### DIFF
--- a/core/model/schema/modx.action.fields.schema.xml
+++ b/core/model/schema/modx.action.fields.schema.xml
@@ -46,6 +46,7 @@
         <tab name="modx-page-settings-right-box-left">
             <field name="isfolder" />
             <field name="searchable" />
+            <field name="alias_visible" />
             <field name="richtext" />
             <field name="uri_override" />
             <field name="uri" />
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
@JoshuaLuckers If you have time, look at the order of the fields in "Set Information and Fields" tab, it seems to me that should change the order in accordance with the form of editing the resource.

**Here is the new order:**
modx-resource-main-left _(there is an "id" field, but it doesn't work, in my opinion)_
modx-resource-main-right
modx-resource-content
modx-page-settings-left
modx-page-settings-right
modx-page-settings-right-box-left
modx-page-settings-right-box-right

Mark-H (approved):
👍

Mark-H:
@Ruslan-Aleev Sounds like a new issue might be in order (hah) for that.

alroniks (approved):

PR #14210, opened 2018-12-22
Don't include the resource ID in real time aliases

### LINKED ISSUE #14153: Real-Time resource alias shows resource ID
### Summary
Real-Time resource alias shows the resource ID when editing the title.

### Step to reproduce
Edit an existing resource.
Clear the resource alias field.
Try to change the resource title.

### Observed behavior
Real-Time resource alias shows also the resource ID appended to the title.
_For example:_ email-(10)
Only when saving the resource the alias field is correctly saved without the appended ID.

### Expected behavior
Real-Time resource alias should not show the resource ID when editing the title.

### Environment
MODX 2.7.0-dev

JoshuaLuckers:
### What does it do?
When a real time alias has to be generated it uses the original unmodified value of the title.

### Why is it needed?
PR #13914 made ID's visible in manager pages. This was done by adding it to the title variable that is also used when generating real time aliases.

When generating a real time alias we want to use the original value and let the transliterator take care of string filtering. 

### Related issue(s)/PR(s)
Fixes #14153

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -315,6 +315,15 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             });
         }
     }
+    ,generateAliasRealTime: function(title) {
+        // check some system settings before doing real time alias transliteration
+        if (parseInt(MODx.config.friendly_alias_realtime, 10) && parseInt(MODx.config.automatic_alias, 10)) {
+            // handles the realtime-alias transliteration
+            if (this.config.aliaswasempty && title !== '') {
+                this.translitAlias(title);
+            }
+        }
+    }
     ,templateWarning: function() {
         var t = Ext.getCmp('modx-resource-template');
         if (!t) { return false; }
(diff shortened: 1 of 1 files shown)


OptimusCrime:
What does the [ReUp] tag mean?

JoshuaLuckers:
> What does the [ReUp] tag mean?

I used it wrong:
> [ReUp] - If your updates require a core transport rebuild (such as anything modified in the _build directory, database model changes, or default data changes), then prefix your commit message with this. If you see this message, simply rebuild the core transport and run setup/ again.

OptimusCrime:
Is it taken from [here](https://docs.modx.com/revolution/2.x/getting-started/installation/using-modx-revolution-from-svn)? That is a pretty old doc, and I have not seen many people actually use this prefix.

I would suggest making it a part of the open PR template instead, as it would make it more consistent.

If you agree, then I can open a separate issue for this. Sorry for the offtopic by the way.

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -315,6 +315,15 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             });
         }
     }
+    ,generateAliasRealTime: function(title) {
+        // check some system settings before doing real time alias transliteration
+        if (parseInt(MODx.config.friendly_alias_realtime, 10) && parseInt(MODx.config.automatic_alias, 10)) {
```
Mark-H: That's an interesting approach to checking for a boolean value... I see it was that way before, but maybe worth improving this while you've made it into a nice function?

Ruslan-Aleev:
Everything works for me.
MODX 2.7.0, Chrome 71.0.3578.98

opengeek (approved):

manager/assets/modext/widgets/resource/modx.panel.resource.js
ilyautkin: All system settings has type `string` (field `value` of object [modSystemSetting](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/mysql/modsystemsetting.map.inc.php#L38)). Everywhere boolean system settings checking as string: https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L2201

I think, we can't make this improvement within this PR.

alroniks (approved):

PR #14211, opened 2018-12-23
Prevent symlinks and weblink to target itself

### LINKED ISSUE #14145: Symlinks can target itself
### Summary
Symlink ID's are now validated to make sure the resource exists. It does not check if the given ID is the ID of the resource itself. A symlink to itself should not be possible and invalidate the value.

### Step to reproduce
- Edit a symlink
- Fill in the ID of the resource itself
- Save

### Observed behavior
Value is valid and can be saved.

### Expected behavior
Value should be marked as invalid and give an error explaining that it's not possible to create a symlink to itself.

### Environment
MODX 2.7.0-dev

JoshuaLuckers:
### What does it do?
When updating a symlink it checks if the target is not itself.

### Why is it needed?
Symlinks and weblinks should not be able to target itself.

### Related issue(s)/PR(s)
Fixes #14145

### DIFF
--- a/core/model/modx/processors/resource/update.class.php
+++ b/core/model/modx/processors/resource/update.class.php
@@ -516,6 +516,10 @@ class modResourceUpdateProcessor extends modObjectUpdateProcessor {
             return false;
         }
 
+        if ($targetResource->get('id') === $this->object->get('id')) {
+            $this->addFieldError('modx-symlink-content', $this->modx->lexicon('resource_err_symlink_target_self'));
+        }
+
         return true;
     }
 
--- a/core/model/modx/processors/resource/update.class.php
+++ b/core/model/modx/processors/resource/update.class.php
@@ -518,6 +518,7 @@ class modResourceUpdateProcessor extends modObjectUpdateProcessor {
 
         if ($targetResource->get('id') === $this->object->get('id')) {
             $this->addFieldError('modx-symlink-content', $this->modx->lexicon('resource_err_symlink_target_self'));
+            return false;
         }
 
         return true;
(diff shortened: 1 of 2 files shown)


Jako (dismissed):

core/model/modx/processors/resource/update.class.php
```diff
@@ -539,6 +543,10 @@ public function checkWebLinkTarget() {
                 $this->addFieldError('modx-weblink-content', $this->modx->lexicon('resource_err_weblink_target_nf'));
                 return false;
             }
+
+            if ($targetResource->get('id') === $this->object->get('id')) {
+                $this->addFieldError('modx-weblink-content', $this->modx->lexicon('resource_err_weblink_target_self'));
```
alroniks, line 549: But function returns true even if error showed. Should be there `return false;`?

JoshuaLuckers: You are right. When I saw your comment I was thinking, why didn't I noticed it when testing? If it would return true the resource should save but it didn't.. 
Then I checked and figured out that this method returns false if there are errors set: https://github.com/modxcms/revolution/blob/b8e8ac20122878d08bb525a51b472ea77dca3e27/core/model/modx/processors/resource/update.class.php#L167

Jako: The modObjectUpdate processor does not go further if any field error is set during beforeSet or beforeSave. So return true/false is not really needed. https://github.com/modxcms/revolution/blob/b8e8ac20122878d08bb525a51b472ea77dca3e27/core/model/modx/modprocessor.class.php#L862 It is just used to avoid unnecessary tests.

Mark-H (approved):

alroniks (approved):

Issue #13015, opened 2016-05-12
Add a test for the build process in _build/test


Jako:
### Summary

Before merging #12985 there was no error present in Travis but the build process was not successful after (see #13003). Maybe this could be checked somehow.

alroniks:
Closing as unit tests were added.

PR #14213, opened 2018-12-23
Add a Unit Test for testing the build process

JoshuaLuckers:
### What does it do?
It tests if the build script is successful and if the `core.transport.zip` file is created.

### Why is it needed?
To make sure changes don't break the build process.

### Related issue(s)/PR(s)
Fixes issue #13015


Mark-H (approved):
Yay, tests!

alroniks (approved):

PR #14214, opened 2018-12-24
Don't force the manager date time format for the date of birth field

JoshuaLuckers:
### What does it do?
Don't force the manager date time format for the date of birth field.

### Why is it needed?
If we do force it and the manager date format is set diffrently than 'Y-m-d' the date of birth can not be saved.

### Related issue(s)/PR(s)
Issue fixed #11873

**Note:** later we should actually take the manager date format in consideration and fix the conversion properly. **_This is just a temporary solution_** (the issue mentioned above gets a lot of traffic, giving the impression people are actually affected by it).


Mark-H (approved):
A better fix would be a submitFormat option.. but that's apparently ExtJS 4+! 😐 

Pragmatic, short-term fix.

alroniks (approved):

Issue #14212, opened 2018-12-23
Disable Gravatar by default


JoshuaLuckers:
### Summary
Gravatar should be disabled by default for new installations. 

### Why is it needed?
Gravatar is a third party service used to show a profile image when a user has not uploaded a profile picture. In default installations `enable_gravatar` is `true` by default. 

Information is collected automatically when using Gravatar as described in their [Privacy Policy](https://automattic.com/privacy/). This means when installing MODX the first admin user that logs in is almost certain of being tracked by a third party service.

### Suggested solution(s)
Disable Gravatar by default for new installations or make it possible to enable/disable it in the setup UI.

### Related issue(s)/PR(s)
None.

JoshuaLuckers:
> I do not actually see the problem. Users who have a Gravatar photo setup, also agreed on this privacy policy when they did. So there is no problem to be solved i think?

@gerbenvandekraats it’s better to talk about it here in the issue and keep the PR for technical discussions ☺️.
The problem is that is enabled by default. Even if a user doesn’t have a Gravatar the request is still made. The third party receives information about this user like his/her email address, IP etc. Therefore it should be opt-in instead of opt-out.

gerbenvandekraats:
Sorry, my mistake. Let's talk here ;)

I didn't think about the request from non-existing Gravatar users. You're right. Disable by default is indeed the way to go then. Thanks for explaining.

ghost:
Just to weigh in with my $.02 worth. Disabling the Gravatar is the first thing I have done on every new site since the beginning. Well... sometimes I create my not-found resource first.

Mark-H:
I wonder if a gravatar proxy would do away with the privacy concerns?

JoshuaLuckers:
The email address of the user is still send to Gravatar (as md5 hash). In Europe it might also cause some GDPR issues.

philipwhiuk:
Is there a strong argument why Gravatar functionality is in the minimal core rather than as a plugin (which could be optionally bundled but not enabled/installed)?

OptimusCrime:
@JoshuaLuckers If you consider md5 hashes as one way hashing (not considering the weak type of hashing it is, and how rainbow tables might make it possible to find the original email from a md5 hash), then no, this is not a GDPR issue.

GDPR says nothing about anonymous collection of statistics. As long as the information can not identify one particular user or person, then it does not violate GDPR.

JoshuaLuckers:
[...]

PR #14215, opened 2018-12-24
Disabled Gravatar by default for new installations

JoshuaLuckers:
### What does it do?
Gravatar is disabled by default for new installations

### Why is it needed?
Gravatar is a third party service used to show a profile image when a user has not uploaded a profile picture.

Information is collected automatically when using Gravatar as described in their [Privacy Policy](https://automattic.com/privacy/). This means when installing MODX the first admin user that logs in is almost certain of being tracked by a third party service.

### Related issue(s)/PR(s)
#14212

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2031,7 +2031,7 @@ $settings['xhtml_urls']->fromArray(array (
 $settings['enable_gravatar']= $xpdo->newObject('modSystemSetting');
 $settings['enable_gravatar']->fromArray(array (
   'key' => 'enable_gravatar',
-  'value' => true,
+  'value' => false,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'manager',


gerbenvandekraats:
I do not actually see the problem. Users who have a Gravatar photo setup, also agreed on this privacy policy when they did. So there is no problem to be solved i think?

alroniks (approved):
I agree with arguments.

Mark-H (approved):

opengeek:
Wait, wat?

alroniks:
Do you disagree, Jason? @opengeek 
I'm wondering what to do: merge or not this one.

opengeek (approved):
Go ahead and merge, since everyone seems to be on board with this.

Ruslan-Aleev:
@JoshuaLuckers gravatar is also displayed in widgets, probably PR is needed for widgets

JoshuaLuckers:
> @JoshuaLuckers gravatar is also displayed in widgets, probably PR is needed for widgets

If you don't mind opening an issue describing how I can check this that would be great! 😄

Ruslan-Aleev:
@JoshuaLuckers There is no problem, but just not logical behavior, it seems to me, see the image below:

gravatar

JoshuaLuckers:
> @JoshuaLuckers There is no problem, but just not logical behavior, it seems to me, see the image below:
> 
> gravatar

Thanks, I made some suggestions to this PR based on your info: https://github.com/modxcms/revolution/pull/14544

PR #14221, opened 2018-12-27
Small style improvements for the quick create button 

JoshuaLuckers:
### What does it do?
Changes the icon being used for the quick create button in the tree and centers it vertically

### Why is it needed?
A better looking MODX 3

### Related issue(s)/PR(s)
None


Mark-H (approved):
I also fixed the icon centering in #14225, so there might be a small conflict depending on which one gets merged first.

alroniks (approved):

Issue #13969, opened 2018-07-04
MODX 3: small graphical glitches in resource tree


Ruslan-Aleev:
### Summary

1) When you delete resources (checkbox in resource settings) - they are duplicated in the resource tree (until you refresh the page)
![delete_res](https://user-images.githubusercontent.com/12523676/42276853-5ced6810-7fa6-11e8-90b2-3a273d30aa34.png)

2) Visual inaccuracy (as it seems to me), goes from MODX Revo:
unpublished resources look brighter in the resource tree than hidden resources from the menu. In a large resource tree, "font-style: italic" of unpublished resources is not noticeable, but publishing is more important than hiding from the menu. Need styles do the opposite, or add some additional notation for unpublished resources (icon color, or change the icon).
![publish_res](https://user-images.githubusercontent.com/12523676/42278077-a050e5ec-7faa-11e8-9484-0b1833415737.png)

MODX 3,
Goole Chrome 67.0.3396.99

SnowCreative:
You can easily add your own CSS to style the tree elements however you like. Here's a sample for one site I worked on:

.unpublished, .unpublished a span, .unpublished i.icon, .unpublished i.icon-large {
	color:#999 !important;
	font-style: italic;
}
.hidemenu a span {
	font-style: normal;
}
.unpublished.hidemenu a span {
	color:#a8a8a8 !important;
}

This makes the text for Published the main blue color, Unpublished plain gray, and Hide from Menu italic (regardless of published setting). Save your styles in a CSS file ("manager.css" or whatever), create a Plugin called "customManagerStyles" (or whatever), set the System Event to "OnManagerPageInit", and use this code, subsitituting the actual path to your CSS file:

$modx->regClientStartupHTMLBlock('<link rel="stylesheet" type="text/css" href="'.MODX_BASE_URL.'css/manager.css" />');

Jako:
Just an idea. Why not use different icons for published/unpublished/hidden resources/folders or use some badge on the icon.

Mark-H:
This does seem like something we could improve in MODX 3, but I'm not quite sure what the best solution is. 

For context, here's what they look like in 2.7:

![schermafbeelding 2018-12-28 om 21 35 03](https://user-images.githubusercontent.com/312944/50527602-a8f81880-0ae9-11e9-9ead-ff3a872e860d.jpg)

And in current 3.x:

[...]

PR #14225, opened 2018-12-28
[3.x] Clean up some styles, less borders -> more depth, better emphasis

Mark-H:
### What does it do?

With this PR I'm tweaking the 3.x styles a little. This PR is focused on the overall manager frame and a little bit on the resource panel. I worked mostly on the tabs, trees, and header area. 

The primary thing I tried to do was get rid of all.the.borders. There were so many, especially in the tree and the header area, all begging for your attention, causing the manager to feel really confusing and cluttered. Where appropriate, I replaced those with more subtle shadows to convey some more depth/hierarchy instead. 

In some places (like the breadcrumbs) I just removed backgrounds and borders entirely to let it sit against the background. 

All in all, I think this simplifies and focuses the design a bit more. There are still other areas I'd like to work on (grids and introducing some more color), but that's for another day.

### Why is it needed?

The redesign is one of the key features in MODX3, but it wasn't exactly sitting right with me just yet. The early designs by Rinze were much simpler and cleaner than the latest 3.x, so while I didn't actually look up those earlier designs I did try to bring it more in line again with the simplicity that stood out to me about those designs.

### Related issue(s)/PR(s)

N/a.


Mark-H:
Some screenshots for quick comparison:

## Resource update before

resource-before

## Resource update after

resource-after

## Template update before

template-before

## Template update after
template-after

JoshuaLuckers:
I'm in favour of removing the amount of borders!

Two things I would like to note:
- I do miss the distinction in tabs after applying these changes. 
- It would also be a good idea to make the breadcrumb separator a little darker.

Ruslan-Aleev:
Thank you, MODX becomes more beautiful :)
@Mark-H look at the styles of resources hidden from the menu and unpublished resources, in my opinion, they should be reversed, here in more detail - https://github.com/modxcms/revolution/issues/13969

JoshuaLuckers (dismissed):

SnowCreative:
Much better use of space. Thanks.

JoshuaLuckers (dismissed):

Mark-H:
Rebased to resolve a conflict.

JoshuaLuckers (approved):

Jako:
Is there enough space on less wide screens (between the tabs and the buttons topright)? Otherwise the buttons could be less high.

Mark-H:
I'm not sure if I fully understand what you're asking, but the overall layout is responsive (in a weird extjs way), and not affected by this PR. If there are responsive issues I'd be happy to take a look at that in a different PR.

opengeek (approved):

PR #14228, opened 2018-12-31
Don't remove static files if automated static files are turned off

Jako:
This really fixes the issue where old static files where removed even if the automate static files system setting was turned off for that element type. It was only partly fixed in #14208, which added a check for removing the file, when the static file path was changed.

### What does it do?
A check for the automatic static files system setting is added to the modElementRemoveProcessor class.

### Related issue(s)/PR(s)
Issue: #14206, #140206
PR that created the issue: #14135

### DIFF
--- a/core/model/modx/processors/element/remove.class.php
+++ b/core/model/modx/processors/element/remove.class.php
@@ -16,6 +16,9 @@
  * @subpackage processors.element
  */
 abstract class modElementRemoveProcessor extends modObjectRemoveProcessor {
+    public $staticFilePath;
+    public $staticFile;
+
     public function cleanup() {
         $this->clearCache();
     }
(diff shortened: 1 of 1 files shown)


core/model/modx/processors/element/remove.class.php
```diff
@@ -24,18 +27,42 @@ public function clearCache() {
...
-            @unlink($this->staticFilePath);
+        if ($this->isStaticFilesAutomated()) {
+            /* Remove file. */
+            $count = $this->modx->getCount($this->classKey, array('static_file' => $this->staticFile));
+            if ($this->staticFilePath && $count === 0) {
+                @unlink($this->staticFilePath);
+            }
+
+            /* Check if parent directory is empty, if so remove parent directory. */
+            $pathinfo = pathinfo($this->staticFilePath);
```
JoshuaLuckers, line 38:
```suggestion
            $dirname = pathinfo($this->staticFilePath, PATHINFO_DIRNAME);
```

core/model/modx/processors/element/remove.class.php
```diff
@@ -24,18 +27,42 @@ public function clearCache() {
...
+
+            if (!empty($pathinfo['dirname'])) {
```
JoshuaLuckers, line 40: To me it would make more sense to check for things like this in the `cleanupStaticDirectories` method to prevent code duplication.

core/model/modx/processors/element/remove.class.php
```diff
@@ -24,18 +27,42 @@ public function clearCache() {
...
+                $this->cleanupStaticDirectories($pathinfo['dirname']);
```
JoshuaLuckers, line 41:
```suggestion
                $this->cleanupStaticDirectories($dirname);
```

JoshuaLuckers (changes requested):

Jako:
@JoshuaLuckers I agree with your review, but code changes that are not bugfixes should be done in 3.x.

JoshuaLuckers (approved):

Mark-H (approved):
👍

Issue #14067, opened 2018-08-29
Insert Template Variables-tab before the resource Settings-tab


Ruslan-Aleev:
The location of the TV-tab before Settings-tab is more appropriate in terms of UX / UI:

1) By default, the TV tab is located after the settings, but most of the time the user works in the Document-tab and the TV-tab - it is not clear why they are separated by the rarely used Settings-tab. Especially since Document-tab and the TV-tab relate to content.

2) The new MODX-user or content-manager will be intuitively easier to find where all content of the resource located. And even if you do not remember the name of the tab, it's easier to remember that the content in the first 2 tabs.

![new_tabs](https://user-images.githubusercontent.com/12523676/44796426-06bd3800-abbe-11e8-968e-cdcdb6533894.png)

Most likely I unnecessarily pay attention to tabs location, but still ;)

p.s. It is clear that you can move the TV after the content in Settings or remake the location using the Manager Customization

Mark-H:
Moved tabs in #14250

PR #14232, opened 2019-01-02
Insert TV-tab before Settings-tab in resource editing form

Ruslan-Aleev:
### Why is it needed?
The location of the TV-tab before Settings-tab is more appropriate in terms of UX / UI. More details in the related issue.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14067


alroniks (approved):

Mark-H:
Would it be better to have a change like this target 3.0? While not a technical breaking change, it does seem like a big change for the user experience.

ghost:
Doesn't seem intuitive to me, TVs are optional. Resources always have content, settings and Groups. And to be technical TVs were never intended to be extra Resource content fields, they are awkward to use for that purpose which usually includes sorting and searching. I expect to see their use as content fields be reduced as creating proper custom resources with custom fields becomes easier.

PR #14236, opened 2019-01-03
Changed the order of tabs in the Template and TV form

Ruslan-Aleev:
### Why is it needed?
Changed the order in the form of "Template" and "TV" in accordance with the frequency of use of the tab.
Tabs are arranged in different forms in different ways. For example, the tab "Properties" in the form of "Plugin" in last place, and in the form of "Template" and "TV" - in second place.

### Related issue(s)/PR(s)
Not directly, but the logic is the same - https://github.com/modxcms/revolution/pull/14232


JoshuaLuckers:
It would be better to target 3.x.

Ruslan-Aleev:
@JoshuaLuckers It seems to me that PR is more useful for 2.x.
These changes for UX are not critical, even now, not being in the manager panel, it is difficult to remember their current position, but the fact that the tabs are located in different places is more of a bug.
For https://github.com/modxcms/revolution/pull/14232 also, UX is not critical, in my opinion, and useful in 2.x.

JoshuaLuckers:
First of all, thanks for taking time to make MODX better by submitting this PR. 

Like @Mark-H mentioned in #14232 these changes alter the behaviour of the User Interface most users are already used to.
Even if the position of the tabs is not consistent across various places in the manager, changing them will break backwards compatibility. Users might, for example, have automated browser scripts that can break when the order of the tabs change, therefore it's not a good idea to include them in a patch release like 2.7.1 (the next release of MODX).
These changes should be in a minor release like 2.8.x. At the moment there is no plan for a 2.8.x release and therefore they should be targeting 3.x.

Ruslan-Aleev:
@JoshuaLuckers Perhaps you are right, but I disagree with you :)
Let Reviewers decide this question, or reject this PR.

alroniks:
I also guess that it should be targeted to 3.x because it is not a bug. Even if it will improve the current version. We should follow our goal - make stable current version fixing existing bugs only in this version and focused on the next major version.

Mark-H (changes requested):
As discussed, please target 3.x.

Ruslan-Aleev:
@Mark-H Changed, here too - https://github.com/modxcms/revolution/pull/14232

Issue #14242, opened 2019-01-08
Make navigation elements as links with the correct href attribute


Ruslan-Aleev:
### Summary
In Windows and Linux, there is a problem with handling links (buttons) in the MODX manager:

1) In Windows / Linux OS the middle button click event does not hold the proper information in the onclick event, but it does in onmousedown (onmouseup). The onclick event returns only the left mouse button (so left click or ctrl + left click works) - see the pictures below.
_As a result, extJS does not provide a universal link processing solution for different OS._

![incorrect_link](https://user-images.githubusercontent.com/12523676/50841121-6ce07580-137d-11e9-9d18-9c3f2a18eb81.png)

![correct_link](https://user-images.githubusercontent.com/12523676/50841118-6baf4880-137d-11e9-9844-34c911589c97.png)

2) Modern browsers can block links of the form window.open(), confusing them with advertising banners.

**A universal solution would be to make navigation elements in the form of links with the physical attribute href.** It refers to icons and buttons for quickly creating elements, links without href, trash manager.

This problem concerns Windows and Linux (but these are about 90% of users), in Mac OS everything works basically correctly.
And the href attribute is for links and buttons in the tree of elements (perhaps somewhere else, but not sure).

### Environment
Windows, Linux OS, MODX 2.7 (and earlier), modern browsers

Ibochkarev:
### Triage (current 3.x)

Still a fair UX/a11y request. No discussion yet.

Manager navigation still leans on JS (`window.open` / loadPage-style handlers) instead of real `<a href="…">` for many chrome actions. Middle-click / open-in-new-tab stays broken on Win/Linux; popup blockers can interfere.

Not a functional data bug, but still worth keeping for progressive enhancement of manager links (tree quick-create, trash, etc.).

PR #14243, opened 2019-01-08
Added physical links to the href attribute for elements in the tree

Ruslan-Aleev:
Added physical links to the href attribute for elements in the tree. Without correct links in the href attribute, the middle and right mouse buttons in Windows / Linux do not work. Read more here - https://github.com/modxcms/revolution/issues/14242

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14059

### DIFF
--- a/manager/assets/modext/util/utilities.js
+++ b/manager/assets/modext/util/utilities.js
@@ -234,7 +234,7 @@ Ext.override(Ext.tree.TreeNodeUI,{
 
         var cb = Ext.isBoolean(a.checked),
             nel,
-            href = this.getHref(a.href),
+            href = n.attributes.page,
             iconMarkup = '<i class="icon'+(a.icon ? " x-tree-node-inline-icon" : "")+(a.iconCls ? " "+a.iconCls : "")+'" unselectable="on"></i>',
             elbowMarkup = n.attributes.pseudoroot ?
                 '<i class="icon-sort-down expanded-icon"></i>' :
--- a/manager/assets/modext/util/utilities.js
+++ b/manager/assets/modext/util/utilities.js
@@ -234,7 +234,7 @@ Ext.override(Ext.tree.TreeNodeUI,{
 
         var cb = Ext.isBoolean(a.checked),
             nel,
-            href = n.attributes.page,
+            href = a.page,
             iconMarkup = '<i class="icon'+(a.icon ? " x-tree-node-inline-icon" : "")+(a.iconCls ? " "+a.iconCls : "")+'" unselectable="on"></i>',
             elbowMarkup = n.attributes.pseudoroot ?
                 '<i class="icon-sort-down expanded-icon"></i>' :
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (dismissed):

OptimusCrime:
This looks good! Any idea what `getHref` actually does?

Ruslan-Aleev:
I also did not understand what getHref is used for.
The getHref call is in the files:
/manager/assets/ext3/ext-all-debug.js
/manager/assets/ext3/ext-all.js
/manager/assets/modext/modx.jsgrps-min.js
/manager/assets/modext/util/utilities.js

Jako:
It does nothing special, if href is set. Otherwise it returns a hash or an empty space depending on the browser:
```
    getHref : function(href){
        return Ext.isEmpty(href) ? (Ext.isGecko ? '' : '#') : href;
    },
```

Jako:
If a.page contains an url everywhere and can't be empty, I am fine with the solution. Otherwise surround that thing with `this.getHref(a.page)`

Ruslan-Aleev:
@Jako Empty url will be, for example, for images in files.
And what is the fundamental difference between the empty attribute href="" and the href="#" (for empty url or with getHref check)?

ghost:
href="#" will cause a jump to the top of the page. https://stackoverflow.com/questions/5637969/is-an-empty-href-valid

Ruslan-Aleev:
@Jako It’s probably better to wrap getHref, I just don’t understand what this check gives :)

Ruslan-Aleev:
@sottwell No, the page will not move to the top, now all links in the tree with href="#"

Jako:
If I am right, an empty href attribute in an a tag creates cursor issues in some browsers (and maybe does not work properly even with javascript). So it is filled with an # for some browsers (not isGecko - whatever that really means in ExtJS).

Jako:
ExtJS 3.4 is quite old and has to handle all those nasty browser incompatibilities to that time. Maybe some checks are not needed anymore.

OptimusCrime:
@Jako See this: https://en.wikipedia.org/wiki/Gecko_(software)

Jako:
@OptimusCrime Thanks. Gecko should be deprecated now, but maybe isGecko means something different in ExtJS (I don't know the test).

OptimusCrime:
It should be deprecated indeed. Calls to `isGecko`, `isGecko2` , and `isGecko3` is found all over the source. I found this:

```
        isWebKit = check(/webkit/),

[...]

        isGecko = !isWebKit && check(/gecko/),
        isGecko2 = isGecko && check(/rv:1\.8/),
        isGecko3 = isGecko && check(/rv:1\.9/),
```

It basically sniffs the user agent to see if the browser is Firefox or some other Gecko based browser.  

This test should be irrelevant in 2019.

ilyautkin:
PR was tested. Found bug:

Open resource updating page. Hover any link in the tree. All links in tree gets current page url as href attribute:

1zpun4

Browser - Crome 71.0.3578.80 (64 bit) for Ubuntu

ilyautkin:
PR is tested, it works.
Now middle click is available. Awesome =)

JoshuaLuckers (approved):

opengeek (approved):

Issue #14240, opened 2019-01-08
Vertical Tabs explode if TV names longer than 169px


IamNeo1:
The template variable container is defined with a fixed width of 169px. If the names of the template variables are longer, the whole layout of the vertical tabs explodes.

![modx_long_tv_names_1](https://user-images.githubusercontent.com/8747516/50828344-67fdd080-1341-11e9-9691-90bd153f56f7.PNG)

It is simple to reproduce and simple to fix.

After each update of the manager CSS I have to search and adjust the two places where the width is fixed to 169px. It is already commented as "dirty hack". Why not just set it to "auto". Then the content area adapts.

index.css 6068 rule: `.vertical-tabs-panel .vertical-tabs-header`

![modx_long_tv_names_2](https://user-images.githubusercontent.com/8747516/50828541-f5d9bb80-1341-11e9-9a70-fe317403e4fb.PNG)

index.css 6112 rule: `.vertical-tabs-panel .vertical-tabs-header .x-tab-strip-wrap ul.x-tab-strip > li.x-tab-strip-active`
![modx_long_tv_names_3](https://user-images.githubusercontent.com/8747516/50828641-3c2f1a80-1342-11e9-82e7-3ab64255ffb5.PNG)

See here with "auto" at both rules
![modx_long_tv_names_4](https://user-images.githubusercontent.com/8747516/50828810-bd86ad00-1342-11e9-9ab6-93ae34ec0c5a.PNG)

Please fix that.

Ruslan-Aleev:
@IamNeo1 <del>I can't reproduce this bug, how do you test it?</del>

PR #14247, opened 2019-01-09
Fix Vertical Tabs explode if TV names longer than 169px #14240

Ibochkarev:
### Why is it needed?
Change width Vertical Tabs explode if TV names longer from 169px to auto

### Related issue(s)/PR(s)
#14240

### DIFF
--- a/_build/templates/default/sass/_tabs.scss
+++ b/_build/templates/default/sass/_tabs.scss
@@ -165,7 +165,7 @@
 }
 
 ul.x-tab-strip-top li:first-child {
-  /*margin-left: -1px*/;
+  /*margin-left: -1px*/
   margin-left: 0;
 }
 
@@ -301,7 +301,7 @@ ul.x-tab-strip-bottom {
     float: left;
     margin-bottom: -10000px; /* dirty hack to make vertical tabs container stretch to bottom */
     padding-bottom: 10000px !important; /* dirty hack to make vertical tabs container stretch to bottom */
-    width: 169px !important; /* aligns the vertical tabs with the TVs tab left edge, will not work that nicely with non-english langs */
+    width: auto !important; /* aligns the vertical tabs with the TVs tab left edge, will not work that nicely with non-english langs */
 
     .x-tab-strip-wrap {
       background-color: transparent; /* as vertical tab panels are nested ones too, do not apply the background color for nested tab panels */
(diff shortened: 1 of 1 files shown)


_build/templates/default/sass/_tabs.scss
```diff
@@ -165,7 +165,7 @@
 }
 
 ul.x-tab-strip-top li:first-child {
-  /*margin-left: -1px*/;
+  /*margin-left: -1px*/
```
JoshuaLuckers, line 168: Now that this line has been edited it makes more sense to be completely removed.

_build/templates/default/sass/_tabs.scss
```diff
@@ -301,7 +301,7 @@ ul.x-tab-strip-bottom {
     float: left;
     margin-bottom: -10000px; /* dirty hack to make vertical tabs container stretch to bottom */
     padding-bottom: 10000px !important; /* dirty hack to make vertical tabs container stretch to bottom */
-    width: 169px !important; /* aligns the vertical tabs with the TVs tab left edge, will not work that nicely with non-english langs */
+    width: auto !important; /* aligns the vertical tabs with the TVs tab left edge, will not work that nicely with non-english langs */
```
JoshuaLuckers, line 304: The comment for this line is no longer valid after this change.

_build/templates/default/sass/_tabs.scss
```diff
@@ -338,7 +338,7 @@ ul.x-tab-strip-bottom {
             border-right-color: $white;
             box-shadow: none; /* removes the active tab strip on top */
             color: $tabActiveText;
-            width: 169px; /* make the active li 1px more wide to cover the containers right border, this makes the right border on inactive tabs necessary as the whole tab-strip wrap gets wider */
+            width: auto; /* make the active li 1px more wide to cover the containers right border, this makes the right border on inactive tabs necessary as the whole tab-strip wrap gets wider */
```
JoshuaLuckers, line 341: Is the comment still valid?

JoshuaLuckers (changes requested):
Thanks for taking time to submit this PR and make MODX better. 

~~Since this is your first time contributing you have to sign the CLA first. For more info:~~
~~- [Becoming a Contributor](https://docs.modx.com/community/contribute/becoming-a-contributor)~~
~~- [Fill out and send in a Contributor License Agreement (CLA)](http://develop.modx.com/contribute/cla/)~~

Mark-H:
One thing to check before this gets merged is how the vertical tabs behave in modal windows. I seem to recall some of the `#hacky` was to make them behave nicely in such contexts... that said, I can't think of any places in the core that uses vtabs in modals, so that may just be extras (like ContentBlocks).

Ruslan-Aleev:
<del>I can't reproduce this bug.</del>
My mistake, you had to set a name larger than 169px in ONE WORD.

Ibochkarev:
Found a better solution to correct this error in the PR #14317

PR #14251, opened 2019-01-09
Changed the order of tabs in the Template and TV form

Ruslan-Aleev:
### Why is it needed?
Changed the order in the form of "Template" and "TV" in accordance with the frequency of use of the tab.
Tabs are arranged in different forms in different ways. For example, the tab "Properties" in the form of "Plugin" in last place, and in the form of "Template" and "TV" - in second place

### Related issue(s)/PR(s)
Not directly, but the logic is the same - https://github.com/modxcms/revolution/issues/14067

### DIFF
--- a/manager/assets/modext/widgets/element/modx.panel.template.js
+++ b/manager/assets/modext/widgets/element/modx.panel.template.js
@@ -263,14 +263,6 @@ MODx.panel.Template = function(config) {
 				}]
 			}]
         },{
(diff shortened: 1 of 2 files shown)


Ruslan-Aleev:
@Mark-H I closed the old PR with tabs and created new on 3.x base

JoshuaLuckers (approved):

Mark-H (approved):

PR #14255, opened 2019-01-11
Change the dependencies install display

Jako:
### What does it do?
- Open the dependency tab, when the dependencies are not met.
- Change MODX/PHP dependencies display, since they can't be downloaded/installed at that place.
- Use less space for installed packages/available versions.

### Why is it needed?
Before:
Before
After:
After

### Related issue(s)/PR(s)
None known

### Needs Build
Due to CSS changes the the theme needs `grunt rebuild`


JoshuaLuckers:
Will this also change the order when updating an extra? I would prefer to see the changelog first there.

Jako:
The dependencies are only displayed, if the package has some. But then they are first. I will try to add a tab change, if the dependencies are not met.

Jako:
Now the cangelog is first again and the tab is changed, when the dependencies are not met.

manager/assets/modext/workspace/package.panels.js
```diff
@@ -35,9 +35,9 @@ MODx.panel.PackageMetaPanel = function(config) {
 };
 Ext.extend(MODx.panel.PackageMetaPanel,MODx.VerticalTabs,{
 	updatePanel: function(meta, record){
-		if(meta.changelog != undefined){
-			this.addTab(_('changelog'), 'changelog', meta);
-		}
+        if(meta.changelog != undefined){
```
JoshuaLuckers, line 38:
```suggestion
        if (meta.changelog !== undefined) {
```

manager/assets/modext/workspace/package.panels.js
```diff
@@ -147,9 +147,9 @@ Ext.extend(MODx.panel.PackageBeforeInstall, MODx.panel.PackageMetaPanel,{
         Ext.getCmp('package-list-reset').show();
         installBtn.hide().signature = '';
         setupoptionsBtn.hide();
-		if(meta.changelog != undefined){
-			this.addTab(_('changelog'), 'changelog', meta);
-		}
+        if(meta.changelog != undefined){
```
JoshuaLuckers, line 150:
```suggestion
        if (meta.changelog !== undefined) {
```

JoshuaLuckers (commented):
For a few lines the indentation has changed from tabs to spaces, since these lines have been changed it's good to add strict comparisons for these lines.

Jako:
Now it uses less space for installed packages (requirements are met)

manager/assets/modext/workspace/package.panels.js
Jako: Let's do that in 3.0

manager/assets/modext/workspace/package.panels.js
Jako: Let's do that in 3.x

JoshuaLuckers (dismissed):

Jako:
'available' and 'not available' display value for PHP and MODX.

Jako:
Now we 'just' need a displayname and name column in package dependencies. pdoTools is found but i.e. SEO Pro and Image+ could not be installed as dependency, see #12666 

This should be fixed with #14261 and #14260.

JoshuaLuckers (approved):

manager/assets/modext/workspace/package.panels.js
Mark-H: IMO there's nothing wrong with correcting code style on things you're editing in 2.x. As these lines are already marked as changed, correcting the code style wouldn't cause extra conflicts.

Mark-H:
What's a good extra to test this against?

Jako:
I’ll send one by mail/slack.

Jako:
The Crosslinks package contains PHP/MODX dependencies. If I remember right, the BlogIt package contains extras dependencies.

ghost:
I know that Flatso has dependencies; I believe it was intentionally designed to demonstrate the use of them as well as some kind of theming system.

opengeek (approved):

PR #14257, opened 2019-01-11
Fix margin .modx-page-header 

Ibochkarev:
### What does it do?
Corrects indents at heading

### Why is it needed?
Fix header indents - add margins property

Before fix:
Before

After fix:
After

### Related issue(s)/PR(s)
Not know


Ibochkarev:
@Jako @JoshuaLuckers Please close. I am create new PR in branch 3.x

Issue #12126, opened 2014-11-10
Wrong escaping for image template variables


stefan-scholl:
(Only tested in 2.3.1-pl, sorry.)
Had a directory with "&" in it and the manager showed it as "&amp;"

E.g. "pictures/Something & Something/nice-logo.png" ==> "pictures/Something &amp; Something/nice-logo.png"

When this resource got updated the wrong path was in the template variable and the image wasn't displayed anymore.

I guess it's the wrong escaping in the Smarty template manager/templates/default/element/tv/renders/input/image.tpl
Default is escaping for HTML. You need to specify JavaScript.

``` diff
--- manager/templates/default/element/tv/renders/input/image.tpl.orig   2014-11-10 09:50:24.849809382 +0100
+++ manager/templates/default/element/tv/renders/input/image.tpl        2014-11-10 09:51:26.834895745 +0100
@@ -13,7 +13,7 @@
         xtype: 'displayfield'
         ,tv: '{$tv->id}'
         ,renderTo: 'tv-image-{$tv->id}'
-        ,value: '{$tv->value|escape}'
+        ,value: '{$tv->value|escape:'javascript'}'
         ,width: 400
         ,msgTarget: 'under'
     {literal}
@@ -32,8 +32,8 @@
         xtype: 'modx-panel-tv-image'
         ,renderTo: 'tv-image-{$tv->id}'
         ,tv: '{$tv->id}'
-        ,value: '{$tv->value|escape}'
-        ,relativeValue: '{$tv->value|escape}'
+        ,value: '{$tv->value|escape:'javascript'}'
+        ,relativeValue: '{$tv->value|escape:'javascript'}'
         ,width: 400
         ,allowBlank: {if $params.allowBlank == 1 || $params.allowBlank == 'true'}true{else}false{/if}
         ,wctx: '{if $params.wctx}{$params.wctx}{else}web{/if}'
```

JoshuaLuckers:
Fixed by merging PR #14263

PR #14263, opened 2019-01-12
Fix escaping for image template variables #12126

Ibochkarev:
### What does it do?
Had a directory with "&" in it and the manager showed it as "&"
E.g. "pictures/Something & Something/nice-logo.png" ==> "pictures/Something & Something/nice-logo.png"

When this resource got updated the wrong path was in the template variable and the image wasn't displayed anymore.

### Why is it needed?
Wrong escaping in the Smarty template manager/templates/default/element/tv/renders/input/image.tpl
Default is escaping for HTML. You need to specify JavaScript.

### Related issue(s)/PR(s)
#12126


JoshuaLuckers (approved):
This PR does fix the described issue. While testing I stumbled upon another bug I will report later (directories with & in the name don't work properly in the media browser).

Mark-H (approved):

Issue #13954, opened 2018-07-03
MODX 3: Manager Sidebar - Width is increased after browser resize


jonleverrier:
### Summary
If you resize your browser window to a small size and then increase it, the sidebar width becomes very large and overlaps content below.

### Step to reproduce
1. Load the manager
2. Shrink your browser in width to see the "mobile" view
3. Resize your browser back again

### Observed behavior
The sidebar becomes large.
The content below the sidebar is no longer full width.

![screen shot 2018-07-03 at 11 57 38](https://user-images.githubusercontent.com/2373940/42213474-22ae2f1a-7eb9-11e8-826e-dc42aa7f0a39.png)

### Expected behavior
The sidebar to retain it's original width after resize.

### Environment
MODX3 Alpha v 3.0.0-dev
From 3.x branch and https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99
Firefox 61.0
Safari 11.1

jonleverrier:
Is this a problem inherited from MODX 2.x does anybody know?

Mark-H:
Fixed in #14271 - though we have seemed to regress and the manager no longer works properly on mobile.

PR #14271, opened 2019-01-13
Fix Width Manager Sidebar - is increased after browser resize #13954

Ibochkarev:
### What does it do?
If you resize your browser window to a small size and then increase it, the sidebar width becomes very large and overlaps content below.

### Why is it needed?
Change min-width #modx-header -> max-width.

Before fix:
Before

After fix:
After

### Related issue(s)/PR(s)
#13954

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -20,7 +20,7 @@
 
 #modx-header {
   background: $colorSplash;
-  min-width: $navbarWidth;
+  max-width: $navbarWidth;
   position: absolute;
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):
This PR does solve the problem as described. The manager is still not usable on smaller screens.

Mark-H (approved):

Mark-H:
The manager used to work on smaller screens in 2.x, so that it doesn't work anymore in 3.x is a regression. We'll need to get that sorted - but not in this PR. Thank you @Ibochkarev.

Issue #14177, opened 2018-12-03
Incorrect display TV with "Tag" input type


Ruslan-Aleev:
### Summary
In the resource, the TV field with the "Tag" input type is displayed incorrectly if not set Input Option Values. For other TVs, if Input Option Values is not set, there is no input field, for Tag, the field appears, but incorrect.

![tag_error](https://user-images.githubusercontent.com/12523676/49387962-77836b80-f73c-11e8-83e6-577cdc157443.png)

### Environment
MODX 2.7

jonleverrier:
I can reproduce the issue too.

Issue #11823, opened 2014-07-25
[2.3.1] Checkbox TV doesn't display in Manager


aesop1:
In 2.3.0 or 2.3.1 my previously visible checkbox TVs (the checkbox itself) are not visible (they are there, I can mouse over them and see the pointer). CSS issue.

It appears the new Manager style for checkboxes isn't being applied to the TVs after the standard HTML checkboxes are hidden via the CSS opacity (opacity: 0) attribute at line 21 of the Manager's index.css file:

.x-form-check-wrap .x-form-checkbox, .x-form-check-wrap input[type=checkbox], .x-form-check-wrap .x-form-radio, .x-fieldset-checkbox-toggle legend .x-form-checkbox, .x-fieldset-checkbox-toggle legend input[type=checkbox], .x-fieldset-checkbox-toggle legend .x-form-radio {
cursor: pointer;
opacity: 0; /_This is hiding the standard checkboxes_/
position: absolute; 
top: 0;
left: 0;
width: 18px;
height: 18px;
z-index: 1;
}

Ruslan-Aleev:
Added empty check - https://github.com/modxcms/revolution/pull/14275

Ibochkarev:
### Triage (current 3.x)

Still looks open. #14275 (empty option check) was closed unmerged.

Cause from the thread: manager CSS hides the native checkbox (`opacity: 0`); the visible icon hangs off `label.x-form-cb-label`. Checkbox TV options without a label text never get that label, so the control is invisible but still clickable.

`checkbox.tpl` still sets `boxLabel` from `$item.text` with no empty fallback. Keep open until a fix lands (placeholder label / always render `boxLabel`).

PR #14275, opened 2019-01-15
Added a empty check for option values in TV

Ruslan-Aleev:
### What does it do?
Added a empty check for option values in TV for which the option values are required.
If the option values are not set, an error message is displayed.
Also, unnecessary scripts are not generated for such TV, if the option values are not set.
Slightly corrected code.

### How to check:
1) Create a TV with an input type for which option values are required, for example, checkbox.
2) Do not set in the option values when TV create.
3) Go to the resource where this TV is and save.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14177 
#11823


JoshuaLuckers (commented):
Instead of committing the same type of changes for each file individually it's better to group them together in one commit. This ensures a good history of the changes and makes it easier to, for example, undo the changes if needed.

Ruslan-Aleev:
Thanks for the info, I haven't figured out much with git yet.
Can you give a correct example?

Jako:
The integrators could do a squash commit.

JoshuaLuckers:
Can you provide some more info and steps how we can validate the changes you made? I tested the changes and they do not seem to work for me.

Ruslan-Aleev:
@JoshuaLuckers Added information to the PR description.

JoshuaLuckers:
Thanks, now it makes sense. 

Is there a reason why the presence of "Input Option Values" isn't checked when creating/editing a Template Variable?

Ruslan-Aleev:
Yes, it is odd, in general are a lot of bugs in the form of creating a TV, for example, the "Allow Blank" field does not work everywhere (for some TV it does not exist at all, although it should be for all TV), the "Input Option Values" field is not needed for some TVs, but it is.
I think, checking the "Input Option Values" when creating a TV is a more competent variant.

Ibochkarev:
Guys, we probably forgot about this PR. It is necessary to restore justice and check it. :sunglasses:

Ruslan-Aleev:
Can anyone help with this issue - https://github.com/modxcms/revolution/issues/14424?
Then this PR can be closed, because I did not do it very well here.

Issue #14262, opened 2019-01-12
Core Finder not seeing moved core in 2.7.x?


thewhiterabbit:
### Summary
Installing fresh 2.7.x from stable recommended on the download page. Not able to lock down MODX core for security.

### Step to reproduce
1) Moved core to parent directory (public_html equivalent on my server)
2) Get the screen for Core Finder
3) Enter the new core path verbatim
4) Click "Submit"
ERROR: Your MODX_CORE_PATH is invalid; please specify the correct path in the field above and click Submit.
5) Cleared cache in core directory
6) Click "Submit" again...
ERROR: Your MODX_CORE_PATH is invalid; please specify the correct path in the field above and click Submit.
![screenshot_2019-01-11_18-46-40](https://user-images.githubusercontent.com/5334222/51068913-32bfdf80-15da-11e9-99dc-d7fbd8e31965.png)

### Observed behavior
Not recognizing the moved core when it's typed into Core Finder input field.
Still shows the error message; ERROR: Your MODX_CORE_PATH is invalid; please specify the correct path in the field above and click Submit.

I even chmod everything to 777 for now just to make sure it wasn't permissions issue.

### Expected behavior
Should be setting the new core path in the config file and moving to the next setup screen.

### Environment
MODX version 2.7.x,  Apache/2.4.18, MySQL  Ver 14.14 Distrib 5.7.24, Chrome

Jako:
The path should contain a trailing slash. This should be part of the description or the error message.

thewhiterabbit:
Thank you, Jako. Okay, let me try that again.

Do you think it may be a better idea if we add a conditional trigger that adds the trailing slash silently in the background were it not included, to ease the confusion and perpetuate the ease of installation?

I try to avoid error control, as much as possible, from reporting back to the end user unless there are clear instructions on what the exact cause is, or a reference code.

This error was very vague and highly confusing.

Update: Adding the trailing slash worked. Thank you again, Jako.

Ibochkarev:
Fixed. In Increased error message incorrect #14276 . Can be closed.

PR #14276, opened 2019-01-15
Enhance the error message of a wrong MODX_CORE_PATH

Jako:
### What does it do?
Add a message, that the core path has to contain a trailing slash.

### Why is it needed?
Make that mistake more obvious.

### Related issue(s)/PR(s)
#14262

### DIFF
--- a/setup/templates/findcore.php
+++ b/setup/templates/findcore.php
@@ -101,7 +101,7 @@ if (!is_writable(MODX_SETUP_PATH . 'includes/config.core.php')) {
 } else {
 ?>
                         <span class="field_error">ERROR: Your MODX_CORE_PATH is invalid; please specify the correct path in the
-                        field above and click Submit.</span>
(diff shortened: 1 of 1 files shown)


alroniks (approved):

JoshuaLuckers (approved):

JoshuaLuckers:
More specific validation would even be better which might be nice to have in 3.x

Mark-H (approved):

SnowCreative:
Couldn't the script just add a slash on the end if there is none? One line of code would do it. This comes under the category "unneccessarily geeky for non-programmers," just like when I get the annoying message on a purchasing page, "credit card number must not contain spaces or dashes." People commonly enter spaces or dashes, so just have your script strip them out!

Jako:
You are welcome to prepare a PR with one line of code. It will have a better chance to be accepted, if you don't use "unneccessarily geeky for non-programmers".

Issue #5537, opened 2011-08-17
Tab rank in FC goes up in increments of 10 every time I save the Set


ghost:
paramaniac created Redmine issue ID 5537

I am using Form Customization to move all my TVs from the TV tab to the Create/Edit tab.

_Rant_
Shouldn't have to do this because all custom TVs should appear on the bottom of the create/edit tab by default like MODx Evo, why do they appear on their own tab in Revo!!!
_Rant over_

When I use the Tab Rank to order them and then save the FC Set, and then return to the set: via Security > Form Customization > Edit Profile > Edit Set
all the values I put in for tab rank have gone up by 10! Save the Set again and return, and the values have gone up by 10 again!
I have even set all the tab ranks to 0 and it still happens.

Flowan:
This issue remains in 2.5.7 - It's even worse, all rank-values fall back to 0

gpsietzema:
#modxbughunt #1point to @Flowan

sonicpunk:
@gpsietzema saw you give the point here live!

OptimusCrime:
This issue is really, really weird. Should definitely be fixed.

alroniks:
Good news! Issue solved and PR is merged.

PR #14281, opened 2019-01-16
Fixed bug with unexpected incremented TV ranks in Form Customization

ilyautkin:
### What does it do?
Fix for issue https://github.com/modxcms/revolution/issues/5537

### Why is it needed?
There are bug in form customisation when updating FC set - ranks for TVs goes up in increments of 10 every saving.

### Related issue(s)/PR(s)
Issue https://github.com/modxcms/revolution/issues/5537

### DIFF
--- a/core/model/modx/modformcustomizationset.class.php
+++ b/core/model/modx/modformcustomizationset.class.php
@@ -151,7 +151,8 @@ class modFormCustomizationSet extends xPDOSimpleObject {
                         break;
                     case 'tvMove':
                         $tvArray['tab'] = $rule->get('value');
-                        $tvArray['rank'] = ((int)$rule->get('rank'))-10;
+                        /* subtract 20 from rank that have been added in update processor */
+                        $tvArray['rank'] = ((int)$rule->get('rank'))-20;
                         if ($tvArray['rank'] < 0) $tvArray['rank'] = 0;
                         break;
                 }


Mark-H:
The code change seems straightforward enough - but do you know what's going on in the first place that it needs to remove 20 from the rank? That seems so random 😅

ilyautkin:
> The code change seems straightforward enough - but do you know what's going on in the first place that it needs to remove 20 from the rank? That seems so random

All rules are sorted by `rank` field before applying. Ranks are specified directly [in the code](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/security/forms/set/update.class.php#L96). Also rank field is used to sort TV in the region and user can change `rank`. MODX use dirty hack to prevent conflicts of ranks - adding for TV ranks [20 points](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/security/forms/set/update.class.php#L305).

Certainly the best solution is adding field `order` for sort rules. Then we can use `rank` field only for sort TVs. But this solution will change MODX model and I think it is some dangerous.

OptimusCrime:
This is one of those "wtf is that magic number". If this had been a constant with a good name to begin with, we would not have this conversation, and this bug might never have existed.

Mark-H (approved):
Thanks for the explanation. Super random number, but hey, that's what you get in code from a decade ago ;)

JoshuaLuckers (approved):

Issue #11579, opened 2014-06-25
[2.3] Manager TopNav Menu Icons a Little "Unorthodox"


mindeffects:
Hi there,
call me picky but I don't get the logic of the new manager topnav icons: The main elements have an "arrow down" while it is obvious clear that they are "buttons". The menu items inside a dropdown WITH children don't have any icon, which is obviously not clear. – To be honest, it is only "Reports" which would need an icon like "arrow right".

Maybe the whole topnav menu could loose some weight if ALL arrow would be dropped?! We don't need them!

![image](https://user-images.githubusercontent.com/458619/51240764-b54ef300-197c-11e9-9426-b4841664e638.png)

The arrows in the RED circles could all be dropped (and also all the arrow in the right part of the top-menu bar).
Inside the GREEN circles "arrows to the right" would be nice to have.

cheers
Oliver

Ibochkarev:
Actually for versions 2.6.x, 2.7.x and 3.x

JoshuaLuckers:
This will be fixed by the redesign in MODX 3.
Related PR #14292

PR #14289, opened 2019-01-16
Added an additional .sub class in the menu for li which have child items

Ruslan-Aleev:
### What does it do?
Added an additional class .sub in the menu for li in #modx-footer .modx-subnav which have child items. 
Added an extra icon to the .sub class (:after).

sub_info

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/11579

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -190,16 +190,16 @@
         }
       }      
       &.sub {
-      &:after {
-        @extend %pseudo-font;
-        position: absolute;
-        color: $subnavDescriptionColor;
-        content: fa-content($fa-var-caret-right);
-        font-size: 14px;
-        margin-top: 0;
-        top: 20px;
-        right: 15px;
-        }
+        &:after {
+          @extend %pseudo-font;
+          position: absolute;
+          color: $subnavDescriptionColor;
+          content: fa-content($fa-var-caret-right);
+          font-size: 14px;
+          margin-top: 0;
+          top: 20px;
+          right: 15px;
+          }
       }
       a {
         border-radius: $borderRadius;
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -190,15 +190,15 @@
         }
       }      
       &.sub {
-        &:after {
-          position: absolute;
-          top: 12px;
-          right: 15px;
-          width: 6px;
-          height: 6px;
-          content: " ";
-          border-radius: 50%;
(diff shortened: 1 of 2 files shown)


Ibochkarev:
Please make a change to the Scss:
- icon is used
- A menu with a drop-down list can be seen immediately, and not when you hover.
- fix position

```
&:hover {
  &:after {
    border-right-color: $subnavBgHover;
  }
  
}
&.sub {
  &:after {
    @extend %pseudo-font;
    position: absolute;
    color: inherit;
    content: fa-content($fa-var-caret-right);
    font-size: 14px;
    margin-top: 0;
    top: 20px;
    right: 15px;
  }
}
```
**After fix:**
https://github.com/Ibochkarev/other-repo/blob/757f0bd2696e73fb330420993828f50451d58c47/%2314289.png

JoshuaLuckers (commented):
The changes don't seem to be working. Are they working for you?

Ibochkarev:
@JoshuaLuckers

_build/templates/default/sass/_navbar.scss
```diff
@@ -188,6 +188,18 @@
         &:after {
           border-right-color: $subnavBgHover;
         }
+      }      
+      &.sub {
+      &:after {
```
JoshuaLuckers: Please add indentation.

_build/templates/default/sass/_navbar.scss
```diff
@@ -188,6 +188,18 @@
...
+        @extend %pseudo-font;
+        position: absolute;
+        color: $subnavDescriptionColor;
+        content: fa-content($fa-var-caret-right);
```
JoshuaLuckers:
```suggestion
        content: $fa-var-caret-right;
```

JoshuaLuckers (changes requested):

_build/templates/default/sass/_navbar.scss
alroniks: fa-content is correct here, it's a new function from the updated font-awesome

Ruslan-Aleev: @Alroniks This is probably worth changing in the future, because the old variant is found here and in other scss files.

Ruslan-Aleev:
Close PR, because suggested changes from a non-updated 3.x branch

JoshuaLuckers:
Sorry for suggesting changes that didn't need to be done @Ruslan-Aleev.

Ruslan-Aleev:
@JoshuaLuckers This is for the better :) because @Alroniks helped find out that my branch was not updated.

Issue #11751, opened 2014-07-20
[2.3] Misplaced slide-in/slide-out button of Resource tree when scrollbar is visible 


gadgetto:
The slide-in/slide-out button of Resource tree is misplaced when scrollbar is visible + it overlaps the scrollbar. Also the padding between tree border and button changes.

![screenshot 2014-07-20 23 12 01](https://cloud.githubusercontent.com/assets/876737/3638569/33d6bcb8-1053-11e4-8013-f6b1a209fcf7.png)

alroniks:
It seems issue solved by just merged pr.

PR #14298, opened 2019-01-19
Fix position slide-in/slide-out button of Resource tree

Ibochkarev:
### What does it do?
Fix position slide-in/slide-out button of Resource tree.

### Why is it needed?
Fix styles

**Before**:
Before
**After**:
After

### Related issue(s)/PR(s)
Issue #11751

### DIFF
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -93,10 +93,10 @@
     background-repeat: no-repeat;
     background-position: center left;
     border-radius: 0 $borderRadius $borderRadius 0;
-    left: -9px;
+    left: -25px;
     top: 68px;
     width: 11px; /* to have left and right 3px equal space from the arrow which is 5px */
-    height: 41px;
+    height: 39px;
     opacity: 0.5;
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
Unfortunately, this variant does not work for Windows OS, we will need to think a little longer :)

JoshuaLuckers:
@Ruslan-Aleev can you provide a screenshot?

Ruslan-Aleev:
@JoshuaLuckers 

left_1

left_2

Ruslan-Aleev:
@JoshuaLuckers The problem is not the position of the icon, but the fact that the scroll is located close to the element tree.
More correctly to increase the width of the tree by the width of the icon, and increase the wrapper of the tree by the width of the scroll. The width of the scroll also should be the largest among the different OS scrolls.

@Ibochkarev and I have already discussed this moment.

Ruslan-Aleev:
It works on Windows OS now :)

JoshuaLuckers (approved):

opengeek (approved):

Mark-H (approved):

PR #14300, opened 2019-01-20
Solution for long dropdown submenu

GulomovCreative:
### What does it do?
1. Add new system setting `topmenu_submenu_max_items`
2. Check this setting and moves remainings items to new item with label "More"
3. Add right arrow/triangle for menu items that have children

**Process gif**

ux

screenshot-revolution-2019 01 20-18-23-00

### Why is it needed?
Description from #11939

> When there's a long Extras dropdown menu, and not enough screen space, the bottom items get cut off. It's not possible to scroll down or access them.

### Related issue
Closes #11939

### DIFF
--- a/manager/controllers/default/header.php
+++ b/manager/controllers/default/header.php
@@ -310,7 +310,7 @@ class TopMenu
         //$output .= '<ul class="modx-subnav">'."\n";
 
 		$moreMenu = '';
-		if($maxItems && count($menus) > $maxItems) {
+		if ($maxItems && count($menus) > $maxItems) {
 			$moreMenu = array_slice($menus, $maxItems);
 			$menus = array_slice($menus, 0, $maxItems);
 		}
--- a/manager/controllers/default/header.php
+++ b/manager/controllers/default/header.php
@@ -61,6 +61,8 @@ class TopMenu
         $this->controller =& $controller;
         $this->modx =& $controller->modx;
         $this->showDescriptions = (boolean) $this->modx->getOption('topmenu_show_descriptions', null, true);
+        $this->mainNavParent = (string) $this->modx->getOption('main_nav_parent', null, 'topnav', true);
+        $this->subMenuMaxItems = (int) $this->modx->getOption('topmenu_submenu_max_items', null, 0, true);
     }
 
     /**
@@ -77,7 +79,7 @@ class TopMenu
         $mainNav = $this->modx->smarty->getTemplateVars('navb');
         if (empty($mainNav)) {
             $this->buildMenu(
-                $this->modx->getOption('main_nav_parent', null, 'topnav', true),
+                $this->mainNavParent,
                 'navb'
             );
         }
@@ -194,7 +196,7 @@ class TopMenu
 
             if (!empty($menu['children'])) {
                 $menuTpl .= '<ul class="modx-subnav">'."\n";
-                $this->processSubMenus($menuTpl, $menu['children']);
+                $this->processSubMenus($menuTpl, $menu['children'], ($name == $this->mainNavParent ? $this->subMenuMaxItems : 0));
                 $menuTpl .= '</ul>'."\n";
             }
             $menuTpl .= '</li>'."\n";
@@ -303,9 +305,15 @@ class TopMenu
      *
      * @return void
      */
-    public function processSubMenus(&$output, array $menus = array())
+    public function processSubMenus(&$output, array $menus = array(), int $maxItems = 0)
     {
         //$output .= '<ul class="modx-subnav">'."\n";
+
+		$moreMenu = '';
+		if($maxItems && count($menus) > $maxItems) {
+			$moreMenu = array_slice($menus, $maxItems);
+			$menus = array_slice($menus, 0, $maxItems);
+		}
 
         foreach ($menus as $menu) {
             if (!$this->hasPermission($menu['permissions'])) {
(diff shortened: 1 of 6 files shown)


JoshuaLuckers:
PR’s with new functionality should target the 3.x branch

Ruslan-Aleev:
@JoshuaLuckers This problem is not relevant for 3.x. I think better in 2.x.
Although PR with an icon in the list is present for 3.x

manager/controllers/default/header.php
```diff
@@ -303,10 +305,16 @@ public function hasPermission($perms)
      *
      * @return void
      */
-    public function processSubMenus(&$output, array $menus = array())
+    public function processSubMenus(&$output, array $menus = array(), int $maxItems)
```
JoshuaLuckers:
```suggestion
    public function processSubMenus(&$output, array $menus = array(), $maxItems = false)
```

MODX should be compatible with PHP 5.6 and higher, scalar type hints are available in PHP 7+ only. 
A default value should be used to maintain backwards compatibility.

manager/controllers/default/header.php
```diff
@@ -340,6 +355,13 @@ public function processSubMenus(&$output, array $menus = array())
             $this->childrenCt++;
         }
 
+		if(!empty($moreMenu)) {
```
JoshuaLuckers:
```suggestion
		if (!empty($moreMenu)) {
```

manager/controllers/default/header.php
```diff
@@ -328,9 +336,16 @@ public function processSubMenus(&$output, array $menus = array())
...
+			$classes = '';
             if (!empty($menu['children'])) {
+				$hasChildren = true;
+				$classes = ' class="right"';
+			}
+            $smTpl .= '<a'.$attributes.$classes.'>'.$menu['text'].$description.'</a>'."\n";
+
+            if ($hasChildren) {
                 $smTpl .= '<ul class="modx-subsubnav">'."\n";
                 $this->processSubMenus($smTpl, $menu['children']);
```
JoshuaLuckers, line 350: The maximum amount of items should also apply to child submenu's.

manager/controllers/default/header.php
```diff
@@ -340,6 +355,13 @@ public function processSubMenus(&$output, array $menus = array())
             $this->childrenCt++;
         }
 
+		if(!empty($moreMenu)) {
+			$output .= '<li><a class="right">'.$this->modx->lexicon('more').'</a>'."\n";
+			$output .= '<ul class="modx-subsubnav more">'."\n";
+			$this->processSubMenus($output, $moreMenu);
```
JoshuaLuckers, line 361: The maximum amount of items should also apply here.

JoshuaLuckers (changes requested):

manager/controllers/default/header.php
```diff
@@ -303,10 +305,16 @@ public function hasPermission($perms)
      *
      * @return void
      */
-    public function processSubMenus(&$output, array $menus = array())
+    public function processSubMenus(&$output, array $menus = array(), int $maxItems)
     {
         //$output .= '<ul class="modx-subnav">'."\n";
 
+		$moreMenu = '';
+		if($maxItems && count($menus) > $maxItems) {
```
JoshuaLuckers:
```suggestion
		if ($maxItems && count($menus) > $maxItems) {
```

Jako:
It should be relevant for 3.x, too. Scrolling in menus should be made avoidable, if possible.

Ruslan-Aleev:
@Jako Yes, literally a couple of hours found the same problem in MODX3, although I had not noticed before. Therefore, it is not clear what to do with PR.

manager/controllers/default/header.php
GulomovCreative: You're right

manager/controllers/default/header.php
GulomovCreative: Do you think that the amount of items in the third level of the menu will be so much that you need to hide it?

JoshuaLuckers: It's possible, if it requires too much change it would be good to do that in 3.x instead.

Mark-H:
So, I see 2 scenarios for this (really cool!) improvement.

We either approve it for 2.x, in its current form, and then when 2.x gets merged into 3.x that'll likely cause weird conflicts that need to be manually resolved.

Or, this PR is updated to target 3.x instead. That probably means a bit more work to get things looking just as nice, but it'll avoid that responsibility falling on the integrators doing a merge (which is typically complicated enough, and those integrators may not always know all the intricacies of individual PRs when doing that). 

So (for completely selfish reasons :D), my vote would go to making it ready for 3.x instead. That then also adds another great reason to hurry with finishing that up for a proper release, and we could add a sensible default (maybe 10?) to the setting that controls it so it kicks in automatically :)

GulomovCreative:
> So, I see 2 scenarios for this (really cool!) improvement.
> 
> We either approve it for 2.x, in its current form, and then when 2.x gets merged into 3.x that'll likely cause weird conflicts that need to be manually resolved.
> 
> Or, this PR is updated to target 3.x instead. That probably means a bit more work to get things looking just as nice, but it'll avoid that responsibility falling on the integrators doing a merge (which is typically complicated enough, and those integrators may not always know all the intricacies of individual PRs when doing that).
> 
> So (for completely selfish reasons :D), my vote would go to making it ready for 3.x instead. That then also adds another great reason to hurry with finishing that up for a proper release, and we could add a sensible default (maybe 10?) to the setting that controls it so it kicks in automatically :)

@Mark-H I got it. I will make it ready for 3.x. Thx!

JoshuaLuckers:
@GulomovCreative is this still on your radar? If you need any help please let us know.

Mark-H:
Closing as the consensus is to recreate this for 3.

Ruslan-Aleev:
Guys, maybe it's worth merge this fix into the 2.x after all? On some projects it is not possible to work with the menu in principle.

PR #14301, opened 2019-01-22
Bugfix for thumbnails not rendering for s3 image URL's with spaces.

gregcrown:
### What does it do?
Updates file source URL's with encoded spaces.

### Why is it needed?
Media sources with spaces in URL's (example: s3) were unable to generate thumbnails.

### Related issue(s)/PR(s)
No related issues.


Jako (commented):
urlencode or rawurlencode is not an option?

Jako:
Will S3 media sources be deprecated in 3.x? The API has changed for a lot of new buckets and the current code does not work anymore.

JoshuaLuckers:
> Will S3 media sources be deprecated in 3.x? The API has changed for a lot of new buckets and the current code does not work anymore.

Doesn't Flysystem cover this in 3.x?

Jako:
Flysystem supports this, but it has to be configured.

There is an extra available for new S3 buckets.

gregcrown:
> urlencode or rawurlencode is not an option?

I explored this option however the variables that are affected already contained full URL's. To use urlencode or rawurlencode would require breaking up the complete URL's into segments and then rebuilding the URL. Explicitly replacing the spaces removes the extra steps.

Please suggest a more elegant approach if you have one though. 👍

Jako (dismissed):

gregcrown:
I needed to amend my commit to include a check for URL match in:

core/model/phpthumb/phpthumb.class.php

I should have done this to begin with. This will allow filesystem images to be handled without the URL fix.

core/model/phpthumb/phpthumb.class.php
```diff
@@ -280,6 +280,9 @@ public function purgeTempFiles() {
 	public function setSourceFilename($sourceFilename) {
 		//$this->resetObject();
 		//$this->rawImageData   = null;
+		if (preg_match('#^[a-z0-9]+://#i', $sourceFilename)) {
```
JoshuaLuckers: Please add a comment with an explanation to indicate what this does and why it's needed.

JoshuaLuckers (changes requested):

core/model/phpthumb/phpthumb.class.php
gregcrown: This is the same URL checking logic that exists on lines 581 and 906. Should we add comments to all of these instances as well?

JoshuaLuckers: Only here is fine with me.

Jako: That class is external code and can't be changed.

Jako (changes requested):

Ibochkarev:
@gregcrown Please continue to work on this PR. Thanks

gregcrown:
This PR has been updated by reseting the original phpThumb and making the required changes to the MODx managed modPhpThumb class. I've also added a comment above the URL detection for clarity.

Jako (approved):

JoshuaLuckers (approved):

Mark-H (approved):

Mark-H:
Hey @gregcrown, I was about to merge this but it looks like you haven't signed the MODX CLA yet. Could you please fill out the form at https://develop.modx.com/contribute/cla/ and let me know when you've done so? 

Thanks!

gregcrown:
Accepted! Thanks for the heads up!

Issue #14302, opened 2019-01-23
In 2.7.0 installing extras gives deprecation error for different exta


BobRay:
### Summary
When I install an Extra that has a routing-based menu item, I see an error message in red referencing another extra which doesn't.

For example, when installing the new version of Orphans, which contains no action object and has one routing-based menu item, I see this:

> modAction support is deprecated since version 2.3.0. Support for modAction has been replaced with routing based on a namespace and action name. Please update the extra with the namespace **sitecheck** to the routing based system.

Many people won't read to the end of this and will assume that Orphans has a deprecated menu system. I don't think this is caused by a left-over action because the action is removed in the Orphans validator, unless the system is responding to an older version of Orphans that is still in core/packages, but in that case the namespace in the message would be 'orphans.' 

A similar error message gets written to the error log when you drop down the Extras menu, so I don't think it's necessary for this to appear when extras are installed.

### Step to reproduce

Install an extra with a modAction. Install another extra with a routing-based menu. 

### Observed behavior
You seen the error message in the console during package installation

### Expected behavior
No error message

### Environment
MODX version, apache/nginx and version, mysql version, browser, etc. Any relevant information.
MODX 2.7.0
Localhost/XAMPP
PHP 7.0

sebastian-marinescu:
Very annoying, as my initial automated setup (with Gitify package-installation) now "fails" (still usable, but red and not green with a big ok anymore).

PR #14303, opened 2019-01-23
Log deprecated modAction for the current installed package only during package install/uninstall

Jako:
### What does it do?
Set a modX/xPDO class option, that allows to bypass the $this->xpdo->deprecated call in the modMenu->getSubMenus method while running the install/uninstall processors.

### Why is it needed?
Log deprecation of modAction for the current installed package only during package install/uninstall, since the log entry is visible in red in the installation logs and the deprecated message is often triggered by a previous installed package.

### Related issue(s)/PR(s)
#14302

### DIFF
--- a/core/model/modx/modmenu.class.php
+++ b/core/model/modx/modmenu.class.php
@@ -116,7 +116,9 @@ class modMenu extends modAccessibleObject {
             // allow 2.2 and earlier actions
             $deprecatedNamespace = $menu->get('action_namespace');
             if (!empty($deprecatedNamespace)) {
-                $this->xpdo->deprecated('2.3.0', 'Support for modAction has been replaced with routing based on a namespace and action name. Please update the extra with the namespace ' . $deprecatedNamespace . ' to the routing based system.', 'modAction support');
+                if (!defined('MODX_PACKAGE_INSTALL') || MODX_PACKAGE_INSTALL == $deprecatedNamespace) {
+                    $this->xpdo->deprecated('2.3.0', 'Support for modAction has been replaced with routing based on a namespace and action name. Please update the extra with the namespace ' . $deprecatedNamespace . ' to the routing based system.', 'modAction support');
+                }
                 $namespace = $deprecatedNamespace;
             }
             if ($namespace != 'core') {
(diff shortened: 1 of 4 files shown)


Mark-H:
I'm not entirely convinced this is necessary.. is it really a problem that the package installation shows which packages need to be updated from modAction?

opengeek:
I concur with Mark. This seems helpful to know.

Jako:
As BobRay stated in his issue, the warnings are often caused by other previous installed packages. And the warning is red.

For testing, you could install Collections and afterwards a package that installs an own menu entry, that does not use modAction. The warnings during installing the second package are caused by Collections.

Mark-H:
I understand it shows the message for unrelated packages, but I still don't personally consider that a real problem. When updating packages, I find the messages useful to see what other packages will also need attention. As it will likely take a little time for all packages to catch up, it's a useful periodic reminder.

I'm probably a 3 on the [SSOGAF](http://blog.capwatkins.com/the-sliding-scale-of-giving-a-fuck) though, so if this is truly bothering others more than not having it bothers me, then I'm okay with giving it a thumbs up.

Jako:
The deprecated issues should be logged only once a day. At the moment I switch them off everywhere, to be not annoyed, because there are too much extras that still use i.e. modAction.

The 'other packages' messages will show up soon enough in the error log.

Jako:
The last change restrict the logging of modAction deprecation to the current package during install/uninstall.

opengeek (changes requested):
I do not feel a define is a good choice here. Perhaps a static variable would be better?

Jako:
@opengeek Now it uses a modX/xPDO class option.

Mark-H:
This seems pretty nasty to me.. does it have to be an array? I think the code would be a lot cleaner if it were just a simple string.

Jako:
I introduced the array, because a package could install another package inside. I can't remember where, but I have seen this in a theme package.

alroniks:
Tickets extra installs two packages, for example: since this line https://github.com/bezumkin/Tickets/blob/master/_build/resolvers/resolve.setup.php#L135

opengeek (approved):

Issue #11888, opened 2014-08-09
Installer can't find core path when core is moved/renamed


Pont:
When doing an upgrade of an advanced install with the core moved/renamed the installer doesn't set the core path correctly and thus displays the findcore page. This could be avoided if the installer tried to use the config.core.php in the websites root. Something like this in `/setup/includes/config.core.php` solves the issue.

``` php
if (!include(MODX_INSTALL_PATH . 'config.core.php')) {
    define('MODX_CORE_PATH', MODX_INSTALL_PATH . 'core/');
    if (!defined('MODX_CONFIG_KEY')) define('MODX_CONFIG_KEY', 'config');
}
define ('MODX_SETUP_KEY', '@advanced@');
```

Mark-H:
Fixed in #14304

PR #14304, opened 2019-01-23
Check config in root for search moved core path

ilyautkin:
### What does it do?
Check `config.core.php` file when installing MODX. Now we can find moved or renamed core path automaticly.

### Why is it needed?
Installer can't find core path when core is moved/renamed

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/11888

### DIFF
--- a/setup/includes/config.core.php
+++ b/setup/includes/config.core.php
@@ -1,4 +1,10 @@
 <?php
-if (!defined('MODX_CORE_PATH')) define('MODX_CORE_PATH', MODX_INSTALL_PATH . 'core/');
+if (!defined('MODX_CORE_PATH')) {
+    if (file_exists(MODX_INSTALL_PATH . 'config.core.php')) {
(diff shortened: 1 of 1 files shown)


opengeek:
Why don't you just edit/replace the one in `setup/includes`? This isn't really doing anything automatic.

Mark-H:
Isn't that what this PR does @opengeek? Sorry, I'm a little confused.

opengeek:
No, this one just looks for one in the root of setup to get values from. The one in setup/includes should just be edited or replaced by one in the same way one could be put in the root.

opengeek:
Or wait, MODX_INSTALL_PATH is the location where modx is to be installed. Duh! I guess this is alright then.

opengeek (approved):

Mark-H (approved):

Mark-H:
Thanks @ilyautkin!

Issue #13948, opened 2018-07-03
MODX 3: login page checkbox


meshkov:
### Summary
checkbox on the login page is off-center
![2018-07-03_10-24-55](https://user-images.githubusercontent.com/1703670/42204978-d836dca0-7eab-11e8-9735-c33787618b9d.png)

### Environment
MODX 3.0
Windows 10
Google Chrome 67.0.3396.99

jonleverrier:
I wonder if this is a Windows issue. As I can't reproduce on;
MacOSX 10.13.4
MODX3 Alpha v 3.0.0-dev

Firefox 61.0
![screen shot 2018-07-03 at 16 56 39](https://user-images.githubusercontent.com/2373940/42227585-1e44b2ae-7ee2-11e8-82ce-9561397c8588.png)

Chrome Version 67.0.3396.99
![screen shot 2018-07-03 at 16 57 24](https://user-images.githubusercontent.com/2373940/42227638-371f384e-7ee2-11e8-8261-ba09d17038fa.png)

Safari 11.1
![screen shot 2018-07-03 at 16 58 12](https://user-images.githubusercontent.com/2373940/42227676-4ddec5a4-7ee2-11e8-99cc-31b29055aa8a.png)

meshkov:
Test on MacOSX 10.3.5 in Chrome Version 67.0.3396.99 and all is well.

But problem exist on Windows 10 in Google Chrome 67.0.3396.99.

And on iOS 11.4 Safari:
![photo_2018-07-03_22-39-23](https://user-images.githubusercontent.com/1703670/42241139-3389672e-7f12-11e8-8f18-2dc2d2f78e5a.jpg)
![photo_2018-07-03_22-39-50](https://user-images.githubusercontent.com/1703670/42241222-7b2bd1ca-7f12-11e8-969f-d1a6bc9d79ef.jpg)

Ibochkarev:
This problem has been fixed earlier.

**Tested in browsers:**
Google Chrome 71.0.3578.98
Safari 12.0.2 (13606.3.4.1.4)
Firefox Developer Edition 65.0b9 (64-bit)

MacOS High Sierra 10.13.6 (17G4015)
MODX 3.0.0-pl

![Tested](https://github.com/Ibochkarev/other-repo/blob/54a25c71f12f6a92d132f71a6340d34188fda2a3/%2313948.png?raw=true)

I vote for closing this problem.

Ruslan-Aleev:
[...]

PR #14306, opened 2019-01-23
Correction on the consistency of design and buttons

Ibochkarev:
### What does it do?
Improves the consistency of the design, and the color of the buttons and their state when you hover the mouse

### Why is it needed?
**Fixed**:
- button color, hover colors
- removed unnecessary styles after replacing colors
- replaced red, yellow, darkGray and orange colors with clearer and more suitable
- changed styles of borders of panels and tabs

Before fix:

After fix:

After

### Related issue(s)/PR(s)
Issue #13958

PR #14310, opened 2019-01-26
Minor authorization page fixes styles

Ibochkarev:
### What does it do?
With this PR, I adjust the styles a little 3.x. This PR is focused on changing the visual representation of the authorization page.

I have corrected:
- centered and increased the size of the logo
- text indents, input field descriptions
- link colors, input field descriptions
- seems to have centered the position of the checkbox and the text =)
- font sizes
- button colors their hover behavior
- styles at the form input
- styles of error message and successful operation

### Why do you need it?
Redesign is one of the key features of MODX3 and I tried to bring the styles of this page a little closer to the overall design.

Before:
modx revolution 2019-01-26 05-59-49
modx revolution 2019-01-26 05-59-36

After:
login modx revolution - 2019-01-30 18-14-40
login modx revolution - 2019-01-30 18-15-14
login modx revolution - 2019-01-30 18-15-30

### Related issue(s)/PR(s)
#13948


_build/templates/default/sass/_colors-and-vars.scss
```diff
@@ -19,8 +19,12 @@ $black: darken($colorSplash, 42.5%); /* generate the black from the $colorSplash
 $yellow: #FFEA80;
 $orange: #F49548;
 $red: #BE0000;
+$lighterRed: #FFEEEE;
 $green: #6CB24A;
+$lighterGreen: #EFFCF6;
+$greener: darken($green, 12%);
 $blue: #3697cd; /* default theme $colorSplash */
+$havelockBlue: #507dd7;
```
JoshuaLuckers:
```suggestion
$havelockBlue: #507DD7;
```

Improve code quality by using capitalization consistently.

christianseel:
I like the success/error styles! However the logo looks way too big for me now. I'm also not sure about the inputs as they look somehow disabled now?

Is the (button) green the same as the current primary color of the manager interface? Because at the time when we were working on the login, the previous color was used as primary in the manager.

christianseel:
Ah, just found https://github.com/modxcms/revolution/pull/14306 which seems to be related and which streamlines the button colors 👍🏻

_build/templates/default/sass/_colors-and-vars.scss
Jako: Are there editor styles available for this?

Ibochkarev:
> I'm also not sure about the inputs as they look somehow disabled now?

@christianseel I ask you to test these changes. Since there is always the possibility to make adjustments.

I will fix the logo. 

Thanks for the feedback. I tried to do in the same style.

_build/templates/default/sass/_colors-and-vars.scss
Ibochkarev: > Are there editor styles available for this?

@Jako in 137 line

Jako: I meant by this file: https://github.com/modxcms/revolution/blob/3.x/.editorconfig

Jako: In PhpStorm these styles could be predefined in https://www.jetbrains.com/help/phpstorm/settings-code-style-scss.html. Maybe there is a way to put those settings in the .editorconfig.

Ibochkarev: Open #14311

alroniks (changes requested):
After the quick look:
- New blue color seems to bold in comparing with old one. I'd prefer to return it.
- As some font properties were changed it's strongly required test it on Latin letters too (screenshots).
- I don't agree with replacing focus stroke by shadows. Availability is completely lost.

JoshuaLuckers (changes requested):
Keep the key points in the original plan in mind when changing the login screen:
- [MODX 3 Plan - Login screen](https://modx3.org/plan/login-screen)
- [Another example of the login screen](https://modx3.org/updates/week-7)

Mark-H:
@JoshuaLuckers We honestly don't have to stick to what was posted on modx3.org a year ago if a better proposal comes around. If you have specific points of feedback, please elaborate.

JoshuaLuckers:
I agree that we don't have to stick to something that's proposed a long time ago. We do have to keep in mind the designs that where proposed. Deviations from this design should have proper argumentation (that's what I'm missing right now).

Mark-H:
I agree with @Alroniks on the focus status (the inset box shadow is a little too subtle), and I'd probably take out the letter-spacing on the ghost buttons/password reset link. Makes it look a little stretched and inconsistent with other text on latin text:

schermafbeelding 2019-01-29 om 14 36 15

Those two things changed would earn my ✔️ . 

I don't really have an opinion on the blue color. I think it's an interesting shade of blue, but if it's not used anywhere else but the login page, it may make sense to restore that to something more consistent. 

@JoshuaLuckers: Design is very subjective of course, but in my opinion the screenshots made it clear that it adds more polish to the design, which is more than enough argumentation in my book ;)

alroniks (approved):
Now it looks much better! 👍

JoshuaLuckers (approved):

Ibochkarev:
The last word is yours @Mark-H  =)

Mark-H (approved):

PR #14318, opened 2019-01-31
Clarify the description of adding icons to templates

Ibochkarev:
### What does it do?
Clarify the description of adding icons to templates

### Why is it needed?
The user was initially unclear what icons are used and in what form they should be indicated in the field for entering Icons. This PR improves the description.

### Related issue(s)/PR(s)
NA


core/lexicon/en/template.inc.php
```diff
@@ -33,7 +33,7 @@
 $_lang['template_err_remove'] = 'An error occurred while removing the template.';
 $_lang['template_err_save'] = 'An error occurred while saving the template.';
 $_lang['template_icon'] = 'Icon';
-$_lang['template_icon_description'] = 'Optional. A custom icon class for all resources with this template.';
+$_lang['template_icon_description'] = 'Optional. A custom Font Awesome (Free) icon class for all resources with this template. For example: icon-home or fa-home.';
```
Jako: `css icon class` instead of `Font Awesome (free) icon class`?

Ibochkarev: I am change.

Ruslan-Aleev: @Jako Without clarifying Font Awesome, it’s not obvious which class to use, especially for new users.

Ibochkarev: @Jako I agree with Ruslan. This description was born during the discussion.

Jako: Then write `(i.e. a Font Awesome class)`, since that part depends on the css of the manager.

JoshuaLuckers:
```suggestion
$_lang['template_icon_description'] = 'Optional. A custom CSS icon class for all resources with this template. In example a Font Awesome (free) class such as icon-home or fa-home.';
```

JoshuaLuckers (dismissed):

Jako (dismissed):

core/lexicon/en/template.inc.php
```diff
@@ -33,7 +33,7 @@
 $_lang['template_err_remove'] = 'An error occurred while removing the template.';
 $_lang['template_err_save'] = 'An error occurred while saving the template.';
 $_lang['template_icon'] = 'Icon';
-$_lang['template_icon_description'] = 'Optional. A custom icon class for all resources with this template.';
+$_lang['template_icon_description'] = 'Optional. A custom CSS icon class for all resources with this template. In example a Font Awesome (free) class such as icon-home or fa-home.';
```
Mark-H: I think "In example" should be "For example".

JoshuaLuckers (approved):

Mark-H (approved):

PR #14319, opened 2019-01-31
Refactored tag input renreder to fix rendering with empty options list

alroniks:
### What does it do?
It fixes the issue with displaying tag input component (TV) when options list set as empty.

### Why is it needed?
It needed to avoid glitches that described in the issue.

### Related issue(s)/PR(s)
#14177

### DIFF
--- a/core/model/modx/processors/element/tv/renders/mgr/input/tag.class.php
+++ b/core/model/modx/processors/element/tv/renders/mgr/input/tag.class.php
@@ -32,11 +32,11 @@ class modTemplateVarInputRenderTag extends modTemplateVarInputRender
             if (count($option) === 1) {
                 $option[] = $option[0];
             }
-            list($label, $meaning) = $option;
+            list($inputOptionText, $inputOptionValue) = $option;
             $options[] = array(
-                'value' => htmlspecialchars($meaning, ENT_COMPAT, 'UTF-8'),
-                'text' => htmlspecialchars($label,ENT_COMPAT,'UTF-8'),
-                'checked' => in_array($meaning, $value, false),
+                'value' => htmlspecialchars($inputOptionValue, ENT_COMPAT, 'UTF-8'),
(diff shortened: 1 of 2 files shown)


core/model/modx/processors/element/tv/renders/mgr/input/tag.class.php
```diff
@@ -13,42 +13,43 @@
...
+            $option = is_array($option) ? $option : explode('==', $option);
+            if (count($option) === 1) {
+                $option[] = $option[0];
             }
-
-            $opts[] = array(
-                'value' => htmlspecialchars($itemValue,ENT_COMPAT,'UTF-8'),
-                'text' => htmlspecialchars($item,ENT_COMPAT,'UTF-8'),
-                'checked' => $checked,
+            list($label, $meaning) = $option;
```
Jako: Why not `$value` and `$text`? The variables are only used for placeholder values afterwards and could use their name.

alroniks: `$value` already comes as the function parameter, I can not redefine it in the loop. About text yes, makes sense but label and meaning look more like a pair that text and meaning.

Jako: Oh, I did not see that.

Meaning is somehow the wrong word.  What about `$tagText` and `$tagValue`?

JoshuaLuckers: `$inputOptionText`and `$inputOptionValue`?

Mark-H (dismissed):

alroniks:
Hi @Mark-H I've improved wording in variables a bit, please approve it again when you'll able to.

JoshuaLuckers (approved):

Mark-H (approved):

Issue #14273, opened 2019-01-14
[enhancement] System settings: change default value


Ruslan-Aleev:
### Summary
In the System Settings, some default settings are chosen, in my opinion, incorrectly.

Friendly URL:
_FURL Alias Character Restriction Method (friendly_alias_restrict_chars) - pattern_
**Correctly put "alphanumeric"**, because It happens that in the current variant can stay incorrect characters in url, but with "alphanumeric" there is no such problem and this option will is suitable for most users.

Gateway:
_Strict Request Method (request_method_strict) - no_
**It is better to put "Yes"**, because in the current version, if FURL is enabled, then the url opens with both index.php and sef-url, which is bad for search engines.

Back-end Manager:
_Manager Favicon URL (manager_favicon_url) - empty_
**Why not put the favicon MODX by default?** Firstly, without a favicon, the admin panel looks unfinished and it is more convenient to navigate through the browser tabs with a favicon. Secondly, if the user changes this setting, he would in any case change it, i.e. time is not added for him.

_Resource Tree Node Fallback Field - pagetitle_
The settings of the "Resource Tree Node Field" and "Resource Tree Node Fallback Field" have the same value - "pagetitle", **but it is more logical for the "Resource Tree Node Fallback Field" to put another field, for example, "alias".**

_Manager Date Format (manager_date_format), Manager Time Format (manager_time_format) and Week start (manager_week_start)_
These settings are more correct for USA, but for most countries the time and date format is different.

Site:
_Automatic Template Assignment (automatic_template_assignment) - parent_
**Correct to put "sibling"**, because more often, resources with the same template are siblings pages, for example, news or articles.

_Published default (publish_default) - no_
In my opinion, resources are published more often than not published. **Better to set "yes"**.

_Hide From Menus Default (hidemenu_default) - no_
Same comment as "Published default".

**Perhaps some settings should be left as is, or change some other, in any case, we need a discussion.**

rthrash:
I'd personally be OK with the new default permission suggestions. I know I've been caught on more than one occasion wondering where the heck the content went (to go back and remember to, you know, actually publish it…).

ant-tar:
Please add label "needs-docs" - I'll update doc settings, default values

opengeek:
We shouldn't be changing documentation based on issues. Only based on actual changes made in PRs once they are accepted/merged.

ant-tar:
Sure @opengeek I will be happy with this label after accepting the changes :)

[...]

Issue #13876, opened 2018-04-23
[in progress] Making a clean install more complete by adding some settings in the setup


gpsietzema:
### Summary
*Please note: this is a work in progress*

A clean install of MODX is currently too clean. It's missing some very essential things, which will make the install more tailored to your needs. Therefore we need to enhance the current setup page with some more options, maybe even with an 'Make MODX more personal' button to expand these settings.

The following settings are blank or default to something very specific:

| Key | What to do with it? |
|---|---|
|date_timezone|Is currently not set, which might mess up certain timestamps.|
|cultureKey|Should be set in setup|
|fe_editor_lang|Should be set in setup|
|locale|Is currently empty and should be set in setup|
|manager_week_start|Is set to 0 (sunday) right now. Should also be able to set this to 1 (monday)|
|manager_data_format|Is set to US-version, should be setup in setup|
|manager_time_format|Is set to US-version am/pm, should be changeable to 24h format|

The following System Settings should have a different default setting:

| Key | Change |
|---|---|
| preserve_menuindex |Currently True, should be False.|
| resource_tree_node_name_fallback |Should be longtitle or menutitle|
| site_name| Is currently 'MODX Revolution', while this should really be the first thing to set for your site|
| site_status| Is set to True, while you should be able to set it at start. Websites usually start unpublished|

alroniks:
FURLs used everywhere now and it also should be possible to turn on it during the setup (with notices about required changes in nginx/apache config files).

JoshuaLuckers:
I would also love to see some kind of default error page. I think a lot of (new) users don't know that it's a setting. Maybe this should be a separate issue.

Jako:
@JoshuaLuckers Good Idea: A default error_page and maybe the default unauthorized_page should be created during the setup.

Jako:
The ht.access file should be renamed automatically by an option. A nginx config file should be prepared, if thats possible.

meshkov:
url_scheme,  because #13269 the setting server_protocol does not do anything

Ruslan-Aleev:
Related with https://github.com/modxcms/revolution/issues/14238 and https://github.com/modxcms/revolution/issues/14273

Ruslan-Aleev:
It seems to me that you should not add all the fields to the installation process.
It would be more correct to change the default settings, for example, for days and time, since most countries use others.

JoshuaLuckers:
[...]

PR #14322, opened 2019-02-02
System settings: change default value

Ibochkarev:
### What does it do?
- Automatic Template Assignment (`automatic_template_assignment`) - `parent` changed to `sibling`, Because more often, resources with the same template are siblings pages, for example, news or articles.
- FURL Alias Character Restriction Method (`friendly_alias_restrict_chars`) - `pattern` changed to `alphanumeric`. Because It happens that in the current variant can stay incorrect characters in url, but with "alphanumeric" there is no such problem and this option will is suitable for most users.
- Manager Time Format `manager_time_format` Is set to US-version `am/pm`, changed to `24h format`
- Week start `manager_week_start` Is set to `0 (sunday)` right now, changed to to `1 (monday)`
- Published default (`publish_default`) - `no`, changed to `yes`. Resources are published more often than not published. 
- Resource Tree Node Fallback Field `resource_tree_node_name_fallback` Is set to `pagetitle`, changed to `alias` 
- Preserve Menu Index When Duplicating Resources `preserve_menuindex`  Currently `True`, Changed to `False`
- Manager Favicon URL (`manager_favicon_url`) - `empty`. Added file favicon.ico and registered path `/manager/templates/default/images/favicon.ico`

### Why is it needed?
A clean install of MODX is currently too clean. It's missing some very essential things, which will make the install more tailored to our needs.

### Related issue(s)/PR(s)
#14273 #13876


ghost:
I'm pretty sure that each enhancement should have its own PR.

Ibochkarev:
@sottwell What is specified in #13876  - will be decided by a separate PR when we rewrite the installation page.

Here I combined, only that which can already be specified.

JoshuaLuckers:
I agree with @sottwell that some changes are better off in separate PR’s (I.e. the favicon).

JoshuaLuckers:
To clarify: separate PR's for the different enhancements makes it easier to accept and test them. It also prevents stalling enhancements that don't need discussion/changes.

Ibochkarev:
This PR will be divided into several. For ease of discussion and decision on confirmation.

PR #14323, opened 2019-02-03
Fixed changing language settings during installation / upgrade

Ruslan-Aleev:
### What does it do?
When installing and updating, no matter how the language was selected - in the editor settings (the "fe_editor_lang" key) was always "en". Added a change of settings depending on the selected.
Perhaps in MODX3 this setting will not be.

Added check of "cultureKey" settings when updating.

### Related issue(s)/PR(s)
NA

### DIFF
--- a/setup/includes/new.install.php
+++ b/setup/includes/new.install.php
@@ -234,7 +234,7 @@ if ($usemb) {
     $setting->save();
 }
 
-/* if language != en, set cultureKey, manager_language, manager_lang_attribute to it */
+/* if language != en, set cultureKey, fe_editor_lang, manager_lang_attribute, manager_language to it */
 $language = $settings->get('language','en');
 if ($language != 'en') {
     /* cultureKey */
@@ -253,14 +253,14 @@ if ($language != 'en') {
     $setting->set('value',$language);
     $setting->save();
 
-    /* manager_language */
+    /* fe_editor_lang */
     $setting = $modx->getObject('modSystemSetting',array(
-        'key' => 'manager_language',
+        'key' => 'fe_editor_lang',
     ));
     if (!$setting) {
         $setting = $modx->newObject('modSystemSetting');
         $setting->fromArray(array(
-            'key' => 'manager_language',
+            'key' => 'fe_editor_lang',
             'namespace' => 'core',
             'xtype' => 'textfield',
             'area' => 'language',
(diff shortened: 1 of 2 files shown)


Jako:
Since this is not really a bug, it should be handled in 3.x

Ruslan-Aleev:
Why not a bug? So I chose Russian, installed MODX, all languages changed, except for the editor’s language, see the picture below.

lang_select

Why choose a language during installation, if then need to configure it again?

Ruslan-Aleev:
@Mark-H What do you think? Is this not a bug when the functional behaves differently than expected?

Jako:
Where is that fe_editor_lang used? https://github.com/modxcms/revolution/search?q=fe_editor_lang&unscoped_q=fe_editor_lang

Thats the reason, why it is not a bug IMHO.

Ruslan-Aleev:
Yes, I agree that this setting is useless, which is why I wrote that in MODX3 it might not be.
The logic is as follows:
- Bring the behavior of this setting to normal = fix in 2.7.1
- Delete unused settings along with others = improvement in MODX3

Maybe some components use it?

Jako:
Maybe it is an old Evo setting: https://github.com/modxcms/evolution/search?q=fe_editor_lang

I don't think that it is used wide.

alroniks:
This setting used for configuring TinyMCE in MIGX. So, it would be nice to mark it as deprecated and delete later. 

Here a lot of such examples https://github.com/search?l=PHP&p=2&q=fe_editor_lang&type=Code

alroniks:
So, based on described behavior it's definitely a bug and should be fixed. But, it's not a regression bug of 2.7.0 so probably it should go to 3.x.

Jako:
> This setting used for configuring TinyMCE in MIGX. So, it would be nice to mark it as deprecated and delete later.

It is not used in MIGX, if this search is right: https://github.com/Bruno17/MIGX/search?q=fe_editor_lang

Jako:
Most of the search results are system settings in site backups in a public repository.

Some are part of the TinyMCE package plugin: https://github.com/modxcms/TinyMCE/blob/production/core/components/tinymce/tinymce.plugin.php 

TinyMCE RTE does not use that setting:
https://github.com/modxcms/tinymce-rte/search?q=fe_editor_lang

devlanda:
According to docs it is a deprecated setting: https://docs.modx.com/revolution/2.x/administering-your-site/settings/system-settings/fe_editor_lang

For backward compatibility e.g. TinyMCE it is not necessary, because it will look at system settings cultureKey and manager_language, as long as fe_editor_lang is not set.
https://github.com/modxcms/TinyMCE/blob/6eef255292f486424fad2ad113e684243e16475a/core/components/tinymce/tinymce.plugin.php#L34

Jako:
So this could be closed, since the setting is already deprecated? It does not make sense to patch this for just one version.

Mark-H:
Agreed, would be nice to get that setting removed.

Issue #14238, opened 2019-01-06
[enhancement] MODX setup: changing fields and their order


Ruslan-Aleev:
### Summary
1) When installing MODX, in the db setting, in the current version the "Database name" and "Database login name" fields are separated by the "Database password" field, but very often these fields have the same value and they are filled in via copy / paste, they must be nearby.
**The correct order, in my opinion, should be:**
...
"Database name"
"Database login name"
"Database password"
...

p.s. In the /core/config/config.inc.php file, need also place variables next to each other ($dbase, $database_user).

2) When setting up an administrator, the "Admin username" and "Admin password" fields are separated by the "Admin email" field, but the "Username - Password" combination should always be nearby, this is already standard :)
**The correct order, in my opinion, should be:**
...
"Admin email"
"Admin username"
"Admin password"
"Confirm password"
...

3) Add the **"Site Settings" item** to the MODX installation and add the **"Site Status"** field (enabled or not), uncheck the box and you can not worry about indexing the unfinished site.
Also add the **"Site name"** field, a useful setting, is found in other CMS, and it is logical to set the Site name when installing the site.
I am not sure, but it also makes sense to add the "Site email" field, which is now in the "System settings" - the key "emailsender" and is equal to "Admin email", but in fact these should be different fields.

Perhaps it makes sense to add more important fields, or not to add them at all, but the order of the fields (points 1 and 2) needs to be changed precisely.
**Anyway, we need a discussion.**

_Upcoming MODX3 installer design:_
![3-connect](https://user-images.githubusercontent.com/12523676/50730618-eeac8500-116a-11e9-8d7c-7366be17ce55.png)

Mark-H:
Where'd you get that screenshot from?

Ruslan-Aleev:
@Mark-H :) I found it here - https://modx3.org/updates/week-18-june-11-17-2018

Mark-H:
Thanks. Someone should implement that design... maybe I can find some time in the next few weeks. :)

Ruslan-Aleev:
It would be great, but, besides design in MODX3, if the item "Site settings" will be added, then need to select useful settings, in addition to "Site status", "Site name". Need a discussion.

Ibochkarev:
> Thanks. Someone should implement that design... maybe I can find some time in the next few weeks. :)
 
@Mark-H If you provide mockups, I can join

Ibochkarev:
[...]

PR #14325, opened 2019-02-03
Change the Manager date and time format after a clean installation.

Ibochkarev:
### What does it do?
- Changes the Manager Time Format `manager_time_format` values ​​to 24 hour format after a clean installation.
- Changes the Manager Date Format `manager_date_format` values ​​to `d-m-Y` format after a clean installation.

### Why is it needed?
To change a specific setting. In most cases, the 24 hour format and format date `d-m-Y` is used.

### Related issue(s)/PR(s)
#13876

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -963,7 +963,7 @@ $settings['mail_smtp_user']->fromArray(array (
 $settings['manager_date_format']= $xpdo->newObject('modSystemSetting');
 $settings['manager_date_format']->fromArray(array (
   'key' => 'manager_date_format',
-  'value' => 'Y-m-d',
+  'value' => 'd-m-Y',
   'xtype' => 'textfield',
   'namespace' => 'core',
   'area' => 'manager',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1008,7 +1008,7 @@ $settings['manager_js_document_root']->fromArray(array (
 $settings['manager_time_format']= $xpdo->newObject('modSystemSetting');
 $settings['manager_time_format']->fromArray(array (
   'key' => 'manager_time_format',
-  'value' => 'g:i a',
+  'value' => 'H:i',
   'xtype' => 'textfield',
   'namespace' => 'core',
   'area' => 'manager',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -963,7 +963,7 @@ $settings['mail_smtp_user']->fromArray(array (
 $settings['manager_date_format']= $xpdo->newObject('modSystemSetting');
 $settings['manager_date_format']->fromArray(array (
   'key' => 'manager_date_format',
-  'value' => 'd-m-Y',
+  'value' => 'Y-m-d',
   'xtype' => 'textfield',
   'namespace' => 'core',
   'area' => 'manager',


opengeek (commented):
I don't understand why you would change the default date format to one that is not internationally accepted. The time format I can partially understand, but even that is debatable as a default.

Ruslan-Aleev:
@opengeek This format "d-m-Y" is popular - https://en.wikipedia.org/wiki/Calendar_date#Gregorian,_day-month-year_(DMY), and certainly more popular than "Y-m-d"
Time format - https://en.wikipedia.org/wiki/24-hour_clock

Ruslan-Aleev:
The Russian-language wikipedia has a scheme, see below.
In this case, Wikipedia can be trusted, in my opinion.

date_info

opengeek:
> This format "d-m-Y" is popular - https://en.wikipedia.org/wiki/Calendar_date#Gregorian,_day-month-year_(DMY), and certainly more popular than "Y-m-d"

The format may be popular across the world, but it is not used in, for example, the United States at all. It is, in fact, confusing to US users. That is why we use the more agnostic format by default.

Ruslan-Aleev:
Strange statement. Users from the USA (who use "m-d-Y") are not confused by the "Y-m-d" format, but at the same time confused by the "d-m-Y" format, which is spread all over the world?
Shouldn't we consider the majority of users?

opengeek:
It's about ambiguity. Consider 01-05-2019. Is this January 5, 2019 (what it would mean in the US) or June 1, 2019 (what it would mean with this format)? This format is ambiguous. We decided to avoid ambiguity with the default value because this can be interpreted differently in different locales. This was a conscious choice.

Ruslan-Aleev:
Yes, you are right, we did not think about it. Perhaps we should change the format to "d-M-Y"? This will remove ambiguity and will not confuse majority of users. Or leave as is.
This date is used only in the manager panel, in the fields of the elements the date is saved in the standard way.

Mark-H:
I agree with Jason that the default date format should stay at Y-m-d, that's the simplest and most unambiguous form that anyone around the world can understand. I'm from a country that does d/m/Y but am routinely confused online with any format that isn't Y-m-d, because you need context to figure out if you're looking at d-m or m-d.

Writing out the month with d M Y could be an option to solve the ambiguity. But I think the manager mostly shows dates in grids, where numbers may be easier to compare.

Changing the time format to 24hr instead of am/pm I do agree with. If you haven't grown up with am/pm that can be quite confusing especially around noon/midnight, so 24hr is a more sane default.

Jako:
Maybe we should introduce some refinements after setup by switching those settings with locale based combinations. The same could be done with the manager mail texts/subjects (forgot password etc.).

JoshuaLuckers:
I would love to see a more advanced setup that makes those settings customisable in the setup process. Maybe we can create a issue to discuss this.

Ruslan-Aleev:
@JoshuaLuckers I offered to make some in the installer - https://github.com/modxcms/revolution/issues/14238
But here we need not to overdo it, if there are a lot of fields and until you put everything in - you will change your mind to install MODX :)

Ibochkarev:
In the last commit I left only the time format change for the 24 hour format.

I propose to make this change.

And when we make a new installation page, it will be possible to combine it with the previously described proposals for removing additional settings with a clean installation and add new features and additional settings there.

alroniks:
> But here we need not to overdo it, if there are a lot of fields and until you put everything in - you will change your mind to install MODX :)

Not need to include all of them to the installer and also it makes sense to set some of them automatically based in the current locale in the browser, so a user will have to just agree with settings or change any of them to fit their needs. And this step should be optional to allow skipping it and to allow installing via command line.

Mark-H (approved):

alroniks (approved):

PR #14327, opened 2019-02-03
Change the default request_method_strict and friendly_alias_restrict_chars

Ibochkarev:
### What does it do?
- FURL Alias Character Restriction Method (`friendly_alias_restrict_chars`) - `pattern` changed to `alphanumeric`. 
- Strict Request Method (`request_method_strict`) - `no` changed to `yes`. 

### Why is it needed?

1. FURL Alias Character Restriction Method (`friendly_alias_restrict_chars`). Because It happens that in the current variant can stay incorrect characters in url, but with `alphanumeric` there is no such problem and this option will is suitable for most users.
2. Strict Request Method (`request_method_strict`). Because in the current version, if FURL is enabled, then the url opens with both index.php and sef-url, which is bad for search engines.

### Related issue(s)/PR(s)
#14273

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1442,11 +1442,7 @@ $settings['request_controller']->fromArray(array (
 $settings['request_method_strict']= $xpdo->newObject('modSystemSetting');
 $settings['request_method_strict']->fromArray(array (
   'key' => 'request_method_strict',
-<<<<<<< HEAD
   'value' => '0',
-=======
-  'value' => 0,
->>>>>>> d8941327784a9af4817acc22ca24644da4d1d603
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'gateway',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1442,7 +1442,7 @@ $settings['request_controller']->fromArray(array (
 $settings['request_method_strict']= $xpdo->newObject('modSystemSetting');
 $settings['request_method_strict']->fromArray(array (
   'key' => 'request_method_strict',
-  'value' => '0',
+  'value' => true,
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (changes requested):
A few Unit Tests are failing because of these changes.

Ibochkarev:
> A few Unit Tests are failing because of these changes.
@JoshuaLuckers  Please help me with troubleshooting.

JoshuaLuckers:
This test fails: https://github.com/modxcms/revolution/blob/13c6b22b8abfeb8e3436eaaff6fa5e67041068f2/_build/test/Tests/Model/Request/modRequestTest.php#L102-L122

It's testing this method:
https://github.com/modxcms/revolution/blob/1e4a45c1b0e412c46fe4b8880a6e939ed222f23c/core/model/modx/modrequest.class.php#L144-L163

Caused by the new default setting for `request_method_strict`.

Jako:
Changing `friendly_alias_restrict_chars` is ok, but activating `request_method_strict` could cause infinite redirect loops in some situations (i.e. in CustomRequest). That part has to be checked.

Mark-H:
Past the feature/breaking change cut-off for alpha1, so pushing further.

JoshuaLuckers:
@Mark-H @Jako do we approve changing `friendly_alias_restrict_chars` from `pattern` to `alphanumeric`?

Jako:
I favor a better iconv based solution than the current. The current solution transliterates from unicode to unicode if it works the same as 2.x, it should transliterate from unicode to ascii.

JoshuaLuckers:
> I favor a better iconv based solution than the current. The current solution transliterates from unicode to unicode if it works the same as 2.x, it should transliterate from unicode to ascii.

How can we accomplish this?

Jako:
Maybe #13932 covers this. But I did not check a lot of transliteratons.

PR #14328, opened 2019-02-03
Change default automatic_template_assignment, tree node fallback, and preserve_menuindex

Ibochkarev:
### What does it do?
- Automatic Template Assignment (`automatic_template_assignment`) - `parent` changed to `sibling`, Because more often, resources with the same template are siblings pages, for example, news or articles.
- Resource Tree Node Fallback Field `resource_tree_node_name_fallback` Is set to `pagetitle`, changed to `alias`
- Preserve Menu Index When Duplicating Resources `preserve_menuindex` Currently `True`, Changed to `False`

### Why is it needed?
A clean install of MODX is currently too clean. It's missing some very essential things, which will make the install more tailored to our needs.

### Related issue(s)/PR(s)
#13876 #14273

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -117,7 +117,7 @@ $settings['automatic_alias']->fromArray(array (
 $settings['automatic_template_assignment']= $xpdo->newObject('modSystemSetting');
 $settings['automatic_template_assignment']->fromArray(array (
     'key' => 'automatic_template_assignment',
-    'value' => 'parent',
+    'value' => 'sibling',
     'xtype' => 'textfield',
     'namespace' => 'core',
     'area' => 'site',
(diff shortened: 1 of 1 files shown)


MrRoco (approved):

JoshuaLuckers (approved):

Ibochkarev:
@Jako For this change in system settings, is it necessary to write an upgrade script?

Jako:
That would be a breaking change and it should be better done after some user input. The install script has to query those changes. There is no mechanism for this at the moment.

OptimusCrime:
@Jako Can't an upgrade to 3.x justify breaking changes? I feel like it can. If this is documented properly, there is no reason we can not change old (and bad) behavior imo.

Jako:
Yes, but a breaking change in the system settings could be prevented with the right query page. All those settings could be set before by a special request and should not be changed without a question (except for security related settings).

Jako (approved):

Ibochkarev:
@Mark-H Please check this PR

Mark-H (approved):

Issue #14320, opened 2019-02-01
File with two dots in name can be uploaded, but not deleted


sebastian-marinescu:
### Summary
The MODX media-manager allows uploading of files with two dots ("**..**") in the filename, 
but deleting it is impossible.  

### Step to reproduce
Create a txt-file with the name `datei-mit_zwei..punkten.txt`. 
Upload it via media-manager. Try to delete it afterwards. 

### Observed behavior
Trying to delete it throws an error: 
![image](https://user-images.githubusercontent.com/3798871/52120349-4e883580-261c-11e9-88eb-f7a7538e7884.png)
(the error says: "Please provide a valid directory")

### Expected behavior
It should either not allow uploading such files in the first place, or if it does, make sure it can be deleted.

### Environment
MODX 2.7.0-pl

SnowCreative:
Same story for files uploaded with "?" or "&" in the file name. Isn't there some way we can prevent bad characters from being used in upload file names?

Ruslan-Aleev:
Possibly related with https://github.com/modxcms/revolution/issues/15271

sebastian-marinescu:
Just tested it in `v2.8.5` and `v3.0.2`. 

It still doesn't work. 
The original filename is: `datei-mit-2..-punkten.txt`

In `v2.8.5` it doesn't find the file and it looks like it strips both dots `..` from the filename. 
It also looks like MODX thinks it is a directory.
![image](https://github.com/modxcms/revolution/assets/3798871/c6d092e2-c133-4b22-9cda-f9a6df2331e0)

In `v3.0.2` it doesn't find the file and it looks like it strips one dot `.` from the filename. 
![image](https://github.com/modxcms/revolution/assets/3798871/8c26c1da-e8fb-4043-acf3-e1a7fe7d5251)

Could be related to https://github.com/modxcms/revolution/issues/16351

PR #14330, opened 2019-02-05
Change sanitize behavior for files with two dots in filename

tolanych:
### What does it do?
For files uploaded via media-manager name with two or more dots will sanitize as when creating file.
**UPD:**
Change behavior. Now sanitize path before filename and allow filename had two dots. Update related browser/file processors.

### Why is it needed?
To avoid the mistake like described in issue (cannot be removed).
And for consistency with behavior creating file.

### Related issue(s)/PR(s)
This fixes issue #14320

### DIFF
--- a/core/model/modx/processors/browser/file/remove.class.php
+++ b/core/model/modx/processors/browser/file/remove.class.php
@@ -33,12 +33,13 @@ class modBrowserFileRemoveProcessor extends modProcessor {
         if (empty($file)) {
             return $this->modx->error->failure($this->modx->lexicon('file_err_ns'));
         }
-        $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars(pathinfo($file, PATHINFO_DIRNAME)));
-        if (!empty($directory)) {
-            $directory .= DIRECTORY_SEPARATOR;
-        }
-        $name = htmlspecialchars(end(explode( DIRECTORY_SEPARATOR, $file )));
-        $path = $directory.$name;
+        $oldlocale = setlocale(LC_ALL, 0);
+        setlocale(LC_ALL,'C.UTF-8');
+        $pathinfo = pathinfo($file);
+        setlocale(LC_ALL,$oldlocale);
+        $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars($pathinfo['dirname']));
+        $name = htmlspecialchars($pathinfo['basename']);
+        $path = $directory.DIRECTORY_SEPARATOR.$name;
 
         $loaded = $this->getSource();
         if (!($this->source instanceof modMediaSource)) {
--- a/core/model/modx/processors/browser/file/remove.class.php
+++ b/core/model/modx/processors/browser/file/remove.class.php
@@ -33,10 +33,10 @@ class modBrowserFileRemoveProcessor extends modProcessor {
         if (empty($file)) {
             return $this->modx->error->failure($this->modx->lexicon('file_err_ns'));
         }
-        $oldlocale = setlocale(LC_ALL, 0);
-        setlocale(LC_ALL,'C.UTF-8');
         $pathinfo = pathinfo($file);
-        setlocale(LC_ALL,$oldlocale);
+        if ($pathinfo['dirname'].DIRECTORY_SEPARATOR.$pathinfo['basename'] != $file) {
+            $this->modx->log (modX::LOG_LEVEL_ERROR, 'Could not prepare the filepath ' . $file . '. Please set a valid UTF8 capable locale in the MODX system setting "locale".');
+        }
         $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars($pathinfo['dirname']));
         $name = htmlspecialchars($pathinfo['basename']);
         $path = $directory.DIRECTORY_SEPARATOR.$name;
--- a/core/model/modx/processors/browser/file/remove.class.php
+++ b/core/model/modx/processors/browser/file/remove.class.php
@@ -34,7 +34,10 @@ class modBrowserFileRemoveProcessor extends modProcessor {
             return $this->modx->error->failure($this->modx->lexicon('file_err_ns'));
         }
         $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars(pathinfo($file, PATHINFO_DIRNAME)));
-        $name = htmlspecialchars(pathinfo($file, PATHINFO_BASENAME));
+        if (!empty($directory)) {
+            $directory .= DIRECTORY_SEPARATOR;
+        }
+        $name = htmlspecialchars(end(explode( DIRECTORY_SEPARATOR, $file )));
         $path = $directory.$name;
 
         $loaded = $this->getSource();
@@ -44,7 +47,7 @@ class modBrowserFileRemoveProcessor extends modProcessor {
         if (!$this->source->checkPolicy('remove')) {
             return $this->failure($this->modx->lexicon('permission_denied'));
         }
-        $success = $this->source->removeObject($file);
+        $success = $this->source->removeObject($path);
 
         if (empty($success)) {
             $errors = $this->source->getErrors();
--- a/core/model/modx/processors/browser/file/remove.class.php
+++ b/core/model/modx/processors/browser/file/remove.class.php
@@ -33,7 +33,9 @@ class modBrowserFileRemoveProcessor extends modProcessor {
         if (empty($file)) {
             return $this->modx->error->failure($this->modx->lexicon('file_err_ns'));
         }
-        $file = preg_replace('/[\.]{2,}/', '', $file);
+        $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars(pathinfo($file, PATHINFO_DIRNAME)));
+        $name = htmlspecialchars(pathinfo($file, PATHINFO_BASENAME));
+        $path = $directory.$name;
 
         $loaded = $this->getSource();
         if (!($this->source instanceof modMediaSource)) {
--- a/core/model/modx/processors/browser/file/rename.class.php
+++ b/core/model/modx/processors/browser/file/rename.class.php
@@ -40,10 +40,21 @@ class modBrowserFileRenameProcessor extends modProcessor {
         }
 
         $oldFile = $this->getProperty('path');
-        $oldFile = preg_replace('/[\.]{2,}/', '', htmlspecialchars($oldFile));
-        $name = $this->getProperty('name');
-        $name = preg_replace('/[\.]{2,}/', '', htmlspecialchars($name));
-        $success = $this->source->renameObject($oldFile, $name);
+        $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars(pathinfo($oldFile, PATHINFO_DIRNAME)));
+        if (!empty($directory)) {
+            $directory .= DIRECTORY_SEPARATOR;
+        }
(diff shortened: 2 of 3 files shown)


core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -873,7 +873,7 @@ public function uploadObjectsToContainer($container,array $objects = array()) {
             }
 
             $newPath = $this->fileHandler->sanitizePath($file['name']);
-            $newPath = $directory->getPath().$newPath;
+            $newPath = ltrim(strip_tags(preg_replace('/[\.]{2,}/', '', htmlspecialchars($directory->getPath().$newPath))));
```
Jako: Why not move this in sanitizePath?

tolanych: I have fears to break other uses case fileHandler->sanitizePath.
This method might have been used before in external components and their processors

Jako: O.k. Then we maybe have to get the reason why two dots in the filename create that issue. 

Some sanitize method on the filename removes too much elsewhere. It just should remove the possibility to ascend in the path and strips those two dots instead of slashes/backslashes surrounded by two dots. But I did not searched for this ...

tolanych: Now I've realised that my decision (sanitize in method uploadObjectsToContainer) is not clear from the point of consistency.

This behavior I get from processor [browser/file/create](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/browser/file/create.class.php#L33)
This filter process directory path and filename.

In processor [browser/file/upload](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/browser/file/upload.class.php#L49) the filter in fact process only directory path, but not filename.

In processor [browser/file/remove](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/browser/file/remove.class.php#L36) preparing correct filepath without two dots and as result don't found file.

This had resulted in bug described in issue #14320.

As a solution propose for browser/file processors (create, upload, remove) to remove only two dots from path before filename and allow filename had two dots.
How are you viewing this?

sebastian-marinescu: If then it's also allowing to delete a file with two or multiple dots, than I'm alright with this 🚀

Ruslan-Aleev:
Files are renamed when they are loaded, but files with two dots in the name cannot be deleted if they are not loaded via the file manager (for example, via ftp).

core/model/modx/processors/browser/file/remove.class.php
```diff
@@ -33,7 +33,12 @@ public function process() {
         if (empty($file)) {
             return $this->modx->error->failure($this->modx->lexicon('file_err_ns'));
         }
-        $file = preg_replace('/[\.]{2,}/', '', $file);
+        $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars(pathinfo($file, PATHINFO_DIRNAME)));
+        if (!empty($directory)) {
+            $directory .= DIRECTORY_SEPARATOR;
+        }
+        $name = htmlspecialchars(end(explode( DIRECTORY_SEPARATOR, $file )));
```
Jako: You could use pathinfo for both parts.

core/model/modx/processors/browser/file/rename.class.php
```diff
@@ -40,10 +40,21 @@ public function process() {
...
+        }
+        $name = htmlspecialchars(end(explode( DIRECTORY_SEPARATOR, $oldFile )));
+        $oldFile = $directory.$name;
+
+        $newFile = $this->getProperty('name');
+        $directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars(pathinfo($newFile, PATHINFO_DIRNAME)));
+        if (!empty($directory)) {
+            $directory .= DIRECTORY_SEPARATOR;
+        }
+        $name = htmlspecialchars(end(explode( DIRECTORY_SEPARATOR, $newFile )));
```
Jako: Same here.

core/model/modx/processors/browser/file/remove.class.php
tolanych: I did so the first time.
Pathinfo with PATHINFO_BASENAME and PATHINFO_FILENAME  have trouble with filename with unciode symbols and more dots.
So after testing and catch more bugs I replace this on more simple - end->explode for filename.

e.g. http://codepad.org/73SaPqcF, first 4 characters in filename are loses.

Jako: pathinfo is locale aware, so you have to set an UTF8 capable locale before:

```
$oldlocale = setlocale(LC_ALL, 0);
setlocale(LC_ALL,'C.UTF-8');
$pathinfo = pathinfo($file);
setlocale(LC_ALL,$oldlocale);
```

Jako: See: http://codepad.org/nz0uth22

Jako: You have to check, if 'C.UTF-8' is available.

core/model/modx/processors/browser/file/rename.class.php
tolanych: @Jako,

my last fix with  [_setlocale_](https://github.com/modxcms/revolution/pull/14330/commits/aad01d99d61a886e51b4f6e8d1b8ebcd400c50b4) is not right yet?

core/model/modx/processors/browser/file/rename.class.php
```diff
@@ -40,10 +40,20 @@ public function process() {
         }
 
         $oldFile = $this->getProperty('path');
-        $oldFile = preg_replace('/[\.]{2,}/', '', htmlspecialchars($oldFile));
-        $name = $this->getProperty('name');
-        $name = preg_replace('/[\.]{2,}/', '', htmlspecialchars($name));
-        $success = $this->source->renameObject($oldFile, $name);
+        $oldlocale = setlocale(LC_ALL, 0);
+        setlocale(LC_ALL,'C.UTF-8');
```
Jako: Please check here, wether setlocale is successful. Otherwise an error should be logged, describing possible issues. Maybe the  'C.UTF-8' string should be configurable.

tolanych: This trouble solving if MODX Setting "locale" is correct encoding for current server environment.
Now I don't run setlocale manually, just log about encoding if function pathinfo return not what was expected

Ibochkarev:
@Jako Please check the changes.

JoshuaLuckers:
@cla-bot check

SnowCreative:
Shouldn't the sanitizer also remove characters like ampersand and pound sign? Those cause problems as well, and shouldn't be allowed in file names. I have clients who upload files like this all the time and then ask me why linking to these files on their websites doesn't work.
See also #13412

Ibochkarev:
## Summary

Fixes the inability to remove or rename files whose names contain two or more dots (e.g. `report.backup.2019.pdf`, `my..file.txt`). Previously, `preg_replace('/[\.]{2,}/', '', ...)` stripped consecutive dots from the entire path, so the stored path no longer matched the actual file on disk (especially for files uploaded via FTP or created outside the manager).

**Main changes:**
- **Create:** Remove dot-stripping from the filename; only the directory path is sanitized for `..`.
- **Remove:** Split path via `pathinfo()`, sanitize only the directory for `..`, keep basename intact; add locale warning when `pathinfo` returns unexpected results.
- **Rename:** Same split/sanitize approach for both old and new paths; add locale warning.

## Suggestions

### 1. Rebase for 3.x — different structure

The PR targets 2.x (`core/model/modx/processors/browser/file/`). In 3.x, the logic lives in:

- `core/src/Revolution/Processors/Browser/Browser.php` — `sanitize()` method
- `core/src/Revolution/Processors/Browser/File/Create.php`, `Remove.php`, `Rename.php`

3.x `Browser::sanitize()` uses `preg_replace('/\.(?![\w\-\~\s])/u', '', $file)`, which removes dots not followed by word chars. That also alters filenames like `file..name` → `filename`, so the same bug exists. The fix should be applied in 3.x by either:

- Extending `Browser` with `sanitizePath()` (path only) and `sanitizeFilename()` (no dot removal), or
- Splitting path/name in each processor and sanitizing only the path for `..`.

### 2. Rename processor — second parameter to `renameObject`

In the PR, `renameObject($oldFile, $newFile)` is called with `$newFile = $directory . DIRECTORY_SEPARATOR . $name`. If the form sends only the new filename (e.g. `file.newname.txt`), `pathinfo` yields `dirname=''`, so `$newFile` becomes `DIRECTORY_SEPARATOR . 'file.newname.txt'` (e.g. `/file.newname.txt`).

**Recommendation:** Confirm whether `renameObject` expects `(oldPath, newName)` or `(oldPath, newPath)`. If it expects only the new name, pass `$name` (basename) instead of `$newFile`

### 3. pathinfo and locale — validation logic

The check `$pathinfo['dirname'] . DIRECTORY_SEPARATOR . $pathinfo['basename'] != $file` is meant to detect locale/encoding issues with `pathinfo`. Edge cases:

- Paths with trailing slashes: `path/to/dir/` vs `path/to/dir`
- Windows vs Unix separators
- `pathinfo('')` returns `['dirname'=>'.','basename'=>'']`

**Recommendation:** Normalize the path (e.g. trim trailing slashes) before comparing. Consider `realpath` or a more robust path comparison. Document that the MODX `locale` setting should use a UTF-8–capable locale for correct handling of non-ASCII filenames.

### 4. pathinfo — empty dirname for bare filenames

For `pathinfo('file.txt')`, `dirname` is `'.'` and `basename` is `'file.txt'`. So `$directory = preg_replace('/[\.]{2,}/', '', htmlspecialchars('.'))` → `'.'`. Then `$path = '.' . DIRECTORY_SEPARATOR . 'file.txt'`. For remove, the input is usually a full path; for rename’s new name it may be just a filename. Ensure the logic handles both cases correctly.

### 5. Ampersand and pound sign (SnowCreative)

@SnowCreative noted [#13412](https://github.com/modxcms/revolution/issues/13412): filenames with `&` or `#` can break links. This PR does not address that. Consider tracking as a separate issue or extending sanitization in a follow-up.

---

### 6. setlocale check (Jako)

@Jako suggested checking whether `setlocale(LC_ALL, 'C.UTF-8')` succeeds before use. The current approach avoids `setlocale` and instead logs when `pathinfo` returns unexpected results. That is reasonable; ensure the log message is clear and actionable for admins.

## Assessment

### Strengths

- Fixes the core bug: files with multiple dots can be removed and renamed.
- Consistent behavior across create, remove, and rename.
- Path traversal protection kept: `..` is still removed from the directory part.
- Filenames like `report.backup.2019.pdf` are preserved.
- Logging when `pathinfo` fails helps diagnose locale/encoding issues.

### Concerns

- PR targets 2.x; 3.x needs an equivalent change.
- Possible mismatch with `renameObject` API (path vs name).
- pathinfo validation may have edge cases.
- Ampersand/pound in filenames remain a separate concern.

---

## Verdict

**Approve with minor changes** — recommend:

1. Verify `renameObject` expects `(oldPath, newName)` or `(oldPath, newPath)` and adjust the rename logic if needed.
2. Harden pathinfo validation (trailing slashes, empty input).
3. Plan a 3.x port with the same path/name split and sanitization rules.
4. Consider a follow-up for `&` and `#` in filenames per #13412.

The fix is targeted and addresses #14320. The path/name split is the right approach to keep path traversal protection while allowing dots in filenames.

Issue #14227, opened 2018-12-30
[enhancement] System Settings: Combine the categories "Friendly URL" and "Gateway"


Ruslan-Aleev:
### Summary
In the "System Settings" section, need to combine the two related categories "Friendly URL" and "Gateway" into one - **"Friendly URL & Gateway"**, since in the category "Gateway" of 4 settings 3 are associated with FURL.

![system_settings_2](https://user-images.githubusercontent.com/12523676/50547727-e202cb00-0c58-11e9-99d6-ed4d1b642a0b.png)

alroniks:
@Ruslan-Aleev Could you send one issue per request in the future, please? Because there are described two different issues related to the system settings that should be fixed and tested separately.

Ruslan-Aleev:
Despite the fact that Jason was against - https://github.com/modxcms/revolution/pull/14332#issuecomment-461099149
It still seems to me that these settings are better combined, for example, with the new name “Routing” (as Jason suggested), since duplication of pages in the search index is not a pleasant thing :)
And the settings are still connected, although they can be used separately.

PR #14332, opened 2019-02-05
Combine system settings the categories "Friendly URL" and "Gateway"

Ibochkarev:
### What does it do?
Combine system settings the categories "Friendly URL" and "Gateway" into one general group Friendly URL & Gateway

### Why is it needed?
In the category "Gateway" of 4 settings 3 are associated with FURL.

### Related issue(s)/PR(s)
#14227


core/lexicon/en/setting.inc.php
```diff
@@ -50,7 +49,7 @@
 $_lang['setting_update'] = 'Update Setting';
 $_lang['settings_after_install'] = 'As this is a new install, you are required to control these settings, and change any that you may wish to. After you\'ve controlled the settings, press \'Save\' to update the settings database.<br /><br />';
 $_lang['settings_desc'] = 'Here you can set general preferences and configuration settings for the MODX manager interface, as well as how your MODX site runs. Double-click on the value column for the setting you\'d like to edit to dynamically edit via the grid, or right-click on a setting for more options. You can also click the "+" sign for a description of the setting.';
-$_lang['settings_furls'] = 'Friendly URLs';
+$_lang['settings_furls_gateway'] = 'Friendly URL & Gateway';
```
alroniks:
```suggestion
$_lang['settings_furls_gateway'] = 'Friendly URLs & Gateway';
```
Plural variant more correct here.

alroniks (changes requested):

Mark-H:
Please correct me if I'm wrong, but I don't think upgrading an existing site will automatically adjust the categories for existing settings. That's okay (I don't think it's needed to write an upgrade script for that), but as the furls and gateway lexicons are removed, I think that may mean those categories will no longer be translated after applying this change.

To deal with that, I'd suggest not creating a new area (`furls_gateway`), but instead adjust the lexicon for `furls` and change the settings that are in `gateway` to `furls`. Then if you also leave in the `gateway` lexicon entry, that should make a better upgrade experience. 

Does that make sense?

opengeek:
I don't understand why this is needed?

opengeek:
The gateway settings are not all friendly URL related.

opengeek:
If anything, why not change it to "Routing" if we are going to merge them all together?

Ruslan-Aleev:
The settings in the "Gateway" section on 3/4 refer to the "Friendly URL" settings, i.e. they are related. And, for example, for seo it’s not correct, if urls like "/contacts.html" and "/index.php?id=5" are available on the site at the same time (configured in the gateway). Therefore, as it seems to me, it is right to combine them.
About the name "Routing" - it is not clear for a simple user.
In the name "Friendly URL & Gateway" there is a "Friendly URL" in the title, which is a little easier to understand :)

opengeek:
In that case, I'm going to say no to this one. I think they need to remain separate since the gateway is a different issue than Friendly URLs. Even though some of the gateway settings relate to Friendly URLs, they are still a separate group of settings that someone not using Friendly URLs might be looking for. I don't see this as a necessary change.

Ruslan-Aleev:
There is a need for SEO, although perhaps you are right. In this question the statistics of using or not using FURLs in MODX would be useful.

JoshuaLuckers:
Instead of doing work for changes that might or might not be approved it's better to create issues for changes like this where it can be discussed first and be implemented when reached a consensus.

Ibochkarev:
@JoshuaLuckers H. This improvement has been made by Issue #14227 . In it only Ivan answered. Therefore, I decided to make this PR.

**If he is not accepted, life will not end there =)**

Issue #14105, opened 2018-10-02
Stored XSS via document resources


AgelxNash:
Create new document with pagetitle
```</title><img src=# onerror=alert(1); /><!--```

XSS work in: 
 - Update
 - Quick edit
 - View manager logs

AgelxNash:
CVE-2018-20756

Issue #14104, opened 2018-10-02
Stored XSS: extended user fields


AgelxNash:
**Container name**
```
<img src=# onerror=alert(1); />
```

**Attribute name**
```
<img src=# onerror=alert(2); />
```

AgelxNash:
CVE-2018-20757

Issue #14103, opened 2018-10-02
Stored XSS: User Settings


AgelxNash:
![Alt text](https://monosnap.com/image/1SJ98Z3yuvdTJNDahdEkYKweQFjyD2.png)

AgelxNash:
CVE-2018-20758

Issue #14102, opened 2018-10-02
Stored XSS: User Photo


AgelxNash:
Stored XSS: The application is vulnerable to stored XSS.

Step to reproduce

Under Manage -> Users sources choose Create New User and enter the Xss Payload ```" onerror="alert(1)``` in the User Photo field and click on save.
The application renders the entered script and displays a pop-up whenever the page is being visited by the user.
Observed behavior
The application processes the html tags or scripts and it is getting stored in the database.

Expected behavior
It should not accept any scripts or html tags.

Environment
MODX version:MODX Revolution 2.6.5-pl

AgelxNash:
CVE-2018-20755

PR #14335, opened 2019-02-06
Fixes bunch of various XSS issues in the manager

AgelxNash:
### What does it do?
Fix XSS

### Why is it needed?
Security reason

### Related issue(s)/PR(s)
#14102 #14103 #14104 #14105

### DIFF
--- a/core/model/modx/moduser.class.php
+++ b/core/model/modx/moduser.class.php
@@ -907,7 +907,7 @@ class modUser extends modPrincipal {
         $path = $source->prepareSrcForThumb($this->Profile->photo);
 
         return $this->xpdo->getOption('connectors_url', null, MODX_CONNECTORS_URL)
-            . "system/phpthumb.php?zc=1&h={$height}&w={$width}&src={$path}";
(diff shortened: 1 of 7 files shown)


alroniks (approved):

Ibochkarev (approved):

Ibochkarev:
Related #14094

Mark-H (approved):
Thanks for working on this, @AgelxNash.

Ruslan-Aleev:
Some recent PR, affecting XSS, broke the generation of the tree, we need to test it more carefully.

resource_tree

Ibochkarev:
@AgelxNash I ask you once again to check the changes in this PR - I also confirm the incorrect operation after applying these changes.

In the resource tree, the code is added - which was inserted for verification and it does not disappear even after cleaning in the place where I inserted it for testing.

Overwriting files from the distribution kit fixes this bug.

Mark-H:
Sounds like a new issue for that is in order, that definitely needs to get fixed.

Ruslan-Aleev:
@Mark-H  The problem was found, here is - https://github.com/modxcms/revolution/pull/14335/files#diff-5f9a4ceb2f6f68d22416f98adcc4fdb5R250

Issue #14341, opened 2019-02-07
Stored XSS: Media Source


Ibochkarev:
### Summary
Stored XSS: Media Source

### Step to reproduce
Go to the Media menu, then media sources. Create or edit an existing media source and insert `<img src=# onerror=alert(1); />`

![screencast 2019-02-07 16-13-07](https://user-images.githubusercontent.com/2138260/52404785-0b363700-2af4-11e9-844a-fe6cf8507591.gif)

### Observed behavior
The application processes the html tags or scripts and it is getting stored in the database.

### Expected behavior
It should not accept any scripts or html tags.

### Environment
MODX Revolution 2.7.0-pl

Mark-H:
I think this is a duplicate of #14094 that was fixed in #14336 just 22 hours ago? Can you still reproduce in the latest 2.x?

Ibochkarev:
@Mark-H  
**Still found**:
In Content types - /manager/?a=system/contenttype
In Profile - /manager/?a=security/forms
In the panels - /manager/?a=system/dashboards
In access policy - /manager/?a=security/permission
In the parameter sets - /manager/?a=element/propertyset
In lexicons - /manager/?a=workspaces/lexicon
In namespaces - /manager/?a=workspaces/namespace

I will try to test the latest version of github and try to fix it myself on the weekend!

Mark-H:
I've created PR #14344 that applies XSS protection for all grid columns by default. There are still places where object names are inserted into a header dynamically, without running them through htmlEncode, so those will still need manual fixing.

Ibochkarev:
@Mark-H It is excellent!

Ibochkarev:
Fix to #14344

Issue #14094, opened 2018-09-26
Stored XSS


AnusyaAngamuthu:
Stored XSS: The application is vulnerable to stored XSS.

Step to reproduce

1. Under Media->Media sources choose Create New Media source and enter the Xss Payload <img src=xss onerror=alert("giraffee")> in the Name field and click on save.
2. The application renders the entered script and displays a pop-up whenever the page is being visited by the user.

Observed behavior
The application processes the html tags or scripts and it is getting stored in the database.

 Expected behavior
It should not accept any scripts or html tags.

 Environment
MODX version:MODX Revolution 2.6.5-pl

ghost:
Disagree. In the first place, this can only be done by a trusted Manager login. Access to Media Source management is controlled by ACLs, so lesser-trusted managers are easily excluded. 

In the second place, in a CMS such as MODX, it is a feature that such functions as assigning names or attributes of elements are highly flexible, allowing all sorts of programmatic assignments.

Mark-H:
The manager being a powerful tool should not be an excuse to "disagree" with vulnerabilities. It's a textbook example of XSS vulnerabilities, there's no discussion possible about that. 

I think we all agree it's not the most severe of issues, though. Otherwise @AnusyaAngamuthu would've reported this to the private mailing list.

I looked at a fix for this, but unfortunately it seems that it's not specific to the sources grid. There is nothing specific in `MODx.grid.Sources` that is making this vulnerability possible (e.g. a custom tpl that forgot a htmlent filter like we've seen elsewhere), meaning it goes up the tree to `MODx.grid.Grid` and even `Ext.grid.EditorGridPanel` from what I can tell. Perhaps even beyond that. That suggests practically every grid in the manager may be vulnerable for similar issues. 

There are [some suggested](https://stackoverflow.com/questions/6073037/how-to-escape-html-entities-in-grid) [solutions](https://www.sencha.com/forum/showthread.php?302339-HTML-encoding-grid-data-by-default) on the web to make grids encode by default, but they would likely cause unexpected side-effects in both core and third party components that do expect markup to be accepted (columns that include a link or buttons, for example, would likely also be encoded). Would perhaps be better to address that in 3.0, but that does mean we're leaving the door open to all sorts of XSS issues until then. 

(As a side note, the XSS also affects the Files tree. That should be patched separately.)

Omeryl:
@Mark-H 

[...]

Issue #14342, opened 2019-02-07
Stored XSS: Context


Ibochkarev:
### Summary
Stored XSS: Context

### Step to reproduce
Go to the Settings menu, then Contexts. Create or edit an existing context and insert `<img src=# onerror=alert(1); />`
![russian inn - revista rusa en espana russian inn - 2019-02-07 16-29-12](https://user-images.githubusercontent.com/2138260/52405533-b2679e00-2af5-11e9-9d96-c088c3bbce17.jpg)
![web111 russian inn - revista rusa en espana russian inn - 2019-02-07 16-28-47 1](https://user-images.githubusercontent.com/2138260/52405534-b2679e00-2af5-11e9-80b1-5f1edfdbc545.jpg)

### Observed behavior
The application processes the html tags or scripts and it is getting stored in the database.

### Expected behavior
It should not accept any scripts or html tags.

### Environment
MODX Revolution 2.7.0-pl

OptimusCrime:
Does it also print the alert, or is that stripped out?

Ibochkarev:
@OptimusCrime An example of execution can be seen here: #14341  Runs the same.

JoshuaLuckers:
~Fixed in 3.x by merging PR #14344~

JoshuaLuckers:
Sorry, PR #14344 didn't solve this particular case.

PR #14344, opened 2019-02-07
Protect MODx.grid.Grid against XSS vulnerabilities by default

Mark-H:
### What does it do?

When a MODx.grid.Grid/MODx.grid.LocalGrid implementation does not define a renderer for a column, automatically add the Ext.util.Format.htmlEncode renderer which protects the grid from XSS vulnerabilities. 

When a renderer is set, no action is taken, and the implementation is assumed to handle its own protections. This allows implementations to still do things like add links or buttons into grids. 

### Why is it needed?

The manager is very prone to XSS vulnerabilities. This has been discussed in different places, most notably #14094. In the past week, a lot of XSS-related issues [were raised](https://github.com/modxcms/revolution/issues?utf8=%E2%9C%93&q=is%3Aissue+xss) and [also fixed](https://github.com/modxcms/revolution/pulls?q=is%3Apr+xss), but those are all putting bandaids on the core problem, which is that ExtJS was not configured in a way to treat any value as potentially hostile user input.

This PR changes that and automatically escapes all values in grids, unless otherwise instructed. The PR also escapes values from a combobox that are rendered through an `editor` configuration. 

That's also why this targets 3.x - while security issues would typically go into a bugfix release, it is likely that compatibility issues in both core and third party extras will surface where those grids took advantage of the lack of encoding to insert links/buttons/other html into values. 

This does not fix all XSS issues in the manager, but it does stop a lot of them (like [the list here](https://github.com/modxcms/revolution/issues/14341#issuecomment-461382325)).  

### Related issue(s)/PR(s)

#14094

### DIFF
--- a/manager/assets/modext/widgets/core/modx.combo.js
+++ b/manager/assets/modext/widgets/core/modx.combo.js
@@ -163,12 +163,13 @@ Ext.extend(MODx.combo.ComboBox,Ext.form.ComboBox, {
 });
 Ext.reg('modx-combo',MODx.combo.ComboBox);
 
-Ext.util.Format.comboRenderer = function (combo,val) {
-    return function (v,md,rec,ri,ci,s) {
+Ext.util.Format.comboRenderer = function (combo, val) {
+    return function (v, md, rec, ri, ci, s) {
         if (!s) return v;
         if (!combo.findRecord) return v;
         var record = combo.findRecord(combo.valueField, v);
-        return record ? record.get(combo.displayField) : val;
+        val = record ? record.get(combo.displayField) : val;
+        return Ext.util.Format.htmlEncode(val);
     }
 };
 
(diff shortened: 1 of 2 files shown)


alroniks (dismissed):

Mark-H:
Sorry @Alroniks - found another potential vector that could be fixed globally (automatically rendered combos from an editor configuration), so added another commit after you approved.

JoshuaLuckers (approved):

alroniks (approved):

Ibochkarev (approved):

opengeek (approved):

Ibochkarev:
Related #14342

Ruslan-Aleev:
One of the recent PR links broke the generation of the tree, in my opinion, PR is related to XSS, we need to test it more carefully.

resource_tree

Ibochkarev:
@Mark-H Please see my comment https://github.com/modxcms/revolution/pull/14335#issuecomment-462100686

PR #14346, opened 2019-02-08
Add input prompt per setting to script

MrRoco:
Still uses default values but also able to receive prompt info to setup modx

### What does it do?
Add input prompt on running script to set custom values.

### Why is it needed?
So the file doesn't need to be changed for installing modx via this script thus being processed as changed in git

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/test/generateConfigs.sh
+++ b/_build/test/generateConfigs.sh
@@ -1,8 +1,29 @@
 #!/bin/bash
 
-DBNAME="revo_test"
-DBUSER="root"
-DBPASS=""
+read -p "Enter your database [revo_test]: " input
+DBNAME=${input:-revo_test}
+
+read -p "Enter your database user [root]: " input
+DBUSER=${input:-revo_test}
+
+read -p "Enter your database []: " input
+DBPASS=${input:-}
+
+read -p "Enter your host [unit.modx.com]: " input
+MDXHOST=${input:-unit.modx.com}
+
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
This is a work in progress.

JoshuaLuckers (dismissed):

_build/test/generateConfigs.sh
```diff
@@ -1,8 +1,49 @@
...
+    read -s -p "Enter your database password []: " input
+    DBPASS=${input}
+
+    read -p "Enter your host [unit.modx.com]: " input
+    MDXHOST=${input}
+
+    read -p "Enter your language [en]: " input
+    MDXLANG=${input}
+
+    read -p "Enter your modx admin userinput [admin]: " input
```
Mark-H: "userinput"

Ibochkarev:
@MrRoco Please make a change.

_build/test/generateConfigs.sh
JoshuaLuckers:
```suggestion
    read -p "Enter your modx admin user [admin]: " input
```

Ibochkarev:
@MrRoco Please confirm the changes from @JoshuaLuckers

JoshuaLuckers (approved):

opengeek (approved):

PR #14348, opened 2019-02-08
Using cltr/cmd and click will open the url in a new tab/window again

JoshuaLuckers:
### What does it do?
Using cltr/cmd and click will open the url in a new tab/window again for ExtJS elements that use `loadPage()` to open URLs

### Why is it needed?
The function to detect the ctrl/cmd + click did a wrong comparison causing the behaviour a lot of people use to misbehave.


Mark-H (approved):

Ruslan-Aleev (approved):
For Windows OS, it works as described :)
Although with the old version everything worked the same.

alroniks (approved):

opengeek (approved):

Issue #14167, opened 2018-11-29
Deleting resources - small bugs


Ruslan-Aleev:
1) You can restore only a children resource in the trash manager (it disappear from the list in trash manager), but the parent will not be restored. And if you click "Purge resource", then everything will be deleted, but the message will be only about the parent.

Need a connection between deleted resources - parent and children.
**It makes sense to activate and freeze the setting "Deleted" in children resources until the parent is deleted OR restore parent if children been restored OR change the parent to 0 and set the resource to unpublised, if it has been restored and its parent has been deleted.**

p.s. It is odd that you can create or restore a children resource if the parent has been deleted.

![del_bug_2](https://user-images.githubusercontent.com/12523676/49252938-cda39680-f43e-11e8-8bd7-0bc97548dc15.png)

2) If you delete a resource through its setting, then its child resources will be deleted, but at the same time they will be marked as not deleted in the tree, which can confuse users.

See https://github.com/modxcms/revolution/issues/14167#issuecomment-535401489

3) If we transfer the Home page or other system page ("404", "Site is not available") inside any resource, then by deleting this prent resource we will delete the system pages as well.
However, if they are located in the root, then it will not be possible to delete system pages, the check will work.

Jako:
~~I don't think that 2. could be solved nice. Purge them too could be too harsh. We need some 'search for orphan resources' processor otherwise, since those resources can't be found in the tree after purging.~~

~~Deleted parents and not deleted children could happen in other circumstances too.~~

Jako:
I just checked that on a current installation. The child resources are purged too, even if they are not deleted. They don't stay as orphaned resources. So the child resources 'just' have to be listed in the trash manager.

Ibochkarev:
@Jako Hello! Can you fix it

Ruslan-Aleev:
Now there is a problem in that if you delete a resource through its setting, then its child resources will be deleted, but at the same time they will be marked as not deleted in the tree, which can confuse users.

![deleted](https://user-images.githubusercontent.com/12523676/65671992-55573300-e059-11e9-961d-99a27fc73fa6.png)

Although if you delete via the context menu, the child resources will also be marked as deleted.

If you have a large resource tree and a lot of deleted resources in the trash, then you may simply not understand what will be deleted, which is a serious problem.

**I think that it is necessary to mark child resources deleted if the parent has been deleted** (as delete works through the context menu now)

A detailed discussion was here https://github.com/modxcms/revolution/pull/14350

Issue #14349, opened 2019-02-08
Delete checkbox does not change deletedby/deletedon


Jako:
### Summary
Delete checkbox does not change deletedby/deletedon.

### Step to reproduce/Observed behavior
Delete a resource with the deleted checkbox in the resource. The column deletedby and deletedon in the trash manager do not change. The opposite could only seen directly in the database. Deleting a resource with the context menu changes both fields.

### Expected behavior
The column deletedby and deletedon should change.

### Environment
MODX 2.7

Ruslan-Aleev:
Yes, I can reproduce the bug.

tolanych:
I'm looking at the present implementation. I have got options.
1) Should just update fields deletedby/deletedon...
2) Should duplicate all delete logic in update processor...
3) or just call delete/undelete resource processor inside method update processor? (possibly trouble with addLock, but can run delete processor in cleanup method after removeLock())

Jako:
I would run run the processor to delete the childs too, but make sure to handle $modx->error right. That part is a bit annoying when running a processor without using $this->success/$this->failure (or wherever the reset of $modx->error happens).

Issue #14158, opened 2018-11-28
[Feature] Hide deleted resources in manager tree since 2.7 has a trash management page


winformatic:
### Summary
It would be great to hide deleted resource elements in the tree view in the manager. Since 2.7 has a trash management page it is not necessary to see the red strike-through resources in the tree view anymore.

Maybe it could be activated via a system setting to hide deleted elements from the tree?

### Environment
MODX 2.7, php7.2

winformatic:
Found in the table _modx_site_content_ the field _show_in_tree_ and the getnodes.class.php file in core\model\modx\processors\resource\getnodes.class.php  interprets this field in the where statements.

If I switch the _show_in_tree_ value from **1** to **0**, the element is hidden in the treeview but still exists in the trash manager, which is exact the behavior I was asking for.

So we can simply use a plugin which handles the hiding:

**OnResourceDelete / OnResourceUndelete**
```
<?php
$currentResource = $modx->event->params['resource'];
if ($modx->event->name == 'OnResourceDelete') {
    $result = $currentResource->set('show_in_tree',0);
    $currentResource->save();
} elseif($modx->event->name == 'OnResourceUndelete') {
    $result = $currentResource->set('show_in_tree',1);
    $currentResource->save();
}
```

Jako:
Changing any resource field (except of deleted and deletedon/deletedby) is not the right way to hide those resources in the tree. There has to be a system setting and in the getnodes query deleted resources have to be excluded.

winformatic:
Hey Jako,
I totally agree with you. The above mentioned plugin has some difficulties with other extras that hide resources. It's a simple workaround and the feature should be implemented in the core.

Ruslan-Aleev:
At the Russian-speaking forum there was a discussion about the new format of deleting resources: _if there are a lot of them, then all the resources cannot be deleted at once, the pagination in the deleted resources manager interferes_.
Perhaps in addition to the setting "Hiding remote resources from the tree," **we can also make the setting "Use delete manager or not", which will enable the manager or allow you to delete resources in the old way.**

PR #14350, opened 2019-02-08
Show all resources to purge in the trash manager and change the tree options for deleted resources

Jako:
### What does it do?
- Add the not deleted children of deleted resource parents to the trash manager grid. 
- Don't show the create/quick create context menu  and the plus button for a deleted resource.

### Why is it needed?
- The not deleted children will be purged during purging the parents, so they have to be shown in the grid and in the list of purged resources.
- It makes no sense to create resources where they could be purged afterwards.

### Related issue(s)/PR(s)
#14167


JoshuaLuckers (approved):

Ruslan-Aleev:
@Jako Can you mark resources as deleted if their parents are deleted?
This would be visually useful, because it is immediately visible in the tree that the resource will be deleted and the user will not be confused.

Jako:
@Ruslan-Aleev It would be strange, when a file, that does not have the 'deleted' property enabled, is shown as deleted.

At the moment, all resources in a tree are marked as deleted, when the parent is deleted (maybe this works only with the context menu - but thats a different issue see #14349). So if you undelete afterwards a subtree, this is a user decision and should be visible in the tree.

Ruslan-Aleev:
> It would be strange, when a file, that does not have the 'deleted' property enabled, is shown as deleted.

Yes, it is true, but there is confusion. The user notes them that they are not deleted, but they are also displayed in the trash manager, isn’t there a  logical conflict?

> So if you undelete afterwards a subtree, this is a user decision and should be visible in the tree.

Such resources should not be deleted in the trash manager, but they will be deleted (you won't even restore them until the parent is restored).

It would be correct either to mark them deleted (for example, activate and block the deleted checkbox), or to change their parent.

Ruslan-Aleev:
As a result, we can undelete resource, but if its parent is deleted, the resource is also wil be deleted, so I suggest marking and freezing the deleted  checkbox for such a resource, which will immediately let the user know that something is wrong.
If we do not mark the deleted checkbox, then visually everything in the tree looks fine, but the resource will be deleted from the trash manager, and this is more serious situation.

Jako:
The additional purged resources could be marked special in that trash manager grid and in the popup. Would that be enough?

JoshuaLuckers:
I think it makes more sense to "undelete" the parent if you "undelete" a child.

Ruslan-Aleev:
@JoshuaLuckers Yes, I agree, in a issue I described several variants:
1) Activate and freeze the setting "Deleted" in children resources until the parent is deleted
2) Restore parent if children been restored
3) Change the parent to 0 and set the resource to unpublised, if it has been restored and its parent has been deleted.

Ruslan-Aleev:
@Jako I do not know, here we need additional discussion. I’m still in favor of marking resources as deleted (if the parent is deleted), because in fact, such resources will be deleted.
And if there are a lot of deleted resources, then in the trash manager you may not notice the marks (since pagination is used there).
By the way, it is not clear what will be in the tree if this function is implemented - https://github.com/modxcms/revolution/issues/14158

@Alroniks @Mark-H @opengeek Can you help with the discussion?

Jako:
@Ruslan-Aleev 
- Freeze the deleted setting is not an option, because sometimes you want to delete all but one item. This could be handled by deleting the parent and restore just that item and the parent.
- Restore the parent could be an option (or a warning, what could happen without restore the parent).
- Change the parent to 0 is not an option, since that can't be reverted easy.

Ruslan-Aleev:
Probably yes, it is better to restore more than necessary than to delete more than necessary :)
The option of restoring a deleted parent, in the case of restoring a deleted child, is the best option.

Jako:
I suggest to do get this merged into 2.7.1 and do the changes for restoring the parent(s) during restoring a child in 3.x, since that would be a 'behavior change'. I will create a new issue for this.

Jako:
This PR is still valid at the moment for the following list:

- if a resource is deleted by quickedit or by checkbox in the resource edit, the children are not deleted
- if the resource is deleted by changing the deleted field directly in the database, this is the same
- undeleted resources could be dragged into deleted parents

Then the undeleted children of a deleted resource would be purged in the trash manager without showing them there.

Ruslan-Aleev (approved):

opengeek (approved):

SnowCreative:
OK, I know I'm a bit late to the party here, but to me the most user-friendly option would be this:
If there is an undeleted resource under a deleted folder (either because someone dragged it there or manually undeleted it after the parent was deleted), when purging the parent, an alert would pop up like the following:
_The resource "[resource name]" is not marked for deletion but its parent will now be deleted. Delete this resource as well, or move it upwards in the tree?_
Then you'd have the option to either delete the undeleted resource, or move it up in the tree to the closest undeleted parent.

Ruslan-Aleev:
Pop-up windows have inconvenience: some users may not pay attention to this.
There is an issue - https://github.com/modxcms/revolution/issues/14368

Jako:
We will refactor the modResource model for deleted resources in MODX3: https://github.com/modxcms/revolution/issues/14368#issuecomment-463370037

JoshuaLuckers:
This PR is already merged and the ongoing conversation will end up in oblivion. It's better to move the conversation to an existing or new issue.

Issue #14352, opened 2019-02-09
Form Customization: Menu Title / Form Title


achterbahn:
### Summary
Actually the menu entry for FC is "Manager Customization". But the title of the form is "Form Customization". 

### Step to reproduce
Click manager menu "Manager Customization" and watch the title in the form opened.
![image](https://user-images.githubusercontent.com/1411919/52520360-3f7f3400-2c69-11e9-8e15-1ab635dbe4ec.png)

### Observed behavior
--

### Expected behavior
Identical naming 

### Environment
MODX => 2.7.0

opengeek:
Fixed via #14780

PR #14353, opened 2019-02-09
Change of the name and description in the top menu of Form Customization

Ibochkarev:
### What does it do?
Change of the name and description in the top menu of Form Customization

### Why is it needed?
Fix lexicon keys for name and description Form Customization in the top menu

### Related issue(s)/PR(s)
#14352


JoshuaLuckers (commented):
I think the lexicon `bespoke_manager` and `bespoke_manager_desc` can be deleted as well.

Ibochkarev:
@JoshuaLuckers Indeed, these lexicons are not used anywhere else.

JoshuaLuckers (dismissed):

Ibochkarev:
@markh please check.

Mark-H (changes requested):
The [primary key for installing menu items is the `text`](https://github.com/modxcms/revolution/blob/6bae073eddcdd3f340ac066b9dd77789f05caf8f/_build/transport.core.php#L192). I think that means that the current "Manager Customization" menu item would be left in-place, without its lexicon as that's also removed, alongside a new item for FC. That seems to call for an upgrade script that removes the old bespoke_manager menu item on upgrade.

I also think this should be targeting 3.x, not 2.x?

Ibochkarev:
This PR I close and create a new 3.x base

PR #14357, opened 2019-02-09
Fix Stored XSS: Context

Ibochkarev:
### What does it do?
Fix Stored XSS: Context

### Why is it needed?
Security reason

### Related issue(s)/PR(s)
#14342

### DIFF
--- a/manager/assets/modext/widgets/system/modx.grid.context.js
+++ b/manager/assets/modext/widgets/system/modx.grid.context.js
@@ -32,12 +32,14 @@ MODx.grid.Context = function(config) {
             ,width: 150
             ,sortable: true
             ,editor: { xtype: 'textfield' }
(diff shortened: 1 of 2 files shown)


Ibochkarev:
This PR in the current form corrects the error when you edit and save context data. In the list of contexts, the error is not fixed. I take for revision.

Mark-H:
All grids in 3.x automatically get the htmlEncode renderer applied, so I'm wondering if we can ignore reports/PRs like this targeting 2.x?

Mark-H (approved):
As there's already a PR for this I'll approve it, but I suggest no more precious development time is spent on adding these renderers to other grids when they are all automatically applied in 3.x.

alroniks (approved):

PR #14360, opened 2019-02-10
Minor changes in Media Source Browser

### LINKED ISSUE #14362: Detailed file description area always shows the first file of several selected
### Summary
Detailed file description area shows the first file of several selected

### Step to reproduce
Select multiple files in Media Source Browser

### Observed behavior
Description of the first file will be displayed. It isn't correct. (Note the order of the selected files)

_Do not pay attention to the "undefined"_ label

screenshot-revolution-2019 02 10-22-55-19

### Expected behavior
Description of the last selected file should be displayed (Note the order of the selected files)

screenshot-revolution-2019 02 10-22-57-03

### Environment
MODX 3.0.0-pl

### LINKED ISSUE #14363: Missing file size translitions in Media Source Browser
### Summary
File size, or more precisely, "bytes" and "KB" don't have lexicon entries

### Step to reproduce
Select the file in Media Source Browser

screenshot-revolution-2019 02 10-23-10-56

### Observed behavior
The words "bytes" and "KB" are displayed without translation

### Expected behavior
They must have a translation

### Environment
MODX 3.0.0-pl

GulomovCreative:
### What does it do?
1. Display last file description
2. Add and use some lexicon entries

### Why is it needed?
If user selects multiple files, then will be displayed a description of the first, not the last file. It isn't correct.

### Related issue(s)/PR(s)
This closes #14362 and closes #14363


Ibochkarev (approved):

Mark-H:
Is there an issue that describes what's being fixed here?

GulomovCreative:
> Is there an issue that describes what's being fixed here?

@Mark-H No, but i described the problem here

> If user selects multiple files, then will be displayed a description of the first, not the last file. It isn't correct.

### Step to reproduce

1. Select one file (Description of this file will be displayed. Correct)
2. Select one more file (Description of the first file will be displayed. It isn't correct.)

This PR solves this problem. Now the description of the last selected file will be displayed.

**+** Missing keys for "bytes" and "KB" and the **last_modified** key was used instead of the correct **file_last_modified**

If necessary, i can create the issue

GulomovCreative:
@Mark-H I created issues and noted them in description

Jako (approved):

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -275,8 +275,9 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
         var node = this.getSelectedNodes();
         var detailPanel = Ext.getCmp(this.config.ident+'-img-detail-panel').body;
         var okBtn = Ext.getCmp(this.ident+'-ok-btn');
+		var keys = Object.keys(node);
```
Mark-H, line 278: Just to note this, `Object.keys` is only available as of IE9. According to the [docs](https://docs.modx.com/revolution/2.x/getting-started/server-requirements), the expected browser support goes back to IE8 which [is apparently down to 0.14% yay](https://caniuse.com/usage-table). 

Perhaps 3.0 is a good time to bump the minimum to, say, IE11? :)

Mark-H (approved):

manager/assets/modext/widgets/media/modx.browser.js
JoshuaLuckers: Since this PR is marked as ready to be merged; are we going to bump the minimum IE version MODX supports in 3.x?  Maybe we should drop support for IE in 3.x

Issue #14365, opened 2019-02-11
To update the version of phpThumb to support WebP


Ibochkarev:
### Summary
To update the version of phpThumb to support WebP

### Why is it needed?
Have an up-to-date version of phpThumb with support for progressive webp image format.

### Suggested solution(s)
https://github.com/JamesHeinrich/phpThumb/commit/884abbdb1b045c16b58b9b3f0969d93e043a7b26

### Related issue(s)/PR(s)
#14355 #14356

Ibochkarev:
It can be closed. PR #14366 merged

PR #14356, opened 2019-02-09
Registration of WebP format in the Types of downloaded images setting

Ibochkarev:
### What does it do?
Registers the WebP format in the default settings to allow downloads.

### Why is it needed?
Registration of web format in the Types of downloaded images setting `upload_images `

### Related issue(s)/PR(s)
#14355

Issue #14355, opened 2019-02-09
Webp format and MODX is not friend yet.


iWatchYouFromAfar:
### Summary
As you probably know, the format of webp images is becoming more and more relevant. If you do not know, then here's an article on Wikipedia - https://en.wikipedia.org/wiki/WebP.

In MODX there are several problems that I discovered while trying to work with this format. What I want to tell below.

### Step to reproduce & Observed behavior
**Problem 1.**
Go to the settings and look for the key ** upload_images **. The description of the setting tells us that here we can specify the format of images that can be loaded into the _assets/images/_ folder. Specify the format webp and when you try to download the image in this folder, we get an error.

![1](https://user-images.githubusercontent.com/16403917/52524776-fe5f4200-2cb1-11e9-91e5-8c0748ee5254.gif)

**Improvement 1.**
After our unsuccessful attempt to upload an image, look for the setting with the ** upload_files ** key and specify the wepb format there. After that, we successfully upload the image. Perhaps it is necessary to register this format by default in future versions?

**Problem 2.**
As you can see, the preview image in the webp format is not available for us to see. This is sad, because MODX modern CMS!

![2](https://user-images.githubusercontent.com/16403917/52524785-0e772180-2cb2-11e9-9250-557828feb55e.gif)

### Observed behavior
**Problem 1.**
if I understood correctly, after specifying the format of the webp, we should easily upload the image to this directory?

**Problem 2.**
FTW We must see the preview!

### Environment
PHP 7.2
MODX 2.7.0

JoshuaLuckers:
WebP is not supported by one of the major browsers. And if I’m not mistaken phpthumb doesn’t support WebP either.

Ibochkarev:
> WebP is not supported by one of the major browsers. And if I’m not mistaken phpthumb doesn’t support WebP either.

At the moment, phpthumb does not support the format of the webp as this issue says 
https://github.com/JamesHeinrich/phpThumb/issues/107.

I added a PR that changes the setting and registers this format in the list of allowed downloads.

muzzwood:
It looks like there's a PR adding this to phpthumb as of 4 hours ago.
https://github.com/JamesHeinrich/phpThumb/commit/884abbdb1b045c16b58b9b3f0969d93e043a7b26

JoshuaLuckers:
MODX and `webp` are now friends! Thanks a lot to all involved.

PR #14366, opened 2019-02-11
Update phpThumb 1.7.15-201902101903

Ibochkarev:
### What does it do?
Update phpThumb 1.7.15-201902101903
Add support image format WebP
WebP is a google requirement for page optimization, a very useful feature.

### Why is it needed?
To solve the issue in #14365 #14355 #14356 

### Related issue(s)/PR(s)
#14365 #14355 #14356


opengeek (changes requested):
I believe we exclude the phpThumb.php file—can anyone else confirm this? Also, do not change the file permissions on files when submitting PRs like this. I realize it is most likely a result of the source repository, but in this case, you can simply change the permissions before submitting the PR. All the PHP files modified in this PR are attempting to change permissions from 100644 -> 10755.

Ibochkarev:
@opengeek I fixed the permissions on the files on 644 and deleted the phpThumb.php file

Jako (approved):

alroniks (approved):

opengeek (approved):

Issue #14309, opened 2019-01-25
Error caching lexicon topic


sebastian-marinescu:
### Summary
Occasionally, seldom and still regularly, I get the error `Error caching lexicon topic` in the logs.

I'd really like to know what is causing this. **Why** is this happening? 

### Step to reproduce
Use MODX, best with some Extras active. 

### Observed behavior
Sometimes nothing, but the log-message. Sometimes really no translated lexicon-tags in the front-end, which go away by itself, or after clearing cache. Sometimes blank page?

### Expected behavior
Working lexicon system.

### Environment
All versions I've been using so far, in all environments. 

### Related 
https://forums.modx.com/thread/82511/error-caching-lexicon-topic
https://community.modx.com/t/error-caching-lexicon-topic/3154/2
https://github.com/modxcms/revolution/issues/7324
https://github.com/modxcms/revolution/issues/9667

Jako:
~~One source of this issue would be a lexicon file that does not create a $_lang array.~~

~~Put `return false;` in one lexicon file or create a lexicon file without any content. After clearing the cache, that error message should be triggered. Maybe you have a file like this somewhere in the tree. But the name of the file should be logged.~~

Sorry this creates a different error message:
[2019-02-13 11:15:49] (DEBUG @ /somepath/core/model/modx/modlexicon.class.php : 259) An error occurred while trying to cache lexicon/de/core/trash (lexicon/language/namespace/topic)

sebastian-marinescu:
Thank you very much @Jako for this information. 
I'll test/check this soon and will report back.

sebastian-marinescu:
So in the last 3 days these logs accumulated: 

[2019-01-27 18:14:31] (ERROR @ /somepath/www/core/model/modx/modcachemanager.class.php : 349) Error caching lexicon topic [lexicon/en/babel/default](https://pastebin.com/xpDYPtnE)
[2019-01-28 20:00:04] (ERROR @ /somepath/www/core/model/modx/modcachemanager.class.php : 349) Error caching lexicon topic [lexicon/de/mxcalendars/default](https://pastebin.com/WcnHfGNw)
[2019-01-28 20:00:04] (ERROR @ /somepath/www/core/model/modx/modcachemanager.class.php : 349) Error caching lexicon topic [lexicon/en/batcher/default](https://pastebin.com/cFKXydfN)
[...]

PR #14367, opened 2019-02-12
Enhance the caching lexicon topic error message

Jako:
### What does it do?
Enhance the caching lexicon topic error message.

### Why is it needed?
To locate the reason for #14309.

### Related issue(s)/PR(s)
#14309


JoshuaLuckers (approved):

opengeek:
I'm not sure how this change is going to help find the source of the problem for #14309. Can you clarify?

Jako:
Since the cache issue is totally random according to #14309, it could be a start to have the cache options logged and dig deeper afterwards.

Mark-H:
I don't really see this helping find the issue either. The cache options can be seen from configuration and I seriously doubt the problem is the values being empty. 

More likely, it has a problem writing the file. My guess is that in most cases the file is locked because another simultaneous process is writing it, and any other case will need deeper inspection to figure out.

Issue #12991, opened 2016-04-28
Firefox 46 on Linux brings GTK3 compatibility and shows discrepencies in MODX manager theme.


muzzwood:
### Summary

MODX Manager theme relies on browser defaults for scroll bars and form input text. With new Firefox, those defaults are not what they used to be. In the interest of uniformity across platforms, default colours for form text and scroll bars should be specified in CSS.
### Step to reproduce

Open the MODX manager on any Linux desktop environment with a dark theme.
### Observed behavior

See attached screenshots
### Expected behavior

Form text should be readable. Scroll bars should not be dark.
### Environment

MODX 2.5.0. Ubuntu Gnome 15.10. Firefox 46.0

If you look closely you can see there is white text in the username field
[http://imgur.com/ZATRkNn](http://imgur.com/ZATRkNn)

Note the dark scroll bar
[http://imgur.com/1XMhjnL](http://imgur.com/1XMhjnL)

Jako:
In my eyes it is more a Firefox than a MODX issue. So please tell us, why those default colors should not be specified in the internal browser css.

digitalpenguin:
You're right, it's a Firefox issue. But it means that the MODX manager is now unusable on this platform. With a couple of lines of CSS, it could be made usable.

Mark-H:
Yeah that should get fixed, especially the input text. Is it like that on every form across the manager?

digitalpenguin:
Yes, every input/textarea field as far as I can see

Jako:
Hmm, there are no reset css rules in https://github.com/modxcms/revolution/tree/2.x/_build/templates/default/sass
where this would belong to.

digitalpenguin:
I'm not sure of the order the files are compiled but what about here: 
https://github.com/modxcms/revolution/blob/2.x/_build/templates/default/sass/_colors-and-vars.scss
?

jaygilmore:
I wonder what @jpdevries has to say on this matter? While a seamless app experience is desirable caution may need to be taken to alter a user's browser settings.

Mark-H:
Is there really a question wether white-on-white text is bad? It's unusable, that's a bug.

Jako:
It is a bug of Firefox (or Gnome) that switches the text color to white when some theme is changed to dark.

digitalpenguin:
[...]

PR #14369, opened 2019-02-12
Add normalize.css

Ibochkarev:
### What does it do?
Add normalize.css

### Why is it needed?
Normalize.css is an alternative to CSS resets. 

**The aims of normalize.css are as follows:**
- The aims of normalize.css are as follows:
- Preserve useful browser defaults rather than erasing them.
- Normalize styles for a wide range of HTML elements.
- Correct bugs and common browser inconsistencies.
- Improve usability with subtle improvements.
- Explain the code using comments and detailed documentation.
- It supports a wide range of browsers (including mobile browsers) and includes CSS that normalizes HTML5 elements, typography, lists, embedded content, forms, and tables.

Despite the project being based on the principle of normalization, it uses pragmatic defaults where they are preferable.

### Related issue(s)/PR(s)
#12991


Jako:
This will not help with #12991, since it does not set the default input color/background.

Ibochkarev:
@Jako If this does not help with this problem, it will still improve the correctness of displaying the default styles of different browsers into a single view.

Jako:
Things like this should be started in 3.x. It is an enhancement, that could have an impact for the whole manager styling.

Ibochkarev:
~~I will close this PR. Add Normalize.css, together with a complete refactoring of the styles and corrections in the administrative panel in the branch 3.x.~~

Ibochkarev:
@Jako I changed the branch to 3.x. Please check.

_build/templates/default/package.json
```diff
@@ -21,6 +21,9 @@
   "dependencies": {
     "bourbon": "^5.1.0",
     "bourbon-neat": "^3.0.0",
-    "@fortawesome/fontawesome-free": "^5.7.0"
+    "@fortawesome/fontawesome-free": "^5.7.0",
+    "font-awesome": "^4.7.0",
```
Jako: `"font-awesome": "^4.7.0"` has to be removed.

_build/templates/default/package.json
```diff
@@ -21,6 +21,9 @@
...
+    "node-neat": "^1.7.2",
```
Jako: This line should be there, when 2.x is merged into 3.x

_build/templates/default/sass/index.scss
```diff
@@ -22,6 +22,7 @@ $fa-css-prefix: icon;
 
 $fa-css-prefix: fa;
 
+@import "normalize/_normalize";
```
Jako: You have added the normalize-scss package to package.json but you have included the files of that package in this commit too (and this twice: i.e. normalize/normalize/_variables.scss and normalize/_variables.scss!). Adding the files is superfluous.

Please refer here to the files in this line by descending/ascending the tree with multiple `../` until you get to `node_modules/normalize-scss/sass/_normalize.scss` (please check the path).

Jako (changes requested):

_build/templates/default/sass/index.scss
Ibochkarev: In the gruntfile.js I made similar to Font-awesome copying files from node_modules folder. Next is the connection of one normalize file.scss . The way I fixed it.

Jako: But copying is not needed, since there are no files referenced by url like font files for the browser.

Jako: Copying the scss files of fontawesome is too much too. The location of the font files could be set with a sass variable (if thats not done already) and the scss files could be included directly from the node_modules folder.

Ibochkarev: Fixed

Jako: Thanks!

Jako (approved):

_build/templates/default/sass/index.scss
```diff
@@ -1,3 +1,4 @@
+@import "../node_modules/normalize-scss/fork-versions/default/normalize";
```
Jako: according to the [docs](https://github.com/JohnAlbin/normalize-scss) this should be used (Approach 2 – no mixin): 

`@import "[path to]/normalize-scss/sass/normalize/import-now"`

Jako (changes requested):

_build/templates/default/sass/index.scss
Jako: The fork-versions are meant to be edited after copying.

Ibochkarev: @Jako Thanks

Jako (approved):

Mark-H:
The reported issue was closed as no longer being a problem so I wonder if this is a necessary change and if this isn't going to cause new issues due to changing defaults?

muzzwood:
That specific issue was closed as Firefox changed how they did their rendering.
However, as it may be an issue again in the future, this could be worth doing now if it doesn't cause too many problems.
I don't mind at all if it's held off until there is another problem though. It's not a big priority for now.

OptimusCrime:
I am in favor of introducing a normalize stylesheet, even if it does not solve a problem here and now. Having a default that is not vendor specific makes life easier for everyone.

Ibochkarev:
@OptimusCrime thanks my friend!

Mark-H (approved):
Fair enough. I just hope it doesn't break ExtJS styles that depend on legacy behaviour ;)

PR #14371, opened 2019-02-12
Fix wrong width of textarea and text TVs

Ibochkarev:
### What does it do?
Fix wrong width of textarea and text TVs

### Why is it needed?
- Refactoring styles _form.scss
- Add styles to correct width of textarea and text TVs

### Related issue(s)/PR(s)
#13323


Ibochkarev:
Close for revision!

PR #14377, opened 2019-02-14
Fix watch grunt task where used outdated and already removed growl plugin

alroniks:
### What does it do?
It replaces `growl` by `notify`

### Why is it needed?
To get the task `grunt watch` working.
It happened after replacing growl by notify by in one place I forgot to replace the call.

### Related issue(s)/PR(s)
#14315


Jako (approved):

opengeek:
Can you recreate this with a branch not pushed to the origin repository?

Issue #9092, opened 2012-11-12
need an extra permission, sort


modxbot:
rdaneeel created Redmine issue ID 9092

Clients can't reorder documents in the resource tree without saving permissions.
Would be nice if there is a seperate 'order' permission in acces policy.
So the can reorder the pages

Mark-H:
Closing for inactivity.

PR #14378, opened 2019-02-14
add extra permission - sort

iWatchYouFromAfar:
### What does it do?
1 - Added new permission in ACL (Access Control List) with the name "resource_tree_sort"
2 - At the resource sorting in the resource tree, I set the condition for checking the availability of this permission, otherwise we return an error with an explanation.

Just let me show you.

### Why is it needed?
So, we can control the resolution of sort resources in the tree, which affects the value in their menuindex field. And it does not affect the full editing of resources (including the menuindex ).

### Related issue(s)/PR(s)
Fix for https://github.com/modxcms/revolution/issues/9092


core/lexicon/en/resource.inc.php
```diff
@@ -72,6 +72,7 @@
 $_lang['resource_err_new_parent_nf'] = 'New parent resource with id [[+id]] not found.';
 $_lang['resource_err_remove'] = 'An error occurred while trying to remove the resource.';
 $_lang['resource_err_save'] = 'An error occurred while trying to save the resource.';
+$_lang['resource_err_sort'] = 'An error occurred while trying to move a resource in the resource tree.';
```
Jako, line 75: Could there be two error messages? One for not allowed and one for error occured?

Jako (changes requested):

Jako:
The menuindex field has to be hidden on the resource edit page. And the resource save processor (or better modResource) should not change the menuindex then.

Jako:
And drag&drop should be not allowed in the tree.

core/lexicon/en/resource.inc.php
iWatchYouFromAfar: I think that we can think about this or rename the lexicon. It is necessary to understand in which case to display an error occured message ...

Jako:
It is already available (with a different approach): https://github.com/modxcms/revolution/pull/377/files

I'm not sure wether to close this PR.

iWatchYouFromAfar:
In connection with the above requirements from Jako and the release of version 2.7.1 - I decided to close this PR for some updates and include it new function into version 3.x.

I'll do a new PR later!

Issue #4841, opened 2011-05-24
Customized Icons/Text for content types


modxbot:
everettg_99 created Redmine issue ID 4841

When building a document repository using thousands of static resources, it can be really difficult to tell which MODx Static Resources are representing PDFs or .docs or .xls files because they all use the same icon.  It would save a lot of time and be much easier for the user if there were a way to customize the display of MODx resources based on content type.

One idea:  Each time you define a content type (e.g. .pdf), allow a custom icon to be specified.
Second idea: Allow the menu text to be customized, e.g. a system setting with placeholders: the default would simply be something like

<pre>
[[+menutitle]] ([[+id]])
</pre>

But you could get a lot of mileage out of that simple template if there were also a placeholder for the file extension or for the various content type attributes (e.g. content type name).

OR, even better, do both.

Thanks.

rtripault:
While this is still not possible, a lot has been done about resources' icons.

I guess the closest solution would be to define a template per content-type, with the appropriate icon.

But some cleanup in the way tree icons are handled (ie. a dedicated class/service) and handling content types would still be nice to have!

adamwintle:
As far as I know this can now be done on a per-resource level. @sottwell @theboxer can you confirm this is the case?

gpsietzema:
As far as I know this cannot be done per-resource, but per template. Would be nice though :)

OptimusCrime:
Idealy it would override the template icon?

alroniks:
> Idealy it would override the template icon?

I guess template icons should override content type icons. Content type always predefined and it seems as a default property when the template can be changed later.

alroniks:
So, an interesting question now. We can add icons to content types, but it means that we should predefine some of the icons for these content types but the current version of font awesome does not have enough icons. 

For example, we have these content types:
- HTML - not file html icon, only html5, that looks not nice - https://fontawesome.com/icons/html5?style=brands
- XML - not any icon
- text - not any appropriate icon
- CSS - not file icon, but brand icon - https://fontawesome.com/icons/css3?style=brands
- javascript - not file icon, only brand - https://fontawesome.com/icons/js?style=brands
- RSS - icon exists, but not file icon  - https://fontawesome.com/icons/rss?style=solid
- JSON - not any icon, but probably can be applied code icon
- PDF - file icon exists - https://fontawesome.com/icons/file-pdf?style=regular

So, only pdf type has an appropriate icon that does this feature request absolutely useless.

[...]

PR #14383, opened 2019-02-15
Customized icons for content types

alroniks:
### What does it do?
It adds to content types new field (column) with icon class that will be shown in the resource tree.

Also, during work on this, I've fixed one issue about updating grid when CheckColumn plugin is used. See details in https://github.com/modxcms/revolution/pull/14383/commits/4a89afe8453d6c7bb180ad0ad906c05f086774cf

### Why is it needed?
It will allow to easier determine that resource has a different content type. 
NOTE: Template icon can override the content type icon as content type more basic entity than a template.

For HTML content type icon is set to empty to not change existing condition.

Small animation about how it works.

content-type-icons

### Related issue(s)/PR(s)
#4841

### DIFF
--- a/core/model/modx/processors/resource/getnodes.class.php
+++ b/core/model/modx/processors/resource/getnodes.class.php
@@ -436,11 +436,13 @@ class modResourceGetNodesProcessor extends modProcessor {
 
         $contentType = $resource->getOne('ContentType');
         if ($contentType && $contentType->get('icon')) {
+            $iconCls = [];
             $iconCls[] = $contentType->get('icon');
         }
 
         $template = $resource->getOne('Template');
         if ($template && $template->get('icon')) {
+            $iconCls = [];
             $iconCls[] = $template->get('icon');
         }
 
--- a/core/model/modx/processors/resource/getnodes.class.php
+++ b/core/model/modx/processors/resource/getnodes.class.php
@@ -436,17 +436,15 @@ class modResourceGetNodesProcessor extends modProcessor {
 
         $contentType = $resource->getOne('ContentType');
         if ($contentType && $contentType->get('icon')) {
-            $iconCls = [];
-            $iconCls[] = $contentType->get('icon');
+            $iconCls = [$contentType->get('icon')];
         }
 
         $template = $resource->getOne('Template');
         if ($template && $template->get('icon')) {
-            $iconCls = [];
-            $iconCls[] = $template->get('icon');
+            $iconCls = [$template->get('icon')];
         }
 
-        if (!count($iconCls)) {
+        if (empty($iconCls)) {
             $iconCls[] = $this->modx->getOption('mgr_tree_icon_' . $classKey, null, 'tree-resource', true);
         }
 
--- a/core/model/modx/processors/resource/getnodes.class.php
+++ b/core/model/modx/processors/resource/getnodes.class.php
@@ -426,21 +426,26 @@ class modResourceGetNodesProcessor extends modProcessor {
             }
         }
 
-        // Check for an icon class on the resource template
-        $tplIcon = $resource->Template ? $resource->Template->icon : '';
-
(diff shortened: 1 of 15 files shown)


core/model/modx/processors/resource/getnodes.class.php
```diff
@@ -426,21 +426,28 @@ public function prepareResourceNode(modResource $resource) {
...
-        $classKeyIcon = $this->modx->getOption('mgr_tree_icon_' . $classKey, null, 'tree-resource', true);
+        $iconCls = [];
 
-        if (!empty($tplIcon)) {
-            $iconCls[] = $tplIcon;
-        } else {
-            $iconCls[] = $classKeyIcon;
+        $contentType = $resource->getOne('ContentType');
+        if ($contentType && $contentType->get('icon')) {
+            $iconCls = [];
```
JoshuaLuckers:
```suggestion
            array_shift($iconCls);
```

core/model/modx/processors/resource/getnodes.class.php
```diff
@@ -426,21 +426,28 @@ public function prepareResourceNode(modResource $resource) {
...
+            $iconCls[] = $contentType->get('icon');
+        }
+
+        $template = $resource->getOne('Template');
+        if ($template && $template->get('icon')) {
+            $iconCls = [];
```
JoshuaLuckers:
```suggestion
            array_shift($iconCls);
```

core/model/modx/processors/resource/getnodes.class.php
```diff
@@ -426,21 +426,28 @@ public function prepareResourceNode(modResource $resource) {
...
+            $iconCls[] = $template->get('icon');
+        }
+
+        if (!count($iconCls)) {
```
JoshuaLuckers:
```suggestion
        if (empty($iconCls)) {
```
We don't need to know how large the array is only if it's empty.

core/model/modx/processors/resource/getnodes.class.php
Mark-H: This line seems unnecessary entirely ($iconCls is already initialised empty in line 435) and I'm not sure what array_shift would accomplish on an empty array?

manager/assets/modext/widgets/system/modx.grid.content.type.js
```diff
@@ -137,6 +146,10 @@ Ext.extend(MODx.grid.ContentType,MODx.grid.Grid,{
         });
         window.show(e.target);
     }
+
+    ,renderIconField: function (v, md, rec) {
+        return new Ext.XTemplate('<i class="icon icon-lg {icon}"></i>&nbsp;&nbsp; {icon}').apply(rec.data);
```
Mark-H: worth applying a htmlEncode to prevent introducing new XSS issues

core/model/modx/processors/resource/getnodes.class.php
JoshuaLuckers: I suggested it to make sure things won't break if someone adds a new if statement with the same kind of logic before this one.

Ruslan-Aleev:
<del>I tested it, for some reason I don’t see an icon for "icon-json", see the picture below.</del>
I forgot to rebuild styles, everything works well.

Ruslan-Aleev (dismissed):
Everything works as expected.

core/model/modx/processors/resource/getnodes.class.php
alroniks: `array_shift` deletes only first element but the idea is to clear the whole array before setting new icons.

alroniks:
@Mark-H @JoshuaLuckers I've improved the code a bit but without changing how it should work.

JoshuaLuckers (approved):

Mark-H (approved):

PR #14386, opened 2019-02-15
Fix the bug with copying TVs when caption is empty

### LINKED ISSUE #14381: Bug with copying TVs caption
### Summary
If you copy a TV variable and leave the caption empty for the new variable, then "Duplicate + the description of the initial variable" will be written to it.

### Step to reproduce
1. Create any TV with any name and caption.
2. Copy this TV.
3. When MODX asks you for the name and caption of the new variable - make the caption empty.
4. Save this TV, look at its caption.

### Observed behavior
tvcaption

### Expected behavior
Empty caption for for new TV

### Environment
MODX 2.7.0

GulomovCreative:
### What does it do?
Removing a check for the presence of a caption field

### Why is it needed?
Caption field for TV is not required and may be empty. So, during copying if you leave the field blank, then be written "Copy (TV caption)"

### Related issue(s)/PR(s)
Closes #14381

### DIFF
--- a/core/model/modx/processors/element/tv/duplicate.class.php
+++ b/core/model/modx/processors/element/tv/duplicate.class.php
@@ -116,20 +116,6 @@ class modTemplateVarDuplicateProcessor extends modElementDuplicateProcessor {
     }
 
     /**
-     * Get the new caption for the duplicate
-     * @return string
-     */
-    public function getNewCaption()
-    {
-        $caption = $this->getProperty($this->captionField);
-        $newCaption = !empty($caption)
-            ? $caption
-            : $this->modx->lexicon('duplicate_of', array('name' => $this->object->get($this->captionField)));
-
-        return $newCaption;
-    }
-
-    /**
      * Set the new caption to the new object
      * @param $caption
      * @return string
@@ -146,7 +132,7 @@ class modTemplateVarDuplicateProcessor extends modElementDuplicateProcessor {
      */
     public function beforeSave()
     {
-        $caption = $this->getNewCaption();
+        $caption = $this->getProperty($this->captionField);
         $this->setNewCaption($caption);
 
         return parent::beforeSave();
--- a/core/model/modx/processors/element/tv/duplicate.class.php
+++ b/core/model/modx/processors/element/tv/duplicate.class.php
@@ -116,6 +116,15 @@ class modTemplateVarDuplicateProcessor extends modElementDuplicateProcessor {
     }
 
     /**
+     * Get the new caption for the duplicate
+     * @return string
+     */
+    public function getNewCaption()
+    {
+        return $this->getProperty($this->captionField);
+    }
+
+    /**
      * Set the new caption to the new object
      * @param $caption
      * @return string
@@ -132,7 +141,7 @@ class modTemplateVarDuplicateProcessor extends modElementDuplicateProcessor {
      */
     public function beforeSave()
     {
-        $caption = $this->getProperty($this->captionField);
+        $caption = $this->getNewCaption();
         $this->setNewCaption($caption);
 
         return parent::beforeSave();


Jako:
Maybe I am wrong, but removing the logic here is wrong. 

If I am right, the logic in the line 125 of the old code should just be swapped.

GulomovCreative:
> Maybe I am wrong, but removing the logic here is wrong.
> 
> If I am right, the logic in the previous line 125 should just be swapped.

@Jako I didn't understand what it was about

Jako:
The old code updates the existing caption with the duplicate lexicon entry. But only when it is empty (which is the real bug here).

GulomovCreative:
> The old code updates the existing caption with the duplicate lexicon entry. But only when it is empty (which is the real bug here).

@Jako Do you have any suggestion? Because I do not see the point in the existence of this method. The caption field can be blank

GulomovCreative:
Correct me if I'm wrong

ghost:
If the caption is empty and left empty, what would show in the TV tab for this TV? Perhaps some default is better than nothing?

GulomovCreative:
> If the caption is empty and left empty, what would show in the TV tab for this TV?

The name field will be displayed, it is required.

Again, my code does not break anything. MODX allows you to left empty the caption field without my code. But, we have bug during copying TVs. With my code MODX will still offer the default value, but if you want to leave the caption field empty, there will be no bug.

GulomovCreative:
Has anyone tested this PR?) Because everything works correctly. But it doesn't matter to me. Suggest other solutions, I will be glad.

GulomovCreative:
> The old code updates the existing caption with the duplicate lexicon entry. But only when it is empty (which is the real bug here).

@Jako This method is not needed. Why does MODX decide for the user what to save, if there are no any restrictions?

Is this correct on UX?

Doesn’t the user have to find out about something unacceptable before saving?

Jako:
Your code works fine, but If an old caption exist, the duplicate has the same caption as the original.

The old code would have changed this with the duplicate lexicon, if it worked right, so I suggest just to swap the condition in line 125 of the old code.

Then a duplicate with an empty caption will stay with an empty caption and the duplicate will get a 'Duplicate of ' prefix in english.

GulomovCreative:
> Your code works fine, but If an old caption exist, the duplicate has the same caption as the original.
> 
> The old code would have changed this with the duplicate lexicon, if it worked right, so I suggest just to swap the condition in line 125 of the old code.
> 
> Then a duplicate with an empty caption will stay with an empty caption and the duplicate will get a 'Duplicate of ' prefix in english.

 'Duplicate of ' with prefix value still gets with the js and is offered to the user, is this not enough?

Jako:
Ok, I missed that js part ...

Jako (dismissed):

Ruslan-Aleev (dismissed):
Everything works as described

core/model/modx/processors/element/tv/duplicate.class.php
```diff
@@ -115,20 +115,6 @@ public function duplicateMediaSources() {
         }
     }
 
-    /**
-     * Get the new caption for the duplicate
-     * @return string
-     */
-    public function getNewCaption()
```
Mark-H: It might be worth to have a discussion about wether or not removing public methods of a processor constitutes a breaking change*, but until that discussion has been had, I think we should err on the side of caution and not remove any public class methods/variables such as this one in a patch release. 

Could you please restore this method? It can have the new behavior you're doing in `beforeSave`, but the method signature should stay the same. 

---

* According to semver, a change in a "public api" is a breaking change, and a public method of a class does count as a public API... but one could also argue that the "public api" of a processor in MODX is the `$modx->runProcessor($action)` syntax, which isn't affected by this removal in the slightest.

Mark-H (changes requested):

core/model/modx/processors/element/tv/duplicate.class.php
GulomovCreative: @Mark-H What do you offer?
Restore the method which will return the caption?

GulomovCreative: Is that better?

Jako (approved):

Ibochkarev:
@Mark-H please check again this PR. Thanks

Mark-H (approved):

Issue #14382, opened 2019-02-15
2.7.0-pl upgrade script is running on installing 2.7.1


Jako:
### Summary
The 2.7.0-pl upgrade script is running on installing 2.7.1.

### Step to reproduce
Install MODX 2.7.1 on a modx 2.7.0 installation. The alias_visible column is added again and throws an error, since it already exists.

### Observed behavior
The alias_visible column is added again and throws an error, since it already exists.

### Expected behavior
The alias_visible column is not added again.

### Environment
2.7.1

Ibochkarev:
I can reproduce this error.

PR #14396, opened 2019-02-16
Don't execute the upgrade script for the last installed version again

Jako:
### What does it do?
Fix the version compare for detecting the needed upgrade scripts.

### Why is it needed?
The scripts of the last installed version are executed again because of the equal sign. With the change only the scripts larger than the last installed version are executed.

### Related issue(s)/PR(s)
#14382

### DIFF
--- a/setup/includes/modinstallversion.class.php
+++ b/setup/includes/modinstallversion.class.php
@@ -125,7 +125,7 @@ class modInstallVersion {
                 if (is_dir($path.$script)) continue;
                 $sc = str_replace('.php','',$script);
 
-                if (version_compare($this->version,$sc,'<=')) {
+                if (version_compare($this->version,$sc,'<')) {
                     $scripts[] = $path.$sc.'.php';
                 }
             }


Mark-H:
Don't we want an upgrade script named 2.7.2-pl to be executed when upgrading to 2.7.2? This seems like it would cause that script to not be executed.

Jako:
$this->version contains the previous version at that time during my tests. See: https://github.com/modxcms/revolution/blob/bec29b09cdd87145a710fb87a6bb8bc7f59065cc/setup/includes/modinstallversion.class.php#L143-L150

Jako:
Maybe the variable could be renamed too.

Mark-H (approved):
Gotcha. Thanks for explaining.

alroniks (approved):

rthrash:
This pull request has been mentioned on **Community**. There might be relevant details there:

https://community.modx.com/t/db-errors-when-updating-to-2-7-1-error-updating-table-for-class-modresource/435/3

PR #14398, opened 2019-02-16
Fix modprocessor deprecated create_function()

kudashevs:
### What does it do?
Change preg_replace_callback call with deprecated create_function() with to new one with anonymous function.

### Why is it needed?
Due to multiple warning in log "(WARN @\www\core\model\modx\modprocessor.class.php : 314) PHP deprecated: Function create_function() is deprecated" and the fact that "This function create_function() has been deprecated as of PHP 7.2.0. Relying on this function is highly discouraged." I think better to fix it.

### Related issue(s)/PR(s)
I made a search through opened issues and pull requests and didn't find any.


kudashevs:
Just find same issue in another file and made a new pull request #14399.

Issue #14338, opened 2019-02-06
Trash manager grid resource title not correctly escaped


travisbotello:
### Summary
Using script tag inside content field of a deleted resource causing resource's title in trash manager grid being messed up.

### Step to reproduce
- Create a resource with `pagetitle` *test* and content like this:
```
<script>
var test = "test";
</script>
```
- Delete this resource
- Switch to trash manager

### Observed behavior
In grid you will see the resource's title being rendered `">Test` and tooltip content preview `var test =`

### Expected behavior
Resource's title in the grid shouldn't be messed up by content.

### Environment
MODX 2.7.0 advanced

JoshuaLuckers:
This problem is solved by merging PR #14401

PR #14401, opened 2019-02-17
Fix, added escape field content for tooltip

tolanych:
### What does it do?
Added escape html entities on server-side in processor trash/getlist and on frontend in renderTooltip

### Why is it needed?
image

### Related issue(s)/PR(s)
#14338


Jako (approved):

alroniks (approved):

manager/assets/modext/widgets/resource/modx.grid.trash.js
```diff
@@ -469,6 +469,7 @@ Ext.extend(MODx.grid.Trash, MODx.grid.Grid, {
                 + ((record.json.longtitle) ? '<p><strong>' + _('long_title') + ':</strong> ' + record.json.longtitle + '</p>' : '')
                 + ((record.data.parentPath) ? '<p><strong>' + _('trash.parent_path') + ':</strong> ' + record.data.parentPath + '</p>' : '')
                 + ((record.json.content) ? '<p><strong>' + _('content') + ':</strong> ' + Ext.util.Format.ellipsis(record.json.content.replace(/<\/?[^>]+>/gi, ''), 100) + '</p>' : '');
+            preview = Ext.util.Format.htmlEncode(preview);
```
Mark-H, line 472: When merging, please double check that this does not cause all the `<p><strong>`'s to be rendered as text. From reading the code I don't think it will, as the qtip has to be added into the div first and presumably that will decode the html to render it appropriately, but just for making sure.

Mark-H (approved):

manager/assets/modext/widgets/resource/modx.grid.trash.js
JoshuaLuckers: Done! Works as described 👍

PR #14402, opened 2019-02-17
Fix Duplicate behavior on resource update page

### LINKED ISSUE #4834: Inconsistent 'Duplicate' command behavior
gadamiak created Redmine issue ID 4834

Duplicating a resource behaves inconsistently when used via context menu in resource tree and via a button on resource edit form. The former opens a dialog with option to include children resources and setting title for a duplicate and allows to cancel duplication. The latter issues warning about all child resources being duplicated and creates new resource. User is presented with a form for a new resource and if cancels edition an empty untitled resource is left.

Both should have the same behavior of the former (there was an issue for it), ie. context menu command.

GulomovCreative:
### What does it do?
Adds normal dialog option like duplicating via context menu in resource tree

ux

### Why is it needed?
Description from related issue #4834 

> Duplicating a resource behaves inconsistently when used via context menu in resource tree and via a button on resource edit form. The former opens a dialog with option to include children resources and setting title for a duplicate and allows to cancel duplication. The latter issues warning about all child resources being duplicated and creates new resource. User is presented with a form for a new resource and if cancels edition an empty untitled resource is left.
>
> Both should have the same behavior of the former (there was an issue for it), ie. context menu command.

### Note for testers
1. There are a few bugs, but they are not related to this PR. Later, I or someone will post the issues.
2. Checks for the existance on node in this PR are needed for these cases when resource hide in tree (for example: miniShop2 uses show in tree 0 to hide products).

### Related issue(s)/PR(s)
Closes #4834

### DIFF
--- a/core/lexicon/en/resource.inc.php
+++ b/core/lexicon/en/resource.inc.php
@@ -48,7 +48,6 @@ $_lang['resource_delete_confirm'] = 'Are you sure you want to delete this resour
 $_lang['resource_description'] = 'Description';
 $_lang['resource_description_help'] = 'This is an optional description of the resource.';
 $_lang['resource_duplicate'] = 'Duplicate';
-$_lang['resource_duplicate_confirm'] = 'Are you sure you want to duplicate this resource? Any item(s) it contains will also be duplicated.';
 $_lang['resource_edit'] = 'Edit';
 $_lang['resource_editedby'] = 'Edited By';
 $_lang['resource_editedon'] = 'Edited On';
--- a/manager/assets/modext/sections/resource/update.js
+++ b/manager/assets/modext/sections/resource/update.js
@@ -55,8 +55,7 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
     ,duplicateResource: function(btn,e) {
         var t = Ext.getCmp('modx-resource-tree');
         var id = this.config.resource;
-        var nodeId = this.config.record.context_key + '_' + id;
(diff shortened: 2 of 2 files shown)


GulomovCreative:
I have an idea. Make a checkbox, which would give the choice to redirect after the success duplication to the new duplicated resource or not. 

What do you think? I think this is good UX behavior.

GulomovCreative:
Default value from resource update page let it be true, but from resource tree false

GulomovCreative:
Any suggestions?

Ruslan-Aleev:
> I have an idea. Make a checkbox, which would give the choice to redirect after the success duplication to the new duplicated resource or not.

I think this is a very good idea. But this should work both in the tree and on the button.
It also makes sense to add this functionality for other elements (template, chunk, etc.).

ant-tar:
> I think this is a very good idea. But this should work both in the tree and on the button.
> It also makes sense to add this functionality for other elements (template, chunk, etc.).

Totally agree with @GulomovCreative  and @Ruslan-Aleev , this will be very useful .

Jako:
This should better go into 3.x

GulomovCreative:
> This should better go into 3.x


(︶︹︺)

JoshuaLuckers:
For 2.x we only accepts bug fixes, this is a great addition but also introduces a new "feature" and therefore it's better to have this target 3.x.

GulomovCreative:
I will send other PR

PR #14411, opened 2019-02-19
Refactored behavior for duplicating resources and elements

### LINKED ISSUE #4834: Inconsistent 'Duplicate' command behavior

### LINKED ISSUE #14409: Title input field is deactivated when copying child resources
### Summary
Title input field is deactivated when copying child resources

### Step to reproduce
When duplicating a resource, if it has child resources when switching the checkbox (Copy child resources), the field for entering the title for the duplicate is deactivated.

### Expected behavior
The checkbox (copy child resources) in the checked state should not deactivate the field named

### Environment
MODX version 2.7.1

GulomovCreative:
### What does it do?
1. Adds normal resource duplication dialog like duplicating via context menu in resource tree
2. Adds new **redirect to duplicate** option before duplicating resources and elements (Templates, TVs, Chunks, Snippets and Plugins)
3. Fixs misbehavior duplicate children checkbox
4. Adds related headers for duplicating modal

ux

### Why is it needed?
Good UX behavior + description from related issue #4834

> Duplicating a resource behaves inconsistently when used via context menu in resource tree and via a button on resource edit form. The former opens a dialog with option to include children resources and setting title for a duplicate and allows to cancel duplication. The latter issues warning about all child resources being duplicated and creates new resource. User is presented with a form for a new resource and if cancels edition an empty untitled resource is left.
>
>Both should have the same behavior of the former (there was an issue for it), ie. context menu command.

### Note for testers
1. There are a few bugs, but they are not related to this PR. Later, I or someone will post the issues.
2. Checks for the existance on node in this PR are needed for these cases when resource hide in tree (for example: miniShop2 uses show in tree 0 to hide products).

### Related issue(s)/PR(s)
Closes #4834 and closes #14409

### DIFF
--- a/core/lexicon/en/default.inc.php
+++ b/core/lexicon/en/default.inc.php
@@ -133,7 +133,6 @@ $_lang['edit_tv'] = 'Edit TV';
 $_lang['editing'] = 'Editing: [[+name]]';
 $_lang['editedon'] = 'Edit Date';
 $_lang['editing_form'] = 'Editing Form';
-$_lang['element_duplicate'] = 'Duplicate Element';
 $_lang['element_duplicate_values'] = 'Duplicate Resource Values?';
 $_lang['element_name_new'] = 'Name of New Element';
 $_lang['element_caption_new'] = 'Caption of New Element';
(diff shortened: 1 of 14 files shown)


Mark-H (approved):

JoshuaLuckers (approved):

rainbowtiger:
So, this is adding the same options to the Duplicate button as well, right? It wasn't clear from the description, and the animation doesn't show that; it just shows the context menu for the tree.

Issue #11902, opened 2014-08-16
MySQL : modUserProfile.gender as tinyint


rtripault:
`modUserProfile.gender` stores the user gender as an integer.
There obviously won't be much choices there, so suggesting to change the column type as tinyint (as in MsSQL schema) instead of int

bezumkin:
I believe we could close this issue in 2018 because of, you know, many new genders in the world.

@Mark-H

PR #14412, opened 2019-02-20
Alter modUserProfile gender field to tinyint from int

alroniks:
### What does it do?
Changes type of gender field to `tinyint` instead of `int` as this field shouldn't contain a lot of information.

### Why is it needed?
- to take less place and memory for storing this data
- to keep consistency between different database drivers (sqlsrv uses `tinyint` as well).

### Related issue(s)/PR(s)
#11902

### DIFF
--- a/core/model/modx/mysql/moduserprofile.map.inc.php
+++ b/core/model/modx/mysql/moduserprofile.map.inc.php
@@ -157,7 +157,7 @@ $xpdo_meta_map['modUserProfile']= array (
     ),
     'gender' => 
     array (
-      'dbtype' => 'int',
+      'dbtype' => 'tinyint',
       'precision' => '1',
       'phptype' => 'integer',
       'null' => false,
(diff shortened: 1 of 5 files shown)


alroniks:
I am wondering, it is strongly required migration script for altering this field during the upgrade?

opengeek (changes requested):
Yes, you will need to modify the field in the upgrade script. Does this need a similar change for sqlsrv?

JoshuaLuckers:
What if we want to support [all possible genders](http://genderfluidsupport.tumblr.com/gender/) like Facebook en Tumblr do?

Jako:
`[-128 .. 127]` should be enough for genders

alroniks:
@opengeek I've added migration scripts for both: mysql and sqlsrv, because for sqlsrv it was missed too. Please, check.

Mark-H (approved):

opengeek (approved):

Issue #13201, opened 2016-11-30
Google Chrome translates parts of the Manager it shouldn't


jpdevries:
### Summary
Google Chrome translates parts of the Manager it shouldn't.

### Step to reproduce
Set the manager language to a language other than your primary language. For example, change it from English to Dutch. Chrome will now ask you if you want to translate the Manager page. Say yes. Notice that not only the interface, but the content itself is also translated. This means that if you save the resource your client's content could get translated! Yikes.

### Observed behavior
Client gets mad. Or laughs if they have a good since of humor.

### Expected behavior
MODX should prevent Chrome from translating certain parts, such as the content area, of the Manager.

### Environment
All environments, Google Chrome.

There is a fix for this, and I'll try and work on a PR but this might be something to just keep in mind when the Manager is rebuilt. There is a special class you can use that tells Google translate to leave that part of the DOM alone.

Mark-H:
If this can be implemented in todays manager, that would be welcome.

JoshuaLuckers:
There is an HTML attribute we can use to solve this: [HTML's translate attribute](https://www.w3.org/International/questions/qa-translate-flag)

alroniks:
Here https://support.google.com/webmasters/answer/79812?hl=en said
> When we recognize that the contents of a page are not in the language that the user is likely to want to read, we often provide a link to a translation in the search results. In general, this gives you the chance to provide your unique and compelling content to a much larger group of users. However, there may be situations where this is not desired. This meta tag tells Google that you don't want us to provide a translation for this page.

So, adding `<meta name="google" content="notranslate" />` to the manager pages should solve the issue.

PR #14414, opened 2019-02-20
 Prevent Google Chrome translations from the admin panel

Ibochkarev:
### What does it do?

- Prevent Google Chrome translations from the admin panel 
- Refactoring

### Why is it needed?
To specify a meta Tag to prevent Google translation to translate text in admin panel

### Related issue(s)/PR(s)
#13201


JoshuaLuckers (changes requested):
Please only commit the change with the newly add meta tag. I do see some new formatting and other code changes that should not be part of this PR.

Ibochkarev:
@JoshuaLuckers fixed

JoshuaLuckers (approved):

Mark-H (approved):

PR #14415, opened 2019-02-21
Change System font stack

Ibochkarev:
### What does it do?
Change System font stack.
Defaulting to the system font of a particular operating system can boost performance because the browser doesn't have to download any font files, it's using one it already had. That's true of any "web safe" font, though. The beauty of "system" fonts is that it matches what the current OS uses, so it can be a comfortable look.

### Why is it needed?
Change the current base font set to suggested:
`font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif;`

**This set of fonts is used:**
- https://medium.com/
- https://wptavern.com/wordpress-4-6-to-drop-open-sans-in-the-admin-in-favor-of-system-fonts

**Before**:

**After**:

**Before**:

**After**:

**Before**:

**After**:

**Before**:

**After**:

**More information here:**

- https://css-tricks.com/snippets/css/system-font-stack/
- https://css-tricks.com/implementing-system-fonts-booking-com%E2%80%8A-%E2%80%8Aa-lesson-learned/

**Environment**:
- MODX 3.x
- MacOS 10.13.6 (17G5019)
- Google Chrome: 72.0.3626.109

### Related issue(s)/PR(s)
NA

### DIFF
--- a/_build/templates/default/sass/_colors-and-vars.scss
+++ b/_build/templates/default/sass/_colors-and-vars.scss
@@ -193,7 +193,7 @@ $winFooterBg: $white;
 $winFooterBorderColor: $white;
 
 /* Font stacks */
-$bodyfonts: 'Helvetica Neue', Helvetica, Arial, Tahoma, sans-serif;
+$bodyfonts: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif;
 $bodyFontSize: 13px;
 $altfonts: titilliumregular, $bodyfonts; /* unused */
 $altfontslight: titilliumlight, $bodyfonts; /* unused */


petyagrill:
### It looks very nice

**Before:**
_ modx revolution
_ home _ modx revolution

**After:**
_ home _ modx revolution 2
_ home _ modx revolution 3


**Environment**:

* MODX 3.x
* Ubuntu 18.04.2 LTS
* Google Chrome: 72.0.3626.96

JoshuaLuckers (approved):

Ibochkarev:
@Mark-H Please check this PR

Mark-H (approved):
I will review PRs even if you don't mention me @Ibochkarev ;) Just a little busy with other things so it may take me a little longer every now and then.

PR #14421, opened 2019-02-22
Update editorconfig

Ibochkarev:
### What does it do?
Updates the rules in the file .for editorconfig

### Why is it needed?
**Specifies**:
- the default file encoding of `charset = utf-8`
- the maximum length of a line is 100 characters.

**For files with extension .md:**
- the maximum length of a line in characters is not limited
- no spaces are removed at the end of the lines

### Related issue(s)/PR(s)
NA

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,7 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 tab_width = 4
-max_line_length = 100
+max_line_length = off
 
 [**.js]
 tab_width = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,11 +1,14 @@
+# https://editorconfig.org
 root = true
 
 [*]
+charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 tab_width = 4
+max_line_length = 100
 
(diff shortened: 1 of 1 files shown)


Jako (dismissed):

.editorconfig
```diff
@@ -1,11 +1,14 @@
...
 root = true
 
 [*]
+charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 tab_width = 4
+max_line_length = 100
```
alroniks: Why 100? By PSR-2 (at least for PHP) it said:
> There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less.

So, 120 fits more.

JoshuaLuckers: It's a soft limit, I don't think we should enforce a max line length.

alroniks: I agree @JoshuaLuckers Probably it should be removed or set as `off` to allow use editor settings.

Ibochkarev: I set the value `max_line_length` to `off`

JoshuaLuckers (approved):

Jako (approved):

Mark-H (approved):

PR #14423, opened 2019-02-22
Changing the menuindex of the updating resource after sorting in tree

### LINKED ISSUE #14299: [bug] Change resource meniundex through drag and drop at the time of resource editing
### Summary
If you edit a resource and at the same time move it in the tree through drag and drop, and then save it, the resource position (menuindex) will remain the same at the time of opening the resource.

However, if you edit a resource and change its parent (by moving through drag and drop to another folder) and then save, everything will be fine with the new parent.

This bug is pretty old, but I did not find it in the list. Customers often complained about this issue.

**More comment info** - https://github.com/modxcms/revolution/issues/14299#issuecomment-456017453

### Environment
MODX 2.7.0

GulomovCreative:
### What does it do?
Fixes the bug when sorting resources and locate on resource update page at the same time.

ux

### Why is it needed?
Description from #14299 

> If you edit a resource and at the same time move it in the tree through drag and drop, and then save > it, the resource position (menuindex) will remain the same at the time of opening the resource.
>
> However, if you edit a resource and change its parent (by moving through drag and drop to another folder) and then save, everything will be fine with the new parent.
>
> This bug is pretty old, but I did not find it in the list. Customers often complained about this issue.

### Related issue(s)/PR(s)
Closes #14299

### DIFF
--- a/core/model/modx/processors/resource/sort.class.php
+++ b/core/model/modx/processors/resource/sort.class.php
@@ -24,6 +24,8 @@ class modResourceSortProcessor extends modProcessor {
 
     public $source;
     public $target;
+    public $activeTarget;
+    public $menuindex;
     public $point;
 
     public $autoIsFolder = true;
@@ -50,6 +52,7 @@ class modResourceSortProcessor extends modProcessor {
         $this->fireBeforeSort();
 
         $target = $this->getProperty('target', '');
+        $activeTarget = $this->getProperty('activeTarget', '');
         $source = $this->getProperty('source', '');
         $point = $this->getProperty('point', '');
 
@@ -84,6 +87,10 @@ class modResourceSortProcessor extends modProcessor {
             $this->getProperty('source_type'),
             $this->getProperty('source_pk')
         );
+
+        if (!empty($activeTarget) && $this->getActiveTargetMenuindex($activeTarget)) {
+            return $this->success('', array('menuindex' => (int) $this->menuindex));
+        }
 
         return $this->success();
     }
@@ -197,6 +204,17 @@ class modResourceSortProcessor extends modProcessor {
         $this->moveAffectedContexts($lastRank);
 
         return true;
+    }
+
+    public function getActiveTargetMenuindex(int $target) {
+        $resource = $this->modx->getObject('modResource', $target);
+
+        if($resource instanceof modResource) {
+            $this->menuindex = $resource->get('menuindex');
+        	return true;
+        }
+
+        return false;
     }
 
     public function moveToContext() {
--- a/core/model/modx/processors/resource/sort.class.php
+++ b/core/model/modx/processors/resource/sort.class.php
@@ -209,7 +209,7 @@ class modResourceSortProcessor extends modProcessor {
     public function getActiveTargetMenuindex(int $target) {
         $resource = $this->modx->getObject('modResource', $target);
 
-        if($resource instanceof modResource) {
+        if ($resource instanceof modResource) {
             $this->menuindex = $resource->get('menuindex');
         	return true;
         }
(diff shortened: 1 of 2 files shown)


GulomovCreative:
Come on, who's a tough guy? 😎

core/model/modx/processors/resource/sort.class.php
```diff
@@ -199,6 +206,17 @@ public function sortContexts() {
         return true;
     }
 
+    public function getActiveTargetMenuindex(int $target) {
+        $resource = $this->modx->getObject('modResource', $target);
+
+        if($resource instanceof modResource) {
```
JoshuaLuckers:
```suggestion
        if ($resource instanceof modResource) {
```

GulomovCreative:
If suddenly someone did not understand why I send the ID of the editing resource and not the dragging node: If you try to get the menuindex dragging node, it will work well until the moment when you sort the other resources and the sorting of the editing resource changes in DB but we will not know it.

Ruslan-Aleev (dismissed):
Everything works for me

JoshuaLuckers (dismissed):

core/model/modx/processors/resource/sort.class.php
```diff
@@ -199,6 +206,17 @@ public function sortContexts() {
         return true;
     }
 
+    public function getActiveTargetMenuindex(int $target) {
```
Mark-H: Unfortunately I don't think we can't use scalar type hinting yet - last I heard the target minimum PHP version for 3.x is PHP 5.6, while scalar type hints were introduced in 7.0.

core/model/modx/processors/resource/sort.class.php
```diff
@@ -199,6 +206,17 @@ public function sortContexts() {
...
+        $resource = $this->modx->getObject('modResource', $target);
```
Mark-H: The `getObject` method allows you to pass in a variety of params as the second param - including raw SQL. 

While you're currently typehinting in the method, which would cause $target to always be integer, making getObject treat it as the primary key... if you change the method signature to support PHP 5.6 per my other comment, you should make sure to either cast `$target` to an int here to prevent it from being a potential SQLI vulnerability.

core/model/modx/processors/resource/sort.class.php
```diff
@@ -199,6 +206,17 @@ public function sortContexts() {
...
+
+        if ($resource instanceof modResource) {
+            $this->menuindex = $resource->get('menuindex');
+        	return true;
```
Mark-H: As it looks like there's only one place `getActiveTargetMenuindex` is used, I'd suggest it's better code to return the value you seek (the menuindex) from the method directly, instead of `true`/`false` and setting it to a class property.

Mark-H (changes requested):
There's a few things in the sort processor that I'd prefer to see handled differently.

core/model/modx/processors/resource/sort.class.php
Mark-H: Never mind this, we'll bump the requirements to PHP7 for MODX3. Yay, scalar type hints!

core/model/modx/processors/resource/sort.class.php
Mark-H: Per other comment, the scalar typehint can stay, so this comment is also no longer relevant.

core/model/modx/processors/resource/sort.class.php
Mark-H: This would still be great to fix.

core/model/modx/processors/resource/sort.class.php
GulomovCreative: > Never mind this, we'll bump the requirements to PHP7 for MODX3. Yay, scalar type hints!

Seriously? (• ◡•)

OptimusCrime: Seriously?

Mark-H (approved):

JoshuaLuckers (approved):

JoshuaLuckers:
I'll merge this PR later and also make sure these changes get in PR #14534

Issue #14385, opened 2019-02-15
Unpacking zip file allows uploading arbitrary files


AgelxNash:
### Summary
Bypass **upload_files** settings

### Step to reproduce
- Upload zip file containing files not allowed by upload_files setting
- Unpack

ghost:
I must admit this is one of the reasons I wanted the "unpack" feature in the first place. But I suppose it's not really a good idea to be able to do that.

Was anything ever done about the "rename" feature? I also used to simply upload files with a .txt sufix, then rename the file with .php suffix or whatever it was I wanted to upload.

Of course, opening an FTP/SFTP client and uploading necessary files from that is much preferable, but sometimes I just feel lazy.

Mark-H:
> Was anything ever done about the "rename" feature? I also used to simply upload files with a .txt sufix, then rename the file with .php suffix or whatever it was I wanted to upload.

Yes, that was fixed in #13240.

opengeek:
Can this not be avoided by preventing the upload of a zip file via the upload_files setting?

Ruslan-Aleev:
@opengeek In the context menu there is a whole item on unpacking the zip-archive, i.e. in addition to the preventing .zip, we need to think about unpack functionality (remove it or something else).

PR #14433, opened 2019-02-28
Check extracted files for allowed extensions by option

Jako:
### What does it do?
Check extracted files for allowed extensions by option and enable that option in the file/unpack processor

### Why is it needed?
Without that patch, not allowed files could be unpacked.

### Related issue(s)/PR(s)
#14385

### DIFF
--- a/core/xpdo/compression/xpdozip.class.php
+++ b/core/xpdo/compression/xpdozip.class.php
@@ -234,7 +234,7 @@ class xPDOZip {
             $allowedFileTypes = (!is_array($allowedFileTypes)) ? explode(',', $allowedFileTypes) : $allowedFileTypes;
         } else {
             $allowedFiles = $this->xpdo->getOption('upload_files') ? explode(',', $this->xpdo->getOption('upload_files')) : array();
-            $allowedImages = $this->xpdo->getOption('upload_images') ? explode(',', $this->xpdo->getOption('upload_files')) : array();
+            $allowedImages = $this->xpdo->getOption('upload_images') ? explode(',', $this->xpdo->getOption('upload_images')) : array();
             $allowedMedia = $this->xpdo->getOption('upload_media') ? explode(',', $this->xpdo->getOption('upload_media')) : array();
             $allowedFlash = $this->xpdo->getOption('upload_flash') ? explode(',', $this->xpdo->getOption('upload_flash')) : array();
             $allowedFileTypes = array_unique(array_merge($allowedFiles, $allowedImages, $allowedMedia, $allowedFlash));
--- a/core/xpdo/compression/xpdozip.class.php
+++ b/core/xpdo/compression/xpdozip.class.php
@@ -142,13 +142,24 @@ class xPDOZip {
      *
      * @param string $target The path of the target location to unpack the files.
      * @param array $options An array of options for the operation.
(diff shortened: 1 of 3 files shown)


core/xpdo/compression/xpdozip.class.php
```diff
@@ -209,4 +220,32 @@ public function hasError() {
...
+     * @return bool
+     */
+    private function checkFiletype($filename)
+    {
+        if ($this->xpdo->getOption('allowedFileTypes')) {
+            $allowedFileTypes = $this->getOption('allowedFileTypes');
+            $allowedFileTypes = (!is_array($allowedFileTypes)) ? explode(',', $allowedFileTypes) : $allowedFileTypes;
+        } else {
+            $allowedFiles = $this->xpdo->getOption('upload_files') ? explode(',', $this->xpdo->getOption('upload_files')) : array();
+            $allowedImages = $this->xpdo->getOption('upload_images') ? explode(',', $this->xpdo->getOption('upload_files')) : array();
```
Mark-H: the second getOption here as `upload_files` instead of `upload_images`

Jako: Oh, that must be wrong on the original code too.

Jako: Gotcha: https://github.com/modxcms/revolution/blob/bcb5ab3eac04dcdd834db5af5b775c011cd82385/core/model/modx/sources/modfilemediasource.class.php#L576

Mark-H (approved):
Nice work. I didn't think it'd be possible to keep both the ability to unzip files, *and* prevent the reported issue, but thanks for proving me wrong :)

alroniks (approved):

Ruslan-Aleev:
@Jako can you check the bug, please? Zip-archive can be unpacked only from the file tree, it does not work in the media browser, and an error occurs in the console.

opengeek (commented):
This absolutely should never have been merged. This modifies an xPDO file.

opengeek:
Doesn't this also destroy the transport packaging functionality if someone has this setting enabled?

Jako:
> @Jako can you check the bug, please? Zip-archive can be unpacked only from the file tree, it does not work in the media browser, and an error occurs in the console.

I don't get an error in the console. Nothing happens: No Ajax call and no message for the user (2.7.1 + all current merges here).

But this can't be caused by the changes in this PR.

If I am right, the issue occurs because the unpackFile method of manager/assets/modext/widgets/system/modx.tree.directory.js is currently not available in manager/assets/modext/widgets/media/modx.browser.js, but unpack is displayed in the context menu.

Jako:
> This absolutely should never have been merged. This modifies an xPDO file.

But the extract has to be changed in the xPDO file. I thought it would be ok, to post the PR here and merge the file change afterwards into xPDO.

~~The change does not touch the unpacking of the transport files, since a system setting can't override the check in https://github.com/Jako/revolution/blob/7b77bd872b73bebeb65c3876066e202917470958/core/xpdo/compression/xpdozip.class.php#L151 because $options is an array. The xPDOZip unpack method has to be explicit called with $options = null to make the method use a (normally not existing) system setting.~~

Sorry, I was wrong here: there misses a default in $this->_options to catch an override by a (normally not existing) system setting.

Issue #12284, opened 2015-01-21
System settings - Use of xtype = numberfield


pepebe:
I've noticed that xtype => 'numberfield' (with the exteption of 'tree_root_id') is almost never used in transport.core.system_settings.php?

https://github.com/modxcms/revolution/blob/cd690b16c303823697562c89efb1a8b2045bc429/_build/data/transport.core.system_settings.php

I might be overlooking something, but is there a good reason why we don't use this type for things like 'site_start', 'error_page', etc?

Cheers,

pepebe

Ibochkarev:
@pepebe what are your suggestions for change?

alroniks:
@Ibochkarev He suggested changing default content type for settings that should contain any ID (resource id or element id or another id) to number field instead of text field to avoid issues when code expects valid number but gets a string with a text.

JoshuaLuckers:
When I have time I will fix this 😄

Mark-H:
Fixed in #14447

PR #14447, opened 2019-03-02
Improved xtypes and values for system settings

JoshuaLuckers:
### What does it do?
Checked the system settings and changed the value to the right type and also checked if they could use another xtype like `numberfield`.

### Why is it needed?
Improve the code base.

### Related issue(s)/PR(s)
Closes issue #12284

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -292,8 +292,8 @@ $settings['cache_resource']->fromArray(array (
 $settings['cache_resource_expires']= $xpdo->newObject(modSystemSetting::class);
 $settings['cache_resource_expires']->fromArray(array (
   'key' => 'cache_resource_expires',
-  'value' => '0',
-  'xtype' => 'textfield',
+  'value' => 0,
+  'xtype' => 'numberfield',
   'namespace' => 'core',
   'area' => 'caching',
   'editedon' => null,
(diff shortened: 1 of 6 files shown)


alroniks (changes requested):
Upgrade script required I guess.

JoshuaLuckers:
> Upgrade script required I guess.

I agree 👍.

alroniks:
I am wondering what if somebody added value that does not fit new xtype, maybe we should check the type and skip them, showing the warning? I think it almost impossible but still.

JoshuaLuckers:
> I am wondering what if somebody added value that does not fit new xtype, maybe we should check the type and skip them, showing the warning? I think it almost impossible but still.

Good point. There is already a check if the system setting is using the `old_xtype`, I will try to add the check you propose another time. 

If someone else wants to add such a check: feel free to help!

JoshuaLuckers:
@Alroniks I gave your suggestion some thought and maybe we don't need such a check. The xtype is used when you edit the system setting via the grid. Only if you edit the value it will check if it's a valid value for the defined xtype. 

A warning if the value is invalid for the new xtype would be nice but could be done in another PR. What do you think?

_build/data/transport.core.system_settings.php
```diff
@@ -829,7 +829,7 @@
 $settings['link_tag_scheme']->fromArray(array (
   'key' => 'link_tag_scheme',
   'value' => -1,
-  'xtype' => 'textfield',
+  'xtype' => 'numberfield',
```
Mark-H: This should be text, can also be `full`, `https`, `abs`, etc.

setup/includes/upgrades/common/3.0.0-update-xtypes-system-settings.php
```diff
@@ -0,0 +1,182 @@
...
+    ],
+    [
+        'key' => 'log_level',
+        'old_xtype' => 'textfield',
+        'new_xtype' => 'numberfield',
+    ],
+    [
+        'key' => 'link_tag_scheme',
+        'old_xtype' => 'textfield',
+        'new_xtype' => 'numberfield',
```
Mark-H: This one should be removed per other comment.

Mark-H (changes requested):
Spotted one incorrect change, other than that, nice work.

Mark-H (approved):

Mark-H (approved):

PR #14449, opened 2019-03-02
Fixes the bug when matches names of policy before saving

GulomovCreative:
### What does it do?
1. Fix behavior when matching names of access policy (See picture below)
2. Adds asterisks to required fields on creating policy window (See picture below)
3. Change the call of non-existent "policy_err_name_ns" lexicon entry

### Why is it needed?
If we type the name of the exiting Access Policy being on update page, then we see the next error message "An error occurred while trying to save the Policy", and all. It is not clear what is the error.

**Before:**

before

**After:**

after

And when we create a policy, at once we can't understand what fields are required

**Before:**

before2

**After:**

after2

By the way, someone indicated a call to a nonexistent lexicon key "policy_err_name_ns"

### Related issue(s)/PR(s)
I did not create a issue, everything seems to be clear here.

### DIFF
--- a/core/model/modx/processors/security/access/policy/create.class.php
+++ b/core/model/modx/processors/security/access/policy/create.class.php
@@ -30,7 +30,7 @@ class modAccessPolicyCreateProcessor extends modObjectCreateProcessor {
     public function beforeSet() {
         $name = $this->getProperty('name');
         if (empty($name)) {
-            $this->addFieldError('name',$this->modx->lexicon('policy_err_name_ns'));
+            $this->addFieldError('name',$this->modx->lexicon('field_required'));
         }
 
         if ($this->doesAlreadyExist(array('name' => $name))) {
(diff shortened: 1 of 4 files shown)


opengeek:
I don't understand the issue being solved here. Can someone that understands better describe it?

ghost:
Apparently if you try to create a Policy with the same name as an existing Policy, you get the ambiguous Javascript error. The PR improves the duplicate name error.

The second part adds an asterisk by the label of the required fields.

GulomovCreative:
> I don't understand the issue being solved here. Can someone that understands better describe it?

My English is not clear?)

Mark-H (approved):

Jako:
Maybe we could create that required asterix with a css class + rule. Since that could work better with rtl text. But thats a different issue.

Jako (approved):

GulomovCreative:
> Maybe we could create that required asterix with a css class + rule. Since that could work better with rtl text. But thats a different issue.

Yes, this is good idea

Issue #14437, opened 2019-02-28
[bug] In the description of the system settings html tags are not processed


Ruslan-Aleev:
### Summary
In version 2.7.1, there was a lot of PR on protection from XSS with html-escaping tag output.
Now in the system settings, in the description, the tags are not processed, see the picture below.

![setting_error](https://user-images.githubusercontent.com/12523676/53592149-1d653b80-3baf-11e9-8fb0-f0fc84c1922d.png)

### Environment
MODX 2.7.1

Mark-H:
Unless we reintroduce XSS issues, this can't easily be fixed.

Jako:
A striptags could fix the current output.

Ruslan-Aleev:
@Mark-H it seems the issue has been fixed through https://github.com/modxcms/revolution/pull/14453

Ibochkarev:
Fixed in #14453
@modxbot close

Mark-H:
Glad I was proven wrong on this issue ;)

PR #14453, opened 2019-03-03
Allow safe html in system setting description

Jako:
### What does it do?
Create a safeHtml method in MODx.util, that strips all not allowed html tags, html attributes and javascript in the href attribute.

### Why is it needed?
Allow a reduced set of html attributes in the system setting grid.

It should be checked carefully, wether there could get some executable script though this.  

### Related issue(s)/PR(s)
#14437

### DIFF
--- a/manager/assets/modext/util/utilities.js
+++ b/manager/assets/modext/util/utilities.js
@@ -129,13 +129,22 @@ MODx.util.safeHtml = function (input, allowedTags, allowedAttributes) {
     var tags = /<\/?([a-z][a-z0-9]*)\b[^>]*>/gi,
         attributes = /([a-z][a-z0-9]*)\s*=\s*".*?"/gi,
         commentsAndPhpTags = /<!--[\s\S]*?-->|<\?(?:php)?[\s\S]*?\?>/gi,
-        hrefJavascript = /href(\s*?=\s*?(["'])javascript:.*?\2|\s*?=\s*?javascript:.*?(?![^> ]))/gi;
-    return input.replace(commentsAndPhpTags, '')
-        .replace(tags, function ($0, $1) {
+        hrefJavascript = /href(\s*?=\s*?(["'])javascript:.*?\2|\s*?=\s*?javascript:.*?(?![^> ]))/gi,
+        length;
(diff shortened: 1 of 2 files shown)


JoshuaLuckers (dismissed):

Mark-H:
Because the regex assumes valid markup, it's possible to evade this filter with a string that isn't valid markup:

```
<<SCRIPT>img src=/ onerror=alert('gotcha')</SCRIPT>>
```

(Added to the setting description, it does require the description to be expanded, but a better security researcher may know ways around that.)

I'm thinking it may be needed to call the filter recursively, but after just spinning my browser into an infinite loop I think it's getting too late for me to suggest an actual implementation :p

Jako:
Grr! But thanks for the test.

Jako:
The replacing iterates now. 

It misses a detection for attributes without a quote sign. I hope this could be handled.

Jako:
Those html attributes without a quote sign can't be stripped without deeper parsing. Now the html event attributes are invisible masked. I hope this is enough.

OptimusCrime:
Would it not be better to just encode all the special characters? Regexing "safe" content is next to impossible, and I bet there are more than a few ways to get around the stip function still.

Jako:
Some system settings have links and formatted text inside of the description. That is no more possible at the moment. The script is quite harsh against all invalid tags and I don’t think that an attack through this is possible (but I am not absolutely sure).

OptimusCrime:
Hm, okey. Well, I guess, if you really wanted to make it secure you would have to remove all the enhanced text then. This is by definition still exposed to XSS vulnerabilities, but I am sure this fix works good enough.

Jako:
What do you mean with enhanced text?

Jako:
The script strips all tags that are not allowed, it strips all tag attributes, that are not allowed. It removes javascript links in the href attribute and it inserts an invisible space in text with `onwhatever=` to make not stripped event attributes (because of a missing quote) invalid.

Ibochkarev:
@opengeek Please check this PR

Mark-H (approved):
I've not been able of bypassing this filter, good job 👍

JoshuaLuckers (approved):

Issue #14095, opened 2018-09-28
Parser fails on nested output filters


goldsky:
### Summary
Related issue: https://github.com/modxcms/revolution/issues/13250 .
Parser fails to process nested output filters, and returns truncated tags.

### Step to reproduce
Create a chunk called 'rowChunk' with this content:
```
[[+id:is=`1`:then=`[[+id]]`:else=`
    [[+id:is=`2`:then=`[[+id]]`:else=`
        [[+id:is=`3`:then=`[[+id]]`:else=`More`]]
    `]]
`]]
```

Then use the chunk as this:
```
[[!$rowChunk?
&id=`2`
]]
```
Try with ``` &id=`3` ```
### Observed behavior
It returns
```
More`]] `]] 
```

### Expected behavior
It should return the defined value.

### Environment
MODX 2.6.5-pl

Jako:
@bezumkin Do you have an idea, what could happen here?

bezumkin:
@Jako Problem is [here](https://github.com/modxcms/revolution/blob/e59f4e0de39adf5b3091ff0281d3062f4cdadb2f/core/model/modx/filters/modinputfilter.class.php#L54-L57)

This expression [couldn't handle nested conditions](https://regex101.com/r/vzttSe/1). Here what is happening:
![screenshot 2019-01-17 at 17 27 39](https://user-images.githubusercontent.com/1257284/51312188-4a391580-1a7d-11e9-8b67-390a74d0b11b.png)

I can't fix it and, frankly, I don't use anything like this because of Fenom.

JoshuaLuckers:
I will give it a try in the upcoming days 👍

goldsky:
I am not sure if this is related, but how about changing the MODX's parsing order for nested tags, from inner outwards to outer inwards?

That means, the logic condition should start from the first tag and stop immediately once it meets the condition, without going further into the deeper nested tag(s).

bezumkin:
@goldsky MODX parser always process tags inside out. Changing this will brake everything.

goldsky:
Yeah, I know that irony.
It shouldn't be at the first place.

JoshuaLuckers:
I submitted a PR to fix this issue. Feel free to test the changes and provide feedback.

Ibochkarev:
Fixed #14458 
@goldsky Thanks
@modxbot close

PR #14458, opened 2019-03-05
Fix the parser when using nested output filters

JoshuaLuckers:
### What does it do?
Order the matching results in the order they match and don't include the full matched string. This way to parser doesn't break.

### Why is it needed?
It fixes a problem where nested output filters might break in certain situations.

### Related issue(s)/PR(s)
Fixes issue #14095

### DIFF
--- a/_build/test/Tests/Model/Element/modChunkTest.php
+++ b/_build/test/Tests/Model/Element/modChunkTest.php
@@ -93,6 +93,10 @@ class modChunkTest extends MODxTestCase {
             ['<p>Hello, Mark!</p>', ['name' => 'Mark']],
             ['<p>Having fun.</p>', [],'<p>Having fun.</p>'],
             ['<p>Test 1</p>', ['number' => 1],'<p>Test [[+number]]</p>'],
+            ['1', ['id' => '1'], '[[+id:is=`1`:then=`[[+id]]`:else=`[[+id:is=`2`:then=`[[+id]]`:else=`[[+id:is=`3`:then=`[[+id]]`:else=`More`]]`]]`]]'],
+            ['2', ['id' => '2'], '[[+id:is=`1`:then=`[[+id]]`:else=`[[+id:is=`2`:then=`[[+id]]`:else=`[[+id:is=`3`:then=`[[+id]]`:else=`More`]]`]]`]]'],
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
This PR is a work in progress.

Jako:
All tests are failing at the moment …

JoshuaLuckers:
>All tests are failing at the moment … 

Yes, forgot to make this PR a draft (can’t be changed afterwards).

JoshuaLuckers:
@cla-bot check

Jako:
I don't get the clue on the change of the regular expression. What do you test with this? It includes the next signs until a colon or a equal sign or the full pattern. This does not sound right to me.

The regular expression that retrieves the content of the surrounding backticks after a modifier has to include everything and can't stop at the first nested backtick of a MODX tag inside. So it 'just' has to ignore the content between all nested MODX tags. I don't know if that's possible in a direct way, since you have to detect opening and closing nested pairs of MODX tags.

JoshuaLuckers:
@Jako to see what it does: https://regex101.com/r/N6dGM0/1

You're right, what's going on here isn't right. I added a new chunk of test data to the Unit Test (without this change the test failed) to make sure I didn't break anything. 
Back then I didn't know it would have impact elsewhere and break things (there was also no Unit Test testing the other part).

As you can see in issue #16186 I've now added a Unit Test in the right place, in my local repo I added test data that fails the Unit Test: https://github.com/JoshuaLuckers/revolution/blob/004b877ad0e597e77b19d4a6a09477495bdde4ad/_build/test/Tests/Model/Filters/modInputFilterTest.php

> I don't know if that's possible in a direct way, since you have to detect opening and closing nested pairs of MODX tags.
Correct, I spend several hours to fix it but it's not as easy as it seems.

 I'm in favor of reverting this PR, what do you think @opengeek ?

Jako:
If I am right, this issue occurs, when the MODX tags inside of the modifier chain options are not replaced. Normally the parser should have replaced them already. But if they are uncached, this is not the case. Can there be are other reasons?

I am also in favor of reverting this PR.

rthrash:

https://community.modx.com/t/modx-3-generated-don-t-cache-tags-seem-not-to-work-anymore/5247/21

rthrash:

https://community.modx.com/t/modx-3-generated-don-t-cache-tags-seem-not-to-work-anymore/5247/23

PR #14459, opened 2019-03-05
Fix using wrong option for upload_images

Jako:
### What does it do?
Fix using the wrong option for checkFiletype

### Why is it needed?
The wrong option was used before.

### Related issue(s)/PR(s)
Detected by a review of #14433


Mark-H (approved):

Jako:
Maybe it is in the S3 code too.

Jako:
I better remove the branch here and work in my repository.

Mark-H:
Hm? Why do you need to remove the branch? If it's also in the S3 code, you could just fix it in that and either force push or just add an extra commit :)

Jako:
I have accidentally created the branch on modxcms/revolution instead of Jako/revolution by directly editing the file on the modxcms/revolution on GitHub.

Mark-H:
Ahhhh, I hadn't noticed that. :P

PR #14460, opened 2019-03-05
Fix using wrong option for upload_images

Jako:
### What does it do?
Using the right option for upload_images.

### Why is it needed?
Fix using wrong option for upload_images.

### Related issue(s)/PR(s)
Detected during a review of #14433


Jako:
Shit. There seems to be CRLF issues in the S3 media source code.

Mark-H (approved):

JoshuaLuckers (approved):

Issue #14006, opened 2018-07-19
MODX 3: Buttons - different button groups


jonleverrier:
### Summary
We have a great new button pattern when editing resources;

![screen shot 2018-07-19 at 21 15 06](https://user-images.githubusercontent.com/2373940/42964975-e7d8513a-8b98-11e8-8af5-198638651858.png)

### Observed behavior
I wonder if this should be applied to all button groups to make it consistent?

**button group when editing template, tv, chunk, snippet, plugin:**
/manager/?a=element/template/update&id=1
/manager/?a=element/tv/update&id=1
/manager/?a=element/chunk/update&id=1
/manager/?a=element/snippet/update&id=1
/manager/?a=element/plugin/update&id=1
![screen shot 2018-07-19 at 21 16 51](https://user-images.githubusercontent.com/2373940/42965053-2a1224fe-8b99-11e8-83c7-f9f5da1fc5b1.png)

**button group when updating a user:**
/manager/?a=security/user/update&id=1
![screen shot 2018-07-19 at 21 20 09](https://user-images.githubusercontent.com/2373940/42965184-92a97120-8b99-11e8-9974-c878f1482afa.png)

**button group when editing a form customisation set:**
/manager/?a=security/user/update&id=1
![screen shot 2018-07-19 at 21 20 50](https://user-images.githubusercontent.com/2373940/42965232-ad2bc8a4-8b99-11e8-915d-9eda1f2a87ac.png)

**button group when updating a dashboard group:**
/manager/?a=system/dashboards/update&id=1
![screen shot 2018-07-19 at 21 20 50](https://user-images.githubusercontent.com/2373940/42965282-cf1d5cb6-8b99-11e8-96c1-374083e89508.png)

**button group when updating a context**
/manager/?a=context/update&key=web
![screen shot 2018-07-19 at 21 36 59](https://user-images.githubusercontent.com/2373940/42965995-fa8e950c-8b9b-11e8-80c3-7748c3cdc001.png)

**button group when updating a user group**
/manager/?a=security/usergroup/update&id=1
![screen shot 2018-07-19 at 21 38 51](https://user-images.githubusercontent.com/2373940/42966066-353ca3ba-8b9c-11e8-9171-3c1e6cbf4df8.png)

**button group when updating an access policy / policy template**
/manager/?a=security/access/policy/update&id=2
/manager/?a=security/access/policy/template/update&id=1
[...]

PR #14463, opened 2019-03-06
Using a single pattern for elements action buttons like a resource

GulomovCreative:
### What does it do?
Uses a single pattern for action buttons like a resource.

screenshot-revolution-2019 03 03-18-35-08
screenshot-revolution-2019 03 03-18-41-35

### Why is it needed?
On the pages of creating and updating resources, some buttons are hidden in one button with ellipsis. My PR copies them for elements.

### Related issue(s)/PR(s)
None


Ibochkarev:
Related to #14006

Mark-H (approved):
I like this. When merging, each action should be carefully checked to make sure there are no dangling commas or something that's hard to spot from the diff alone.

JoshuaLuckers (approved):

PR #14464, opened 2019-03-06
Fix icon layout in action buttons

GulomovCreative:
### What does it do?
Fix icon layout in action buttons.

### Why is it needed?
Icons have different widths and because of this they are not aligned.

**Before:**

screenshot-revolution-2019 03 04-01-41-35

**After:**

screenshot-revolution-2019 03 04-01-39-59

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1456,7 +1456,7 @@ iframe[classname="x-hidden"] {
       .icon {
         text-align: center;
         margin-left: 10px;
-        width: 18px;
+        @extend .#{$fa-css-prefix}-fw;
       }
     }
   }
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1456,6 +1456,7 @@ iframe[classname="x-hidden"] {
       .icon {
         text-align: center;
         margin-left: 10px;
+        width: 18px;
       }
     }
   }


Mark-H (dismissed):

Jako (dismissed):

_build/templates/default/sass/index.scss
```diff
@@ -1456,6 +1456,7 @@ iframe[classname="x-hidden"] {
       .icon {
         text-align: center;
         margin-left: 10px;
+        width: 18px;
```
JoshuaLuckers:
```suggestion
        @extend .#{$fa-css-prefix}-fw;
```

To prevent unnecessary changes down the road It would be better to use the fixed-width utility FontAwesome provides. If my suggestion is approved you can also remove the `text-align: center;`.

JoshuaLuckers (changes requested):

_build/templates/default/sass/index.scss
Jako: Thats better!

JoshuaLuckers (approved):

JoshuaLuckers:
Thanks @GulomovCreative 👍

GulomovCreative:
> Thanks @GulomovCreative 👍

Not at all, thank you @JoshuaLuckers

Jako (approved):

Mark-H (approved):

Mark-H:
Merged, thanks.

PR #14466, opened 2019-03-06
[ReUp] Fix image content being too wide for the dashboard widget.

muzzwood:
### What does it do?
Sets max-width on img tags within the body of a dashboard-block. 

### Why is it needed?
Currently images in the feed are not resized to the container and this means horizontal scrollbars are appearing. Adding `max-width:100%;` to the img tags allows images to fit properly.
See image:
screenshot_2019-03-06 dashboard modx revolution

### Related issue(s)/PR(s)
None

This PR changes a SASS file, so the CSS will need to be regenerated.

### DIFF
--- a/_build/templates/default/sass/_dashboard.scss
+++ b/_build/templates/default/sass/_dashboard.scss
@@ -87,6 +87,9 @@
       list-style: circle outside;
       padding: 0 0 0 15px;
     }
+    img {
+      max-width:100%;
+    }
     .draggable {
       cursor: move;
     }


Ibochkarev:
I would add another property for the image `height: auto;`

muzzwood:
`auto` is the default setting isn't it?

Jako (approved):

Ibochkarev:
https://getbootstrap.com/docs/4.0/content/images/ for example, it also uses both properties. But can be to leave and so. This not necessarily =)

JoshuaLuckers (approved):

JoshuaLuckers:
> `auto` is the default setting isn't it?

Correct, `auto` is the default value for the `height` property.

Ibochkarev (approved):

Mark-H (approved):

Ibochkarev:
Fixing one thing - we broke another. It was necessary to apply this change to` #modx-news-feed-container`.
screencast 2019-03-15 19-12-42

Now the display of the avatar in another widget has broken.

I will fix this bug.

PR #14469, opened 2019-03-07
Set default for check_filetype

Jako:
### What does it do?
Set a default for the check_file type option.

### Why is it needed?
The check_file type option introduced with #14433 could be overridden by a (normally not existing) xPDO option. This should be quite impossible after that patch.

### Related issue(s)/PR(s)
#14433

### DIFF
--- a/core/xpdo/compression/xpdozip.class.php
+++ b/core/xpdo/compression/xpdozip.class.php
@@ -40,7 +40,9 @@ class xPDOZip {
 
     public $xpdo = null;
     protected $_filename = '';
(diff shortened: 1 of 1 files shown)


Mark-H (changes requested):
My bad for not catching this on the earlier PR, but xPDO file changes should be made in xPDO first and can then be incorporated back to MODX. Could you do a PR to https://github.com/modxcms/xpdo that covers the changes needed for this feature?

Jako:
I will prepare a PR. But this way we create another merging bottleneck.

Issue #14470, opened 2019-03-08
MODX 3: Lexicon (login.inc.php) "Good night" should be "Good evening" in English. 


muzzwood:
### Summary
"Good night" isn't used in contemporary English as a greeting. It's typically used as a farewell.

### Step to reproduce
Go to the manager login page.

### Observed behavior
n/a

### Expected behavior
The English lexicon value for `$_lang['login_greeting_night']` should be changed to `<strong>Good evening</strong>, welcome back!`

I would have done a PR for this but I'm not sure if changes are made directly or through the Crowdin system. (?)

Ibochkarev:
You can send a PR to correct, then the guys will synchronize with Crowdin for translation into other languages

muzzwood:
Thanks, will do!

Jako:
I would prefer 'Hello Nightowl' ;-)

Serious: In german the meaning of `good night` is similar than in english. So we should better think what could be set as welcome string in the night or completely remove that entry.

ghost:
I think a generic greeting would be the better idea. If someone wants a time- or season-specific greeting that is easy enough to create.

muzzwood:
Haha night owl works for me! 
My line of thinking was that perhaps some languages need differentiation between evening and night whereas English doesn't. So, in that case it'd be best to keep the entry and just double up the values on the English (and German?) side.

PR #14471, opened 2019-03-08
Good night changed to Good evening on the login screen. (en)

muzzwood:
### What does it do?
Changes the English lexicon file for the manager login screen. Good night has been replaced with Good evening.

### Why is it needed?
 Good night is used as a farewell rather than a greeting in contemporary English.

### Related issue(s)/PR(s)
#14470


muzzwood:
Hmm I switched branch for this new PR but it seems to be including the commit from my previous PR. 
What's the best course of action to prevent that?

Edit:
A bit of research seems to suggest I should've used `git stash` before switching to the new branch. Is that correct?

_build/templates/default/sass/_dashboard.scss
```diff
@@ -87,6 +87,9 @@
       list-style: circle outside;
       padding: 0 0 0 15px;
     }
+    img {
+      max-width:100%;
```
JoshuaLuckers: I don't think this should be part of this PR.

JoshuaLuckers (changes requested):

JoshuaLuckers:
> Hmm I switched branch for this new PR but it seems to be including the commit from my previous PR.
> What's the best course of action to prevent that?
> 
> Edit:
> A bit of research seems to suggest I should've used `git stash` before switching to the new branch. Is that correct?

Always create a new branch (branch from 3.x in this case) for PR's.

JoshuaLuckers:
If you pull in the latest changes from 3.x you probably don't have to change anything 😄

Jako:
The following should work:

1. Fetch all changes from all remotes
2. Stash your changes
3. Checkout a new branch on base of 3.x
4. Rebase that branch with upstream/3.x
5. Pop your changes from the stash

```
git fetch --all
git stash
git checkout -b myFeature 3.x 
git rebase upstream/3.x
git stash pop
```

(This depends on that you have set up an upstream remote to modxcms/revolution)

After this, you could commit your stashed changes and create a PR.

muzzwood:
Thanks all, I'm going to re-do this one.

PR #14482, opened 2019-03-12
Update subMenuBuilder so "installer-entry" is always at the top

### LINKED ISSUE #11694: Make Installer always top menu item
Screenshot: http://sepiariver.co/1nF7Xav

The "Installer" is arguably the most important or at least it's different enough from the installed components that it should some how be demarcated.

sebastian-marinescu:
Due to the default-value of menuindex in menu-entries of newly installed extras being most of the time "0", the installer entry will jump up and/or down after an extra gets installed/updated. 

Therefore it should at least be fixed at the bottom of the dropdown-menu. 

Fixes #11694 pragmatically

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -502,7 +502,7 @@ $_lang['setting_new_folder_permissions'] = 'New Folder Permissions';
 $_lang['setting_new_folder_permissions_desc'] = 'When creating a new folder in the File Manager, the File Manager will attempt to change the folder permissions to those entered in this setting. This may not work on some setups, such as IIS, in which case you will need to manually change the permissions.';
 
 $_lang['setting_package_installer_at_top'] = 'Pin Package-Installer at top';
-$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the package-installer menu-entry inside the Extras-menu will be pinned at the top - regardless of its "menuindex"-attribute. If disabled, the package-installer will be positioned according to its menuindex.';
+$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the Installer entry will be pinned to the top of the Extras menu. Otherwise it will be positioned according to its menuindex.';
 
 $_lang['setting_parser_recurse_uncacheable'] = 'Delay Uncacheable Parsing';
 $_lang['setting_parser_recurse_uncacheable_desc'] = 'If disabled, uncacheable elements may have their output cached inside cacheable element content. Disable this ONLY if you are having problems with complex nested parsing which stopped working as expected.';
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -501,6 +501,9 @@ $_lang['setting_new_file_permissions_desc'] = 'When uploading a new file in the 
 $_lang['setting_new_folder_permissions'] = 'New Folder Permissions';
 $_lang['setting_new_folder_permissions_desc'] = 'When creating a new folder in the File Manager, the File Manager will attempt to change the folder permissions to those entered in this setting. This may not work on some setups, such as IIS, in which case you will need to manually change the permissions.';
 
+$_lang['setting_package_installer_at_top'] = 'Pin Package-Installer at top';
+$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the package-installer menu-entry inside the Extras-menu will be pinned at the top - regardless of its "menuindex"-attribute. If disabled, the package-installer will be positioned according to its menuindex.';
+
 $_lang['setting_parser_recurse_uncacheable'] = 'Delay Uncacheable Parsing';
 $_lang['setting_parser_recurse_uncacheable_desc'] = 'If disabled, uncacheable elements may have their output cached inside cacheable element content. Disable this ONLY if you are having problems with complex nested parsing which stopped working as expected.';
 
--- a/core/model/modx/modmenu.class.php
+++ b/core/model/modx/modmenu.class.php
@@ -101,6 +101,8 @@ class modMenu extends modAccessibleObject {
         $c->where(array(
             'modMenu.parent' => $start,
         ));
+        // make sure installer is always on top
+        $c->sortby('FIELD(modMenu.text, "installer")','DESC');
         $c->sortby($this->xpdo->getSelectColumns('modMenu','modMenu','',array('menuindex')),'ASC');
         $menus = $this->xpdo->getCollection('modMenu',$c);
         if (count($menus) < 1) return array();
--- a/core/model/modx/modmenu.class.php
+++ b/core/model/modx/modmenu.class.php
@@ -101,8 +101,12 @@ class modMenu extends modAccessibleObject {
         $c->where(array(
             'modMenu.parent' => $start,
         ));
-        // make sure installer is always on top
-        $c->sortby('FIELD(modMenu.text, "installer")','DESC');
+        
+        if ($this->xpdo->getOption('package_installer_at_top', null, true)) {
+            // make sure installer is always on top
+            $c->sortby('FIELD(modMenu.text, "installer")','DESC');
+        }
+        
         $c->sortby($this->xpdo->getSelectColumns('modMenu','modMenu','',array('menuindex')),'ASC');
         $menus = $this->xpdo->getCollection('modMenu',$c);
         if (count($menus) < 1) return array();
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -165,7 +165,7 @@ $topNavMenus[2]->fromArray(array (
 /* Installer */
 $children[0]= $xpdo->newObject('modMenu');
 $children[0]->fromArray(array (
-  'menuindex' => 10,
+  'menuindex' => 0,
   'text' => 'installer',
   'description' => 'installer_desc',
   'parent' => 'components',
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -165,7 +165,7 @@ $topNavMenus[2]->fromArray(array (
 /* Installer */
 $children[0]= $xpdo->newObject('modMenu');
 $children[0]->fromArray(array (
-  'menuindex' => 0,
+  'menuindex' => 10,
   'text' => 'installer',
   'description' => 'installer_desc',
   'parent' => 'components',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1160,6 +1160,15 @@ $settings['modx_charset']->fromArray(array (
   'area' => 'language',
   'editedon' => null,
 ), '', true, true);
+$settings['package_installer_at_top']= $xpdo->newObject('modSystemSetting');
+$settings['package_installer_at_top']->fromArray(array (
+  'key' => 'package_installer_at_top',
+  'value' => true,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'manager',
+  'editedon' => null,
+), '', true, true);
 $settings['principal_targets']= $xpdo->newObject('modSystemSetting');
 $settings['principal_targets']->fromArray(array (
   'key' => 'principal_targets',
--- a/core/lexicon/de/setting.inc.php
+++ b/core/lexicon/de/setting.inc.php
@@ -501,9 +501,6 @@ $_lang['setting_new_file_permissions_desc'] = 'Nach dem Hochladen einer neuen Da
 $_lang['setting_new_folder_permissions'] = 'Verzeichnisrechte für neue Verzeichnisse';
 $_lang['setting_new_folder_permissions_desc'] = 'Nach dem Anlegen eines neuen Ordners im Dateimanager versucht dieser, die Verzeichnisrechte in die zu ändern, die in dieser Einstellung gespeichert sind. Dies könnte in einigen Konfigurationen evtl. nicht funktionieren, z.B. bei Verwendung des IIS-Webservers. In diesem Fall müssen Sie die Rechte selbst ändern.';
 
-$_lang['setting_package_installer_at_top'] = 'Package-Verwaltung oben fixieren';
-$_lang['setting_package_installer_at_top_desc'] = 'Wenn aktiviert, wird der Package-Verwaltung-Menüeintrag im Extras-Menü oben fixiert – unabhängig von seinem "menuindex"-Attribut. Wenn deaktiviert, wird die Package-Verwaltung entsprechend ihrem Menüindex positioniert.';
-
 $_lang['setting_parser_recurse_uncacheable'] = 'Parsen nicht gecachter Elemente verzögern';
 $_lang['setting_parser_recurse_uncacheable_desc'] = 'Wenn diese Option deaktiviert ist, kann die Ausgabe von Elementen, die nicht gecacht werden sollen (Ausrufungszeichen am Anfang des MODX-Tags, siehe <a href="https://rtfm. modx. com/revolution/2.x/making-sites-with-modx/tag-syntax#TagSyntax-Caching" target="_blank">MODX-Dokumentation</a>), innerhalb des Inhalts von gecachten Elementen gecacht werden. Deaktivieren Sie diese Option NUR DANN, wenn Sie Probleme mit dem Parsen von komplexen verschachtelten Elementen haben und dieses nicht wie erwartet funktioniert.';
 
--- a/core/lexicon/de/setting.inc.php
+++ b/core/lexicon/de/setting.inc.php
@@ -501,6 +501,9 @@ $_lang['setting_new_file_permissions_desc'] = 'Nach dem Hochladen einer neuen Da
 $_lang['setting_new_folder_permissions'] = 'Verzeichnisrechte für neue Verzeichnisse';
 $_lang['setting_new_folder_permissions_desc'] = 'Nach dem Anlegen eines neuen Ordners im Dateimanager versucht dieser, die Verzeichnisrechte in die zu ändern, die in dieser Einstellung gespeichert sind. Dies könnte in einigen Konfigurationen evtl. nicht funktionieren, z.B. bei Verwendung des IIS-Webservers. In diesem Fall müssen Sie die Rechte selbst ändern.';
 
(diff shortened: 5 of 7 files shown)


Mark-H (changes requested):
With all due respect and appreciation for opening a PR, I don't consider putting it at the bottom an improvement. [Posted an alternative suggestion in the issue](https://github.com/modxcms/revolution/issues/11694#issuecomment-472009581).

sebastian-marinescu:
> With all due respect and appreciation for opening a PR, I don't consider putting it at the bottom an improvement. [Posted an alternative suggestion in the issue](https://github.com/modxcms/revolution/issues/11694#issuecomment-472009581).

Changed the behavior to suggested one.

JoshuaLuckers:
@sebastian-marinescu did you look if this was possible?

> What I'd suggest we try (I haven't looked at the relevant code yet) is to add some extra code into the menu builder that makes sure the installer is always first. 
> 
> Something like an `ORDER BY field(modMenu.key, 'installer') ASC` (but in xPDO) before the menuindex sort is applied.

_Originally posted by @Mark-H in https://github.com/modxcms/revolution/issues/11694#issuecomment-472009581_

Mark-H:
Hey @sebastian-marinescu - did you determine sorting in the SQL as I suggested wasn't possible and that the html-replacement technique is best? Just hoping for some feedback from you here as I'm not clear on what this PR is at and 2.7.2 is coming soon.

Ibochkarev:
Fixed #14494 

@modxbot close

Ibochkarev:
@modxbot open

SnowCreative:
Even better would be to make the installer/updater scripts NOT CHANGE the menuindex on already installed items. Every time an extra gets updated, its position in the menu changes. Whatever order has been set by users should be retained after updates. New extras could go at the bottom, and then people could drag them where they want in the menu.

The same thing goes for menu parents. For example, I use the ClientConfig extra, and it makes the most sense to me and my clients to keep that under the Site (Content) menu. Every update moves it back under the Extras menu, and I have to move it back where I want it manually.

Mark-H:
That's not something the core does - that's each extra saying their menu object should be updated.

Jako:
So it is a package build issue. And these methods are in the core.

Ibochkarev:
@Mark-H @Jako Guys, what in the end on this PR needs to be done?

Mark-H:
@Ibochkarev [In the related issue](https://github.com/modxcms/revolution/issues/11694#issuecomment-472009581) I posted an alternative solution before this PR was created that handled it in the menu generation and I've asked if @sebastian-marinescu tried that or not. If my idea, which seems like a cleaner solution, was not feasible I'd happily merge this but I'm not sure if it was tried.

sebastian-marinescu:
Sorry for ghosting people. 

I just wasn't sure back then where/how the menu gets grabbed. 
Trying to re-track it, I think I found it: 

- The [modManagerController](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modmanagercontroller.class.php#L178) loads the header, which is just an include of the [header.php](https://github.com/modxcms/revolution/blob/2.x/manager/controllers/default/header.php) in the default manager-theme

- Inside the header.php the [`buildMenu`](https://github.com/modxcms/revolution/blob/2.x/manager/controllers/default/header.php#L150)-method gets the [menus from cache](https://github.com/modxcms/revolution/blob/2.x/manager/controllers/default/header.php#L225)

- If it's not already cached, it [rebuilds the menu-cache](https://github.com/modxcms/revolution/blob/2.x/manager/controllers/default/header.php#L242)

- Inside the `getSubMenus`-method it [sorts by `menuindex`](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modmenu.class.php#L104)

So this is where you would add an extra sort-criteria @Mark-H ? 
```
$c->sortby($this->xpdo->getSelectColumns('modMenu','modMenu','',array('menuindex')),'ASC');
``` 
FYI: @Ibochkarev

Mark-H:
Yep, that's what I was hoping for. :)

sebastian-marinescu:
So I think I found the solution and pushed it to my [patch-3](https://github.com/sebastian-marinescu/revolution/tree/patch-3) branch - but it doesn't show here on the PR like it used to do...

Anyone any idea how I can update the commits here?

sebastian-marinescu:
Oh okay, the friendly cla-bot unfroze the situation - nice :) 
Hopefully this is the right way now @Mark-H @opengeek

Jako:
Would it be possible to have folders first (maybe by an option)?

JoshuaLuckers:
I would be happy with this alone!

sebastian-marinescu:
> Would it be possible to have folders first (maybe by an option)?

I'm not sure I understand. Do you mean for the two entries in "Manage"? 
image

Would probably be possible, by checking if it is a parent - but I don't want to mix this PR up: I just want the "Installer" entry to not jump around anymore, and I'm hoping to get that into the next release.

Jako:
I often put folders first in the Extras menu to avoid things like in this screenshot at the bottom of the browser window.

Cut off menu

With your patch a submenu could start only at the second entry, which is a little less space to the bottom. it would be great, if entries with a submenu could stay in the before the Installer (by enabling a system setting).

Ruslan-Aleev:
> With your patch a submenu could start only at the second entry, which is a little less space to the bottom. it would be great, if entries with a submenu could stay in the before the Installer (by enabling a system setting).

Maybe we should add this fix to the 2.x branch - https://github.com/modxcms/revolution/pull/14300? And for 3.x right away.

JoshuaLuckers (approved):
I think it's a good idea to solve this little annoyance this way. In 3.x we can improve it further by implementation one of the suggestions.

sebastian-marinescu:
> With your patch a submenu could start only at the second entry, which is a little less space to the bottom. it would be great, if entries with a submenu could stay in the before the Installer (by enabling a system setting).

@Jako I see now, thx. So your setting should/would only be relevant to the "Extras" main-menu or to all? 
And while I understand your use-case, I'd still be in favor of having the "installer" as first, *always* - then your special-folders. It would still work. 

Another thing I wanted to point out is, that your component descriptions take a lot of space away - 
maybe you could check out "AdminTools", which hides the descriptions and therefore gives much more space: 

image

SnowCreative:
I really don't understand why the consensus is that the installer should always be at the top. Once a site is set up, that is going to be the LEAST USED item in the list. Putting it at the bottom, or else just sorting it like everything else, makes a lot more sense to me.

Ruslan-Aleev:
> Once a site is set up, that is going to be the LEAST USED item in the list.

Personally, I am enraged that the "Installer" jumps in menu :)
And while you **set up the site once**, you get tired of looking for the installer in the menu.

SnowCreative:
> > Once a site is set up, that is going to be the LEAST USED item in the list.
> 
> Personally, I am enraged that the "Installer" jumps in menu :)
> And while you **set up the site once**, you get tired of looking for the installer in the menu.

If it's sorted alphabetically, you can find it. If it's always at the bottom, you can find it. When everything is jumping around, as they do now, you can't find any of the plugins, including the installer.

Ruslan-Aleev:
> If it's sorted alphabetically, you can find it. If it's always at the bottom, you can find it.

- Alphabetically, it is located only for the English language.
- The bottom location is not convenient, because the menu item can be clipped by the screen (@Jako pointed out about this above).
- **Positioning on top** is the most effective UX option, in my opinion.

SnowCreative:
> > If it's sorted alphabetically, you can find it. If it's always at the bottom, you can find it.
> 
>     * Alphabetically, it is located only for the English language.
> 
>     * The bottom location is not convenient, because the menu item can be clipped by the screen (@Jako pointed out about this above).
> 
>     * **Positioning on top** is the most effective UX option, in my opinion.
If it gets clipped by the screen at the bottom, then moving it to the top will make something else get clipped by the screen instead. That's a separate issue, which also needs to get fixed. As for alpha sorting, the sorting would have to be done depending on the language selected for the manager itself. There is a separate thread, I believe, asking for alpha sorting of plugins rather than using menuindex as is now done, and which is responsible for making the menu a mess since the menuindexes get reset with every update. Any custom permission you've put on a plugin also gets wiped out when the plugin is updated, which is yet another issue.

Why can't we just install a system setting to determine where to put the installer, instead of trying to force everyone to have it in the same place?

Or, better yet, why not just **remove the possibility for package updates to reset the menuindexes at all**. That would fix everything, and everybody could position the plugins and installer exactly where they want them, without them jumping around after every update.

Ruslan-Aleev:
>  As for alpha sorting, the sorting would have to be done depending on the language selected for the manager itself.

Sorting will be done by lexicon key, not translation. Make sorting by translation - is a difficult task.

> Why can't we just install a system setting to determine where to put the installer, instead of trying to force everyone to have it in the same place?

In my opinion, the fact that the menu will be on top suits most of the users (based on the discussion in the issue/PR), I didn't think there would be any disputes :)

sebastian-marinescu:
> I really don't understand why the consensus is that the installer should always be at the top. Once a site is set up, that is going to be the LEAST USED item in the list. Putting it at the bottom, or else just sorting it like everything else, makes a lot more sense to me.

@rainbowtiger 
I strongly disagree, as it still remains the most important menu-link inside the components-menu (e. g. for maintenance/updates or by the fact that a website is constantly growing and in need of new or swapped extras over time). 
The corresponding issue #11694 has been around *for years*, and the consensus is clear. 

That's why I also feel like @Ruslan-Aleev when it jumps around.
> Personally, I am enraged that the "Installer" jumps in menu :)

And as it is the most important link, putting it as first is a good choice for UX, as also pointed out. 

However, you have valid points in your [comment](https://github.com/modxcms/revolution/pull/14482#issuecomment-660125660) about the different other issues, especially that maybe the *rest* of the menu should be sorted alphanumerically and not by menuindex 👍

rthrash:
Top +100. After you have set up a site you should still maintain your extras and keep them up to date. This makes that ever-so-slightly less inconvenient. 

I doubt a discussion about its location would have ever cropped up had it been pinned at the top from the beginning.

SnowCreative:
> > As for alpha sorting, the sorting would have to be done depending on the language selected for the manager itself.
> 
> Sorting occurs by lexicon key, not translation.
> 

If plugins were to be sorted alphabetically, they would be sorted by the lexicon values (the translations), not the "keys". No menus are sorted this way now, so this would have to be an added feature. The complexity of this approach is just another good argument for preventing plugin updates from altering the menuindexes at all, so that nothing would jump around after an update.

SnowCreative:
> I strongly disagree, as it still remains the most important menu-link inside the components-menu (e. g. for maintenance/updates or by the fact that a website is constantly growing and in need of new or swapped extras over time).

Keeping plugins updated is what is important, not the location of the installer menu item. It you know where that menu item is always going to be, regardless of location, then what's the problem? Adding or swapping plugins over time is a very occasional need. I use SiteDash to do my plugin updating across all sites, so it's not AT ALL important to me to have the installer at the top of the menu in individual sites. An intuitive and useful menu arrangement for my clients is far more important to me. I want to be able to find ALL of my plugins easily, not just the installer. And, I would argue that other plugins in my sites are just as "important" as the installer. 

> However, you have valid points in your [comment](https://github.com/modxcms/revolution/pull/14482#issuecomment-660125660) about the different other issues, especially that maybe the _rest_ of the menu should be sorted alphanumerically and not by menuindex 👍

Actually, alpha sorting is only a second-best option, as far as I'm concerned. Leaving the menuindexes alone is by far the most preferable, because then everybody could set up this menu however they want, and nothing would shift after updates. "Creative Freedom", right? For example, if I have a client who uses a particular plugin every day, I want that to be at the top of the menu, not the installer or some other plugin that comes before it alphabetically.

core/model/modx/modmenu.class.php
```diff
@@ -101,6 +101,8 @@ public function getSubMenus($start = '') {
         $c->where(array(
             'modMenu.parent' => $start,
         ));
+        // make sure installer is always on top
```
JoshuaLuckers: Let's give users a choice and make it optional (simply on/off) in 2.x.

JoshuaLuckers (changes requested):

core/model/modx/modmenu.class.php
Ruslan-Aleev: If we make a system setting, then it should be enabled by default. It doesn't make much sense, though. Most users favor placing the "Installer" at the top.

SnowCreative: Excellent! Great for in-the-minority folks like me who want it elsewhere.

JoshuaLuckers: And we should fix the issue with the menu indexes in 3.x and look into the other great suggestions!

Jako: Updating an extra resets not only the menuindex. It resets also the parent of menu entries. Both should stay the same.

SnowCreative: It also resets the permissions.

Jako: Ok, that should be moved to a different issue.

sebastian-marinescu: I added a new system-setting `package_installer_at_top` and an if-clause around the installer-sorting

sebastian-marinescu:
I realized this PR is still open, so I attempted to add a new system-setting `package_installer_at_top`, which puts the installer-menu-entry at the top, if activated.

Jako:
In case you don't know it: https://extras.modx.com/package/menukeeper

sebastian-marinescu:
> In case you don't know it: https://extras.modx.com/package/menukeeper

Nice, I didn't! I'm still hoping to get this PR through though 😄

SnowCreative:
> In case you don't know it: https://extras.modx.com/package/menukeeper

Looks great!

Jako (approved):

Jako:
The PR misses an english lexicon entry.

sebastian-marinescu:
I've added lexicon-entries for the languages I can handle. 

@Jako Can you please check again, if it's okay?

Jako:
English should be the only addition. The others are handled by Crowdin.

sebastian-marinescu:
Okay, missed that part. So should I delete the other 2 language-entries? Or can it stay there and will be overridden?

Jako:
You have to remove the de and ro lexicon files.

sebastian-marinescu:
Done! And thank you @Jako, it's my first time committing a PR with system-settings 👍

sebastian-marinescu:
Hi @opengeek, can we get this into [v2.8.6](https://github.com/modxcms/revolution/milestone/104)? 

And `port-to-v3` means someone would cherrypick or merge this also into 3.x or do I need to open a separate/new PR for 3.x?

opengeek:
> Hi @opengeek, can we get this into [v2.8.6](https://github.com/modxcms/revolution/milestone/104)?
> 
> And `port-to-v3` means someone would cherrypick or merge this also into 3.x or do I need to open a separate/new PR for 3.x?

We can get it into 2.8.6 if we get a consensus on approval (at least one more approving review). And I could attempt to cherry-pick it for 3.x, but it would be preferable if a PR explicitly targeting 3.x was submitted.

theboxer (dismissed):

sebastian-marinescu:
Yay!! 🎉 

Preparing another PR for 3.x now 👍

core/lexicon/en/setting.inc.php
```diff
@@ -501,6 +501,9 @@
 $_lang['setting_new_folder_permissions'] = 'New Folder Permissions';
 $_lang['setting_new_folder_permissions_desc'] = 'When creating a new folder in the File Manager, the File Manager will attempt to change the folder permissions to those entered in this setting. This may not work on some setups, such as IIS, in which case you will need to manually change the permissions.';
 
+$_lang['setting_package_installer_at_top'] = 'Pin Package-Installer at top';
+$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the package-installer menu-entry inside the Extras-menu will be pinned at the top - regardless of its "menuindex"-attribute. If disabled, the package-installer will be positioned according to its menuindex.';
```
smg6511: Same as my suggestion for the 3.x version of this PR, tweak to make more concise.
```suggestion
$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the Installer entry will be pinned to the top of the Extras menu. Otherwise it will be positioned according to its menuindex.';
```

smg6511 (changes requested):
Tested and can approve with requested lexicon change mentioned above (assuming others agree with my tweaks). And, yes I'm one who'd always like to see the Installer at the top, so I do like this feature ;-)

Jako:
I like to have some folders as the first entry and because of this I wrote the MenuKeeper extra.

sebastian-marinescu:
Should be good now, right? @opengeek @smg6511

smg6511 (approved):

sebastian-marinescu:
Thanks @smg6511 👍 

@opengeek 
It seems we got all approvals, can we stil get this into v2.8.6 pretty please?

sebastian-marinescu:
And... is something wrong with the cla-bot, or do I have to sign it again? 

image

sebastian-marinescu:
@Mark-H @opengeek @JoshuaLuckers @smg6511 

Can one of you, super-pretty please, label it with the [v2.8.6-milestone](https://github.com/modxcms/revolution/milestone/104), so it doesn't slip away?

Also I think the "port-to-v3" label isn't necessary anymore, as the v3-pr went through. 
~And out of curiosity: the label "blocked" is because of the currently not working cla-bot?~
And after the updates, it should also not be labeled "blocked" anymore.

sebastian-marinescu:
Thank you very much, Mark! Now I'm calm :))

opengeek:
@sebastian-marinescu — I have not yet integrated this PR because @Mark-H and @JoshuaLuckers still have requested changes and did not approve their reviews.

@Mark-H — are you approving this since you added the v2.8.6 label? I'm confused by the comment thread here.

sebastian-marinescu:
Thank you for the clarification @opengeek 
IIRC I solved all change requests, but we can wait for Mark and Joshua to confirm 👍

rthrash:
@opengeek I believe that @Mark-H is good with the update after all the changes made below his change request and his assigning the previous milestone. @JoshuaLuckers is now outdated, since the toggle was added to the system settings, so should be good to go. Mark, Joshua, please confirm if my review of the reviews is correct! :)

Mark-H (approved):

Issue #14457, opened 2019-03-05
The inconsistency of the names in the menu and in the section


Ruslan-Aleev:
### Summary
In the main menu item (and description) has one name, and the section is called differently. Moreover, the section may also differ in the section title and title in the browser.

![menu_title](https://user-images.githubusercontent.com/12523676/53797696-20b24b80-3f50-11e9-9289-1ff45da1521c.png)

Partially the problem was discussed in PR - https://github.com/modxcms/revolution/pull/14353

Related with https://github.com/modxcms/revolution/issues/13554

Issue #14291, opened 2019-01-17
[enhancement] MODX3: Reorganize the main menu


Ruslan-Aleev:
### Summary
The main menu in MODX has items that are non specific and do not give an understanding of the nested items, for example, "Manage" and the "Gear icon" on the right. Until you hover over on item - it is not clear what elements are inside.

There are also items that are connected, but located in different places, see the picture below.

![menu_info](https://user-images.githubusercontent.com/12523676/51331976-4b297180-1a94-11e9-9e8a-b6fba8089cb4.png)

In MODX3 we will need to think over the structure of the main menu, and reorganize the items.

JoshuaLuckers:
Related issues: 
- #13942
- #11694

Ruslan-Aleev:
**Possible structure of the main menu in MODX3, from left to right in MODX 2.x (in MODX3 from top to bottom).**

I was here trying to collect menu items based on logical connections, maybe in the future something will be added or removed. We need a discussion.

1) **Content:**
- Clear Cache / Refresh URIs (item now in "Manage->Clear Cache / Refresh URIs")
- Remove Locks (item now in "Manage->Remove Locks")
- Resource Schedule (item now in "Manage->Reports->Site Schedule")
- Content Types
- Import Static Resources (?)
- Import HTML (?)

Moved all the items that are related to the content in the one section.
Removed the menu items that are in the widget ("New resource", "Preview Site").
"Import Resources and HTML" perhaps will be removed from the core in the future (https://github.com/modxcms/revolution/issues/14216#issuecomment-449869312).

2) **Media:**
- Media Browser
- Media Sources

There are no changes.

3) **Packages (old name "Extras"):**
- Installer
- Packages Settings (item now in "Gear->System Settings")
- Packages Lexicons (item now in "Gear->Lexicons")
- Namespaces (item now in "Gear->Namespaces")

In MODX, the core is also a component, but which is already preinstalled. All the listed settings are related to the components (and the core, and installed from the outside), it is logical to combine them. Thats make it clear that the settings and lexicons of the package can change in the same system sections, but depending on the namespace (at the beginning of acquaintance with MODX it is not immediately clear).

4) **Access (new item):**
- Access Control Lists (item now in "Gear->Access Control Lists")
- Resource Groups (https://github.com/modxcms/revolution/pull/14855#issuecomment-652141921)
- Flush Your Permissions (item now in "Manage->Flush Your Permissions")
- Users (item now in "Manage->Users")
[...]

Issue #13554, opened 2017-07-17
Inconsistent terminology in MODX


davidpede:
### Summary
A number of new clients have had issue with inconsistent action button labels, confirmation + warning dialog, different labels for the same thing etc. Has caused confusion and annoyance with some of them.

Referring to things like:

- Delete action labelled **Delete** and **Remove** in different menus/buttons
- Edit action labelled  **Edit** and **Update** in different menus/buttons
- Create action labelled  **Create** and **New** in different menus/buttons
- Folders referred to as **Directory** and **Folder** in different places

- Warning/Confirmation dialogs not helpful and totally inconstant across resources, elements, file/folder, grids etc (especially delete)
- Warning dialog terminology doesn't match action label
(There is more but you get the idea)

I'm happy to look at this but can I get some feedback first:

- Do people think this is worth looking at?
- Is this the right place or should it go through the MAB manager group?
- Are there any guidelines for making changes to lexicon files?

### Related
#13065, #4834, #1188, #764, #13448, #13497

### Environment
MODX all versions

Mark-H:
- Yes
- While the MAB could take a role in this, don't let that stop you. Perhaps you'll end up draft a set of guidelines on what type of terminology to use. If you then find a member of the MAB that is on board with that to turn, they can act as editor for a recommendation and put it up for a MAB vote to make it "official".
- English lexicons are managed through this repo. Translations via CrowdIn. 

To propose changes you'd use the same process as other contributions. Create a git branch, make and commit your changes, send a pull request. I'd recommend branching of 2.x and also sending your changes to 2.x, so these updates could be incorporated into 2.6/2.7. 

The lexicon files are in https://github.com/modxcms/revolution/tree/2.x/core/lexicon/en. Having a search tool will help find where each entry is. ;)

Keep in mind lexicons are cached by default. Set cache_lexicon_topics to no so you wont have to clear the cache after each change.

sottwell:
The 'Delete' and 'Remove' labels are tricky; resources that are 'deleted' are just flagged in the database for deletion, while 'remove' actually refers to permanently deleting something from the database.

davidpede:
Not a simple task but I think it will come down to writing style and striking the balance between simple, consistent UX and technical accuracy.

Were in process of getting more formalised feedback from the clients that have raised these issues which will help a lot and material design will also influence the strategy. Some good reading:

https://material.io/guidelines/style/writing.html#writing-language
[...]

PR #14483, opened 2019-03-12
Corrected menu items and section headers

Ruslan-Aleev:
### What does it do?
Corrected menu items and section headers, buttons.
Corrected some inaccuracies in the menu: deleted points, somewhere there were points in the description, but somewhere not.
Corrected some inaccuracies in the import section: the buttons used the lexicon from the menu topic, and not the lexicon from the import topic, it was wrong.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14457


Ruslan-Aleev:
It seems to me that the item "Extras" is more correctly to call "Packages", because In this section, the packages will be installed.
But at the same time, in the description of the section "Namespaces", the "Add-ons" should be called as "Components", because in this section, the components are located (installed packages).

core/lexicon/en/topmenu.inc.php
```diff
@@ -17,7 +17,7 @@
 $_lang['api_docs_desc'] = 'Complete API documentation for MODX Revolution.';
 $_lang['bespoke_manager'] = 'Manager Customization';
 $_lang['bespoke_manager_desc'] = 'Manage bespoke MODX Managers';
-$_lang['components'] = 'Extras';
+$_lang['components'] = 'Packages';
```
opengeek, line 20: No. Extras are what could be installed by a package, but they are not the same thing.

opengeek (changes requested):
These commits include a lot of miscellaneous changes that should not be combined in my opinion.

core/lexicon/en/topmenu.inc.php
Ruslan-Aleev: Why so complicate? We can leave only "Packages" and "Components" - installed packages.
Why need for the entities "Extras" and "Add-ons"?

opengeek: Add-Ons/Extras/Components <-- these are interchangeable and different than packages. I'm fine with picking one of those, but replacing any of those with packages is simply wrong. It's not complicated at all.

opengeek: The more important issue is these things need to be discussed independently from all the miscellaneous changes included in this single PR.

Ruslan-Aleev: Issue hang in 2017, for example, https://github.com/modxcms/revolution/issues/13554 but there is no progress in the discussion.

JoshuaLuckers:
Changes are more likely to be accepted if you make a new PR for each "group" of changes. Especially when there are changes that need discussion. I believe it's also more clear what the definition of a "Extra" is.

Thanks again for helping make MODX better @Ruslan-Aleev 😎 !

Ruslan-Aleev:
@JoshuaLuckers Yes, about several groups that I have in vain combined - no questions, it is worth splitting into separate PRs.
But there is a question to the wording of @opengeek _"these things need to be discussed"_, there is a discussion, but Code Owners do not participate in the discussion, and even if they participate, the result of the discussion is not clear.
For example, here https://github.com/modxcms/revolution/issues/14291 there is a structure, there is a discussion, but there is no result of the discussion. We are waiting for more discussion or what? If there was a specific comment from Owners, we do this and that, or we leave it this way and that - this is the result, but instead we get a discussion for discussion. What is not productive, in my opinion.
Owners probably have a lot of work and do not have time to clearly state their position in the comments, but there are issue that have been hanging since 2009 and this is depressing.

Issue #14201, opened 2018-12-14
Improve getChunk to allow file based and inline tpl 


sdrenth:
### Summary
The getChunk method from pdoTools allows you to use file based and inline tpl's. So it doesn't require a modChunk reference in the database. 

I think the core getChunk method should be improved to allow also file based and inline TPLs and would like to know who else would like to see this improvement into the core of MODX.

pdoTools uses a system setting for the base path of where all static elements live, the new  core system setting `static_elements_basepath` could be used for this as well.

Examples:
```
'tplWrapper' => '@FILE /elements/chunks/menu/wrapper.tpl'

AND

'tplWrapper' => '@INLINE <ul>[[+wrapper]]</ul>`
```
### Why is it needed?
We like to work file based and for performance it would be better if we can bypass the database when working file based. And also for small templates it can be a lot easier to create an inline TPL then creating a new modChunk snippet for it in the database.

In short, it makes life easier and is also faster.

### Suggested solution(s)
I think the getElement method in the modParser class should be changed to also support file based and inline tpls. 

**What do you guys think about the idea and the suggested solution?**

Oetzie:
Maybe also use Fenom from pdoTools (or another template engine) for parsing templates and chunks instead of MODx tags/output filters?

JoshuaLuckers:
@Oetzie if that is something you would like to see I suggest creating a new issue/feature request instead.

mvoevodskiy:
I will do it.

PR #14490, opened 2019-03-16
Improve elements to allow file and inline bindings

mvoevodskiy:
### What does it do?
Added bindings @CODE, @INLINE, @FILE bindings with elements: snippets, chunk. This available for templates and plugins, but it's not callable as "$modx->getChunk() " / "$modx->runSnippet()".
Also, this bindings available in resource content or templates body and other parsible places. 

### Why is it needed?
This ability do simplier site development because you can not create chunks and snippets in database in all cases.

Also, fixed bug with static elements: now correctly check static file modofication time and if it's more than cache file, cache file will be rewrited.

Small fix PHP notice in modmediasource.class.php.

### How works `@FILE` binding
`@FILE` binding creates temporary static element (chunk, snippet, template). Then static element works as usual.   
You can use 2 system settings:
* __static_elements_default_mediasource__ for choose media source. If not selected, MODX wiil be use default media source ("Filesystem" or other)  
* __static_elements_basepath__ for specify elements directory from media source root.  

If you leave blank both settongs, specify full path for `@FILE` binding from site root.
If you fill __static_elements_basepath__ setting, path for `@FILE` binding should be relative from path in setting.

### How works `@INLINE` for chunks
You can use MODX tags in inline chunks with both: `[[ ... ]]` or  `{{ ... }}`
`{{ ... }}` need to support MODX tags in inline chunk when it specified at other chunk or template.

#### Files
_chunks/example.tpl_:
```
[[++site_name]]
<h2>Hello, [[+modx.user.username]]!</h2>
link to home page: [[~1]]
<p>[[+param1]]</p>
<p>[[+param100]]</p>
```

_snippets/example.php_:
```
<?php
return $modx->user->username . ' Param: ' . $snParam;
```

#### PHP
_chunks_:
```
$chunk = '@INLINE {{++site_name}} <i>[[++emailsender]]</i> ';
echo $modx->getChunk($chunk));

$chunkName = '@FILE chunks/example.tpl';
echo $modx->getChunk($chunkName, array('param1' => 'value1 :)', 'param100' => 'cache success!!!'));
```
_snippets_:
```
$snippet = '@FILE snippets/example.php';
// echo $modx->runSnippet($snippet, array('snParam' => 'testValue and Cache!!'));
```

#### HTML
_chunks_:
```
[[$@INLINE {{++site_name}} <i>[[++emailsender]]</i>]]
[[$@FILE chunks/example.tpl ? &param1=`value 1 :)` &param100=`cache success!!!`]]
```
_snippets_:
```
[[@FILE snippets/example.php]]
```

### Related issue(s)/PR(s)
[#14201](https://github.com/modxcms/revolution/issues/14201)

### Thanks to @bezumkin 
Thanks to @bezumkin and his pdoTools. A part of logic for creating dynamic elements taken from `pdoTools::_loadElement()`.


Mark-H:
I'll need to look at this more closely later, but it looks like this would allow syntax like `[[!@CODE: runArbitraryPHP(); ]]`. 

In 3.x, the `@EVAL` binding was removed because it could be used in TVs to run arbitrary code. This proposal seemingly would allow any access to anything editable in MODX to result in arbitrary code execution, making it a potential very disastrous vulnerability.

mvoevodskiy:
> I'll need to look at this more closely later, but it looks like this would allow syntax like `[[!@CODE: runArbitraryPHP(); ]]`.
> 
> In 3.x, the `@EVAL` binding was removed because it could be used in TVs to run arbitrary code. This proposal seemingly would allow any access to anything editable in MODX to result in arbitrary code execution, making it a potential very disastrous vulnerability.

Disabled this. `@CODE` and `@INLINE` only for chunks and templates.

mvoevodskiy:
Updated description.

JoshuaLuckers (changes requested):
There is a typo (*temlplates*) in this path: `core/elements/temlplates/.gitignore`

mvoevodskiy:
> There is a typo (_temlplates_) in this path: `core/elements/temlplates/.gitignore`

Done.

sdrenth:
@mvoevodskiy Thank you for putting in the great work! I think your PR should be rebased because of the big PR that has been merged earlier where all classes where refactored which is causing merge conflicts. 

Would be great to see this PR merged 👍

sdrenth:
@mvoevodskiy Any thoughts? Would be great to see this PR merged :)

Ibochkarev:
@opengeek @Mark-H If Michael does a rebase of this PR, will he be accepted or in doubt?

Mark-H:
Well, I can guarantee that it wont get accepted it if it is not rebased ;) 

I personally am not 100% sold, but not 100% against either. Probably leaning slightly in favour, but not enough yet to smash that green button. This is one of those things that I don't use often myself, so more user feedback and testers indicating it works the way they'd want it to work would need to bump my confidence in this PR. 

IMO, this is the type of thing a MAB decision would be useful on to determine consensus, just to make sure it's well considered and people have had a chance to poke holes (so they can be plugged) before it goes to implementation. 

Code-wise it looks okay. `core/elements/` would need to be added into `_build/build.xml` for it to be included in the transport, some loose type checks could be turned into a strict type check, and we can use short array syntax in 3.x, but I see no immediate blocking issues from a code point of view. Tests would be good. 

Seems closely related to the automatic static elements introduced in 2.7 which had the same problem of neither Jason nor myself having much plans of using it. :P

sdrenth:
@Mark-H What's not to like about this pull request? We talked with all website makers and this is what everybody needs in their workflow. I think it would be a great improvement enabling MODX developers to work file based.

Mark-H:
You've talked with _all the site builders_, and _everybody_ wants this in their workflow, really? I know Sterc is all about this sort of things in your workflows, but please don't extrapolate that to the entire community. The fact that I personally don't expect to use this makes your bold claim false ;)

I'm **not** saying that I don't like this PR. All PRs are awesome because they're an opportunity to make MODX better. This one, too. 

When it comes to new features, I consider it my responsibility as integrator to think ahead and try to determine if it's something that fits MODX strategically. Is a proposal something that will still benefit  MODX in 5 years or will we regret having added it (and committing to maintaining it) in the same way everyone hates on ExtJS? What may be unintended side effects, or potential maintenance issues in the future? Is it really something that should be in core, or is it more of a responsibility for extras? Is it the right way to solve a certain problem/use case or can we do better?

I suggested it would be a good idea to get wider community consensus on (e.g. MAB) to make a fair decision on wether it's something we want to do and commit to or not. I'll happily put such a strategic decision over my personal preferences. For many PRs I'm comfortable making a decision myself, but for this one I'm unsure as it's just not a part of my typical workflow, unlike other people. 

Just to state this (again), that does **not** mean I'm against this PR, it means I'm **undecided**. That's a very big difference.

Often if I'm undecided (and low on the SSOGAF), I just let a PR sit until others discuss it and reach a consensus, but in this case @Ibochkarev specifically asked for my thoughts, and the PR has been open for a while without much discussion about the feature so I figured I'd at least try to address the question for as much as I have made up my mind about it.

rthrash:

https://community.modx.com/t/allowing-file-and-inline-bindings-for-chunks-and-snippets-14490/1496/1

Mark-H:
I've opened a topic in the MAB board: https://community.modx.com/t/allowing-file-and-inline-bindings-for-chunks-and-snippets-14490/1496.. and apparently @rthrash is a bot now that automatically cross-posts, that's cool :P

BobRay:
NewsPublisher uses {} tags in the Resource content field to allow it to work with allow_tags_in_post set to false. Standard tags are converted to ]] on load and {{ tags are converted to ]] on save. I don't think the PR creates a problem for NP, but I'm not sure.

JoshuaLuckers:
To make a well-argued decision I would like to gather some more pro's and con's.

> ### Why is it needed?
> This ability do simplier site development because you can not create chunks and snippets in database in all cases.

If you can not create chunks and snippets why would it be good to provide method(s) to bypass this? 

> We talked with all website makers and this is what everybody needs in their workflow. I think it would be a great improvement enabling MODX developers to work file based.

How does this file based workflow look like? What are other possible solutions for this problem?

In recent projects it would have made my life as a developer easier if I could have used `@INLINE for chunks`. 

I'm not in favor of the `@FILE binding` since I'm not totally convinced we will not shoot ourself in the foot (security wise and user/developer friendliness).

gpsietzema:
I thought about this and try to add some of our thoughts to this.

> To make a well-argued decision I would like to gather some more pro's and con's.
> 
> > ### Why is it needed?
> > This ability do simplier site development because you can not create chunks and snippets in database in all cases.
> 
> If you can not create chunks and snippets why would it be good to provide method(s) to bypass this?

The reason to bypass it is from a performance-standpoint. So at the start we used Static Elements, so we were finally able to work in our favorite IDE (phpstorm). Then we came to the conclusion that pdoTools/Fenom had a way of using static elements by bypassing the database, which made it perform better. We did some serious benchmarking on this for large websites and pdoTools/Fenom was way faster. So it's not like we're killing chunks/templates/snippets, it's just that they are not in the elements-tree in MODX in our projects, but in an elements folder in the MODX-filesystem.

> 
> > We talked with all website makers and this is what everybody needs in their workflow. I think it would be a great improvement enabling MODX developers to work file based.
> 
> How does this file based workflow look like? What are other possible solutions for this problem?
> 

Exactly the same as you would do with the current Static Elements in MODX, but just without the database entries which reference to your actual static files.


> In recent projects it would have made my life as a developer easier if I could have used `@INLINE for chunks`.
> 
> I'm not in favor of the `@FILE binding` since I'm not totally convinced we will not shoot ourself in the foot (security wise and user/developer friendliness).

It's a tough call. I'm not entirely sure about where to draw the line. If we protect our users from using this, because we're afraid they will get hacked... I dunno. Can we allow people to make custom snippets? Because they could evaluate user-generated-input there as well. We could disable it by default and have big ass warning message popup to never include user generated content in there. Or we can have the @FILE to only include text and not other MODX placeholders or code...

JoshuaLuckers:
Thanks @gpsietzema for sharing hour thoughts! 
I would like to see this PR rebased and to have some tests include inside `modParserTest`.

Mark-H:
This is now past the feature/breaking change cut-off for alpha1 so pushing further. Shame there's not been discussion from the other MAB members about it.

CrazyBoy49z:
This is great PR for working with Git. This is in the kernel of pdotools that many use for phpstorm and Git

Ibochkarev:
Guys, are there any progress on this PR? Can we add this improvement?

Ibochkarev:
## Summary

Adds **@CODE**, **@INLINE**, and **@FILE** bindings for chunks and snippets so elements can be defined inline or loaded from files without database entries. Available in templates, plugins, resource content, and other parseable places. **@CODE** and **@INLINE** are restricted to chunks and templates only (per Mark-H security feedback). Also fixes static element cache invalidation and a PHP notice in `modMediaSource`.

**Main changes:**
- `modParser::getElement()` — parses `@BINDING` prefix, creates temporary elements for inline/file content before falling back to sourceCache/DB.
- `modElement` — corrects `getSourceFile`/`getMetaData` path handling when filename includes source base path.
- `modScript` — fixes static file mtime check so cache is invalidated when source file is newer.
- `modMediaSource` — fixes "Only variables should be passed by reference" notice for `array_pop(explode(...))`.
- New directories: `core/elements/chunks/`, `snippets/`, `templates/` with `.gitignore`.

## Suggestions

### 1. Rebase required — 2.x vs 3.x structure

The PR targets `core/model/modx/` (2.x layout). The 3.x codebase uses `core/src/Revolution/`. All changes must be applied to:

- `core/src/Revolution/modParser.php` (not `core/model/modx/modparser.class.php`)
- `core/src/Revolution/modElement.php` (not `core/model/modx/modelement.class.php`)
- `core/src/Revolution/modScript.php` (not `core/model/modx/modscript.class.php`)
- `core/src/Revolution/Transport/modMediaSource.php` (not `core/model/modx/sources/modmediasource.class.php`)

`modx->loadClass($class)` is already used in 3.x `getElement()` before the sourceCache lookup; the binding logic should be inserted **before** that lookup, using the raw `$name` before `realname()`.

### 2. Add `core/elements/` to transport

As noted by @Mark-H, `core/elements/` must be included in `_build/build.xml` so the chunks/snippets/templates directories are packaged in the distribution. Otherwise the directories will not exist in fresh installs.

### 3. @FILE path security

`@FILE` accepts a path from `static_elements_basepath` or site root. If the path can be influenced by user input (e.g. from a TV or resource field), path traversal (`../`) could expose files outside the intended directory.

**Recommendation:** Validate and sanitize the path before use — e.g. resolve to realpath, ensure it stays under the base path, and reject `..` segments. Consider restricting `@FILE` to trusted contexts (templates, static chunks) if user content is parsed.

### 4. NewsPublisher `{{ }}` compatibility

@BobRay noted that NewsPublisher uses `{{ }}` in resource content (converted to `[[ ]]` on load/save). The PR converts `{{`/`}}` to `[[`/`]]` only for **modChunk** and **modTemplate** inline content. Ensure this does not conflict with NewsPublisher’s own handling when both are used together. A short note in the docs or a test would help.

### 5. modParser::getElement — `$sourceId` undefined when using default source

In the @FILE branch, when no `static_elements_default_mediasource` is set:

```php
} else {
    $source = modMediaSource::getDefaultSource($this->modx);
    $source->initialize();
    $path = file_exists($filename) ? $filename : '';
    // ...
}
```

`$sourceId` is not set, but later:

```php
'source' => $sourceId,
```

This can pass an undefined variable. **Recommendation:** Set `$sourceId = 0` or the default source ID when using the filesystem fallback.

### 6. modElement — `getMetaData` vs path handling

The change replaces `!$source->getMetaData($filename)` with logic that strips `$sourceBasePath` from `$filename` when building `$checkFilename`. Ensure `getMetaData` in 3.x expects a path relative to the source base (or absolute) and that the new logic matches that contract. Verify with a media source that uses a non-root base path.

### 7. Tests

@JoshuaLuckers requested tests in `modParserTest`. Add coverage for:

- `@INLINE` chunks: `[[$@INLINE {{++site_name}}]]` and `[[$@INLINE [[++site_name]]]]`
- `@FILE` chunks: `[[$@FILE chunks/example.tpl]]` with a fixture file
- `@CODE` restricted: `[[!@CODE: phpinfo();]]` as snippet should **not** execute (element not found)
- `realname()` with `@CODE`/`@INLINE`/`@FILE` prefixes
- Static element cache invalidation when source file is newer than cached file

---

### 8. Strict types and short array syntax

@Mark-H suggested using strict type checks where appropriate and short array syntax `[]` instead of `array()` for 3.x consistency.

## Assessment

### Strengths

- Enables file-based workflows (IDE, Git) without database entries, similar to pdoTools.
- Security: @CODE and @INLINE limited to chunks and templates, avoiding arbitrary PHP in TVs/snippets.
- Fixes static element cache invalidation (source file mtime check).
- Fixes PHP notice in `modMediaSource` for `array_pop(explode())`.
- Clear system settings: `static_elements_basepath`, `static_elements_default_mediasource`.
- `{{ }}` support in inline chunks allows MODX tags inside chunks that are embedded in other chunks/templates.

### Concerns

- PR is based on 2.x structure; rebase to 3.x is required.
- @FILE path handling needs validation to prevent traversal.
- `$sourceId` may be undefined in the default-source branch.
- No automated tests yet.
- Consensus was not reached (MAB discussion); PR is blocked.

## Verdict

**Approve with changes** — after rebase to 3.x and addressing:

1. Rebase onto 3.x and update all file paths.
2. Add `core/elements/` to `_build/build.xml`.
3. Fix `$sourceId` when using default media source.
4. Add path validation for @FILE to prevent traversal.
5. Add tests in `modParserTest` for bindings and cache invalidation.
6. Use short array syntax and strict types where appropriate.

The feature aligns with file-based development workflows and pdoTools usage. The security restrictions on @CODE/@INLINE are appropriate. Once rebased and the above points are addressed, the PR would be in good shape for merge.

Issue #14244, opened 2019-01-08
Update the defaults in htaccess, perhaps deprecate some things, cleanup


guyinpv:
### Summary
Parts of htaccess may not be necessary any more.

Let's look at this section:

```
# Rewrite www.domain.com -> domain.com -- used with SEO Strict URLs plugin
#RewriteCond %{HTTP_HOST} .
#RewriteCond %{HTTP_HOST} ^www.(.*)$ [NC]
#RewriteRule ^(.*)$ https://%1/$1 [R=301,L]
```

1) In the above, why does it have to mention using a plugin? Redirecting to non-www is a common thing to do and has nothing to do with a plugin. This makes it sound like using this redirect somehow requires the plugin to work. This could be reworded, and not mention plugins at all.

2) The style of redirect is not normal looking. Here is a standard way this is done these days:

```
RewriteCond %{HTTP_HOST} ^www\.example\.com [NC]
RewriteRule ^(.*)$ http://example.com/$1 [L,R=301]
```
I'm sure there is a reason for the style of redirect used, but if it can be done in 2 lines instead of 3? It's worth looking at.

Now the next section:

```
# or for the opposite domain.com -> www.domain.com use the following
# DO NOT USE BOTH
#
#RewriteCond %{HTTP_HOST} !^$
#RewriteCond %{HTTP_HOST} !^www\. [NC]
#RewriteCond %{HTTP_HOST} (.+)$
#RewriteRule ^(.*)$ https://www.%1/$1 [R=301,L] .
```

3) Redirecting from non-www to www can also be done in two lines:

```
RewriteCond %{HTTP_HOST} ^example.com [NC]
RewriteRule ^(.*)$ http://www.example.com/$1 [L,R=301,NC]
```
I understand there could be good reasons for the way it is done, I just don't think it's clear. Maybe add some further explanatory text so people don't change it for the worse?

4) At the very end of the code is a lone period just sitting there after `[R=301,L]`, I'm assuming this is not supposed to be there at all? Is it a typo or bug? I imagine it could cause issues?

5) There is a big commented section about htc files for support on Windows XP or something. This is way outdated now, all that stuff can probably just go away.

6) There is a section about register_globals. All this can be removed or deprecated because register_globals has already been removed from PHP as of 5.4.0 anyway.

[...]

PR #14491, opened 2019-03-16
Offer a new version of the file htaccess

Ibochkarev:
### What does it do?
**Added sections:**
- List of Media types
- Reducing MIME type security risks
- Server software information

**Updated sections:**
- Suppressing the `www.` at the beginning of URLs
- Forcing the `www.` at the beginning of URLs
- Compression
- Cache expiration
- Character encodings

**Removed sections:**
- `AddType text/x-component .htc` - There is a big commented section about htc files for support on Windows XP or something. This is way outdated now, all that stuff can probably just go away.
- `php_flag register_globals Off` - All this can be removed or deprecated because register_globals has already been removed from PHP as of 5.4.0 anyway.
- `php_flag zlib.output_compression On` and `php_value zlib.output_compression_level 5` - Even the suggestion regarding Zlib compression is iffy. The php_flag and php_value directives only work if PHP is run as an Apache module. I don't think almost anybody does this do they? It won't work when in CGI mode, so uncommenting these just go to server error most of the time.
Either remove it or add notes that this will break the server if it doesn't work.

### Why is it needed?

- Update the .htaccess file included in the distribution by default
- Add or update the description to some sections

### Related issue(s)/PR(s)
Issue: #14244

### DIFF
--- a/ht.access
+++ b/ht.access
@@ -12,97 +12,348 @@ RewriteEngine On
 RewriteBase /
 
 
+# Media types
+# Serve resources with the proper media types (f.k.a. MIME types)
+
+<IfModule mod_mime.c>
+
+  # Data interchange
+  AddType application/atom+xml                        atom
+  AddType application/json                            json map topojson
+  AddType application/ld+json                         jsonld
+  AddType application/rss+xml                         rss
+  AddType application/vnd.geo+json                    geojson
+  AddType application/xml                             rdf xml
+
+  # JavaScript
+  AddType text/javascript                             js mjs
+
+  # Manifest files
+  AddType application/manifest+json                   webmanifest
+  AddType application/x-web-app-manifest+json         webapp
+  AddType text/cache-manifest                         appcache
+
+  # Media files
+  AddType audio/mp4                                   f4a f4b m4a
+  AddType audio/ogg                                   oga ogg opus
+  AddType image/bmp                                   bmp
+  AddType image/svg+xml                               svg svgz
+  AddType image/webp                                  webp
+  AddType video/mp4                                   f4v f4p m4v mp4
+  AddType video/ogg                                   ogv
+  AddType video/webm                                  webm
+  AddType video/x-flv                                 flv
+  AddType image/x-icon                                cur ico
+
+  # WebAssembly
+  AddType application/wasm                            wasm
+
+  # Web fonts
+  AddType font/woff                                   woff
+  AddType font/woff2                                  woff2
+  AddType application/vnd.ms-fontobject               eot
+  AddType font/ttf                                    ttf
+  AddType font/collection                             ttc
+  AddType font/otf                                    otf
+
+  # Other
+  AddType application/octet-stream                    safariextz
+  AddType application/x-bb-appworld                   bbaw
+  AddType application/x-chrome-extension              crx
+  AddType application/x-opera-extension               oex
+  AddType application/x-xpinstall                     xpi
+  AddType text/calendar                               ics
+  AddType text/markdown                               markdown md
+  AddType text/vcard                                  vcard vcf
+  AddType text/vnd.rim.location.xloc                  xloc
+  AddType text/vtt                                    vtt
+  AddType text/x-component                            htc
+
+</IfModule>
+
+
+# Character encodings
+
+# Serve all resources labeled as `text/html` or `text/plain`
+# with the media type `charset` parameter set to `UTF-8`.
+
+AddDefaultCharset utf-8
+
+
+# Serve the following file types with the media type `charset`
+# parameter set to `UTF-8`.
+
+<IfModule mod_mime.c>
+    AddCharset utf-8 .atom \
+                     .bbaw \
+                     .css \
+                     .geojson \
+                     .ics \
+                     .js \
+                     .json \
+                     .jsonld \
+                     .manifest \
+                     .markdown \
+                     .md \
+                     .mjs \
+                     .rdf \
+                     .rss \
+                     .topojson \
+                     .vtt \
+                     .webapp \
+                     .webmanifest \
+                     .xloc \
+                     .xml
+</IfModule>
+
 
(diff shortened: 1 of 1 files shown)


Ibochkarev:
This is not the final option, ready to listen to thoughts about this proposal. The old file is outdated and needs to be updated.

ht.access
```diff
@@ -12,97 +12,348 @@ RewriteEngine On
...
-#RewriteCond %{HTTP_HOST} (.+)$
-#RewriteRule ^(.*)$ https://www.%1/$1 [R=301,L] .
+# (!) NEVER USE BOTH WWW-RELATED RULES AT THE SAME TIME!
+
+# (1) The rule assumes by default that both HTTP and HTTPS
+#     environments are available for redirection.
+#     If your SSL certificate could not handle one of the domains
+#     used during redirection, you should turn the condition on.
 
+# RewriteCond %{HTTPS} !=on
```
Jako, line 147: Why is this restriction used? If it is necessary: Why not the opposite, since https should be the default?

ht.access
```diff
@@ -12,97 +12,348 @@ RewriteEngine On
...
+# RewriteCond %{HTTP_HOST} !^www\. [NC]
+# RewriteCond %{SERVER_ADDR} !=127.0.0.1
+# RewriteCond %{SERVER_ADDR} !=::1
+# RewriteRule ^ %{ENV:PROTO}://www.%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 
 
 # Rewrite secure requests properly to prevent SSL cert warnings, e.g. prevent 
 # https://www.example.com when your cert only allows https://secure.example.com
+
```
Jako, line 156: Use the following condition here and the rewrite rule could contain a variable hostname with %1:
`#RewriteCond %{HTTP_HOST} (.+)$`

Jako (changes requested):

JoshuaLuckers (changes requested):
The additions seem to come from a [MIT licensed file](https://github.com/h5bp/html5-boilerplate/blob/master/dist/.htaccess).

> Using `.htaccess` files slows down Apache

Do we really want to add all these new rules/configurations?

Furthermore the RewriteRules should come as early as possible since they are the most likely to be changed and putting them down in the file isn't user friendly.

Ibochkarev:
> Do we really want to add all these new rules/configurations?

@JoshuaLuckers The associated issue received a proper response from colleagues from the community. Therefore, I offered my own vision of this file in Draft mode.
Therefore, I am ready to make changes to this PR.

Rules for RewriteRules specified @Jako  I will correct.

Waiting for suggestions from the community and integrators for this file. It must be clearly refresh!

guyinpv:
To be honest here, I don't think the htaccess file needs almost anything except the general index file rewrite.

The reason is simple, unless MODX **requires** something to work, ANY other optimizations, cache settings, headers, compression, HTTPS or www testing should be configured by the end user appropriate for their project.

Further, creating cache settings tends to be done by plugins who are built to do such things anyway.

Maybe an exception can be made for security settings though, like protecting MODX system folders/files with a `<FilesMatch>` or `<Files>` section. Maybe turn off directory indexing for example, block access to htaccess itself, dot folders and files, etc.

We have two options; give people every kind of interesting setting, or just the basics.
ProcessWire has a host of settings: https://github.com/ryancramerdesign/ProcessWire/blob/master/htaccess.txt
While WordPress has the basics: https://codex.wordpress.org/htaccess

I also think doing a ton of cool stuff only for Apache users really leaves a hole for Nginx users. We should provide a matching config for Nginx, or at least instructions how to setup similar options in it.

ghost:
I would prefer a basic ht.access file, with a gist or maybe text file somewhere to have various possible settings in a format easy to copy/paste as required.

The same file could also have a section of useful nginx configurations.

Ibochkarev:
In the end, what do we do with this PR?

Mark-H:
It's a good idea to refresh the provided ht.access, but let's simplify it to only what MODX need and perhaps the minimum security things (locking core, dot-files/directories); put examples of more advanced configurations into the docs instead.

ht.access
guyinpv: Note that for auto-ssl systems, they need to read the .well-known folder without HTTPS. You'll note in most htaccess files they make an exception for .well-known (and others) before redirecting to HTTPS.

ht.access
```diff
@@ -12,97 +12,348 @@ RewriteEngine On
...
+# you should choose one of the alternatives and redirect the other
+# one.
 #
-# or for the opposite example.com -> www.example.com use the following
-# DO NOT USE BOTH
+# (!) NEVER USE BOTH WWW-RELATED RULES AT THE SAME TIME!
+
+# RewriteCond %{HTTPS} !=on
+# RewriteCond %{HTTP_HOST} ^www\.(.+)$ [NC]
+# RewriteRule ^ %{ENV:PROTO}://%1%{REQUEST_URI} [R=301,L]
```
guyinpv, line 128: Can you use ENV:PROTO without first setting it manually? I've seen this used first before you can access that:
```
RewriteCond %{HTTPS} =on
RewriteRule ^(.*)$ - [env=proto:https]
RewriteCond %{HTTPS} !=on
RewriteRule ^(.*)$ - [env=proto:http]
```

JoshuaLuckers:
> It's a good idea to refresh the provided ht.access, but let's simplify it to only what MODX need and perhaps the minimum security things (locking core, dot-files/directories); put examples of more advanced configurations into the docs instead.

Good idea to put the example in the docs. We still have to consider the suggestions mentioned in this draft PR before it can be used as example in the docs.

Ibochkarev:
I will delete all the new studies that are indicated here and leave only those that were considered in the issue https://github.com/modxcms/revolution/issues/14244.

Everything else will write in the new documentation.

Ibochkarev:
Leave it as it is.

Issue #14489, opened 2019-03-15
Error in core\model\phpthumb\phpthumb.class.php 990


ghoulru:
### Summary
Notice: Use of undefined constant IMG_WEBP - assumed 'IMG_WEBP' in \core\model\phpthumb\phpthumb.class.php on line 990

### Step to reproduce
Rename "core" and "manager" directories/
Create TV named "image". Media sources: default and custom, same problem.
don't show images thumbnails in /manager/

### Observed behavior
If i comment or delete code from line 990
```
if ($imagetypes & IMG_WEBP) {
$this->thumbnailFormat         = 'webp';
				$AvailableImageOutputFormats[] = 'webp';
			}
```
everythig is ok.

screenshots
[error in code](http://перекраска-дома.рф/phpthumb_modx.png)
[no thumbnail](http://перекраска-дома.рф/modx_manager.png)

### Environment
MODX 2.7.1, php 7.0.8 , mysql 5.5, Fitefox 65.0.2. Using openserver 5.2.2

JoshuaLuckers:
The constant `IMG_WEBP` is available as of PHP 5.6.25 and PHP 7.0.10, respectively. 
Source: [PHP Predefined Constants](https://secure.php.net/manual/en/image.constants.php)

```
/**
 * Used as a return value by imagetypes() Available as of PHP 5.6.25 and PHP 7.0.10, respectively.
 * @link https://php.net/manual/en/image.constants.php
 */
define('IMG_WEBP', 32);
```

Ibochkarev:
Fixed #14494

@modxbot close

PR #14494, opened 2019-03-17
Fix for undefined constant IMG_WEBP

JoshuaLuckers:
### What does it do?
Define the IMG_WEBP constant.

### Why is it needed?
The IMG_WEBP constant is used as a return value by imagetypes() and available as of PHP 5.6.25 and PHP 7.0.10, respectively. This would cause some notices when using an not up-to-date install of, for example, PHP 7.

### Related issue(s)/PR(s)
Fixes issue #14489

### DIFF
--- a/core/model/phpthumb/modphpthumb.class.php
+++ b/core/model/phpthumb/modphpthumb.class.php
@@ -33,6 +33,13 @@ class modPhpThumb extends phpThumb
         $this->modx =& $modx;
         $this->config = $config;
 
+        if (version_compare(PHP_VERSION, '5.6.25', '<')
+            || (version_compare(PHP_VERSION, '7.0.0', '>=')
+                && version_compare(PHP_VERSION, '7.0.10', '<'))) {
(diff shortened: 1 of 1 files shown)


OptimusCrime:
What about

```
if (!defined('IMG_WEBP')) {
    define('IMG_WEBP', 32);
}
```

?

JoshuaLuckers:
> What about
> 
> ```
> if (!defined('IMG_WEBP')) {
>     define('IMG_WEBP', 32);
> }
> ```
> ?

I did it this way because I believe it should only be defined in these specific cases. In the future it's also easier to spot once we increase the minimum required PHP version beyond those specific versions.

Ibochkarev:
This way we fix it in the second branch, but in the 3rd branch, everything is put through a composer. Would be nice to send a fix to the repository phpThumb?

Mark-H (approved):

alroniks (approved):

Issue #14504, opened 2019-03-21
New design for installer


Ibochkarev:
### Summary
It is rewritten the pages of the installer in accordance with the designed design.

### Why is it needed?
![1-welcome](https://user-images.githubusercontent.com/2138260/54737119-9d902700-4bd8-11e9-9faf-7115aebaa2eb.png)
![2-settings](https://user-images.githubusercontent.com/2138260/54737120-9d902700-4bd8-11e9-924e-e171738d1486.png)
![3-connect](https://user-images.githubusercontent.com/2138260/54737121-9e28bd80-4bd8-11e9-90ba-6ca508073fca.png)
![4-test](https://user-images.githubusercontent.com/2138260/54737123-9e28bd80-4bd8-11e9-9559-75cc199032f1.png)
![5-installed](https://user-images.githubusercontent.com/2138260/54737126-9e28bd80-4bd8-11e9-817d-f14830b997e6.png)

### Suggested solution(s)
- [x] Add rules to gruntfile.js for compiling styles for the installer
- [x] Connect normalize to reset styles
- [x] Delete unused images
- [x] Delete unnecessary style files after assembly via Grunt
- [x] Rewrite layout using flexbox

### Related issue(s)/PR(s)
N/A

This process I'll do the other day. Please assigned me for this request.

Also, please comment on what else needs to be improved. This will be the first step in this improvement.

Ruslan-Aleev:
Related with https://github.com/modxcms/revolution/issues/14238, https://github.com/modxcms/revolution/issues/13876

Ibochkarev:
can be closed. made in #14507

PR #14507, opened 2019-03-24
New design for installer

Ibochkarev:
### What does it do?
Describe the technical changes you did.

- [x] Add rules to gruntfile.js for compiling styles for the installer
- [x] Connect normalize to reset styles
- [x] Delete unused images
- [x] Delete unnecessary style files after assembly via Grunt
- [x] Rewrite layout using flexbox
- [x] Change the order of fields to fill

### Why is it needed?
It is rewritten the pages of the installer in accordance with the designed design.

### Related issue(s)/PR(s)
#14504 #14238


OptimusCrime:
Want to share some print screens of how this looks?

jonleverrier:
I think the designs are from the related issue @OptimusCrime https://github.com/modxcms/revolution/issues/14504

Ibochkarev:
@jonleverrier Yes. That's right.

JoshuaLuckers (commented):
Please don't commit the compiled CSS. The integrator who merges the PR will compile the final CSS.

Ibochkarev:
> Please don't commit the compiled CSS. The integrator who merges the PR will compile the final CSS.

@JoshuaLuckers  ok! I fixed this

Ibochkarev:
Please test.

header.tpl
```diff
@@ -0,0 +1,51 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <title>{$app_name} {$app_version} &raquo; {$_lang.install}</title>
+
+    <meta charset="utf-8">
+
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+
```
JoshuaLuckers, line 9:
```suggestion
    <base href="/setup/">
```

Without `base href` assets won't load if you run setup without trailing slash (https://example.com/setup instead of https://example.com/setup/)

_build/templates/default/sass/installer.scss
```diff
@@ -0,0 +1,762 @@
...
+    width: 100%;
+    padding-right: 15px;
+    padding-left: 15px;
+    margin-right: auto;
+    margin-left: auto;
+    max-width: 1140px;
+}
+
+header {
+    margin-top: 50px;
```
JoshuaLuckers: 50px is a lot of margin, especially when your screen doesn't have much view height. 
I'm in favour of removing the `margin-top`.

_build/templates/default/sass/installer.scss
```diff
@@ -0,0 +1,762 @@
...
+    .content-inside {
+        padding: 20px;
+
+        .wrapper {
+            background: $white;
+            padding: 30px;
+            border-radius: 5px;
+            max-width: 890px;
+            height: auto;
+            min-height: 540px;
```
JoshuaLuckers: This is also pretty high, especially when your screen doesn't have much view height.

setup/templates/options.tpl
```diff
@@ -1,128 +1,144 @@
...
-    <input type="button" onclick="MODx.go('welcome');" value="{$_lang.back}" />
-</div>
+
+    <div class="content-wrap">
+        <h2>{$_lang.options_title}</h2>
+
+        <div class="options">
+            <div class="options-wrap">
+
+                <div class="option-item">
```
JoshuaLuckers: I would expect the whole element to be a trigger for the option. Using a label element would be the best approach (keep in mind that [only inline elements are allowed inside a label](https://stackoverflow.com/a/4461962))

setup/templates/contexts.tpl
```diff
@@ -11,57 +11,101 @@ Ext.onReady(function() {literal}{{/literal}
...
+    </div>
 </div>
+
 <br />
 
 <div class="setup_navbar">
-<input type="submit" id="modx-next" name="proceed" value="{$_lang.next}" />
-<button type="button" id="modx-back" onclick="MODx.go('database');">{$_lang.back}</button>
+    <input type="submit" id="modx-next" class="button" name="proceed" value="{$_lang.next} &#xf054;" />
+    <button type="button" id="modx-back" class="button" onclick="MODx.go('database');">&#xf053; {$_lang.back}</button>
```
JoshuaLuckers: The icon doesn't work (in the other templates an `input` element is used instead of a `button`)

JoshuaLuckers (commented):
Thanks for the amazing work! I made some comments on specific parts of the changes.

I've also attached a screenshot as reference for the `height` comments.
modx3-new-setup-height

_build/templates/default/sass/installer.scss
Ibochkarev: Agree

setup/templates/contexts.tpl
Ibochkarev: Fixed

_build/templates/default/sass/installer.scss
Ibochkarev: Fixed

setup/templates/options.tpl
Ibochkarev: Fixed

opengeek:
Please, stop merging 3.x back into your PR branches. Can you remove that? If you need to update the PR, rebase it on top of the latest 3.x instead.

setup/templates/options.tpl
JoshuaLuckers: It would be nice to have the same behaviour for the "Advanced Options" checkboxes.

alroniks:
It seems pretty ready for testing and merging after. Please, have a look.

alroniks (dismissed):
Despite the fact that I also made edits, for me, everything looks good. Very obvious errors are corrected, the remaining trifles should be corrected after the merge.

JoshuaLuckers:
I'll test it this weekend.

JoshuaLuckers (dismissed):

alroniks:
I rebased the branch to get it to work properly. Please, have a look @JoshuaLuckers @Mark-H @opengeek

alroniks:
@opengeek I also fixed rule for checking upgradable versions. It should be inverted as here https://github.com/modxcms/revolution/pull/14507/commits/1fc4a952c3b739c33cf27ea9474d32a3fcc5e925

JoshuaLuckers (approved):
The height of the "content" still isn't fixed and going to the next step should go back to the top. Furthermore the installer seems to work fine.

PR #14518, opened 2019-04-01
Fix padding in _windows.scss

Ruslan-Aleev:
### What does it do?
Removed `.x-window.modx-window .x-window-body {padding-top: 0;}`

### Why is it needed?
Missing indent in package update window.

Before
before

After
after

I do not understand why the style is:
`.x-window.modx-window .x-window-body {padding-top: 0;}`

**I removed this styles and did not notice the changes, perhaps the community knows what these lines are for.**

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/_windows.scss
+++ b/_build/templates/default/sass/_windows.scss
@@ -103,16 +103,6 @@
     overflow-y: auto;
     padding: 15px;
   }
-  &.modx-window {
-    .x-window-body {
-      padding-top: 0;
-    }
-    &.modx-console, &.modx-alert, &.modx-confirm {
-      .x-window-body {
-        padding-top: 15px;
-      }
-    }
-  }
 
   .x-panel-bwrap {
     background: $winBodyBg;


Jako:
It was introduced in #13038. Please check the side effects of the removement. Maybe there are better ways to solve the double padding.

Ruslan-Aleev:
Yes, if remove {padding-top: 0;} then a double padding appears (padding x-window-body + padding from the label).
However, the problem here is wider, since the double padding is still present in different forms, i.e. no standard format.

For example, in the form of editing a template or resource.

padding-top-1

padding-top-2-1

padding-top-2-2

In the form of a resource, by the way, a completely different class is responsible for padding - `.main-wrapper`

It’s good to bring everything to the same format, or do it all in MODX3.
Or just add an additional class to the package update window, and then the pop-up windows will be without double padding. _But this is not a very good option, because there is no standard format for padding in other forms._

**Or leave a double padding for all forms (`.x-window-body and .main-wrapper`), i.e. lead to standard format.**

@Jako Any thoughts on this?

Issue #6703, opened 2012-01-16
Turn off autocomplete on Title field


modxbot:
rethrash created Redmine issue ID 6703

The Title field needs to have autocomplete set to off, because it tends to autofill other page titles on larger sites with similar field names.

<pre>
autocomplete="off"
</pre>

OptimusCrime:
This is a like/dislike case. I personally have no problems with the autocomplete.

pixelchutes:
+1 for adding / disabling autocomplete. While I generally like the (browser) feature, it would be nice if it were context aware and did not pull in suggestions from other sites. I _think_ that was @rthrash's issue with it.

rtripault:
I guess #11895 is also related

JoshuaLuckers:
I have a love/hate relationship with autocompletion. I propose to add a user setting to enable/disable autocompletion when managing resources.

Mark-H:
#14526 adds a setting for it... do we honestly need to make things like this configurable? :/ 

We already have so many settings, so I'd like to see we either just disable autocomplete, or leave it as-is. This is not really that big of a deal that it's worth the extra complexity IMO.

smg6511:
We should make some sort of move on this or close it. Personally, the only time I find autocomplete (AC) useful in the context of the manager forms is when doing dev/testing (mainly for modx itself, not client back ends). What about making a single setting used to turn off AC system-wide?

rthrash:
It should just be off, period. If you need dev testing, there are scripting tools to do that that don't require autocomplete.

smg6511:
@rthrash - I totally agree. Unfortunately, after playing with this today, I found it's virtually impossible to control. There was a trick that worked until recently (setting autocomplete to "new-password"), but it seems Google has further exerted its control over this subject and, let's face it, Chrome being the overwhelmingly-dominant browser makes it the _only_ one to pay attention to when deciding whether to pursue this further.

That said, shall we just close this?

rthrash:
Lots of devs will use different browsers, too. But yeah, Chrome is the dominant one. 

I still think the fields should be explicitly set to `autocomplete="off"`

PR #14526, opened 2019-04-02
Adds ability to turn on/off autocompletion for 'pagetitle' field

DettRoxx:
### What does it do?
Adds ability to turn on/off autocompletion for 'pagetitle' field with system setting `resource_pagetitle_autocomplete`

### Why is it needed?
to avoid long sheet of available autocompletions

### Related issue(s)/PR(s)
#6703

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -625,6 +625,9 @@ $_lang['setting_resolve_hostnames_desc'] = 'Do you want MODX to try to resolve y
 $_lang['setting_resource_tree_node_name'] = 'Resource Tree Node Field';
 $_lang['setting_resource_tree_node_name_desc'] = 'Specify the Resource field to use when rendering the nodes in the Resource Tree. Defaults to pagetitle, although any Resource field can be used, such as menutitle, alias, longtitle, etc.';
 
+$_lang['setting_resource_pagetitle_autocomplete'] = 'Enables autocomplete for pagetitle field';
+$_lang['setting_resource_pagetitle_autocomplete_desc'] = 'Select \'Yes\' if you want to see hints/autocomplete data based on the previous entered values for the pagetitle field of the resource.';
+
 $_lang['setting_resource_tree_node_name_fallback'] = 'Resource Tree Node Fallback Field';
 $_lang['setting_resource_tree_node_name_fallback_desc'] = 'Specify the Resource field to use as fallback when rendering the nodes in the Resource Tree. This will be used if the resource has an empty value for the configured Resource Tree Node Field.';
 
--- a/core/lexicon/ru/setting.inc.php
+++ b/core/lexicon/ru/setting.inc.php
@@ -634,6 +634,9 @@ $_lang['setting_resolve_hostnames_desc'] = 'Хотите ли вы, чтобы M
 $_lang['setting_resource_tree_node_name'] = 'Поле для названия узла в дереве ресурсов';
 $_lang['setting_resource_tree_node_name_desc'] = 'Укажите поле ресурса, которое будет использоваться в качестве названия узла в дереве ресурсов. По умолчанию поле «pagetitle», любое поле ресурса может быть использовано: «menutitle», «alias», «longtitle», и т.п.';
 
+$_lang['setting_resource_pagetitle_autocomplete'] = 'Включить автозаполнение для поля заголовка';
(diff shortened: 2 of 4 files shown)


core/lexicon/ru/setting.inc.php
```diff
@@ -634,6 +634,9 @@
 $_lang['setting_resource_tree_node_name'] = 'Поле для названия узла в дереве ресурсов';
 $_lang['setting_resource_tree_node_name_desc'] = 'Укажите поле ресурса, которое будет использоваться в качестве названия узла в дереве ресурсов. По умолчанию поле «pagetitle», любое поле ресурса может быть использовано: «menutitle», «alias», «longtitle», и т.п.';
 
+$_lang['setting_resource_pagetitle_autocomplete'] = 'Включить автозаполнение для поля заголовка';
```
alroniks, line 637: It should be removed from the pull request, all strings will be translated via [Crowdin](https://crowdin.com/project/modx-revolution).

manager/assets/modext/core/modx.js
```diff
@@ -95,6 +96,14 @@ Ext.extend(MODx,Ext.Component,{
         });
     }
 
+    ,initAutocompletePagetitle: function() {
+        Ext.ComponentMgr.onAvailable("modx-resource-pagetitle", function(field) {
+            var setting_pagetitle_autocomplete = +MODx.config.resource_pagetitle_autocomplete;
+            var autocomplete = setting_pagetitle_autocomplete ? "on" : "off";
+            field.defaultAutoCreate = {tag: "input", type: "text", size: "20", autocomplete: autocomplete, msgTarget: 'under' }
```
alroniks, line 103: Why the size is 20?

core/lexicon/en/setting.inc.php
```diff
@@ -625,6 +625,9 @@
 $_lang['setting_resource_tree_node_name'] = 'Resource Tree Node Field';
 $_lang['setting_resource_tree_node_name_desc'] = 'Specify the Resource field to use when rendering the nodes in the Resource Tree. Defaults to pagetitle, although any Resource field can be used, such as menutitle, alias, longtitle, etc.';
 
+$_lang['setting_resource_pagetitle_autocomplete'] = 'Enables autocomplete for pagetitle field';
+$_lang['setting_resource_pagetitle_autocomplete_desc'] = 'Select \'Yes\' if you want to see hints/autocomplete data based on the previous entered values for the pagetitle field of the resource.';
```
alroniks, line 629: Could you change quotes to `"` for "Yes" or rephrase the string to avoid issues with backslashed symbols?

alroniks (changes requested):
Need to apply some changes. Also, I am wondering, is it should be applied only for pagetitle field? What about longtitle and others?

manager/assets/modext/core/modx.js
DettRoxx: I copied from here (default code of modx.js)
image

and change only logic for autocomplete

core/lexicon/en/setting.inc.php
DettRoxx: i take example as 'standart' from here, here single quotes with backslash 
image

alroniks: I understand, but the existing code does not mean automatically that it is correct. Those string also should be fixed in the future.

manager/assets/modext/core/modx.js
DettRoxx: i think (if we can write more than 20 characters in pagetitle field - 'size' is only default, but overrides by ExtJS for every field with strict 'size' option)

Ruslan-Aleev:
An interesting issue, but narrowly focused :) Maybe we need to make the setting for all fields of the resource, as suggested earlier, or even the fields of the elements?
p.s. Although autocomplete never bothered me.

DettRoxx:
Yes, sounds good. Like 'exclude templates' in ms2Gallery, but for fields. I'll try to change logic to this functionality

JoshuaLuckers:
Having a setting to turn off autocompletion for the page title alone is fine. We don't need to enable/disable autocompletion for the other fields (and if we do create new PR's for them).

If you need any help solving the requested changes please let us know.

Mark-H:
We're now past the cutoff for alpha1, and as this is a new feature it's being pushed to a later release.

Mark-H:
Devil's advocate here, does this need a setting or can we just disable autocomplete for these fields? We're not filling in addresses or other types of "standard" content usually that can benefit from autocompletion are we?

smg6511:
I have a PR that addresses this globally (#16225) that is ready for merging but please take note of [my comment](https://github.com/modxcms/revolution/issues/6703#issuecomment-1182473333) re autocomplete being difficult to impossible to control with Chrome.

That said, should we consider closing this PR?

Issue #14520, opened 2019-04-02
regClientScript async defer


ghoulru:
Add  new option can be useful:
`void regClientScript (string $src, [boolean $plaintext = false][, $async=''])`
where $async = async | defer
and in front looks like: 
<script src="/path/to/script.js" async></script>

For asynchronous script loading.
Thanks.

PR #14527, opened 2019-04-02
Added 'async' optional param to 'regClientScript' and 'regClientStartupScript'

DettRoxx:
### What does it do?
Added optional param 'async' for methods **regClientScript** and **regClientStartupScript** of modx object, which allows to get 

`<script type="text/javascript" src="/assets/js/script.js" async></script>` 
**or** 
`<script type="text/javascript" src="/assets/js/script.js" defer></script>`

**instead of**
``<script type="text/javascript" src="/assets/js/script.js"></script>``

### Why is it needed?
to avoid "blocking-like" synchronous behaviour when browser builds DOM and stops and execute scripts tags, allows to use asynchronous behaviour instead

### Related issue(s)/PR(s)
#14520

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -1548,9 +1548,11 @@ class modX extends xPDO {
      * tag of an HTML response.
      * @param boolean $plaintext Optional param to treat the $src as plaintext
      * rather than assuming it is JavaScript.
+     * @param string $async String literal 'async' or 'defer' which is added 
+     * as attribute to html script tag
      * @return void
      */
-    public function regClientStartupScript($src, $plaintext= false) {
+    public function regClientStartupScript($src, $plaintext= false, $async='') {
         if (!empty ($src) && !array_key_exists($src, $this->loadedjscripts)) {
             if (isset ($this->loadedjscripts[$src]))
                 return;
@@ -1560,7 +1562,7 @@ class modX extends xPDO {
             } elseif (strpos(strtolower($src), "<script") !== false) {
                 $this->sjscripts[count($this->sjscripts)]= $src;
             } else {
-                $this->sjscripts[count($this->sjscripts)]= '<script type="text/javascript" src="' . $src . '"></script>';
+                $this->sjscripts[count($this->sjscripts)]= '<script type="text/javascript" src="' . $src . '" '.htmlentities($async).'></script>';
             }
         }
     }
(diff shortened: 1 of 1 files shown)


DettRoxx:
Need to update UnitTests for this PR

core/model/modx/modx.class.php
```diff
@@ -1548,9 +1548,11 @@ public function regClientCSS($src, $media = null) {
      * tag of an HTML response.
      * @param boolean $plaintext Optional param to treat the $src as plaintext
      * rather than assuming it is JavaScript.
+     * @param string $async String literal 'async' or 'defer' which is added 
+     * as attribute to html script tag
      * @return void
      */
-    public function regClientStartupScript($src, $plaintext= false) {
+    public function regClientStartupScript($src, $plaintext= false, $async='') {
```
alroniks, line 1555: Only two params allowed - `async` and `defer`, so we should add check if passed param is correct or just make it possible to turn it into a boolean parameter.

alroniks (changes requested):

core/model/modx/modx.class.php
DettRoxx: fixed. now params checks correct values

JoshuaLuckers: We should add an attributes parameter. The script tag can also have other attributes like `id`.

Mark-H: There are also other methods available (e.g. `regClientHTML`) so I don't think we need an option for every possibility here.

JoshuaLuckers: I was thinking about something like this:
```
public function regClientStartupScript($src, $plaintext= false, $attributes = [])
```

`$attributes = ['async', 'defer', 'id' => ['test'], 'etc'];`

DettRoxx: but for what? i opened w3 specs for 'script' tag:
we see:

image

JoshuaLuckers: You are referring to W3 schools. Not the w3 specs.
[See this list of supported attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Attributes). 
This also includes [all global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes).

Mark-H:
This just needs the tests to be updated to get my approval.

Mark-H (changes requested):
Failing tests need to be updated.

JoshuaLuckers:
@DettRoxx if you need any help fixing the test please let us know 👍

Mark-H:
Closing due to lack of interaction - if someone wants to pitch in to get this up to speed again that'd be awesome.

Issue #5225, opened 2011-07-11
Allow in-line editing in FC sets grid


gadamiak:
gadamiak created Redmine issue ID 5225

FC profiles grid allows in-line editing of name and description for a profile. FC sets grid doesn't. It would be useful to allow editing of at least description field for a set in FC sets grid. Ideally all fields would be editable in-line: action, description, template, constraint. That would speed up working with FC.

ilyautkin:
Confirmed for MODX 2.7.0

smg6511:
This has been implemented in 3.0, both for the Profiles and Sets grids.

PR #14529, opened 2019-04-03
Allow in-line editing in FC sets grid

DettRoxx:
### What does it do?
Adds ability to edit as inline fields in grid in FC sets as in FC Profiles grid

### Why is it needed?
Increase productivity

### Related issue(s)/PR(s)
#5225

### DIFF
--- a/core/model/modx/processors/security/forms/set/updatefromgrid.class.php
+++ b/core/model/modx/processors/security/forms/set/updatefromgrid.class.php
@@ -1,0 +1,29 @@
+<?php
+/*
+ * This file is part of MODX Revolution.
+ *
+ * Copyright (c) MODX, LLC. All Rights Reserved.
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ */
+
+require_once (dirname(__FILE__) . '/update.class.php');
+/**
+ * Update a FC Profile from grid
+ *
+ * @package modx
+ * @subpackage processors.security.forms.profile
+ */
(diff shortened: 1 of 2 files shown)


core/model/modx/processors/security/forms/set/updatefromgrid.class.php
```diff
@@ -0,0 +1,29 @@
...
+        $data = $this->getProperty('data');
+        if (empty($data)) return $this->modx->lexicon('invalid_data');
+        $properties = $this->modx->fromJSON($data);
+        $this->setProperties($properties);
+        $this->unsetProperty('data');
+
+        return parent::initialize();
+    }
+}
+return 'modFormCustomizationSetUpdateFromGridProcessor';
```
alroniks, line 29: Better to write `return modFormCustomizationSetUpdateFromGridProcessor::class` here.

alroniks:
This RP will be on hold before Jason finish his refactoring.

Mark-H (approved):

JoshuaLuckers:
@DettRoxx thanks for the patience, is it possible for you to refactor this PR?

Mark-H:
The processor needs to be redone in the new way. If someone wants to pitch in here to help that would be awesome.

Mark-H (changes requested):
Processor needs to be redone after the processor refactor.

theboxer:
The processor already exists in the 3.x codebase, to make this work only the changed JS file is needed, with updated processor's path. (Tested and works fine)

opengeek:
This was incorporated via #14781

Issue #14530, opened 2019-04-03
Bug when deleting the parent item in the "Menu" section


Ruslan-Aleev:
### Summary
If you delete the parent item in the menu, the **child item is no longer displayed in the tree**, but it is not deleted.
It will not be possible to create an item with the same name and it is in the list of parents.

![menu-delete](https://user-images.githubusercontent.com/12523676/55484803-3774bc80-563a-11e9-9af4-9efce3a0b280.png)

![menu-delete-2](https://user-images.githubusercontent.com/12523676/55484804-3774bc80-563a-11e9-9296-7135e6f3aa36.png)

When creating a new item, the last edited item is automatically assigned as a parent, even if it has been deleted and when saving item an error pops up that the parent is not found.

![menu-delete-3](https://user-images.githubusercontent.com/12523676/55484949-7e62b200-563a-11e9-8587-c2ba4d15db00.png)

### Environment
MODX 2.7.1
https://github.com/modxcms/revolution/issues/14220
https://github.com/modxcms/revolution/issues/14521

alroniks:
So, I've started work on it and faced with a problem. What to do with nested menus?
Somebody accidentally can remove top menu (even after confirming popup) and in case removing all menus will go away directly from a database, it can be painful to restore them from the database.
The second option - move a menu to the level of deleted.

Please, vote using emoji for this comment. 👍 for the removing, 👎 against removing that also means moving to the upper level.

Jako:
With a warning 'removing' the children should be ok.

Jako:
If a parent is removed, the children should be removed (with a warning). Otherwise it will be a lot of work to delete a whole tree. Moving items up is not a normal behavior for me.

jcdm:
The 'next level' would be to have a marked-as-deleted for menu items from which they can be undeleted, but that would increase the scope a fair bit so I couldn't say if it would have much value in terms of value for users for the investment.

JoshuaLuckers:
Maybe we could ask the users to move the items “up” or to another element?

Mark-H:
Fixed in #14532

PR #14532, opened 2019-04-04
Remove menu item together with all nested menus

alroniks:
### What does it do?
It removes all nested items when the parent menu was deleted.

### Why is it needed?
Before: when parent menu is deleted nested submenus become ghosts. It may confuse users, especially when they try to create a new item but get an error that item already exists. And it is the truth but it is impossible to manage these ghosts by UI. Now the whole branch should be deleted.

### Related issue(s)/PR(s)
#14530


core/lexicon/en/menu.inc.php
```diff
@@ -25,7 +25,7 @@
 $_lang['menu_parent'] = 'Parent Menu';
 $_lang['menu_parent_err_ns'] = 'Parent menu not specified!';
 $_lang['menu_parent_err_nf'] = 'Parent menu not found!';
-$_lang['menu_remove'] = 'Delete Menu Item';
+$_lang['menu_remove'] = 'Delete Menu';
 $_lang['menu_top'] = 'Top Menu';
```
Ruslan-Aleev: Maybe it makes sense to change "Top Menu" to "Main Menu"? Since in the MODX3 menu is no longer on top.

core/lexicon/en/menu.inc.php
```diff
@@ -16,7 +16,7 @@
 $_lang['lexicon_key'] = 'Lexicon Key';
 $_lang['lexicon_key_desc'] = 'The text, or lexicon key, that will be used for rendering the title text of this page in the menu.';
 $_lang['menu_create'] = 'New Menu';
-$_lang['menu_confirm_remove'] = 'Are you sure you want to remove this menu?';
+$_lang['menu_confirm_remove'] = 'All nested menus also will be deleted. Are you sure you want to remove this menu? ';
```
JoshuaLuckers: To keep it consistent with the lexicon [`resource_delete_confirm`](https://github.com/modxcms/revolution/blob/cfa09fabf3914d6012f7c4606cf1e9e601bd95a4/core/lexicon/en/resource.inc.php#L47) I propose to use this lexicon instead:
```suggestion
$_lang['menu_confirm_remove'] = 'Are you sure you want to delete this menu?<br />NOTE: Any nested menus will also be deleted!';
```

Ruslan-Aleev: Then it is better to write "to delete" instead of "to remove", as in the resource lexicon.

core/model/modx/processors/system/menu/remove.class.php
```diff
@@ -16,20 +16,47 @@
...
+    {
+        $criteria = ['parent' => $menu->get('text')];
+
+        if (!$this->modx->getCount($this->classKey, $criteria)) {
+            return;
+        }
+
+        foreach ($this->modx->getIterator($this->classKey, $criteria) as $subMenu) {
+            $this->removeNested($subMenu);
+            $this->modx->runProcessor('system/menu/remove', ['text' => $subMenu->get('text')]);
```
Mark-H, line 50: Is this `$this->modx->runProcessor` call needed if it's already calling `$this->removeNested()` above?

JoshuaLuckers: I think `$this->modx->runProcessor` is needed but `$this->removedNested()` isn't because it's being called in `beforeRemove()`.

alroniks: It's needed because of it a recursive call, `removeNested` removes only items that inside, `runProcessor` removes the current item.

alroniks: @JoshuaLuckers Both needed. The first one (in `beforeRemove`) starts removing after that it runs recursively to reach all levels.

JoshuaLuckers (approved):

Mark-H (approved):

Mark-H:
Thanks @Alroniks :)

PR #14534, opened 2019-04-05
PSR-4 / xPDO 3.x Model Refactor

opengeek:
### What does it do?
Work in progress on xPDO 3.x model refactor of the Revo core featuring PSR-4 autoloading.

### Why is it needed?
To allow the core model to be maintained via xPDO 3.x and to enable PSR-4 autoloading.

### Related issue(s)/PR(s)
n/a

- [x] Finish converting core/model/modx/processors/ to core/src/Processors/
- [ ] Convert sqlsrv model classes
- [x] Remove core/model/modx/*
- [x] Figure out what to do with manager/controllers to enable PSR-4 autoloading without losing ability to rename or relocate the manager/ directory
- [ ] Work on recommended practices for extras to add their component src to the autoloader dynamically (see xPDO->addPackage() PSR-4 support)
- [x] Refactor unit tests for the new classes


opengeek:
There is a lot still to do on this…

OptimusCrime:
This looks great. I was just wondering if GitHub wrongfully reports new files instead of moved files? Are all these classes brand new, or are they just moved from their previous location in core?

opengeek:
They are being manually converted to their new location in `core/src/` and then `core/model/modx/` will only contain a few files to serve as faux class loaders for BC purposes.

gpsietzema:
Nice work! Am I correct in saying that this will just require a lot of time and most strategic thinking has been done? More people can now contribute towards this PR.

JoshuaLuckers:
> Nice work! Am I correct in saying that this will just require a lot of time and most strategic thinking has been done? More people can now contribute towards this PR.

Yes, people are welcome to help convert the processors 👍 ~(without reformatting the code)~.

alroniks:
Added search processor refactoring https://github.com/opengeek/revolution/pull/4

alroniks:
@JoshuaLuckers New processors are new files so I guess code reformatting also can be applied.

JoshuaLuckers:
> Figure out what to do with manager/controllers to enable PSR-4 autoloading without losing ability to rename or relocate the manager/ directory

Can't we simply move `manager/controllers` to `core/src/Manager`?  Various functions in  modManagerController have to be changed to use the autoloader for the core classes of course.

opengeek:
@JoshuaLuckers I don't think they should be in the core at all, honestly. Perhaps we should just leave them as is for now. We can address them at least in a separate PR if the consensus is to convert them and add them to the autoloader dynamically based on the location of the manager folder. This also needs to be able to load theme specific controllers.

opengeek:
I believe @Alroniks is working on the security processors.

rthrash:
This pull request has been mentioned on **Community**. There might be relevant details there:

https://community.modx.com/t/mab-2019-meeting-2-notes-may-2-2019/731/1

sergant210:
> Figure out what to do with manager/controllers to enable PSR-4 autoloading without losing ability to rename or relocate the manager/ directory

Connectors use core classes. But what to do with the core directory? That is the question.
```
"autoload": {
        "psr-4": {
            "MODX\\Revolution\\": "core/src/"
        },
        "classmap": [
            "core/xpdo/"
        ]
},
```

JoshuaLuckers:
Searching in the System Settings is broken. Those issue can be fixed after we merge this PR.

JoshuaLuckers:
Changing values for settings doesn't seem to work either.

JoshuaLuckers (dismissed):

setup/lang/en/default.inc.php
```diff
@@ -213,6 +213,7 @@
 $_lang['toggle'] = 'Toggle';
 $_lang['toggle_success'] = 'Toggle Success Messages';
 $_lang['toggle_warnings'] = 'Toggle Warnings';
+$_lang['upgrade_version_unsupported'] = 'Upgrading from MODX [[+version]] is not supported by this release. You will need to upgrade to MODX 2.6 or later before upgrading to this release.';
```
alroniks, line 216: Why 2.6 and not 2.7?

opengeek: Probably could have supported older, but didn't want to deal with it. There were no upgrade scripts for 2.6, so no reason not to support 2.6+.

PR #14511, opened 2019-03-27
Improvements for recently edited resources

wax100:
### What does it do?
Fix #14334 & #14333

### Why is it needed?
Describe the issue you are solving.

### Related issue(s)/PR(s)
#14334 & #14333

Issue #11264, opened 2014-04-05
Recent Resources: No date and time? FR: View personal manager log!


mindeffects:
Just a little FR/improvement:
When someone wants to know WHEN he edited something, the "Recent Resources" (User > Profile > Recent Resources) does not provide that information. It only says "that" one edited a resource but nothing else/more. It tells my the ID, which I can figure out when I edit the resource (at least in the URL because the ID is nowhere else mentioned (any more)) but I have no chance to get the exact date/time if I have no access to the "manager logs" – which most users don't have.

Maybe a "View personal manager log" (only for MY stuff) would be easy to implement in exchange for the now used "Recent Resouces" listing.

And the functionality is already there, in the "manager logs" listing – just drop some of those select boxes and you are there.

Mark-H:
Date is shown and results filtered on user in #14540.

PR #14540, opened 2019-04-08
Added date and time to the list of recently edited resources in the personal profile

igorsuhinin:
### What does it do?
Adds date and time to the list of recently edited resources in the personal profile.

### Why is it needed?
Currently the list of recently edited resources doesn't show the date & time when the resource was edited.  There is also a small bug: the list of YOUR recently edited resources in the personal profile shows the most recently edited resources without filtering by the current user ID.

### Related issue(s)/PR(s)
Issue #11264 
PR #14511

### DIFF
--- a/core/model/modx/processors/security/user/getrecentlyeditedresources.class.php
+++ b/core/model/modx/processors/security/user/getrecentlyeditedresources.class.php
@@ -91,7 +91,7 @@ class modUserGetRecentlyEditedResourcesProcessor extends modObjectGetListProcess
             return [];
         }
 
-        $resourceArray = $object->get(array('id','pagetitle','description','published','deleted','context_key'));
+        $resourceArray = $resource->get(array('id','pagetitle','description','published','deleted','context_key', 'editedon'));
         $resourceArray['pagetitle'] = htmlspecialchars($resourceArray['pagetitle'], ENT_QUOTES, $this->modx->getOption('modx_charset', null, 'UTF-8'));
 
         $row = array_merge($row, $resourceArray);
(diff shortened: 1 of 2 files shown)


core/model/modx/processors/security/user/getrecentlyeditedresources.class.php
```diff
@@ -58,8 +58,13 @@ public function initialize()
      */
     public function prepareQueryBeforeCount(xPDOQuery $c)
     {
+        $user = $this->getProperty('user');
         $q = $this->modx->newQuery($this->classKey, ['classKey:IN' => $this->classKeys]);
         $q->select('MAX(id), item');
+        if (!empty($user)) {
+            $q->where(['user' => $user]);
+            $c->where(['user' => $user]);
```
alroniks, line 66: Is it required here?

alroniks: Seems like accidentally duplicated lines. Need to leave the first one.

igorsuhinin: > `Seems like accidentally duplicated lines. Need to leave the first one`

Not really, both are required. Please note, there are two different variables **_$q_** and **_$c_**.

I just tried to make as less changes as I could. So I just used the variables which were added before.

alroniks: Ok, I see now. 👍

Mark-H (approved):

Mark-H:
FYI This PR also includes changes already proposed in #14511.

Issue #14501, opened 2019-03-20
Move "Deleted" checkbox from the resource settings


Ruslan-Aleev:
### Summary
In the resource editing tab on the right there are the publishing group, it seems to me that it is right there to place the "Deleted" checkbox.

Firstly, the checkbox is logically associated with the publication, secondly, it will be immediately see that the resource has been deleted or it will be easier to delete it (other settings do not change often, unlike deletion).

![delete](https://user-images.githubusercontent.com/12523676/54695950-5c4e3780-4b44-11e9-8236-8de6ff4ee682.png)

Mark-H:
Fixed in #14545

PR #14545, opened 2019-04-10
Move "Deleted" checkbox from the resource settings to the main tab

igorsuhinin:
### What does it do?
It moves "Deleted" checkbox from the resource settings to the main tab

### Why is it needed?
Looks like this checkbox should be placed in the "Publishing" section as described in the related feature request.

### Related issue(s)/PR(s)
Feature request #14501 

### Comments
I'm not really sure if my decision of using the property "padding: '10px 0 0'" is correct. There could be some special class or something else for the layout (layout: 'column') to follow the general design but I couldn't find it. If anyone has the better idea please let me know.

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -752,16 +752,33 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,getState: function() {
                 return { collapsed: this.collapsed };
             }
-            ,items: [{
-                xtype: 'xcheckbox'
-                ,boxLabel: _('resource_published')
-                ,hideLabel: true
-                ,description: '<b>[[*published]]</b><br />'+_('resource_published_help')
-                ,name: 'published'
-                ,id: 'modx-resource-published'
-                ,inputValue: 1
-                ,checked: parseInt(config.record.published)
-            },{
+            ,items: [
+                [{
+                layout: 'column'
+                ,border: false
+                ,padding: '10px 0 0'
+                ,items: [{
+                        xtype: 'xcheckbox'
+                        ,boxLabel: _('resource_published')
+                        ,hideLabel: true
+                        ,description: '<b>[[*published]]</b><br />'+_('resource_published_help')
+                        ,name: 'published'
+                        ,id: 'modx-resource-published'
+                        ,inputValue: 1
+                        ,checked: parseInt(config.record.published)
+                    },{
+                        xtype: 'xcheckbox'
+                        ,boxLabel: _('deleted')
+                        ,description: '<b>[[*deleted]]</b>'
+                        ,hideLabel: true
+                        ,cls: 'danger'
+                        ,name: 'deleted'
+                        ,id: 'modx-resource-deleted'
+                        ,inputValue: 1
+                        ,checked: parseInt(config.record.deleted) || false
+                    }]
+                }]
+            ,{
                 xtype: 'xdatetime'
                 ,fieldLabel: _('resource_publishedon')
                 ,description: '<b>[[*publishedon]]</b><br />'+_('resource_publishedon_help')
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -752,16 +752,33 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
...
-                ,hideLabel: true
-                ,description: '<b>[[*published]]</b><br />'+_('resource_published_help')
-                ,name: 'published'
-                ,id: 'modx-resource-published'
-                ,inputValue: 1
-                ,checked: parseInt(config.record.published)
-            },{
+            ,items: [
+                [{
+                layout: 'column'
```
JoshuaLuckers:
```suggestion
                layout: 'form'
```

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -752,16 +752,33 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
...
+                ,border: false
+                ,padding: '10px 0 0'
```
JoshuaLuckers:
```suggestion
```

JoshuaLuckers (changes requested):

manager/assets/modext/widgets/resource/modx.panel.resource.js
igorsuhinin: @JoshuaLuckers Joshua, looks like if I do like you proposed then the layout will be changed from the table view to the simple lines. Please see the attached files below (before and after your changes):
before
after

I believe it would be better to show those options in one line. However I'm not sure how to make them both to take 50% of the available width... Do you have any idea?

Thanks,
Igor

JoshuaLuckers: My screen was smaller and they didn't align properly, that's why I moved them to their own row instead. The length of the text in the label is also depending on the lexicon and might break the alignment (can be longer/shorter in other languages).

Ruslan-Aleev: In my opinion, the variant with a width of 50% is better :)

JoshuaLuckers: I agree with it looking better (when there is enough space). We have to figure out if there is a better solution where it looks good in all (or most) situations.

Ruslan-Aleev: I checked languages, everywhere there is enough space for 50%. The only thing that with small windows need to move the deleted line to a new line without padding.

delete_right

igorsuhinin: Guys, I've tried a lot of different formats here including, for example, table view:

`...
                layout: 'table'
                ,layoutConfig: {
                    columns: 2
                }
...
`
However probably due to limitation of the current design (it's not responsive) the result looks quite ugly :(

Table View:
table-view

So currently I'm thinking that the Joshua's idea to distribute the elements by each own row can be the best one. Any other ideas?

JoshuaLuckers: Maybe there is a class that we can use in combination with the `column` layout that has the padding and automatically puts the label on a new row when needed. I'm sure we can figure this out 😎.

igorsuhinin: Yeah, I would be happy to know if there is such class and how it is named :) Maybe someone else will join our discussion and let us know then.

JoshuaLuckers: I tried various things without luck. ExtJS 3 really isn't mobile friendly 😞.

igorsuhinin: So, what would you recommend, Joshua? Should I change the code or leave it as is then?

Mark-H: @JoshuaLuckers, what's your recommendation on this?

JoshuaLuckers: I recommend applying my suggestions and put them in a `row` layout instead of `column`. It might not look perfect but at least it doesn't break alignment (which looks worse in my opinion).

JoshuaLuckers (approved):
Thanks!

Mark-H (approved):

Mark-H:
Thanks @igorsuhinin.

PR #14553, opened 2019-04-12
Added styles for scroll

Ruslan-Aleev:
### What does it do?
Added styles for scrolling.
In MODX on Windows OS there were flaws with a tree on the left due to the large width of the scroll. 

This PR made the scroll universal for different OS and it became more beautiful :)

Before
before

After
scroll

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1805,3 +1805,23 @@ iframe[classname="x-hidden"] {
     margin-top: 0 !important;
   }
 }
+
+::-webkit-scrollbar, ::-webkit-scrollbar-thumb {
+  width: 1rem;
+  height: 1rem;
+  border: .25rem solid transparent;
+  border-radius: .5rem;
+  background-color: transparent;
+}
+
+::-webkit-scrollbar-thumb {
+  box-shadow: inset 0 0 0 1rem rgba(85,108,136,.2);
+}
+
+::-webkit-scrollbar-thumb:hover {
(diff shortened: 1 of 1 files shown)


_build/templates/default/sass/index.scss
```diff
@@ -1805,3 +1805,23 @@ iframe[classname="x-hidden"] {
...
+::-webkit-scrollbar, ::-webkit-scrollbar-thumb {
+  width: 1rem;
+  height: 1rem;
+  border: .25rem solid transparent;
+  border-radius: .5rem;
+  background-color: transparent;
+}
+
+::-webkit-scrollbar-thumb {
+  box-shadow: inset 0 0 0 1rem rgba(85,108,136,.2);
```
JoshuaLuckers: It would be nice if we could [use an existing](https://github.com/modxcms/revolution/blob/3.x/_build/templates/default/sass/_colors-and-vars.scss) or add a new variable for at least the color.

_build/templates/default/sass/index.scss
```diff
@@ -1805,3 +1805,23 @@ iframe[classname="x-hidden"] {
...
+}
+
+::-webkit-scrollbar-thumb:hover {
+  box-shadow: inset 0 0 0 1rem rgba(85,108,136,.3);
```
JoshuaLuckers: Same as above.

JoshuaLuckers (approved):

Ibochkarev (approved):

Mark-H (approved):

PR #14554, opened 2019-04-13
[SQL optimization] Removed FULLTEXT index

sergant210:
Little SQL optimization. For discussion...

### What does it do?
Removed FULLTEXT index from the mysql and sqlsrv schemas of the site_content table.

### Why is it needed?
This is a multi-columns index ('pagetitle', 'longtitle', 'description', 'introtext' and 'content').  I think it was supposed for searching in the table of contents. But it's not used in the core. 
Reason to remove:
- it takes up a lot of disk space;
- it's very expensive for INSERT/UPDATE/DELETE operations.

Let developers or DBAs create it themselves if it needed.
 
### Related issue(s)/PR(s)
No.

### DIFF
--- a/manager/assets/modext/widgets/element/modx.panel.chunk.js
+++ b/manager/assets/modext/widgets/element/modx.panel.chunk.js
@@ -71,7 +71,7 @@ MODx.panel.Chunk = function(config) {
                             'keyup': {scope:this,fn:function(f,e) {
                                 var title = Ext.util.Format.stripTags(f.getValue());
                                 title = _('chunk')+': '+Ext.util.Format.htmlEncode(title);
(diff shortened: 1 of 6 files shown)


JoshuaLuckers:
I'm not convinced that removing the index solves a problem. 

> Reason to remove:
> 
> * it takes up a lot of disk space;

How much of a problem is this? 

> * it's very expensive for INSERT/UPDATE/DELETE operations.

What about SELECT operations? Those are mostly done on the table.

> Let developers or DBAs create it themselves if it needed.

The same can be said about removing the index.

sergant210:
I'm so sorry. I will remove this PR.

JoshuaLuckers:
No need to remove it. Just trying to get a discussion going 😊

sergant210:
It was created for discussion :)
But I need to recreate it right way using custom branch. Could you please save your comment for new PR?

PR #14559, opened 2019-04-16
Fix resource checkbox animation

ilyautkin:
### What does it do?
Fix animation of checkboxes in resource creating/updating page.

### Why is it needed?
When we click any resource checkbox (for example, `Published`), it change state. But type of animation - fading, not slide.
screen-2019-04-16_09 17 22

With this change animation will work correctly:
screen-2019-04-16_09 30 05

### Related issue(s)/PR(s)
-

### DIFF
--- a/_build/templates/default/sass/_forms.scss
+++ b/_build/templates/default/sass/_forms.scss
@@ -685,7 +685,7 @@ input::-moz-focus-inner {
           z-index: 10;
         }
         &:after {
-          left: 1.6em;
+          left: 0.1em;
           top: 0.8em;
           margin-top: -.65em;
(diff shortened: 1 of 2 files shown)


Ibochkarev (dismissed):

Ruslan-Aleev (approved):
Everything works )

checkbox

ilyautkin:
Switched on slider is moved to the right as everywhere in iOS and Android.

Ibochkarev (approved):

JoshuaLuckers (approved):

Mark-H (approved):

PR #14560, opened 2019-04-17
Fix mediasource downloading

wax100:
### What does it do?
Fix Url when you download files from media Source (not Filesystem Media Source)

### Why is it needed?
Because current JS use **r.object.url**, that contain base path of media source and we get /uploads/uploads/ instead of /uploads/ Url. That cause 404 error.

### DIFF
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -198,7 +198,7 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
             ,listeners: {
                 'success':{fn:function(r) {
                     if (!Ext.isEmpty(r.object.url)) {
-                        location.href = MODx.config.connector_url+'?action=browser/file/download&download=1&file='+r.object.url+'&HTTP_MODAUTH='+MODx.siteId+'&source='+this.config.source+'&wctx='+MODx.ctx;
+                        location.href = MODx.config.connector_url+'?action=browser/file/download&download=1&file='+node.attributes.pathRelative+'&HTTP_MODAUTH='+MODx.siteId+'&source='+this.config.source+'&wctx='+MODx.ctx;
                     }
                 },scope:this}
             }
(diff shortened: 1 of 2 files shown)


manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -198,7 +198,7 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
             ,listeners: {
                 'success':{fn:function(r) {
                     if (!Ext.isEmpty(r.object.url)) {
```
Ruslan-Aleev, line 200: Is there no need to change the `if (!Ext.isEmpty(r.object.url))` too? And in another file.

JoshuaLuckers (approved):

manager/assets/modext/widgets/media/modx.browser.js
JoshuaLuckers: This checks if there is an URL in the returned result. The value `node.attributes.pathRelative` is always defined and therefore not empty. 

I most likely have to reconsider my approval for this PR. It's better to fix the `modBrowserFileDownloadProcessor` processor and return the right URL so we can use the actual result instead of the "workaround".

Mark-H: My gut says this also should be sorted server-side - otherwise, what's the `url` for?

Mark-H (changes requested):
Should be fixed in the processor, not the javascript.

Mark-H:
Any update on this @wax100? My previous comment was a little brief, but I believe this should be sorted in the server-side generation of the url rather than the client-side. If you don't agree or if you  need help updating the PR, please do let me know.

Mark-H:
This PR needs to be updated in the next 24hours or so to make it into the alpha1 release.

Ibochkarev:
> Any update on this @wax100? My previous comment was a little brief, but I believe this should be sorted in the server-side generation of the url rather than the client-side. If you don't agree or if you need help updating the PR, please do let me know.



> Should be fixed in the processor, not the javascript.

@Jako Can you help with a solution in processors?

Mark-H:
No response in over 2 years, so closing this. I also believe there have been others tackling this issue differently, e.g. #15512, and I'm not currently reproducing this issue.

Issue #8371, opened 2012-08-08
File Browser allows silent overwriting of files


MokoJumbie:
MokoJumbie created Redmine issue ID 8371

The File Browser should notify you if a file will be overwritten when uploading files. Currently, a file can be overwritten on upload without even a warning.

RinzeVanDerBrug:
Going to test this today.

RinzeVanDerBrug:
Able to reproduce this.

Tested this in MODX 2.5.5. When uploading a file with a name which already exists in the file manager, the system will tell you your upload was successful. When you refresh the file manager afterwards, you'll see that the existing file is still there and the file you uploaded 'successfully' is nowhere to be found. 
![image](https://cloud.githubusercontent.com/assets/25863377/23552441/269f6048-001c-11e7-8087-f2a7b5355618.png)

Confirmed. #modxbughunt

meshkov:
Tested it in MODX 3.0 and it is still actual for MODX 3.0.
And actual for drag&drop packages in package manager. 
You can see also on the demo in #13851

iWatchYouFromAfar:
This problem is very relevant so far. Almost on all sites that i create, i have to make image renaming plugins to avoid duplicate names. At the same time, silent renaming does not work if you rename a file already uploaded to the server. 

I reproduced on the current dev branch MODX 3.x. Any ideas for fixing this problem?
![file-browser](https://user-images.githubusercontent.com/16403917/51986356-1f08ea00-24b1-11e9-9cf3-66e95b67e84e.gif)

sebastian-marinescu:
Silently overwriting doesn't seem good. A confirmation of overwriting the file would be the best UX here.

pyrographics:
Rather than failing to upload or overwriting a file it should add _1, _2, _3 or similar to the filename of the duplicates.  This is what most browser downloads do.

Ibochkarev:
Improved in #15232 

@modxbot close

PR #14514, opened 2019-03-30
Fix #5700 & #8371

webinmd:
Fix form customization for weblink and symlink #5700
Fix identical file names on upload #8371

Related issue(s)/PR(s)
#5700
fc-weblink

#8371
identical_file_names

PR #14566, opened 2019-04-24
Generate new file if file already exists

sdrenth:
### What does it do?
When uploading a file it checks if a file already exists with the same file path. If so, it will generate a unique filename and upload it as a new file instead of overwriting the existing file.

**Example:**
- A file /directory/image.png already exists
- When uploading a new image.png file to the /directory/ it will upload the file as /directory/image-01.png

### Why is it needed?
This will prevent files from being overwritten if an image already exists with the same file name.

### Relevant issue(s)

#8371

### DIFF
--- a/core/model/modx/sources/modmediasource.class.php
+++ b/core/model/modx/sources/modmediasource.class.php
@@ -1062,6 +1062,12 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             }
 
             $newPath = $container . $this->sanitizePath($file['name']);
+
+            /* Check if file already exists. */
+            if ($this->filesystem->has($newPath)) {
+                $newPath = $this->generateUniqueFilename($newPath);
+            }
+
             try {
                 if (!$this->filesystem->put($newPath, file_get_contents($file['tmp_name']))) {
                     $this->addError('path', $this->xpdo->lexicon('file_err_upload'));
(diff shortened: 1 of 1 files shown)


alroniks (changes requested):
I don't like such hidden and silence behavior. It would be much better to show confirm window and as a user what to do and suggest create new but also leave the ability to override. As it works on operation systems (windows, mac os, etc.)

gpsietzema:
I agree, that was what we were also aiming for. We're working on multiple usability improvements for 3.x this week, but the ideal situation (multiple choice - let the user choose) was impossible to do in the amount of time we have at our hands this week. So we settled for this solution, which is still a huge improvement for the ultra-silent-just-overwrite-the-shit-out-of-it current situation.

Mark-H:
What about simply refusing the upload (file already exists)? Plugins can implement whatever business logic best suits the specific user without this necessarily being something we have to get right for everyone in the core. 

(That's implemented in #14514)

Jako:
@gpsietzema If it would be not a degression (you can't upload the same file again to overwrite it), I would be comfortable with this patch.

sdrenth:
@gpsietzema What are your thoughts on this one?

JoshuaLuckers:
It's all about telling users what action to take to solve the issue or/and inform them about what decisions the system has made for them instead. I don't mind if the file gets another name if I'm being told what and why. 

The best solution would be to ask the user what action to take when a file with the same name exists:
1. Cancel the upload
2. Overwrite the old file
3. Keep both

gpsietzema:
I agree to what you're saying, but for us it's not an option to make this work now, because it requires a ton of work and possible rewrite of half of that media browser. So I'd go with this first as an improvement to the current way it works...

Mark-H (changes requested):
If a file already exists the upload should be rejected, generating a new file name should be left to plugins (which is already possible).

gpsietzema:
Fine, we'll just use this in our own amazing MODX-install.

sdrenth:
Since this PR won't be merged into the core I'll close this PR. The ultimate fix was described in this issue by @JoshuaLuckers and I agree that that would be the best solution, but we are unable to fix the issue as described. Therefor I created a new issue for this one and added the described solution by @JoshuaLuckers there and hopefully somebody finds the time to fix the issue as described

PR #14569, opened 2019-04-25
Feature welcome widget

sdrenth:
### What does it do?
Added a new welcome widget for the dashboard page showing a welcome message and current time. The background image can be changed by the newly added system setting `widget_welcome_background`.

### Why is it needed?
Displays a neat welcome message when visiting the dashboard page.

Screenshot 2019-04-25 at 10 03 17


Electrica:
You are seriously? )))

Mark-H:
Two things:

- Where is that image from and under what license is it available?
- Style changes should be made in the sass files (in `_build/templates/default/`), not the output css file.

gpsietzema:
The image is from Unsplash.

sdrenth:
@Mark-H I moved to css changes to the sass files

_build/templates/default/sass/_dashboard.scss
```diff
@@ -355,3 +355,77 @@
...
+    .body {
+      flex: 1;
+      max-height: 100% !important;
+      min-height: 300px;
+      overflow: hidden !important;
+      padding: 0 !important;
+    }
+
+    .time {
+      color: #fff;
```
Ibochkarev: Please change to the variable `$white`

_build/templates/default/sass/_dashboard.scss
```diff
@@ -355,3 +355,77 @@
...
+    .message-wrapper {
+      align-items: center;
+      display: flex;
+      height: 100%;
+      justify-content: center;
+      position: relative;
+      z-index: 1;
+
+      h3 {
+        color: #fff;
```
Ibochkarev: Please change to the variable `$white`

Ibochkarev (changes requested):

manager/templates/default/css/index-min.css
```diff
@@ -1,2 +1,2 @@
...t:"\f07b"}.fa.fa-folder-open-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-folder-open-o:before{content:"\f07c"}.fa.fa-smile-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-smile-o:before{content:"\f118"}.fa.fa-frown-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-frown-o:before{content:"\f119"}.fa.fa-meh-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-meh-o:before{content:"\f11a"}.fa.fa-keyboard-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-keyboard-o:before{content:"\f11c"}.fa.fa-flag-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-flag-o:before{content:"\f024"}.fa.fa-mail-reply-all:before{content:"\f122"}.fa.fa-star-half-o{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-star-half-o:before{content:"\f089"}.fa.fa-star-half-empty{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-star-half-empty:before{content:"\f089"}.fa.fa-star-half-full{font-family:'Font Awesome 5 Free';font-weight:400}.fa.fa-star-half-full:before{content:"\f089"}.fa.fa-code-fork:before{content:"\f126"}.fa.fa-chain-broken:before{content:"\f127"}.fa.fa-shield:before{content:"\f3ed"}.fa.fa-calendar-o{font-family:'Font Awesome 5 Fr
```
Ibochkarev: Please don't commit the compiled CSS. The integrator who merges the PR will compile the final CSS.

manager/templates/default/css/login.css
```diff
@@ -86,12 +86,9 @@ body {
   padding: 0 1.5625rem; }
```
Ibochkarev: Same

manager/templates/default/css/index.css
```diff
@@ -513,33 +513,27 @@ template {
```
Ibochkarev: Same

Ibochkarev (changes requested):

Ibochkarev (dismissed):

JoshuaLuckers:
I don't think that a non-functional dashboard widget like this should be in the core. 

1. It's wasting to much useful space
2. It's showing the time on the server which might confuse users if it deviates from the users time. This can be useful if properly presented to the user. 
3. Do we really want to load more assets like this image in the manager?

gpsietzema:
1. The point is to make this dashboard prettier. That's an important function. Make people happier to use the system.
2. Sure, any ideas?
3. Yes. It's needed for this.

JoshuaLuckers:
> * The point is to make this dashboard prettier. That's an important function. Make people happier to use the system.

Fair enough! The functions provided by the widget in its current state, in my opinion, don't justify the amount of space it uses.  I suggest to provide additional functionality personalised for the current user. 
I'll propose some ideas later. 

> * Sure, any ideas?

Not right now.

> * Yes. It's needed for this.

Haha 🤣. It would be nice to mention the source/link of the image in the code as comment for future reference.

alroniks:
It would be nice to collect those nice widgets in the separated extra. It will allow us to have them updated separately and more frequently and do not waste the time of core integrations for such small improvements. 
It nice and I like it, of course,  but from a common perspective and long term view these things will have to be maintained later in case of changes and this will take time away.

alroniks:
@gpsietzema @sdrenth I can help with creating such extra and I can maintain it if you agree.
Also, as it was described in other issues, a bunch of extras can be suggested user for installation on new setup interface and this extra would be the best candidate for such list.

gpsietzema:
Sounds like a good idea @Alroniks - I really like it.

sdrenth:
I've added the source URL of the default image in: transport.core.system_settings.php

sdrenth:
@Alroniks Any progress on this one? 

Also any thoughts about which widgets should be extracted from the core into an Extra and which should stay in the core?

I feel like most widgets should stay in the core and this one should be added into an extra. But for the extra to also add some more value I feel like there should be some more enhanced dashboard widgets added into that extra. But that could also be something which we could expand later on.

alroniks:
I would like to leave only top buttons, security check and feeds (I don't like them but they probably needed).
Others should be and can be moved to the extra. 
I have not started anything yet, but it should be done at the same time with adding additional screen for an installer with suggested add-ons. To have it preinstalled and keep a good user experience.

gpsietzema:
@Alroniks You can already start with the dashboard-widgets extra; it doesn't need to depend on the additional installer screen. Is anyone working on that by the way? I know @Mark-H made a mockup once.

Mark-H:
Closing as there's consensus to move additional widget types into extras.

Issue #14565, opened 2019-04-23
Animation upload file not working


Electrica:
### Summary
When i upload file, animation not working.
Link https://yadi.sk/i/W4skj_75l0EROg

### Environment
MODX Revolution 3.0.0-pl nginx

alroniks:
Solved.

Issue #13962, opened 2018-07-03
MODX 3: Media Browser - uploads can't be seen until refresh


jonleverrier:
### Summary
When you upload a file in the Media Browser either by using the upload icon or dragging files onto the manager, you can't see that they have been successfully uploaded into you refresh your view.

### Step to reproduce
1. Upload a file using the upload icon or by dragging files on to the browser

### Observed behavior
After uploading a file the page isn't updated automatically. You have to click on the folder you just uploaded to, or refresh the page to use the file.

![modx3_mediabrowser_uploadnorefresh](https://user-images.githubusercontent.com/2373940/42220016-15c64476-7ece-11e8-922f-be63bb2c4939.gif)

### Expected behavior
After upload you can see the file.

### Environment
MODX3 Alpha v 3.0.0-dev
From https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99

meshkov:
Confirm.
Also problem with drag&drop files.
![ifchbiz5j7](https://user-images.githubusercontent.com/1703670/42278146-7bb9222e-7fa2-11e8-848d-b01b9213c158.gif)

Ibochkarev:
I confirm this bug. Not updated in both cases.

3.0.0-pl

alroniks:
Solved.

PR #14570, opened 2019-04-25
Media browser reload

joeke:
### What does it do?
Add the MODx.browser Ext component to the uploader, and call the browser.run() method when uploading via drag-n-drop, the main upload button, or the tree right click upload button.

### Why is it needed?
In 3.x after uploading files in the media browser (through drag-n-drop or upload button) the view does not get updated with the newly uploaded files. This PR fixes that.

### Related issue(s)/PR(s)
- #14565
- #13962

### DIFF
--- a/manager/assets/modext/util/multiuploaddialog.js
+++ b/manager/assets/modext/util/multiuploaddialog.js
@@ -107,7 +107,7 @@
         onInputFileChange: function (e) {
             var files = FileAPI.getFiles(e);
             this.input_file.value = '';
-            this.startUpload(files);
+            this.startUpload(files, this.browser);
         },
 
         startUpload: function (files, browser) {
--- a/manager/assets/modext/util/multiuploaddialog.js
+++ b/manager/assets/modext/util/multiuploaddialog.js
@@ -167,7 +167,7 @@
                             upload.progress.hide();
                         }
                         // Reload the modx-browser tree/dataview
-                        if (browser && typeof browser.run !== "undefined") {
+                        if (typeof browser !== "undefined" && typeof browser.run !== "undefined") {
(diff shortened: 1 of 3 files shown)


manager/assets/modext/util/multiuploaddialog.js
```diff
@@ -166,6 +166,10 @@
                         if (upload.progress) {
                             upload.progress.hide();
                         }
+                        // Reload the modx-browser tree/dataview
+                        if (typeof browser.run !== "undefined") {
```
Mark-H: Could there be a case where `browser` itself is null?

joeke: @Mark-H It shouldn't happen, but yes, it possibly could happen :). I will add a check for `browser`.

Mark-H (dismissed):

manager/assets/modext/util/multiuploaddialog.js
```diff
@@ -166,6 +166,10 @@
                         if (upload.progress) {
                             upload.progress.hide();
                         }
+                        // Reload the modx-browser tree/dataview
+                        if (browser && typeof browser.run !== "undefined") {
```
JoshuaLuckers:
```suggestion
                        if (typeof browser !== "undefined" && typeof browser.run !== "undefined") {
```

JoshuaLuckers (changes requested):

JoshuaLuckers (approved):

sdrenth:
@Mark-H Since you've been reviewing this PR before, could you have another look to see if it is ready to be approved/merged now?

Mark-H (approved):

Issue #14499, opened 2019-03-19
Design Suggestions for MODX3


kolbykruger:
### Summary
This is my (quick) take at what the MODX3 resource panel and resource sections should look like. Inspiration and colors are all from Stripe's manager page. Along with this, I think it would be really beneficial for a system setting to allow you to set the accent color. This color would attribute itself to the save button (?) as well as other borders/focus/switches.

### Why is it needed?
In the current design, there are too many different off-white/grey colors that distract you from your workflow, and take away from the overall clean, modern design.

### Suggested solution(s)
Here are the colors and special css attributes:
Background color: #e3e8ee
Panel color: #f7fafc
Burple (accent): #556cd6
Text color: #2a2f45
Faded Text color: #a3acb9
Sub-text color: #4f566b
Error color: #cd3d64
Line shadow: box-shadow: inset 0 -1px #e3e8ee;
Box shadow: box-shadow: 0 7px 14px 0 rgba(60,66,87, 0.1), 0 3px 6px 0 rgba(0, 0, 0, .07);
Border radius: 4px;

### Screenshots of the suggestions
![resource](https://user-images.githubusercontent.com/10890005/54647562-50e20a00-4a79-11e9-8095-f6321d37ac88.png)
![chunk](https://user-images.githubusercontent.com/10890005/54647561-50e20a00-4a79-11e9-9520-c7e3ba7a864c.png)

alroniks:
Look nice. But why blue instead or current green?

Ruslan-Aleev:
**It looks great!**
The only thing is, I would change the indent of the background-block of the buttons (like in the content block) or remove the background-block completely.

![button-bg](https://user-images.githubusercontent.com/12523676/54681862-db347780-4b26-11e9-8b4a-eb3b48c8fb23.png)

Ibochkarev:
A couple of comments on the proposed changes:
- change green to blue? What for? Green looks consistent and it is in the logo.
- gray is not so agreed look
- also the `box-shadow` property

The current version by color is an exact hit as it should be. I would not like to see changes on these items. 

But there are also things that can be made as changes.

Ruslan-Aleev:
@Ibochkarev The base color can be changed, for example, in the settings, so it is not green, as I understand it.
Personally, I like gray, the usual gray color :)

Ibochkarev:
https://twitter.com/vlijmscherp/status/892410969464418305 This is something new and very interesting in my opinion.

[...]

PR #14572, opened 2019-04-25
Minor design tweaks and fixes for MODX3

joeke:
### What does it do?

1. Removed box-shadow on most of the panels / tabs.
2. Slightly darken main background color (from #f2f2f2 to #f1f1f1) for a bit more contrast between tabs and background.
3. Fix panel collapse toggle arrow to correctly align in center
4. Decrease padding on `modx-action-buttons` element to better align with rest of elements, and prevent overlapping on pages without tabs
5. Make use of `$mainBg` sass variable where hardcoded background with same value was used.
6. Set `.panel-desc` background to white for more consistency with other panels.

### Why is it needed?
To give MODX3 a bit more modern look, and fix some inconsistencies in the sass files.

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/_browser.scss
+++ b/_build/templates/default/sass/_browser.scss
@@ -5,7 +5,6 @@
 .modx-browser-tree {
   background: $white;
   border-radius: $borderRadius;
-  box-shadow: $boxShadow;
 
   .x-window &,
   .modx-browser-rte & {
@@ -18,7 +17,6 @@
 .modx-browser-view-ct {
   background: $white;
   border-radius: $borderRadius;
-  box-shadow: $boxShadow;
   font: $fontSmall;
 
   .x-window &,
@@ -196,7 +194,6 @@
 .modx-browser-details-ct {
   background: $white;
   border-radius: $borderRadius;
-  box-shadow: $boxShadow;
 
   .x-window &,
   .modx-browser-rte & {
--- a/_build/templates/default/sass/_buttons.scss
+++ b/_build/templates/default/sass/_buttons.scss
@@ -261,8 +261,8 @@ button {
   top: 0;
   right: 0;
   left: auto;
-  background: #F2F2F2;
-  padding: 1rem 1.25rem 1rem 1rem;
+  background: $mainBg;
+  padding: 0.8rem 1rem;
   border: 0;
   font-family: $buttonfonts;
   border-radius: $borderRadius;
--- a/_build/templates/default/sass/_colors-and-vars.scss
+++ b/_build/templates/default/sass/_colors-and-vars.scss
@@ -47,7 +47,7 @@ $brandSelectedBg: darken($brandHover, 3%);
 $brandSelectedColor: $colorSplash;
 
 /* Backgrounds */
-$mainBg: #F2F2F2;
+$mainBg: #F1F1F1;
 $lightBg: $lighterGray;
 $mediumBg: $lighterGray;
 $darkBg: $darkestGray;
@@ -135,7 +135,7 @@ $treeBgSelected: #D6E7F8;
 $treeColor: $darkGray;
 $treeColorDesaturated: lighten(desaturate($treeColor,100%), 25%);
 $treeColor_2: #728DA7;
-$treePseudoRootBg: #F2F2F2;
+$treePseudoRootBg: #F1F1F1;
 $treePseudoRootColor: $darkestGray;
 $treePseudoRootOverBg: $brandHover;
 $treePseudoRootOverColor: $darkestGray;
--- a/_build/templates/default/sass/_dashboard.scss
+++ b/_build/templates/default/sass/_dashboard.scss
@@ -7,8 +7,8 @@
 
   .dashboard-button {
     padding: 5px 20px;
-    border: 1px solid $borderColor;
     border-radius: $borderRadius;
+    border: 1px solid transparent;
     background: $white;
     text-decoration: none;
     display: inline-block;
@@ -24,7 +24,7 @@
     }
 
     &:not([disabled]):hover {
-      box-shadow: $boxShadow;
+      border-color: $borderColor;
     }
   }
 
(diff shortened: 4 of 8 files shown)


Ruslan-Aleev:
Can you make screenshots to make the changes clearer?
See also https://github.com/modxcms/revolution/issues/14499 :)

joeke:
@Ruslan-Aleev Yes I will add some screenshots.

joeke:
@Mark-H @opengeek Please wait with reviewing / merging, somehow the modxcms/3.x got merged into this. Will fix it.

manager/templates/default/css/index.css
```diff
@@ -21720,17 +21720,72 @@ body.x-body-masked .x-window-plain .x-window-mc {
           flex-wrap: nowrap;
```
Ibochkarev: Please don't commit the compiled CSS. The integrator who merges the PR will compile the final CSS.

Ibochkarev (changes requested):

JoshuaLuckers:
@joeke if you need any help let us know 👍

joeke:
@Mark-H @JoshuaLuckers @opengeek  PR is fixed now, I've reverted the 3.x merge commit so the correct file changes remain. The 3.x merge and it's commits is however still in the commit list, but the `Files changed` shows the correct changes. So reviewing can commence :).

Ibochkarev:
@joeke Please fix conflicts

joeke:
@Ibochkarev Conflicts should be solved now.

Ibochkarev (approved):

JoshuaLuckers:
I like the manager better with box-shadows and I'm not _yet_ convinced that removing them is an improvement. Can you elaborate why removing them gives the manager a more modern look? 

Screenshot with box-shadows:
modx3-manager-box-shadows

Screenshot without box-shadows:
modx3-manager-box-shadows-removed

Ruslan-Aleev:
@JoshuaLuckers Subjectively :) I like more without shadows

rthrash:
The resource/element/file tree interior looks far better without the box shadows, but I prefer them elsewhere. Perhaps they could be more subtle/less offset, though?

Ibochkarev:
I am for leaving box-shadow. I missed this change ...

Ruslan-Aleev:
@kolbykruger Participate in the discussion? Maybe you have any suggestions?

Ruslan-Aleev:
In my opinion, without a gray background, the description of the section does not attract attention and "hangs in the void", see the picture below.

after

Maybe it makes sense to add a light gray on the background?

kolbykruger:
I agree. I think it needs a grey background to add some depth. 

On another note, why do these show by default? They're helpful for the first time user of MODX, but for most people I think they're unnecessary. They should realistically be moved to a question mark icon that displays a tooltip on hover in my opinion. 

Edit: Or maybe allow us to hide them via a system setting?

rthrash:
And likely rewritten to streamline them a lot. "Here you can…" is a useless introductory clause.

For example on the above screenshot, the following would be a more concise and clearer description:

> Configure your MODX installation by double-clicking the values to edit inline or right-clicking for more options. Settings are available via their respective `[[++key-name]]` placeholders.

IMO descriptions should be shown by default, with `show/hide descriptions` button above the grids (depending on if they're showing or not).

kolbykruger:
As for the box-shadow vs no box-shadow, I prefer the flat version. However it does create some conflict with the sidebar elements. They would probably have to be changed to a different shade of grey. 

I personally still think the interface is too busy with clutter.

joeke:
Hi everybody! Nice to see I've stirred up some conversation 😉. Here are some of our thoughts behind the changes:

The drop-shadows are not very pretty on a lot of places (especially sidebar / tree) and make the interface look more cluttered. Admittedly on some places they don't look all that bad, but for the sake of consistency we dropped (😄) them on all the elements.

The white background on the description field is added to make the active tab + tab contents look more connected. The main reason is the resource editing page, where for example the 'Document' and 'Settings' tabs, when active, are white with also white background on the tab contents. But if you look at the 'Resource Groups' tab it just looks weird, look at these screenshots.

Screen Shot 2019-05-15 at 09 19 43

Screen Shot 2019-05-15 at 09 19 53

Also the gray color of the description (#F0F0F0) is not the same (but very close to) the main body background color (#FAFAFA), which looks weird when they're next to each other (as you can see in the screenshot).

Ruslan-Aleev:
@joeke Maybe it is worth making gray lighter, different from the background or choose a different color?

kolbykruger:
Knowing @joeke 's description I would argue that the textblock gets the same margining as the below fields, and that it get a different background treatment. Either a different grey color, or something that makes it defined as a 'tip' or 'section description'

Edit: It might help to prevent the 'detached' feel, as well as create consistency with the below fields

gpsietzema:
I'd first go with this PR for fixing the weirdness and then create a new one for @kolbykruger 's (imho good) ideas.

JoshuaLuckers (approved):

Mark-H (changes requested):
This PR needs to be rebased (or recreated if that's easier) to prevent including unrelated commits.

Ibochkarev:
@joeke please make rebase

Mark-H:
I've gone ahead and rebased the 2 relevant commits. There was an additional change related to the leftbar setting a minimum width which I removed from the PR as that was already addressed in another recent PR.

Mark-H (approved):

Mark-H:
For the record @joeke, to resolve this sort of issue where unrelated commits are included in a PR, use an interactive rebase instead of merge:

```
git rebase -i upstream/3.x
```

In the edit that pops up, start lines with `pick` that you want to include, and `drop` with PRs to remove. Then save (`esc` + type `:wq`) and the rebase will start applying your commits one-by-one onto `upstream/3.x`. 

When it conflicts, resolve the conflict and `git add` the conflicted files, and continue rebasing by typing `git rebase --continue`. When done rebasing, force push `git push --force origin branch-name` and the PR will be updated and clean.

PR #14574, opened 2019-04-27
Enhance the main application class

sergant210:
For discussion...

### What does it do?
Adds ability for enhancing the main application class (modX). And move the composer autoloader from modX class to the index.php.

### Why is it needed?
This PR allows to extend the modX class correctly following the "Open/Close" principle. All you need is to specify the desired class in the `config.core.php` file. 
One more advantage - no need to pass the modx object to classes like that
```php
function __construct(modX &$modx) {
        $this->modx= & $modx;
    }
```
Now you can get the application object at any part of yuor application
```php
function __construct() {
        $this->modx = modX::getInstance();
    }
```

The same must be done for `manager/index.php` and `connectors/index.php`.

### Related issue(s)/PR(s)
No.

### DIFF
--- a/_build/resolvers/resolve.core.php
+++ b/_build/resolvers/resolve.core.php
@@ -15,7 +15,7 @@ switch (MODX_SETUP_KEY) {
                     . " */\n"
                     . "define('MODX_CORE_PATH', '" . MODX_CORE_PATH . "');\n"
                     . "define('MODX_CONFIG_KEY', '" . MODX_CONFIG_KEY . "');\n"
-                    . "define('MODX_APP_CLASS', 'modX');";
+                    . "define('MODX_APP_CLASS', modX::class);";
                 $written= $cacheManager->writeFile($targetFile, $configContent);
                 $success= $written !== false ? true : false;
             }
--- a/_build/resolvers/resolve.core.php
+++ b/_build/resolvers/resolve.core.php
@@ -15,7 +15,7 @@ switch (MODX_SETUP_KEY) {
                     . " */\n"
                     . "define('MODX_CORE_PATH', '" . MODX_CORE_PATH . "');\n"
                     . "define('MODX_CONFIG_KEY', '" . MODX_CONFIG_KEY . "');\n"
-                    . "?>";
+                    . "define('MODX_APP_CLASS', 'modX');";
                 $written= $cacheManager->writeFile($targetFile, $configContent);
                 $success= $written !== false ? true : false;
             }
--- a/_build/templates/config.core.php.txt
+++ b/_build/templates/config.core.php.txt
@@ -6,3 +6,4 @@
  */
 define('MODX_CORE_PATH', @core-path@);
 define('MODX_CONFIG_KEY', 'config');
+define('MODX_APP_CLASS', 'modX');
--- a/_build/templates/config.core.php.txt
+++ b/_build/templates/config.core.php.txt
@@ -6,4 +6,4 @@
  */
 define('MODX_CORE_PATH', @core-path@);
 define('MODX_CONFIG_KEY', 'config');
-define('MODX_APP_CLASS', 'modX');
+define('MODX_APP_CLASS', modX::class);
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -431,7 +431,7 @@ class modX extends xPDO {
      * @throws xPDOException
      */
     public static function getInstance($id = null, $config = null, $forceNew = false) {
-        $class = __CLASS__;
+        $class = defined('MODX_APP_CLASS') ? MODX_APP_CLASS : __CLASS__;
         if (is_null($id)) {
             if (!is_null($config) || $forceNew || empty(self::$instances)) {
                 $id = uniqid($class);
--- a/index.php
+++ b/index.php
@@ -8,43 +8,55 @@
  * files found in the top-level directory of this distribution.
  */
 
-$tstart= microtime(true);
+$tstart = microtime(true);
+
+if (!function_exists('modx_site_unavailable')) {
+    /**
+     * @param string $errorMessage
+     * @param string $errorPageTitle
+     */
+    function modx_site_unavailable($errorMessage, $errorPageTitle = '') {
+        @include MODX_CORE_PATH . 'error/unavailable.include.php';
+        /* if including is failed */
+        echo "<html><title>Error 503: Site temporarily unavailable</title><body><h1>Error 503</h1><p>{$errorMessage}</p></body></html>";
+        exit();
+    }
+}
 
 /* define this as true in another entry file, then include this file to simply access the API
  * without executing the MODX request handler */
-if (!defined('MODX_API_MODE')) {
-    define('MODX_API_MODE', false);
+defined('MODX_API_MODE') or define('MODX_API_MODE', false);
+
+/* include custom core config and define some important constants */
+@include __DIR__ . '/config.core.php';
+
+defined('MODX_CORE_PATH') or define('MODX_CORE_PATH', __DIR__ . '/core/');
+defined('MODX_APP_CLASS') or define('MODX_APP_CLASS', 'modX');
+
+/* include the composer autoloader */
+if (!file_exists(MODX_CORE_PATH . 'vendor/autoload.php')) {
+    modx_site_unavailable('Site temporarily unavailable - missing dependencies.');
 }
 
-/* include custom core config and define core path */
-@include(dirname(__FILE__) . '/config.core.php');
-if (!defined('MODX_CORE_PATH')) define('MODX_CORE_PATH', dirname(__FILE__) . '/core/');
-
-/* include the modX class */
-if (!@include_once (MODX_CORE_PATH . "model/modx/modx.class.php")) {
-    $errorMessage = 'Site temporarily unavailable';
-    @include(MODX_CORE_PATH . 'error/unavailable.include.php');
-    header($_SERVER['SERVER_PROTOCOL'] . ' 503 Service Unavailable');
-    echo "<html><title>Error 503: Site temporarily unavailable</title><body><h1>Error 503</h1><p>{$errorMessage}</p></body></html>";
-    exit();
-}
+require 'vendor/autoload.php';
 
 /* start output buffering */
 ob_start();
 
-/* Create an instance of the modX class */
-$modx= new modX();
+/* Create an instance of the application class */
+try {
+    $modx = modX::getInstance(MODX_APP_CLASS);
+} catch (Exception $e) {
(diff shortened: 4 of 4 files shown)


Ruslan-Aleev:
@Alroniks  @Jako @JoshuaLuckers @Mark-H @opengeek It would be great to see your comments :) any ideas on this PR?

_build/resolvers/resolve.core.php
```diff
@@ -15,7 +15,7 @@
                     . " */\n"
                     . "define('MODX_CORE_PATH', '" . MODX_CORE_PATH . "');\n"
                     . "define('MODX_CONFIG_KEY', '" . MODX_CONFIG_KEY . "');\n"
-                    . "?>";
+                    . "define('MODX_APP_CLASS', 'modX');";
```
alroniks:
```suggestion
                    . "define('MODX_APP_CLASS', modX::class);";
```

_build/templates/config.core.php.txt
```diff
@@ -6,3 +6,4 @@
  */
 define('MODX_CORE_PATH', @core-path@);
 define('MODX_CONFIG_KEY', 'config');
+define('MODX_APP_CLASS', 'modX');
```
alroniks:
```suggestion
define('MODX_APP_CLASS', modX::class);
```

Mark-H:
Hey @sergant210, I think this looks like a good improvement. Two notes: 

- The composer autoloader would probably be better to stay in the modX class. The index.php is *not* the only way modX can be initialised, so having it at the top of the modx class would make sure it's always available.

- This would likely need to be rebased or potentially reimplemented now that the xPDO 3/namespaces refactor has landed in 3.x. 

I'd also be interested to see if @opengeek has thoughts on this from an architectural point of view, but my opinion is that this is a change that for regular users doesn't change anything (modX is still modX), but would allow a new level of flexibility in the future, which is great.

opengeek:
I am not necessarily in favor of making the modX class itself more extensible, at least not in the current architecture. We already have too much flexibility at times. In fact, I was considering declaring the class `final.` I'm concerned that encouraging extensions to the modX class would open up a whole new set of problems and make upgrade paths even more difficult. I'm also not a fan of using static instance methods to retrieve MODX as a dependency. Dependencies should be explicit, in my opinion.

The location where Composer autoloading gets included is another issue, but I think that should be discussed separately from this PR.

bezumkin:
@opengeek 
> I'm concerned that encouraging extensions to the modX class would open up a whole new set of problems and make upgrade paths even more difficult

What problems? Do you have a lot of problems with custom cache classes? Or session handlers?

it doesn't matter how I'm extending MODX on my side. But right now I need to use custom `index.php` to load custom MODX class. This PR could allow me to do much more, extend and override any MODX protected method for my projects.

> I was considering declaring the class `final`

It really worries me that you could do something like this on your own.

JoshuaLuckers:
We should have an interface for the MODX class.

sergant210:
@JoshuaLuckers Current architecture doesn't allow it. The modX class is a subclass of the xPDO class.

opengeek:
@bezumkin By allowing the modX class to be extended, we make it much more difficult to control an upgrade path. This means anytime we need to change anything in the main class, we would most likely break yours. What would be better is to know what you are extending in the modX class and discuss how that could be attained without using a custom modX class. The movement should be towards making things work as proper dependencies.

opengeek:
>> I was considering declaring the class final

> It really worries me that you could do something like this on your own.

Considering does not mean "on my own." In fact, I just posted about here because of this issue. But if you would like to actually participate in development again after disappearing for a year, I'd love to get more of your input.

Mark-H:
Makes sense to me. The modX class is already doing so much and making it more powerful through extension (as @sergant210 correctly stated, it's also already extending from xPDO, which has been the biggest problem with moving 3.x forward) doesn't seem like the best path forward. 

As @bezumkin stated, we do have extension points for various services already, so anything currently not possible would be better accomplished through such a more decoupled interface.

bezumkin:
@opengeek How will you break my class if I just extend default?

```php
class BezumkinMODX extends modX {

    public function loadUser() {
        // ... And so on
    }

}
```

Mark-H:
@bezumkin Perhaps a new `LoginService` would be introduced that can be passed as optional parameter to `loadUser()`, now your class has an incompatible signature. That's at least going to trigger warnings but as of 7.2 may also throw fatal errors (* tho that seems to be limited to when interfaces are involved).

bezumkin:
@Mark-H and I will get an error and will fix it in 5 seconds.

That is not a problem. You can change session handler signature as well. How often are we introduce new parameters in minor and patch releases, btw?

Looks like this PR not going be approved because "something could happen. We don't know what exactly, but something scary".

Mark-H:
> How often are we introduce new parameters in minor and patch releases, btw?

If a class is open for extension like this, and people may have built custom implementations what _will break_, then yeah, that will definitely hold back any attempt at innovating the internals.

I think you would be able to appreciate that legacy code is holding back innovation today (in 2.x and 3.x), and if this were to be accepted, we're all of a sudden creating a _much_ larger surface for things that may break backwards compatibility. 

Here's an interesting read about `final`: https://matthiasnoback.nl/2018/09/final-classes-by-default-why/ 

I've been making my classes in personal projects for some time now, and that's definitely helping a ton with innovating and improving code over time. I hadn't thought about doing that for `modX` yet, but that definitely makes sense.

sergant210:
@Mark-H Look at it the other way - an alternative way to manage the core. Nobody knows when MODX3 will be released. We can do some things which are not supported the core-team. For example, make xPDO to understand more SQL functions in WHERE clause. And so on.

Mark-H:
@sergant210 I'm 💯% on board with making it possible to add features that are not in the core. That's literally how I earn my paycheck ;)

The *technical* challenges/concerns around this proposal does not mean we can't think of other - better - ways to extend the core beyond its current limits. If you're running into restrictions in xPDO's where clauses being too restricted,  then the question to answer is how we can best address **that**. Perhaps xPDO needs some new hooks for you to do what you need to do. 

You're actually highlighting why inheritance/extending is a pain with that particular example. 

xPDO should be a dependency - a convenient way of securely talking to the database. But it's so deeply coupled that you find yourself trying to _replace the biggest core class_ in order to switch it with something else, a newer version, or just a different configuration. That tells me that adding **more** ways of extending the core class, which is in dire need of being slimmed down, is only going to make that situation worse in the long term.

Mark-H:
Also, I just want to say I really appreciate technical discussions. 

We are at a point where some big steps have been made for 3.x, but there are still lots of things left that different people want to see. Some just want a release yesterday, others want to see more big (or small) changes first. 

Progress is definitely not as fast or steady as _anyone_ would like to see. 

But we do need to talk and explore ideas, like the one in this draft PR. So, before I forget, thank you @sergant210, for the proposal. Even if there are concerns and it may not work out, it's a very worthwhile discussion to have, and I'm really happy we have different voices in the community who care enough to make sure they are heard.

sergant210:
> xPDO should be a dependency - a convenient way of securely talking to the database. 

@Mark-H Ideally Yes. But in the current situation it doesn't work. So I don't understand why  this ability can't be given to developers. Most of us moved to Laravel and only good memories make us to do something for MODX. But no one is interested in our opinion.

JoshuaLuckers:
Everyone’s opinion matters and is taken into consideration. If one of “us” disagrees it doesn’t mean it won’t be accepted.

Mark-H:
I just spent the last two hours discussing this idea with you and Vasily. If I wasn't interested in your opinion, I would have just closed and locked the PR. But because I do care, both about your input and about how we shape MODX going forward, I've tried to explain the technical concerns this proposal comes with, giving you the opportunity to refute those arguments. 

I'm pretty good at changing my opinion when presented with good technical arguments. Explain me something I didn't know. Or show an example of a use case which cannot be solved in a better way.  

Aside from the xPDO example, which I responded to explaining why that actually shows it is a bad idea, you've not addressed the technical concerns. Because you haven't addressed the concerns, you haven't changed my objection. 

To clarify that, **I do care about your opinion and ideas**. That I **don't agree with them** is something completely different. I think it's important to make that distinction, because I do get frustrated when people accuse me personally or the "core team" generally of not listening.

I also don't care who comes up with a certain idea. I don't care about politics or feuds or whatever. I respect every individual that has responded here and everyone brings something worthwhile to the table. If the roles were reversed and Jason had proposed this PR while @sergant210 brought forward the concern that extending `modX` this way is a bad idea, and that perhaps it should be `final`... then I also would've pushed back against Jason because of the technical merit of that concern and the long-term implications.

JoshuaLuckers:
@Mark-H can you share a summary about the technical concerns?

sergant210:
@Mark-H As for you, I have no complaints. 

This PR makes sense for version 2 which is the last working. 

One more example, In some project I have to move the system settings to files.. But MODX loads them from DB in the initialization step. In custom application class I rewrite a couple of methods. And now I must use the custom index.php. But I can do it more correctly if I have this PR.

Maybe for the third version of MODX this PR is not needed, but I think that it can be useful there too.

JoshuaLuckers:
MODX can be initiated in various ways (more details in PR https://github.com/modxcms/revolution/pull/14704).  

This PR will only load `MODX_APP_CLASS` when MODX is initiated the "normal" way but when [loading MODX externally](https://docs.modx.org/current/en/extending-modx/modx-class/loading-externally) it will not initiate the MODX class defined in `MODX_APP_CLASS` thus open up a whole new set of problems and make figuring out what the root problem of an issue/bug is more difficult.

I've seen a few Use Cases mentioned that will be "solved" by this PR but I'm more in favour of finding solutions to make those individual use cases possible instead.

sergant210:
Let's edit the example from the documentaton
```
require_once '/absolute/path/to/modx/config.core.php'; // 
require_once MODX_CORE_PATH.'model/modx/modx.class.php';
#1. Remove the line below
// $modx = new modX();
#2. Add a new instantiation
$modx = modX::getInstance(MODX_APP_CLASS);
$modx->initialize('web');
$modx->getService('error','error.modError', '', '');
```
What's the problem?

P.S.
Or you talk about BC?

JoshuaLuckers:
Why not instantiate the subclass directly in this case?  

I still don't see how any of this is better than manually modifying the `index.php` and instantiate the "enhanced" modX class directly?

sergant210:
@JoshuaLuckers 
1. Risk of losing changes when updating MODX.
2. No ability to get the $modx object like this
```php
public function someMethodOrConstructor() 
{
   $this->modx = modX::getInstance();
}
```

I think we need to stop the discussion. I will close this PR as meaningless. Спасибо!

JoshuaLuckers:
I try to uncover all the advantages and disadvantages so we can make a well thought out decision!

PR #14576, opened 2019-04-30
Added line-through style to unpublished resources

Ruslan-Aleev:
### What does it do?
Changed the style for unpublished resources in the tree to line-through.

**The current version is not informative at all**, we have already discussed this for version 3. And I constantly get confused with the style of states  of resources in the menu, except deleted ones :)
Unpublished resource difficult to find if the tree is large, especially if the resource is not shown in the menu and is not published (font-style: italic does not help much).
Still confusing is that the hidden from the menu, but the published resource looks paler than the unpublished resource, but not hidden from the menu - strange logic.

The line-through style is immediately noticeable and again similar to the deleted of a resource, and both of these states (publication and deletion) give 404, i.e. and visually the logic will be similar.

What it looks like

menu_tree
menu_tree_2

I decided to make PR for version 2, because this will be useful for 2.

### DIFF
--- a/_build/templates/default/sass/_colors-and-vars.scss
+++ b/_build/templates/default/sass/_colors-and-vars.scss
@@ -123,7 +123,7 @@ $deactivatedTextColor: lighten($treeColor, 35%); // do not use 50% here - makes 
 $disabledTextColor: $deactivatedTextColor;
 
 $hidden: lighten($treeText, 10%) !important;
-$unpubText: italic;
+$unpubTextDeco: line-through;
 // locked is already signaled with lock-icon, no need to use other semantic style here
 // as italic is used for "unpublished"
 $lockedText: inherit; // italic;
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -301,7 +301,7 @@
 .unpublished,
 .unpublished a span {
   color: $unpublished;
-  font-style: $unpubText;
+  text-decoration: $unpubTextDeco;
 
   i.icon,
   i.icon-large {
@@ -315,7 +315,7 @@
 
   &.unpublished,
   &.unpublished a span {
-    font-style: $unpubText;
+    text-decoration: $unpubTextDeco;
   }
 
   i.icon,


opengeek:
I'm not in favor of this change. Strike-through just doesn't say "this resource is not yet published" or "this resource is temporarily unpublished" very well. Just my $0.02.

Ruslan-Aleev:
This is subjective :) But in the current version of the style the publication of the resource is not clear at all. Even if you want to understand the styles in the tree - it will be forgotten, because there is no memorable visual logic.

Current styles:
menu_3

And the point, it seems to me, is to attract attention, because publication is the most important parameter, in addition to deletion, and this should be noticeable. In the current version it is not noticeable, especially if the resource tree is large.

ghost:
I don't see the problem; even with my bad eyes I've never had a problem seeing the unpublished resources.

Ruslan-Aleev:
They are visible, but difficult to find if the tree is large.
At least I often come across this in my work, especially if the resource is not shown in the menu and is not published (font-style: italic does not help much). Still confusing is that the hidden from the menu, but the published resource looks paler than the unpublished resource, but not hidden from the menu - strange logic.

Perhaps for others there is no such problem, but I (and clients) constantly get confused and cannot immediately find unpublished resources.
Different variants were discussed here - https://github.com/modxcms/revolution/issues/13969

JoshuaLuckers:
I agree with @Ruslan-Aleev, the current visual indicator for unpublished resources is to difficult to differentiate from the other visual indicators. Maybe we can simply mark them in red without a strike trough? Or in orange?

Ruslan-Aleev:
Yes, also an interesting option. The only thing @Mark-H said was that the problem could be for users with color blindness. But, as it seems to me, the current version, and for users without color blindness creates difficulties.
Perhaps it is worth combining the version with font-style: italic and yellow/orange color?

Ibochkarev:
It will be possible to choose the appropriate color.

image

Ruslan-Aleev:
Here is the variant with orange color (or make a little paler), in my opinion, does not look bad.

color

rthrash:
For consistency of long standing behavior, the unpublished docs should remain italicized. I’m ok with adding a color, but it needs to have enough contrast to pass accessibility checks. 

The presentation of the last screenshot would likely confuse the vast majority of existing users. 

Perhaps we add s small red/green dot (with the right shades for sufferers of pantopia like me) to indicate published status, and black/gray/strikethrough for menu/deleted status?

Ruslan-Aleev:
@rthrash Here we need to remember that the resource can change the icon (template icon), and perhaps the dot will not be noticeable.
Different variants were discussed here - https://github.com/modxcms/revolution/issues/13969

The style can be any, personally the line-through style seems to me optimal, and here's why:
1) Unpublished resource with line-through is immediately visible in the tree and you do not need to add an additional icon or dot. This style is visually closer to the current style than the additional icon or dot.
2) Unpublished resource is in fact closer to deleted (than, for example, hidden from the menu) - both give 404.
3) The variant will work for all types of resources and template icons.
4) There is no need to simplify the state of the resource (for example, unpublished, but visible in the menu), because line-through visually overlaps these styles.

rthrash:
I don’t view unpublished resources as closer to deleted in most cases. They are often works in progress or frequently visited test pages. 

That said, I’m sure there is a solution to this issue that can be found.

Ruslan-Aleev:
Yes, about the pages in the work, there are no questions here, because they will be soon
published.

I’m confused also by the fact that a client can see an unpublished page on a site if he is logged in the manager panel, although in fact the page gives 404 and in the resource tree this is not indicated (font-style: italic is not very noticeable, especially if the page is not shown on the menu). And 404 is serious for the site and it needs to be made as noticeable as possible.

Let's wait for other variants :)

rthrash:
The reason the client can see unpublished pages is so they can actually see how they will look before publishing … very much expected behavior. I've personally never had an issue with italics fonts indicating unpublished, and changing this will likely be poorly received and disruptive for the majority of existing MODX users. 

I suspect a plugin could be written to alter the styles pretty easily for the Manager.

JoshuaLuckers:
> Perhaps we add s small red/green dot (with the right shades for sufferers of pantopia like me) to indicate published status, and black/gray/strikethrough for menu/deleted status?

Adding a coloured dot as one of the status indicators is solution I can be happy with. This behaviour is widely used in for example pages to indicate the status of a service.
Examples:
- [G Suite Status Dashboard](https://www.google.com/appsstatus#hl=en&v=status)
- [Apple System Status](https://www.apple.com/support/systemstatus)

We can keep the current behaviour and add this as additional and optional indicator.

Ruslan-Aleev:
@rthrash It is clear that these styles are not a problem for everyone.
But I would be happy to see the differences in the status of resources more visible.

rthrash:
@Ruslan-Aleev, while the deleted and unpublished states may be too low contrast for low-vision users, and the sizes/fonts/colors, in general, may be a bit off due to quick PS-mockup, would the following help you distinguish the published status better?

image

Ruslan-Aleev:
@rthrash Thank you, also a good variant :)
Add a "title" attribute with a description to understand the dots and that's it.
I will leave this PR with a discussion, but I think it’s worth adding dots to MODX3 or what?
Or will I copy the important comments into the https://github.com/modxcms/revolution/issues/13969 and close this PR?

alroniks:
Colored dots are much better than line-through text. So I vote for closing this PR in favor of proper implementation.

JoshuaLuckers:
@Ruslan-Aleev Yes, the PR for adding the status dots should target 3.x.

PR #14577, opened 2019-05-02
Generate better tag for modElement

grachov:
### What does it do?
This PR updates `modElement::getTag` to generate better tags in some cases as described below.

### Why is it needed?
When using non-default template engines in MODX (for example `Fenom` implementation by @bezumkin) we can pass arrays as properties. When tag is generated for element in such case, those arrays will be replaced with random strings (as arrays are not scalar values). This means that element output can not be taken from elements cache as that random string will not match on next render.
Also we can sort properties to be able to get cached output when similar element calls have different params order.

### DIFF
--- a/core/model/modx/modelement.class.php
+++ b/core/model/modx/modelement.class.php
@@ -224,17 +224,18 @@ class modElement extends modAccessibleSimpleObject {
      */
     public function getTag() {
         if (empty($this->_tag)) {
-            $propTemp = array();
             if (empty($this->_propertyString) && !empty($this->_properties)) {
+                $propTemp = array();
                 foreach ($this->_properties as $key => $value) {
+                    $key = trim($key);
                     if (is_scalar($value)) {
-                        $propTemp[] = trim($key) . '=`' . $value . '`';
-                    }
-                    else {
-                        $propTemp[] = trim($key) . '=`' . md5(uniqid(rand(), true)) . '`';
+                        $propTemp[$key] = $key . '=`' . $value . '`';
+                    } else {
+                        $propTemp[$key] = $key . '=`' . (is_array($value) ? json_encode($value) : md5(uniqid(rand(), true))) . '`';
                     }
                 }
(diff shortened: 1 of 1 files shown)


Mark-H:
This sure looks a lot more useful than a random value. Would be nice to have it target 3.x and include a test or two though? :-)

JoshuaLuckers:
@grachov is it possible to target 3.x and include some unit tests? If not, please let us know, thanks!

core/model/modx/modelement.class.php
```diff
@@ -224,17 +224,18 @@ protected function isStaticFilesAutomated()
...
                 foreach ($this->_properties as $key => $value) {
+                    $key = trim($key);
                     if (is_scalar($value)) {
-                        $propTemp[] = trim($key) . '=`' . $value . '`';
-                    }
-                    else {
-                        $propTemp[] = trim($key) . '=`' . md5(uniqid(rand(), true)) . '`';
+                        $propTemp[$key] = $key . '=`' . $value . '`';
+                    } else {
+                        $propTemp[$key] = $key . '=`' . (is_array($value) ? json_encode($value) : md5(uniqid(rand(), true))) . '`';
```
opengeek, line 234: I think it would be better to use `md5(serialize($value))` here as there is no reason to risk json_encode failures (for instance, if encountering a non-UTF8 character). It would also be much more compact in the cache as a hashed string.

opengeek (changes requested):
If the author does not respond, I will apply this to 3.x and make the requested changes I have suggested in my review.

core/model/modx/modelement.class.php
grachov: Yes, you are right, serialize is better. I wanted to switch to it, but was busy a little. I’ll make the changes to my PR.

opengeek:
I went ahead and cherry-picked this commit on top of 3.x and made the changes I suggested in #14689.

PR #14420, opened 2019-02-22
Change layout on profile pages

GulomovCreative:
### What does it do?
1. Changes layout on user edit page
2. Changes layout on profile edit page (General Information Tab)
3. Changes layout on profile edit page (Reset Password Tab)
4. Outputs missing fields on profile edit page: Address, City, Website and Gender (but I could not output 
 country field, some problem with modx-combo-country. If you know the solution, send it)
5. Adds username to header on profile edit page

**UPD:**
1. Logical resort of fields
2. Adds username to title
3. Adds changing header during keyup username field

Before

before

**After**

after

Before

before2

**After**

after2

Before

before3

**After**

after3

Before

before4

**After**

after4

### Why is it needed?
At least saving space

### Related issue(s)/PR(s)
#13988

PR #14578, opened 2019-05-03
Improving profile edit screen

sdrenth:
### What does it do?
Refactored the edit account page: 
domain.tld/manager/?a=security/profile

- Moved save buttons from the bottom of the page to the top, to improve consistency
- Added fields to edit account page: Gender, address, country, city, zipcode and website.
- Moved change password tab to the General information tab (permissions check is still in place)
- Every field is now 100% wide, but seperated the view into two columns. Personal details are shown on the left side, address + password are shown on the right side.
- Removed changepassword processor since its not being used anymore. I've moved the logic to the /profile/update processor.

### Why is it needed?
This fixes some missing fields, improves consistency and makes the view more clean.

Preview of the refactored view:
Screenshot 2019-05-03 at 13 08 58


Ibochkarev:
Relayted https://github.com/modxcms/revolution/pull/14420

sdrenth:
As discussed with @GulomovCreative I will create a PR to his branch so we can combine these to PR's into one single PR.

Issue #14575, opened 2019-04-27
Allow template variables input type date configurations to be smarty


JANogueira:
### Summary
According to the documentation below:
https://docs.modx.com/revolution/2.x/making-sites-with-modx/customizing-content/template-variables/template-variable-input-types#TemplateVariableInputTypes-Date

We are able to use date TV input type configuration json that can be applied on migx and other extras or customizations.
When using JSON you are not able to allow dynamic `minDateValue`, `maxDateValue`, `minTimeValue` and `maxTimeValue` values due to the `/manager/templates/default/element/tv/renders/input/date.tpl` file that converts JSON string to javascript directly without parsing any possible code inside each JSON config value.

My suggestion is to change `minDateValue`, `maxDateValue`, `minTimeValue` and `maxTimeValue` details on the tpl file to allow smarty parting

### Why is it needed?
This will be important to limit user input when inserting, for instance, ID Card Issue date (you will not have an ID card issued in the future). Any other similar use cases will apply on this idea as well (like purchase dates, etc)

### Suggested solution(s)
Change the file `/manager/templates/default/element/tv/renders/input/date.tpl` to have the following changes:
```
{if $params.minDateValue|default},minDateValue: '{eval var=$params.minDateValue|default}'{/if}
{if $params.maxDateValue|default},maxDateValue: '{eval var=$params.maxDateValue|default}'{/if}

{if $params.minTimeValue|default},minTimeValue: '{eval var=$params.minTimeValue|default}'{/if}
{if $params.maxTimeValue|default},maxTimeValue: '{eval var=$params.maxTimeValue|default}'{/if}

```

By Adding eval, we will be able to use these fields with dynamic dates. Here is an example:
```
          {
            "allowBlank":true,
            "disabledDates":"0000-00-00",
            "disabledDays":"",
            "minDateValue":"",
            "minTimeValue":"",
            "maxDateValue":"{$smarty.now|date_format:'%Y-%m-%d'}",
            "maxTimeValue":"",
            "startDay":"",
            "timeIncrement":"",
            "hideTime":true
          }
```

I believe this can be useful for a lot of people and not only for input type date TVs specially if dealing with CMP.

Do you think this makes sense?

Ibochkarev:
@JANogueira You can prepare PR?

JANogueira:
Hi @Ibochkarev 

Never did a PR before on github.
I will investigate a bit to check how I can do this.

I am still a SVN user that didn't move to git yet.

[...]

PR #14579, opened 2019-05-05
date.tpl input tv: Date and time configurations with nested smarty

JANogueira:
According to the documentation below:
https://docs.modx.com/revolution/2.x/making-sites-with-modx/customizing-content/template-variables/template-variable-input-types#TemplateVariableInputTypes-Date

We are able to use date TV input type configuration json that can be applied on migx and other extras or customizations.
When using JSON you are not able to allow dynamic minDateValue, maxDateValue, minTimeValue and maxTimeValue values due to the /manager/templates/default/element/tv/renders/input/date.tpl file that converts JSON string to javascript directly without parsing any possible code inside each JSON config value.

### What does it do?
This change will allow `minDateValue`, `maxDateValue`, `minTimeValue` and `maxTimeValue` config parameters to accept nested smarty.

### Why is it needed?
This will be important allow user input limit when inserting, for instance, ID Card Issue date (you will not have an ID card issued in the future). Any other similar use cases will apply on this idea as well (like purchase dates, etc)

Final result using the config parameters with Migx (as an example) will be like this:
By Adding eval, we will be able to use these fields with dynamic dates. Here is an example:
```
          {
            "allowBlank":true,
            "disabledDates":"0000-00-00",
            "disabledDays":"",
            "minDateValue":"",
            "minTimeValue":"",
            "maxDateValue":"{$smarty.now|date_format:'%Y-%m-%d'}",
            "maxTimeValue":"",
            "startDay":"",
            "timeIncrement":"",
            "hideTime":true
          }
```

This pull request follows the improvement suggestion posted on issue #14575

### DIFF
--- a/manager/templates/default/element/tv/renders/input/date.tpl
+++ b/manager/templates/default/element/tv/renders/input/date.tpl
@@ -14,12 +14,12 @@ Ext.onReady(function() {
         ,dateFormat: MODx.config.manager_date_format
         ,timeFormat: MODx.config.manager_time_format
         {if $params.disabledDays|default},disabledDays: {$params.disabledDays|default}{/if}
-        {if $params.minDateValue|default},minDateValue: '{$params.minDateValue|default}'{/if}
-        {if $params.maxDateValue|default},maxDateValue: '{$params.maxDateValue|default}'{/if}
+        {if $params.minDateValue|default},minDateValue: '{eval var=$params.minDateValue|default}'{/if}
+        {if $params.maxDateValue|default},maxDateValue: '{eval var=$params.maxDateValue|default}'{/if}
         {if $params.startDay|default},startDay: {$params.startDay|default}{/if}
 
-        {if $params.minTimeValue|default},minTimeValue: '{$params.minTimeValue|default}'{/if}
-        {if $params.maxTimeValue|default},maxTimeValue: '{$params.maxTimeValue|default}'{/if}
+        {if $params.minTimeValue|default},minTimeValue: '{eval var=$params.minTimeValue|default}'{/if}
+        {if $params.maxTimeValue|default},maxTimeValue: '{eval var=$params.maxTimeValue|default}'{/if}
         {if $params.timeIncrement|default},timeIncrement: {$params.timeIncrement|default}{/if}
         {if $params.hideTime|default},hideTime: {$params.hideTime|default}{/if}
 


JoshuaLuckers:
Thanks for taking time to submit this PR and help make MODX better. Since you are a first-time contributor; please sign the the [Contributor License Agreement (CLA)](https://modx.com/community/contribute/cla/).

JANogueira:
> Thanks for taking time to submit this PR and help make MODX better. Since you are a first-time contributor; please sign the the [Contributor License Agreement (CLA)](https://modx.com/community/contribute/cla/).

Hi @JoshuaLuckers 
Done. Thanks for the tip as I was not aware of that detail.

JoshuaLuckers:
I'm wondering if there might be any security implications if we use `eval`.

JANogueira:
Hi @JoshuaLuckers ,

Should not be a problem:
* Evaluated variables are treated the same as templates. They follow the same escapement and security features just as if they were templates.

Source:
https://www.smarty.net/docs/en/language.function.eval.tpl

JoshuaLuckers (approved):

Mark-H:
Evaluating inline templates for a dynamic max date _really_ looks like the wrong hammer. I'm sure there's a much better way to approach that; the default value already accepts the `strtotime` logic - why not make the same work for the min/max dates?

JANogueira:
Hi Mark,

The problem here is that other extras like Migx do a very strict usage of these attributes. As an example, on Migx, I have no other way than using inline smarty to get current date. Otherwise JSON will be broken.

This can also happen on other extras that might use these attributes on the same way.

What would be your suggestion for this?

Mark-H:
I'd much rather see `maxDateValue` accept a value of `today` (or `tomorrow`, `+3 days`, etc) than this templating stuff. That isn't about MIGX, but about how the TV is configured. MIGX just replicates that configuration in its own way.

 I've not recently looked at the core TVs, but I think that type of logic would typically go into the `render()` or `process()` method of the TV class.

JANogueira:
I agree. Your suggestion will greatly simplify the usage and the logic.
Please feel free to drop this PR.

Bruno17:
@JANogueira as response to your MIGX use-case, with this commit, the field-config in MIGX MODX-tags are parsed now, So you can build dynamic configs with a snippet. https://github.com/Bruno17/MIGX/commit/e462de0fc27d009f02457ec3f544c4c3d40f1f8c

JANogueira:
Many thanks @Bruno17! Excellent news!

JoshuaLuckers:
Thanks for taking time to submit this PR and help make MODX better. I'm closing this PR because a better solution for your use case is implemented in MIGX.

Issue #14176, opened 2018-12-03
Make it clear that the value can be changed in MODX tables [ux/ui]


Ruslan-Aleev:
In MODX tables, you can change the value by double-clicking, but this is not obvious to the user.

Need add a class for fields that can be changed by double-clicking and changing the cursor to "pointer" in this class.

![edit_cursor](https://user-images.githubusercontent.com/12523676/49375080-ba354b80-f71c-11e8-9fea-fec72f89a442.png)

Or besides changing the cursor add a tip using `<div ext:qtip="[[++settings_key]]<br>Double-click to edit">` for such fields (should also specify the [[++settings_key]] for setting fields, as well as in the resource fields).

![edit_3_cursor](https://user-images.githubusercontent.com/12523676/49511167-cef01b80-f8a3-11e8-9ee8-aa6776d5ea6a.png)

Together with https://github.com/modxcms/revolution/issues/14125, the user-friendly MODX will improve significantly.

Ruslan-Aleev:
@Jako Can you help with advice or information?
If in the `/manager/assets/modext/widgets/core/modx.grid.settings.js` (script to build system settings) file can specify css at column "Value":
...
```
{
header: _('value')
,dataIndex: 'value'
,sortable: true
,editable: true
,renderer: this.renderDynField.createDelegate(this,[this],true)
,width: 260
,css: 'cursor: pointer; font-style: italic;'
}
```
...
But I did not understand how to add a class or tooltip.
Is it even possible in the current version of the script?

Here's how to look with css:

![settings_grid](https://user-images.githubusercontent.com/12523676/51902120-b1868c00-23d2-11e9-90cf-d079ac62d10d.gif)

Jako:
The tooltip (or any other change of the column) could be created in the renderer property. The method in there has to return the display value. If I remember right, the method could return html tags, too.

Mark-H:
Fixed in #14748

PR #14580, opened 2019-05-08
Added tooltip for editable fields in the "System Settings"

Ruslan-Aleev:
### What does it do?
Added tooltip for editable fields in the "System Settings". Thanks to @GulomovCreative and @sergant210.

**What it looks like:**
tooltip_1

In MODX tables, you can change the value by double-clicking, but this is not obvious to the user.

_Unfortunately, this PR adds this functionality ONLY to the "System Settings", we need to extend this behavior for all editable fields in MODX tables. But I do not have enough knowledge, need help._

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14176


core/lexicon/en/default.inc.php
```diff
@@ -130,6 +130,7 @@
 $_lang['edit_snippet'] = 'Edit Snippet';
 $_lang['edit_template'] = 'Edit Template';
 $_lang['edit_tv'] = 'Edit TV';
+$_lang['editable_field'] = 'Double-click for quick editing.';
```
JoshuaLuckers:
```suggestion
$_lang['editable_field'] = 'Double-click to open edit mode.';
```

I believe this is a more accurate description for the "double-click" action. Sometimes you are able to edit the value directly, sometimes it shows a select box you have to click if you want to change the value.

Mark-H (approved):
Please make sure to mark PRs as ready for review when they're done - I'm going ahead and approving it, but it wont be able to get merged until you mark it as ready.

Ruslan-Aleev:
@Mark-H There is also a question. Can this behavior be somehow extended for all grids via modx.grid.js?

Mark-H:
@Ruslan-Aleev I'm not entirely sure - if so it would likely need to go into `_loadColumnModel()` as that prepares renderers and editors, but I din't immediately see how/where you'd need to add the class to it there.

Ruslan-Aleev:
@Mark-H Clear. We can insert the same code separately in each script with a grid, but this is inconvenient :)

Mark-H:
I've played around with this and almost have a new PR ready that consistently applies it to any grid 🎉 

I'm thinking the bold styling is a little too much though. Maybe you @Ruslan-Aleev or someone else has a better idea on how to style it so it stands out, but isn't in your face _too much_? Maybe applying the link color on hover, or only doing the tooltip?

Ruslan-Aleev:
Good news, thanks! We will look forward to PR :)

As for styles, in addition to tooltips, I think they are still needed to identify property in line (although not so much needed,).
We can, as you rightly pointed out, change the color, or, another option, reduce the transparency (opacity: .7;)
Additional styles will be useful for mobile devices, it seems to me.

Mark-H:
PR created: #14748

Issue #14125, opened 2018-10-29
Show a link to edit an item in its name in the MODX tables


Ruslan-Aleev:
In MODX many settings are edited via the context menu, but this is not obvious, especially if you are not familiar with MODX.
For example in "Context" settings: clicked on the row - nothing happened, clicked 2 times on row - nothing happens either.

- [x] I think it is _worth duplicating the link from the context menu on the item name_, clicked on the link and the settings opened.

**How I see it:**
![link_2](https://user-images.githubusercontent.com/12523676/47645390-50101080-db8a-11e8-8570-5031e1a4376c.png)

I think links should be added to all such tables, this is more understandable for the user than the context menu.
p.s. By the way, in the "Manage -> Users" section the links are already made, but only it is not visible until you hover on it.

**Also a good variant with icons suggested by @Jako:**

- [x] We should also add a icon-button column to all grids in the manager. The 'real' links of some manager pages should be inserted there too (to make those buttons be clickable with a command-click/controll-click while opening the link in a new tab).

Done with https://github.com/modxcms/revolution/pull/14806

![icons_2](https://user-images.githubusercontent.com/148371/47647571-a9bb0f80-db77-11e8-968c-47eb2219bd2f.png)

Jako:
We should better add a button column like in [SwitchTemplate](https://github.com/Jako/SwitchTemplate/blob/master/source/js/mgr/widgets/switchtemplate.grid.js#L204-L238) to all grids in the manager. The 'real' links of some manager pages should be inserted there too (to make those buttons be clickable with a command-click/controll-click while opening the link in a new tab).

Screenshot:
<img width="1056" alt="bildschirmfoto 2018-10-29 um 12 30 15" src="https://user-images.githubusercontent.com/148371/47647571-a9bb0f80-db77-11e8-968c-47eb2219bd2f.png">

Ruslan-Aleev:
@Jako Yes, the icons are also useful, but the links on the names of the lines are more noticeable and understandable. Ideal variant - make and links on names and icons.

Jako:
Some grids could be edited inline and the links can't be used there, so we would have different UX for inline grids, which is a bad habit.

Ruslan-Aleev:
@Jako In the "System settings", for example, inline editing, a window pops up by clicking on the link.
The goal is not to change the functional, but to copy the link from the context menu to the name, the functional already exists.
Or are you talking about something else, you can specify an example of inline editing?

[...]

PR #14581, opened 2019-05-10
Displayed a link to the context in the grid

Ruslan-Aleev:
### What does it do?
In MODX many settings are edited via the context menu, but this is not obvious, especially if you are not familiar with MODX.
I think it is worth add the link on the item name, clicked on the link and the page with settings was opened.

What it looks like:

contexts-links

icons_grid

**Unfortunately, there are 1 unresolved questions:**
- How to display a link to an element in the grid if there is no update controller, for example, in the "System Settings"?

Need help.

In general, the task is not easy, because there is no logic in the name of the scripts, somewhere .grid.js, somewhere .panel.js, and in both scripts tables can be generated.
Some tables are initially hidden in nested other tables, for example, in the section "Form Customization".

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14125

### DIFF
--- a/manager/assets/modext/widgets/system/modx.grid.context.js
+++ b/manager/assets/modext/widgets/system/modx.grid.context.js
@@ -32,6 +32,9 @@ MODx.grid.Context = function(config) {
             ,width: 150
             ,sortable: true
             ,editor: { xtype: 'textfield' }
+            ,renderer: function(value, p, record){
+                return String.format('<a href="?a=context/update&key={0}" title="{1}" class="x-grid-link">{2}</a>', record.data.key, _('context_update'), Ext.util.Format.htmlEncode( value ) );
+            }
         },{
             header: _('description')
             ,dataIndex: 'description'
--- a/manager/assets/modext/widgets/system/modx.grid.context.js
+++ b/manager/assets/modext/widgets/system/modx.grid.context.js
@@ -8,17 +8,6 @@
  */
 MODx.grid.Context = function(config) {
     config = config || {};
-    this.buttonColumnTpl = new Ext.XTemplate('<tpl for=".">'
-        + '<tpl if="action_buttons !== null">'
-        + '<ul class="x-grid-buttons">'
-        + '<tpl for="action_buttons">'
-        + '<li><i class="icon {className} icon-{icon}" title="{text}"></i></li>'
-        + '</tpl>'
-        + '</ul>'
-        + '</tpl>'
-        + '</tpl>', {
-        compiled: true
-    });
     Ext.applyIf(config,{
         title: _('contexts')
         ,id: 'modx-grid-context'
@@ -162,6 +151,7 @@ Ext.extend(MODx.grid.Context,MODx.grid.Grid,{
         //this.refresh();
         return true;
     }
+
     ,clearFilter: function() {
         this.getStore().baseParams = {
             action: 'context/getList'
@@ -224,24 +214,31 @@ Ext.extend(MODx.grid.Context,MODx.grid.Grid,{
         this.refresh();
     }
 
-    ,buttonColumnRenderer: function () {
+    ,buttonColumnRenderer: function (a,id,row,ri) {
+        var p = row.data.perm;
         var b = [];
-        b.push({
-            className: 'update',
-            icon: 'pencil-square-o',
-            text: _('context_update')
-        });
-        b.push({
-            className: 'duplicate',
-            icon: 'files-o',
-            text: _('context_duplicate')
-        });
-        b.push({
-            className: 'remove',
-            icon: 'trash-o',
-            text: _('context_remove')
-        });
-        return this.buttonColumnTpl.apply({
+        if (p.indexOf('pedit') != -1) {
+            b.push({
+                className: 'update',
+                icon: 'pencil-square-o',
+                text: _('context_update')
+            });
+        }
+        if (p.indexOf('pnew') != -1) {
+            b.push({
+                className: 'duplicate',
+                icon: 'files-o',
+                text: _('context_duplicate')
+            });
+        }
+        if (p.indexOf('premove') != -1) {
+            b.push({
+                className: 'remove',
+                icon: 'trash-o',
+                text: _('context_remove')
+            });
+        }
+        return this.getButtonColumnTpl().apply({
             action_buttons: b
         });
     },
--- a/manager/assets/modext/widgets/system/modx.grid.context.js
+++ b/manager/assets/modext/widgets/system/modx.grid.context.js
@@ -49,17 +49,17 @@ MODx.grid.Context = function(config) {
         },{
             header: _('description')
             ,dataIndex: 'description'
-            ,width: 600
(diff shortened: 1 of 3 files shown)


Ruslan-Aleev:
Added action icons in the grid. Need test. Then it makes sense to add such icons for other manager grids.
Strongly useful for UI / UX.

icons_grid

Ruslan-Aleev:
_Somehow, we need to consider permissions for actions icons (change styles for inactive icons)_, because somewhere there will be a delete action, and somewhere not (in PR they are, but a click pops up a message indicating lack of access). In general, the question is not simple, but necessary.

**As I understand it, there must be something like this, in future:**

icons_grid

manager/assets/modext/widgets/system/modx.grid.context.js
```diff
@@ -8,6 +8,17 @@
  */
 MODx.grid.Context = function(config) {
     config = config || {};
+    this.buttonColumnTpl = new Ext.XTemplate('<tpl for=".">'
+        + '<tpl if="action_buttons !== null">'
+        + '<ul class="x-grid-buttons">'
+        + '<tpl for="action_buttons">'
+        + '<li><i class="icon {className} icon-{icon}" title="{text}"></i></li>'
```
Mark-H: The variables should be escaped to prevent XSS, it's been a while but I think the syntax in extjs was `{className|htmlEncode}` or something similar?

manager/assets/modext/widgets/system/modx.grid.context.js
```diff
@@ -8,6 +8,17 @@
  */
 MODx.grid.Context = function(config) {
     config = config || {};
+    this.buttonColumnTpl = new Ext.XTemplate('<tpl for=".">'
```
Mark-H: If we're going to apply this style of buttons to other grids (which for consistency would be a good idea), this property should be moved to the `MODx.grid.Grid` object to make it easily available. And added to the documentation. :)

Mark-H:
Hey @Ruslan-Aleev, is this ready for review or are you still working through it?

Ruslan-Aleev:
To be honest, I could not display the icon (gear), which on a click will show a drop-down menu (not enough knowledge).
And I didn’t understand what we finally decided? We discussed in Slack, but there was no concrete solution.

Mark-H:
Fair enough. I'd be happy to accept this PR with some minor changes and to look at the gear icon for a later PR.

Mark-H:
So does that thumbs up mean you're ready to mark this for review @Ruslan-Aleev? ;)

Ruslan-Aleev:
I did not understand the outcome of the discussion:
1) We leave this option with 3 icons (edit, copy and delete) ONLY for the "Contexts" section? And then we need to do the same PR for other grids?
Or do we need to somehow expand these icons in MODx.grid.Grid and this functionality will be available easier for all grids? (I am for this option).
2) Do we really need protection against XSS here? I set class names in the code, they are not taken from the outside.

Mark-H:
- `buttonColumnTpl` should be moved up to `MODx.grid.Grid` - that way only the `buttonColumnRenderer` has to be implemented in different grids. Each grid would need to be updated manually anyway, but we can at least share that template to simplify it and make it easier to update the markup in the future, if needed. 
- Yes XSS protection is necessary. Skipping it for the classes is fine, but it definitely needs to be added to the text.

Ruslan-Aleev:
Yes, I also forgot that we need to somehow track access to the action in icon.
For example, for the web and mgr contexts, “delete” action in the drop-down menu will not appear, however the icon will be and if you click on it, there will be an access error message. Although it is more correct to somehow remove the action from the icon and change its class to inactive (for example adding transparency to the icon).

Ruslan-Aleev:
In general, while this PR is far from merging and I do not know how to make it work.
The other day I will consult with the guys how to bring it to working condition :)

Ruslan-Aleev:
Thanks to @GulomovCreative for help with this PR!
Now PR can be tested :)

manager/assets/modext/widgets/system/modx.grid.context.js
```diff
@@ -142,68 +151,119 @@ Ext.extend(MODx.grid.Context,MODx.grid.Grid,{
...
+                className: 'remove',
+                icon: 'trash-o',
+                text: _('context_remove')
+            });
+        }
+        return this.getButtonColumnTpl().apply({
+            action_buttons: b
+        });
+    },
+    onClick: function (e) {
```
Mark-H, line 245: I've not tested it yet, but I wonder, could we get rid of the onClick here if each of the action buttons is assigned a `handler` instead? 

Maybe not because they're not extjs-type buttons.. but worth a check to make the per-grid implementation easier perhaps?

Ruslan-Aleev:
@Mark-H a naive question: is it possible to somehow extend the **getMenu** function https://github.com/modxcms/revolution/blob/9694b60434f1e8b50520b378cc9de661ab5fe342/manager/assets/modext/widgets/system/modx.grid.context.js#L98 for icon buttons? Here almost all the functionality is already registered, maybe it will be easier?

Mark-H:
A getMenu-like analogue for the action buttons is indeed what I'd like to see (and why I asked about being able of using handlers). Just defining the options in the grid, and the rest being handled transparently in `MODx.grid.Grid` would be ideal. 

I don't think it should re-use the `getMenu` method; that goes back to the discussion we had before about too many buttons... the context menu could offer more options than the buttons in such cases.

manager/assets/modext/widgets/system/modx.grid.context.js
Ruslan-Aleev: In my opinion the `handler` does not work in this case, but I'm not sure.
By the way, for packages, the code works similarly - https://github.com/modxcms/revolution/blob/3.x/manager/assets/modext/workspace/package.grid.js#L234

Jako: It should be more effective the current way, since each button would need a click handler (or listener) then. At the moment the whole button column gets one click handler assigned.

theboxer:
As Mark pointed out there might (most likely will) be a problem with too many icons in the grid (especially for these that also have a lot of columns) - there's not enough width to fit everything there.

For a Fred, we did a single icon that opens a context menu, so you can either right click the row or left click the gear icon:
gear icons

context menu

Could be leveraged to show this icon by default and have an option on grid to add more icons where it'll make sense.

Ruslan-Aleev:
@theboxer Personally, it seems to me that the actions "Edit" (possibly "Copy" or "Delete") should always be displayed as the most used (to reduce clicks).
And hide the rest in a separate menu, i.e. we will always have a maximum of 2 icons displayed (3 with "Copy" or "Delete"):
icons_grid

theboxer:
Sorry, I missed the gear icon in your screenshot.

Ruslan-Aleev:
There is no gear in the current PR, I could not make it.
The gear may appear in a separate PR, Mark and I discussed this above.

Issue #14143, opened 2018-11-20
Incorrect behavior of the "Uses Template" list in the resource.


Ruslan-Aleev:
The "Uses Template" list in resource will open only when **clicking on the button** with arrows, although, for example, in the "Settings"-tab, lists ("Resource Type", "Content Type", etc.) are opened by **clicking on any place** in the list row.

![list_error](https://user-images.githubusercontent.com/12523676/48767819-015d1e80-ecd1-11e8-9c9a-9e4bda58bbb8.png)

MODX 2.6.5

alroniks:
It is an editable field, so after click field becomes in focus for entering values as a usual text field. Other similar fields allow only select existing variants.
So I would say that it is not a bug.

Ruslan-Aleev:
@Alroniks In the list of templates there are only existing variants, or not? And catch the button is less conveniently.
This behavior appeared not so long ago, right?

alroniks:
@Ruslan-Aleev Nope, last changes from 2016, and they unrelated. So it exists for a long time.

Ruslan-Aleev:
@Alroniks :) Very strange, noticed it quite recently and constantly clicked on the entire list instead of the button, apparently, did not pay attention.

JoshuaLuckers:
The template list being an editable field does yield some odd behaviour. If I type in a name of a non-existing template and save the resource it automatically defaults to the base template.

Ruslan-Aleev:
By the way, in the form of quickly creating / editing a resource, the template selection list works by clicking the full width of the list.

Ruslan-Aleev:
@Jako @opengeek @rthrash Any thoughts on this? This is not a bug, but I do not know if this functionality is needed here?

rthrash:
It makes sense to treat it as a traditional select list. I consider it a poor UX.

PR #14586, opened 2019-05-15
Made template selection list as traditional select list in Resource editing form

Ruslan-Aleev:
### What does it do?
The template selection is an editable field, but to select a template such functionality is not needed, and to expand the entire list of templates, you need to click on the button on the right, which is not convenient.

Now the **list works across the entire width**, a bit more convenient for the UX.

template

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14143

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -617,10 +617,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,name: 'template'
             ,id: 'modx-resource-template'
             ,anchor: '100%'
-            ,editable: true
-            ,typeAhead: true
-            ,typeAheadDelay: 300
-            ,forceSelection: true
+            ,editable: false
             ,baseParams: {
                 action: 'element/template/getList'
                 ,combo: '1'


Jako:
If you have a lot of templates the typeahead could filter the list. I would not remove that possibility.

Ruslan-Aleev:
Yes, I thought about it. To be honest, in my practice I didn’t see a large number of templates. It seems to me that if there are a lot of templates, then the site structure is poorly thought out :)
We need a discussion.

JoshuaLuckers:
In its current state the PR should target 3.x

Ruslan-Aleev:
@JoshuaLuckers Yes, there certainly is not a bug, but so minor functional changes that I decided that PR would be useful in 2.x.
If the discussion decides that these changes are not needed, then i simply close this PR.

JoshuaLuckers:
The arrow down icon doesn't really imply that it's the trigger for the dropdown. Only when you hover over the element it does look like it triggers something (because of the background color).

The input and trigger element needs some distinction, maybe by using a border and/or applying a background color to the element that triggers the dropdown.

Such a change could go into 2.x because:
1. It's a fix to improve usability.
2. The behaviour of the template selection list stays the same.

Mark-H:
I'm not sure yet what's better - automatically opening the list or having the ability to search (yes, there are sites with structures that are a little poorly thought out, or large sites that simply have lots of different page types). 

Can we get the best of both worlds with `triggerAction: 'all'`?

PR #14587, opened 2019-05-17
Recently Edited Resources Widget fix: Set ID to object ID instead of row id

sdrenth:
What does it do?
Enables users to go to the page when using the buttons in the widget

Why is it needed?
Used to be going to an error page, because the row ID is used instead of the ID of the object itself.


Ibochkarev:
Related #14511

JoshuaLuckers:
@sdrenth can you give any insights about your approach v.s how it's done in PR #14511?

sdrenth:
@JoshuaLuckers I'm fine with the fix as provided in https://github.com/modxcms/revolution/pull/14511. So this PR can be closed if it is up to me.

PR #14590, opened 2019-05-19
Change of the name and description in the top menu of Form Customization

Ibochkarev:
### What does it do?
Change of the name and description in the top menu of Form Customization

### Why is it needed?
Fix lexicon keys for name and description Form Customization in the top menu

### Related issue(s)/PR(s)
#14352


Ibochkarev:
> Mark-H left a comment
>The primary key for installing menu items is the text. I think that means that the current "Manager >Customization" menu item would be left in-place, without its lexicon as that's also removed, >alongside a new item for FC. That seems to call for an upgrade script that removes the old 
 bespoke_manager menu item on upgrade.I also think this should be targeting 3.x, not 2.x?

https://github.com/modxcms/revolution/pull/14353#pullrequestreview-230555833

JoshuaLuckers (changes requested):
This requires an upgrade script that updates the old `text` and `description` values to the new ones.

opengeek:
Incorporated via #14780

PR #14593, opened 2019-05-22
Check if modx-content is defined before calling doLayout

jscaltreto:
I have been having an issue where occasionally widgets would fail to load properly in the manager. I tracked it down to the call to doLayout() and found that sometimes when the listener gets called getCmp('modx-content') returned undefined. If I had to guess I'd say it's a race condition wherein sometimes the listener fires too early. Adding a check has resolved the issue entirely and I haven't encountered any ill effects.

### What does it do?
Adds a check to verify modx-content is defined before calling doLayout

### Why is it needed?
Intermittent issue of certain widgets not loading in the manager

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -317,7 +317,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,autoDestroy: true
                 ,listeners:{
                     load: function(){
-                        cmp = Ext.getCmp('modx-content');
+                        var cmp = Ext.getCmp('modx-content');
                         if(typeof cmp !== "undefined") {
                             cmp.doLayout(); /* Fix layout bug with absolute positioning */
                         }
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -315,10 +315,13 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,groupField: this.config.groupBy || 'name'
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
-				,listeners:{
+                ,listeners:{
                     load: function(){
-						Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
-					}
+                        cmp = Ext.getCmp('modx-content');
+                        if(cmp !== undefined) {
+                            cmp.doLayout(); /* Fix layout bug with absolute positioning */
+                        }
+                    }
                 }
             });
         } else {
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -331,10 +334,13 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,remoteSort: this.config.remoteSort || false
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
-				,listeners:{
+                ,listeners:{
                     load: function(){
-						Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
-					}
+                        cmp = Ext.getCmp('modx-content');
+                        if(cmp !== undefined) {
```
JoshuaLuckers:
```suggestion
                        if(typeof cmp !== "undefined") {
```

JoshuaLuckers (dismissed):

jscaltreto:
@opengeek @Mark-H Can you have a look at this and  give feedback/merge as necessary? Thanks!

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -315,10 +315,13 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,groupField: this.config.groupBy || 'name'
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
-				,listeners:{
+                ,listeners:{
                     load: function(){
-						Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
-					}
+                        cmp = Ext.getCmp('modx-content');
```
Mark-H: Bit nitpicking, but this should be `var cmp = ...`; it's implicitly creating the variable now.

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -331,10 +334,13 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,remoteSort: this.config.remoteSort || false
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
-				,listeners:{
+                ,listeners:{
                     load: function(){
-						Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
-					}
+                        cmp = Ext.getCmp('modx-content');
```
Mark-H: Same here ^

Mark-H (changes requested):
Variables should be explicitly defined, see inline comments.

manager/assets/modext/widgets/core/modx.grid.js
JoshuaLuckers:
```suggestion
                        var cmp = Ext.getCmp('modx-content');
```

manager/assets/modext/widgets/core/modx.grid.js
JoshuaLuckers:
```suggestion
                        var cmp = Ext.getCmp('modx-content');
```

jscaltreto:
Good catch @Mark-H.

JoshuaLuckers (approved):

Mark-H (approved):

PR #14607, opened 2019-06-14
Add widget welcome dashboard changes

devlanda:
### What does it do?
Changes to the 'Add Widget' window in the Manager.

It changes the order. It puts the text "Please select a Widget to add to your Dashboard." above the 'Add Widget'-field label.
It also removes the unnecessary description below the 'Size' select box. There are a lot other sizes, so the text does not cover the size options that are available.

image

### Why is it needed?
It makes the manager-interface more intuitive for end users.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/manager/assets/modext/widgets/modx.panel.welcome.js
+++ b/manager/assets/modext/widgets/modx.panel.welcome.js
@@ -282,6 +282,11 @@ Ext.extend(MODx.window.DashboardWidgetAdd, MODx.Window, {
 
     getFields: function (config) {
         return [{
+            hideLabel: true,
+            xtype: 'displayfield',
+            html: _('widget_add_desc'),
+            anchor: '100%'
+        }, {
             fieldLabel: _('widget_add'),
             id: this.ident + '-widget',
             xtype: 'modx-combo-dashboard-widgets',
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/modx.panel.welcome.js
```diff
@@ -308,23 +313,13 @@ Ext.extend(MODx.window.DashboardWidgetAdd, MODx.Window, {
...
             fieldLabel: _('widget_size'),
             id: this.ident + '-size',
             xtype: 'modx-combo-dashboard-widget-size',
             name: 'size',
             value: 'half',
             anchor: '100%'
-        }, {
-            hideLabel: true,
-            xtype: 'displayfield',
-            html: _('widget_size_desc'),
```
alroniks, line 326: If it is unused need to delete the string from lexicon to avoid a mess.

alroniks (changes requested):

manager/assets/modext/widgets/modx.panel.welcome.js
devlanda: @Alroniks it is still needed for the more detailed view in modx.panel.dashboard.widget.js

alroniks: Good to know! ty @devlanda

alroniks (approved):

Mark-H (approved):

alroniks:
~~The interesting question, but how get such a window? Could you please provide relative URI to action in the manager? @devlanda~~
Sorry, nevermind, I found.

PR #14609, opened 2019-06-25
set published status when pub_date exactly == now

intersel:
### What does it do?
when using the create or update processors of resources,
set published status to "published" (1) when pub_date exactly == now

### Why is it needed?

In the function setFieldDefaults when creating a resource, $scriptProperties['published'] status is not set if the pub_date is exactly equal to "now" when we create the resource....

When updating the resource, we can consider too that the 'published' status should be considered to "published" (1) when pub_date is equal to "now".

This PR fixes these two points...

### DIFF
--- a/core/model/modx/processors/resource/create.class.php
+++ b/core/model/modx/processors/resource/create.class.php
@@ -224,7 +224,7 @@ class modResourceCreateProcessor extends modObjectCreateProcessor {
                 $scriptProperties['pub_date'] = 0;
             } else {
                 $scriptProperties['pub_date'] = strtotime($scriptProperties['pub_date']);
-                if ($scriptProperties['pub_date'] < $now) $scriptProperties['published'] = 1;
+                if ($scriptProperties['pub_date'] <= $now) $scriptProperties['published'] = 1;
                 if ($scriptProperties['pub_date'] > $now) $scriptProperties['published'] = 0;
             }
         }
--- a/core/model/modx/processors/resource/update.class.php
+++ b/core/model/modx/processors/resource/update.class.php
@@ -344,7 +344,7 @@ class modResourceUpdateProcessor extends modObjectUpdateProcessor {
             } else {
                 $strPubDate = $publishDate;
                 $publishDate = strtotime($publishDate);
-                if ($publishDate < $now) { /* if we're past publish date, publish resource */
+                if ($publishDate <= $now) { /* if we're past publish date, publish resource */
                     $this->setProperty('published',true);
                     $this->setProperty('publishedon',$strPubDate);
                     $publishDate = 0;


opengeek:
I don't understand the point of this. If you want it set to published when you create it, then set it. You should NOT be setting pub_date unless you are intending to change the published status at a specific point in time.

intersel:
perhaps I set the pub_date to publish my resource and for any reason it is just.... now when I do it.... 
why should I go to a nowhere land in the code when it is exactly "now" ? 
The case occurs to me in a specific use case (rest api using modx api processors) and was quite hard to figure out why the doc was not published...

opengeek:
pub_date is ONLY to trigger automatic publishing on a future date; unpub_date is ONLY to trigger automatic publishing on a future date. You should not be using these for any other reason.

intersel:
Actually, I can't get your point... the code is:
if ($scriptProperties['pub_date'] < $now) $scriptProperties['published'] = 1;
if ($scriptProperties['pub_date'] > $now) $scriptProperties['published'] = 0;
So what when $scriptProperties['pub_date'] == $now?
IMO, It should simply set it to 1...
the case could happen... even if it is a bad use of the processor according to you... it should give me an error message if it is that bad... 
Is there any collateral potential issue I don't see?

opengeek:
You should not be setting pub_date at all if your only intention is to create a Resource that is published. Simply set the published flag. Setting a pub_date implies that you are setting a future date for the published flag to be changed. I do not understand why there is any code in the processor at all which deals with setting the published flag based on the pub_date, as it seems unnecessary and counter-intuitive to the purpose of pub_date. I believe this may be because someone did not understand the code and no one who did noticed the change. I'll look into it deeper shortly.

Bruno17:
seems to be there like this for very long

https://github.com/modxcms/revolution/commit/b41ed7e4eb6f1bfe31a2a01ac4198c12b12cd30e#diff-82b7ce78ba995a8afa343ff659b87f85R83

Bruno17:
seems something like that was even there in etomite
https://github.com/rasel-sk/etomite/blob/master/manager/processors/save_content.processor.php#L95
and also still there in evolutioncms
https://github.com/evolution-cms/evolution/blob/2.0.x/manager/processors/save_content.processor.php#L142

intersel:
wouah... Sorry for having the insolence to dare to touch such a prehistoric code ;)

Mark-H (approved):
While I agree with @opengeek that technically speaking it doesn't make sense to have a future publish date that is before the current time, I have no objections against making that existing logic apply one second earlier.

Bruno17:
I was just trying to find, when and why it was implemented that way

JoshuaLuckers (approved):

PR #14571, opened 2019-04-25
Adding shortcuts dashboard widget

sdrenth:
### What does it do?
Adds a shortcuts widget to the dashboard with an overview of all MODX related shortcuts.

### Why is it needed?
Provides a handy overview for every MODX user to speed up their workflow with MODX.

**Note:** I've added the newObject in dashboard_widgets.php with array key 9 instead of 8 so it can be easily merged together with PR: https://github.com/modxcms/revolution/pull/14569 

Screenshot 2019-04-25 at 14 41 14

PR #14625, opened 2019-06-28
Adding shortcuts dashboard widget 

sdrenth:
# Rebased pull request of PR https://github.com/modxcms/revolution/pull/14571

### What does it do?
Adds a shortcuts widget to the dashboard with an overview of all MODX related shortcuts.

### Why is it needed?
Provides a handy overview for every MODX user to speed up their workflow with MODX.

**Note:** I've added the newObject in dashboard_widgets.php with array key 9 instead of 8 so it can be easily merged together with PR: https://github.com/modxcms/revolution/pull/14569 

Screenshot 2019-04-25 at 14 41 14

### Related PR's
https://github.com/modxcms/revolution/pull/14571


manager/controllers/default/dashboard/widget.shortcuts.php
```diff
@@ -0,0 +1,78 @@
...
+use MODX\Revolution\modDashboardWidgetInterface;
+
+/**
+ * @package modx
+ * @subpackage dashboard
+ */
+class modDashboardWidgetShortcuts extends modDashboardWidgetInterface
+{
+    /** @var modX $modx */
+    public $modx;
```
JoshuaLuckers: This is not necessarily needed here since `modDashboardWidgetInterface::class` already defines the variable.

sdrenth: @JoshuaLuckers Thank you for your feedback. I've removed those lines in my latest commit.

JoshuaLuckers (approved):

Mark-H:
As discussed in the related #14569, additional widgets would be better in an extra.

Ruslan-Aleev:
@Mark-H are you sure about this widget? In my opinion, it should be installed in the manager, because here is the **key combination of the manager** (there’s no way to find out about the key combination).
This widget in fact not an additional one (such as weather, statistics, and others), but a system one, it just was not at the start of MODX :)
And this is a simple widget, it is unlikely that problems with it may arise in the future.

Ruslan-Aleev:
**Get this widget back!**

widget

Mark-H:
There was consensus about this in #14569 - I'm not sure if they've released it as an extra since.

Ruslan-Aleev:
Yes, I read the discussion, but this widget is not an additional widget:

> In my opinion, it should be installed in the manager, because here is the key combination of the manager (there’s no way to find out about the key combination).
This widget in fact not an additional one (such as weather, statistics, and others), but a system one, it just was not at the start of MODX :)
And this is a simple widget, it is unlikely that problems with it may arise in the future.

Ruslan-Aleev:
I agree about all the additional widgets, but I think this widget should be in the manager.

> I'm not sure if they've released it as an extra since.

In my opinion, no, there was no release.

> There was consensus about this in #14569

Ivan offered to collect in extra, but I would not call it a consensus. In a year, he managed to completely leave the MODX project, and return again.
In general, I am confused by the fact that many controversial issues are resolved quickly. And obvious bug fixes (even those already sent to PR) will hang unattended for months. Some kind of lame logic.

JoshuaLuckers:
I'm not sure if a widget like this is useful to be present all the times. Maybe it should be part of the first login dialog? 

But since this PR is closed please open a new issue and/or PR.

Issue #14616, opened 2019-06-26
Does not work button Optimize Database


Ibochkarev:
### Summary
Does not work button Optimize Database

### Step to reproduce
Go to menu **Manage** -> **Reports** -> **System Info tab **Database tables** click button Optimize Database

### Observed behavior
When I try to search by system settings, I do not search for anything and I get an error `Processor not found: system/databasetable/optimizeDatabase`
 and in the error log

```
[2019-06-27 00:07:02] (ERROR @ /Volumes/HDD Data/web/refactoring/core/src/Revolution/modX.php : 1793) Processor /Volumes/HDD Data/web/refactoring/core/src/Revolution/Processors/system/databasetable/getlist.php does not exist; Array
(
    [action] => system/databasetable/getlist
)

[2019-06-27 00:07:43] (ERROR @ /Volumes/HDD Data/web/refactoring/core/src/Revolution/modX.php : 1793) Processor /Volumes/HDD Data/web/refactoring/core/src/Revolution/Processors/system/databasetable/optimizeDatabase.php does not exist; Array
(
    [action] => system/databasetable/optimizeDatabase
)
```

And console error:
```
ext-all.js:21 Uncaught TypeError: Cannot read property 'stopEditing' of undefined
    at constructor.refresh (ext-all.js:21)
    at MODx.grid.DatabaseTables.onStoreException (modx.jsgrps-min.js:1)
    at h.Event.fire (ext-all.js:21)
    at constructor.fireEvent (ext-all.js:21)
    at constructor.<anonymous> (ext-all.js:21)
    at h.Event.fire (ext-all.js:21)
    at Ext.data.HttpProxy.fireEvent (ext-all.js:21)
    at Ext.data.HttpProxy.onRead (ext-all.js:21)
    at Ext.data.HttpProxy.<anonymous> (ext-all.js:21)
    at Ext.data.Connection.handleResponse (ext-all.js:21)
```

### Environment
MODX 3 with the latest commit https://github.com/modxcms/revolution/commit/26430024e33db91b599e57a5b9c530b19cd6be31
PHP 7.1.2
MySQL 5.7.23

Ibochkarev:
Fixed in #14629 . You can close.

PR #14629, opened 2019-06-28
Fix the system actions related to the database widgets

JoshuaLuckers:
### What does it do?
Call the driver specific processors in the system database widget(s)

### Why is it needed?
There are no generic actions for some processors with driver specific processors after the refactor.

This is a quick fix. 

### Related issue(s)/PR(s)
Closes issue #14616

### DIFF
--- a/core/src/Revolution/Processors/System/DatabaseTable/GetList.php
+++ b/core/src/Revolution/Processors/System/DatabaseTable/GetList.php
@@ -1,0 +1,49 @@
+<?php
+/*
+ * This file is part of MODX Revolution.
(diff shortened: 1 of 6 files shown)


opengeek:
Why are these generic processors missing?

JoshuaLuckers:
I have no idea, I can reimplement them if that's more preferred.

opengeek (changes requested):
I think it would be preferable to restore the generic processors for these actions.

JoshuaLuckers:
@opengeek done!

PR #14643, opened 2019-07-14
Also check for scheme value as string

### LINKED ISSUE #13396: Link Tag scheme -1 is not handled correctly with Friendly URLs due to comparison error
### Summary
If you use link_tag_scheme -1 makeUrl in modcontext.class.php outputs an empty URL string and subsequently it is rendered as a full scheme URL and cached. Tags with explicit scheme properties or query properties work as expected.

### Step to reproduce
Turn on Friendly URLs; set link_tag_scheme to -1; set site_start to 1; create the following link tags [[~1]] (must be site_start) [[~2]]; View the href URLs and observe that 2 will be relative while the link to site_start is a full scheme URL.

### Observed behavior
Links built that match the site_start ID build to a full scheme URL, despite the explicit system settings being set to relative URL scheme.

### Expected behavior
All links, including site_start links should build into relative URLs, even if that means href "/".

### Environment
MODX all versions.

### Effect summary
If a mixed http and https site is visited and cached, links to site_start will be built and cached with the protocol the first visitor happened to use.

Breaks the concept of using agnostic link tags and scheme settings.

The ex-client issue that led me to this bug was them not wanting cert costs, but their cache constantly having HTTPS links (due to visitors, newer browsers, etc). In modern browsers this was **causing massive access issues, with many customers saying the site was 'always down'.** (I read them the riot act RE SSL, don't worry).

#### Problem with code
Scheme args can arrive as either integers or strings, but only integer instances are tested for:

<pre>
if ($config['friendly_urls'] == 1) {
	if ((integer) $id === (integer) $config['site_start']) {
		$alias= ($scheme === '' || $scheme === -1) ? $config['base_url']  : ''; // ORIGINAL
		// Inline scheme arg arrives as INT and is processed as base_url as expected
		// Plain links to home arrive as STRING and the alias is set to ''
		// Plain links to home with queries arrive as STRING and the alias is set to '' but they resolve to '?query=value' due to later query string handling
		$found= true;
	} else {
		$alias= $this->getResourceURI($id);
		if (!$alias) {
			$alias= '';
			$this->xpdo->log(xPDO::LOG_LEVEL_WARN, '`' . $id . '` was requested but no alias was located.');
		} else {
			$found= true;
		}
	}
	
} elseif (array_keys(array((string) $id), $this->resourceMap, true) !== false) {
	$found= true;
}
</pre>

### Code fix
Also test for string instances of scheme arg -1:
<pre>
if ($config['friendly_urls'] == 1) {
	if ((integer) $id === (integer) $config['site_start']) {   
	 $alias= ($scheme === '' || $scheme === -1 || $scheme === '-1') ? $config['base_url']  : ''; 
	 // HANDLES STRINGS
	 // All home links resolve to '/' as expected, queries are added after /
		
		$found= true;
	} else {
		$alias= $this->getResourceURI($id);
		if (!$alias) {
			$alias= '';
			$this->xpdo->log(xPDO::LOG_LEVEL_WARN, '`' . $id . '` was requested but no alias was located.');
		} else {
			$found= true;
		}
	}
 
} elseif (array_keys(array((string) $id), $this->resourceMap, true) !== false) {
	$found= true;
}
</pre>

### Related issues
Later on when the function tests for schemes it also only tests for integer -1, meaning any string -1 instances enter this code block with no effect. Possible effect: if the switch had a default case this would cause more unexpected makeUrl results.
<pre>
$host= '';
if ($target === null && $scheme !== -1 && $scheme !== '') { // this should also test STRING versions of -1
</pre>

johnpenny:
### What does it do?
Adds missing scheme checks.

### Why is it needed?
The missing scheme checks break the concept of the scheme setting, and has major side effects on some setups, see issue: #13396

### Related issue(s)/PR(s)
fixes #13396

### DIFF
--- a/core/model/modx/modcontext.class.php
+++ b/core/model/modx/modcontext.class.php
@@ -230,7 +230,7 @@ class modContext extends modAccessibleObject {
             }
             if (is_object($this->xpdo->context) && $this->get('key') !== $this->xpdo->context->get('key')) {
                 $config = array_merge($this->xpdo->_systemConfig, $this->config, $this->xpdo->_userConfig, $options);
-                if ($scheme === -1 || $scheme === '' || strpos($scheme, 'abs') !== false) {
+                if ($scheme === -1 || $scheme === '-1' || $scheme === '' || strpos($scheme, 'abs') !== false) {
                     $scheme= 'full';
                 }
             } else {
@@ -239,7 +239,7 @@ class modContext extends modAccessibleObject {
 
             if ($config['friendly_urls'] == 1) {
                 if ((integer) $id === (integer) $config['site_start']) {
-                    $alias= ($scheme === '' || $scheme === -1) ? $config['base_url'] : '';
+                    $alias= ($scheme === '' || $scheme === -1 || $scheme === '-1') ? $config['base_url'] : '';
                     $found= true;
                 } else {
                     $alias= $this->getResourceURI($id);


JoshuaLuckers (approved):
If we merge this, we should make sure to also apply this in 3.x

Mark-H (approved):

Mark-H:
Hey @johnpenny, it looks like you haven't signed our CLA yet. Could you please fill it out at https://modx.com/community/contribute/cla/ and post back here when you have?

Thank you!

johnpenny:
@Mark-H I've signed it. Thanks.

PR #14658, opened 2019-08-05
Resolved required 0 integer values being incorrectly marked as undefined

ian-emsens-sb:
### What does it do?
Resolved required `0` integer values being incorrectly marked as `undefined`

### Why is it needed?
Sending a `0` value in a `POST`/`PUT` request should be possible to facilitate enumerable TV fields, switching boolean values, ...

### Related issue(s)/PR(s)
None that I know of.

### DIFF
--- a/core/model/modx/rest/modrestcontroller.class.php
+++ b/core/model/modx/rest/modrestcontroller.class.php
@@ -140,7 +140,7 @@ abstract class modRestController {
 	    $missing = array();
 	    foreach ($fields as $field) {
 	        $value = $this->getProperty($field);
-	        if (empty($value)) {
+	        if (empty($value) && (is_int($value) && $value == 0)) {
 	            $missing[] = $field;
 	            if ($setFieldError) {
                     $this->addFieldError($field,$this->modx->lexicon('rest.err_field_required'));
--- a/core/model/modx/rest/modrestcontroller.class.php
+++ b/core/model/modx/rest/modrestcontroller.class.php
@@ -139,8 +139,12 @@ abstract class modRestController {
 	public function checkRequiredFields(array $fields = array(),$setFieldError = true) {
 	    $missing = array();
 	    foreach ($fields as $field) {
-	        $value = $this->getProperty($field);
-	        if (empty($value) && $value !== "0") {
+            $value = $this->getProperty($field);
+
+            $isEmptyString = empty($value) && $value !== "0";
+            $isNotBase = !is_int($value) && !is_bool($value);
+
+	        if ($isEmptyString && $isNotBase) {
 	            $missing[] = $field;
 	            if ($setFieldError) {
                     $this->addFieldError($field,$this->modx->lexicon('rest.err_field_required'));
--- a/core/model/modx/rest/modrestcontroller.class.php
+++ b/core/model/modx/rest/modrestcontroller.class.php
@@ -140,7 +140,7 @@ abstract class modRestController {
 	    $missing = array();
 	    foreach ($fields as $field) {
 	        $value = $this->getProperty($field);
-	        if (empty($value) && $value != 0) {
+	        if (empty($value) && $value !== "0") {
 	            $missing[] = $field;
 	            if ($setFieldError) {
                     $this->addFieldError($field,$this->modx->lexicon('rest.err_field_required'));
--- a/core/model/modx/rest/modrestcontroller.class.php
+++ b/core/model/modx/rest/modrestcontroller.class.php
@@ -140,7 +140,7 @@ abstract class modRestController {
 	    $missing = array();
 	    foreach ($fields as $field) {
 	        $value = $this->getProperty($field);
-	        if (empty($value) && (is_int($value) && $value == 0)) {
+	        if (empty($value) && $value != 0) {
 	            $missing[] = $field;
 	            if ($setFieldError) {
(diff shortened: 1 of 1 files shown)


ian-emsens-sb:
Self-reviewed and spotted some logic errors.

JoshuaLuckers:
Thanks for taking time to submit this PR and help make MODX better. Since you are a first-time contributor; please sign the the [Contributor License Agreement (CLA)](https://modx.com/community/contribute/cla/) and post back here when you have.

JoshuaLuckers:
The documentation for the `checkRequiredFields` function describes its behaviour as follows: Check for any empty fields.

By targeting 2.x the proposed changes will, in my opinion, break backwards compatibility because people might relay on the fact that `checkRequiredFields` uses the PHP-function `empty()` as source of truth. 

Furthermore I think it's a bit odd that required fields can't be "empty". In my opinion a required field can have a "empty" value.

ian-emsens-sb:
> The documentation for the checkRequiredFields function describes its behaviour as follows: Check for any empty fields.

The empty check is preserved, three exclusions to this condition have been added

1. The value is a string value of "0"
2. The value is an integer
3. The value is a boolean

Without this added functionality it is impossible to post `String` "0", `Integer` "0" or `Boolean` "0" because they all evaluate to `true` when processed by `empty()` while not being **actually** empty.

> Furthermore I think it's a bit odd that required fields can't be "empty". In my opinion a required field can have a "empty" value.

An empty value would equate to either `null` or `""` both of which do not pass the `empty()` check in PHP. In other words, the scenario you describe as possible `a required field can have a "empty" value` is currently impossible, regardless of my added changes.

JoshuaLuckers:
> The empty check is preserved, three exclusions to this condition have been added

For 2.x we only accept bug fixes without changing existing behaviour. This PR will change the behaviour of the function and therefore I think it's better to target 3.x instead of 2.x.

And if we target 3.x we should reconsider how the `checkRequiredFields` works because:
1. A required field should be able to have a value that's considered "empty".
2. If we check if the value of a required field is empty, who decides what's considered "empty"?

Mark-H:
I don't think "required" means something should be allowed empty - that's precisely what it's meant to check _against_. You don't want a required name to be left empty; and a checkbox that's allowed to be skipped is by its nature optional.

JoshuaLuckers:
When a request is handled by `modRestController` it calls the `checkRequiredFields` method. Assume we do a POST request it calls `checkRequiredFields` using `$this->postRequiredFields`.

The PHPDoc for this variable:
> /** @var array $postRequiredFields An array of required field keys that must be passed for POST requests */

This makes me believe it only checks if the keys are present in the request and that it shouldn't matter what values those fields have.

The method `checkRequiredFields` doesn't only check if the key is present it also checks if the field values are not empty (as described by the PHPDoc of the function).
> Check for any empty fields

The implementation of the `checkRequiredFields` method does exactly as documented. It checks the required fields for empty values using the PHP-function `empty` as source of truth.

It would make more sense to have a distinction between properties required to be present and properties required to have a value.

Maybe we can consider a `postRequirePresenceFields` and `checkRequirePresentFields` function that doesn't check for (empty) values. This way we could keep the current `checkRequiredFields` behaviour.

ian-emsens-sb:
> It would make more sense to have a distinction between properties required to be present and properties required to have a value.

In what case could you justify the presence of a variable without it having a value? Why even send it to the endpoint if it contains no value.

A flag such as `shouldUnpublish` or something comes to mind but surely you'd want that as a seperate endpoint e.g. PUT `api/unpublish/:id`.

The below request body just seems like very strange input into an API to me. Hell, I'm not even sure how you would program a nothing-value (not null, not undefined, not '', not 0, not false, not '0').
```
{
  property: 
}
```

Either way, I feel like we're getting a bit off topic here. If @Mark-H agrees that we should not break backwards compatability then this PR needs more work or needs to be re-targeted.

ian-emsens-sb:
To re-iterate, this PR basically adjusts the definition of empty for input into the API.

The current definition uses just `empty()` as its source of truth.

The proposed definition uses `empty()` but prevents the API from marking `"0"`, `0` and `false`values as empty. The values I've just mentioned all evaluate to `true` when passed into `empty()`

JoshuaLuckers:
> The proposed definition uses `empty()` but prevents the API from marking `"0"`, `0` and `false`values as empty. The values I just mention all evaluate to `true` when passed into `empty()`

Also consider to check for 0.0 (0 as a float).

JoshuaLuckers:
@opengeek what do you think? I'm ok with approving this PR if we also add support for 0.0 (0 as float).

opengeek:
My only concern is with changing behavior for 2.x with this (as Mark suggested this would do), but overall, I don't use this code in MODX (and likely never will). I'd prefer to get input from others that do use modRestController before deciding if this should be incorporated into a patch release as a bug fix. As far as I can tell, it is not used anywhere in the core, so it would be good to get input from any developers who have used it in Extras.

Jako:
I have no issues with that change and I am using modRestController in several packages. The behaviour is a bug in the current state.

opengeek:
@Mark-H? @JoshuaLuckers? Yay or nay on this one for 2.x?

JoshuaLuckers (approved):

JoshuaLuckers:
@opengeek @Mark-H Yay!

Mark-H (approved):

Issue #14639, opened 2019-07-09
Multiple processors not found after installing MODX 3


Ruslan-Aleev:
### Summary

1) When I go to the "Elements" tab in the tree, an error pops up **"Processor not found: element/getnodes"**

![elements_bug](https://user-images.githubusercontent.com/12523676/60929583-f5a38400-a2c2-11e9-9a8c-6191297f5bc6.png)

2) If I clearing the cache from the "Manage -> Clear Cache" pops up **"Processor /public_html/core/src/Revolution/Processors/system/clearcache.php does not exist ..."**

![cache_bug](https://user-images.githubusercontent.com/12523676/60929581-f50aed80-a2c2-11e9-882f-81f22c34fa9d.png)

3) If I go to the section "Manage -> Reports -> Error Log" then the page freezes.

![logs_bug](https://user-images.githubusercontent.com/12523676/60929585-f5a38400-a2c2-11e9-8505-9bde4faca9a4.png)

4) In the "Contexts" section, the processor is also not found.

![contexts_bug](https://user-images.githubusercontent.com/12523676/61054342-51275c00-a400-11e9-8b03-688ffdf857c5.png)

5) Also missing trash icon.

### How I installed MODX3

_I used the hosting web-console:_

- cd m3/public_html/
- rm -rf *
- git clone https://github.com/modxcms/revolution .
- git remote add upstream https://github.com/modxcms/revolution
- git checkout 3.x
- composer-php7.1 install
- cd _build
- cp build.config.sample.php build.config.php
- cp build.properties.sample.php build.properties.php
- php7.1 transport.core.php

_Then the normal installation started through **setup/**_

p.s. Perhaps these errors have already been described earlier, but I did not see.

### Environment
MODX 3

Server information:
Ubuntu
Apache: 2.4.10
MySQL: 5.7.21-20-beget
Nginx: 1.13.4
Perl: 5.26.1
PHP: 5.2  5.3  5.6  7.0  7.1  7.2  7.3
Python: 2.7

JoshuaLuckers:
I'm unable to reproduce any of these errors. Is this still an issue with the latest changes from 3.x?

Ruslan-Aleev:
Unfortunately, I can not check yet. As soon as I check the new changes, I’ll let you know.

Ruslan-Aleev:
@JoshuaLuckers Unfortunately, there are errors even with the latest changes.
I added the server information to the description.

[...]

PR #14662, opened 2019-08-10
Refactor references to processors

JoshuaLuckers:
### What does it do?
I made a start to rename all references to processors, starting with the "action" values.

### Why is it needed?
The legacy fallback does not work in all cases, especially when using a case sensitive filesystem.

### Related issue(s)/PR(s)
- issue #14639

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -202,7 +202,7 @@ Ext.extend(MODx,Ext.Component,{
         MODx.Ajax.request({
             url: MODx.config.connector_url
             ,params: {
-                action: 'system/clearcache'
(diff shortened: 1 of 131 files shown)


bezumkin (approved):

JoshuaLuckers:
Some of the "missed" actions can be found by using this search query in PhpStorm (match case, regex and file mask set to *.js): `'(browser|context|element|resource|search|security|source|system|workspace)\/[a-z]+(.*)'`

For example, the new TV processor: `core/src/Revolution/Processors/Element/TemplateVar/Create.php` will not replace `'element/tv/create'` (which it does in my PR).
There are also some other processors that are renamed which have to be done.

PR #14664, opened 2019-08-10
Fixed transparency in .png for TV

Ruslan-Aleev:
### Upd
I made a mistake with the &ra=1 parameter, it is not needed, &f=png is enough. &f=png adds transparency for png, but does not touch jpg.
Apparently when I tested only the &f=png parameter, the picture was cached.

### What does it do?
Pretty old bug with transparency in .png for TV. Regardless of the fixes previously, there was still no transparency.

**In general, png with white edges is not noticeable in the standard manager area, because the background is white, but sometimes artifacts appear on the png background.**

**For MODX3 this bug is also relevant.**

f=png

### DIFF
--- a/core/model/phpthumb/phpthumb.class.php
+++ b/core/model/phpthumb/phpthumb.class.php
@@ -1732,8 +1732,8 @@ class phpthumb {
 					if (is_array($getimagesize) && isset($getimagesize[2]) && ($getimagesize[2] == IMAGETYPE_PNG)) {
 						// explicitly exclude PNG from "-flatten" to make sure transparency is preserved
 						// https://github.com/JamesHeinrich/phpThumb/issues/65#issuecomment-256454015
-						$commandline .= ' -flatten';
 					} else {
+						$commandline .= ' -flatten';
 						$commandline .= ' -density '.phpthumb_functions::escapeshellarg_replacement($this->dpi);
 					}
 				}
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
Are there other solutions for this problem? If I understand the code correctly this will convert all thumbnails to the PNG file format and rotate them a small bit.

Ruslan-Aleev:
I did not see any other competent decision.
The image will rotate by 1 degree, but will not be converted to png, in general, the parameter f=png is displayed for any images (regardless of the file extension) in mediasources, as I understand it, to work with transparency, but I'm not sure about that.

https://github.com/modxcms/revolution/blob/2.x/core/model/modx/sources/modfilemediasource.class.php#L273

The output will be .jpg even with f=png
img

Ruslan-Aleev:
In general, png with white edges is not noticeable in the standard manager area, because the background is white, **but sometimes artifacts appear on the png background.**

I did this fix more for MIGX, because there it is more noticeable in the grid - background is not always white (and also decided to send for MODX).

MIGX grid with fix
img2

JoshuaLuckers:
I'm still curious why rotating fixes the transparency. Do you have any idea why that's the case? 

I prefer to fix the root cause of this issue (if possible).

Ibochkarev:
@JoshuaLuckers I also solved this problem by trial, as suggested by Ruslan.

Ruslan-Aleev:
@JoshuaLuckers Yes, it would be the right option to find the root cause :)
I don’t know why rotating fixes the transparency, but I did not see the right solution in network, apparently the problem is not very important.
But the problem is definitely in phpthumb.

JoshuaLuckers (dismissed):

Ruslan-Aleev:
@JoshuaLuckers I made a mistake with the &ra=1 parameter, it is not needed, &f=png is enough. &f=png adds transparency for png, but does not touch jpg.
Apparently when I tested only the &f=png parameter, the picture was cached.

JoshuaLuckers (approved):

Mark-H (approved):

PR #14671, opened 2019-08-13
Added an additional .sub class in the menu for nested li

Ruslan-Aleev:
### What does it do?
Added an additional .sub class in the menu for nested li which have child items. 
Added an extra icon to the .sub class.

Because there is already a branch in 3.0, I don’t see the point of why this function is not added for 2.7.2 either.

sub_info

Because commits were from the old version of the branch, I recreated PR

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/11579

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -121,6 +121,17 @@
   }
 }
 
+#modx-navbar .sub {
+  &:after {
+    border: 5px solid transparent;
+    border-left: 5px solid rgb(96,114,124);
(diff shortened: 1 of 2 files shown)


JoshuaLuckers (approved):

Mark-H (approved):

opengeek:
This was squashed and merged into 2.x as 590d270ba. I hoped that squashing and merging this via CLI would work to mark the PR merged since there is a squash and merge option here in the UI now, but I guess it doesn't. SMH.

Issue #14670, opened 2019-08-13
MODX3: When saving a document, the form for creating a new document opens


Ruslan-Aleev:
### Summary
When creating and saving a document instead of continuing editing the current document, a form for creating a new document opens.

### Expected behavior
Continue editing the current document.

![doc_new](https://user-images.githubusercontent.com/12523676/62932858-01a6c800-bdd2-11e9-8dd2-628023374bee.gif)

### Environment
MODX3

ghost:
Are there settings for this behavior? A long time ago there were options and settings to select the behavior or to set a default.

Ruslan-Aleev:
I did not find such a setting

JoshuaLuckers:
Did you test this on the 3.x branch or with the changes from PR #14662?

Ruslan-Aleev:
I looked in the version of MODX3 with @bezumkin  edits (in my opinion, the essence is the same)

Ibochkarev:
Confirmed, #14662 does not fix this bug

JoshuaLuckers:
Maybe it’s related to https://github.com/modxcms/revolution/issues/14667 

I will check it tomorrow

JoshuaLuckers:
It's not related, ~I'm not sure yet what's causing this issue~.

Ibochkarev:
Fixed in #14676 
@modxbot close

PR #14676, opened 2019-08-15
Fix an issue when process/action values are CamelCased

JoshuaLuckers:
### What does it do?
Process/action values can contain CamelCased values, this will compares the lowercased values.

### Why is it needed?
When creating a new document it didn't load the update controller after saving.

### Related issue(s)/PR(s)
Fixes issue #14670


Mark-H (approved):

Ruslan-Aleev (approved):
Works for me! Thanks

Issue #14638, opened 2019-07-09
Dragging categories in the item tree does not work


Ruslan-Aleev:
### Summary
Dragging categories into the item tree does not work, although you can drag the category into another and the save animation pops up.

![cat_bug](https://user-images.githubusercontent.com/12523676/60927872-6b582180-a2bc-11e9-8169-295735a8dbfb.gif)

### Environment
MODX 2.7.1 >

Ruslan-Aleev:
Dragging to a category does not work for MODX 3.x too, although **you can drag it to the root category**:

![cats](https://user-images.githubusercontent.com/12523676/66839437-771c4980-ef77-11e9-9741-9f27af3a6f90.gif)

Issue #14667, opened 2019-08-12
Remove references to old core actions


JoshuaLuckers:
### Summary
In MODX 3.x `modConfigJsProcessor`/ `ConfigJs` doesn't load actions for backwards compatibility. 

This line of code is removed in `ConfigJs` (the processor responsible to  output $modx->config to JSON).

```
// Load actions for backwards compatibility (DEPRECATED)
$actions = $this->modx->request->getAllActionIDs();
$o .= 'MODx.action = ' . $this->modx->toJSON($actions) . ';';
```

### Step to reproduce
1. Checkout the 3.x branch with all recent changes.
2. Drag and drop a resource in the tree
3. Open the developer console

### Observed behavior
An error appears: `TypeError: undefined is not an object (evaluating 'MODx.action['resource/update']')`

### Expected behavior
No error. 

### Environment
MODX 3

Mark-H:
Considering actions have been deprecated since 2.2 I'm okay with that... though that will cause breakage in extras.

JoshuaLuckers:
> Considering actions have been deprecated since 2.2 I'm okay with that... though that will cause breakage in extras.

I'm not sure who removed it from the code. We can always reconsider the removal and put it back if to many (unmaintained) extra's break that are being used "heavily".

JoshuaLuckers:
Or create a “compatibility mode” extra!

JoshuaLuckers:
It has been "agreed" on to add back support for the deprecated actions and remove support in a future versions.

PR #14681, opened 2019-08-17
Add back support for deprecated actions

JoshuaLuckers:
### What does it do?
Add back support for deprecated actions

### Why is it needed?
We are already changing enough in 3.x, we want to maintain some backwards compatibility.
This fixes an issue where moving resources didn't work in 3.x

### Related issue(s)/PR(s)
~Fixes issue #14638~

### DIFF
--- a/core/src/Revolution/Processors/System/ConfigJs.php
+++ b/core/src/Revolution/Processors/System/ConfigJs.php
@@ -113,6 +113,10 @@ class ConfigJs extends modProcessor
         $o .= $this->modx->toJSON($c);
         $o .= '; MODx.perm = {};';
 
+        // Load actions for backwards compatibility (DEPRECATED)
+        $actions = $this->modx->request->getAllActionIDs();
+        $o .= 'MODx.action = ' . $this->modx->toJSON($actions) . ';';
+
         if ($this->modx->user) {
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
What about MODX 2.7.2? For this branch, this bug is also relevant.

Mark-H (approved):

Mark-H:
@Ruslan-Aleev this is not relevant to 2.x as far as I know. This code is already in 2.x, but was removed from 3.x only in the massive refactor that landed recently.

Ruslan-Aleev:
@Mark-H I mean, the issue in this PR there is also in MODX 2.7 - https://github.com/modxcms/revolution/issues/14638
Maybe the question is wider than backward compatibility?

JoshuaLuckers:
My mistake, this PR isn't a fix for #14638. 
This PR fixes the issue where moving resources in 3.x didn't work (I can't find the related issue).

Mark-H:
Oh, sorry, I didn't see that issue. [These lines of code already exist in 2.x](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/system/config.js.class.php#L115-L117), so it's not the solution for that issue. This is in relation to #14667

Ruslan-Aleev:
Got it, thanks for the info!

Issue #14669, opened 2019-08-13
The rss folder during installation is not created


Ibochkarev:
### Summary
The `rss` folder during installation is not created along the path `core/cache/rss/`

### Step to reproduce
Go to Error Log

### Observed behavior

`[2019-08-13 07:48:20] (ERROR @ /home/s19424/www/core/vendor/simplepie/simplepie/library/SimplePie.php : 1474) User warning: /home/s19424/www/core/cache/rss/ is not writable. Make sure you've set the correct relative or absolute path, and that the location is server-writable.
`

### Expected behavior
The folder must be created during installation

### Environment
MODX 3 with the latest commit

Mark-H:
Fixed in #14682

PR #14682, opened 2019-08-17
Create the RSS cache directory if it doesn't exist

JoshuaLuckers:
### What does it do?
When loading the RSS feed, check if the RSS cache directory exists, if not: create it.

### Why is it needed?
Prevent the error log to be filled with errors

### Related issue(s)/PR(s)
Fixes issue #14669

### DIFF
--- a/core/src/Revolution/Processors/System/Dashboard/Widget/Feed.php
+++ b/core/src/Revolution/Processors/System/Dashboard/Widget/Feed.php
@@ -56,7 +56,12 @@ class Feed extends modProcessor
     {
         $feed = new \SimplePie();
 
-        $feed->set_cache_location($this->modx->getOption('core_path') . 'cache/rss/');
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Processors/System/Dashboard/Widget/Feed.php
```diff
@@ -56,7 +56,12 @@ public function loadFeed($url)
     {
         $feed = new \SimplePie();
 
-        $feed->set_cache_location($this->modx->getOption('core_path') . 'cache/rss/');
+        $cachePath = $this->modx->getOption('core_path') . 'cache/rss/';
```
opengeek: I believe this needs to use the `cache_path` (or `xPDO::OPT_CACHE_PATH`) setting.

opengeek (changes requested):

Mark-H (approved):

PR #14689, opened 2019-08-20
Generate sortable tag signature with serialized array values

opengeek:
### What does it do?
Improves the tag signature created by `modElement::getTag()`.

### Why is it needed?
When using non-default template engines in MODX (for example, the Fenom implementation by @bezumkin) arrays can be passed as properties. When a tag is generated for the element in these cases, the array will be replaced with the hash of a random string (as arrays are not scalar values). This means that element output can not be matched from the element cache since the random string will never match.
This also allows properties to be so sorted so that the cached output can be matched when similar element tags have a different property order.

### Related issue(s)/PR(s)
Replaces #14577 targeting 3.x instead of 2.x, and includes changes suggested in my review of that PR.

### DIFF
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -268,16 +268,18 @@ class modElement extends modAccessibleSimpleObject
     public function getTag()
     {
         if (empty($this->_tag)) {
-            $propTemp = [];
             if (empty($this->_propertyString) && !empty($this->_properties)) {
+                $propTemp = array();
                 foreach ($this->_properties as $key => $value) {
+                    $key = trim($key);
                     if (is_scalar($value)) {
-                        $propTemp[] = trim($key) . '=`' . $value . '`';
+                        $propTemp[$key] = $key . '=`' . $value . '`';
                     } else {
-                        $propTemp[] = trim($key) . '=`' . md5(uniqid(rand(), true)) . '`';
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (commented):
It would be nice to have a Unit Test for this (modElement already has a few tests).

opengeek:
I look forward to your contribution, @JoshuaLuckers ! LOL!

opengeek:
In all seriousness, I'm not sure there is an easy way to test the array support in this because the default implementation shipped with the core has no way to express a property value as an array. Array property values are something exclusively being done in 3rd party add-ons. I'll see if I can find a way to do this, but I'm afraid it will take some mocking of the parser to emulate something like Fenom or Twig.

JoshuaLuckers:
I was wondering how I could test this PR, that’s why I suggested a Unit Test.

core/src/Revolution/modElement.php
```diff
@@ -268,16 +268,18 @@ protected function isStaticFilesAutomated()
     public function getTag()
     {
         if (empty($this->_tag)) {
-            $propTemp = [];
             if (empty($this->_propertyString) && !empty($this->_properties)) {
+                $propTemp = array();
```
JoshuaLuckers, line 272:
```suggestion
                $propTemp = [];
```

Issue #14568, opened 2019-04-25
Make MODX Setup safer and inaccessible to non owners


donShakespeare:
### Why is it needed?

1. Prevent someone from overwriting/hijacking one's site.
2. Users like myself often keep the setup folder around (renamed of course) due to the hassle of getting a new setup folder in order to rerun upgrade to clear out some database/path glitch.
3. Many users (as it will continue to be observed - for whatever human reasons) will uncheck the option that auto deletes the setup folder - oh and will ignore the only single Manager Dashboard warning.
4. In case the Server glitches and just refuses to honor the installation's request to delete the setup folder.

### Solution 1.

Access to setup folder would depend on the presence/absence of a specific file. See Typo3 for example https://docs.typo3.org/typo3cms/InstallationGuide/QuickInstall/TheInstallTool/Index.html

Best to have the said extensionless file in the **core** folder so that it can benefit from being outside webroot (e.g not in `/setup/` folder).
Maybe:
 `core/setup/FIRST_INSTALL`
 `core/setup/UPGRADE`
 `core/setup/ADVANCED`

The absence of all of these files would render Setup inaccessible.
The presence of any or all of these files would make available the respective installation/upgrade option (applicable to solution 2)

![image](https://user-images.githubusercontent.com/5103329/56715698-41478800-66ed-11e9-8054-7e73fd052a0d.png)

- These files would be manually created by: the admin or automated scripts.
- If any of these files exist at anytime (even during an installation/upgrade), MODX Manager should red-alert the admin with badgering popups on every Manager page (not only on Dashboard - many admins rarely look at this page)

![image](https://user-images.githubusercontent.com/5103329/56716383-13634300-66ef-11e9-9531-9a8eefd8c6e9.png)

### Solution 2.

A cleaner approach, which is implemented by Vanilla Forums: https://docs.vanillaforums.com/developer/troubleshooting/

Setup would be only accessible according to the presence/absence of a clause in the `core/config/config.inc.php`

```
<?php
$database_type = 'mysql';
$database_server = 'localhost';

$setup_mode = 'UPGRADE'; // 'FIRST_INSTALL' || 'ADVANCED' 
```
### Solution 3.

Or finally, request an access key upon installation. Automated install/upgrade scripts could do this as well.
The key would have to match that found in `core/config/config.inc.php`
```
<?php
$database_type = 'mysql';
$database_server = 'localhost';

[...]

PR #14692, opened 2019-08-22
Lock setup after it is successfully executed

opengeek:
### What does it do?
Writes a .locked/ directory in the setup/ folder upon successful completion of running installation or upgrade.

### Why is it needed?
To prevent unauthorized users from unexpectedly running setup/ if the folder is left in place (or even renamed).

### Related issue(s)/PR(s)
#14568


Mark-H:
It seems to be working _so well_ that it even prevents the redirect to the manager login (`/setup/index.php?action=login` => `FATAL ERROR: MODX Setup cannot continue.`).

Omeryl:
@Mark-H Is there a reason it needs to hit that specific URL for the redirect to the manager? Can we just embed the manager link on that template, skipping the need for the 3xx redirect?

Mark-H:
Just doing some testing, when moving `$install->lock()` to `setup/controllers/login.php` it does work as expected:

```
if ($install->settings->get('cleanup')) {
    $install->removeSetupDirectory();
}
else {
    $install->lock();
}
```

@Omeryl The `login` controller actually does the cleanup of the `setup/` folder and temporary settings cache file before the redirect. While it does seem like something that could be handled in `complete`, it does serve a purpose.

Omeryl:
@Mark-H So, the problem then becomes that unless someone clicks that login button we'd never lock the setup though?

Mark-H:
It's a hard button not to press, and is also currently responsible for removing the setup if that was checked, so I don't have a problem with that. ;) 

Schermafbeelding 2019-08-23 om 00 52 40 1

Omeryl:
@Mark-H I guess the problem is, the only reason we're in this situation is because of the edge cases where people don't delete the setup and shoot themselves in the foot. So if you don't do it on set up complete, and still require user action, we'd still be letting people shoot themselves here. I feel like it'd be beneficial to perhaps continue this lock, refactor the login stuff out to somewhere else (perhaps manager first login if it detects the lockfile, prompt the sudo user?), and embed a link to the manager.

opengeek:
@Mark-H I had wanted to do it when they click proceed after successful installation, but yeah. Duh. I could move the check out of the preloadChecks to solve the problem with doing it before the login controller.

donShakespeare:
>the **only reason** we're in this situation is because of the **edge cases** where people don't delete the setup and shoot themselves in the foot.

not quite the only reason as many have attested. Plus, this safety measure or similar is implemented by other respectable CMSes as discussed earlier. Just saying that a solution in general is (would have been) the default way to go.

This is awesome.

Omeryl:
@donShakespeare The problem I'm arguing is that if we only put the lockfile in place _after_ the user clicks the "Login" button, there's still going to be a good number of cases where setup is left open. But if this is the way we want to do it, and leave that edge case, I'm fine with it. Just putting my two cents in, the CMSes you list put that lock file in as soon as the setup is completed, not after clicking some "login" button.

donShakespeare:
Yes, you are absolutely correct.
Locking the folder upon user interaction defeats the purpose of this noble pursuit.

opengeek:
Alright. I tried a different approach. It should be a lot better this way, locking right after the sessions are cleared. This also blocks CLI install, though I'm wondering if that is even necessary. Discuss.

Ruslan-Aleev:
I think we should transfer the checkbox with the removal of the "Setup" folder to the "Installation Summary" stage:

install

Or move the checkbox in front of the login button:

install2

_Because at the stage of entering the manager panel, the button blocks all attention (as mentioned above). And below the button, the checkbox looks like the checkbox "Remember me", if you do not pay attention._

Mark-H (approved):

JoshuaLuckers (approved):

PR #14608, opened 2019-06-21
Widget: Recently edited resources, date/time from system settings

devlanda:
### What does it do?
Dashboard widget: 'Recently edited resources'. It takes the date and time format from the system settings:

- manager_date_format
- manager_time_format

Instead of using a fixed format, '%B %d, %Y' which is defined in recentlyeditedresources.tpl.

### Why is it needed?
Date and time formatting should be in line with the system settings as preferences differ per geographical area.

### Related issue(s)/PR(s)
N/A

PR #14693, opened 2019-08-23
Edited resource date and time from system settings, after refactor

devlanda:
### What does it do?
Dashboard widget: 'Recently edited resources'. It takes the date and time format from the system settings:

- manager_date_format
- manager_time_format

Instead of using a fixed format, '%B %d, %Y' which is defined in recentlyeditedresources.tpl.

### Why is it needed?
Date and time formatting should be in line with the system settings as preferences differ per geographical area.

### Related issue(s)/PR(s)
This PR is based on https://github.com/modxcms/revolution/pull/14608. That needed changes because of the massive refactor.

### DIFF
--- a/core/src/Revolution/Processors/Security/User/GetRecentlyEditedResources.php
+++ b/core/src/Revolution/Processors/Security/User/GetRecentlyEditedResources.php
@@ -92,6 +92,14 @@ class GetRecentlyEditedResources extends modObjectGetListProcessor
         $resourceArray = $resource->get(['id','pagetitle','description','published','deleted','context_key', 'editedon']);
         $resourceArray['pagetitle'] = htmlspecialchars($resourceArray['pagetitle'], ENT_QUOTES, $this->modx->getOption('modx_charset', null, 'UTF-8'));
 
+        $dateFormat = $this->modx->getOption('manager_date_format');
+        $timeFormat = $this->modx->getOption('manager_time_format');
+
+        $resourceArray['editedon_date'] = date($dateFormat, strtotime($resourceArray['editedon']));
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
Thanks for taking time to redo the original PR! 

Since we are improving things with this PR it might be a good to use the `DateTimeImmutable` class instead of the `date` function. What do you think?

devlanda:
> Thanks for taking time to redo the original PR!
> 
> Since we are improving things with this PR it might be a good to use the `DateTimeImmutable` class instead of the `date` function. What do you think?

I made some changes. So, it uses the class `DateTimeImmutable` now.

JoshuaLuckers (approved):

opengeek:
Do not merge this PR as is. This will need to be cherry-picked since this PR branch contains a bunch of unrelated merge commits.

sdrenth:
@opengeek Does this PR needs to be adjusted or will the integrators handle the cherry pick themselves?

Mark-H:
In general, it's always a nice thing to rebase when conflicts or additional commits get into a PR, just because there's a bigger supply of contributors than integrators at the moment :)

Mark-H (approved):

Mark-H:
I can't seem to push after cherry picking and have asked Jason for advice. A rebase without the merge commits would be great or perhaps Jason knows how to get around the branch protections that seem to be getting in my way.

Mark-H:
I've gone ahead and rebased this PR for you. For how to do that yourself in the future, please see my description here: https://github.com/modxcms/revolution/pull/14572#issuecomment-529673427

Issue #14663, opened 2019-08-10
Different panel templates in the manager panel


Ruslan-Aleev:
### Summary
MODX has different panel design templates for some sections of the manager panel: some with a top tab, some without (see screen below). And because of the different templates, the styles are not correct displayed.

We need to add a tab to all panels so that the styles display correctly and make it easier to edit style in the future.

![template_panel](https://user-images.githubusercontent.com/12523676/62826171-e10b2200-bbc7-11e9-8029-6f83b136e1e5.png)

---

This is not difficult, for example, **for the "Namespaces" section (/manager/assets/modext/workspace/namespace/modx.namespace.panel.js) need change**:

```
...

,items: [{
    html: _('namespaces')
    ,id: 'modx-namespaces-header'
    ,xtype: 'modx-header'
},{
    layout: 'form'
    ,items: [{
        html: '<p>'+_('namespaces_desc')+'</p>'
        ,xtype: 'modx-description'
    },{
        xtype: 'modx-grid-namespace'
		,cls:'main-wrapper'
        ,preventRender: true
    }]
}]

...
```

to:

```
...

,items: [{
    html: _('namespaces')
    ,id: 'modx-namespaces-header'
    ,xtype: 'modx-header'
},MODx.getPageStructure([{
    title: _('namespaces')
    ,layout: 'form'
    ,items: [{
        html: '<p>'+_('namespaces_desc')+'</p>'
        ,xtype: 'modx-description'
    },{
        xtype: 'modx-grid-namespace'
        ,cls:'main-wrapper'
        ,preventRender: true
    }]
}])]

...
```

### Environment
MODX 2.7>

Mark-H:
Fixed in #14695

PR #14695, opened 2019-08-23
Added tabs for manager section templates

Ruslan-Aleev:
### What does it do?
I made the design of the manager sections to one view, adding tabs on top.

template_panel
Removed unnecessary styles for sections without tabs (since there were no tabs, these styles were added).

Slightly fixed the styles for `.panel-desc`:

before
after
There were useless styles, especially for the resource, I simplified them. In the future, styles will be easier to edit, and `.panel-desc` does not look bad with a white background.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14663


Mark-H (dismissed):

JoshuaLuckers:
If your window is a bit smaller then usual these changes will add a vertical scrollbar when viewing system settings.

Before: 
Screenshot 2019-09-01 at 13 17 05

After: 
Screenshot 2019-09-01 at 13 17 18

Ruslan-Aleev:
@JoshuaLuckers Yes you are right. I wanted to make the template in "System Settings" the same as in the "Lexicons", but I did not check the width of the screen.
However, there is also a problem for the "Lexicons", see below:

lexicons

**Maybe it makes sense to remove the explanatory text for both templates?**

@Mark-H what do you think?

JoshuaLuckers:
The labels add value and should stay. Is it possible to place them about the elements they describe?

Ruslan-Aleev:
Another option is to reduce the width of the elements :)
Correcting the layout will not work, because output in tables.
Placing the label inside the lists will not work either, because in the main section controller, the first in the list should be "core" (here I'm not sure).

Mark-H:
@Ruslan-Aleev  I don't know what the best way to solve that is, sorry.

Ruslan-Aleev:
I think we can reduce the elements width. @JoshuaLuckers what do you think?
Although this does not help, because with a minimum screen width, the labels will not fit in any way.

JoshuaLuckers:
If you can solve the conflict I'll approve this PR and we can figure out a solution with a new PR.

Ruslan-Aleev:
Well, in general, the purpose of this PR was to add tabs on top.
I added labels myself, because similar ones were in “Lexicons” and I thought that in “System Settings” labels would be useful.
But this problem does not apply to this PR, it has always been.
How to solve it - I don’t know, I can return it as it was - by deleting the labels from the "System Settings", but the labels in the "Lexicons" always were, and they also did not fit in width.

JoshuaLuckers:
Revert the changes not part of the original intend and create a new issue for the labels (and issues they cause). 😄

Ruslan-Aleev:
@JoshuaLuckers I removed the labels in the "System Settings", but here the problem is in extJS, because all layouts made with tables.
This behavior on all grids of the MODX manager panel, I will create an issue, but in my opinion it is useless as long as there is extJS in the manager panel =)

JoshuaLuckers (approved):

Mark-H (approved):

Mark-H:
Thanks @Ruslan-Aleev. 

Just so you know for the next time, an interactive rebase (`git rebase -i upstream/3.x`) is prefered over merging 3.x into your branch to avoid extra commits making their way in. See my comment [here](https://github.com/modxcms/revolution/pull/14572#issuecomment-529673427) for a bit more detail over how you can do that.

PR #14704, opened 2019-09-01
Make \modX and other class_aliases for backwards compatibility available in all requests

### LINKED ISSUE #14687: MODX 3 Errors with snippets
### Summary
1) When trying to install a snippet, an error is written to the console, although the snippet is installed:
`MODX\Revolution\modX::loadClass is deprecated since version 3.0. Replace references to class modx.modNamespace with MODX\Revolution\modNamespace to take advantage of PSR-4 autoloading.`

snippet

2) When trying to unnistall / reinstall the snippet, an error is written to the console at each step, although the snippet is unnistalled / reinstalled:
snippet2

3) When a snippet is called in a resource (template, chunk or in content) - **the page crashes with a 500 error.**

### Environment
MODX 3

Mark-H:
## What does it do?

Moves the class_alias for the 2.x `modX` class, and several xPDO-related aliases, to a new `core/include/deprecated.php` file which is autoloaded. 

### Why is it needed?

At its previous location in the BC `modx.class.php` file, the class alias would only be available in scenarios where the `modX` class was manually loaded, e.g. in a connector or external code interacting with MODX.

However in cases where MODX is initiated normally (e.g. a manager or frontend request), the class alias was not available. 

This causes third party code that run in a normal request (e.g. a snippet or plugin) to break if they're typehinting against `modX`. 

For example.. [a certain plugin](https://github.com/modmore/ClientConfig/tree/release-2.0) with a [getService call in a plugin that fires OnMODXInit](https://github.com/modmore/ClientConfig/blob/release-2.0/core/components/clientconfig/elements/plugins/clientconfig.plugin.php#L40) that has a `modX` typehint [in its service constructor](https://github.com/modmore/ClientConfig/blob/release-2.0/core/components/clientconfig/model/clientconfig/clientconfig.class.php#L42).. would result in a fatal error `Uncaught TypeError: Argument 1 passed to ClientConfig::__construct() must be an instance of modX, object given`, whereas with this patch it loads as expected. 

~I've also considered putting the class_alias in [the list of `class_alias`es](https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/modX.php#L45), but because that list is ahead of the `modX` class definition that causes an error as well. It could be done at the bottom of the `modX.php` file but that seems a little off. So I settled on the constructor as a reasonable place for it.~

Following the discussion in the PR, all class_alias references are now in a separate file that is included automatically with the composer autoloader. 

### Related issue(s)/PR(s)

Fixes #14687


JoshuaLuckers:
Did you consider putting it at the bottom of the file (outside the class definition)?

Mark-H:
> I've also considered putting the class_alias in the list of class_aliases, but because that list is ahead of the modX class definition that causes an error as well. **It could be done at the bottom of the modX.php file but that seems a little off**. So I settled on the constructor as a reasonable place for it.

If there's consensus to have it at the bottom of the file instead of in the constructor I'm happy to adjust the PR; this just seemed cleaner. I'm about a 2 on the SSOGAF on where this should go, so long as it's somewhere it will always run ;)

Mark-H:
Joshua noted that this actually doesn't work when loading MODX externally, so this PR needs more work. Moving below the class could be an option,  but that might not work when used [in this way](https://docs.modx.org/current/en/extending-modx/modx-class/loading-externally) either.

opengeek:
I was thinking we should be loading all of these in a simple include for BC aliases. This would be done automatically if calling the old core/model/modx.class.php, otherwise, it would be done optionally based on some type of flag indicating we want to enable BC support. Where it is done in the modX class is I suppose the only issue to be resolved here.

Mark-H:
@opengeek I'm happy with a generic include as well. 

What about always including that, while also putting up a notice that it will be removed in, say, v3.2 or v3.3? It's a bit tricky to get it to load in all cases - just modx.class.php isn't enough (see the use case I described in the PR).

opengeek:
@Mark-H I was thinking we could do a config define to enable BC for global namespace references, but I'm also okay with just deciding to remove an always-on option at some point in the future.

JoshuaLuckers:
I like this solution:
> Collect the aliases in a separate file, deprecated.php and include it additionally with composer.

Source: [PHP: Using class_alias to maintain BC while moving/renaming classes](https://www.schmengler-se.de/en/2016/09/php-using-class_alias-to-maintain-bc-while-move-rename-classes/)

Mark-H:
As `modx.class.php` loads the autoloader, and so does `modX.php`... that does sound like it could do the trick.

JoshuaLuckers:
> As `modx.class.php` loads the autoloader, and so does `modX.php`... that does sound like it could do the trick.

We can give it a try! 😄

Mark-H:
I'll try to update the pr tomorrow or Monday :)

Mark-H:
Updated to a new include. Note that `composer dump-autoload` needs to be executed when testing the PR to update the autoloader.

PR #14720, opened 2019-09-07
Use fully qualified class names for modTemplateVar

JoshuaLuckers:
### What does it do?
Use the fully qualified class names for modTemplateVar when possible

### Why is it needed?
To make sure we have the right class names.

### Related issue(s)/PR(s)
Fixes #14714

PR #14721, opened 2019-09-07
Use fully qualified class names for modTemplate

JoshuaLuckers:
### What does it do?
Use the fully qualified class names for modTemplate when possible

### Why is it needed?
To make sure we have the right class names.

### Related issue(s)/PR(s)
Related PR #14720


alroniks (approved):
:+1:

Mark-H (approved):

Issue #14702, opened 2019-08-29
Lock for elements (snippets, templates, tv’s) does not work good


dimitrihilverda:
### Summary
When using the checkbox Lock [element] for editing, only on chunks the acl's are respected

### Step to reproduce 
as administrator lock some elements, (chunks and snippets) for editing. 

### Observed behavior
Then login with a user that is not allowed to edit_locked, try to edit a chunk with a lock: you get a notice that the chunk is locked and editing is not possible, do the same with a locked snippet, the edit window is opened and I can remove the lock and save the snippet.

### Expected behavior
On all elements that are locked I expect the same notice as with the locked chunks

### Environment
MODX 2.7.1

PR #14723, opened 2019-09-07
Check permissions for locked elements

JoshuaLuckers:
### What does it do?
When editing a locked snippets, templates, template variables, and plugins make sure the user has the right permissions to do so

### Why is it needed?
To make sure only users with the `edit_locked` permissions can edit locked elements.

### Related issue(s)/PR(s)
Fixes issue #14702

### DIFF
--- a/core/lexicon/en/plugin.inc.php
+++ b/core/lexicon/en/plugin.inc.php
@@ -25,6 +25,7 @@ $_lang['plugin_err_create'] = 'An error occurred while creating the plugin.';
 $_lang['plugin_err_ae'] = 'A plugin already exists with the name "[[+name]]".';
 $_lang['plugin_err_invalid_name'] = 'Plugin name is invalid.';
 $_lang['plugin_err_duplicate'] = 'An error occurred while trying to duplicate the plugin.';
+$_lang['plugin_err_locked'] = 'Plugin is locked.';
 $_lang['plugin_err_nf'] = 'Plugin not found!';
 $_lang['plugin_err_ns'] = 'Plugin not specified.';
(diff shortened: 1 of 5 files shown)


Mark-H:
Is the PR missing new lexicons for snippets/chunks/tvs or were they already there (but not used)?

opengeek:
I think there is some confusion. Locked refers to the locking mechanism associated with Resources only.

Mark-H:
Elements can also be locked, though that is indeed a different mechanism than resource locks (and it didn't really work as they could still be edited).

opengeek:
@Mark-H they can, but it is a much simpler mechanism and has never been related to ACLs. The "lock" mechanism on elements is simply to prevent non-administrators from being able to edit it.

opengeek:
Ok, I'm wrong about the ACLs, but I thought the mechanism has always been to prevent saving, not to prevent loading the element into the edit view. The check is in modElementUpdateProcessor::beforeSave().

opengeek (changes requested):
After a discussion with @Mark-H, we have determined that the issue is that the lock test in modElementUpdateProcessor::beforeSave() is happening after the `locked` property is set/unset from the request. 🤦‍♂The `locked` property was never intended to prevent loading the edit view for an element, but only to prevent saving changes to it. There are other ACL permissions for blocking access altogether. So, the test for locked needs to happen before the property is set/unset from the request, AND, if chunks are already being prevented from loading via this permission, that needs to be reverted.

Mark-H:
I've created a new PR #14739 that fixes this in the processors, rather than the controllers, so am closing this one. Let me know if you disagree @JoshuaLuckers.

PR #14731, opened 2019-09-10
Corrected profile layout

Ruslan-Aleev:
### What does it do?
Corrected the profile layout in accordance with the user layout.

user_1

user_2

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14420

### DIFF
--- a/manager/assets/modext/sections/security/profile/update.js
+++ b/manager/assets/modext/sections/security/profile/update.js
@@ -120,64 +120,108 @@ MODx.panel.UpdateProfile = function(config) {
         ,items: [{
             columnWidth: .5
             ,items: [{
-                xtype: 'textfield'
-                ,fieldLabel: _('user_full_name')
+                id: 'modx-user-fullname'
                 ,name: 'fullname'
-                ,maxLength: 255
+                ,fieldLabel: _('user_full_name')
+                ,xtype: 'textfield'
                 ,allowBlank: false
                 ,anchor: '100%'
-            }, {
-                xtype: 'textfield'
-                ,fieldLabel: _('email')
+                ,maxLength: 255
+            },{
+                id: 'modx-user-email'
                 ,name: 'email'
+                ,fieldLabel: _('user_email')
+                ,xtype: 'textfield'
                 ,vtype: 'email'
                 ,allowBlank: false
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

manager/assets/modext/sections/security/profile/update.js
```diff
@@ -208,38 +252,60 @@ Ext.extend(MODx.panel.UpdateProfile,MODx.FormPanel,{
     },
     getItemsRight: function (config) {
         var items = [{
+            id: 'modx-user-country'
+            ,name: 'country'
+            ,fieldLabel: _('user_country')
+            ,xtype: 'modx-combo-country'
+            ,anchor: '100%'
+            ,value: ''
```
Jako, line 260: Please remove the value. Maybe #14732 will be solved with that.

Ruslan-Aleev: This does not help, although the bug now works differently :)

country_2

Jako: Without `value: ''` the default value is '0' there. Where is that panel filled?

Ruslan-Aleev: I do not quite understand your question. If you delete `value: ''`, then when refresh page (sometimes), instead of an empty string, `0` appears, like in gif.
This is relevant for this files (with or without this PR): https://github.com/modxcms/revolution/blob/3.x/manager/assets/modext/widgets/security/modx.panel.user.js#L600 and https://github.com/modxcms/revolution/blob/3.x/manager/assets/modext/sections/security/profile/update.js#L237.
This is relevant for this sections of the manager panel (**Country** field): `/manager/?a=Security/User/Update&id=1` and `/manager/?a=security/profile` .

Jako: Do you see the flickering value just before the blank value occurs? It shows `DZ`, but the keys i.e. in core/lexicon/country/en.inc.php are lowercase. Is that user created in MODX? How did you get the `DZ` value in that field?

Mark-H (approved):

Mark-H:
🙌

PR #14735, opened 2019-09-11
Added adaptive styles for MODX browser

Ruslan-Aleev:
### What does it do?
In the previous version of adaptive styles, the MODX browser did not work at all, this PR fixes this.

By the way, for **MODX 2.7 the browser also does not work**, we can copy these styles for MODX 2.7.2

browser

### Related issue(s)/PR(s)
None


Mark-H (approved):

Mark-H:
🙌

PR #14748, opened 2019-09-21
Add hover effect and tooltip for editable grid columns [#14176]

### LINKED ISSUE #14176: Make it clear that the value can be changed in MODX tables [ux/ui]

Mark-H:
### What does it do?

Adds a special renderer to all grids (* at least, all grids that define `columns` as a config property; some grids like the settings grid that build a custom ColumnModel need additional work, as shown in this PR as well) that automatically adds a class to grid cells that are editable, as well as a tooltip.

Through the CSS, the cursor is changed, and the color set to the splash color on hover. Open to improvement ideas for the styling.

### Why is it needed?

Make it clear what columns can be edited inline, which is not visible anywhere currently.

This is an enhancement for #14580 which added similar tweaks only to the settings grid. Following on the discussion there, this PR adds it globally to all grids. 

### Related issue(s)/PR(s)

Fixes #14176
Follows up on the discussion in #14580

(Requires rebuild)

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -540,6 +540,12 @@ a.x-grid-link:focus {
   cursor: pointer;
   &:hover, &:focus {
     color: $colorSplash;
+    > div::after {
+      @include awesome-font;
+      content: fa-content($fa-var-pen);
+      margin-left: 0.5em;
+      color: $colorSplash;
+    }
   }
 }
 
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -536,6 +536,12 @@ a.x-grid-link:hover,
 a.x-grid-link:focus {
   text-decoration: underline;
 }
+.x-editable-column {
+  cursor: pointer;
+  &:hover, &:focus {
+    color: $colorSplash;
+  }
+}
 
 /* panel stylings */
 .modx-page-header,
--- a/core/lexicon/en/default.inc.php
+++ b/core/lexicon/en/default.inc.php
@@ -133,7 +133,6 @@ $_lang['edit_tv'] = 'Edit TV';
 $_lang['editing'] = 'Editing: [[+name]]';
 $_lang['editedon'] = 'Edit Date';
 $_lang['editing_form'] = 'Editing Form';
-$_lang['editable_field'] = 'Double-click to edit this value.';
 $_lang['element_duplicate_values'] = 'Duplicate Resource Values?';
 $_lang['element_name_new'] = 'Name of New Element';
 $_lang['element_caption_new'] = 'Caption of New Element';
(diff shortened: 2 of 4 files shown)


Ruslan-Aleev (approved):
It works, thanks!

tooltip

Ruslan-Aleev:
Although the color change is almost invisible.

Mark-H:
What about a pencil icon added to the end of the value on hover?

editable

Ruslan-Aleev:
An interesting solution :) It looks cool

JoshuaLuckers:
It's great to see people being informed about "hidden" functionality but do we really want to trigger peoples attention all the time with a tooltip? Maybe we can add a small delay to it? Or maybe only show it when hovering the pen icon?

Mark-H:
It does pop up a lot. Maybe the tooltip isn't needed with the pen icon, is that clear enough?

Ruslan-Aleev:
Yes, probably, pen icons will be enough :)
Although the example in gif is grids where all the fields are corrected, basically there are not many such types of grids.

Ruslan-Aleev:
**It seems to me that the pen icon are enough.**

The only thing is that new MODX users may not understand where and how many times to click (most likely they will click on the icon once and only then twice), but I don’t think that this is a problem :)

JoshuaLuckers:
> It does pop up a lot. Maybe the tooltip isn't needed with the pen icon, is that clear enough?

Or maybe show the tooltip when hovering the pen icon? 
Another thing I was thinking about is opening the "edit" mode when hovering an editable field but I'm not sure if that is possible and practical.

ghost:
Please no! There is already more than enough stuff popping in and out whenever I move the mouse around on a form!

Mark-H:
I'll remove the tooltip to only leave the pen icon. Probably thursday or over the weekend.

JoshuaLuckers:
> I'll remove the tooltip to only leave the pen icon. Probably thursday or over the weekend.

Would it be possible to trigger "edit mode" when clicking the pen icon?

gpsietzema:
I have to say I really love this addition!

Jako:
@JoshuaLuckers Not that easy with the clickable edit icon, because it is a pseudo element. But there are some possibilities: https://stackoverflow.com/questions/25465397/how-to-make-the-area-of-css-pseudo-element-clickable

Mark-H:
Not _the_ weekend I had in mind to work on this, but PR updated. Just the pencil icon and subtle hover effect now.

Ruslan-Aleev (approved):

JoshuaLuckers (approved):
Great work! 

_I did however click the pencil icon a few times expecting it to do something_

Issue #14742, opened 2019-09-19
3.0 Installation Issues


ghost:
### Summary
Installation of 3.0 nightly build fails

### Step to reproduce
download nightly build, try to install

### Observed behavior
If in a subdirectory, paths are not found, there is no styling and installation cannot continue beyond the first unstyled page, all links appear to go to the site root /setup/

If in the root, installation proceeds past the first page, to an empty second page, and fails with a fatal error when trying to create the database.

``` Warning: array_merge(): Expected parameter 2 to be an array, null given in /Applications/MAMP/htdocs/core/vendor/xpdo/xpdo/src/xPDO/xPDO.php on line 302 ```

``` Fatal error: Uncaught Error: Class '\xPDO\Om\\xPDODriver' not found in /Applications/MAMP/htdocs/core/vendor/xpdo/xpdo/src/xPDO/xPDO.php:1890 Stack trace: #0 /Applications/MAMP/htdocs/core/vendor/xpdo/xpdo/src/xPDO/xPDO.php(303): xPDO\xPDO->getDriver() #1 /Applications/MAMP/htdocs/setup/includes/modinstall.class.php(468): xPDO\xPDO->__construct(NULL, NULL, NULL, Array, Array) #2 /Applications/MAMP/htdocs/setup/includes/modinstall.class.php(171): modInstall->_connect(NULL, NULL, NULL, NULL, Array) #3 /Applications/MAMP/htdocs/setup/includes/test/modinstalltest.class.php(315): modInstall->getConnection() #4 /Applications/MAMP/htdocs/setup/includes/test/modinstalltest.class.php(51): modInstallTest->_checkDatabase() #5 /Applications/MAMP/htdocs/setup/includes/modinstall.class.php(247): modInstallTest->run(NULL) #6 /Applications/MAMP/htdocs/setup/controllers/summary.php(27): modInstall->test(NULL) #7 /Applications/MAMP/htdocs/setup/includes/request/modinstallrequest.class.php(80): include('/Applications/M...') #8 /Applications in /Applications/MAMP/htdocs/core/vendor/xpdo/xpdo/src/xPDO/xPDO.php on line 1890 ```

### Expected behavior
It should install from within a subdirectory.

The 2.7.2 nightly build installs as expected.

### Environment
modx-3.0.0-dev-20190918
MAMP 5.5
[...]

PR #14756, opened 2019-09-24
Load the setup assets when MODX is inside a subdirectory

JoshuaLuckers:
### What does it do?
Don't set a base href for the setup.

### Why is it needed?
It would prevent the assets (CSS/JS) to be loaded if MODX was inside a subdirectory.

### Related issue(s)/PR(s)
Issue #14742


Mark-H:
3 minutes to merge, is that a record?

opengeek:
@Mark-H LOL! Don't be jelly!

PR #14763, opened 2019-09-25
Add Font Awesome 5 Solid

wax100:
### What does it do?
Fix some FontAwesome icons

### Why is it needed?
U can see it in Minishop2 products grid
http://prntscr.com/payxkv

After css fix http://prntscr.com/payxxr


Mark-H:
The index.css file should not be edited; the source files for that are in _build/templates/default.

wax100:
> The index.css file should not be edited; the source files for that are in _build/templates/default.

Yes, I see... look like bug in FA or I am mistaken)

Mark-H:
It might just need to integrated differently through the sass files. Perhaps another `@import` specific for the solid icons? I'm not sure.

Ibochkarev:
This font is already included
https://github.com/modxcms/revolution/blob/3.x/_build/templates/default/sass/index.scss#L9

image

PR #14774, opened 2019-10-01
Fixed a broken MODX browser for mobile screens

Ruslan-Aleev:
### What does it do?
Fixed the broken MODX browser and fixed some styles for mobile screens.
The browser styles, of course, is not perfect, but at least it works.

b

a

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/_browser.scss
+++ b/_build/templates/default/sass/_browser.scss
@@ -240,3 +240,84 @@
     background-image: url($imgPath + 'modx-theme/transparency-pattern.png');
   }
 }
+
+@include media($mobile) {
+  .modx-browser {
+    top: 15px !important;
+    max-height: 100% !important;
+    overflow-y: scroll;
+  }
+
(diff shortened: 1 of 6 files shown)


JoshuaLuckers (approved):

opengeek:
Is this really fixing behavior that once worked but no longer does? Or is this making the MODX browser work on mobile screens for the first time?

Ruslan-Aleev:
Browser did not work in my opinion, from the first versions where the adaptive appeared, 2.6 or earlier. Browser started working for the first time, at least I don’t remember that it worked at all.

JoshuaLuckers:
@opengeek is it ok with you if we merge this into 2.x? :)

Ruslan-Aleev:
@JoshuaLuckers , I adjusted the styles of the icons in the tree, did it more correctly than before. Test please.

Ibochkarev (approved):

PR #14775, opened 2019-10-01
Reduce buttons and TV category headers for mobile devices

Ruslan-Aleev:
### What does it do?

- Reduce action buttons in #modx-action-buttons
- Reduce TV category tab headers

to save space.

**Before:**
b

**After:**
a

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/_buttons.scss
+++ b/_build/templates/default/sass/_buttons.scss
@@ -292,6 +292,11 @@ button {
     }
   }
 
+  .x-toolbar-left {
+    width: auto !important;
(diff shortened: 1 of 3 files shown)


JoshuaLuckers:
What's the reason for decreasing the TV tab size?

Ruslan-Aleev:
On mobile devices, there is much less space and a scroll appears, screenshots above - these are screenshots from the desktop browser, just with a smaller width.

JoshuaLuckers (approved):

Issue #14654, opened 2019-08-01
MODX 3: Setup Installation summary error field color


meshkov:
### Summary
if there are errors, then the color of error fields is gray instead of red
![2019-08-01_11-03-09](https://user-images.githubusercontent.com/1703670/62280594-ed1f1300-b454-11e9-92e6-e6663639aa7c.png)

### Step to reproduce
during installation, change permissions on some folder
`chmod 000 connectors`

### Expected behavior
should be red like this:
![2019-08-01_12-01-40](https://user-images.githubusercontent.com/1703670/62280693-1a6bc100-b455-11e9-8a21-f47fb85b4158.png)

### Environment
MODX 3 installation with composer

Ibochkarev:
I will fix it this week.

Mark-H:
I don't think you've worked on this yet, right @Ibochkarev?

Mark-H:
PR sent: #14708

Ibochkarev:
Fixed #14708 @modxbot close

meshkov:
I try with MODX 3 with last comment installation with composer and it not fix:
![image](https://user-images.githubusercontent.com/1703670/66909581-ad100b00-f015-11e9-837a-ff90fbad1ef1.png)

Jako:
Sorry, thats a different issue.

Ruslan-Aleev:
@meshkov I fixed it in PR - https://github.com/modxcms/revolution/pull/14785, try to test.
Perhaps then this PR will be introduced with minor changes.

PR #14785, opened 2019-10-10
Fixed some styles for the installer

### LINKED ISSUE #14900: Вug front-end Install screen one
## Bug report

Russian language

Screen

Install screen

### Environment
MODX Revolution 3.0.0-alpha1 , Browser Firefox 71.0 (64-bit)

Ruslan-Aleev:
### What does it do?
Fixed some styles for the installer. Changed colors, borders, boldness, changed checkboxes via icons, made more universal error messages, deleted duplicate styles, etc.

install_0
install_1
install_2
install_3

I wanted to make a design, as in sketches - https://modx3.org/updates/week-18-june-11-17-2018.
In my opinion, that green in sketches looks better.

This PR is for discussion, if the community does not accept the new green color, we can return the old green :)

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14654
https://github.com/modxcms/revolution/pull/14507

### DIFF
--- a/_build/templates/default/sass/installer.scss
+++ b/_build/templates/default/sass/installer.scss
@@ -48,10 +48,10 @@ body {
 }
 
 body {
-    color: #333333;
+    color: $mineShaft;
     font: $baseText;
     -webkit-font-smoothing: antialiased;
-    background: #f1f2f7;
+    background: $wildSand;
     display: flex;
     flex-direction: column;
 
@@ -76,7 +76,7 @@ body {
     cursor: pointer;
     line-height: 1;
     font-size: 18px;
-    padding: 13px 25px;
+    padding: 15px 30px;
     width: 170px;
     transition: background-color .2s ease-out;
     background: $green;
@@ -137,7 +137,7 @@ body {
         .wrap {
             display: block;
             padding: .5rem .7rem;
-            border: 2px solid $softGray;// $blue
+            border: 1px solid $softGray;// $blue
             margin: 5px;
             background-color: $lightestGray;
             border-radius: 3px;
@@ -162,10 +162,9 @@ body {
         input:checked ~ .wrap,
         input:focus ~ .wrap {
             border-color: $blue;
-            background-color: lighten($blue, 45%);
+            background-color: lighten($blue, 38%);
         }
     }
-
 }
 
 [type="checkbox"] {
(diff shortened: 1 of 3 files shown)


Ibochkarev:
@Ruslan-Aleev Nice! Thanks

_build/templates/default/sass/installer.scss
```diff
@@ -1,6 +1,12 @@
 @import "../node_modules/normalize-scss/sass/normalize/import-now";
 @import "colors-and-vars";
 
+$darkBlack: #343434;
+$lightGray: #F6F6F6;
+$softGray: #D6D6D6;
+$blue: #4A90E2;
+$green: #14C8A0;
```
JoshuaLuckers: To keep the colours consistent across MODX it's better to change the values inside [_colors-and-vars.scss](https://github.com/modxcms/revolution/blob/3.x/_build/templates/default/sass/_colors-and-vars.scss)

JoshuaLuckers (commented):
I like the new green but I'm a bit worried about the color contrast. I'm really interested in @sottwell's opinion about this.

_build/templates/default/sass/installer.scss
Ruslan-Aleev: I understand this, but I say that this PR is for discussion, so that global styles are not edited if the community does not accept the changes.

Mark-H:
In current, "Next" button contrast ratio is 2.59. With this PR it's less with 2.14. For AA compatibility we need at least 4.5, which means a dark grey-ish (*or dark green hue) color rather than white. The test section seems fine though Chrome is being a pain in not showing the contrast.

Ruslan-Aleev:
Is contrast really important? The same WordPress does not really care about this: just a gray button on a gray background :)

Mark-H:
WordPress is an excellent example, thank you for bringing that up ;)

WordPress has a dedicated accessibility team with the goal "All new or updated code released in WordPress must conform with the Web Content Accessibility Guidelines 2.0 at level AA"

https://make.wordpress.org/accessibility/handbook/

And specifically: https://make.wordpress.org/accessibility/handbook/design/use-of-color/#about-color-contrast


While MODX does not currently have a dedicated team/handbook/defined goal relating to accessibility, WCAG 2/AA is a really good baseline to work towards. I'm happy if we at least _move in the right direction_.

Ruslan-Aleev:
Interesting, but I still won’t believe that this gray button in WP is more contrast than the green button in PR.
ustanovka_wordpress

Mark-H:
[That appears to be a score of 5.34, above the 4.5 minimum for AA](https://webaim.org/resources/contrastchecker/?fcolor=656565&bcolor=F5F5F5). 

The bright green in the setup button definitely stands out more than the grays in the WP example, but contrast is about the difference between the foreground and background of the button itself. White on bright green just isn't a great fit where contrast is concerned. Making the text color a really dark green for example would much more quickly meet the minimum.

JoshuaLuckers:
Wrong repo guys 😋.

Ruslan-Aleev:
> ... but contrast is about the difference between the foreground and background of the button itself.

Thank you. Now it is more clear.

JoshuaLuckers:
Jokes aside, @Mark-H did a great color suggestion we should try to improve the color contrast ratio.

Ruslan-Aleev:
Yes, even in the current version, the manager panel colors do not pass the test,we will need to come up with something.
https://webaim.org/resources/contrastchecker/?fcolor=ffffff&bcolor=528738

Mark-H:
If completed in the next 24 hours or so this could potentially still be included in alpha1, but it's cutting it real close so I'm marking it for the next milestone.

Ruslan-Aleev:
There is a question with colors (https://github.com/modxcms/revolution/pull/14785/files#diff-238eaef74a8ad5b87131006e1af93bf5R4), I can remove color vars and then PR can be accepted.
**And the question of color can be raised before the next release.**

Ruslan-Aleev:
@Mark-H I adjusted the styles. The contrast ratio has remained the same.
I will create a separate issue and in the future we should consider new colors for better contrast.

JoshuaLuckers:
The Advanced Options checkboxes look disabled (the first one is, the second one isn't):

Screen Shot 2020-01-26 at 13 00 43

When the install fails, the message indicating the failure has no background

Screen Shot 2020-01-26 at 13 02 00

Relevant element:
Screen Shot 2020-01-26 at 13 02 30

Ruslan-Aleev:
@cla-bot check

Ruslan-Aleev:
@JoshuaLuckers I changed styles for checkboxes and notifications:

installer_1

installer_2

Ruslan-Aleev:
@cla-bot check

JoshuaLuckers:
Amazing work. I have some small suggestions I would love to get your feedback on:

**Welcome**
Make the selected language have a different background color (like how it is in alpha 2).
Only the thin border now indicates that it's selected, having a different background makes it more clear what language is selected.

**Options**
Consider making the selected install option (new, upgrade or advanced) also have a different background color for the same reasons as mentioned above.

**Install completed**
The checkbox has a blue background. I had the impression the checkbox was selected. This wasn't the case. Maybe use a different background color for checkboxes that are unchecked?

Ruslan-Aleev:
@JoshuaLuckers show visually on the screenshots what you mean, not everything is understood.

JoshuaLuckers:
Image of the unchecked checkbox, because of the color it might give people the impression it's checked.

Screen Shot 2020-02-07 at 11 47 11

Current background of the selected language:
Screen Shot 2020-02-07 at 18 45 37

Ruslan-Aleev:
@JoshuaLuckers I corrected the styles, see gif.
For installation type I changed the color of the icons, **not the background**, because the background cannot be set for a nested span (there is a hidden checkbox, and next to the span, the border is set using the span):

installer_color2

I also changed the commits using force-push, but now the travis check fails, which is strange.

JoshuaLuckers (approved):
Great work!

Issue #14597, opened 2019-06-03
MODX3: Icon for "container" is not displayed


Ruslan-Aleev:
### Summary
The resource does not change the icon in the tree, even if the resource is a "container".

**MODX 2.7.1:**

![icon_1](https://user-images.githubusercontent.com/12523676/58821447-23ebcf00-8646-11e9-8425-d5142c5877eb.png)

**MODX 3:**

![icon_2](https://user-images.githubusercontent.com/12523676/58821448-23ebcf00-8646-11e9-8f24-bb2a11e215f5.png)

Possibly related to https://github.com/modxcms/revolution/pull/14383

### Environment
MODX 3

Mark-H:
Fixed in #14786

PR #14786, opened 2019-10-11
Fixed display of icons in the tree

Ruslan-Aleev:
nodes

Moved styles below **styles for file types**, as otherwise they were in charge - https://github.com/modxcms/revolution/blob/4b0cf0a9f93cb9e708fde6877f42b5655265db86/_build/templates/default/sass/_tree.scss#L675

For example, a resource with a file type of **.pdf** would always have a pdf icon, even if this resource is a symbolic link or folder. Now there is no such problem.

And I changed the classes from `.tree-folder` to` .parent-resource`, `.icon-folder`. Because currently there is no `.tree-folder` class in the tree.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14597

### DIFF
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -164,48 +164,6 @@
   width: 12px;
 }
 
-/* icons for the various tree elements */
-.tree-context:before {
-  @extend %pseudo-font;
-  content: fa-content($fa-var-globe);
-}
-
-.x-tree-node-expanded .tree-folder:before {
-  @extend %pseudo-font;
-  content: fa-content($fa-var-folder-open);
-}
-
-.x-tree-node-collapsed .tree-folder:before,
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

Mark-H:
I'd love a brief (1-2 sentence) explanation of what this does to fix the problem. You know, like it asks in the pull request template which you removed ;)

It's not clear from me looking just at the diff, in part because it doesn't show inline changes (if any?) because the entire code block was moved... is that what fixed it or did something else change?

Ruslan-Aleev:
@Mark-H I added a description

JoshuaLuckers:
@Mark-H ping!

Mark-H:
@JoshuaLuckers Pong!

Mark-H (approved):

Issue #14428, opened 2019-02-26
A save animation appears above the logo when editing a resource


Ruslan-Aleev:
### Summary
If you update the panel **when editing a resource** and drag any resource in the tree, the save animation will appear above the logo once, see the picture below.
After refreshing the page, the bug will be repeated.

![menu_animate_bug](https://user-images.githubusercontent.com/12523676/53422486-041a8e80-39f9-11e9-98ef-ac490c7a6d11.gif)

### Environment
MODX3, Windows 7, Google Chrome

Ruslan-Aleev:
As it turned out, this bug is also relevant for version 2.7.

Ibochkarev:
Fixed #14790 
@modxbot close

PR #14787, opened 2019-10-11
Fixed a bug with appearance of save animation over logo

Ruslan-Aleev:
### What does it do?
If you edit a resource and drag the same resource in the tree, a save animation appears next to the logo.

save-anim

This PR fixes this bug through increasing `z-index`.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14428

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -22,7 +22,7 @@
   background: $colorSplash;
   max-width: $navbarWidth;
   position: absolute;
-  z-index: 2;
+  z-index: 99999;
   height: 100%;
 }
 


JoshuaLuckers:
I believe it's better to fix the actual issue, the animated element being in the wrong place. It looks to be related to the animated effect you get when, for example, you drag and drop resources in the tree.

Ruslan-Aleev:
I do not think that someone will correct this differently, because work will take more time than possible profit :) I tried to find out why this happens, but I did not find it.
The element appears before `</body>`, here is the GIF:

save-anim-2

JoshuaLuckers (approved):
> I do not think that someone will correct this differently, because work will take more time than possible profit :)

I agree and the actual underlying issue can always be fixed later down the road.

Mark-H (changes requested):
This has worked in the past so should be fixed properly. A z-index of 99999 also sounds arbitrary and not based on any particular other z-index that we're trying to +1.

Ruslan-Aleev:
> This has worked in the past so should be fixed properly

This bug is also relevant for 2.x, i.e. this did not work in the past :)
And z-index can be set **35001**, if it matters, because z-index 35000 stands at the div with animation, see gif above.

JoshuaLuckers:
> I do not think that someone will correct this differently, because work will take more time than possible profit :)

It took a few hours but I think I found the root cause of this issue!

Ruslan-Aleev:
@JoshuaLuckers Thank for your time, a little later I will test your PR and close my :)

PR #14795, opened 2019-10-16
Added a link and buttons to the context grid

Ruslan-Aleev:
It did not work out correctly to rebase, and the discussion in the old PR was not specific :) - had to create another PR.

### What does it do?
Changed the order of functions in a `modx.grid.context.js` file so that in the future it would be easier to edit them.

Added a link and buttons to the context grid. **I figured out how to add a context menu by clicking on the gear.**
In general, it seems to me that the most useful option is the option with the actions "Edit" and "Copy", and all actions will be in the "Gear-menu" (or maybe it’s worth leaving just “Edit” and the "Gear-menu").

What it looks like:

contexts

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14125
Old PR - https://github.com/modxcms/revolution/pull/14581

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -528,13 +528,28 @@ textarea.x-form-field {
 }
 
 a.x-grid-link {
-  color: $mainText;
+  color: $colorSplash;
+  text-decoration: underline;
+}
+
+a.x-grid-link:hover, a.x-grid-link:focus{
   text-decoration: none;
 }
 
-a.x-grid-link:hover,
-a.x-grid-link:focus {
-  text-decoration: underline;
+.x-grid-buttons {
+  text-align: center;
+}
+
+.x-grid-buttons li {
+  cursor: pointer;
+  display: inline-block;
+  font-size: 1.1em;
+  line-height: .7;
+  margin-right: 10px;
+}
+
+.x-grid-buttons li:last-child {
+  margin-right: 0;
 }
 
 /* panel stylings */
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -232,7 +232,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
             ,url: this.config.url
             ,params: p
             ,listeners: {
-            	'success': {fn:this.refresh,scope:this}
+                'success': {fn:this.refresh,scope:this}
             }
         });
     }
@@ -255,7 +255,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,url: this.config.url
                 ,params: p
                 ,listeners: {
-                	'success': {fn:function() {
+                    'success': {fn:function() {
                         this.removeActiveRow(r);
                     },scope:this}
                 }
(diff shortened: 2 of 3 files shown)


Ruslan-Aleev:
@Mark-H

> I've not tested it yet, but I wonder, could we get rid of the onClick here if each of the action buttons is assigned a handler instead?
> Maybe not because they're not extjs-type buttons.. but worth a check to make the per-grid implementation easier perhaps?

@Jako

> It should be more effective the current way, since each button would need a click handler (or listener) then. At the moment the whole button column gets one click handler assigned.

Ruslan-Aleev:
@Mark-H how will you have time, test this PR, if everything is fine, I would try add the same functionality to other grids.

theboxer:
Heya,

Personally I think it's too much of a hustle and repetitive coding for adding this feature.

The new action column, onClick handler and buttonColumnRenderer should be added to the `MODx.grid.Grid` and have a param to turn it on/off. The function `buttonColumnRenderer` should call `getActions` or similar function to get the array of buttons to render, where the `gear` action/icon should be added automatically, so user defines only grid specific actions.

Using classes to figure out what action to take seems inappropriate (I know Revo uses it on several places, but I don't think we have to continue this horrible pattern), why don't we use something like `data-action="duplicate"`?  

From this point also comes the update for `onClick` handler, having a specific attribute for the action, the onClick handler can be more generic, look if this attribute exists on the target element and if there is a function with a specific name (for example `actionDuplicate`) and if yes let's execute it. This way you'll only have to define which actions to show and their "handlers", without a need to customise the general `onClick` function for every grid.

With all of this, the changes on the grid that should show the action column should be setting the appropriate parameter for the grid, defining the `getActions` function and appropriate action handlers.

Also one comment for the context grid change, when there is the edit icon in the action column, I'd say the link from the context name is redundant. Especially when there's the double-click hooked up to the inline editing, which forces you to double click the column but you can't aim at the context name as that'd redirect you to the edit page.

Ruslan-Aleev:
@theboxer Yes, it sounds great, I completely agree with you, I will be glad to see your PR :)
I did my best, working with ready-made solutions.
And the code is yes, it often repeats, for example, a similar solution - https://github.com/modxcms/revolution/blob/3.x/manager/assets/modext/workspace/package.grid.js#L234

theboxer:
I'm not saying there is no repetitive code in the current codebase. I'm suggesting to make it better, especially if this should be a new major release.

Ruslan-Aleev:
I understand this, I mean, if we make a universal solution, we need to remember to fix the repeating code in the current code base (not just in new), I gave it as an example.

JoshuaLuckers:
To prevent this to stagnate: who's going to move forward with this? 

@Ruslan-Aleev is this something you can do with the general outline given by @theboxer ?

Ruslan-Aleev:
My knowledge of js is not enough to implement the @theboxer proposals, so I will be glad to see his more competent PR :)

Ruslan-Aleev:
Although I would like to see a comment from @Mark-H , as he offered to accept my version earlier (I do not think that he did not take into account other options):

> Fair enough. I'd be happy to accept this PR with some minor changes and to look at the gear icon for a later PR.

itsjesterZ:
I've tested this on multiple browser on desktop, and on mobile, I have to say that it functions just fine on all browsers and is easy to use. 
The tests were done by using Browserstack on the browsers below.

- macOS (latest) Chrome/Firefox/Safari
- Windows (latest) Chrome/Firefox/Safari/Edge
- Android (Latest) Chrome
- iOS (Latest) Chrome/Safari

I do have an opinion on the buttons, wouldn't it be easier if you kept only the gear button with a sub menu for duplicate, update, and delete instead of having separate buttons for duplicate and update, or vice versa buttons for each function without the gear button.

Ruslan-Aleev:
@itsjesterZ Thanks for testing!
I do not quite agree with you. It is necessary not to hide frequent actions in the menu - for UX it is more useful (fewer clicks).

theboxer:
Here's the update: #14806

Mark-H:
Shouldn't be needed anymore with #14806, so closing. Please reopen if you don't agree. 

Thanks for the discussion and collaboration folks :)

PR #14797, opened 2019-10-16
Add more aliases

### LINKED ISSUE #14796: [3.x] Breaking extras with CMPs
I've tested few extras (all had custom manager pages) and none of them worked out of the box.

Changes required for making them work were easy - all needed to fix processors that were extending `abstract mod*Processor` and add the namespace to the extended processor; adjust the namespace for `modExtraManagerController`. After those changed everything (usually) started to work as expected.

This might be solve with aliasing abstract classes:
- modExtraManagerController
- mod*Processor

Other class worth of aliasing might be `modResource` as its `filterPathSegment` method is commonly used.

theboxer:
### What does it do?
Adds aliases for abstract processors, modExtraManagerController and modResource classes

### Why is it needed?
Fixes BC of many extras that are using CMPs or processors in general.

### Related issue(s)/PR(s)
Resolves #14796

### DIFF
--- a/core/include/deprecated.php
+++ b/core/include/deprecated.php
@@ -19,3 +19,20 @@ class_alias(\xPDO\Transport\xPDOTransport::class, \xPDOTransport::class);
 class_alias(\xPDO\Transport\xPDOObjectVehicle::class, \xPDOObjectVehicle::class);
 
 class_alias(\MODX\Revolution\modX::class, \modX::class);
+
+// Processors
+class_alias(\MODX\Revolution\modProcessor::class, \modProcessor::class);
+class_alias(\MODX\Revolution\modObjectProcessor::class, \modObjectProcessor::class);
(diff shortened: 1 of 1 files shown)


opengeek:
I suppose this brings up the question of where do we draw the line on what classes to alias.

JoshuaLuckers:
Unless they have been marked deprecated earlier I think it makes sense to class alias most of the "old" classes. 

Per class alias we should mark them deprecated using PHPDoc.
I also think it would be good to call `trigger_error` with as message type `E_USER_DEPRECATED` informing what class they should use instead of the old one.

Mark-H (approved):

Mark-H:
Adding `modManagerRequest` would also be helpful as I've had 2 sites throw a fatal error on trying to upgrade due to the context setting on the mgr context.

Issue #14564, opened 2019-04-18
MODX resources can be "saved" without a pagetitle


pmacswebteam:
### Summary
MODX resources can be "saved" without a pagetitle

### Step to reproduce
1. Create a resource with a "space" in the pagetitle field
1. Save the resource
1. Resource is automatically renamed to "Untitled Resource"
1. Removed "Untitled Resource" text & replace with a "space"
1. Save

### Observed behavior
Resource is saved with no pagetitle

### Expected behavior
User should be prompted to enter a pagetitle with real text (not just spaces) before being able to save.

### Environment
MODX version, apache/nginx and version, mysql version, browser, etc. Any relevant information.

MODX 2.7.1

Ruslan-Aleev:
The bug, as it turned out, is still relevant for **Templates, Categories and Files**. You can create them with spaces instead of a name (for a file there will be, for example: "   .txt").

Ruslan-Aleev:
For MODX3, this problem is also relevant, although for resources the bug works slightly differently.

smg6511:
This issue, among other validation-related ones, is resolved in my pull request #15146.

smg6511:
Please ignore the referenced pull above (#15146) ... I jumped the gun, so to speak, when replying that it addressed this particular issue. I removed my comment, but the pull reference remains.

smg6511:
OK, I'm going to jump in and work up a solution. One question: Should there _ever_ be an instance where a resource, element, or file name is allowed to begin with white space? My feeling is no. The answer to this question may inform how to go about making the change.

smg6511:
As I work on this, I find myself questioning the usefulness of auto-generating the "Untitled Resource" title and alias to begin with (at least in the way it currently happens). As I play with it (in both 2.x and 3.x), it seems to only apply when you happen to enter only whitespace into the pagetitle field. (Forgive me if I'm missing something in that regard.) As one would expect, an initially-empty field generates a validation error. So why would this not be the case for what is programatically-empty (i.e., only spaces)? Note that I realize there's a nuance in resource naming in that there's a use-case for allowing "0" as a pagetitle/alias, which is why the processors should test for null or an empty string rather than using php empty().

If the intention is to allow the saving of one or more untitled documents, then I'd argue it needs to be gone about in a different and more predictable way. Brainstorming it a bit, maybe ...

1. Create a boolean system setting, such as "Auto-Generate Page Titles." If this was set to "Yes" then...
[...]

PR #14800, opened 2019-10-19
Added check for multiple and single space instead of pagetitle

Ruslan-Aleev:
### What does it do?
In the current version, you can save a resource with one or several spaces instead of a pagetitle.
This PR fixes this if there are only spaces in the pagetitle - the pagetitle will become "Untitled Resource".

Deleted an unnecessary `$_lang['untitled_resource']` line (found only in the lexicon file), the rest of the code uses `$_lang['resource_untitled']`.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14564

### DIFF
--- a/core/lexicon/en/resource.inc.php
+++ b/core/lexicon/en/resource.inc.php
@@ -156,7 +156,6 @@ $_lang['symlink_help'] = 'The address of the object you wish to reference with t
 $_lang['symlink_message'] = 'A symlink is a symbolic link to another resource in your site which is forwarded to without changing the URL.';
 $_lang['symlink_new'] = 'New Symlink';
 $_lang['template_variables'] = 'Template Variables';
-$_lang['untitled_resource'] = 'Untitled Resource';
 $_lang['weblink'] = 'Weblink';
 $_lang['weblink_create'] = 'Create Weblink';
 $_lang['weblink_create_here'] = 'Weblink';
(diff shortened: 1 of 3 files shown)


JoshuaLuckers:
I consider it odd behaviour when we create a resource the page title being treated differently than when we update a resource.

Wouldn't it make more sense to align the way we handle the page title in 3.x when creating/updating a resource?

JoshuaLuckers:
@Mark-H @opengeek imo it makes more sense if we "prepared" the page title inside the model, what do both of you think?

sepiariver:
> @Mark-H @opengeek imo it makes more sense if we "prepared" the page title inside the model, what do both of you think?

+1 on comment by @JoshuaLuckers
My 2 cents is the processors have too much responsibility as it is, and the models don't know enough about how to validate themselves.

Ibochkarev:
@Mark-H @opengeek Guys, are there any progress on this PR?

Mark-H:
I can't reproduce the full behavior described in the related issue #14564 anymore since merging #15146 just now which tightens the validation and error handling further. A single space is still considered not empty and creates the "Untitled resource",  but trying to remove that no longer passes the validation to "unset" the pagetitle as reported, so I'm closing this PR and the related issue.

I'd be okay with preventing a single space from being accepted and turning into "Untitled resource" but don't have any strong feelings on that subject.

Issue #7695, opened 2012-05-02
User Profile: Firstname & lastname field instead of fullname


modxbot:
markh created Redmine issue ID 7695

Some applications reusing the existing security system require a separate first and last name, which, while possible by using extended fields or custom fields, is not as intuitive as it should be. 

I'd like to propose replacing the User Profile's fullname field with two separate fields for first and last name, and for backwards compatibility creating a fullname alias to firstname + lastname (or even better, a configurable way of concatenating first and lastname, such as "lastname, firstname").  This would allow more flexibility compared to what is offered now and make it easier to integrate with custom systems that expect them separately. 

With something like this we may need to consider the middle thingy in a name (don't think it's common in the US, but it is down here):

<pre>
firstname +(de|van (den|der)?|ten)+ lastname
</pre>

No idea what the English term for that is. It's tussenvoegsel in Dutchy.

JoshuaLuckers:
We should add fields for the first- and last name but also keep one for the fullname.
This makes it possible to use an alternative full name. First name: John, last name: Doe and full name J. Doe

JoshuaLuckers:
[To split or not to split?](https://www.w3.org/International/questions/qa-personal-names#singlefield)
> If designing a form or database that will accept names from people with a variety of backgrounds, you should ask yourself whether you really need to have separate fields for given name and family name .
> 
> This will depend on what you need to do with the data, but obviously it will be simpler, where it is possible, to just use the full name as the user provides it.

@Mark-H shared links with more information about this complex topic: https://github.com/modxcms/revolution/pull/14803#issuecomment-558621063).

After doing more research I don't think adding a first- and last name field to the core is a good idea.

SnowCreative:
Really, why was this closed? Having first and last name fields is such basic data. Without it, there is no way to sort by last name reliably, as Jay commented 8 years ago. I was just going to add a request to add this. What is the issue with adding this to the core? The extra "ExtraFields" was recently released that makes adding fields to the table possible, but I'd rather not have to depend on that.

JoshuaLuckers:
@SnowCreative did you read my last comment and the links with more information about this? 
Having a first- and last name field seems as something basic, but it actually isn't: https://www.w3.org/International/questions/qa-personal-names

Therefore I decided it's best if we keep the most simple form that works in all cases, a full name.

SnowCreative:
[...]

PR #14803, opened 2019-10-22
First- and last name fields for modUserProfile

JoshuaLuckers:
### What does it do?
It adds two fields to `modUserProfile`; 
1. `first_name` 
2. `last_name`

Also the corresponding input fields have been added so the fields can be set when creating/updating a user.

### Why is it needed?
As mentioned in issue #7695 
> This would allow more flexibility compared to what is offered now and make it easier to integrate with custom systems that expect them separately.

### Related issue(s)/PR(s)
Issue #7695


theboxer:
Wouldn't make sense to either remove the `fullname` field, or hide it and automatically compose it from those new `first_name` and `last_name` fields?

rthrash:
@theboxer how would you handle legacy data if removing?

theboxer:
Migration script and split the full name to first and last name ;)

rthrash:
What about middle names? Or people with three names (Tommy Lee Jones, Jennifer Love Hewitt…). Or people with titles or generational suffixes that are important to their identity? (Sr/Jr/III, Esq, PhD, MD, e.g., Robert Downey Jr)

theboxer:
Fair point, but without any change in the PR, how is your registration form going to look like? Please enter full name, first name and last name?

Even Google doesn't care about your  suffixes.
image

Middle name/Second name can go with the first name to the same column.

JoshuaLuckers:
Setting a first- and last name should be made possible in the first place. Once this PR is approved and merged we can share our thoughts about "what's next" in the original issue and/or new PR. 😄

theboxer:
That's the thing, as I'm not for merging the PR as is, cause it presents two new fields that have same functionality as the one we already have, so you'll have to maintain all three fields manually. That's why I suggested removing the fullname field, or leaving it as automated version that will return first_name + last_name.

I'm 100% for having separate fields for first & last name ;)

JoshuaLuckers:
> it presents two new fields that have same functionality as the one we already have

Depends; maybe someone wants to show his full name as:
- "given name" + "last name"
- "first name" + "last name"
- "initials" + "last name"
- etc.

Ruslan-Aleev:
I am also for hiding the "full name" field, but creating it dynamically from "first name" + "last name".
Now (with 3 name fields) some kind of excessive flexibility is obtained.

JoshuaLuckers:
My point being: Whatever is decided about the full name field shouldn’t prevent the first- and last name fields to be added.

JoshuaLuckers:
In my opinion the same applies on adding additional functionality to both fields. Once this PR is approved and merged others are able to work on that.

opengeek (changes requested):
I'm of the opinion that a full solution should be described and included as part of the PR before any change is made to the schema.

JoshuaLuckers:
I've checked the input fields for account creation by various services like Google, Facebook and several webshops. Most services only have inputs for the first- and last name. I think most of us agree on only adding those fields.

Storing a users name in the full name field should be discouraged by marking it deprecated. The field should be removed from the schema in MODX 4.

To get a full name we add a new get method `getFullName()` in `modUserProfile` that combines the first- and last name. The order in which both fields are combined can be determined by a parameter (e.g `$lastNameFirst = false`). If the full name field already has a value it will be returned instead of the first- and last name combination.

theboxer:
Few points:
- If you only introduce a new method to get a full name, extras and snippets that are using it will have to be updated, instead I'd suggest overriding the `get` method in the `modUserProfile` and if someone asks for `fullname` it'll return same result as the `getFullName` method you suggested.
- IMHO only discouraging and marking field as deprecated won't provide consistency, everyone should be forced to use these new fields, which will require a migration script to split the fullname into the first/last name fields; overriding the set method and fill first/last name when trying to set fullname. Without these steps you'll never will be able to count on that first_name is there (even though you know fullname is set) which will lead to devs keep using the fullname or creating own mechanics of getting the first/last name from the fullname.

JoshuaLuckers:
> * If you only introduce a new method to get a full name, extras and snippets that are using it will have to be updated, instead I'd suggest overriding the `get` method in the `modUserProfile` and if someone asks for `fullname` it'll return same result as the `getFullName` method you suggested.

Instead of overriding the `get` method for this specific use case I suggest a reusable and extendable solution. xPDOObject has field aliases, maybe it's worth adding some kind of virtual fields next to it.
When trying to access a virtual field field it will call a method based on a pattern or defined callable.

E.g: 
```
class User extends xPDOSimpleObject
{
 /**
     * An optional array of virtual fields.
     * @var array
     */
    public $_virtualFields = ['full_name'];

    protected function _getFullName()
    {
        return $this->first_name . '  ' . $this->last_name;
    }
}
```

> * IMHO only discouraging and marking field as deprecated won't provide consistency, everyone should be forced to use these new fields
I agree.

> which will require a migration script to split the fullname into the first/last name fields; overriding the set method and fill first/last name when trying to set fullname. 

I suggest the same solution as above, a virtual field with a `set` method based on a pattern or callable.

Mark-H:
https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/

Good guidance here as well: https://www.w3.org/International/questions/qa-personal-names

JoshuaLuckers:
Thanks for sharing @Mark-H! The information provided by those two links is very helpful. 

To be continued!

Mark-H:
We're now past the cut-off for breaking changes/new features for alpha1, so pushing milestone to 
beta.

JoshuaLuckers:
[To split or not to split?](https://www.w3.org/International/questions/qa-personal-names#singlefield)
> If designing a form or database that will accept names from people with a variety of backgrounds, you should ask yourself whether you really need to have separate fields for given name and family name .
> 
> This will depend on what you need to do with the data, but obviously it will be simpler, where it is possible, to just use the full name as the user provides it.

After doing more research (thanks for the links @Mark-H) I'm no longer in favor of having a first- and last name field in the core.

PR #14808, opened 2019-10-23
Add permissions to the languages list array for Language Toggle

### LINKED ISSUE #14755: MODX 3 - Menu items in manager missing "permissions" index
### Summary
While viewing any page in the manager the following is logged once for each menu item that exists:
```
[2019-09-24 17:51:37] (WARN @ /home/vagrant/code/modx3/public_html/manager/controllers/default/header.php : 314) PHP notice: Undefined index: permissions
```
So although this is only being logged as a warning, I assume this means the permissions are not being set for menu items.
https://github.com/modxcms/revolution/blob/3.x/manager/controllers/default/header.php#L314

### Step to reproduce
Set logging to 2 or above then load a page in the manager.

### Environment
MODX 3.0-dev ,PHP 7.3

theboxer:
### What does it do?
Adds `'permissions' => ''` to the languages sub menu when generating it.

### Why is it needed?
The menu handler is expecting `permissions` item in the menu entry to work properly. The generated language toggle submenu is missing this `permissions` item and throwing a warning.

### Related issue(s)/PR(s)
Resolves #14755

### DIFF
--- a/core/src/Revolution/modMenu.php
+++ b/core/src/Revolution/modMenu.php
@@ -106,6 +106,7 @@ class modMenu extends modAccessibleObject
                 'action' => 'language',
                 'params' => '&switch=' . $code,
                 'namespace' => 'core',
+                'permissions' => ''
             ];
         }
 


JoshuaLuckers:
Wouldn't it make more sense to solve this by checking if the `$menu` array has the `permissions` key?

theboxer:
That was other approach I was considering, but as menu has the `permission` in the model and languages submenu is generated artificially without the model, I thought this is a better approach.

JoshuaLuckers (approved):

Mark-H (approved):

Issue #14814, opened 2019-11-01
[modx3] Including composer.lock in the repository


Mark-H:
I just ran into an issue with a git installation of 3.x that was solved by running `composer update` - even though I'd already done `composer install` earlier, which installs the _locked_ versions.

This may have been previously touched on, and I seem to recall @opengeek not being too fond of including `composer.lock` into the repository, but I'm a little fuzzy on the details and want to invite people that use git installations to weigh in.

## Proposal: lock down dependencies by committing composer.lock

The benefit of locking down the composer.lock file is consistency. Without the composer.lock file, git installations are not guaranteed to run any particular version of any particular dependency. 

In my case today, `xpdo/xpdo` was out of date. But it can also go the other way, where a (potentially incompatible) newer version of a dependency is installed. 

### Pros

- Guaranteed version usage. Users should be encouraged to use `composer install` (or `composer install --no-dev`) after a `git pull`. 
- More control over supported versions. 

### Cons

- More work for integrators as new versions of dependencies will not be automatically available to people using a git installation; the "official" composer.lock file needs to be updated first. Users can of course do a manual `composer update` and resolve conflicts later in case of security issues or whatever. This can also be automated with a dependency bot that automatically sends pull requests whenever new versions come out. I'm using [dependabot](https://dependabot.com/) on some projects for that. 

## ...

Anything missing?

alroniks:
We use the composer.lock always in our extensions and deployments to guaranty that proper versions are used. So I vote for adding it.

opengeek:
https://github.com/modxcms/revolution/pull/14815

alroniks:
Seems the issue with 7.0 gone away and we could return to the question about including composer.lock into the repo.

opengeek:
I'm still against this. Different versions of PHP will load different dependencies. It's going to cause problems.

opengeek:
Not to mention if a security issue in a dependency comes out and could simply be updated via composer, any git-based installations will have to update the lock file, which will then be modified and could potentially cause problems updating MODX from git. Or it will require us to immediately release a new version because a dependency needs updating.

alroniks:
Then I suggest close it. I know and understand the reasons why not, we many times faced it in our project (my work), but one difference, that we manage almost all installations, so for us it's not a big problem. Feel free to reopen.

PR #14815, opened 2019-11-01
Commit composer.lock to the repository

opengeek:
### What does it do?
Adds composer.lock to the project to ensure consistent installation of dependencies.

### Why is it needed?
To ensure consistent installation of dependencies when installing from git.

### Related issue(s)/PR(s)
#14814

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,8 +48,6 @@ _build/templates/default/sass/font-awesome
 
 # ignore composer files
 composer.phar
-#composer.json
-composer.lock
 
 # ignore composer vendor dir
 /vendor
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # MODX Revolution
 
-[![Build Status](https://travis-ci.org/modxcms/revolution.svg?branch=2.x)](https://travis-ci.org/modxcms/revolution) [![Slack Chat](https://img.shields.io/badge/chat_in_slack-online-green.svg?longCache=true&style=flat&logo=slack)](https://modx.org)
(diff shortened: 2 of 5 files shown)


Mark-H:
It doesn't have to go into the build/distro, does it?

opengeek:
I would think it should. At the very least the composer.json needs to be there, and it wasn't.

.gitignore
```diff
@@ -49,7 +49,7 @@ _build/templates/default/sass/font-awesome
 # ignore composer files
 composer.phar
 #composer.json
-composer.lock
+#composer.lock
```
JoshuaLuckers: Maybe we can just remove the lines instead?

JoshuaLuckers (approved):

.travis.yml
```diff
@@ -10,7 +10,6 @@ php:
   - 7.3
   - 7.2
   - 7.1
-  - 7.0
```
JoshuaLuckers: Does this mean we make PHP 7.1 the minimum required version for MODX 3?

composer.json
```diff
@@ -43,7 +43,7 @@
         "vendor-dir": "core/vendor"
     },
     "require": {
-        "php": ">=7.0",
+        "php": ">=7.1",
```
JoshuaLuckers: Isn't it a better idea to have a separate commit/pr for the minimum required version bump? If 7.1 is the minimum required version we also have to change the version checks (for example in `manager/index.php`).

JoshuaLuckers (changes requested):

composer.json
opengeek: I had to do this so it passed the checks. 7.0 will no longer pass the unit tests.

JoshuaLuckers: Then we need to officially bump the minimum PHP version required for MODX 3.x to 7.1.

JoshuaLuckers:
@opengeek can we move forward with getting composer.lock in the repo?

opengeek:
Yeah, now 7.0 and 7.1 won't work with composer.lock committed. I'm going to say that we avoid committing this to the repo for now.

Issue #14822, opened 2019-11-11
Error "Class 'modManagerRequest' not found" after upgrading to MODX 3 nightly


JoshuaLuckers:
### Summary
When updating an existing installation of MODX (e.g: 2.7.2) to the latest nightly of MODX 3 it doesn't update the manager context setting `modRequest.class` from `modManagerRequest` to `MODX\Revolution\modManagerRequest`.

### Step to reproduce
1. Install MODX 2.7.2 
2. [Download the MODX 3 nightly](https://modx.com/download/nightlies) build on Oct 30, 2019
3. Go through the upgrade process using `/setup`
4. Once completed open the website and/or manager

### Observed behavior
An error is thrown: 
```
PHP Fatal error:  Uncaught Error: Class 'modManagerRequest' not found in /www/core/src/Revolution/modX.php:1490
Stack trace:
#0 ,/www/manager/index.php(58): MODX\Revolution\modX->getRequest()
#1 {main}
  thrown in ,/www/core/src/Revolution/modX.php on line 1490
```

The value for the mgr context setting `modRequest.class` isn't updated to the right value when updating.

### Expected behavior
Updating an existing installation should also update context settings with the new value:
https://github.com/modxcms/revolution/blob/b3bfde44d42d17064e139699e9059e5de1fc7b13/_build/data/transport.core.context_settings.php#L18

### Environment
MODX 3 nightly build on Oct 30, 2019.

PR #14826, opened 2019-11-15
Modify core class references in modContextSettings

JoshuaLuckers:
### What does it do?
Modify core class references in modContextSettings

### Why is it needed?
This solves the error "Class 'modManagerRequest' not found" after upgrading to MODX 3

### Related issue(s)/PR(s)
Resolves issue #14822

### DIFF
--- a/setup/includes/upgrades/common/3.0.0-update-legacy-class-references.php
+++ b/setup/includes/upgrades/common/3.0.0-update-legacy-class-references.php
@@ -17,7 +17,9 @@ use MODX\Revolution\Hashing\modMD5;
 use MODX\Revolution\Hashing\modNative;
 use MODX\Revolution\Hashing\modPBKDF2;
 use MODX\Revolution\modAccess;
+use MODX\Revolution\modContextSetting;
 use MODX\Revolution\modDocument;
+use MODX\Revolution\modManagerRequest;
 use MODX\Revolution\modResource;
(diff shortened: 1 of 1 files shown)


setup/includes/upgrades/common/3.0.0-update-legacy-class-references.php
```diff
@@ -78,3 +80,9 @@
 $modx->updateCollection($class, ['class_key' => modFileMediaSource::class], ['class_key' => 'sources.modFileMediaSource']);
 $modx->updateCollection($class, ['class_key' => modFTPMediaSource::class], ['class_key' => 'sources.modFTPMediaSource']);
 $modx->updateCollection($class, ['class_key' => modS3MediaSource::class], ['class_key' => 'sources.modS3MediaSource']);
+
+/* modify core class references in modContextSettings */
+$class = modContextSetting::class;
+$table = $modx->getTableName($class);
+
+$modx->updateCollection($class, ['value' => modManagerRequest::class], ['key' => 'modRequest.class']);
```
Mark-H: Should this, like the other examples above, only update where the _value_ is `modManagerRequest` rather than always change the `modRequest.class` setting?

JoshuaLuckers: Yes, you are right!

PR #14827, opened 2019-11-15
Port modPackageBuilder, modTransportManager and modTransportVehicle

theboxer:
### What does it do?
Ports modPackageBuilder, modTransportManager and modTransportVehicle classes from 2.x to 3.x

### Why is it needed?
They were missing and they are used in a lots of build scripts :)


alroniks (changes requested):
File names do not compatible to PSR-4. Could we avoid `.class` postfix here?

theboxer:
Ouch, sorry. Will update it

theboxer:
Adjusted

JoshuaLuckers (changes requested):
To keep the code consistent with the rest of the code base please apply the PSR-2 code style.

theboxer:
Tried my best :)

JoshuaLuckers:
> Tried my best :)

Thanks, if you could also add the copyright notice that would be great! 😄

opengeek (approved):

PR #14830, opened 2019-11-19
Added element name to browser title

Ruslan-Aleev:
### What does it do?
Added element name in  browser title for:
- Media sources;
- Dashboard widgets and widgets themselves;
- Profiles customization forms.

In the remaining sections, the names are in the title.
And with titles in title it’s easier to navigate in browser tabs.

Before
before

After
after

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/controllers/default/security/forms/profile/update.class.php
+++ b/manager/controllers/default/security/forms/profile/update.class.php
@@ -96,7 +96,7 @@ class SecurityFormsProfileUpdateManagerController extends modManagerController {
      * @return string
      */
     public function getPageTitle() {
-        return $this->modx->lexicon('form_customization');
+        return $this->modx->lexicon('form_customization').': '.$this->profileArray['name'];
     }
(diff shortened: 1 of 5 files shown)


Ibochkarev (approved):

JoshuaLuckers (approved):

Mark-H:
Should be checked for security - I _think_ the `<title>` is already escaped in the smarty template but am not sure. Would not want somebody entering a `</title><script>alert('xss');</script>` name on a media source or something to work.

Ruslan-Aleev:
I tried `<script>alert ('xss');</script>` in the name, in the code and on the page everything is fine, the window does not appear.
And I used the existing name output code in title, for example:
https://github.com/modxcms/revolution/blob/3.x/manager/controllers/default/element/template/update.class.php#L166

Mark-H (approved):

PR #14832, opened 2019-11-20
Changed resource styles in the tree

Ruslan-Aleev:
### What does it do?
Changed the resource styles in the tree, made them more logical:
- Unpublished - the most faded in the tree;
- Hidden from the menu - font-family: italic;

### Why is it needed?
Unpublished resources look brighter in the resource tree than hidden resources from the menu - which is not logical.
Perhaps this only bothers me, but I'm confusing the current style version.

Now there is no such problem.
res

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13969#issuecomment-450426844

### DIFF
--- a/_build/templates/default/sass/_colors-and-vars.scss
+++ b/_build/templates/default/sass/_colors-and-vars.scss
@@ -145,13 +145,14 @@ $treeItemBg: $white;
 $treeItemColor: $mediumGray;
 
 $iconColor: $treeText;
-$unpublished: $treeText;
 
 $deactivatedTextColor: lighten($treeColor, 35%); // do not use 50% here - makes base color nearly invisible
 $disabledTextColor: $deactivatedTextColor;
 
-$hidden: lighten($treeText, 10%) !important;
-$unpubText: italic;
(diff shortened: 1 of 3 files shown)


JoshuaLuckers:
Is it ready for review?

Ruslan-Aleev:
Yes, ready.

JoshuaLuckers:
Before: 
Screenshot 2019-11-27 at 14 12 14

After: 
Screenshot 2019-11-27 at 14 12 19

JoshuaLuckers (approved):

Mark-H:
I could get used to this pattern - dimmed for unpublished resources that aren't visible, and italics for menu visibility. Probably easier for users.

One problem, [the contrast ratio for unpublished elements is only 2.1](https://webaim.org/resources/contrastchecker/?fcolor=B3B2B2&bcolor=FFFFFF) while for a baseline of WCAG AA accessible a ratio of 4.5 is needed. It looks like the colors are the same as before, so this doesn't make it _worse_, but it would be nice if in the future we could make this _better_...

Mark-H (approved):

PR #14836, opened 2019-11-21
Change the value of default_media_source_type system setting

### LINKED ISSUE #14818: Source type of Create New Media Source
### Summary
When creating a new Media Source the source type is "sources.modFileMediaSource". Now it confused
image

### Step to reproduce
Create a new Media Source 

### Environment
MODX 3

theboxer:
### What does it do?
Changes the default value of `default_media_source_type` system setting.

### Why is it needed?
When you try and add new media source, the default value is set to `sources.modFileMediaSource` which is an old one from 2.x.

### Related issue(s)/PR(s)
Resolves #14818


JoshuaLuckers (commented):
Wouldn't it be better to use the value returned from the special `::class` constant when updating the value for `default_media_source_type`?

PR #14839, opened 2019-11-24
Set autocomplete to off on username in user edit form (for 3.x)

### LINKED ISSUE #14699: Autocomplete on Username Field in User Edit Form
### Summary
Autocomplete in the username field in the user edit form can lead to accidental saving of an incorrect username from that originally stored when the user hasn't been assigned a password - for example when imported without one.

### Step to reproduce
Create a user via the API without a password on a domain you have an autocomplete username for.

### Observed behavior
For users with no password, on Chrome the username is filled in with the first entry from my autocomplete list replacing the original username.

### Expected behavior
It should retain the user's actual username.

### Environment
MODX version 2.7.1-pl, apache 2.4, mysql 5.5, Chrome (tested on Linux, Mac & Windows.

I'm unclear as to the benefit of having autocomplete on any of the fields in the manager.
I see #6703 for the removal of autocomplete for the title field of resources too but this issue could be a real nuisance for unwary site admins.

theboxer:
### What does it do?
Set autocomplete to `off` on username field.

### Why is it needed?
To prevent browsers changing username through autocomplete.

### Related issue(s)/PR(s)
Resolves #14699

### DIFF
--- a/manager/assets/modext/widgets/security/modx.panel.user.js
+++ b/manager/assets/modext/widgets/security/modx.panel.user.js
@@ -491,6 +491,7 @@ Ext.extend(MODx.panel.User,MODx.FormPanel,{
             ,description: _('user_username_desc')
             ,xtype: 'textfield'
             ,anchor: '100%'
+            ,autoCreate: {tag: "input", type: "text", size: "20", autocomplete: "off", msgTarget: "under"}
             ,listeners: {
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

rthrash:
This should probably also go into a 2.7.x when another one is released.

opengeek:
@rthrash see #14838

Issue #14539, opened 2019-04-07
Need refactoring of system settings


Ruslan-Aleev:
### Summary
In the "System Settings" a lot of settings initially. But besides this, there are also settings that are not used in the system at all, but hang in the list.

Those we need refactoring of "System Settings" section:

1) It is necessary to remove outdated system settings.
2) Some settings are not in the right categories, for example:
**publish_default** is located in the "Site" section, but the **default_duplicate_publish_option** setting is located in the "Back-end Manager" section, although, in my opinion, everything related to resources should be in the "Site" section.
**log_snippet_not_found** setting should be in the "Back-end Manager" section.
**default_media_source** and **default_media_source_type** settings should be in the "File System" section etc.
Need to check the categories of system settings.

webinmd:
These settings are not used anywhere in the code.
Most likely this is the legacy of Evo (WebSignup & WebLogin)

- webpwdreminder_message
- websignupemail_message

Ruslan-Aleev:
### In my opinion, these settings do not work (maybe i'm wrong):

**~~Authentication and Security:~~** (https://github.com/modxcms/revolution/pull/14841)
- ~~udperms_allowroot (found only in "System Settings")~~
- ~~webpwdreminder_message (found only in "System Settings")~~

**~~Rich-Text Editor:~~** (https://github.com/modxcms/revolution/pull/14843)
- ~~editor_css_path (found only in "System Settings")~~
- ~~editor_css_selectors (found only in "System Settings")~~

**~~File System:~~** (https://github.com/modxcms/revolution/pull/14886)
- ~~filemanager_path (although it is indicated as deprecated, the setting is found in the files)~~
- ~~filemanager_path_relative (found only in "System Settings")~~
- ~~filemanager_url (although it is indicated as deprecated, the setting is found in the files)~~
- ~~filemanager_url_relative (found only in "System Settings")~~
- ~~rb_base_dir (?)~~
- ~~rb_base_url (?)~~
- ~~strip_image_paths (found only in "System Settings")~~
- ~~use_browser (found only in "System Settings")~~

**~~Lexicon and Language:~~** (https://github.com/modxcms/revolution/pull/14848)
- ~~fe_editor_lang (found only in "System Settings")~~

**~~Back-end Manager:~~** (https://github.com/modxcms/revolution/pull/14868)
- ~~manager_js_cache_file_locking (found only in "System Settings")~~
- ~~manager_js_cache_max_age (found only in "System Settings")~~
[...]

PR #14843, opened 2019-11-26
Removed unused system settings for "Rich-Text Editor" section

Ruslan-Aleev:
### What does it do?
Removed unused system settings for "Rich-Text Editor" section.

- editor_css_path
- editor_css_selectors

### Related issue(s)/PR(s)

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -478,24 +478,6 @@ $settings['default_content_type']->fromArray(array (
   'area' => 'site',
   'editedon' => null,
 ), '', true, true);
-$settings['editor_css_path']= $xpdo->newObject(modSystemSetting::class);
-$settings['editor_css_path']->fromArray(array (
-  'key' => 'editor_css_path',
-  'value' => '',
-  'xtype' => 'textfield',
-  'namespace' => 'core',
-  'area' => 'editor',
-  'editedon' => null,
-), '', true, true);
-$settings['editor_css_selectors']= $xpdo->newObject(modSystemSetting::class);
-$settings['editor_css_selectors']->fromArray(array (
-  'key' => 'editor_css_selectors',
-  'value' => '',
-  'xtype' => 'textfield',
-  'namespace' => 'core',
-  'area' => 'editor',
-  'editedon' => null,
-), '', true, true);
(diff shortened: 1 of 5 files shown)


Mark-H:
I think the old tinymce package may use this? 

That's obviously a poor reason to keep it in core, but if that's the case it's a good idea to make sure the package is updated to use a different setting first.

Ibochkarev (approved):

Ruslan-Aleev:
> I think the old tinymce package may use this?

@Mark-H I searched for this setting on github, there were no direct matches, only in MODX, as it seems to me.

Ruslan-Aleev:
The old version of tinyMCE will not work with MODX 3, in any case. After all, the version of the editor after the release of MODX 3 will have to be updated for compatibility with MODX 3, and it would be disadvantageous to update the old version tinyMCE in any case.

JoshuaLuckers:
> I think the old tinymce package may use this?
> 
> That's obviously a poor reason to keep it in core, but if that's the case it's a good idea to make sure the package is updated to use a different setting first.

Seems like something that should be documented in the docs [(Upgrading MODX)](https://docs.modx.org/current/en/getting-started/maintenance/upgrading).

Mark-H:
[I've opened an issue for TinyMCE](https://github.com/modxcms/TinyMCE/issues/30) and [for the 3.0 upgrade documentation](https://github.com/modxorg/Docs/pull/125), but am wondering if perhaps we should make a special exception for these settings and _don't actively remove them_ upon upgrade? 

While that would leave an orphaned setting, that setting could potentially contain information that's still relevant for people's site. 

On the other hand... that's a really old extra and I don't see any documentation explaining users how to use it, so we could also just go ahead with this and let Tiny figure it out. I'm a little conflicted, so opinions are welcome.

JoshuaLuckers:
Maybe remove the settings only when the value equals the default install value?

Mark-H (approved):
I've decided it can't hurt to force tinymce to update for the first time in a few years while cleaning up the core.

donShakespeare:
If no editor is the default editor for MODX there might be no planetary reason for keeping a specific strictly-related editor setting lying around in the MODX core.

Wise decision!

PR #14847, opened 2019-11-27
delete copies of the lexicon

CrazyBoy49z:
### What does it do?
delete copies of the lexicon key setting_passwordless_activated and setting_passwordless_activated_desc

### Why is it needed?
https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/setting.inc.php#L845-L846

https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/setting.inc.php#L851-L852

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -847,6 +847,3 @@ $_lang['setting_passwordless_activated_desc'] = 'When enabled, users will enter 
 
 $_lang['setting_passwordless_expiration'] = 'Passwordless login expiration';
 $_lang['setting_passwordless_expiration_desc'] = 'How long a one-time login link is valid in seconds.';
-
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Good catch! 

There are other duplicates related to the passwordless login:
- https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/setting.inc.php#L848-L849
- https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/setting.inc.php#L854-L855

JoshuaLuckers (approved):

Ibochkarev (approved):
Good find.

CrazyBoy49z:
> Good find.

When translating into Ukrainian, the same key came across

Mark-H (approved):

PR #14848, opened 2019-11-27
Removed unused system settings for "Lexicon and Language" section

Ruslan-Aleev:
### What does it do?
Removed unused system settings for "Lexicon and Language" section.

- fe_editor_lang

### Related issue(s)/PR(s)

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -532,15 +532,6 @@ $settings['failed_login_attempts']->fromArray(array (
   'area' => 'authentication',
   'editedon' => null,
 ), '', true, true);
-$settings['fe_editor_lang']= $xpdo->newObject(modSystemSetting::class);
-$settings['fe_editor_lang']->fromArray(array (
-  'key' => 'fe_editor_lang',
-  'value' => 'en',
-  'xtype' => 'modx-combo-language',
-  'namespace' => 'core',
-  'area' => 'language',
-  'editedon' => null,
-), '', true, true);
 $settings['feed_modx_news']= $xpdo->newObject(modSystemSetting::class);
 $settings['feed_modx_news']->fromArray(array (
   'key' => 'feed_modx_news',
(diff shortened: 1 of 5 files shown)


Mark-H:
Rather than adding a bunch of different upgrade scripts, I think it's fine to add any extra settings that are removed into `setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php` - the code in each "setting cleanup" migration script is practically the same.

Ruslan-Aleev:
I agree with this, I even discussed it with @JoshuaLuckers , we decided to make it different files:

> If the removal of some system settings require diffrent migrations its easier to have them in separate files. Having a lot of files is not an issue

I think in the next PR I will combine the settings for deletion into one file.

Ruslan-Aleev:
@Mark-H there is another question.
Some files that are affected in this PR are infused after - https://github.com/modxcms/revolution/pull/14849
But why doesn’t the conflict appear in this PR or something I don’t understand?

Mark-H:
Sorry, where did you discuss that with Joshua? I must've missed that. While we need different files for different versions (i.e. 3.1.0 will require another file) this seems like unnecessary duplication. But if that was already discussed maybe I'm missing some good arguments for separated files.

Git is pretty good at avoiding conflicts most of the time, only when it can't automatically figure out how to deal with multiple changes to the same file you'll see a conflict. ;)

Ruslan-Aleev:
We discussed this in Slack :) Anyway, I think in the next PR I will combine the settings for deletion into one file.

Mark-H:
Ahh okay. ;)

JoshuaLuckers:
Via a direct message on Slack, Ruslan asked me:
>  isn’t it right to place all the settings in one file (indicating all the keys), and remove the extra files
> Example:
- setup/includes/upgrades/common/3.0.0-remove-upload-flash-system-setting.php
- setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
- setup/includes/upgrades/common/3.0.0-remove-tv-eval-system-setting.php

My reaction: 
> for now it's ok to have them in different files, it's ok to do it for your PR as well.
> If the removal of some system settings require different migrations its easier to have them in separate  files. Having a lot of files is not an issue
> We can always merge them into one eventually

**My reasoning:**
More settings might be removed before 3.0 is released, some of them might need additional checks ([like this one maybe](https://github.com/modxcms/revolution/pull/14843#issuecomment-559869657)). Having separate files for now makes it easier for the PR author to add those additional checks and migrations. 

**Conclusion**
Yes, having them in one file is better. Separate files cause unnecessary duplication. But it can't harm to have them in separate files.

Mark-H (approved):

Mark-H:
It's probably easier (for settings that don't require additional migrations) to put them in the same file, so you don't also have to duplicate the include in each database driver folder... but yeah, no real harm in it this way.

PR #14855, opened 2019-11-29
Reorganized the main menu

Ruslan-Aleev:
### What does it do?
Changed the order of the main menu items.
The items from the "Manage" section have moved to other logically related sections and a new "Access" item has been added and that's it.

In principle, we can leave the "Manage" section, where there will be "Clear cache", "Delete Locks", "Manager Action" and "Error log" (the "Site Schedule" item will move to "Content", "Reports -> System Info" item will move to "System").

**BUT the addition of the "Access" item seems to me to be a very useful improvement to the menu.**

### What does it look like now:

**All menu**
menu

---

**Content**
content

---

**Extras**
extras

---

**Account**
user
The language switch is now in the system section. The language switch can be left in "Account" section, but it seems to me that this is more about the "System" than the "Account".

---

**Access**
access
All items related to users and access are moved to one section.

---

**System**
system_1
system_2

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14291

### DIFF
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -12,7 +12,7 @@ use MODX\Revolution\modMenu;
 
 $menus = [];
 
-$menus[0]= $xpdo->newObject(modMenu::class);
+$menus[0] = $xpdo->newObject(modMenu::class);
 $menus[0]->fromArray([
     'menuindex' => 0,
     'text' => 'topnav',
@@ -22,9 +22,9 @@ $menus[0]->fromArray([
     'action' => '',
 ], '', true, true);
 
-$menus[1]= $xpdo->newObject(modMenu::class);
+$menus[1] = $xpdo->newObject(modMenu::class);
 $menus[1]->fromArray([
-    'menuindex' => 0,
+    'menuindex' => 1,
     'text' => 'usernav',
     'description' => 'usernav_desc',
     'parent' => '',
@@ -34,7 +34,7 @@ $menus[1]->fromArray([
 
 
 /* ***************** CONTENT MENU ***************** */
-$topNavMenus[0]= $xpdo->newObject(modMenu::class);
+$topNavMenus[0] = $xpdo->newObject(modMenu::class);
 $topNavMenus[0]->fromArray([
   'menuindex' => 0,
   'text' => 'site',
@@ -42,13 +42,13 @@ $topNavMenus[0]->fromArray([
   'parent' => 'topnav',
   'permissions' => 'menu_site',
   'action' => '',
-  'icon' => '<i class="icon-file-text-o icon icon-large"></i>',
+  'icon' => '<i class="icon-file-text-o icon"></i>',
 ], '', true, true);
 
 $children = [];
 
 /* New Resource */
-$children[0]= $xpdo->newObject(modMenu::class);
+$children[0] = $xpdo->newObject(modMenu::class);
 $children[0]->fromArray([
   'menuindex' => 0,
   'text' => 'new_resource',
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
It would be very helpful (for documentation and reviewing) to have comparison between the old and new menu "structure" with an overview of the items that have been moved (and to where), deleted and added.

Mark-H:
Not sure if I agree with all these changes. In particular the "Packages" ("Extras") menu - that's where installed extras go, and I don't think settings/lexicons/namespaces should also be there. Settings/lexicons/namespaces are administrator features, while (most) extras are meant for editors to use. I get why you'd want to highlight where settings for extras are, but don't think we should mix up admin/editor menu items too much.

I do like the introduction of an "Access" menu, but I'd probably prefer that as a submenu of system (gear icon). Again, that's admin functionality. 

Not sure how I feel about the "Manager" menu yet.. 

> There are inaccuracies in the current menu, and I would not want users to get used to the menu, which may change in the future.

As a counter point to that; people are already used to the current menu as well and there's ton of documentation that refers to, for example, "Settings > Access Control Lists". If we change menu locations, that will also need to be updated or it will cause additional confusion.

Ruslan-Aleev:
> Settings/lexicons/namespaces are administrator features, while (most) extras are meant for editors to use. I get why you'd want to highlight where settings for extras are, but don't think we should mix up admin/editor menu items too much.

On the one hand, I agree with this, but on the other not :) Changing the theme of a text editor is a function of the editor or administrator? If it’s an administrator, then the ability to install packages is also an administrator’s function, and therefore the entire "Packages" section is an administrator’s function.

The sections **"Access" and "Manager"** can be moved down, as in the image below. I see no reason to make them nested in the "System" section (gear), this will only add unnecessary clicks, there enough place in menu.

By the way, it may be worthwhile for the bottom menu to also display the text of the item, and not just icons.

bottom_m

> As a counter point to that; people are already used to the current menu as well and there's ton of documentation that refers to, for example, "Settings > Access Control Lists". If we change menu locations, that will also need to be updated or it will cause additional confusion.

Yes, I completely agree with this point.

Ruslan-Aleev:
@opengeek @rthrash @Jako what do you think?

JoshuaLuckers:
Where I can find a particular item is always confusing me, especially the items under "Manage" and "Settings" (gear icon). 
Some items that are now located under "Settings" also indicate they are to *manage* something (e.g: lexicons, property sets, namespaces).

JoshuaLuckers:
It would be nice that if we decide to move/change the menu items to gather feedback during the beta period (with an option to revert back to the "old" menu).

Ruslan-Aleev:
> It would be nice that if we decide to move/change the menu items to gather feedback during the beta period (with an option to revert back to the "old" menu).

Yes, it would be helpful. The question is how to get this feedback from users?

Ruslan-Aleev:
Can we display a widget in the alpha version with a link to a google-poll, where we can clarify various questions, including the menu (with a link to issue / PR, etc.)? It would be useful to get statistics and understand specific positive and negative points from users.

Mark-H:
This is past the cut-off for breaking changes/new features in the alpha1 build, so assigned to beta. If this warrants more discussion or research I'd suggest doing that somewhere else (perhaps on the forum or with slack).

JoshuaLuckers:
@Ruslan-Aleev it would be awesome if you could make a table with the current and suggested menu structure and post it on the [MODX community](https://community.modx.com/c/ux-and-feedback) for discussion.

_build/data/transport.core.menus.php
```diff
@@ -33,224 +33,250 @@
 ), '', true, true);
 
 
-/* ***************** CONTENT MENU ***************** */
+/* ********** CONTENT MENU ********** */
```
mishanthrop: MODX size reducing :)

_build/data/transport.core.menus.php
```diff
@@ -33,224 +33,250 @@
...
-  'description' => '',
-  'parent' => 'topnav',
-  'permissions' => 'menu_site',
-  'action' => '',
+    'menuindex' => 0,
+    'text' => 'site',
+    'description' => '',
+    'parent' => 'topnav',
+    'permissions' => 'menu_site',
+    'action' => '',
```
mishanthrop: Useless changes

rthrash:
I'm a big fan of the thought that went into this PR—it's really well-reasoned, so thank you—but want to consider the impact quite carefully as you suggested. For example, Resource Groups, in my mind, probably belongs under Access. 

I could also see separating the Installer from the Extras menu entirely … that menu just becomes a place where installed Extras live. The installer would go in the system/gear/cog menu, and if no extras were installed, a link to the MODX Extras on the web could live (or a copy of the Installer menu … though again, you don't want Bill the copy editor installing `phpThumbOf Cache Manager` … an alternative blank state could be something like, "Ask your Site Admin for awesome Extras from the <link>MODX Extras repository</link>…").

Import menus items … feels like a good opportunity to push to Extras while we can. Less to maintain in the core.

theboxer:
> I could also see separating the Installer from the Extras menu entirely

Exactly what I wanted to point out :) However, if no extras are installed I'd vote for hiding that Extras menu.

Ruslan-Aleev:
> I could also see separating the Installer from the Extras menu entirely … that menu just becomes a place where installed Extras live.

The logic is clear here, and we even partially discussed this question above.
But to move the "Package Installer" from the "Packages" section is very strange for users. There will be so many questions and indignation :)

Need to discuss.

ghost:
Having the installer in the "Extras" menu is logical and intuitive, but it must be configured to always be at the top of the list.

Ruslan-Aleev:
@JoshuaLuckers @Mark-H @theboxer @rthrash Guys, so where did we end up? What's the plan with this PR?
Also there is a discussion of the menu in https://github.com/modxcms/revolution/issues/14291 And the latest version of the menu, based on the discussion - https://github.com/modxcms/revolution/issues/14291#issuecomment-459953084

JoshuaLuckers:
> @JoshuaLuckers @Mark-H @theboxer @rthrash Guys, so where did we end up? What's the plan with this PR?
> Also there is a discussion of the menu in #14291 And the latest version of the menu, based on the discussion - [#14291 (comment)](https://github.com/modxcms/revolution/issues/14291#issuecomment-459953084)

Hi Ruslan, my idea is to gather community input during the Snowup next week so we can move forward with this!

Ruslan-Aleev:
@JoshuaLuckers A good idea! Thank :)

Ruslan-Aleev:
@JoshuaLuckers @Mark-H @rthrash Guys, I significantly simplified this PR, the changes are now not as significant as they were before.
In fact, the items from the "Manage" section have moved to other logically related sections and a new "Access" item has been added and that's it (see PR description).

I took into account almost all the previous comments, in my opinion this variant is more understandable.

rthrash:
Have screenshots with the new version by chance? Looking forward to seeing it. :)

Ruslan-Aleev:
@rthrash I changed the PR description with new screenshots :)

PR #14858, opened 2019-11-29
Rename and move the base Object-related processors

Mark-H:
### What does it do?

This modernises the structure and naming of the base processors, inline with the earlier massive refactor of all implementing processors. While this commit technically breaks all third party processor compatibility - that has already been broken anyway. 

By doing this refactor now, we can prepare the documentation _before_ the first alpha appropriately and benefit from the simplified naming in the future without adding another set of breaking changes down the road.

Also some minor/code style issues have been resolved (objects always passed by reference, strict bool checks, base DuplicateProcessor not marked as abstract, wrong classes being called in database table processors).

Brought you by PhpStorm's a-bloody-mazing refactoring capabilities.

### Why is it needed?

While great strides have been made, I think we can do even better with modernising the codebase further. This is one of the things that can help bring greater clarity and code separation to the core, by making sure all processor related classes are in one specific, logical, place. 

### Related issue(s)/PR(s)

N/a.

### DIFF
--- a/_build/test/MODxTestCase.php
+++ b/_build/test/MODxTestCase.php
@@ -11,6 +11,7 @@
 */
 namespace MODX\Revolution;
 
+use MODX\Revolution\Processors\ProcessorResponse;
 use PHPUnit\Framework\TestCase;
 use xPDO\xPDOException;
 
@@ -54,18 +55,18 @@ abstract class MODxTestCase extends TestCase {
     /**
      * Check a MODX return result for a success flag
      *
-     * @param modProcessorResponse $result The result response
+     * @param ProcessorResponse $result The result response
      * @return boolean
      */
     public function checkForSuccess(&$result) {
-        if (empty($result) || !($result instanceof modProcessorResponse)) return false;
+        if (empty($result) || !($result instanceof ProcessorResponse)) return false;
         return !$result->isError();
     }
 
     /**
      * Check a MODX processor response and return results
      *
-     * @param modProcessorResponse $result The response
+     * @param ProcessorResponse $result The response
      * @return array
      */
     public function getResults(&$result) {
--- a/_build/test/Tests/Processors/Browser/DirectoryTest.php
+++ b/_build/test/Tests/Processors/Browser/DirectoryTest.php
@@ -13,7 +13,7 @@ namespace MODX\Revolution\Tests\Processors\Browser;
 
 
 use Exception;
-use MODX\Revolution\modProcessorResponse;
+use MODX\Revolution\Processors\ProcessorResponse;
 use MODX\Revolution\modX;
 use MODX\Revolution\MODxTestCase;
 use MODX\Revolution\Processors\Browser\Directory\Create;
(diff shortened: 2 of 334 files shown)


core/src/Revolution/Processors/Object/CreateProcessor.php
```diff
@@ -8,9 +8,11 @@
  * file that was distributed with this source code.
  */
 
-namespace MODX\Revolution;
+namespace MODX\Revolution\Processors\Object;
```
JoshuaLuckers: I would refrain from using [reserved words](https://www.php.net/manual/en/reserved.other-reserved-words.php)(object; as of PHP 7.2) in namespaces.
> Reserved words cannot be used to name a class, interface or trait, and they are also prohibited from being used in namespaces.

Mark-H: That... does seem like a problem. I'm surprised phpstorm isn't flagging that.. 

It's getting late and my brain is fairly fried, but I can think of 2 alternatives.

1. Rename the namespace to `xPDOObject` (e.g. `\MODX\Revolution\Processors\xPDOObject\CreateProcessor`), so you still benefit from the clean `class MyProcessor extends CreateProcessor`. 
2. Move up one directory/namespace and reintroduce `Object` into the classname (e.g. `\MODX\Revolution\Processors\ObjectCreateProcessor`). Still some benefit from pulling processors out of `\MODX\Revolution`, but not quite as neat in its own directory. 

Good catch!

JoshuaLuckers: > That... does seem like a problem. I'm surprised phpstorm isn't flagging that..

It's possible, but prohibited, to use them.

> It's getting late and my brain is fairly fried, but I can think of 2 alternatives.
> 
> 1. Rename the namespace to `xPDOObject` 
> 2. Move up one directory/namespace and reintroduce `Object` into the classname 

I have a third alternative: rename `Object` to `Model` (e.g. `\MODX\Revolution\Processors\Model\CreateProcessor`) 

> Good catch!

Thanks! 😄

Mark-H: Updated to use `Model` - that's a pretty good alternative.

Mark-H:
As an aside, I also have another PR coming that's based off this branch that removes support for flat-file processors and generally refactors `modX::runProcessor`. [Previewable here](https://github.com/Mark-H/revolution/pull/9), will send it upstream when this one is merged.

JoshuaLuckers (approved):

opengeek:
I may have misunderstood the scope of what you were proposing here. I think, by doing this, we have created a much larger BC break because now ALL processors, regardless if they extend core processors, will have to be refactored to work in 3.x. Correct me if I am wrong, but before this change, any custom processor that just used modProcessor and modProcessorResponse would still work when referencing the legacy class names, no?

Mark-H:
Yeah, those are added as aliases. I've not done super extensive testing, but the extras I had installed continued to work. 

All processor will need to be refactored regardless of this change - at least when the class alias file is no longer automatically loaded which we talked about for v3.3 or so.

opengeek (approved):
Oh, I gotcha. Fair enough. Approved!

Issue #14744, opened 2019-09-19
MODX3: Make links in the manager in one format


Ruslan-Aleev:
### Summary
In the current version of MODX 3, the manager has links of various formats, for example:
`/manager/?a=element/template/update&id=1`
and
`/manager/?a=Context/Update&key=web`

It seems to me that we need to bring the links to the same format (in lower case, for example) so that:
- to avoid possible bugs in the future, like https://github.com/modxcms/revolution/pull/14662
- maintain visual compliance

### Environment
MODX 3

Issue #13639, opened 2017-10-16
2.6.0-dev: Header for Chunks: wrong format


achterbahn:
### Summary
Title for chunks is wrongly formated

### Step to reproduce
Create a chunk and compare the title to a snippet/TV/... 

### Observed behavior
`<div id="modx-chunk-header" class=" panel-desc" style="height: auto;">Chunk: test</div>`

### Expected behavior
`<h2 id="modx-chunk-header" class=" modx-page-header" style="height: auto;">Chunk: test</h2>
`
### Environment
2.6.0-dev

PR #14864, opened 2019-12-02
Added item links to grids

Ruslan-Aleev:
### What does it do?
- Added links to the item in the list, in grids.
Now you can go inside the element without calling the context menu.
- Also linked items with links in grids that are not directly related (opens in new tab).
- Cleaned the code: fixed tabs, removed extra spaces, added comments.
In some grids, I changed the order of functions to a more logical one (for example, **before:** createContext, filter, search, updateContext, deleteContext, ... **after:** createContext, updateContext, deleteContext, filter, search, ...)

### What does it look like:
link_grids

The changes are very useful for UX, they reduce the number of clicks and it becomes more convenient to work: now it’s easy to open non-directly related elements, but before you had to look for an element by navigating through sections in the manager panel.

Thanks a lot to @GulomovCreative for help in PR!

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14125 (paragraph 1)

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -644,7 +644,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
         });
     }
 
-    ,rendLink: function(v,attr) {
+    ,renderLink: function(v,attr) {
         var el = new Ext.Element(document.createElement('a'));
         el.addClass('x-grid-link');
         el.dom.title = _('edit');
@@ -1063,7 +1063,7 @@ Ext.extend(MODx.grid.LocalGrid,Ext.grid.EditorGridPanel,{
         });
     }
 
-    ,rendLink: function(v,attr) {
+    ,renderLink: function(v,attr) {
         var el = new Ext.Element(document.createElement('a'));
         el.addClass('x-grid-link');
         el.dom.title = _('edit');
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -252,7 +252,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
             ,url: this.config.url
             ,params: p
             ,listeners: {
-            	'success': {fn:this.refresh,scope:this}
+                'success': {fn:this.refresh,scope:this}
             }
         });
     }
@@ -275,7 +275,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,url: this.config.url
                 ,params: p
                 ,listeners: {
-                	'success': {fn:function() {
+                    'success': {fn:function() {
                         this.removeActiveRow(r);
                     },scope:this}
                 }
@@ -334,10 +334,10 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,groupField: this.config.groupBy || 'name'
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
-				,listeners:{
+                ,listeners:{
                     load: function(){
-						Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
-					}
+                        Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
+                    }
                 }
             });
         } else {
@@ -350,10 +350,10 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,remoteSort: this.config.remoteSort || false
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
-				,listeners:{
+                ,listeners:{
                     load: function(){
-						Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
-					}
+                        Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
+                    }
                 }
             });
         }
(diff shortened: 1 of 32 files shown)


JoshuaLuckers:
The underlines make it look restless. How does it look when you show them on hover only?

Ruslan-Aleev:
I specifically made them to be underlined, and when you hover, the underline disappears.
Otherwise, you can’t see that the link in the grid, for example, in the "Users" section (/manager/?a=security/user), has been linked since the 2.x version, did you know about this? :)
But maybe you have another style option?

JoshuaLuckers:
> Otherwise, you can’t see that the link in the grid

In the Media Sources grid this looks fine but in the grid with access permissions it looks very restless (to me). 

>for example, in the "Users" section (/manager/?a=security/user), has been linked since the 2.x version, did you know about this? :)

I didn't know that, I might have noticed it when all columns had the "click" behaviour. 

> But maybe you have another style option?

Only thing that comes to mind is to show the underline only on hover. But this is just my opinion, I won't prevent this from being merged!

manager/assets/modext/widgets/element/modx.grid.template.tv.js
```diff
@@ -129,14 +133,43 @@ Ext.extend(MODx.grid.TemplateTV,MODx.grid.Grid,{
         }
         return m;
     }
+
     ,updateTV: function(itm,e) {
-        MODx.loadPage('Element/TemplateVar/Update', 'id='+this.menu.record.id);
+        MODx.loadPage('element/tv/update', 'id='+this.menu.record.id);
```
JoshuaLuckers, line 138: Why did you change this? 
To prevent issues it's better to have MODX connector actions match the classname of the corresponding processor (CamelCase).

This also applies to the other action urls in this PR.

Ruslan-Aleev:
> In the Media Sources grid this looks fine but in the grid with access permissions it looks very restless (to me).

Here I agree, but there are not many grids where there are many columns with links, mainly related to access, in other grids, as a rule, there is 1 link per row.

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -644,6 +644,17 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
         });
     }
 
+    ,rendLink: function(v,attr) {
```
JoshuaLuckers: A better name for this function would be `renderLink` (or is there a particular reason why it's called `rendLink`?).
```suggestion
    ,renderLink: function(v,attr) {
```

manager/assets/modext/widgets/element/modx.grid.template.tv.js
Ruslan-Aleev: The `Element/TemplateVar/Update` does not work, `element/tv/update` is the same action in the TV tree and used it.
As for the register, I did not know which one to use, because both uses in MODX 3, I created an issue https://github.com/modxcms/revolution/issues/14744 on this topic.

manager/assets/modext/widgets/core/modx.grid.js
Ruslan-Aleev: Yes, probably the new name is clearer, I will correct it later.

manager/assets/modext/widgets/element/modx.grid.template.tv.js
Ruslan-Aleev: Most in code, in my opinion, use lowercase actions, so I used lowercase.

manager/assets/modext/widgets/core/modx.grid.js
Mark-H: It's probably called that because that's the pattern used for other renderers too (e.g. `rendYesNo`)

manager/assets/modext/widgets/element/modx.grid.template.tv.js
JoshuaLuckers: Actions should be in camel case, matching the name/structure of the processor class. The lowercase actions work because MODX (tries to) converts them camelcase. 

For reference also see this PR: https://github.com/modxcms/revolution/pull/14871

If only the lowercase actions works we need to figure out why the camel case version doesn't work and fix it.

Mark-H: Note that this loads a **controller**, not a **processor**. Controllers haven't been touched yet for 3.0.

Ruslan-Aleev: Actions work in any case. In the examples, they just are not correctly indicated:
- Element/**TemplateVar**/Update -> Element/**TV**/Update or element/tv/update
- security/access/usergroup/**namespace**/remove -> Security/Access/UserGroup/**AccessNamespace**/Remove or security/access/usergroup/accessnamespace/remove

Ruslan-Aleev: And the bugs associated with https://github.com/modxcms/revolution/issues/14639 are due to the fact that Linux considers files in a different register (files exactly).
As I understand it, but maybe I'm wrong.
Now actions work both in lower and upper case (the server is the same).
**Although they should lead to one format, in any case.**

Mark-H: Just to reiterate, #13639 is about **processors**. This fix is about a **controller**. Totally different thing. 

Uppercasing paths is new to **processors** only. Processors also still work with lowercase paths. Processors also convert legacy "TV" to new "TemplateVar".

**Controller** behavior has not changed in 3.0 (yet). [That splits the path and uppercases the first character regardless of how it is provided.](https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/modManagerResponse.php#L264-L274) That could use CamelCase if we'd wanted, but it doesn't have to. 

(Your fix in this line of code, specifically converting `TemplateVar` back to `tv` is correct)

JoshuaLuckers: > Note that this loads a **controller**, not a **processor**. Controllers haven't been touched yet for 3.0.

Aha! Conversation solved. 😄

Ruslan-Aleev:
So do I end up editing actions or not? To fix less in the future? 
Because there is currently no standard in writing actions in MODX 3 - https://github.com/modxcms/revolution/issues/14744.
Although I like lower case more :)

JoshuaLuckers (approved):

Mark-H (approved):
Will test it in more detail when merging. Thanks.

PR #14867, opened 2019-12-04
Removed unused system settings for "Caching" section

Ruslan-Aleev:
### What does it do?
Removed unused system settings for "Caching" section.

- cache_db_expires
- cache_expires
- cache_format

### Related issue(s)/PR(s)

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -190,15 +190,6 @@ $settings['cache_db']->fromArray(array (
   'area' => 'caching',
   'editedon' => null,
 ), '', true, true);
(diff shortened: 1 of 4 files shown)


Mark-H (changes requested):
I'm 95% sure these are used in the xPDO cache handling, which might refer to them using the `xPDO::OPT_CACHE_EXPIRES`/`xPDO::OPT_CACHE_FORMAT` and dynamically generated `cache_{partition}_expires` causing a simple search to not show them.

Ruslan-Aleev:
Yes, you're right, these settings are registered in xPDO: https://github.com/modxcms/xpdo/blob/3.x/src/xPDO/xPDO.php#L58

PR #14872, opened 2019-12-05
Fix sizing of the actions column

### LINKED ISSUE #14859: [3.x] Grid column sizing might be way off due to gear icon for context menu
### Summary

In https://github.com/modxcms/revolution/pull/14806, a new column is dynamically added to all (incl third party) grids, with a width of 50.

Works great, until you're loading a third party grid that is not defined in "pixel-ish" values, but in "percentage-ish". Because widths are relative to the total, one could be defining grid columns with widths of "0.2", "0.5", "0.3" for a total of 1, and ExtJS will calculate that to the proper pixel values on render.

Now imagine what that looks like when a magic column with a default width of 50 is added... your total is 51, 50 of which goes to the gear icon.

### Step to reproduce

Install an extra that defines "percentage-ish" widths for columns instead of "pixel-ish". I ran across it with ClientConfig's admin view, but I'm sure there are others. 

### Observed behavior

Schermafbeelding 2019-11-29 om 23 43 18

### Expected behavior

I'm not quire sure - mostly posting this to see what others and in particular @theboxer think. It's a simple fix, where the extra can define the width of the column (`config.actionsColumnWidth`)... but maybe there are some bright ideas on other solutions.

### Environment

3.x branch from today

theboxer:
### What does it do?
Figures out if columns are using percentage width and if yes, uses `0.1` as a default, instead of `50`.

### Why is it needed?
When using percentage based width on columns, actions column is so huge :)

### Related issue(s)/PR(s)
Resolves #14859

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -22,7 +22,6 @@ MODx.grid.Grid = function(config) {
         ,showPerPage: true
         ,stateful: false
         ,showActionsColumn: true
-        ,actionsColumnWidth: 50
         ,disableContextMenuAction: false
         ,menuConfig: {
             defaultAlign: 'tl-b?'
@@ -96,16 +95,41 @@ MODx.grid.Grid = function(config) {
     }
 
     if (config.showActionsColumn) {
+        var defaultActionsColumnWidth = 50;
+
+        var isPercentage = function(columns) {
+            for (var column of config.columns) {
+                if (column.width && (column.width > 1)) {
+                    return false;
+                }
+            }
+
+            return true;
+        };
+
         if (config.columns && Array.isArray(config.columns)) {
+            if (config.actionsColumnWidth === undefined) {
+
+                if (isPercentage(config.columns)) {
+                    defaultActionsColumnWidth = 0.1;
+                }
+            }
+
             config.columns.push({
-                width: config.actionsColumnWidth || 50
+                width: config.actionsColumnWidth || defaultActionsColumnWidth
                 ,renderer: this.actionsColumnRenderer.bind(this)
             });
         }
 
         if (config.cm && config.cm.columns && Array.isArray(config.cm.columns)) {
+            if (config.actionsColumnWidth === undefined) {
+                if (isPercentage(config.cm.columns)) {
+                    defaultActionsColumnWidth = 0.1;
+                }
+            }
+
             config.cm.columns.push({
-                width: config.actionsColumnWidth || 50
+                width: config.actionsColumnWidth || defaultActionsColumnWidth
                 ,renderer: this.actionsColumnRenderer.bind(this)
             });
         }
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -98,8 +98,8 @@ MODx.grid.Grid = function(config) {
         var defaultActionsColumnWidth = 50;
 
         var isPercentage = function(columns) {
-            for (var column of config.columns) {
-                if (column.width && (column.width > 1)) {
+            for (var i = 0; i < columns.length; i++) {
+                if (columns[i].width && (columns[i].width > 1)) {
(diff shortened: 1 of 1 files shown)


Mark-H (approved):
That's pretty clever!

JoshuaLuckers:
Let's say we have 4 columns, using percentages for widths that make up a total of 1 (all columns have a width of 0.25). If the new default action column is then added the total width exceeds `1`. 

What happens when the total width of all columns exceed `1`? Will it cause any issues?

Mark-H:
@JoshuaLuckers Regardless how column widths are defined, in this percentage or pixel-ish way, ExtJS still calculates the width for each column. If you have columns of 2.5, 2.5, 5 and 10 width, in a panel that has 600 pixels, it will those to take up 75, 75, 150, and 300px respectively. Same widths will be used if you define 200, 200, 400, 800 - it scales those numbers to the actual width available. 

This PR just makes sure that the now automatically added column doesn't take over the entire grid by specifying a default of 50 while all other columns are a combined 1.

JoshuaLuckers:
Thanks for the explanation! 

> If you have columns of 2.5, 2.5, 5 and 10 width, in a panel that has 600 pixels, it will those to take up 75, 75, 150, and 300px respectively. 

In this example, the default width for the action column would be `50` or not?

Mark-H:
Yeah. Perhaps not the best examples, just wanted to illustrate how ExtJS calculates the width.

JoshuaLuckers:
> Yeah. Perhaps not the best examples, just wanted to illustrate how ExtJS calculates the width.

It's a good example because it shows there is room for improvement in calculating the default width 😄.

Mark-H:
I don't think that's necessarily a use case we have to cover, though. This just helps with the migration to 3 for what I think is a fairly common way (<= 1) along with the pixel-ish based.

JoshuaLuckers (approved):

opengeek:
```
Running "uglify:jsgrps" (uglify) task
[JS_Parse_Error [SyntaxError]: Unexpected token name «of», expected punc «;»
  at JS_Parse_Error.get (<anonymous>:69:23)
  at formatError (internal/util/inspect.js:910:19)
  at formatRaw (internal/util/inspect.js:703:14)
  at formatValue (internal/util/inspect.js:591:10)
  at inspect (internal/util/inspect.js:221:10)
  at formatWithOptionsInternal (internal/util/inspect.js:1703:40)
  at formatWithOptions (internal/util/inspect.js:1576:10)
  at Object.Console.<computed> (internal/console/constructor.js:276:10)
  at Object.log (internal/console/constructor.js:286:61)
  at /Users/opengeek/www/revo/_build/templates/default/node_modules/grunt-contrib-uglify/tasks/uglify.js:144:17
  at Array.forEach (<anonymous>:null:null)
  at Object.<anonymous> (/Users/opengeek/www/revo/_build/templates/default/node_modules/grunt-contrib-uglify/tasks/uglify.js:78:16)
  at Object.<anonymous> (/Users/opengeek/www/revo/_build/templates/default/node_modules/grunt/lib/grunt/task.js:252:15)
  at Object.thisTask.fn (/Users/opengeek/www/revo/_build/templates/default/node_modules/grunt/lib/grunt/task.js:70:16)
  at Object.<anonymous> (/Users/opengeek/www/revo/_build/templates/default/node_modules/grunt/lib/util/task.js:294:30)
  at Task.runTaskFn (/Users/opengeek/www/revo/_build/templates/default/node_modules/grunt/lib/util/task.js:244:24)
  at Task.<anonymous> (/Users/opengeek/www/revo/_build/templates/default/node_modules/grunt/lib/util/task.js:293:12)
  at /Users/opengeek/www/revo/_build/templates/default/node_modules/grunt/lib/util/task.js:220:11
  at processTicksAndRejections (internal/process/task_queues.js:75:11)
] {
  message: 'Unexpected token name «of», expected punc «;»',
  filename: 'widgets/core/modx.grid.js',
  line: 101,
  col: 28,
  pos: 3274
}
```

^^ Error when running `grunt build`

PR #14873, opened 2019-12-06
Format timestamp to date & time in edit file panel

### LINKED ISSUE #14750: MODX3: Incorrectly displayed date in the file editing form
### Summary
The date is incorrectly displayed in the file editing form for MODX 3, although there is no such problem in MODX 2.7.

m3

m2

**It is necessary to display the date and time in accordance with the date format in the system settings.**

### Environment
MODX 3

theboxer:
### What does it do?
Format timestamp using system settings (`manager_date_format` & `manager_time_format`) into a friendlier format.

### Why is it needed?
In the edit file panel, last accessed & last modified dates are displayed as a timestamp.

### Related issue(s)/PR(s)
Resolves #14750

### DIFF
--- a/manager/assets/modext/sections/system/file/edit.js
+++ b/manager/assets/modext/sections/system/file/edit.js
@@ -110,14 +110,14 @@ MODx.panel.EditFile = function(config) {
                     ,name: 'last_accessed'
                     ,id: 'modx-file-last-accessed'
                     ,anchor: '98%'
-                    ,value: config.record.last_accessed || ''
+                    ,value: ((new Date(parseInt(config.record.last_accessed) * 1000)).format(MODx.config.manager_date_format + ' ' + MODx.config.manager_time_format)) || ''
                 },{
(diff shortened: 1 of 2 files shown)


Jako:
Two questions: 
- Why not use a renderer method? Is that not possible with statictext?
- Why does the connector returns a value that has to be multiplied with 1000?

JoshuaLuckers:
> * Why does the connector returns a value that has to be multiplied with 1000?

Timestamps in JavaScript are in milliseconds.

theboxer:
Renderer method is only available in grids, not in inputs.

Jako:
Thanks, everything is fine then.

Jako (approved):

JoshuaLuckers:
Is it possible to create a function for it instead to improve reusability and readability?

JoshuaLuckers (approved):
Thanks a lot @theboxer !

Mark-H (approved):

Issue #9960, opened 2013-06-04
Fire OnResourceDuplicate for duplicated children


BobRay:
bobray created Redmine issue ID 9960

I think this would do it (untested), but there may be a better way.

Remove the OnResourceDuplicate invocation from the resource/duplicate processor.

Put this code just below the line in the modResource->duplicate() method where $duplicateChildren is set:

<pre>
$this->modx->invokeEvent('OnResourceDuplicate',array(
            'newResource' => &$newResource,
            'oldResource' => &$oldResource,
            'newName' => $newName,
            'duplicateChildren' => $duplicateChildren,
            'prefixDuplicate' => $prefixDuplicate,
            'publishedMode' => $publishedMode,
        )); 
</pre>

OptimusCrime:
Hey @BobRay, want to explain this issue a bit? I do not understand what this fixes. Also, feel free to submit a PR.

BobRay:
Sure.  Users could have plugins that do stuff when a resource is duplicated (I know I did at one time).  If you duplicate a resource and select "duplicate children," the plugin may not perform the actions on the children when they're duplicated.

This probably shouldn't be fixed, because the plugin can check the properties to see if children have been duplicated and act accordingly.

Closing.

Mark-H:
Per discussion in #14874, this should be fixed with [solution 1 described here](https://github.com/modxcms/revolution/pull/14874#issuecomment-751983997)

PR #14874, opened 2019-12-06
Move OnResourceDuplicate invoke event

sdrenth:
### What does it do?
Moved the invokeEvent of the OnResourceDuplicate event to the modResource class where the actual duplication is processed, so the event is invoked every time a resource is duplicated and not only when the processor is used.

### Why is it needed?
Currently there is no way to hook in when duplicating child resources or when resources are duplicated when using the context duplicate functionality.  This is needed so plugins can hook into this process as well when, for instance if you'd like to duplicate custom database values related to a resource.

Also I think it is expected that this event is invoked every time a resource is duplicated and not only when the processor is used (then it only runs for the parent resource).

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/9960

### DIFF
--- a/core/src/Revolution/Processors/Resource/Duplicate.php
+++ b/core/src/Revolution/Processors/Resource/Duplicate.php
@@ -81,7 +81,6 @@ class Duplicate extends Processor
             return $this->failure($this->newResource);
         }
 
-        $this->fireDuplicateEvent();
         $this->logManagerAction();
 
         return $this->success('', [
@@ -102,22 +101,6 @@ class Duplicate extends Processor
             $canAddChildren = false;
         }
         return $canAddChildren;
-    }
-
-    /**
-     * Fire the OnResourceDuplicate event
-     * @return void
-     */
-    public function fireDuplicateEvent()
-    {
-        $this->modx->invokeEvent('OnResourceDuplicate', [
-            'newResource' => &$this->newResource,
-            'oldResource' => &$this->oldResource,
-            'newName' => $this->getProperty('name', ''),
-            'duplicateChildren' => $this->getProperty('duplicate_children', false),
-            'prefixDuplicate' => $this->getProperty('prefixDuplicate', false),
-            'publishedMode' => $this->getProperty('published_mode', 'preserve'),
-        ]);
     }
 
     /**
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1252,6 +1252,15 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
             }
         }
 
+        $this->xpdo->invokeEvent('OnResourceDuplicate',array(
+            'newResource' => $newResource,
+            'oldResource' => $this,
+            'newName' => $newName,
+            'duplicateChildren' => $duplicateChildren,
+            'prefixDuplicate' => $prefixDuplicate,
+            'publishedMode' => $publishedMode,
+        ));
+
         return $newResource;
 
     }


JoshuaLuckers:
Transcript of the discussion between on Slack:
> **Jason Coward** (@opengeek)
> it's not done that way because we at one time tried to keep events out of the model
> I mean, that's the whole purpose of having a separate processor layer
> of course, I see that several events have been pushed down into the model, and now we have an inconsistent mess

> **sdrenth** (@sdrenth)
> I can understand that you’d like to have that done in a consistent manner, but I think how it is working at this moment is not how it’s supposed to be/expected to be, because now there is no way to hook in if a child resource is duplicated or if a context is duplicating the resources. But that might be very helpful if you have custom database models related to resource and want to duplicate the values for those models as well.

> **Jason Coward** (@opengeek)
> but duplication is not something that should be part of the resource model itself; that's something you do with a resource, not as a resource — in any case, it needs some discussion

JoshuaLuckers:
Wether the `duplicate` function should be part of the model or not;
I do think it makes sense to have events inside the model because:
1. Events should be trusted to be fired, they need to be triggered from a single source of truth. 
2. A model/object is the single source of truth when it comes to data access, manipulation and mutations (especially when it's an Active record object).

If a model has a function to duplicate itself then I also assume it to be triggered if there is a `duplicate` event I can subscribe to.

opengeek:
> If a model has a function to duplicate itself then I also assume it to be triggered if there is a duplicate event I can subscribe to.

No. That is the function of business logic—which is the only reason we have a processors layer—not a data model. The data model should really only be about interactions with the data storage mechanism. If events are firing in the data model that are related to custom business logic, then you cannot use the data model to manipulate data without fear of triggering custom business logic processes. We already run into this problem is places where people use teleport to try and work with the data model and it triggers plugins that should not be firing when simply importing/exporting the data objects.

sdrenth:
@opengeek I still believe the nicest way to handle these kind of flows is by invoking events in the model and tend to agree with @JoshuaLuckers here. I'm not aware of any issues with invoking events in the model and teleport or if that might arise more problems in general and not only for the teleport use-case. Maybe we should also be looking at how these issues could be resolved while keeping the events in the model.

Regarding the business logic I think thats the reason why we use plugins right? And in order to handle the business logic in a correct way we should be able to hook into events when desired, the MODX core should provide that by invoking events in the right places, this might not always be in a processor.

Asides from that you're against invoking events in de model, I would like to know how you would like the issue as described in this PR to be resolved.

Also I lean towards an implementation as how Eloquent models do:
https://laravel.com/docs/5.8/eloquent#events

So to invoke the same events for every object (create/update/delete) and to extend these for some models (duplicate) if needed.

Oetzie:
I agree with @sdrenth. An event should be triggered in the model each time a model is created, updated, duplicated or removed. The plugin it self should check if he needs to be fired or not. So if you have a problem with Teleport and firing plugins I quess the logic of the plugin is not correct.

JoshuaLuckers:
> > If a model has a function to duplicate itself then I also assume it to be triggered if there is a duplicate event I can subscribe to.
> 
> No. That is the function of business logic—which is the only reason we have a processors layer—not a data model. The data model should really only be about interactions with the data storage mechanism. If events are firing in the data model that are related to custom business logic, then you cannot use the data model to manipulate data without fear of triggering custom business logic processes. We already run into this problem is places where people use teleport to try and work with the data model and it triggers plugins that should not be firing when simply importing/exporting the data objects.

Wouldn't it be better to turn off the firing of events at this case be a better idea? If you have an alarm system that should trigger whenever you access an "object" there should be no way you can access it without triggering the alarm unless you turn it off. Or isn't that conform the Active record pattern?

sdrenth:
@JoshuaLuckers I think that would be a great idea, perhaps by doing a `$modx->setOption('invoke_events', false);` which of course is enabled by default and checked upon in the invokeEvent method?

This way you will be able to disable invoking events temporarily for just a part of your code and enable it afterwards.

```
// My code using invoke events.
$modx->setOption('invoke_events', false);

// My code without invoking events.
$modx->setOption('invoke_events', true);

// My code using invoke events again.
```

Mark-H:
In case of the `save` method on a model, a teleport/gitify/whatever "import" script calling save and firing an event could lead to things that probably shouldn't happen (for example a "product" being imported firing a plugin that sends data to a product information management system), because that likely already ran on the environment it was exported from. 

In that case, the model/data layer should only be responsible for persisting the information, and it should not fire any events doing so. That same layer also doesn't necessarily transform any data, that's all in the processor and should stay there. That's probably agreeable by all here.

But that's the `save` method - it does get more muddy with the `duplicate` method that is more of a user interaction than just a read-old-and-persist-new-object. There's ample business logic in duplicate methods about how to deal with related objects. And with inheritance like we have it, having that on the model makes sense (a moregallery resource type needs different business logic for the duplicate method than a regular resource). 

But then you do run into the issue faced here - the business logic for duplicating child resources being outside of the processor causing the event to only be fired for the top-level resource. While moving the event (as in this PR) is a very pragmatic short term fix, I do share Jason's architectural concerns about it not necessarily being the right place for it. 

As an alternative solution, maybe we can move the child duplication logic out of the model and into the processor instead. Let the model `duplicate` method focus on its own entities (like TVs, or images in moregallery) and let the processor determine what needs to be duplicated and moved where. That might make the processor code more complex than the current recursive approach... but perhaps architecturally better.

Mark-H:
Moving to the milestone for beta as this likely wont make the alpha1 cut-off.

JoshuaLuckers:
**This comment is only for the events/triggers discussion, I will share my thoughts about the duplicate function later**

Remember, we are using the active record pattern here (without using a table class). 
> An object that wraps a row in a database table or view, encapsulates the database access, and adds domain logic on that data.

Source: [Martin Fowler - Active Record](https://www.martinfowler.com/eaaCatalog/activeRecord.html)

[MySQL](https://dev.mysql.com/doc/refman/8.0/en/triggers.html) and [MSSQL](https://www.mssqltips.com/sqlservertip/5909/sql-server-trigger-example/) both have triggers that are associated with a table and triggered when certain events occur on that table.

> triggers can be used to enforce business rules and data integrity, query other tables

Source: [SQL Docs - DML Triggers](https://docs.microsoft.com/en-us/sql/relational-databases/triggers/dml-triggers?view=sql-server-ver15)

Normally we would have an entity and table class where the table class would be responsible for insert/update/delete actions (and the best place to trigger events) but since we don't have separate classes for tables the object itself seems to be the best place to trigger events. Or maybe we need callbacks instead?

In Ruby on Rails there are "[Active Record Callbacks](https://guides.rubyonrails.org/active_record_callbacks.html)" that allow you to trigger logic before or after an alteration of an object's state. This is also part of the object.

opengeek:
a) xPDO is not a pure active record implementation. It is in fact a hybrid. But I'm not going to get into a philosophical argument about design patterns for this.

b) xPDO already has callbacks for methods that change the state in the database.  This is also not what we are discussing.

c) This is about MODX and its domain logic layer versus the model persistence layer. I'm saying that this duplicate method belongs in the domain logic layer (i.e. processors) along with all other MODX events. For consistency, and to keep the database model out of the business of what happens in the current MODX UX. As Mark implied, we can accomplish this without shoving events into the persistence layer, and all I'm asking is we look into doing it that way.

sdrenth:
@opengeek @Mark-H Sorry for the late reply, I can see why you'd like that logic to live in processors instead of the model itself. 

**Solution 1**
So in other words, to fix this correctly we would call the duplicate method in the duplicate processor with duplicateChildren set to false and copy this piece of code into the processor itself:
```
if (!$this->checkPolicy('add_children')) return $newResource;

$criteria = array(
    'context_key' => $this->get('context_key'),
    'parent' => $this->get('id')
);

$count = $this->xpdo->getCount('modResource',$criteria);

if ($count > 0) {
    $children = $this->xpdo->getIterator('modResource',$criteria);

    /** @var modResource $child */
    foreach ($children as $child) {
        $child->duplicate(array(
            'duplicateChildren' => true,
            'parent' => $newResource->get('id'),
            'prefixDuplicate' => $prefixDuplicate,
            'overrides' => !empty($options['overrides']) ? $options['overrides'] : false,
            'publishedMode' => $publishedMode,
            'preserve_alias' => $preserve_alias,
            'preserve_menuindex' => $preserve_menuindex
        ));
    }
}
```

And of course add the invokeEvent so it's called every time a resource or child resource is duplicated. Only downside is that we have duplicate code doing the same stuff now. 

**Solution 2**
Another option would be to keep the invokeEvent in the model class but passing a parameter invokeEvent in order to run this logic or not and having a default value of false will ensure it is only invoked when explicitely told to do so:
```
  if (isset($options['invokeEvent']) && $options['invokeEvent']) {
          $this->xpdo->invokeEvent('OnResourceDuplicate',array(
              'newResource' => $newResource,
              'oldResource' => $this,
              'newName' => $newName,
              'duplicateChildren' => $duplicateChildren,
              'prefixDuplicate' => $prefixDuplicate,
              'publishedMode' => $publishedMode,
          ));
  }
```

Any thoughts or maybe any better ideas?

Ibochkarev:
@opengeek @Mark-H please give your answers

opengeek:
@JoshuaLuckers — why was this assigned to me?

Mark-H:
Probably to make a decision on what to do with this PR 😛 

Re-reading the comments, I'd say this PR as-is is to be rejected, but [solution 1 Sander pitched here](https://github.com/modxcms/revolution/pull/14874#issuecomment-751983997) would be acceptable in the future?

opengeek:
@sdrenth — do you want to update this PR with your proposed solution 1, or should we close this one and you'll provide a new PR with that solution?

sdrenth:
@opengeek please proceed by closing this one and creating a new one.

Issue #13269, opened 2017-01-28
Honor value of "server_protocol"


OptimusCrime:
### Current state

Right now, as of MODX 2.x, the setting `server_protocol` does not do anything (?). The setting has the values `http` and `https`. Anyone would expect the choice here to be honored by MODX, but it does not. The actual protocol is fetched automatically by evaluating the request which is done by inspecting various server settings and values.

### Potential

Sometimes you want to set the value absolute, and sometimes the value is not evaluated correctly by MODX because of various shared host solutions (se relevant issues). Sometimes you also know that you are either always serving everything over http, or you know that everything is served over https.

### Idea

- Add a new value (default) that is automatic. Perhaps just call it `automatic` (?).
- The automatic value evaluates as best as it can if the request is http or https. This value is needed for sites that is served over both http and https.
- http value should NEVER allow to serve over https.
- https value should NEVER allow to serve over http.

### Relevant issues

#13259 and perhaps more

sottwell:
What does Wayfinder et. al. use to determine how to generate full URLs? If it just uses the current automatic settings, you'd end up with all kinds of mixed up URLs because of caching.

OptimusCrime:
The automatic setting it how MODX does things today, as you can not force either http or https automatically. MODX determine if the request is a secure one by doing this: https://github.com/modxcms/revolution/blob/fbf126690f6ce903faeaeb9f1d958211fc2203c9/core/docs/config.inc.tpl#L49-L57

gadgetto:
+1 for honoring **server_protocol** setting to override automatic fetching of protocol

christianseel:
Isn't that what `site_url` is doing right now? If you set that as a system or context setting to something like `https://mysite.com/`, MODX will generate all URLs with https. That's what we're doing at our projects…

From my perspective there is no reason to change the `server_protocol` behavior.

#modxbughunt

OptimusCrime:
After I opened this issue i learned that `server_protocol` does nothing. I should in fact be removed.

What I actually wanted was `link_scheme` which defines if links should be created with or without SSL.

OptimusCrime:
I have learned that we need to specify the protocol much earlier than when we have the system settings available. This means that this approach is essentially worthless. A better configuration file which is better at identifying SSL and reverse proxies would be better. I have a PR for this ready in #13594. While it does not solve the problem completely, it should make some edge cases work out of the box.

rthrash:
[...]

PR #14877, opened 2019-12-07
Removed unused system settings for "System and Server" section

Ruslan-Aleev:
### What does it do?
Removed unused system settings for "System and Server" section.

- `resolve_hostnames`
- `server_protocol` (https://github.com/modxcms/revolution/issues/13269)
The server_protocol setting is confusing, I saw posts, for example, when setting up redirects, people are looking for the cause of the problem, including in this setting, not knowing that it does not work at all :)

These settings are not used in 3.x; they were specified in Evolution:
- **resolve_hostnames** https://github.com/modxcms/evolution/blob/v1.0.0/install/setup.sql#L847
- **server_protocol** https://github.com/modxcms/evolution/blob/v1.0.0/install/setup.sql#L837

### Related issue(s)/PR(s)

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -84,9 +84,6 @@ $_lang['setting_login_homepage'] = 'Login Home Page';
 $_lang['setting_login_homepage_desc'] = 'Enter the ID of the document you want to send the user to after he/she has logged in. <strong>NOTE: Make sure the ID you enter belongs to an existing document, and that it has been published and is accessible by this user!</strong>';
 
 // system settings
-$_lang['setting_access_policies_version'] = 'Access Policy Schema Version';
-$_lang['setting_access_policies_version_desc'] = 'The version of the Access Policy system. DO NOT CHANGE.';
-
 $_lang['setting_allow_forward_across_contexts'] = 'Allow Forwarding Across Contexts';
 $_lang['setting_allow_forward_across_contexts_desc'] = 'When true, Symlinks and modX::sendForward() API calls can forward requests to Resources in other Contexts.';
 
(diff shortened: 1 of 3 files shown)


core/lexicon/en/setting.inc.php
```diff
@@ -84,9 +84,6 @@
 $_lang['setting_login_homepage_desc'] = 'Enter the ID of the document you want to send the user to after he/she has logged in. <strong>NOTE: Make sure the ID you enter belongs to an existing document, and that it has been published and is accessible by this user!</strong>';
 
 // system settings
-$_lang['setting_access_policies_version'] = 'Access Policy Schema Version';
-$_lang['setting_access_policies_version_desc'] = 'The version of the Access Policy system. DO NOT CHANGE.';
-
```
JoshuaLuckers: I believe these should stay because the related system setting seems to be used during the setup/upgrade of MODX:

https://github.com/modxcms/revolution/blob/c44faaf70d96ad351d713e7d8cced20124cbeed4/setup/includes/upgrade.install.php#L161-L330

Ruslan-Aleev: Yes, but I deleted only the lexicon, the setting itself is not present in the system settings grid. https://github.com/modxcms/revolution/blob/3.x/_build/data/transport.core.system_settings.php

Mark-H: That looks like it might be a remnant from the upgrade to v2.0.5. I believe we're not planning to support 3.0 upgrades from anything but 2.6/2.7, so that looks like code which could also be removed.

Ruslan-Aleev: I was wrong. This setting is specified during the upgrade. Now I will return the lexicon.

Ruslan-Aleev: @Mark-H Sent a commit before your comment :)
I think that all lexicons can be cleaned with a separate PR group. Let this lexicon remain for now.

Mark-H (approved):

PR #14890, opened 2019-12-12
Header breadcrumbs

### LINKED ISSUE #14769: MODX3: Expand breadcrumbs
### Summary
In the current version, breadcrumbs, in fact, work only for child resources.

It would be useful if breadcrumbs would work for related sections.

**For example:**
Media Sources `(/manager/?a=source)` -> Media Source: Filesystem `(/manager/?a=Source/Update&id=1)`

This is relevant for Contexts, Users, Form Customization, Dashboards, Widgets and for others.

### Related issue(s)/PR(s)
None

theboxer:
### What does it do?
Adds a helper function to generate header breadcrumbs for pages. Implements it for sources & users. There are more places where it can be implemented, but I didn't know if this will be a wanted enhancement or not. So if the idea gets approved I'll add it to other places.

### Why is it needed?
As described in the issue #14769, it seems nice to have breadcrumbs in more places than just in resources.

### Related issue(s)/PR(s)
Resolves #14769

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -1491,7 +1491,7 @@ iframe[classname="x-hidden"] {
       order: 1;
       font: normal 20px $bodyfonts;
       color: $darkestGray;
-      margin: 0;
+      margin: 0 !important;
       padding-left: 0;
 
       @include grid-media($desktop) {
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -593,7 +593,7 @@ a.x-grid-link:focus {
   #modx-content > .x-panel-bwrap > .x-panel-body > .x-panel {
     margin: 0;
 
-    #modx-resource-breadcrumbs {
+    .modx-header-breadcrumbs {
       margin-top: 1.25rem;
       font-weight: bold;
       box-sizing: border-box;
@@ -605,7 +605,7 @@ a.x-grid-link:focus {
       }
     }
 
-    #modx-resource-breadcrumbs {
+    .modx-header-breadcrumbs {
       width: 100% !important;
     }
   }
@@ -1481,13 +1481,13 @@ iframe[classname="x-hidden"] {
 }
 
 /* New Header */
-#modx-resource-breadcrumbs {
+.modx-header-breadcrumbs {
   .breadcrumbs {
     display: flex;
     flex-wrap: wrap;
     align-items: baseline;
 
-    #modx-resource-header {
+    h2 {
       order: 1;
       font: normal 20px $bodyfonts;
       color: $darkestGray;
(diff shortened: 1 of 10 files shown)


Ruslan-Aleev:
Thanks! Useful improvement!

_build/templates/default/sass/index.scss
```diff
@@ -1481,13 +1481,13 @@ iframe[classname="x-hidden"] {
...
 /* New Header */
-#modx-resource-breadcrumbs {
+.modx-header-breadcrumbs {
   .breadcrumbs {
     display: flex;
     flex-wrap: wrap;
     align-items: baseline;
 
-    #modx-resource-header {
+    h2 {
```
Ruslan-Aleev, line 1490: For h2, need to add `margin: 0 !important`, because `.modx-page-header` adds extra indentation at the top.

Ruslan-Aleev (approved):

manager/assets/modext/widgets/source/modx.panel.source.js
```diff
@@ -64,7 +64,7 @@ MODx.panel.Source = function(config) {
                                 ,anchor: '100%'
                                 ,listeners: {
                                     'keyup': {scope:this,fn:function(f,e) {
-                                        Ext.getCmp('modx-source-header').getEl().update(_('source')+': '+f.getValue());
+                                        Ext.getCmp('modx-source-header').getEl().update(f.getValue());
```
Ruslan-Aleev: Maybe here it is also worth outputting in the format `Ext.util.Format.htmlEncode(f.getValue())`? I do not know if it is necessary, but the different output variant `f.getValue()` in modx.panel.user.js and modx.panel.source.js is confusing.

Mark-H:
I like where this is going :)

Mark-H:
Other than the fix @Ruslan-Aleev mentioned for making sure we don't introduce XSS opportunities - what is needed for marking this for review @theboxer?

theboxer:
I'll go through more manager pages and see where else it can be implemented. After that I'll remove the draft tag from it.

Mark-H:
Would be nice to get this into alpha2 (along with some new docs explaining how to use it), even if it's not yet implemented in other sections.

theboxer:
I wanted to work on in during Christmas vacation, seems like I failed...

Mark-H:
's okay, I wanted to not gain any weight during the Christmas vacation and failed too..

ghost:
Have you looked at ModDevTools? It has a lot of breadcrumbs for resources and tabs for seeing what element is used where.

Mark-H:
@cla-bot check

theboxer:
@cla-bot check

theboxer:
Documentation was already added...

Ruslan-Aleev:
Thank you for your work, cool done!
But, as it seems to me, for the `"Form Customization" (/manager/?a=security%2Fforms)` section it’s also worth adding crumbs, there is a structure like in the "Users" section.

Mark-H (approved):

PR #14898, opened 2019-12-14
Clean up legacy core/model files

JoshuaLuckers:
### What does it do?
Clean up legacy core/model directories and files.

### Why is it needed?
When upgrading from 2.x to 3.x most people merge directories and replaces files. 
To prevent issues down the road it's best to clean up the old core/model directories and files.

### Related issue(s)/PR(s)
none


Mark-H:
That's a recursive function, so just specifying `core/model/modx` should do fine (and be easier to grasp).

rthrash:

https://community.modx.com/t/modx3-nightly-installed-now-manager-displays-error/1705/9

JoshuaLuckers:
@Mark-H there is 1 file that should stay there, modx.class.php. That’s why I didn’t delete the whole folder.

Mark-H:
Ah, right. Forgot about that.

Mark-H (approved):

opengeek (approved):

PR #14860, opened 2019-11-30
Add missing attribute "unsigned" in schema

andchir:
### What does it do?
Added missing attribute "unsigned" in schema and re-generate classes.

### Why is it needed?
See issue description.

### Related issue(s)/PR(s)
#12419 #14745

Issue #12419, opened 2015-04-14
Schema int type mismatched due to attributes="unsigned" not being placed on Foreign Keys


wshawn:
At first I thought this was an isolated occurrence, but have found it all over the place modx.mysql.schema.xml.

Examples:
### modUserProfile

``` xml
<field key="internalKey" dbtype="int" precision="10" phptype="integer" null="false" index="unique" />
```

Which is a FK for modUser id which has attributes="unsigned".
### modResource

``` xml
<field key="parent" dbtype="int" precision="10" phptype="integer" null="false" default="0" index="index" />
```

Which is a FK for modResource id which has attributes="unsigned".

This can be found all over the schema, wherever a Foreign Key is matched to a xPdoSimpleObject Primary Key.

Essentially, any record over 2147483647 in the child relationships will be mismatched and probably cause errors. 

xPdoSimpleObject can have a maximum of 4294967295 due to it being unsigned. This leaves quite a lot of room for records to no longer be synced due to a simple error.

The entire schema requires review and correction with an emphasis on any relation to the 'id' primary key - even within the same class.

wshawn:
Still exists in 2.4.0

alroniks:
Potentially can break backward compatibility, so tagged into 3.x version.

JoshuaLuckers:
The PR to resolve this this issue has been merged into 3.x

PR #14902, opened 2019-12-15
PR: Add missing attribute "unsigned" in schema

andchir:
### What does it do?
Added missing attribute "unsigned" in schema and re-generate classes.

### Why is it needed?
See issue description.

### Related issue(s)/PR(s)
#12419 
Previous PR: #14860


Mark-H:
Excellent @andchir - thanks so much for recreating the pull request with just the right commits. :) 

@opengeek, is the `_build/build.schema.php` added here needed? From my standalone xPDO 3 usage I think there should be a `vendor/bin/xpdo parse-schema` call to use instead, but I'm not sure what that would be exactly for core, what parameters and such were used. If you have that command handy, I'll happily add that into the 3.x documentation to explain the process for schema changes for future contributors.

opengeek:
@Mark-H no, that build.schema.php should not be added. However, I just tried to use the core/vendor/bin/xpdo script and it's not working at all for some reason. It's just repeatedly returning:

```<html><title>Error 503: Site temporarily unavailable</title><body><h1>Error 503</h1><p>Site temporarily unavailable; missing dependencies.</p></body></html>%```

Let me investigate and then I'll update you with the proper command for the docs and handle merging this.

PR #14905, opened 2019-12-16
Replace Pelago/Emogrifier to InlineStyle

bezumkin:
### What does it do?
Replaces Pelago/Emogrifier to InlineStyle

### Why is it needed?
While Emogrifier block us from use PHP 7.4 we need to replace it

### Related issue(s)/PR(s)
#14899

### DIFF
--- a/core/src/Revolution/Mail/modPHPMailer.php
+++ b/core/src/Revolution/Mail/modPHPMailer.php
@@ -13,7 +13,7 @@ namespace MODX\Revolution\Mail;
 
 use Exception;
 use MODX\Revolution\modX;
-use Pelago\Emogrifier;
+use InlineStyle\InlineStyle;
 use PHPMailer\PHPMailer\PHPMailer;
 
 /**
@@ -221,8 +221,10 @@ class modPHPMailer extends modMail
         try {
             if (strpos($this->mailer->ContentType, 'html') !== false) {
                 if (!empty($this->mailer->Body)) {
-                    $emogrifier = new Emogrifier($this->mailer->Body);
-                    $this->mailer->Body = $emogrifier->emogrify();
+                    $html = new InlineStyle($this->mailer->Body);
+                    /** @noinspection PhpParamsInspection */
+                    $html->applyStylesheet($html->extractStylesheets());
+                    $this->mailer->Body = $html->getHTML();
                 }
             }
             $sent = $this->mailer->send();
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
Is it wise to introduce a new dependency that looks unmaintained?

bezumkin:
@JoshuaLuckers I don't think that this dependency should be maintained actively.
I mean, what could be new for parsing CSS from one tag and putting it inline to another tags?

Anyway, we have other option - just to remove this function from core.
Screenshot 2019-12-16 at 22 42 26

alroniks:
@JoshuaLuckers I think there is an example of an advantage using a composer when another one can easily replace one tool.

Jako:
Switching to a probably unmaintained but working repository is IMHO a straightforward solution. If the code changes to switch to that dependency are that minimal, I am fine with the solution.

Using an active repository of maintainers, that don't have understand the version constraint rules of composer, should be avoided. All the badges in the MyIntervals/emogrifier repository don't hide this strange composer rules.

core/src/Revolution/Mail/modPHPMailer.php
```diff
@@ -221,8 +221,10 @@ public function send(array $attributes = [])
         try {
             if (strpos($this->mailer->ContentType, 'html') !== false) {
                 if (!empty($this->mailer->Body)) {
-                    $emogrifier = new Emogrifier($this->mailer->Body);
-                    $this->mailer->Body = $emogrifier->emogrify();
+                    $html = new InlineStyle($this->mailer->Body);
+                    /** @noinspection PhpParamsInspection */
```
opengeek, line 225: What is the purpose of this @noinspection tag?

bezumkin: `applyStylesheet` method has wrong string attribute in phpdoc but `extractStylesheets` returns array.

So PhpStorm warn about it, and I suppressed it by `@noinspection`

JoshuaLuckers: Would be useful to have this information in the code as well. 😄

opengeek: Interestingly, that invalid phpdoc is fixed in master branch since May 2018 (see https://github.com/christiaan/InlineStyle/commit/e7dbf8830507d63c032285ae34b2d4e947f99515). That gives me some concern, though not that much if the thing works. I'll see if I can't get someone to create a new release.

opengeek: Going to move forward with this for now though…

opengeek (approved):

PR #14906, opened 2019-12-16
Fixed transparency in .png for TV

Ruslan-Aleev:
### What does it do?
This PR is for 3.x. For branch 2.x, corrections were applied in MODX 2.7.2.

Pretty old bug with transparency in .png for TV. In general, png with white edges is not noticeable in the standard manager area, because the background is white, but sometimes artifacts appear on the png background.

Also made the code a little clearer.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14664

### DIFF
--- a/manager/templates/default/element/tv/renders/input/image.tpl
+++ b/manager/templates/default/element/tv/renders/input/image.tpl
@@ -1,69 +1,71 @@
 <div id="tv-image-{$tv->id}"></div>
+
 <div id="tv-image-preview-{$tv->id}" class="modx-tv-image-preview">
-    {if $tv->value}<img src="{$_config.connectors_url}system/phpthumb.php?w=400&h=400&aoe=0&far=0&src={$tv->value}&source={$source}" alt="" />{/if}
+    {if $tv->value}<img src="{$_config.connectors_url}system/phpthumb.php?w=400&h=400&aoe=0&far=0&f=png&src={$tv->value}&source={$source}" alt="" />{/if}
 </div>
+
 {if $disabled}
-<script type="text/javascript">
-// <![CDATA[
-{literal}
-Ext.onReady(function() {
-    var fld{/literal}{$tv->id}{literal} = MODx.load({
-    {/literal}
-        xtype: 'displayfield'
-        ,tv: '{$tv->id}'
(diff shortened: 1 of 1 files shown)


opengeek (commented):
I'm a little confused on this. The changes you referenced for 2.7.2 via #14664 don't appear to be anywhere near the same. Can you elaborate on how this is the same fix as was applied in #14664 ?

Ruslan-Aleev:
I did not quite understand the comment. The changes are essentially the same - the `&f=png` parameter has been added. And indentation to make the code more readable, but indentation is not important.

JoshuaLuckers:
The indentation changes make it harder to see what has actually changed and make it harder to compare with the changes from #14664.

Ruslan-Aleev:
I can remove the indentation. Return as it was?

opengeek:
I'm not worried about the indentation. But there was something changed in another file to do with `--flatten` and I thought the change in 2.x involved the ra parameter. But I could just be confused, as I said.

Ruslan-Aleev:
I made a mistake in first commit in PR https://github.com/modxcms/revolution/pull/14664/, and edited `--flatten`, but in the end, after the tests, I returned as it was for  `--flatten` (just correcting the link in the comment).
Those adding the parameter `&f=png` is enough.

opengeek:
Ah, ok. In the future, we could avoid that kind of confusion if you would remove (via interactive rebase) or squash commits like that out of your PRs and force push them so that only the relevant changes are part of it. I'm going to proceed with merging this.

Ruslan-Aleev:
> In the future, we could avoid that kind of confusion if you would remove (via interactive rebase) or squash commits like that out of your PRs and force push them so that only the relevant changes are part of it.

I did not think that this is very necessary. The final changes are displayed in the files, everything is clear there.

opengeek:
And I think it is. The idea is to keep the history clean on the main branch and if you keep unnecessary commits in your PRs, all of those show up in the history when they are merged. So, pretty please? With sugar on top?

Ruslan-Aleev:
Yes, the logic is clear, thanks. I will closely monitor the commits.

PR #14912, opened 2019-12-17
Move deprecated class_aliases out of composer and clean up entry points

opengeek:
### What does it do?
Having core/includes/deprecated.php loaded via composer was breaking any use of core/vendor/bin/xpdo because it was forcing the loading of modX before a core.config.php was loaded. This should never be allowed. To make the process more strict, I am moving MODX_CORE_PATH and autoloading out of modX class. Autoloading and loading config.core.php MUST always come first.

### Why is it needed?
To fix problems regenerating the model classes and make entry points more consistent.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,10 +75,7 @@
     "autoload": {
         "psr-4": {
             "MODX\\": "core/src/"
-        },
-        "files": [
-          "core/include/deprecated.php"
-        ]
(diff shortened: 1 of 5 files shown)


JoshuaLuckers (approved):
Don't forget to run `composer dump-autoload`.

PR #14920, opened 2019-12-19
Add a DI container and allow namespaces to define a bootstrap.php

opengeek:
### What does it do?
Adds a DI container extended from Pimple (https://github.com/silexphp/Pimple) for use by MODX components and allows any component which defines a modNamespace to provide a bootstrap.php in the component `path` which is executed on initialization of a modX context.

### Why is it needed?
For 3.x, MODX components need a way to add their paths to the autoloader dynamically. By allowing components to define a bootstrap.php file which is executed when modX is initialized, they can take care of registering their source path(s) with the autoloader, add any custom model packages, add a service class definition to the DI container, and just about anything else they might need to do. This will allow extension packages and the getService() method to be completely deprecated for 3.x in favor of using the DI container and autoloading.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/core/src/Revolution/Services/NotFoundException.php
+++ b/core/src/Revolution/Services/NotFoundException.php
@@ -1,0 +1,13 @@
+<?php
+
+
+namespace MODX\Revolution\Services;
+
+
+use Exception;
+use Psr\Container\NotFoundExceptionInterface;
+
+class NotFoundException extends Exception implements NotFoundExceptionInterface
+{
+
+}
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "erusev/parsedown": "^1.7",
         "inlinestyle/inlinestyle": "^1.2",
         "simplepie/simplepie": "^1.5",
+        "pimple/pimple": "^3.0",
         "ext-curl": "*",
         "ext-dom": "*",
         "ext-gd": "*",
--- a/core/.gitignore
+++ b/core/.gitignore
@@ -1,3 +1,4 @@
 .htaccess
 /components/
 /vendor/
+/bootstrap.php
--- a/core/src/Revolution/Services/Container.php
+++ b/core/src/Revolution/Services/Container.php
@@ -1,0 +1,38 @@
+<?php
+
+
+namespace MODX\Revolution\Services;
+
+
+use Psr\Container\ContainerInterface;
+
+class Container extends \Pimple\Container implements ContainerInterface
+{
+    public function add($id, $value)
+    {
+        $this->offsetSet($id, $value);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function get($id)
+    {
+        if ($this->has($id)) {
+            try {
+                return $this->offsetGet($id);
+            } catch (\Exception $e) {
+                throw new ContainerException($e->getMessage(), $e->getCode(), $e);
+            }
+        }
+        throw new NotFoundException("Dependency not found with key {$id}.");
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function has($id)
+    {
+        return $this->offsetExists($id);
+    }
+}
--- a/core/src/Revolution/Services/ContainerException.php
+++ b/core/src/Revolution/Services/ContainerException.php
@@ -1,0 +1,13 @@
+<?php
+
+
+namespace MODX\Revolution\Services;
+
+
+use Exception;
+use Psr\Container\ContainerExceptionInterface;
+
+class ContainerException extends Exception implements ContainerExceptionInterface
+{
+
+}
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -26,6 +26,7 @@ if (!file_exists(MODX_CORE_PATH . 'vendor/autoload.php')) {
 require_once MODX_CORE_PATH . 'vendor/autoload.php';
 
 use Exception;
+use MODX\Revolution\Services\Container;
 use MODX\Revolution\Error\modError;
 use MODX\Revolution\Error\modErrorHandler;
 use MODX\Revolution\Mail\modMail;
@@ -74,6 +75,10 @@ class modX extends xPDO {
      * @const SESSION_STATE_EXTERNAL
      */
     const SESSION_STATE_EXTERNAL = 2;
+    /**
+     * @var Container The DI services container.
+     */
+    public $services;
     /**
      * @var modContext The Context represents a unique section of the site which
      * this modX instance is controlling.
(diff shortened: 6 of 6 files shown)


opengeek:
As an example, consider the Collections extra, which currently requires an extension package definition and instantiates a service class in a plugin. The bootstrap.php for Collections could look like: 

```
<?php
\MODX\Revolution\modX::getLoader()->addPsr4('Collections\\', $namespace['path'] . 'src/');
$modx->addPackage('Collections\Model', $namespace['path'] . 'src/', null, 'Collections\\');
$modx->services->add('collections', function($c) use ($modx) {
    return new Collections\Collections($modx);
});
```

A simple component that just needed to register itself with the autoloader:
```
<?php
\MODX\Revolution\modX::getLoader()->addPsr4('My\\Component\\', $namespace['path'] . 'src/');
```

JoshuaLuckers:
#needsdocumentation 😋

core/src/Revolution/Services/NotFoundException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+
+namespace MODX\Revolution\Services;
+
+
+use Exception;
+use Psr\Container\NotFoundExceptionInterface;
+
+class NotFoundException extends Exception implements NotFoundExceptionInterface
```
Mark-H, line 10: Perhaps this should extend the `ContainerException`, so a single try/catch is enough to wrap a call to `Container->get()`

JoshuaLuckers: https://www.php-fig.org/psr/psr-11/#33-psrcontainernotfoundexceptioninterface

JoshuaLuckers: See: https://www.php-fig.org/psr/psr-11/#31-psrcontainercontainerinterface

Mark-H: I meant for the MODX-specific implementation (class), though it's good to know the interface does extend.

JoshuaLuckers: Both exceptions have a different purpose:
```
 * @throws NotFoundExceptionInterface  No entry was found for **this** identifier.
 * @throws ContainerExceptionInterface Error while retrieving the entry.
```

opengeek: So can you catch `ContainerExceptionInterface`? Or catch `\Exception` and test if it is an instance of `ContainerExceptionInterface`?

Mark-H: I'm mostly thinking along these lines:

```
try {
   $someservice = $container->get('someservice');
} catch (??? $e) {
   log('install someservice first or whatever');
}
```

In such a case, I don't care wether it's a not found or an internal error to the container or an error in the instantiation of the dependency - all I want to know is, did the call succeed, was my service available.

Using `ContainerExceptionInterface` would definitely work, but that makes the code reliant on the dependency. If we're making available MODX-specific implementations, it would seem safer to look for that specifically. But in the current situation that would mean either doubling up the `catch` to cover both cases (which causes the boilerplate code to double in size for each service you're looking up), or catching the root `\Exception` (which seems like a waste of more detailed implementations). 

I don't necessarily have the answer but think this is the type of architectural detail that's worth thinking about, as it could become a pattern that is implemented into other code as it gets refactored. 

My gut says NotFoundException should be an instance of ContainerException so they can be caught in a single catch statement, and that we should use that instead of code from a dependency directly. I've had issues in Commerce where a dependency update caused breaking changes to extensions because it was exposed directly to users; wrapping that in a Commerce-specific class helps avoid such lock in in the future.

But, as the dependency here is a well-established PSR standard, I can also imagine we'd prefer referring to those directly.

opengeek: Yes, I want to stick to the PSR standard. We are 100% in control of the PSR-11 extension of Pimple here. I can guarantee that the PSR-11 interfaces here will never change. They will use proper semver if they do and it will not affect us unless we choose to adopt them.

As for NotFoundException extending ContainerException, can I even do that if they both implement ContainerInterfaceException? I'm not even sure if that is proper architecture.

JoshuaLuckers: If you don't care about it being a `NotFoundExceptionInterface` or `ContainerExceptionInterface` why not simply catch `Exception`?

theboxer (approved):

theboxer (approved):

theboxer (approved):

core/src/Revolution/Services/NotFoundException.php
opengeek: Any consensus here? Can we move forward with this as-is and address the concern with the exceptions post-haste?

theboxer: I'd go with as-is. Catching `\Exception` seems fine to me.

JoshuaLuckers (approved):

core/src/Revolution/Services/NotFoundException.php
Mark-H: I'm okay with catching the PSR interfaces, but want to play with the PR a little before giving formal approval if that's alright.

opengeek: I'll give you two hours! That enough? LOL!

Mark-H: Ha-ha. If we're going to use this as a primary integration method for the next 5-10 years, why rush?

JoshuaLuckers: One of my PRs is also delayed because we need to come to a consensus first 😅😋

opengeek: Because no one can actually build an extra for 3.x without this?

opengeek: Well thumbs down to you to, buddy! 😅

Mark-H (approved):

Mark-H:
I've stubbed out docs:

- https://docs.modx.com/3.x/en/extending-modx/namespaces#bootstrapping-services
- https://docs.modx.com/3.x/en/extending-modx/di-container
- https://docs.modx.com/3.x/en/extending-modx/di-container/config

I'm not sure if any other services should be available in the container already, only the config appears to be available from within the `bootstrap.php`. Might be worth investigating that separately to see if there might need to be some changes in the ordering to allow other services to be available in the bootstrap.

JoshuaLuckers:
Would be nice to have unit tests for this in the future

PR #14921, opened 2019-12-20
Call getInstance instead of creating directly new instance of processor

theboxer:
### What does it do?
Replaces creating a direct instance of a processor class with calling `getInstance` method. As that's how instance of processor should be created :)

### Why is it needed?
CRCs for example Collections have own Update/Create processors that extend those core ones. MODX know about them only through the `getInstance` implementation on the `MODX\Revolution\Processors\Resource\Update(/Create)` class, that tries to check if there's a different processor for the `class_key`. This doesn't work atm.

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1724,7 +1724,7 @@ class modX extends xPDO {
         // We're limiting to subclasses of Processor to prevent instantiating arbitrary classes
         if (static::isProcessorClass($action)) {
             /** @var Processor $processor */
-            $processor = new $action($this, $scriptProperties);
+            $processor = call_user_func_array([$action, 'getInstance'], [&$this, $action, $scriptProperties]);
 
             return $processor->run();
         }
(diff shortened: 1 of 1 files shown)


core/src/Revolution/modX.php
```diff
@@ -1724,7 +1724,7 @@ public function runProcessor($action = '', $scriptProperties = array(), $options
         // We're limiting to subclasses of Processor to prevent instantiating arbitrary classes
         if (static::isProcessorClass($action)) {
             /** @var Processor $processor */
-            $processor = new $action($this, $scriptProperties);
+            $processor = call_user_func_array([$action, 'getInstance'], [&$this, $action, $scriptProperties]);
```
Mark-H: Perhaps `$action::getInstance($this, $action, $scriptProperties)` would be an additional improvement?

theboxer: `$this` is passed with the `&` in the `call_user_func_array`, it didn't work when passing it directly to `getInstance`, so I used the same notation as at the bottom of the method.

Mark-H: `&` shouldn't be needed when passing directly to a method (cause objects are always passed as reference).. but fair enough.

theboxer: yup, figured that out :)

Mark-H (approved):

Mark-H:
I've merged and also tweaked the third instantiation in the runProcessor method to use it directly vs `call_user_func`.

PR #14927, opened 2019-12-20
Remove modAction and legacy controller support

Mark-H:
### What does it do?

Removing modAction and relevant legacy, which was deprecated in v2.2. Modernise the modManagerResponse and controller interaction. Improve stuff.

### Why is it needed?

Code cleanup. 

### Related issue(s)/PR(s)

Documentation update: https://github.com/modxorg/Docs/pull/160

### To do

- [x] Fix login; currently the modManagerResponse handling is _too_ strict and it wont show the login page
- [x] Figure out what to do with setup/includes/upgrade.install.php
- [x] Add tests to prove the class name handling is correct and can find classes 
- [x] Add removed files to the cleanup migration script 
- [x] Improve error pages
- [x] Check if actions lexicon can be removed, and if so, remove it 
- [x] Remove relevant settings - cache_action_map, others?
- [ ] (optional) Squash all commits either in PR or as part of merge
- [x] Remove modManagerControllerDeprecated


Ruslan-Aleev:
Perhaps it’s worth deleting lexicons as well - https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/action.inc.php

JoshuaLuckers (commented):
I see `file_exists` and `is_readable` in the same if statements. `is_readable` checks if the file exists and is readable so there is no need to check both.

core/src/Revolution/modManagerResponse.php
```diff
@@ -247,30 +213,74 @@ public function checkForMenuPermissions($action)
...
+            }
+
+            if (file_exists($controllersPath . strtolower($action) . '/index.class.php')) {
+                $fullPath = $controllersPath . strtolower($action) . '/index.class.php';
+                break;
+            }
+        }
+
+        // If the file exists, require it, while discarding any content from it
+        if (file_exists($fullPath) && is_readable($fullPath)) {
```
JoshuaLuckers:
```suggestion
        if (is_readable($fullPath)) {
```
> is_readable — Tells whether a file exists and is readable

JoshuaLuckers:
Is `modManagerControllerDeprecated` being removed by this PR as well?

Mark-H:
I think this should now be ready for review and testing. Documentation updates are pending at https://github.com/modxorg/Docs/pull/160, though there are probably more references to modAction in the docs that should be updated.

Mark-H:
@cla-bot check

JoshuaLuckers:
I'm not too familiar with `modAction*` classes and I was wondering how the classes below are related to `modAction` and if they should stay or not:
- modAccessActionDom 
- modActionDom
- modActionField

Mark-H:
Iirc, modActionDom is the "regions" tab in form customization. modActionField the individual fields you can enable/move. Don't think modAccessActionDom is implemented, but supposed to be related to the regions.

opengeek:
Any chance you can squash this PR and force push @Mark-H ?

Issue #14929, opened 2019-12-21
MODX 3 alpha1 - icons visable not actions


OlegShchavelev:
## Bug report

The navigation icon is displayed if there is no context menu.

![изображение](https://user-images.githubusercontent.com/20732384/71304954-5391d700-23de-11ea-90ab-49b6ce6a1e81.png)

### Environment
MODX version, apache/nginx and version, mysql version, browser, etc. Any relevant information.

Ruslan-Aleev:
> The list of package versions does have a context menu - only the most recent version listed does not. So removing the icon is not a proper fix here.

@Mark-H Perhaps we can close this issue? Because there is, in fact, no bug.

Although the components, it seems to me, are more often updated than not and maybe it’s worth hiding the gear-icon? Not sure.

JoshuaLuckers:
We have to investigate whether it is possible to hide/show the gear-icon per row and maybe even what "actions" it shows. For packages that don't have any actions we should hide the gear-icon.

Ibochkarev:
The designated problem remains relevant. Tested on the latest version MODX Revolution 3.0.0-alpha3 (git)

![image](https://user-images.githubusercontent.com/2138260/93712996-0c0dc680-fb7b-11ea-8fbb-2be2bade72b7.png)

smg6511:
This is still quite relevant and there are actually issues in quite a few places, in terms of menu options, buttons, and the context and actions menus not accurately reflecting what one has permissions to do. I've been working on this over the past few days now and will submit a PR in the near future.

@JoshuaLuckers - I've figured out how to handle the actions (gear icon) menu visibility on a per-row basis. But in addition to that issue, none of the other avenues of editing/creating (the column-level editors, the create buttons, the batch [delete] buttons, etc.) are in sync with the permissions level a user has. So I'm sifting through that. 

The real tricky thing is in the Roles grid: We have two (what I'm calling) protected options (_Super User_ and _Member_) that should never be editable, whereas the editability of rest of the grid's options is permissions-dependent. There's no built-in way to enable/disable the editor(s) on a row-by-row basis, so that's proving to require a creative solution—one that's in progress.

[...]

PR #14934, opened 2019-12-22
Disable actionsColumn on package version grid

Ruslan-Aleev:
### What does it do?
Disable actionsColumn on package version grid.
Also I brought the code to the format, as in other grids files.

### Why is it needed?

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14929


JoshuaLuckers (approved):

Mark-H (changes requested):
The list of package versions does have a context menu - only the most recent version listed does not. So removing the icon is not a proper fix here.

PR #14935, opened 2019-12-23
Temporary solution for long dropdown submenu and on mobile width

### LINKED ISSUE #14928: MODX 3 alpha1 -  style limitation package menu
## Bug report

Styles of the menu item "Packages" is not adapted to accommodate a large number of packages.

I understand that the menu items can be moved using the settings. But in my opinion, the administrator should have a choice. How can he arrange the menu items.

Scale 100%

изображение

Scale 67%

изображение

### Environment
MODX version, apache/nginx and version, mysql version, browser, etc. Any relevant information.

GulomovCreative:
### What does it do?
I added overlay bg block for submenus not for beauty, but in order to solve the problem with overlong dropdown submenu. And fixed it on mobile width. This is a temporary solution because I believe that the whole menu needs to be refactor, but I do not have time. I will refactor later.

photo_2019-12-23_21-27-57

screencast-www figma com-2019 11 03-02_20_31

### Why is it needed?
To make it normal

### Related issue(s)/PR(s)
Closes #11939 and closes #14928

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -74,7 +74,8 @@
   #modx-manager-search-icon a,
   #modx-leftbar-trigger a,
   #modx-user-menu a {
-    padding: 12px 0;
+    width: auto;
+    padding: 12px;
   }
 
   #modx-topnav {
@@ -142,9 +143,7 @@
     }
 
     #limenu-user a {
-      display: flex;
-      flex-direction: column;
-      justify-content: center;
+      line-height: 0;
     }
   }
 
@@ -215,6 +214,7 @@
     box-sizing: border-box;
     list-style: none;
     position: absolute;
+    left: 0;
     z-index: 10000;
     opacity: 0;
     visibility: hidden;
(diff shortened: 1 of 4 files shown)


Ruslan-Aleev:
Unfortunately, this solution for the drop-down menu has several disadvantages:
1) Scroll works only for the first-level submenu, for the rest of the submenu levels everything will also be cropped by the screen.
2) There is no scroll itself (panel on the right) and it may not be clear to the user that the menu can be scrolled (this is because overflow cannot be set).

Personally, the solution with the maximum number of submenus seems more universal (with some fixes) - https://github.com/modxcms/revolution/pull/14300

Both PR were implemented by @GulomovCreative , for which many thanks to him! :)

JoshuaLuckers:
> Personally, the solution with the maximum number of submenus seems more universal (with some fixes) - #14300
> 
> Both PR were implemented by @GulomovCreative , for which many thanks to him! :)


I agree, @GulomovCreative it would be awesome to have #14300 in 3.x instead! Do you have time to recreate that PR for 3.x?

Mark-H:
@cla-bot check

GulomovCreative:
>I agree, @GulomovCreative it would be awesome to have #14300 in 3.x instead! Do you have time to recreate that PR for 3.x?

#14300 That decision is disgusting, although it is mine, therefore, I will not recreate it. Menu need refactored

GulomovCreative:
If I have time, I will rewrite this long-suffering menu. Thx you all

PR #14937, opened 2019-12-26
Disable actionsColumn for FC grids

Ruslan-Aleev:
### What does it do?
Disable actionsColumn for FC grids.
fc_1

fc_2

fc_3

Also reduced the width of the ID column on the "Regions" tab (200 wider than necessary).

### Why is it needed?

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/assets/modext/widgets/fc/modx.panel.fcset.js
+++ b/manager/assets/modext/widgets/fc/modx.panel.fcset.js
@@ -231,6 +231,7 @@ MODx.grid.FCSetFields = function(config) {
     });
     Ext.applyIf(config,{
         id: 'modx-grid-fc-set-fields'
+        ,showActionsColumn: false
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/fc/modx.panel.fcset.js
```diff
@@ -298,14 +299,15 @@ MODx.grid.FCSetTabs = function(config) {
...
+        ,showActionsColumn: false
         ,fields: ['id','action','name','form','other','rank','visible','label','type']
         ,autoHeight: true
         ,plugins: [this.vcb]
         ,stateful: false
         ,columns: [{
             header: _('tab_id')
             ,dataIndex: 'name'
-            ,width: 200
+            ,width: 100
```
JoshuaLuckers, line 310: This should be in a separate commit.

Ruslan-Aleev:
Corrected commits.

JoshuaLuckers (approved):

PR #14938, opened 2019-12-26
Disable actionsColumn for ACL grids

Ruslan-Aleev:
### What does it do?
Disable actionsColumn for ACL grids.
Also added comments.

acl_1

acl_2

### Why is it needed?

### Related issue(s)/PR(s)
None


JoshuaLuckers:
If you add a new role there are actions available. They are not available for the two default roles.

Ruslan-Aleev:
Yes, you're right, corrected the commit.

Mark-H (approved):

Issue #14512, opened 2019-03-28
Need refactoring of lexicon topics


Ruslan-Aleev:
### Summary
Lexicon topics contain unnecessary entries, for example, in the chunk.inc.php topic there is an entry with the key "chunk" - https://github.com/modxcms/revolution/blob/2.x/core/lexicon/en/chunk.inc.php#L9, but the same entry with the key "chunk" is in the topic https://github.com/modxcms/revolution/blob/2.x/core/lexicon/en/default.inc.php#L58 which is always included in MODX manager.
Why do we need 2 entries for one key?

There are also duplicate entries within the meaning, but with a different key - https://github.com/modxcms/revolution/blob/2.x/core/lexicon/en/default.inc.php#L18 and https://github.com/modxcms/revolution/blob/2.x/core/lexicon/en/default.inc.php#L272 is already a problem with incorrect key names, see PR https://github.com/modxcms/revolution/issues/14313.

There are unnecessary entries, for example, "Edit TV", "Edit chunk", but it is easier to use one "Edit" entrie 
for both TV and chunk. Why such specific entries for the element? There are not only extra lines in the topics, but errors in the code, i.e. used, for example, and "create" and "create_ELEMENT" at the same time, etc.

There are also entries that are written in the wrong topic.

We need to correctly distribute the entries in the topics, remove unnecessary and remove unused.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14313
https://github.com/modxcms/revolution/issues/13554
https://github.com/modxcms/revolution/issues/1188

Ruslan-Aleev:
In the code, lexicons are output in several ways:
- `[[%lexicon]]` (in .tpl)
- `_('lexicon')` (in .js)
- `return $this->modx->('lexicon')` (in .php)
- `return $this->modx->lexicon($this->objectType . '_err_nf')` (dynamically in .php, mainly for errors), for example - https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/Processors/Element/Create.php#L57
- maybe there are still output options
---
Quickly checking the lexicons, it seems to me that about 40-50% (so many) of the lexicons are not used.
And this is extremely unpleasant for working with translation.

Jako:
I have created a lexicon checker in GPM, that has a few regular expressions to cover all the strings you mentioned. Maybe you could check the MODX core with a similar method?

[...]

PR #14940, opened 2019-12-28
Lexicon check for the core

Jako:
### What does it do?
Add a lexicon check for the core. Cross check the core files with the existing lexicon entries for the usage of a lexicon entry.

### Why is it needed?
To detect easier, which lexicon entries are used.

The script creates 3 files in the _build/lexicon folder. The files are containing missing and superfluous entries and the third one collects usages of the lexicon with variable generated keys (not all are currently collected - only the ones starting with a single/double quote key). The lexicon with variable generated keys can't be copied to a lexicon, so this generated code does not have to be valid - the other ones could be copied.

### Related issue(s)/PR(s)
#14512


Jako:
At the moment the variable object type errors of the core processors are not cross checked i.e. created by `$this->objectType.'_err_ns'` code. The used object types of the core processors could be collected and the the lexicon entries could be cross checked afterwards.

Jako:
The check is currently created on 2.x. It should be valid for 3.x too with a few changes.

JoshuaLuckers:
If connecting to the database fails the script will continue to execute. I don't think this is supposed to happen.
Afterwards 3 new files are created:
- _missing.php
- _superfluous.php
- _variable.php

It's a bit unclear to me what the content of each file means. The syntax of the generated content of `_variable.php` is invalid:
```
<?php
$_lang['area_'.$object->get('area'] = '';
$_lang['area_'.$object->get('area'] = '';
$_lang['db_info_' + MODx.config.dbtype)+'</p>'] = '';
$_lang['refresh_' . $partition) . ': ' . $this->modx->lexicon('refresh_' . ($result ? 'success' : 'failure'] = '';
$_lang['refresh_' . ( $result ? 'success' : 'failure'] = '';
$_lang['refresh_' . ($ctxResult ? 'success' : 'failure'] = '';
```

Jako:
The script crosschecks only files, so there is no need for a DB connection.

The three files are containing missing and superfluous entries and the third one collects usages of the lexicon with variable generated keys (not all are currently collected - only the ones starting with a single/double quote key). The lexicon with variable generated keys can't be copied to a lexicon, so this generated code does not have to be valid - the other ones could be copied.

JoshuaLuckers:
It would be nice to have this documented in the script and/or files it generates.

Mark-H:
@cla-bot check

Issue #12293, opened 2015-01-24
Wanted: A new global xtype to enter small JSON key/value objects


Jako:
It would be nice to have a new xtype that is displayed as a small grid and contains small JSON key/value objects. A grid type system setting or snippet property could be created that way, not nested xPDO where clauses could be clicked together by non techies …

oori:
xtype "propertygrid" does that.     although it isn't simple, you normally have to care for two-way sync into a hidden string field (using "propertychange" and "_updated" listeners, and encoding/decoding JSON)  and more UI related hacks due to vertical space, etc..

```
/**
 * Update window
 */
OO.window.UpdateSurvey = function(config) {
    config = config || {};
    Ext.applyIf(config,{
        ooType:'survey'
        ,width: 650
        ,autoScroll: true
        
        ,fields: [{
            xtype: 'hidden'
            ,name: 'id'
        },{
        xtype:'fieldset',
        title: _('oo.update'),
        collapsible: true,
        collapsed: false,
        items :[
            {
                xtype: 'oo-combo-role'
                ,fieldLabel:_('oo.survey.role')
                ,name: 'role'
                ,anchor: '100%'
            },{
                xtype: 'oo-combo-survey'
                ,fieldLabel: _('oo.survey.id')
                ,name: 'survey'
                ,anchor: '100%'
            },{
                xtype: 'textfield'
                ,fieldLabel:_('oo.survey.title')
                ,name: 'title'
                ,anchor: '100%'
                ,id: 'surveyTitle_update'
            },{
                xtype: 'textfield'
                ,fieldLabel: _('oo.survey.lastSync')
                ,name: 'lastSync'
                ,anchor: '100%'
            },{
                xtype: 'hidden'
                ,fieldLabel:  _('oo.survey.mappedQuestions')
                ,name: 'mappedQuestions'
                ,id: 'mappedQuestions'
                ,anchor: '100%'
                ,listeners: {
                    '_updated': {fn:function(o){
                        var v = Ext.util.JSON.decode(o.getValue());
                        Ext.getCmp('mappedQuestions_grid').setSource(v);
                    },scope:this}
                }
            }]
        },{
            xtype: 'propertygrid'
            ,name: 'mappedQuestions_grid'
            ,id: 'mappedQuestions_grid'
            ,anchor: '100%'
            ,collapsed: true
            ,collapsible: true
            ,autoHeight: true
            ,autoSroll: true

            ,title:  _('oo.survey.mappedQuestions')
[...]

PR #14943, opened 2020-01-02
JSON grid xtype

Jako:
### What does it do?
Add a JSON grid xtype

### Why is it needed?
Use a simple grid editor for JSON values in i.e. the system settings.

It could be configured a bit with the fieldConfig property, containing an array of field configs 

```
fieldConfig: [{
    name: 'whatever', // required, name 'id' is reserved and can't be used
    width: 100, // defaults to 100
    xtype: 'textfield', // defaults to textfield
    allowBlank: true, // defaults to true
    header: _('whatever') // defaults to the name value of the current field config
}];
```

If there is no fieldConfig property set, the following default fieldConfig is used:
```
[{name: 'key'}, {name: 'value'}]
```

### Related issue(s)/PR(s)
#12293

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -1160,7 +1160,7 @@ Ext.grid.PropertyColumnModel=function(a,b){var g=Ext.grid,f=Ext.form;this.grid=a
  * It could be configured with the fieldConfig property, containing an array of field configs:
  *
  * fieldConfig: [{
- *    name: 'whatever', // required
+ *    name: 'whatever', // required, name 'id' is reserved and can't be used
  *    width: 100, // defaults to 100
  *    xtype: 'textfield', // defaults to textfield
  *    allowBlank: true, // defaults to true
@@ -1190,6 +1190,7 @@ MODx.grid.JsonGrid = function (config) {
         hidden: true
     });
     this.fieldConfig = config.fieldConfig || [{name: 'key'}, {name: 'value'}];
+    this.fieldConfig.push({name: 'id', hidden: true});
     this.fieldColumns = [];
     this.fieldNames = [];
     Ext.each(this.fieldConfig, function (el) {
@@ -1198,6 +1199,7 @@ MODx.grid.JsonGrid = function (config) {
             header: el.header || _(el.name),
             dataIndex: el.name,
             editable: true,
+            hidden: el.hidden || false,
             editor: {
                 xtype: el.xtype || 'textfield',
                 allowBlank: el.allowBlank || true,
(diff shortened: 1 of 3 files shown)


manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -1152,3 +1152,188 @@ Ext.ns('Ext.ux.grid');Ext.ux.grid.CheckColumn=function(a){Ext.apply(this,a);if(!
...
+ * MODx JSON Grid
+ */
+MODx.grid.JsonGrid = function (config) {
+    config = config || {};
+    this.ident = config.ident || 'jsongrid-mecitem' + Ext.id();
+    this.buttonColumnTpl = new Ext.XTemplate('<tpl for=".">'
+        + '<tpl if="action_buttons !== null">'
+        + '<ul class="action-buttons">'
+        + '<tpl for="action_buttons">'
+        + '<li><i class="icon {className} icon-{icon}" title="{text}"></i></li>'
```
Mark-H: Could use some html encoding on these values to make it XSS-proof by default

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -1152,3 +1152,188 @@ Ext.ns('Ext.ux.grid');Ext.ux.grid.CheckColumn=function(a){Ext.apply(this,a);if(!
...
+    });
+    this.fieldConfig = config.fieldConfig || [{name: 'key'}, {name: 'value'}];
+    this.fieldColumns = [];
+    this.fieldNames = [];
+    Ext.each(this.fieldConfig, function (el) {
+        this.fieldNames.push(el.name);
+        this.fieldColumns.push({
+            header: el.header || _(el.name),
+            dataIndex: el.name,
+            editable: true, editor: {
```
Mark-H: [minor] Missing a linebreak

Mark-H (dismissed):
From a code review, only a minor potential XSS found, other than that this looks very promising!

manager/assets/modext/widgets/core/modx.grid.js
Jako: Very minor, since the template variables of the ExtJS template are not set by user input. Only directly by the script itself.

Jako:
The requested changes are done and some documentation is added at the top of the new xtype code.

JoshuaLuckers:
Shouldn't this target 3.x instead of 2.x?

Jako:
It should work with both versions.

Jako:
Rebasing does not work well at the moment. Could that be cherry picked?

Mark-H:
If rebasing doesn't work, cherry picking likely does not either. AFAIK, rebasing is much better and more powerful than a cherry pick.

Jako:
My 3.x branch is not at the current state of the modxcms one. That could be the issue with rebasing here on GitHub.

Jako:
No It isn't.

Jako:
I have recreated this on 3.x

Issue #14390, opened 2019-02-15
Unsafe RSS feed


AgelxNash:
### Summary
Potential backdoor

### Step to reproduce
Change the value of settings **feed_modx_news** or **feed_modx_security**

### Observed behavior
- XSS
- XXE
- Social engineering

AgelxNash:
RSS feed also handles MODX tags. This is another attack vector.

Mark-H:
This doesn't really seem like a problem to me, but an administrator level functionality. 

The permissions needed to change settings would also allow you to, for example, change the manager theme or undo other protections that have a setting (upload_files, allow_tags_in_post, etc).

AgelxNash:
Ok:-) Then the security of **ALL** CMS installations depends on who can create post on https://forums.modx.com/board?board=7 or https://forums.modx.com/board?board=294 :hankey:

Mark-H:
Those are both restricted to the MODX team, so I'm okay with putting some trust in that. ;) 

Open to ideas though if you think there are worthwhile improvements. Sounds like changing the URLs is not what you're most worried about now?

AgelxNash:
I worry that access to the site == author of the rss feed

Mark-H:
If you have genuine security worries about these feeds, I would like to understand them. It's better to not discuss security concerns in a public place like this though, if you have a POC (or just a theory), could you please send it to me and the rest of the security team via the security mailing list (security@modx.com)?

AgelxNash:
PoC submitted
Related #14393

Ibochkarev:
@JoshuaLuckers  Wasn't that fixed in #14392 ?

JoshuaLuckers:
> @JoshuaLuckers Wasn't that fixed in #14392 ?

No, that was only to reduce a man in the middle attack.

Jako:
We have MODx.util.safeHtml since https://github.com/modxcms/revolution/pull/14453. Could this be used here?

meshkov:
I think we need to use in links with target="_blank":
`rel="noopener noreferrer"`

Ibochkarev:
@JoshuaLuckers Is it relevant?

PR #14946, opened 2020-01-07
Apply `MODx.util.safeHtml` to make the feed dashboards a bit safer

JoshuaLuckers:
### What does it do?
When loading the RSS feeds in the welcome dashboard apply `MODx.util.safeHtml` to the feed results only allowing the following tags:
`<h1><h2><h3><h4><span><div><ul><li><p><ol><dl><dd><dt><a><br><i><em><b><strong>`

### Why is it needed?
To make the feeds a bit safer to use.

### Related issue(s)/PR(s)
#14390

### DIFF
--- a/manager/assets/modext/widgets/modx.panel.welcome.js
+++ b/manager/assets/modext/widgets/modx.panel.welcome.js
@@ -51,17 +51,17 @@ Ext.extend(MODx.panel.Welcome, MODx.Panel,{
                     success: {
                         fn: function(response) {
                             if (response.success) {
-                                container.update(response.object.html);
+                                container.update(MODx.util.safeHtml(response.object.html, '<h1><h2><h3><h4><span><div><ul><li><p><ol><dl><dd><dt><img><a><br><i><em><b><strong>'));
                             }
                             else if (response.message.length > 0) {
-                                container.update('<p class="error">' + response.message + '</p>');
+                                container.update('<p class="error">' + MODx.util.safeHtml(response.message) + '</p>');
                             }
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/modx.panel.welcome.js
```diff
@@ -51,17 +51,17 @@ Ext.extend(MODx.panel.Welcome, MODx.Panel,{
                     success: {
                         fn: function(response) {
                             if (response.success) {
-                                container.update(response.object.html);
+                                container.update(MODx.util.safeHtml(response.object.html, '<h1><h2><h3><h4><span><div><ul><li><a><br><i><em><b><strong>'));
```
rthrash: Probably should add a few more tags to this as well: `<p><ol><dl><dd><dt>` and maybe ``

JoshuaLuckers: Done.

Mark-H (approved):
Not sure how safe img tags are, even with the `safeHtml` filter, but a welcome improvement nonetheless.

JoshuaLuckers:
> Not sure how safe img tags are, even with the `safeHtml` filter, but a welcome improvement nonetheless.

Not allowing them would be safer.

Jako:
safeHtml could also strip/allow tag attributes, if that helps.

opengeek:
@JoshuaLuckers does this also need to be applied to 3.x?

JoshuaLuckers:
@opengeek yes 😊

Issue #13491, opened 2017-07-07
Purge does not respect access policy #modxbughunt


wuuti:
### Summary
If you purge deleted documents, all documents in all contexts and resource groups which are marked as deleted are purged, no matter if you have access to the context or not. 

### Step to reproduce
Create a second context B and a user group with additional permissions (compared to Content Editor) of 'purge_deleted' only for that context. Create a new user and add it to the new user group. Remove the web context access policy from the new user group. Delete a document in the web context. Logout (or use a second browser) and login with the new restricted user. You should only see the new context B in the resource tree, as you don't have any access to the web context any more. Click the recycle bin icon and confirm the purge. The document in the web context is now purged.

### Observed behavior
The recycle bin icon is active and you can purge documents in contexts you don't have access to. 

### Expected behavior
It should _not_ be possible for user to change (and purging is a change with a huge impact!) documents he is not even allowed to see. So purging should only work on resources the user has access to. 
This is especially a concern in larger installations with different sites in different contexts and with a lot of editors working in their contexts. The issue is even more important as the confirmation alert does not even name the documents which are about to get purged (see related #13448 )

### Environment
MODX version 2.6.0-dev and prior versions

gpsietzema:
#modxbughunt #1point to @wuuti

Ibochkarev:
Fixed #14947 
@modxbot close

Jako:
I don't think that the issue is fixed with #14947. There was no additional check for allowed contexts of the current user added.

PR #14947, opened 2020-01-07
Update broken recycle bin getlist

markwillis82:
### What does it do?
If a resource has no parent (either by being deleted or hitting an ACL policy - the ) in that the resources parent has been removed but not the main resource, if the main resource is marked as deleted the recycle bin breaks with a 500 error as end() cannot get an object

### Why is it needed?
If a user has access to the recycle bin but not access to all resources (I think) - the recycle bin 500's because we can't always get the parents. This just stops the 500 error

### Related issue(s)/PR(s)
#13491

### DIFF
--- a/core/src/Revolution/Processors/Resource/Trash/GetList.php
+++ b/core/src/Revolution/Processors/Resource/Trash/GetList.php
@@ -119,8 +119,13 @@ class GetList extends GetListProcessor
         $parent = $objectArray['parent'];
 
         while ($parent != 0) {
-            $parents[] = $this->modx->getObject(modResource::class, $parent);
-            $parent = end($parents)->get('parent');
+            $o = $this->modx->getObject('modResource', $parent);
+            if($o) {
+                $parents[] = $o;
+            } else {
+                break;
+            }
+            if($parents) $parent = end($parents)->get('parent');
         }
 
         $parentPath = "";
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Processors/Resource/Trash/GetList.php
```diff
@@ -119,8 +119,13 @@ public function prepareRow(xPDOObject $object)
...
         while ($parent != 0) {
-            $parents[] = $this->modx->getObject(modResource::class, $parent);
-            $parent = end($parents)->get('parent');
+            $o = $this->modx->getObject('modResource', $parent);
+            if($o) {
+                $parents[] = $o;
+            } else {
+                break;
+            }
+            if($parents) $parent = end($parents)->get('parent');
```
Mark-H: Seems like this code could be simplified:

```
            $o = $this->modx->getObject(modResource::class, $parent);
            if ($o) {
                $parents[] = $o;
                $parent = $o->get('parent');
            } 
            else {
                break;
            }
```

Mark-H: Note that in MODX3 the `modResource::class` syntax should be used rather than `'modResource'`.

JoshuaLuckers: It would be nice if the name of the variable `$o` were to be changed to something more descriptive.

Mark-H:
@cla-bot check

Mark-H:
@markwillis82 Will you have a chance to make the suggested changes before the planned alpha2 release next week?

markwillis82:
This has been updated.

PR #14954, opened 2020-01-16
Created dedicated method to get a resource preview URL

rtripault:
### What does it do?
Extracted the "resource preview URL" computation into a dedicated method inside modResource class

### Why is it needed?
Previously we were having duplicate & inconsistent code to generate those preview URLs

### Related issue(s)/PR(s)
None found

### DIFF
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -456,35 +456,33 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
      * @return string
      */
     public function getPreviewUrl() {
-        $url = '';
-        if (!$this->get('deleted')) {
-            $this->xpdo->setOption('cache_alias_map', false);
-            $sessionEnabled = '';
-            /** @var modContextSetting|null $ctxSetting */
-            $ctxSetting = $this->xpdo->getObject(
-                'modContextSetting',
-                array(
-                    'context_key' => $this->get('context_key'),
-                    'key' => 'session_enabled'
-                )
-            );
+        if ($this->get('deleted')) {
+            return '';
+        }
+        $this->xpdo->setOption('cache_alias_map', false);
+        $sessionEnabled = '';
+        /** @var modContextSetting|null $ctxSetting */
+        $ctxSetting = $this->xpdo->getObject(
+            'modContextSetting',
+            array(
+                'context_key' => $this->get('context_key'),
+                'key' => 'session_enabled'
+            )
+        );
 
-            if ($ctxSetting) {
-                $sessionEnabled = $ctxSetting->get('value') == 0 ? array('preview' => 'true') : '';
-            }
-
-            $url = $this->xpdo->makeUrl(
-                $this->get('id'),
-                $this->get('context_key'),
-                $sessionEnabled,
-                'full',
-                array(
-                    'xhtml_urls' => false
-                )
-            );
+        if ($ctxSetting && $ctxSetting->get('value') == 0) {
+            $sessionEnabled = array('preview' => 'true');
         }
 
-        return $url;
+        return $this->xpdo->makeUrl(
+            $this->get('id'),
+            $this->get('context_key'),
+            $sessionEnabled,
+            'full',
+            array(
+                'xhtml_urls' => false
+            )
+        );
     }
 
     /**
--- a/core/model/modx/modresource.class.php
+++ b/core/model/modx/modresource.class.php
@@ -451,6 +451,43 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
     }
 
     /**
+     * Compute the "preview URL" for the resource
+     *
+     * @return string
+     */
+    public function getPreviewUrl() {
+        $url = '';
+        if (!$this->get('deleted')) {
+            $this->xpdo->setOption('cache_alias_map', false);
+            $sessionEnabled = '';
+            /** @var modContextSetting|null $ctxSetting */
+            $ctxSetting = $this->xpdo->getObject(
+                'modContextSetting',
+                array(
+                    'context_key' => $this->get('context_key'),
+                    'key' => 'session_enabled'
+                )
+            );
+
+            if ($ctxSetting) {
+                $sessionEnabled = $ctxSetting->get('value') == 0 ? array('preview' => 'true') : '';
+            }
+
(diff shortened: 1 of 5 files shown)


christianseel:
The issue was also, that when you enter the "update" controller of a resource you will get the `?preview=true` at the preview URL (when clicking the preview button), when sessions are disabled. Once you did changes and saved the resource, the preview url was regenerated by the update processor and the logic for the `?preview=true` was missing.

…hope that helps to understand the original issue that was solved here by romain!

rtripault:
Actually this is also happening when previewing a resource from the resource tree contextual menu ;)

Mark-H:
So, I understand it was this way before so it's not your fault or anything, but.... why isn't this using `$modx->getOption`/`$wctx->getOption` rather than fetching a context setting? That seems weird. 😬

christianseel:
As that's executed from the manager and you definitely want the setting value from the resources context, that would mean you have to get/load the context first, in order to use the `$ctx->getOption()`.

Loading just the setting instead of "preparing" the whole context feels reasonable to me. But I might miss something here, Mark? Cheers!

rtripault:
Sorry, i forgot to bring more context.

Some times ago, the ability to have "session-less contexts" has been introduced. The behavior to allow previewing resources in such contexts was to add a `preview=true` query parameter to the resource URL (added in `preview_url`).
This was added in the manager controller (resource update) and was working as expected.
However, as soon as you saved the resource, the manager populated the data from the response of the resource/update processor, which did not populate the `preview_url` attribute with the `preview=true` query parameter.

Some other places were also not taking care of `preview_true`, like listing the nodes of a context (for the resource tree).

So this PR aims to correct this, by adding `modResource::getPreviewUrl` and using it where needed.

Hope this helps better understand this PR.

Happy to try to bring more explanation if need be!

Cheers

core/model/modx/modresource.class.php
```diff
@@ -450,6 +450,43 @@ function __construct(xPDO & $xpdo) {
...
     }
 
+    /**
+     * Compute the "preview URL" for the resource
+     *
+     * @return string
+     */
+    public function getPreviewUrl() {
+        $url = '';
+        if (!$this->get('deleted')) {
```
JoshuaLuckers: I personally prefer to return values early to make the code more readable. 
```php
if ($this->get('deleted')) {
	return $url;
}
```

core/model/modx/modresource.class.php
```diff
@@ -450,6 +450,43 @@ function __construct(xPDO & $xpdo) {
...
+            $ctxSetting = $this->xpdo->getObject(
+                'modContextSetting',
+                array(
+                    'context_key' => $this->get('context_key'),
+                    'key' => 'session_enabled'
+                )
+            );
+
+            if ($ctxSetting) {
+                $sessionEnabled = $ctxSetting->get('value') == 0 ? array('preview' => 'true') : '';
```
JoshuaLuckers: `$sessionEnabled` is already set to `''` a few lines earlier 😄.
```suggestion
                $sessionEnabled = array('preview' => 'true');
```

JoshuaLuckers (commented):
I have some small suggestions but furthermore great improvement! We also need to get this in 3.x once this is approved and merged.

core/model/modx/modresource.class.php
rtripault: Same preference, but just copied over existing code. Will update ASAP

core/model/modx/modresource.class.php
rtripault: From my understanding, if the setting value is `1` (true) or anything else, we should not need to populate the query parameter.

JoshuaLuckers: Something went wrong when I finished my review. I also shared another suggestion but that seems to be missing.

core/model/modx/modresource.class.php
```diff
@@ -450,6 +450,43 @@ function __construct(xPDO & $xpdo) {
...
+            /** @var modContextSetting|null $ctxSetting */
+            $ctxSetting = $this->xpdo->getObject(
+                'modContextSetting',
+                array(
+                    'context_key' => $this->get('context_key'),
+                    'key' => 'session_enabled'
+                )
+            );
+
+            if ($ctxSetting) {
```
JoshuaLuckers: The missing suggestion:
```suggestion
            if ($ctxSetting && $ctxSetting->get('value') == 0) {
```

Mark-H:
@cla-bot check

opengeek:
@cla-bot check

opengeek:
I realize this is intended to fix what is in effect a bug, but by adding a new public method, this PR being merged in for a patch release would essentially circumvent semver rules. If any extra begins using this public method, and it only works on 2.7.3+, but not other 2.7.x releases, we can create issues for users. I realize this is a bit nitpicky in this instance, and I'm willing to overlook it in this case I suppose. What do others think?

opengeek:
The alternative would be to go ahead and bump the next 2.x release to 2.8.0.

JoshuaLuckers:
I don't think we should break the semver rules, even if it might be nitpicky.

JoshuaLuckers:
I propose to fix this in 2.7.x without breaking backwards compatibility and simply use "duplicate" code. This PR could then be applied to 3.x to fix it the proper way.
What do you think @opengeek ?

rtripault:
Sorry for the lack of reactivity folks. PR should have been updated to care take of comments & suggestions.

I indeed did not pay much attention about SemVer, focusing on fixing the issue.
Happy to have it shipped when/however you want. I'm not sure about about PHP requirements for the project, but if PHP >= 5.4, a trait might also be an option (food for thought).

Thanks

core/model/modx/modresource.class.php
```diff
@@ -450,6 +450,41 @@ function __construct(xPDO & $xpdo) {
...
     }
 
+    /**
+     * Compute the "preview URL" for the resource
+     *
+     * @return string
+     */
+    public function getPreviewUrl() {
+        if ($this->get('deleted')) {
+            return '';
```
rtripault, line 460: This is not related to this "issue" per se, but i'm just wondering why one can't preview a deleted resource (granted the permission is given)

JoshuaLuckers: A deleted resource will result in a 404 when requested.

PR #14957, opened 2020-01-16
Allow adding new values to Mulit-select listbox TV

BenceSzalai:
### What does it do?
When using a `Listbox (Single-Select)` type TV one can set the ''Force Selection to List" option effectively allowing the manager to choose from the list OR enter a custom value.
The same option however was not available for `Listbox (Multi-Select)` type TVs, but this PR enables it.

### Why is it needed?
The need to benefit from both the ability to choose a value from a predefined list AND to enter a custom value if needed is not related to the choice if the given input accepts a single value or more as well, so it is not reasonable to allow such option for the Single variant, but not for the Multi.

Since there was a bug that prevented new values to be added to a `Listbox (Multi-Select)` TV, even if the option was enabled, my assumption that it was not allowed in the first place, because it didn't work anyway. 5729093 fixes that, and c3b4f1d adds a very similar TV Input Option as the one already available for `Listbox (Single-Select)`: the ''Force Selection to List".

### Related issue(s)/PR(s)
I've found none.

### DIFF
--- a/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
+++ b/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
@@ -44,7 +44,8 @@ Ext.onReady(function() {
         {if $params.listEmptyText|default}
             ,listEmptyText: '{$params.listEmptyText|default}'
         {/if}
-        ,forceSelection: true
+        ,allowAddNewData: {if $params.forceSelection|default && $params.forceSelection|default != 'false'}false{else}true{/if}
+        ,addNewDataOnBlur: true
         ,stackItems: {if $params.stackItems|default && $params.stackItems|default != 'false'}true{else}false{/if}
         ,msgTarget: 'under'
 
--- a/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
+++ b/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
@@ -54,7 +54,10 @@ Ext.onReady(function() {
             'select': {fn:MODx.fireResourceFormChange, scope:this}
             ,'beforeadditem': {fn:MODx.fireResourceFormChange, scope:this}
             ,'newitem': {fn:function(bs,v,f) {
-                bs.addNewItem({"id": v,"text": v});
+                var item = {};
+                item[bs.valueField] = v;
+                item[bs.displayField] = v;
+                bs.addNewItem(item);
                 MODx.fireResourceFormChange();
                 return true;
             },scope:this}
--- a/core/lexicon/en/tv_widget.inc.php
+++ b/core/lexicon/en/tv_widget.inc.php
@@ -16,6 +16,7 @@ $_lang['combo_allowaddnewdata'] = 'Allow Add New Items';
 $_lang['combo_allowaddnewdata_desc'] = 'When Yes, allows items to be added that do not already exist in the list. Defaults to No.';
 $_lang['combo_forceselection'] = 'Force Selection to List';
 $_lang['combo_forceselection_desc'] = 'If using Type-Ahead, if this is set to Yes, only allow inputting of items in the list.';
+$_lang['combo_forceselection_multi_desc'] = 'If this is set to Yes, only items already in the list are allowed. If No, new values can be entered a well.';
 $_lang['combo_listempty_text'] = 'Empty List Text';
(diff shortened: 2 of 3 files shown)


JoshuaLuckers:
Thanks for taking time to submit this PR and help make MODX better. Since you are a first-time contributor; please sign the the [Contributor License Agreement (CLA)](https://modx.com/community/contribute/cla/).

BenceSzalai:
Hi! I've done that already! Has it not submitted correctly maybe?

BenceSzalai:
By the way, as we are speaking a note regarding 5729093: I could have used a one line fix instead of the more ellaborate one submitted, but I'm not sure about the compatibility of this newer syntax with older systems. I know ModX supports some really old deployments where this may not be welcome:
` bs.addNewItem({ [bs.valueField]: v, [bs.displayField]: v });`
But not sure. If you think it's okay, this one liner is more elegant than:
```
var item = {};
item[bs.valueField] = v;
item[bs.displayField] = v;
bs.addNewItem(item);
```

opengeek:
@Grapestain I need your modx.com username or email address in order to confirm your CLA status. Can you provide that?

BenceSzalai:
It's the same as here! ;)
I've also provided my github username on the CLA form too.

Mark-H:
@cla-bot check

manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
```diff
@@ -53,7 +54,10 @@ Ext.onReady(function() {
             'select': {fn:MODx.fireResourceFormChange, scope:this}
             ,'beforeadditem': {fn:MODx.fireResourceFormChange, scope:this}
             ,'newitem': {fn:function(bs,v,f) {
-                bs.addNewItem({"id": v,"text": v});
+                var item = {};
+                item[bs.valueField] = v;
+                item[bs.displayField] = v;
```
misha5000:
```javascript
var item = {
    [bs.valueField]: v,
    [bs.displayField]: v,
};
```

mishanthrop, line 59:
```javascript
var item = {
    [bs.valueField]: v,
    [bs.displayField]: v,
};
```

JoshuaLuckers (approved):

Ruslan-Aleev:
Unfortunately this does not work properly, see - https://github.com/modxcms/revolution/pull/15069#issuecomment-645930191

BenceSzalai:
Depends what do you mean properly...

In our use-case the key here is that content editors can add their own values to a TV selection list, and use those values in the future by selecting them from the list once they are added. It is not promised anywhere that the TV definition will be updated, and we don't need that either.

Actually this PR does not modify the working logic and behaviour of the "Force Selection to List" option, just makes sure the same behaviour is available for the `Listbox (Multi-Select)` type TV as for the `Listbox (Single-Select)` type TV.

If you think the behaviour should be changed, that is a separate issue, but regardless of what the behaviour is, there is no reason for the Single and the Multi to be different, which is fixed by this PR.

Ruslan-Aleev:
> If you think the behaviour should be changed, that is a separate issue

I agree with that.

> In our use-case the key here is that content editors can add their own values to a TV selection list, and use those values in the future by selecting them from the list once they are added.

And here are the problems. I meant that the values entered by the user are **not displayed in the TV**, in the resource form:

tv_list_m

TV does not work properly, correct me if I am wrong.

BenceSzalai:
Well, I didn't remember from like half a year ago, so I've checked and here's the deal:

In fact it's only half of the solution. The other half is to use SQL for the *Input Option Values* parameter.

Put this in there, and enjoy being able to select any custom values entered on any resource for all resources:

```@SELECT '' AS val UNION ALL SELECT DISTINCT `value` as val FROM `modx_site_tmplvar_contentvalues` WHERE ( `tmplvarid` = 155 ) ORDER BY val ASC;```

where you set the `tmplvarid` (155 in my case) to the ID of the template var as shown on the top of the window.

So that's why I was not bothered more with if ModX populates the list. Because the best list is the list in the db. Only caveat is there's no simple place to add default selection options, but even that can be done like this:

```@SELECT '' AS val UNION ALL SELECT 'Default option 1' AS val UNION ALL SELECT 'Default option 2' AS val UNION ALL  SELECT DISTINCT `value` as val FROM `modx_site_tmplvar_contentvalues` WHERE ( `tmplvarid` = 155 ) ORDER BY val ASC;```

I hope this helps!

Issue #14951, opened 2020-01-15
[3.x] Upgrade MODX 2.7.2 to MODX 3.0.0-alpha1


CrazyBoy49z:
### Summary
Error updating table for class  MODX\Revolution\modUser

```
Ошибка обновления таблицы для класса MODX\Revolution\modUser
Array
(
[0] => 00000
[1] =>
[2] =>
)
```

Ibochkarev:
Fixed in #14960
@CrazyBoy49z Thanks
@modxbot close

PR #14960, opened 2020-01-21
Fix upgrade error "Error updating table for class MODX\Revolution\modUser"

JoshuaLuckers:
### What does it do?
When upgrading, alter the `hash_class` column instead of the non-existing column `hashing_class`.

### Why is it needed?
To fix an error when upgrading and to make sure we change the value of the "right" column. 

### Related issue(s)/PR(s)
Resolves issue #14951


Mark-H (approved):

opengeek:
@cla-bot check

opengeek:
@cla-bot check

PR #14979, opened 2020-01-29
Lexicon check for the core

Jako:
### What does it do?
Add a lexicon check for the core. Cross check the core files with the existing lexicon entries for the usage of a lexicon entry.

3.x version of #14940 

### Why is it needed?
To detect easier, which lexicon entries are used.

The script creates 3 files in the _build/lexicon folder. The files are containing probably missing and superfluous entries and the third one collects usages of the lexicon with variable generated keys (not all are currently collected - only the ones starting with a single/double quote key). The lexicon with variable generated keys can't be copied to a lexicon, so this generated code does not have to be valid - the other ones could be copied.

### Related issue(s)/PR(s)
#14512


_build/lexicon/checklexicon.php
```diff
@@ -0,0 +1,565 @@
...
+                    try {
+                        include $current->getRealPath();
+                    } catch (Exception $e) {
+                        $this->invalidLexicons[] = $current->getFilename();
+                    }
+                }
+            }
+        } else {
+            return 'Could not load the lexicons in the setup language folder "' . $this->setupLexiconPath . $this->language . '/' . '"!';
+        }
```
JoshuaLuckers: It would be better to make a function for this to prevent code duplication.

_build/lexicon/checklexicon.php
```diff
@@ -0,0 +1,565 @@
...
+
+    /**
+     * Add used lexicon keys
+     */
+    private function addKeys()
+    {
+        $directory = new \RecursiveDirectoryIterator($this->scanPath, \RecursiveDirectoryIterator::SKIP_DOTS);
+        $filter = new \RecursiveCallbackFilterIterator($directory, function ($current, $key, $iterator) {
+            /** @var \RecursiveDirectoryIterator $current */
+            if ($current->getFilename()[0] === '.') {
```
JoshuaLuckers, line 259: It's unclear to me what this statement does and why.

_build/lexicon/checklexicon.php
```diff
@@ -0,0 +1,565 @@
...
+                return ($current->isFile() && isset($pathinfo['extension']) && (
+                        $pathinfo['extension'] == 'php' ||
+                        $pathinfo['extension'] == 'js' ||
+                        $pathinfo['extension'] == 'html' ||
+                        $pathinfo['extension'] == 'tpl' ||
+                        $pathinfo['basename'] == 'config.json'
+                    ) &&
+                    strpos($pathinfo['basename'], 'min.js') === false &&
+                    strpos($pathinfo['basename'], 'ext-') !== 0
+                ) ? true : false;
```
JoshuaLuckers: It would be better to make a function for this to make it more understandable and readable.

_build/lexicon/checklexicon.php
```diff
@@ -0,0 +1,565 @@
...
+     * Add lexicon calls in php files:
+     * modx->lexicon('whatever'
+     *
+     * @param string $filename
+     */
+    private function addPhpKeys($filename)
+    {
+        $fileContent = file_get_contents($filename);
+        $results = array();
+        preg_match_all('/(modx|xpdo)->lexicon\((["\'])(.*?)\2\s*[,\)]/m', $fileContent, $results);
```
JoshuaLuckers: We should consider using [Named Capturing Groups](https://www.regular-expressions.info/named.html) here.

_build/lexicon/checklexicon.php
```diff
@@ -0,0 +1,565 @@
...
+     * _('whatever'
+     * createDelegate(this, ['whatever'
+     *
+     * @param string $filename
+     */
+    private function addJsKeys($filename)
+    {
+        $fileContent = file_get_contents($filename);
+        $results = array();
+        preg_match_all('/_\(([\'"])(.*?)\1\s*[,\)]/m', $fileContent, $results);
```
JoshuaLuckers: We should consider using [Named Capturing Groups](https://www.regular-expressions.info/named.html) here.

_build/lexicon/checklexicon.php
```diff
@@ -0,0 +1,565 @@
...
+                    if (strpos($result, '+') === false
+                    ) {
+                        $this->languageKeys[] = $result;
+                    } else {
+                        $this->variableKeys[] = $result;
+                    }
+                }
+            }
+        }
+        preg_match_all('/(createDelegate)\(.*?, \[([\'"])(.*?)\2/m', $fileContent, $results);
```
JoshuaLuckers: We should consider using [Named Capturing Groups](https://www.regular-expressions.info/named.html) here.

JoshuaLuckers (commented):
Very useful addition. I left some suggestions here and there. 

Also; can convert the traditional array syntax to the short array syntax?

JoshuaLuckers (approved):

Issue #14686, opened 2019-08-20
MODX 3 Visual errors with styles


Ruslan-Aleev:
### Summary
- [ ] 1) In the TV, the "Set to default" button is located strangely.
![tv](https://user-images.githubusercontent.com/12523676/63338282-928e1e00-c353-11e9-9a9b-c8928172414b.png)

- [ ] 2) Text is not aligned in height
![valign](https://user-images.githubusercontent.com/12523676/73382208-8c2ba700-42e0-11ea-8673-263e2b8fe108.png)

- [ ] 3) The font in the package installer does not look very nice.
![fonts](https://user-images.githubusercontent.com/12523676/63339494-0e896580-c356-11e9-8dc8-cadb3ed78783.png)

- [ ] 4) In my opinion, red is not the best choice for uninstalled packages.
![package-color](https://user-images.githubusercontent.com/12523676/73382315-b3827400-42e0-11ea-9239-a485df6055b7.png)

- [ ] 5) No padding on top, it looks weird
![padding-top](https://user-images.githubusercontent.com/12523676/73392295-f1889380-42f2-11ea-9002-eea01c7ccaf9.png)

- [ ] 6) some inaccuracies in the tree
![tree_1](https://user-images.githubusercontent.com/12523676/73612110-08d0c500-4602-11ea-8d3c-1bf67ee790f0.png)

![tree_2](https://user-images.githubusercontent.com/12523676/73612111-08d0c500-4602-11ea-8b65-6c1fd8adca5a.png)

### Environment
MODX 3

meshkov:
I think it can be added here:
Go to Files and open Filesystem 

![image](https://user-images.githubusercontent.com/1703670/64236536-41f5f380-cf03-11e9-8377-aa12c58c04c4.png)

Ruslan-Aleev:
https://github.com/modxcms/revolution/issues/14597

Ruslan-Aleev:
https://github.com/modxcms/revolution/issues/14375

Ruslan-Aleev:
![tree_buttons](https://user-images.githubusercontent.com/12523676/66674194-5f8d4a00-ec73-11e9-952a-0fb8e441e911.png)

Ruslan-Aleev:
Text is not centered on the buttons in height:

![bh](https://user-images.githubusercontent.com/12523676/69011930-5fc2e880-0989-11ea-9523-2187ff69a3aa.png)

Ruslan-Aleev:
[...]

PR #14989, opened 2020-02-02
Changed navigation styles in element tree

Ruslan-Aleev:
### What does it do?
Changed the navigation styles in the element tree: buttons and selected item and simplified styles.

**Buttons:**
Before (ovals / not centered in height):
tree_1_before

After:
tree_1_after

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14686 (6)

### DIFF
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -148,23 +148,31 @@
   background: transparent 0 0;
   display: inline-block;
   width: 10px;
-  padding-left: 6px;
-  margin: 0 0 0 -16px;
+  padding-left: 4px;
+  padding-right: 4px;
+  text-align: center;
+  margin: 0;
+}
+
+.x-tree-arrows .x-tree-elbow-plus:before,
+.x-tree-arrows .x-tree-elbow-minus:before,
+.x-tree-arrows .x-tree-elbow-end-plus:before,
+.x-tree-arrows .x-tree-elbow-end-minus:before {
+  @extend %pseudo-font;
+  content: fa-content($fa-var-caret-right);
 }
 
 .x-tree-arrows .x-tree-elbow-minus:before,
 .x-tree-arrows .x-tree-elbow-end-minus:before {
   @extend %pseudo-font;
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Nice improvements for the buttons in the tree.

I personally prefer the blue color for the selected text. As far as I can tell blue is also used by Operating Systems to indicate a selection.

JoshuaLuckers:
I would love to get more input from others about the color for the selected item in the tree.
The other changes are fine with me and I'm happy to approve them.

ghost:
I prefer the blue. I'm very dependent on the colors of such things, since I often don't see other cues like tooltips or small captions or descriptions very well. Too many different colors is confusing.

Ruslan-Aleev:
By the way, green is combined with the "Save" button, but the blue color does not combine with anything in the panel (there is dark blue, but it’s not nearby).
Well, in general, I added green, because it’s kind of cold without such colors. But the question is not very important :) @rthrash what do you think?

JoshuaLuckers:
> By the way, green is combined with the "Save" button, but the blue color does not combine with anything in the panel (there is dark blue, but it’s not nearby).
> Well, in general, I added green, because it’s kind of cold without such colors. But the question is not very important :) @rthrash what do you think?

I still don't agree with green being the best color to indicate the selected element. Green colored elements trigger a users attention and should be used to indicate positive actions and/or positive consequences. 

If we use green for the selected item in the tree it can cause confusion. Did something with positive consequences happen to this item? I only clicked it.. did it save my changes while I didn't click it? We also don't use a green color elsewhere to indicate the selected state (selected inputs have a gray border).

Highlighting the selected item in the tree is "informative". Blue is a widely used color for informative elements.

However, I do agree with you that the current blue feels "cold" and isn't used elsewhere.

Ruslan-Aleev:
> Highlighting the selected item in the tree is "informative". Blue is a widely used color for informative elements.

Fair :) If there are no other color suggestions, I will return the color in the coming days.

JoshuaLuckers (approved):

Ibochkarev (approved):

PR #14994, opened 2020-02-07
Fix package dependency constraints not working

JoshuaLuckers:
### What does it do?
Use the fully qualified classname when checking the latest installed package version for package dependencies

### Why is it needed?
Checking dependency constraints would result in an error because of invalid SQL

### Related issue(s)/PR(s)
-


Mark-H:
Can you please squash your commits?

Mark-H (approved):

Issue #14974, opened 2020-01-28
Setup Welcome Screen Blank on Next


jaygilmore:
### Summary
When installing 3.0-alpha2, after selecting a language and clicking the _Next>_ button to proceed, the screen remains on the Welcome page, however, it is now blank.  
![installer-welcome-blank](https://user-images.githubusercontent.com/378199/73306224-26f98800-41f2-11ea-99bb-a3938de25dd7.png)

### Step to reproduce
Start to install Revo, select a language from the language screen and click _Next>_. 

### Observed behaviour
The **Welcome** page appears to be reloaded as blank after clicking _Next>_ the first time. I retested this and didn't see any errors in the console. 

### Expected behavior
After selecting the language and clicking _Next>_, the user should be presented with the option to change the context key, or click _Next>_ again to take the user to the **Options** screen. 
<img width="937" alt="installer-welcome-b" src="https://user-images.githubusercontent.com/378199/73306172-0e896d80-41f2-11ea-9fd7-0782c628cf5a.png">

### Environment
3.0-alpha2
PHP 7.2.1
MySQL 5.6.38

qodo:
+1 I am seeing this too on both Windows Firefox and Chrome.
There are no error messages in neither the server logs or browser console - it's just blank.

This doesn't prevent the user from continuing to the next step though!

Ibochkarev:
Resolved #14999 
@modxbot close

PR #14999, opened 2020-02-08
Setup improvements

JoshuaLuckers:
### What does it do?
Some technical improvements in the setup

### Why is it needed?
To make the setup process more user friendly

### Related issue(s)/PR(s)
Resolves issue #14742, #14974


setup/index.php
```diff
@@ -40,26 +40,23 @@
 
 /* check for compatible PHP version */
 define('MODX_SETUP_PHP_VERSION', phpversion());
-$php_ver_comp = version_compare(MODX_SETUP_PHP_VERSION, '5.6.0');
-if ($php_ver_comp < 0) {
-    die('<html><head><title></title></head><body><h1>FATAL ERROR: MODX Setup cannot continue.</h1><p>Wrong PHP version! You\'re using PHP version '.MODX_SETUP_PHP_VERSION.', and MODX requires version 5.4.0 or higher.</p></body></html>');
+if (version_compare(MODX_SETUP_PHP_VERSION, '7.0.0', '<')) {
```
alroniks: Does it make sense to parse somehow from the composer.json file to get defined this version from one place?

setup/index.php
```diff
@@ -40,26 +40,23 @@
...
+    die('<html><head><title></title></head><body><h1>FATAL ERROR: MODX Setup cannot continue.</h1><p>Wrong PHP version! You\'re using PHP version '.MODX_SETUP_PHP_VERSION.', and MODX requires version 7.0 or higher.</p></body></html>');
```
alroniks: I know that it is not in the scope of the task but could you please move this HTML template at least to variable or const with HEREDOC value, to avoid duplication and mess on the code?

setup/index.php
JoshuaLuckers: It makes sense to have a single source of truth. We could use `composer.json` if we are guaranteed this file is always available.

setup/index.php
JoshuaLuckers: Great suggestion, it's something I will change before this PR is ready for review.

setup/controllers/language.php
```diff
@@ -35,7 +35,9 @@
     $settings = $install->request->getConfig();
     $settings = array_merge($settings, $_POST);
     $install->settings->store($settings);
-    $this->proceed('welcome');
+
+    $nextAction = (MODX_SETUP_KEY === '@traditional@') ? 'options' : 'welcome';
```
mishanthrop, line 39: Excess brackets.

JoshuaLuckers: @mishantrop Thanks for taking time to review my code. I think the brackets make the code easier to read, don't you agree?

mishanthrop: Without [codings stardarts](https://docs.modx.com/current/en/contribute/code/coding-standards) for this, brackets is a matter of taste. Without brackets, the code looks more concise; it takes less time to read.
At this moment, that's at your discretion.

theboxer:
@JoshuaLuckers any plans on finishing this one?

JoshuaLuckers:
Yes, but I'm not sure when. I hope to have some time in the upcoming weeks.

theboxer (approved):

JoshuaLuckers:
I made some improvements in the way the requirements are validated. This is what it looks like when the requirements are not satisfied:

Screen Shot 2020-07-11 at 11 47 21

_For this example I forced all requirements to always "fail" 🙈_

setup/index.php
```diff
@@ -14,92 +14,15 @@
...
-    die('<html><head><title></title></head><body><h1>FATAL ERROR: MODX Setup cannot continue.</h1><p>Make sure you have uploaded all of the setup/ files; your setup/includes/modinstall.class.php file is missing.</p></body></html>');
-}
 $modInstall = new modInstall();
 if ($modInstall->getService('lexicon','modInstallLexicon')) {
     $modInstall->lexicon->load('default');
 }
 //$modInstall->findCore();
 $modInstall->doPreloadChecks();
-$requestClass = $isCommandLine ? 'request.modInstallCLIRequest' : 'request.modInstallRequest';
+$requestClass = MODX_SETUP_INTERFACE_IS_CLI ? 'request.modInstallCLIRequest' : 'request.modInstallRequest';
```
opengeek, line 25: Where is MODX_SETUP_INTERFACE_IS_CLI defined? I can't find this change in these diffs.

JoshuaLuckers: > Where is MODX_SETUP_INTERFACE_IS_CLI defined? I can't find this change in these diffs.

@opengeek `setup/provisioner/bootstrap.php` line 11:
https://github.com/modxcms/revolution/pull/14999/files#diff-6d29bb4ac4a6b37d567e2441cf473fe0R11

Issue #12591, opened 2015-08-24
[2.4] Date TV ignores min-max-time settings


mindeffects:
### Problem

The "hide time" feature for the TV-type "date" – introduced with #12348 – is nice but has some flaws: instead of storing a fixed time, the current time is being saved. Even if the values min- & max-time are set, they get ignored which results in a more or less "random" time being saved. Setting it to a fixed 00:00:00 would have been much better. Why?

A non-fixed time makes comparisons on the database layer overcomplicate: Find the same date? Difficult. One must strip the time time before the comparison, because a random time was stored and is influencing the result:
`2015-12-31 12:34:56 > 2015-12-31 01:23:45`
`2015-12-31 00:00:00 = 2015-12-31 00:00:00`

Another thing: You need a "time-only" field? Bad luck. Not possible.
### Solution Proposal

The following solutions does not need any new fields, only some logic behind analysing the values and setting the date-time-picker accordingly. And it can create three variants of the date TV.
##### Date only

If min and max time are equal, no time-picker is displayed, the time is set to the (identical) min-max-value and gets saved.
##### Time only

If min and max date are equal, no date-picker is displayed, the date is set to the (identical) min-max-value and gets saved
##### Date and Time

Everything is like it's used to be in the old days. But the time and date min-max-values are taken into account before the date-time-picker is shown and the value are corrected according to the boundries.
### Result

Everybody smiles and is happy. :-)

inreti:
I prefer having separate DateTime, Date and Time TVs with clean semantics. The above would  always be stored in the DB as timestamp where you cannot see which semantics is inside. If we had different TV types here, the TV type only decides over the semantics of the timestamp. If it is only one TV type we also need to interpret the TVs options to clearly understand what the timestamp means...

JoshuaLuckers:
Hiding the time option with the current `date` template variable what do developers and/or users expect regarding the time when selecting a date?
- the time is set to the current time
- the time is set to `00:00:00`
- don't store the time at all

mindeffects:
In no scenario, inside the knows universe, I can imagine storing the `current time` is of any use considering that I want to hide the time.
I would expect the time to be a constant value, like 00:00:00 to make comparisons easier, like `date 1 > date2`. I would also be fine with no time being stored at all (assuming it is not stored as a timestamp, of course).

JoshuaLuckers:
[...]

PR #15000, opened 2020-02-08
Date TV improvements

JoshuaLuckers:
### What does it do?
Added an option to set the hidden format for date TV's when the time is hidden

### Why is it needed?
This makes it possible to hide the time option for users and set the saved time to something custom instead of setting the time to the moment it got saved.

### Related issue(s)/PR(s)
Resolves issue #12591

### DIFF
--- a/manager/assets/modext/util/datetime.js
+++ b/manager/assets/modext/util/datetime.js
@@ -31,6 +31,11 @@ Ext.ux.form.DateTime = Ext.extend(Ext.form.Field, {
      */
     ,hiddenFormat:'Y-m-d H:i:s'
     /**
+     * @cfg {String} hiddenFormat Format of datetime used to store value in hidden field
+     * and submitted to server (defaults to 'Y-m-d H:i:s' that is mysql format)
+     */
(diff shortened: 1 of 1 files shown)


Mark-H:
PR 15000 🎉

Jako (approved):

manager/assets/modext/util/datetime.js
```diff
@@ -30,6 +30,11 @@ Ext.ux.form.DateTime = Ext.extend(Ext.form.Field, {
      * and submitted to server (defaults to 'Y-m-d H:i:s' that is mysql format)
      */
     ,hiddenFormat:'Y-m-d H:i:s'
+    /**
+     * @cfg {String} hiddenFormat Format of datetime used to store value in hidden field
+     * and submitted to server (defaults to 'Y-m-d H:i:s' that is mysql format)
```
Jako: The description has to be slightly changed, but otherwise ok.

JoshuaLuckers: Thanks @Jako, you have to re-approve this PR.

Ibochkarev (approved):

PR #15002, opened 2020-02-08
Move deprecated logging into dedicated table/view

Mark-H:
### What does it do?

Refactors $modx->deprecated to log to a dedicated table rather than the MODX log, making it visible in its own controller. 

### Why is it needed?

Before, deprecated messages (of which there are... quite a few) would go into the error log, causing that to grow rapidly (even on clean installations) and important information going to waste. This makes it more insightful without causing the downsides of the logs getting lost.

afbeelding

### Related issue(s)/PR(s)

N/a


JoshuaLuckers:
I propose to rename the menu item "Error Log" to "Error Reports" (or maybe "Debug Log"?). 
Then add tabs for each  "log type" (errors, deprecations, debug, etc).

JoshuaLuckers:
@Mark-H is this still on your radar?

theboxer:
@Mark-H any chance on finishing this one? :)

Mark-H:
Work is finally getting back to normal so I'll try and get this sorted out in the next month or so. It should be pretty close. I like @JoshuaLuckers's idea of making it a tab on the error log page so will look into that. 

Reviews of the rest of the code are certainly welcome so I can hopefully get it ready in one sitting. ;)

core/lexicon/en/system_events.inc.php
```diff
@@ -12,4 +12,15 @@
 $_lang['error_log_download'] = 'Download Error Log ([[+size]])';
 $_lang['error_log_too_large'] = 'The error log at <em>[[+name]]</em> is too large to be viewed. You can download it via the button below.';
 $_lang['system_events'] = 'System Events';
-$_lang['priority'] = 'Priority';
\ No newline at end of file
+$_lang['priority'] = 'Priority';
+
+$_lang['deprecated_log'] = 'Deprecations Log';
+$_lang['deprecated_log_desc'] = 'From time to time, old functionality is marked as deprecated to be removed in a future version. In such cases you\'ll find deprecation notices that alert you that updates will be necessary in the future.';
```
JoshuaLuckers, line 18:
```suggestion
$_lang['deprecated_log_desc'] = 'From time to time, old functionality is marked as deprecated and to be removed in a future version. In such cases you\'ll find deprecation notices that alert you that updates will be necessary in the future.';
```

Mark-H:
Work did not quite get back to normal after September, actually landed in a pretty deep burnout 😬 If someone else wants to take over this PR to get it done let me know.

core/lexicon/en/system_events.inc.php
Mark-H: I think "marked as deprecated to be removed in a future version" is the correct English here: "deprecated to be removed" being the **one** status it is given. If native speakers disagree please reopen the conversation.

Mark-H:
I've resolved some 3.x conflicts, made sure it creates tables on upgrade, and moved the grid to the error log as Joshua suggested. Ready for review.

Note that as you can also see in the updated screenshot; it may sometimes create duplicate records the first time a message gets logged. That has to do with deprecated messages being persisted in memory temporarily until the end of the request for performance. I don't think it's possible to avoid that without making the process slower.

JoshuaLuckers:
Once again: amazing job!

Would it be possible to fix the conflict in `phpcs.xml`?

Mark-H:
Rebased, should be good to go. :)

JoshuaLuckers (approved):
LGTM!

Ruslan-Aleev:
Thanks for the PR! In my opinion, the `deprecated_log_cleared` pop-up message is not really needed:
- It is clear that the deprecated grid has been cleared;
- When clearing the error log (and other similar grids), no pop-up messages appear;
- Extra code and extra lexicon.

Ruslan-Aleev:
Also, the Package Manager does not work for me, I installed everything from scratch, it gives a 500 error.

pm_500

Mark-H:
@Ruslan-Aleev Have you done a composer install/update? I'd suspect you didn't since #15779 got merged, which added new dependencies for the package manager. 

Otherwise please post the error message. Package manager seems to work fine here.

Ruslan-Aleev:
@Mark-H Yes, I ran composer update - it worked, thanks.
I also found an inaccuracy in the UX: the buttons on the right do not work (they refer to the Error log), which can be confusing, it might be better to hide them from the Deprecations log panel.

dl_buttons

PR #15005, opened 2020-02-11
Fix add script tag to meet W3C requirements

supaweb:
W3C validator output warning "The type attribute is unnecessary for JavaScript resources."

### What does it do?
Default code generates script tag with type attribute that is unnecessary. For example <script type="text/javascript" src="{src}"></script>
This pull request fix this by delete type attribute for script tag

### Why is it needed?
W3C HTML Validator doesn't throw warning. SEO page params grows


supaweb:
@cla-bot check

Jako:
Thanks for helping to make MODX better. But I don't think, that this is urgently needed in the 2.x branch. So please recreate it in the 3.x branch.

PR #15008, opened 2020-02-11
Fix contexts.tpl

Jako:
### What does it do?
- Change $() calls to plain javascript
- Change Ext.get() calls to plain javascript

### Why is it needed?
We don't have jQuery installed and ExtJS is overkill for this.

### DIFF
--- a/setup/templates/contexts.tpl
+++ b/setup/templates/contexts.tpl
@@ -21,10 +21,10 @@ Ext.onReady(function() {literal}{{/literal}
         <label for="context_web_path">{$_lang.context_web_path}:</label>
     </div>
     <div class="col-2">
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Does this mean it never worked?

Jako:
Those checkboxes never worked (the ones handled by $). It was a mix of ExtJS and $ handled ones.

JoshuaLuckers (approved):

JoshuaLuckers:
For people testing: these options are only visible when `MODX_SETUP_KEY` is *not* `@traditional@`.

Jako:
Or if modx is moved/installed into a sub directory of the webroot.

JoshuaLuckers (approved):

Ibochkarev (approved):
Thanks

PR #15009, opened 2020-02-11
Corrected TV form with the Date type

Ruslan-Aleev:
### What does it do?
Corrected the TV form with the date type:
- Changed the order
- Related items grouped

**Before:**
tv-date-before

**After:**
tv-form-after

### Why is it needed?
Improves UI / UX
It became more beautiful :)

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/templates/default/element/tv/renders/inputproperties/date.tpl
+++ b/manager/templates/default/element/tv/renders/inputproperties/date.tpl
@@ -1,12 +1,14 @@
 <div id="tv-input-properties-form{$tv|default}"></div>
-{literal}
 
+{literal}
 <script type="text/javascript">
 // <![CDATA[
 var params = {
-{/literal}{foreach from=$params key=k item=v name='p'}
- '{$k}': '{$v|escape:"javascript"}'{if NOT $smarty.foreach.p.last},{/if}
-{/foreach}{literal}
+{/literal}
+{foreach from=$params key=k item=v name='p'}
+    '{$k}': '{$v|escape:"javascript"}'{if NOT $smarty.foreach.p.last},{/if}
+{/foreach}
+{literal}
 };
 var oc = {'change':{fn:function(){Ext.getCmp('modx-panel-tv').markDirty();},scope:this}};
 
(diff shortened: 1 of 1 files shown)


manager/templates/default/element/tv/renders/inputproperties/date.tpl
```diff
@@ -58,111 +60,145 @@ MODx.load({
...
-        ,id: 'inopt_maxTimeValue{/literal}{$tv|default}{literal}'
-        ,value: params['maxTimeValue'] || ''
-        ,width: 200
+        xtype: 'textfield'
+        ,fieldLabel: _('disabled_dates')
+        ,description: MODx.expandHelp ? '' : _('disabled_dates_desc')
+        ,name: 'inopt_disabledDates'
+        ,id: 'inopt_disabledDates{/literal}{$tv|default}{literal}'
+        ,value: params['disabledDates'] || ''
+        ,anchor: '99%'
```
Jako, line 141: Does `anchor: '100%'` not work?

Jako: Same for the other 99% below.

Ruslan-Aleev: Unfortunately, no, it doesn’t look very good (with `anchor:'100%'`), the fields go beyond.
Although, for example, in https://github.com/modxcms/revolution/pull/14731 there is no such problem. I do not understand why.

Ruslan-Aleev: @Mark-H @JoshuaLuckers maybe you have some thoughts about why this is happening (I'm about `anchor:'100%'`) and how to fix it? :)

JoshuaLuckers: When the anchor is set to 100% the div with class `x-panel-bwrap` (child from div with ID `modx-panel-tv-input-properties`) will overflow horizontally.

JoshuaLuckers: Adding the `container` class on line 24 improves it a bit but still doesn't look perfect. Maybe this helps you find a better solution?

Jako: Maybe there is a border on a 100% width somewhere causing this issue.

Ruslan-Aleev: The simplest solution, it seems to me, is to put a lower %.
Neither extra classes `,cls: '...'`, nor styles `,bodyStyle: '...'`, nor a border `,border: false` - something all the same go beyond.

JoshuaLuckers (approved):

Issue #14991, opened 2020-02-03
Use single quotation marks in queries to support ANSI_QUOTES SQL mode in MySQL


muzzwood:
### Summary
I was unable to see the list of previous versions for specific packages in the manager. The processor was failing and after some debugging found it was due to double quotes being used in the sortby statement.
https://github.com/modxcms/revolution/blob/2.x/core/model/modx/transport/mysql/modtransportpackage.class.php#L65

This is incompatible with ANSI_QUOTES mode for string literals in MySQL 8.
https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

I tested the query with single quotes and that fixed it.

If no one does it first, I can attempt a PR in the next few days.

### Step to reproduce
Have a MySQL database with ANSI_QUOTES mode. Try to view previous packages in the package manager.

### Observed behavior
No data loaded in grid.

### Expected behavior
Data should load.

### Environment
MODX 2.7.2, nginx and PHP7.4, mysql 8 with ANSI_QUOTES enabled.

Ibochkarev:
Fixed #15011 

Thanks for taking time to report this issue and help make MODX better.
Please reopen if you still encounter this issue.

@modxbot close

PR #15011, opened 2020-02-12
Switched string literals to use single quotes so that it is compatibl…

muzzwood:
…e with ANSI_QUOTES sql mode.

### What does it do?
Switched to using single quotes as double quotes are incompatible with ANSI_QUOTES SQL mode. 

### Why is it needed?
With double quotes, the grid to view previous package versions in the package manager can't load any data as the request fails.

### Related issue(s)/PR(s)
See issue: #14991


Jako:
Looks ok, but would that solution be suitable for sqlsrv too?

opengeek:
@Jako it already uses single quotes in the sqlsrv code. I'd really like to see all the strings used in this class use double-quotes for the strings themselves so the single-quotes would not need to be escaped. But I'm going to merge this as is to solve the issue for now.

opengeek (approved):

Issue #14424, opened 2019-02-22
[bug] Input Option Values for TV: bugs with a field and description


Ruslan-Aleev:
### Summary
Recently, an unnecessary Input Option Values for TV was hidden (https://github.com/modxcms/revolution/pull/14040), however there are a couple of bugs:
1) The input field still displayed for TV with types **"Auto-tag", "Hidden", "RichText" and "URL"**, but it is not needed in these TVs.

2) The input field description-label is **always shown**, even if the field itself is hidden, which is strange, see the picture below.

![tv_input_options_bug](https://user-images.githubusercontent.com/12523676/53235714-f0df8a00-36ab-11e9-8397-035b91c54d6e.png)

### Expected behavior
The field should be displayed and **be required** only for fields for which Option Values are needed, for example, the checkbox TV.
And the Option Values field should be checked at the stage of saving, as suggested by @JoshuaLuckers - https://github.com/modxcms/revolution/pull/14275#issuecomment-454428234 (which is a more logical option than I suggested in PR)

For the rest of the TV, this field and its description should be removed.

### Environment
MODX 2.7.1

Jako:
The description is outside of the `element` there. But it has a `for` attribute with the value of the element id. Maybe you could get the description with:

```
Ext.select('label[for="' + element.id + '"]');
```

and hide that, if it is found.

Issue #14311, opened 2019-01-26
Enable autoformat code (before commit)


Jako:
### Summary
The repository code should run through some code formatter like https://prettier.io/

### Why is it needed?
- No need to discuss style in code review

### Suggested solution(s)
Add `prettier` at least to the template build process to make the files in that part formatted. The other part (format the php code) needs some integration in an IDE or in the commit process.

We should also provide the code styles in an [Intellij IDEA code style XML](https://www.jetbrains.com/help/idea/copying-code-style-settings.html) scheme.

The code style of a PR should also be checked by Travis on GitHub and invalid styled code should be rejected.

### Related issue(s)/PR(s)
Nothing special, but changing the default branch to 3.x would be a great time to introduce this change, since after reformatting our code we will have merge issues with code from other branches.

Ibochkarev:
For example https://github.com/csscomb/grunt-csscomb

OptimusCrime:
Should have been done a long time ago. The only problem is all the merge conflicts it is going to cause.

JoshuaLuckers:
How can we move forward with this? I would really like to see a more standardised code base.

Ibochkarev:
Use https://prettier.io/ for our IDE with a common config.

Ibochkarev:
https://www.npmjs.com/package/grunt-prettier

There is a package milking Grunt. You can use it and write a config to format your code.

Ibochkarev:
I studied the information on this topic. I have found:
- prettifer has Pre-commit Hook https://prettier.io/docs/en/precommit.html - this is exactly what we need
- the configuration file is extensive http://json.schemastore.org/prettierrc

We need to take the first step and begin to embody this innovation.

I suggest:
- discuss the set of rules for the .prettierrc configuration file.
- add the given file to the root directory
- and test

Colleagues, your move!

JoshuaLuckers:
Does Prettier provide default rules? Maybe we can use that for a start.

Ruslan-Aleev:
There are libraries of this kind, for CSS, for example, https://github.com/csscomb

---

@Mark-H  specified:

Perhaps we should discuss code styles on the forum. When there's consensus (maybe there already is), existing code should be converted and automatically enforced as part of the checks that run for each PR and commit.

That's not just about SCSS, but also JS and PHP should get that.

Ibochkarev:
> Does Prettier provide default rules? Maybe we can use that for a start.

https://prettier.io/docs/en/configuration.html

PR #15016, opened 2020-02-13
Show "Input Option Values" with labels only on TV where it needed

Ruslan-Aleev:
### What does it do?
In https://github.com/modxcms/revolution/pull/14040, the “Input Option Values” field was hidden on TV forms where they were not needed.

However, https://github.com/modxcms/revolution/pull/14040 has several disadvantages:
1) Some TVs (for example, "Hidden", "RichText") do not have a property rendering template, which is why the "Input Option Values" field is still displayed.
2) TV where the "Input Option Values" field is not needed more than where it is needed.

More competently, in my opinion, hide the "Input Option Values" field by default everywhere, and show only where it is needed ("Radio", "Listbox", etc).

Also, this PR hides the description-label for a hidden "Input Option Values", which has always been shown.

### Why is it needed?
UI/UX improvements

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14424
https://github.com/modxcms/revolution/pull/14040

### DIFF
--- a/manager/assets/modext/widgets/element/modx.panel.tv.js
+++ b/manager/assets/modext/widgets/element/modx.panel.tv.js
@@ -562,9 +562,14 @@ Ext.extend(MODx.panel.TVInputProperties,MODx.Panel,{
     }
 
     ,showInputProperties: function(cb,rc,i) {
+        /**
+         * Hide "Input Option Values" by default
+         */
         var element = Ext.getCmp('modx-tv-elements');
+        var element_label = Ext.select('label[for="' + element.id + '"]');
         if (element) {
-            element.show();
+            element.hide();
+            element_label.hide();
         }
 
         this.markPanelDirty();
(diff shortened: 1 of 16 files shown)


JoshuaLuckers:
I appreciate the code reformatting but they make reviewing the actual changes a bit difficult. There was an issue/PR about the code style for JS/CSS files we should get back to to solve this.

Jako:
For code auto formatting see https://github.com/modxcms/revolution/issues/14311

JoshuaLuckers (approved):
Nice work, thanks! It would be great if we could reduce the code duplication (at a later time).

Ruslan-Aleev:
@JoshuaLuckers There is another option - **move "Input Option Values" to specific TV renders** (checkboxes, radio, etc.) **and make it required**.

As a result, we get:
- You can’t save the TV without setting the “Input Option Values” (in the current version, this can be done - and displays an empty TV in the resource)
- The field "Input Option Values" will be only where necessary, there will be no unnecessary js hiding / displaying of the field.

I couldn’t do it, but it’s not so difficult :) If you have time, take a look.

Ibochkarev (approved):
Worked. Thanks!

image

Issue #15007, opened 2020-02-11
Add transliteration of files when uploading to a media source


Ruslan-Aleev:
### Summary
It would be useful to add transliteration of files when uploading to a media source.

Yes, there are packages that do a great job of this. However, such functionality in the core would not be amiss.

**Again there is transliteration in url, but there is no transliteration of files - it is not logical and look unfinished.**

### Why is it needed?
English-speaking MODX users do not even know how many problems a file can give on server, for example, with a russian name.

Perhaps this was discussed earlier, I did not find.

Mark-H:
Per #15019, this would ideally be implemented by tapping into existing transliteration settings (and not a hardcoded table). Perhaps worth reconsidering the default translit setting values in 3.x to make sure some default transliteration is applied even if people don't know what transliteration is.

ilyautkin:
I'm going to work on this

Ibochkarev:
Improved in #15457

PR #15019, opened 2020-02-18
Add transliteration of files when uploading to a media source

crystaldaking:
Add translit for filenames

### What does it do?
Translates the file name into transliteration of Latin characters when they are found in the file name.

### Why is it needed?
English-speaking MODX users do not even know how many problems a file can give on server, for example, with a russian name.

### Related issue(s)/PR(s)
Related issue #15007

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -812,6 +812,34 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
     }
 
     /**
+     * Translitiration filenames
+     * @param string $fileName
+     * @return string
+     */
+    public function translitFileName ($fileName)
+    {
+        $charList = array(
+            "А"=>"a","Б"=>"b","В"=>"v","Г"=>"g",
+            "Д"=>"d","Е"=>"e","Ё"=>"e","Ж"=>"j","З"=>"z","И"=>"i",
+            "Й"=>"y","К"=>"k","Л"=>"l","М"=>"m","Н"=>"n",
+            "О"=>"o","П"=>"p","Р"=>"r","С"=>"s","Т"=>"t",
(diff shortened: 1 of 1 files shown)


Jako:
Thanks for helping to make MODX better. I don't think that a fixed cyrillic translit table is a suitable solution for all languages. There are a few extras (FileTranslit, FileSluggy etc.), that are doing a lot better transliteration than this solution. Maybe we could adopt one of those as default into the core.

Mark-H:
What @Jako said. Great improvement, but not a fan of the hardcoded table. 

Have you looked at modResource::filterPathSegment, how that handles the transliteration for the resource alias? That supports different solutions based on the `translit` system setting, including iconv and the [translit service/package](https://modx.com/extras/package/translit) that contains several transliteration tables. 

Might also be better if it targeted 3.x instead of 2.x. As media sources have been rewritten in 3.x, that would be a better place to incorporate a change like this.

crystaldaking:
@Mark-H 
Yeah, I just looked at filterPathSegment. Cool.
I agree with Jaco that it’s not worth shoving everything in the core and it’s better to separate everything that can be divided without any problems.

Ruslan-Aleev:
> I agree with Jaco that it’s not worth shoving everything in the core and it’s better to separate everything that can be divided without any problems.

Jako meant that we can adapt the component to the core, something similar (with translit component) is already used for url.
Anyway, problems with files happen, and such functionality in the core will not hurt, it seems to me :)

JoshuaLuckers:
@crystaldaking do you see a chance to continue with this PR?

theboxer:
👋  @crystaldaking is there any chance you'll rework this PR based on the discussion here?

Mark-H:
Closing for now per the discussion. This type of validation would be very useful to avoid issues, but a hardcoded table is not suitable and it would be more appropriate to tap into filterPathSegment which uses user-defined transliterations. Perhaps targetting 3.x and considering if the default translit settings are still correct would be a worthwhile discussion to have in #15007.

PR #15031, opened 2020-03-08
Prevent rewrite the .well-known directory used by LetsEncrypt

Jako:
### What does it do?
Prevent rewrite the .well-known directory

### Why is it needed?
The directory is used by LetsEncrypt and should not been rewritten.

### DIFF
--- a/ht.access
+++ b/ht.access
@@ -19,7 +19,7 @@ RewriteRule "/\.|^\.(?!well-known/)" - [F]
 
 
 # Prevent rewrite the .well-known directory used by LetsEncrypt by rules below of this rule
-RewriteRule "^\.well-known" - [L]
+RewriteRule "^\.well-known/" - [L]
 
 
 # Rewrite www.example.com -> example.com -- used with SEO Strict URLs plugin
--- a/ht.access
+++ b/ht.access
@@ -18,6 +18,10 @@ RewriteBase /
 RewriteRule "/\.|^\.(?!well-known/)" - [F]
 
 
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
What’s the difference with the earlier rule?

Jako:
The earlier rule sends a 403 forbidden for all files/folders named with a starting dot (except of .well-known): https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_f

In my case, I rewrite a www host to a non www host in the .htaccess with the default entries. To make a valid ACME check, the .well-known folder does not have to be rewritten and has to work with and without www in the hostname.

ht.access
```diff
@@ -18,6 +18,10 @@ RewriteBase /
 RewriteRule "/\.|^\.(?!well-known/)" - [F]
 
 
+# Prevent rewrite the .well-known directory used by LetsEncrypt by rules below of this rule
+RewriteRule "^\.well-known" - [L]
```
JoshuaLuckers:
```suggestion
RewriteRule "^\.well-known/" - [L]
```

Make sure that we match the directory `.well-known` and not the file and/or other directories starting with  `.well-known`. Although the previous rule already blocks this, it is still good to apply it here just in case.

JoshuaLuckers (approved):

opengeek:
@Jako can you please push this to your fork instead of the main repository and open a new PR?

Jako:
Sure, that happens, when creating the fork direktly on GitHub. Sorry.

PR #15042, opened 2020-04-01
Corrected TV form with the Number type

Ruslan-Aleev:
### What does it do?
Corrected the TV form with the Number type:
- Changed the order
- Related items grouped

**Before:**
tv-num-1

**After:**
tv-num-2

### Why is it needed?
Improves UI / UX
It became more beautiful :)

### Related issue(s)/PR(s)

### DIFF
--- a/manager/templates/default/element/tv/renders/inputproperties/number.tpl
+++ b/manager/templates/default/element/tv/renders/inputproperties/number.tpl
@@ -29,7 +29,7 @@ MODx.load({
         ,name: 'inopt_allowBlank'
         ,hiddenName: 'inopt_allowBlank'
         ,id: 'inopt_allowBlank{/literal}{$tv|default}{literal}'
-        ,width: 200
+        ,anchor: '100%'
         ,value: (params['allowBlank']) ? !(params['allowBlank'] === 0 || params['allowBlank'] === 'false') : true
         ,listeners: oc
     },{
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
I think the option "allow negative numbers" fits better with the min and max value group although I am not sure if it makes sense to offer this option as we already have a minimum value option. 

How do you (and others) think about this?

Ruslan-Aleev:
It seems to me that these are different groups. That is why I placed it separately.
The minimum value can be, for example, "10", and the maximum "1000" - and here, "Allow negative numbers" looks superfluous.

Ruslan-Aleev:
"Allow negative numbers" is a more global option, it also can applies to decimals.

JoshuaLuckers (approved):

Jako:
Allow negatives can't be replaced with a minimum value. It can't be inifinitive negative then.

JoshuaLuckers:
> Allow negatives can't be replaced with a minimum value. It can't be inifinitive negative then.

Allow negative values: leave min value blank?
Don't allow negative values: set min value to 0?

Jako:
I am not sure, wether the option works that way. Did you check this?

Ruslan-Aleev:
In my opinion, the current option with the choice is more obvious and simpler than setting the minimum value.

Ibochkarev (approved):
It's worked
image

Ruslan-Aleev:
It seems to me that we can add additional condition in https://github.com/modxcms/revolution/blob/3.x/manager/templates/default/element/tv/renders/input/number.tpl#L25 that does not take into "Min Value" with >= 0, if "Allow Negatives" is true.

```
{if $params.minValue|default != '' && is_numeric($params.minValue|default)}
    {if $params.allowNegative|default && $params.allowNegative|default != 'false' && $params.allowNegative|default != 'No'}
        {if $params.minValue|default < 0}
            ,minValue: {$params.minValue|default}
        {/if}
    {else}
        ,minValue: {$params.minValue|default}
    {/if}
{/if}
```

And add a description for "Min Value":

`If you set "Yes" to "Allow Negatives", but specified  >=0 for "Min Value", then "Min Value" will be ignored.`

**But the "Allow Negatives" parameter is superfluous, maybe I should remove it?**

@Jako @JoshuaLuckers What do you think?

JoshuaLuckers:
This is the current behavior:
1. If `Allow Negatives` is set to `No` and `Min Value` is `empty`:
- The input field doesn't accept the `-` character (_great_);
- If the value retrieved from the database is negative it *will be converted* to positive (so -10 becomes 10) without a warning/error (_to me, as a user, unexpected behavior_);
- The minimum value you can insert is `0`.

2. If `Allow Negatives` is set to `No` and `Min Value` is `> 0`:
- Same behavior as above except;
- Gives an error if the value is less then `Min Value`

3. If `Allow Negatives` is set to `Yes` and `Min Value` to `>= 0`:
- The input field accepts the `-` character (negative values can be inserted);
- If the value retrieved from the database is less then `Min Value` it's *not converted*;
- Gives an error if the value is less then `Min Value`. 

I agree with @Ruslan-Aleev, to me as a user,`Allow Negatives` is superfluous.

Ruslan-Aleev:
I made a mistake with the condition above, now I fixed it. Now it should work.

PR #15043, opened 2020-04-01
Corrected TV form with the ResourceList type

Ruslan-Aleev:
### What does it do?
Corrected the TV form with the ResourceList type:
- Changed the order
- Related items grouped
- Removed unused params in TV render in resource
- **Added list check via "Allow blank" instead of "Allow empty choice"**

**Before:**
tv-res-1

**After:**
tv-res-2

### Why is it needed?
Improves UI / UX
It became more beautiful :)

### Related issue(s)/PR(s)

### DIFF
--- a/core/lexicon/en/tv_widget.inc.php
+++ b/core/lexicon/en/tv_widget.inc.php
@@ -113,8 +113,6 @@ $_lang['resourcelist_where'] = 'Where Conditions';
 $_lang['resourcelist_where_desc'] = 'A JSON object of where conditions to filter by in the query that grabs the list of Resources. (Does not support TV searching.)<br/>Examples: [{"template:=":"4"}], [{"pagetitle:!=":"Home"}], [{"parent:IN":[34,56]}]';
 $_lang['richtext'] = 'RichText';
 $_lang['sentence_case'] = 'Sentence Case';
-$_lang['shownone'] = 'Allow Empty Choice';
-$_lang['shownone_desc'] = 'Allow the user to select an empty choice which is a blank value.';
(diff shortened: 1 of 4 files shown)


JoshuaLuckers:
Would it make sense to remove the option "allow empty choice" and add an empty choice to the select list when "allow blank" is set to "no"?

Ruslan-Aleev:
Unfortunately, I do not understand how the "Allow empty choice" parameter is used in the rendering of the resource. It displays "-" at the beginning of the list, but I did not find where exactly this happens.

p.s. By the way, there are other parameters that are not displayed in the TV settings, but are checked in the render of the resource, for example - https://github.com/modxcms/revolution/blob/3.x/manager/templates/default/element/tv/renders/input/resourcelist.tpl#L21

Ruslan-Aleev:
@JoshuaLuckers I changed the PR, check.
I added the display "-" at the top of the list always, because otherwise, the first resource was put at the top of the TV list; and you could save this resource in TV without noticing, because allowBlank = true (I'm about https://github.com/modxcms/revolution/pull/15043/commits/a62db2fd5f2ebf2af80b4a5caf7363830cf684fe).

JoshuaLuckers (approved):
Great work!

JoshuaLuckers:
If this is approved and merged we need to update [the relevant documentation](https://docs.modx.com/3.x/en/building-sites/elements/template-variables/input-types).

Jako:
Typeahead would be nice to have there. Similar as the removed typeahead in #15044.

PR #15044, opened 2020-04-02
Corrected the TV form with the Listbox (Multi-Select) type

Ruslan-Aleev:
### What does it do?
Corrected the TV form with the Listbox (Multi-Select) type:
- Changed the order
- Related items grouped
- **Removed not working params in TV render in resource**

**Before:**
list-m-1

**After:**
listbox-multi-final

### Why is it needed?
Improves UI / UX

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15043

### DIFF
--- a/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
+++ b/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
@@ -4,12 +4,10 @@
     onchange="MODx.fireResourceFormChange();"
     size="8"
 >
-{foreach from=$opts item=item}
-    <option value="{$item.value}" {if $item.selected} selected="selected"{/if}>{$item.text}</option>
-{/foreach}
+    {foreach from=$opts item=item}
+        <option value="{$item.value}" {if $item.selected} selected="selected"{/if}>{$item.text}</option>
+    {/foreach}
 </select>
-
-
 
 <script type="text/javascript">
 // <![CDATA[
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
Do you think we need to look into fixing the options that aren't working or is removing them the best solution?

Ruslan-Aleev:
In my opinion, these options are not needed at all. Delay, list width - who uses them?
And be honest, no one will fix them in future =)

JoshuaLuckers (approved):

Jako:
The life search would be quite nice for a long list, but I don't know, why this does not work here. I could look into that, so please don't merge this at the moment.

Ruslan-Aleev:
In the future, we need to look at this parameters for Listbox (Single-Select) TV

JoshuaLuckers:
@Jako is this still on your radar?

Jako:
It is, but not enough time at the moment.

manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
```diff
@@ -30,22 +28,12 @@ Ext.onReady(function() {
         ,valueField: "value"
         ,resizable: true
         ,allowBlank: {if $params.allowBlank == 1 || $params.allowBlank == 'true'}true{else}false{/if}
-
         {if $params.title|default},title: '{$params.title|default}'{/if}
-        {if $params.listWidth|default},listWidth: {$params.listWidth|default}{/if}
-        ,maxHeight: {if $params.maxHeight|default}{$params.maxHeight|default}{else}300{/if}
-        {if $params.typeAhead|default}
```
Jako: This check was wrong, because the value is a string with 'true' or 'false' in almost every case. So typeAhead is always set to true.

It should be checked like in allowBlank:

```
{if $params.typeAhead == 1 || $params.typeAhead == 'true'}
```

Jako (changes requested):
Please don't remove the typeAhead settings.

Jako:
I have checked the functionality of type ahead in an 2.x installation now. It works and should not be removed. It is very useful in long lists.

screencast 2020-05-02 11-50-46

But switching typeahead off does not work because of my comment above.

manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
Jako: It should be the same issue in listbox-single

Jako:
@Ruslan-Aleev Could you change the PR and don't remove the typeAhead options.

Ruslan-Aleev:
@Jako Yes, I will correct typeAhead for this type of TV and others, when I have time.
Thanks for fixing the bug!

theboxer:
@Ruslan-Aleev what's the state of this PR? Are you going to do the requested updates?

Ruslan-Aleev:
@theboxer It is necessary to solve the issue with the parameters, I'm about - https://github.com/modxcms/revolution/issues/15144.
I can just fix the UX design, but I would like to make PR more productive :)

Ruslan-Aleev:
@theboxer When you have time - test PR.
I, in fact, only deleted `listWidth` (it did not work initially) and `maxHeight` (this parameter was not in the TV settings, but it was in the TV render).
As for the controversial parameters - in the future, we will discuss and make a separate PR.

JoshuaLuckers (approved):

theboxer (approved):

Jako (approved):

Ruslan-Aleev:
@JoshuaLuckers Change label please.

Issue #10337, opened 2013-11-17
Make merge the default in setProperties()


BobRay:
bobray created Redmine issue ID 10337

public function setProperties($properties, $merge = false) {

I'm suggesting $merge = true as the default.

It's a rare case when you want to trash the properties you're not setting.

BobRay:
I just took a look at the current core code, and the use of `public function setProperties()` is very inconsistent and confusing across the multiple instances in the core.  

- modResource has $merge = true as the default
 
- modElement, modRestController, modMediaSource, modParser and modPropertySet have $merge = false as the default

- modLoginController uses setProperty() to do its own merge - no choice.
 
- modManagerController gives you no choice -- merge is false. The whole properties array is replaced with what is passed in.
 
- modProcessor is the opposite. No choice, merge is always true whether you want that or not.

The ones with no choice, especially the last two, can be frustrating when you're writing unit tests.

JoshuaLuckers:
Good points! What if we introduce new methods like `updateProperties`/`updateScriptProperties` for "merging" and `setProperties`/`setScriptProperties` for "replacing"?

Mark-H:
You've clearly looked into this more than I have, but I do want to emphasise that `setProperties` on a database object that has a properties column (like `modResource`) does something _very_ different from setProperties on a controller or processor which is used only in a single "request" or "execution". 

Having different defaults or behaviors in such cases is to be expected and in fact desirable. 

If something merges when it shouldn't for your use (e.g. modProcessor in a unit test), simply reinstantiate the class with the new properties.

JoshuaLuckers:
They do something very different but they still share the same name. Without looking at the implementation it's very confusing for people using these methods. My suggestion was to make the function names more descriptive to make it more descriptive of the actual thing(s) it does.

BobRay:
Before I learned to do a global search of the entire codebase before using a MODX function, I was bitten by this a few times when I assumed that there was only one getProperties() function.

@Mark-H Fair point, but it seems crazy to me that the defaults for modResource and modElement are the reverse of each other.

It can be pretty catastrophic, too, since you might want to reset a small array of properties. If you use setProperties() where the default is replacement, you'll be wiping out all the other properties. I suspect that this mistake is made often with elements (and user extended fields, which IIRC, don't even have setProperties() available). 

[...]

PR #15059, opened 2020-04-21
Merge option for setProperties() methods that don't have a merge option

JoshuaLuckers:
### What does it do?
Add the `merge` parameter to `setProperties()` methods that don't have a merge option.
The current behavior (merge = true /replace = false) is used as default value.

### Why is it needed?
Improve consistency and make it more clear if calling this function will merge or replace the properties set previously. 

### Related issue(s)/PR(s)
Issue #10337

### DIFF
--- a/core/src/Revolution/Processors/Processor.php
+++ b/core/src/Revolution/Processors/Processor.php
@@ -57,11 +57,13 @@ abstract class Processor {
     /**
      * Set the runtime properties for the processor
      * @param array $properties The properties, in array and key-value form, to run on this processor
+     * @param bool $merge Indicates if properties should be merged with existing ones
+     *
      * @return void
      */
-    public function setProperties($properties) {
+    public function setProperties($properties, $merge = true) {
         unset($properties['HTTP_MODAUTH']);
-        $this->properties = array_merge($this->properties,$properties);
+        $this->properties = $merge ? array_merge($this->properties,$properties) : $properties;
     }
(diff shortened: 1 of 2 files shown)


Jako:
I would use `&replace = false` instead of `&merge = true`, because disabling a default enabled property instead of enabling an additional property is somehow strange. Also I have to think twice that not merge means replace. But thats a personal opinion. Both ways are valid.

Jako:
I would use `$replace = false` instead of `$merge = true`, because disabling a default enabled property instead of enabling an additional property is somehow strange. Also I have to think twice that not merge means replace. But thats a personal opinion. Both ways are valid.

JoshuaLuckers:
The reason I used `$merge` is to keep it "consistent" with the other `setProperties` functions. See https://github.com/modxcms/revolution/issues/10337#issuecomment-569443253 for more details.

Jako (approved):

opengeek:
This seems like terrible behavior. Is this really something we want to normalize? Why would you not want the properties to be merged, ever?

JoshuaLuckers:
> This seems like terrible behavior. Is this really something we want to normalize? Why would you not want the properties to be merged, ever?

Yes it is, issue #10337 is about making `merge` the default. This PR only makes it an option instead of "no choice" using the existing behavior as default.

theboxer (approved):

PR #15066, opened 2020-04-26
Collapse/expand all child nodes in elements tree

### LINKED ISSUE #14829: Add button to collapse elements
### Summary
It would be convenient to add a link to the context menu to collapse the entire elements tree (the tree of elements, files, resources), collapse all the trees except the desired one and a collapse separate group (TV, chunks; contexts; file sources).

Often the trees are opened and you have to scroll and look for the desired element, and it would be more convenient to work with such a link.

As I see it (Perhaps it makes sense to add a collapse button for a separate category of elements.):

collapse

I understand that there is a triangle for collapse a group of elements, but “collapse all elements except the current group” is more convenient, as it seems to me.

GulomovCreative:
### What does it do?
Added 2 menu items in root menu of elements tree for collapse/expand all child nodes

### Related issue(s)/PR(s)
Partially closes #14829

### DIFF
--- a/manager/assets/modext/widgets/element/modx.tree.element.js
+++ b/manager/assets/modext/widgets/element/modx.tree.element.js
@@ -563,6 +563,28 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
             });
         }
 
+        if (n.hasChildNodes()) {
+            m.push('-');
+
+            if (n.isExpanded()) {
+                m.push({
+                    text: _('collapse_all')
+                    ,handler: function() {
+                        n.collapseChildNodes();
+                    }
+                });
+            }
+
+            m.push({
+                text: _('expand_all')
+                ,handler: function() {
+                    if (n.isExpandable()) {
+                        n.expand(true);
+                    }
+                }
+            });
+        }
+
         return m;
     }
 
--- a/manager/assets/modext/widgets/element/modx.tree.element.js
+++ b/manager/assets/modext/widgets/element/modx.tree.element.js
@@ -563,26 +563,30 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
             });
         }
 
-        if (n.hasChildNodes()) {
-            m.push('-');
+        if (n.isLoaded()) {
+            var childNodes = n.childNodes;
 
-            if (n.isExpanded()) {
+            if (childNodes.some(function(child) { return !child.leaf; })) { // If any childNode has own children
+                m.push('-');
+    
+                if (n.isExpanded() && childNodes.some(function(child) { return child.isExpanded(); })) { // If any childNode is expanded
+                    m.push({
+                        text: _('collapse_all')
+                        ,handler: function() {
+                            n.collapseChildNodes();
+                        }
+                    });
+                }
+    
                 m.push({
-                    text: _('collapse_all')
+                    text: _('expand_all')
                     ,handler: function() {
-                        n.collapseChildNodes();
+                        if (n.isExpandable()) {
+                            n.expand(true);
+                        }
                     }
                 });
             }
-
-            m.push({
-                text: _('expand_all')
-                ,handler: function() {
-                    if (n.isExpandable()) {
-                        n.expand(true);
-                    }
-                }
-            });
         }
 
         return m;


Ruslan-Aleev:
Nice :)

Ibochkarev:
55ee53bada4942c679f17df1ebd91251

JoshuaLuckers:
Nice work! Is it possible to hide the options when there is nothing to collapse/expand? E.g when there are no "parent" / "child" nodes for a group of elements.
Screen Shot 2020-04-26 at 16 58 13

Ruslan-Aleev:
And the option "Collapse everything except the current" would be useful :)

GulomovCreative:
> Is it possible to hide the options when there is nothing to collapse/expand? E.g when there are no "parent" / "child" nodes for a group of elements.

No, because before loading we don’t know if the node has children or not

image

GulomovCreative:
@JoshuaLuckers this is after loading
image

JoshuaLuckers:
> @JoshuaLuckers this is after loading
> image

Once we know the childNodes (after loading) we can check if we should show the collapse/expand option. They only need to be shown if any of the elements has children. If any of the childNodes has `leaf: false` it means we can expand/collapse.

Thanks again for this PR, I really like it. If you need help let me know!

GulomovCreative:
@JoshuaLuckers You can check last commit, i added more conditions for show

JoshuaLuckers (approved):
Nice work, thanks!

manager/assets/modext/widgets/element/modx.tree.element.js
```diff
@@ -563,6 +563,32 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
             });
         }
 
+        if (n.isLoaded()) {
+            var childNodes = n.childNodes;
+
+            if (childNodes.some(function(child) { return !child.leaf; })) { // If any childNode has own children
```
mishantrop, line 569: If you use cool array functions, why not to use arrow functions?
`if (childNodes.some(function(child) { return !child.leaf; })) {`
VVV
`if (childNodes.some(child => !child.leaf)) {`

manager/assets/modext/widgets/element/modx.tree.element.js
```diff
@@ -563,6 +563,32 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
             });
         }
 
+        if (n.isLoaded()) {
+            var childNodes = n.childNodes;
```
mishantrop, line 567: Hmmm...
`const childNodes = n.childNodes;`?

manager/assets/modext/widgets/element/modx.tree.element.js
```diff
@@ -563,6 +563,32 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
...
+
+            if (childNodes.some(function(child) { return !child.leaf; })) { // If any childNode has own children
+                m.push('-');
+    
+                if (n.isExpanded() && childNodes.some(function(child) { return child.isExpanded(); })) { // If any childNode is expanded
```
mishantrop, line 572: `childNodes.some(function(child) { return child.isExpanded(); })`
VVV
`childNodes.some(child => child.isExpanded())`

mishantrop: Useless comment (`// If any childNode is expanded`)

manager/assets/modext/widgets/element/modx.tree.element.js
GulomovCreative: You came right after ES6 courses?! Do you see any polyfill in MODX?

manager/assets/modext/widgets/element/modx.tree.element.js
GulomovCreative: Hmmm....
`let childNodes = n.comeBackToCourses;`

manager/assets/modext/widgets/element/modx.tree.element.js
GulomovCreative: > Useless comment (`// If any childNode is expanded`)

All 4 of your comments are useless

manager/assets/modext/widgets/element/modx.tree.element.js
mishantrop: Hmmmmmmm...
If a childNodes is never reassigned, using the `const` is better. 

> let childNodes = n.comeBackToCourses;

Ha. Ha. Ha.

Ibochkarev (approved):
Thanks!

theboxer (approved):

PR #15069, opened 2020-05-02
Fix typeAhead parameter check

Jako:
### What does it do?
Fix the check of the typeAhead parameter

### Why is it needed?
With the current check the else branch is never reached, since the value will contain the string 'true' or 'false', which could not be wrong.

### Related issue(s)/PR(s)
Detected during testing #15044

### DIFF
--- a/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
+++ b/manager/templates/default/element/tv/renders/input/listbox-multiple.tpl
@@ -34,7 +34,7 @@ Ext.onReady(function() {
         {if $params.title|default},title: '{$params.title|default}'{/if}
         {if $params.listWidth|default},listWidth: {$params.listWidth|default}{/if}
         ,maxHeight: {if $params.maxHeight|default}{$params.maxHeight|default}{else}300{/if}
-        {if $params.typeAhead|default}
+        {if $params.typeAhead == 1 || $params.typeAhead == 'true'}
             ,typeAhead: true
             ,typeAheadDelay: {if $params.typeAheadDelay|default && $params.typeAheadDelay|default != ''}{$params.typeAheadDelay|default}{else}250{/if}
             ,editable: true
--- a/manager/templates/default/element/tv/renders/input/listbox-single.tpl
+++ b/manager/templates/default/element/tv/renders/input/listbox-single.tpl
@@ -21,7 +21,7 @@ Ext.onReady(function() {
         {if $params.title|default},title: '{$params.title|default}'{/if}
         {if $params.listWidth|default},listWidth: {$params.listWidth|default}{/if}
         ,maxHeight: {if $params.maxHeight|default}{$params.maxHeight|default}{else}300{/if}
-        {if $params.typeAhead|default}
+        {if $params.typeAhead == 1 || $params.typeAhead == 'true'}
             ,typeAhead: true
             ,typeAheadDelay: {if $params.typeAheadDelay|default && $params.typeAheadDelay|default != ''}{$params.typeAheadDelay|default}{else}250{/if}
         {else}
--- a/manager/templates/default/element/tv/renders/input/resourcelist.tpl
+++ b/manager/templates/default/element/tv/renders/input/resourcelist.tpl
@@ -21,7 +21,7 @@ Ext.onReady(function() {
         {if $params.title|default},title: '{$params.title}'{/if}
         {if $params.listWidth|default},listWidth: {$params.listWidth}{/if}
         ,maxHeight: {if $params.maxHeight|default}{$params.maxHeight}{else}300{/if}
-        {if $params.typeAhead|default}
+        {if $params.typeAhead == 1 || $params.typeAhead == 'true'}
             ,typeAhead: true
             ,typeAheadDelay: {if $params.typeAheadDelay && $params.typeAheadDelay != ''}{$params.typeAheadDelay}{else}250{/if}
         {else}


Ruslan-Aleev:
Are you sure about the `typeAhead` parameter? I tried to set `false` only and it still works.
What is the `typeAhead` parameter responsible for? And the `typeAheadDelay` parameter does not work.
Those fixes do not affect his behavior, as it seems to me (I tested for 3.x).

Ruslan-Aleev:
Another issue with the `editable` parameter, in my opinion it should always be `true` and not be included in the `typeAhead` condition (https://github.com/modxcms/revolution/pull/15069/files#diff-f398fbbe2e479ec1e91d444a55b5620dR40)

Jako:
If you enable typeAhead in a ExtJS combo or superboxselect, you get a cursor in the combo. With this cursor you could filter the list of the dropdown. Without typeAhead, the cursor is not available and you can't filter the list.

Ruslan-Aleev:
I figured out, the parameter is not so simple (for example - https://github.com/modxcms/revolution/pull/15069/files#diff-d57f12c4ff93835de3276c38589d93f8R28).

If you set `editable: false`, then the cursor does not appear at all, i.e. it is the `editable` parameter that is **responsible for the cursor**:

list-notedit

If you set `editable: true`, cursor will appear:

list-edit

If you set `typeHead: true` (provided that `editable: true`), the **list will "print the rest text of the list item"**. Here's what it looks like with `typeAheadDelay: 2000` (becomes more noticeable):

list-typehead-2000

Ruslan-Aleev:
In general, in my opinion, the `editable` parameter is enough and it should always be `true`.
Although we can leave the parameters `typeHead` and `typeAheadDelay`, but they are not obvious and do not work without `editable`.

Jako:
Thanks for the clarification. I thought that the typeAhead property is responsible for filtering the list. Maybe this is the case in SuperBoxSelect (Listbox multi-select).

Ruslan-Aleev:
And what do we do with these parameters? In a good way, we need to create a separate setting -  "search through list items" (such as turning on the cursor) and it should be turned on by default.
And `typeHead` settings should be located after.
In the current version, useful functionality (`editable`) is hidden in the `typeHead` condition and is turned off by default.

Jako:
editable and typeAhead could be enabled separatly. So enabling editable should be added for the combos.

Ruslan-Aleev:
There is still a problem with setting "**Force Selection to List:**" (inopt_forceSelection).
It allows you to enter your values not from the list **and this is strange**. And for a multiple list, it does not appear in the resource, although it is displayed strangely on the site.

Listbox (Single-Select):

listbox_s_forse

Listbox (Multi-Select):

listbox_m_forse

Ruslan-Aleev:
I suggest delete the settings `forceSelection` and `typeAheadDelay`, leaving only `ediatble`, we can also leave `typeHead`, but I don’t understand why this setting is needed?
@Jako @JoshuaLuckers @Mark-H What do you think?

theboxer:
`forceSelection` limits editable select boxes to only values that are in the select box, if you set that to `false` you can enter any value you want.

`typeAhead` autosolects the first result in the select box that partially matches what you are typing after the `typeAheadDelay`.

Ruslan-Aleev:
I know that :)
But there are problems in these parameters.

- `forceSelection`:
1) Why let the user enter values other than those specified in the "Input Option Values"?
2) For Listbox (Multi-Select) user-defined values does not appear in the resource, although it is displayed on the site.
https://github.com/modxcms/revolution/pull/15069#issuecomment-645930191

- `typeAhead` and `typeAheadDelay`:
1) `typeAhead` autosolects the first result in the select box. But this is not convenient, because the user may not need the first one in the list. With the `ediatble` parameter, you can filter by typing and select the one you need from the list.
https://github.com/modxcms/revolution/pull/15069#issuecomment-624567577
2) `typeAheadDelay` a completely useless parameter.

Too many options that do similar things.

BenceSzalai:
Please don't remove `forceSelection` setting!
- We are using it.
- It has just been added to **Listbox (Multi-Select)** type TV in [#14957](https://github.com/modxcms/revolution/pull/14957)


> 2. For Listbox (Multi-Select) user-defined values does not appear in the resource form, although it is displayed on the site.

I thought this was the expected behaviour, as why user-defined values would change the TV definition? It's not the role of a  content editor to update or change a TV definition, even if that's happening indirectly. In fact I'd consider that an issue, TV definitions drifting away from their designed state configured by the Admin just because of content editors' input... Also how to handle that from permission perspective, content editors effectively changing TV definitions? I don't think that's the right direction. And also an issue, if the TV definition is updated, the value would stick there forever, even if it is not used any more. And what if someone makes a typo. Than the admin has to go to the TV editor to fix that.

I've shown the easy workaround for populating the TV selection list with both default and user entered values, without a need to update the TV definition, plus unused items are eliminated without manual maintenance after the last usage is removed. Nice and clean. See here: [#14957 (comment)](https://github.com/modxcms/revolution/pull/14957#issuecomment-650092984)

theboxer (approved):

Ruslan-Aleev:
@theboxer There are a question with typeAhead parameter, I would not merge this PR.

theboxer:
@Ruslan-Aleev  I don't see any questions against merging this fix.

You are proposing to nuke this option which others opposed to cause they are using the setting. If there is a need for removing this options, I'd suggest starting a new issue where it can be properly discussed. But no matter what the result will be, the property is broken ATM and this PR fixes it.

Ruslan-Aleev:
@theboxer I'm talking about https://github.com/modxcms/revolution/pull/15069/files#diff-f398fbbe2e479ec1e91d444a55b5620dR40
If typeAhead is turned off, the field will still be editable for typing.

Ruslan-Aleev:
I will create a separate issue so that no one gets confused in the extra discussion.

theboxer:
@Ruslan-Aleev I understand what you are saying, but that's not the goal of this PR. The `typeAhead` param is broken right now and needs fixing. And even without this fix the multiselect's input was editable for typing no matter what you set.

IMHO `editable` should be another standalone param as well (added through a new PR).

Ruslan-Aleev:
Yes, as for the `ediatble` parameter, I created a separate issue so as not to get confused in the discussion.

PR #15070, opened 2020-05-03
Fix duplicating static elements

JoshuaLuckers:
### What does it do?
- Make it possible to use MODX Media Browser to select a file when duplicating static elements.
- Implement `MODx.getStaticElementsPath` when automatic static elements is enabled 

#### Work in progress
- ~~Add a "duplicate-of" prefix to the "static file" when automatic static elements is *not* enabled~~
- ~~Allow users to select/use an existing file as "static file"~~

### Why is it needed?
To solve issues when duplicating static elements. Currently you have to change the filename manually when duplicating static elements.

### Related issue(s)/PR(s)
Issue #14188

### DIFF
--- a/manager/assets/modext/sections/element/chunk/update.js
+++ b/manager/assets/modext/sections/element/chunk/update.js
@@ -55,6 +55,7 @@ Ext.extend(MODx.page.UpdateChunk,MODx.Component, {
             ,type: 'chunk'
             ,name: _('duplicate_of',{name: this.record.name})
             ,source: this.record.source
+            ,openTo: this.record.openTo
             ,static: this.record.static
             ,static_file: this.record.static_file
             ,category: this.record.category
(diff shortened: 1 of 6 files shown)


Jako:
If I remember right, you can't insert an existing path on static file creation even when automatic static elements is not enabled. Please cross check this too.

JoshuaLuckers:
> If I remember right, you can't insert an existing path on static file creation even when automatic static elements is not enabled. Please cross check this too.

You are right, in the duplicate processor there is a check to see if the file already exists.

theboxer:
@JoshuaLuckers any progress on this one?

JoshuaLuckers:
No, I hope to have some time to finish it in the upcoming weeks..

Ibochkarev:
@JoshuaLuckers any progress on this one?

JoshuaLuckers:
It's still on my todo list, but if anyone wants to take over this PR and finish it please let me know

JoshuaLuckers:
This PR is ready for review, once merged others can continue my work if needed!

alroniks:
Hm, I tried to duplicate the existing static chunk and select a new file (created via phpstorm), but I see such an error.
Chunk_ A Chunk _ MODX Revolution

PR #15065, opened 2020-04-25
Corrected TV form with the Legacy multiple list type

Ruslan-Aleev:
### What does it do?
Corrected the TV form with the Legacy multiple list type:
- Changed the order
- Related items grouped
- **Added the ability to set the list size in %**

Later I will correct for TV with other types.

**Before:**
tv-list-legacy-1

**After:**
tv-list-legacy-2

### Why is it needed?
Improves UI / UX

### Related PR(s)
https://github.com/modxcms/revolution/pull/15064
https://github.com/modxcms/revolution/pull/15063
https://github.com/modxcms/revolution/pull/15045
https://github.com/modxcms/revolution/pull/15044
https://github.com/modxcms/revolution/pull/15043
https://github.com/modxcms/revolution/pull/15042
https://github.com/modxcms/revolution/pull/15009

PR #15072, opened 2020-05-05
Corrected width for allowBlank in TV form

Ruslan-Aleev:
### What does it do?
Corrected width for allowBlank in TV form with the type:
- Autotag
- Tag
- Textarea

For the rest of the TV, allowBlank is already fixed in PRs.

**Before:**
blank-before

**After:**
blank-after

### Why is it needed?
Improves UI / UX

### Related PR(s)
https://github.com/modxcms/revolution/pull/15065
https://github.com/modxcms/revolution/pull/15064
https://github.com/modxcms/revolution/pull/15063
https://github.com/modxcms/revolution/pull/15045
https://github.com/modxcms/revolution/pull/15044
https://github.com/modxcms/revolution/pull/15043

### DIFF
--- a/manager/templates/default/element/tv/renders/inputproperties/autotag.tpl
+++ b/manager/templates/default/element/tv/renders/inputproperties/autotag.tpl
@@ -23,7 +23,7 @@ MODx.load({
         ,name: 'inopt_allowBlank'
         ,hiddenName: 'inopt_allowBlank'
         ,id: 'inopt_allowBlank{/literal}{$tv|default}{literal}'
-        ,width: 200
+        ,anchor: '100%'
         ,value: (params['allowBlank']) ? !(params['allowBlank'] === 0 || params['allowBlank'] === 'false') : true
         ,listeners: oc
     },{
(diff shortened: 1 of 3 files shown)


Ibochkarev (approved):
Work

image

image

image

garryn:
Why do we need a full-width dropdown? I can understand the whole idea of bringing it back inline with the main TV dropdown but why is the main TV dropdown full-width?

Just seems like it could be a nightmare for people on large monitors etc.

Ruslan-Aleev:
@garryn It is made for uniformity in design and to avoid potential problems on mobile devices.
The width of the "Input Type" does not bother you, although it is also 100%.
In addition, this list is activated when you click on any place, I see no problems.

allow-width

garryn:
I'm thinking about the end user here, it's no good for them if the arrow is way off to the right of the screen (especially on 4K monitors etc.) - they may just think it's a textfield and be confused when it gives options. My concerns are purely from a UX perspective and I do understand the mobile aspect of this too.

Ruslan-Aleev:
I understand you, however, your concern would be justified if the other elements of the manager panel would meet your requirement ("Input Type", "Output Options" and other elements), but this does not happen, and users are already used to this behavior.
And for other TVs, this design has already been implemented, i.e. except you, it didn’t bother anyone :)

Ruslan-Aleev:
And even if the user confuses the drop-down list with a text field, he will understand this after the first click :)

garryn:
>>> i.e. except you, it didn’t bother anyone :)

Just playing Devil's Advocate and not making assumptions :P

JoshuaLuckers (approved):
I agree with both of you. For now I think it's good to have it consistent with the other inputs but we should definitely look into reorganizing / redesigning the forms.

Issue #6521, opened 2011-12-29
Add "Allow Blank" setting to all TV input type


modxbot:
yogooo created Redmine issue ID 6521

Add the "Allow Blank" setting to all TV input type.
Currently unavailable for type :
- File
- Hidden
- Image
- Rich Text
- URL

I'm currently setting up a website where non-empty images and url TVs are required. "Allow Blank" is a super useful option and it would be even more helpful if it was available for all TV types.

JoshuaLuckers:
Is there any progress on this issue?

JoshuaLuckers:
After doing a bit of research I leave this here as a reference. 

The file and image TV use the `modx-combo-browser` xtype  `manager/assets/modext/widgets/core/modx.combo.js`. 

The component `MODx.combo.Browser` extends `Ext.form.TriggerField` and has support for `allowBlank: false`. However setting `allowBlank: false` in `MODx.combo.Browser` marks the input field as "invalid" because it's empty when rendered. Saving the resource is still possible even if the input field is blank.

Ruslan-Aleev:
Does anyone have any ideas on how to fix it? TV in MODX is one of the coolest features and it is strange that TVs do not work correctly.

JoshuaLuckers:
Thanks for reminding me, I'll put it back on my todo list!

Ruslan-Aleev:
@theboxer When you have the desire and time, can you look at this issue? Maybe some ideas come up.

JoshuaLuckers:
Thanks @Ruslan-Aleev !

PR #15073, opened 2020-05-05
Added allowBlank for TV with URL type

Ruslan-Aleev:
### What does it do?
Added allowBlank for TV with URL type

url-req

### Why is it needed?
For some reason, allowBlank is not for all TVs, which is strange

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/6521 - URL

### DIFF
--- a/manager/templates/default/element/tv/renders/input/url.tpl
+++ b/manager/templates/default/element/tv/renders/input/url.tpl
@@ -1,37 +1,48 @@
 <select id="tv{$tv->id}_prefix" name="tv{$tv->id}_prefix" onchange="MODx.fireResourceFormChange();">
-{foreach from=$urls item=url}
-	<option value="{$url}" {if $url == $selected|default}selected="selected"{/if}>{$url}</option>
-{/foreach}
+    {foreach from=$urls item=url}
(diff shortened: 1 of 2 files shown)


Ibochkarev (approved):
Work!
image

manager/templates/default/element/tv/renders/input/url.tpl
```diff
@@ -1,37 +1,48 @@
...
         ,listeners: { 'select': { fn:MODx.fireResourceFormChange, scope:this}}
     });
+    fld.wrap.applyStyles({
+        display: "inline-block"
+    });
 
-	fld.wrap.applyStyles({
-		display: "inline-block"
-	});
+    var fld = MODx.load({
```
Jako, line 36: Is this addition really needed?

Ruslan-Aleev: Well, this is a text field handler, it is for him that you need `allowBlank: false`, because links can be without the selected type in the drop-down list.
Or did I misunderstand you?

garryn:
This seems a good addition @opengeek (subject to testing of course)

JoshuaLuckers (approved):
Works as intended.

PR #15074, opened 2020-05-05
Added allowBlank for TV with RichText type

Ruslan-Aleev:
### What does it do?
Added allowBlank for TV with RichText type

richtext-req

### Why is it needed?
For some reason, allowBlank is not for all TVs, which is strange

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/6521 - RichText
https://github.com/modxcms/revolution/pull/15073

PR #15075, opened 2020-05-07
Added allowBlank for TV with Image and File type

Ruslan-Aleev:
### What does it do?

Added allowBlank for TV with Image and File type

tv-files-req

### Why is it needed?
For some reason, allowBlank is not for all TVs, which is strange

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/6521 - Image and File
https://github.com/modxcms/revolution/pull/15074
https://github.com/modxcms/revolution/pull/15073

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -106,13 +106,17 @@ Ext.extend(MODx,Ext.Component,{
                     if (cmp.allowBlank !== false) return;
 
                     var applyTo = cmp.applyTo;
+                    var label = cmp.fieldLabel;
                     if (!applyTo) {
-                        var label = cmp.fieldLabel;
-                        cmp.fieldLabel = markdom+label;
-                    } else if (applyTo && applyTo.match(/^tv[\d]*$/i)) { 
-                        var label = document.getElementById(applyTo+'-caption');
-                        var html = markdom+label.innerHTML;
-                        label.innerHTML = html;
+                        if (label) {
+                            cmp.fieldLabel = markdom + label;
+                        } else if (cmp.caption && document.getElementById(cmp.caption)) {
+                            label = document.getElementById(cmp.caption);
+                            label.innerHTML = markdom + label.innerHTML;
+                        }
+                    } else if (applyTo && applyTo.match(/^tv[\d]*$/i)) {
+                        label = document.getElementById(applyTo+'-caption');
+                        label.innerHTML = markdom+label.innerHTML;
                     }
                 }
             });
--- a/manager/assets/modext/widgets/element/modx.panel.tv.renders.js
+++ b/manager/assets/modext/widgets/element/modx.panel.tv.renders.js
@@ -1,6 +1,6 @@
 /**
  * Renders an input for an image TV
- * 
+ *
  * @class MODx.panel.ImageTV
  * @extends MODx.Panel
  * @param {Object} config An object of configuration properties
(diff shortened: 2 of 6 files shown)


JoshuaLuckers:
> I can also display a message that the field is required, but it will disappear only after saving (since the value has not yet been set to the hidden field)

Could this be solved by adding a listener that fires `MODx.fireResourceFormChange` when the `change` event happens instead of the `keydown`? I believe the date TV works similar: https://github.com/modxcms/revolution/blob/f64b6063bd6256106c9a2db89d9febfa850a82ac/manager/templates/default/element/tv/renders/input/date.tpl#L32

Ruslan-Aleev:
Unfortunately, there is no difference.

The value of the hidden field changes, but `listeners: {'change': {fn: MODx.fireResourceFormChange, scope: this}}` will work if the value is changed directly in the field (as I understand it), but since If the field is hidden, then the error does not disappear.

Here's what it looks like if the hidden field is displayed:
img-req-3

Initially, the problem is that the TV with the output of the file browser (image and file) does not have validation, it does not work, and you have to catch the hidden field validation.

_By the way, on TV with the date type error messages are also not shown :)_

Jako:
The hidden field needs the change listener then.

Ruslan-Aleev:
I changed PR, it works, but it seems to me that it can be done more correctly and simply :)

tv-files-req

Jako (changes requested):
Yes, it could be solved a lot simpler: The PR should contain an additional `,msgTarget: 'under'` in the in the defaults section and the already made `,allowBlank: config.allowBlank` addition to FileTV/ImageTV in `modx.panel.tv.renders.js` and no changes in `/renders/input/*.tpl`

Ruslan-Aleev:
@Jako It doesn’t work, your variant adds a message below to select files, which is useful.
But `allowBlank: false` does not work for selecting files, it saves in any case. Therefore, we will need to create a second hidden field anyway.

tv-files-nreq

Ruslan-Aleev:
@Jako @JoshuaLuckers I updated the PR, redid it easier, but without a hidden field I can’t do it anyway.

Jako:
Ok, I forgot to check wether the resource could be saved without the requirement.

Jako:
The file/image tv already contains a hidden field in `manager/assets/modext/widgets/element/modx.panel.tv.renders.js`, but this field seems not to trigger the validation of the resource form.

Jako:
I have changed the code for the image tv
- add the field to the resource form (https://github.com/modxcms/revolution/pull/15075/files#diff-a71262af33263e3041a682eddb38f566R72)
- add a validation for the image tv (https://github.com/modxcms/revolution/pull/15075/commits/458a73da0b6f65484a46c7607e2dcbd085fd373f#diff-a71262af33263e3041a682eddb38f566R58)

The validation could be maybe moved to a method in the xtype.

The same change has to be done for the file tv.

Jako (changes requested):
I have changed the code for the image tv. It has to be done for the file tv.

Ruslan-Aleev:
@Jako Thanks for the help, it looks great. I will test and make for the file TV.

Ruslan-Aleev:
@Jako have you tested your variant? It does not work for me (js and css compression is disabled, tested in incognito mode):

- **Does not save the resource, even if the TV is not empty**
- No * next to the name of required field

files-req-5

Jako:
Now it gets the right value.

Jako:
And now the asterisk is shown on required TVs. This MODx component is a big mess.

Ruslan-Aleev:
As a result, after many edits, we returned almost to what it was :)
It's cool that it works, this functionality is extremely lacking.

Jako:
I tried to remove the annoying MODx.load just to get the asterisk shown on the TV label. It works fine here, but please do a cross check on your installation too.

Jako (approved):

JoshuaLuckers (approved):
Great work!

PR #15076, opened 2020-05-08
Fix setting is_folder after appending resource to another resource

### LINKED ISSUE #14747: Resource setting isfolder is not set to true if you drag another resource into it
### Summary
**Resource setting isfolder is not set to true if you drag another resource into it. But if you create a child resource, then the isfolder will become true (what's the difference?)**

There are a lot of different closed issues about this behavior, but there are no changes and the behavior is strange, as it seems to me.

isfolder

Probably related https://github.com/modxcms/revolution/issues/13722

### Environment
MODX 2.7.1

GulomovCreative:
### What does it do?
Fix the logical mistake in fixParents method of resource sorting processor. $node variable still have unsaved object and $newParent get old parent.

### Why is it needed?
After moving the resource, his new parent is_folder field don't set to true

### Related issue(s)/PR(s)
Closes #14747

### DIFF
--- a/core/src/Revolution/Processors/Resource/Sort.php
+++ b/core/src/Revolution/Processors/Resource/Sort.php
@@ -171,7 +171,7 @@ class Sort extends Processor
         $oldParent = $this->modx->getObject(modResource::class, $nodeObject->parent);
 
         /** @var modResource $newParent */
-        $newParent = $this->modx->getObject(modResource::class, $node->parent);
+        $newParent = $this->modx->getObject(modResource::class, $this->target->id);
 
         if (empty($oldParent) && empty($newParent)) return;
         if ($oldParent->id == $newParent->id) return;


Ruslan-Aleev (approved):
Thank. It works :)

JoshuaLuckers (approved):

Ibochkarev (approved):

opengeek:
This has been discussed so many times. The `isfolder` property should NOT be automatically changed just because you drag a resource into as a child.

Ruslan-Aleev:
Why not? I see no logical reason why the isfolder does not become true when dragging resource into another.
A folder means something in it. Which is logical if you dragged something into the resource.
And the fact that this has been discussed many times indicates that “not changing” - not the right decision.

theboxer (approved):

Issue #15003, opened 2020-02-08
When creating a new element, "undefined" is indicated instead of ID


Ruslan-Aleev:
### Summary
When creating a new element, "undefined" is indicated instead of ID.
Previously, the ID appeared in brackets only after saving the element, and when creating nothing was displayed.

**This happens for all elements: templates, TV, chunks, snippets and plugins.**

### How it look like (for example, in TV)
![tv](https://user-images.githubusercontent.com/12523676/74089622-91e17380-4abc-11ea-8865-03f8030dc8dc.gif)

### Expected behavior
Do not display anything.

### Environment
MODX 3 alpha-2

Ruslan-Aleev:
Here the problem is that in the if-condition the action is indicated in capital letters (https://github.com/modxcms/revolution/blob/3.x/manager/assets/modext/widgets/element/modx.panel.tv.js#L78), but in link is in lowercase letters.
Although for TV there the action is indicated incorrectly, there should be `Element/TV/Create`.

Related with https://github.com/modxcms/revolution/issues/14744

PR #15077, opened 2020-05-08
Fix conditions with request action [JS]

GulomovCreative:
### What does it do?
Fix all condition checks of request action in js

### Why is it needed?
Conditions don't work because request actions in lowercase

### Related issue(s)/PR(s)
null


Ruslan-Aleev:
Related issue - https://github.com/modxcms/revolution/issues/15003

JoshuaLuckers:
I have some ideas and thoughts to better fix this problem across the board. I'll come back to it upcoming Tuesday!

GulomovCreative:
Ok, we'll wait

JoshuaLuckers (approved):

theboxer (approved):

PR #15078, opened 2020-05-08
Changes isfolder field dynamically on active resource

GulomovCreative:
### What does it do?
Changes isfolder field of editing resource after sort process

### Why is it needed?
If we append a resource to the resource being edited, the value of the isfolder field in the DB will change, but it will remain the same in the form of editing. 

### Attention

This PR needs to be tested after merging next PRs: #15076 and #15077 because they are related, but it doesn't make sense to put them together in one PR. 

### Related issue(s)/PR(s)
#15076 #15077

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.tree.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.tree.resource.js
@@ -402,7 +402,7 @@ Ext.extend(MODx.tree.Resource,MODx.tree.Tree,{
                 menuindexField.setValue(o.result.object.menuindex);
             }
             var isfolderFieldCmb = Ext.getCmp('modx-resource-isfolder');
-            if(isfolderFieldCmb && o.result.object.isfolder !== undefined && typeof o.result.object.isfolder === 'boolean'){
+            if(isfolderFieldCmb && typeof o.result.object.isfolder === 'boolean'){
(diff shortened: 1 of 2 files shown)


manager/assets/modext/widgets/resource/modx.tree.resource.js
```diff
@@ -401,6 +401,10 @@ Ext.extend(MODx.tree.Resource,MODx.tree.Tree,{
             if(menuindexField && o.result.object.menuindex !== undefined){
                 menuindexField.setValue(o.result.object.menuindex);
             }
+            var isfolderFieldCmb = Ext.getCmp('modx-resource-isfolder');
+            if(isfolderFieldCmb && o.result.object.isfolder !== undefined && typeof o.result.object.isfolder === 'boolean'){
```
JoshuaLuckers: Is it possible to remove the `o.result.object.isfolder !== undefined` conditional? If I'm not mistaken the conditional `typeof o.result.object.isfolder === 'boolean'` returns false if `o.result.object.isfolder` is `undefined`. 

```suggestion
            if(isfolderFieldCmb && typeof o.result.object.isfolder === 'boolean'){
```

JoshuaLuckers (approved):

Ibochkarev (approved):

theboxer (approved):

PR #15079, opened 2020-05-11
Remove deprecated modClassMap

alroniks:
### What does it do?
It removes old deprecated model class `modClassMap`, that exists in the schema, but not used actually since MODX version 2.2.

### Why is it needed?
To clean up codebase from obsolete and unused code.

### Related issue(s)/PR(s)
It brokes Property Sets separate page (not a tab on each element) a bit, but I would like to fix that page separately, they're a lot of bugs and inconsistent states.


opengeek:
I seem to remember a discussion about not removing this until 3.1.x, but I could be imagining it. @Mark-H, do you recall anything about this?

alroniks:
I do not see any risk in removing it. As it does not used for a long time, better to remove it than support it. Supporting the dead code is as wasting time.

@opengeek @Mark-H @JoshuaLuckers Please, say your opinion.


@Ibochkarev Could you please validate it?

JoshuaLuckers (approved):

Ibochkarev (approved):

core/model/schema/modx.sqlsrv.schema.xml
```diff
@@ -703,7 +685,7 @@
         models to be loaded prior to modX request handling. Useful for when you want to integrate custom data and
         extensions, such as Custom Resource Classes, into the core.
     -->
-    <object class="modExtensionPackage" table="extension_packages" extends="xPDO\Om\xPDOObject">
+    <object class="modExtensionPackage" table="extension_packages" extends="xPDO\Om\xPDOSimpleObject">
```
opengeek, line 688: What is this change doing in here? This would need an upgrade script if you are adding an auto_increment primary key to this table.

alroniks: It fixes sqlsrv schema issue, because actually in the code it extends xPDOSimpleObject.

https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/modExtensionPackage.php#L23
https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/mysql/modExtensionPackage.php#L13

But sqlsrv implementation extends  xPDOObject. https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/sqlsrv/modExtensionPackage.php#L13

So, question where is an issue. Mysql seems configures properly, sqlsrv probably works, maybe not, but who cares? :)

opengeek: Doh! I didn't even notice the sqlsrv schema thing. And yes, who cares, indeed!

PR #15096, opened 2020-06-10
Fix the name of the create tv action

### LINKED ISSUE #15003: When creating a new element, "undefined" is indicated instead of ID

theboxer:
### What does it do?
Uses correct name of the create TV action in the condition

### Why is it needed?
Without the change, it tries to display TV's ID, even though it's unavailable.

### Related issue(s)/PR(s)
Resolves #15003


Ibochkarev (approved):
It works!

JoshuaLuckers (approved):

PR #15097, opened 2020-06-10
Unify manager actions to lowercase

### LINKED ISSUE #14744: MODX3: Make links in the manager in one format

theboxer:
### What does it do?
Tries to normalize manager actions to be lower cased. I tried to find as many CamelCase usages as I could and replace them with lower cased version, but not really sure I got them all.

Also the MODx.getPage / MODx.loadPage now should return only lower case action, which should help keeping this as a standard.

### Why is it needed?
As described in #14744, currently lower case and Camel Case versions are mixed and not all conditions works because they are checking wrong case.

### Related issue(s)/PR(s)
Closes #14744

### DIFF
--- a/manager/assets/modext/sections/resource/update.js
+++ b/manager/assets/modext/sections/resource/update.js
@@ -106,7 +106,6 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
             }
             ,listeners: {
                 success: {fn:function(r) {
-                    //MODx.loadPage('resource/update', 'id='+r.object.id);
                     var panel = Ext.getCmp('modx-panel-resource');
                     if (panel) {
                         panel.handlePreview(true);
@@ -126,7 +125,6 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
             }
             ,listeners: {
                 success: {fn:function(r) {
-                    //MODx.loadPage('resource/update', 'id='+r.object.id);
                     var panel = Ext.getCmp('modx-panel-resource');
                     if (panel) {
                         panel.handlePreview(false);
--- a/manager/assets/modext/sections/resource/update.js
+++ b/manager/assets/modext/sections/resource/update.js
@@ -83,7 +83,7 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
                 success: {fn:function(r) {
                     var response = Ext.decode(r.a.response.responseText);
                     if (response.object.redirect) {
-                        MODx.loadPage('Resource/Update', 'id='+response.object.id);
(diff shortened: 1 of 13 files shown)


Ruslan-Aleev:
Great work! Thanks.

Ibochkarev (approved):

Ruslan-Aleev:
Maybe it is worth https://github.com/modxcms/revolution/pull/15096 edits combine with this PR?

theboxer:
That PR uses wrong action, not just different case...

JoshuaLuckers:
I had some "notes" about this but don't have time to work them out in a good idea/issue at the moment. My ramblings can be found here: https://gist.github.com/JoshuaLuckers/08c7a011a137527c8f75dfce5fdf0fb0

Ibochkarev (approved):
image

image

image

manager/assets/modext/sections/resource/update.js
```diff
@@ -126,7 +126,7 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
             }
             ,listeners: {
                 success: {fn:function(r) {
-                    //MODx.loadPage('Resource/Update', 'id='+r.object.id);
+                    //MODx.loadPage('resource/update', 'id='+r.object.id);
```
JoshuaLuckers:
```suggestion
```

JoshuaLuckers (commented):
If you have time to solve the conflict that would be great! 😄

JoshuaLuckers (approved):

Ruslan-Aleev:
Is there no need a fix this?https://github.com/modxcms/revolution/blob/08ea5ac757dcd7013cc201b77073b94a7f2fb811/manager/assets/modext/widgets/security/modx.grid.access.policy.template.js#L228

https://github.com/modxcms/revolution/blob/08ea5ac757dcd7013cc201b77073b94a7f2fb811/manager/assets/modext/widgets/security/modx.grid.access.policy.js#L190

theboxer:
No, these are not manager actions, but processor names.

Ruslan-Aleev:
@theboxer When editing Sets in Form Customization profile, the manager actions in uppercase, I'm about:

fcs_actions

PR #15098, opened 2020-06-10
Set error message by the name of the field instead of id

### LINKED ISSUE #14982: Error messages are not displayed in the quick update window
### Summary
Error messages are not displayed in the quick update window.
For example, when editing symlink, you cannot specify a link to the link itself, but an error does not appear in a quick window.

### What does it look like
symlink_1

symlink_2

### Environment
MODX 2.x >

theboxer:
### What does it do?
Set field error message for the content field by the field name instead of the id.

### Why is it needed?
In the resource quick edit window, the field ID is always different and based on the window's id, so the error won't show up.

### Related issue(s)/PR(s)
Closes #14982


Ibochkarev (approved):
It works!

PR #15102, opened 2020-06-10
Use 0 if id is undefined

### LINKED ISSUE #15090: When changing a template for a new document - &id=undefined appears in the url
### Summary
When changing a template for a new document - &id=undefined appears in the url:

new_doc

This is because there is no verification of the operation — creation or editing: https://github.com/modxcms/revolution/blob/0c1289f79d35417ae3ef4f2393427a42db368dd6/manager/assets/modext/widgets/resource/modx.panel.resource.js#L339

The bug is not critical, but it looks strange.

### Environment
MODX 2.x>

theboxer:
### What does it do?
Uses `0` if `r.result.object.id` is undefined.

### Why is it needed?
Without the check `undefined` is put to the URL.

### Related issue(s)/PR(s)
Closes #15090

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -358,7 +358,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                     this.warnUnsavedChanges = false;
                     MODx.activePage.submitForm({
                         success: {fn:function(r) {
-                            MODx.loadPage(r.result.object.action, 'id='+r.result.object.id+'&reload='+r.result.object.reload + '&class_key='+ r.result.object.class_key + '&context_key='+ r.result.object.context_key);
+                                MODx.loadPage(r.result.object.action, 'id='+(r.result.object.id || 0)+'&reload='+r.result.object.reload + '&class_key='+ r.result.object.class_key + '&context_key='+ r.result.object.context_key);
                         },scope:this}
                     },{
                         bypassValidCheck: true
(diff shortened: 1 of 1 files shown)


manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -358,7 +358,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                     this.warnUnsavedChanges = false;
                     MODx.activePage.submitForm({
                         success: {fn:function(r) {
-                            MODx.loadPage(r.result.object.action, 'id='+r.result.object.id+'&reload='+r.result.object.reload + '&class_key='+ r.result.object.class_key + '&context_key='+ r.result.object.context_key);
+                                MODx.loadPage(r.result.object.action, 'id='+(r.result.object.id || 0)+'&reload='+r.result.object.reload + '&class_key='+ r.result.object.class_key + '&context_key='+ r.result.object.context_key);
```
Ibochkarev:
```suggestion
                            MODx.loadPage(r.result.object.action, 'id='+(r.result.object.id || 0)+'&reload='+r.result.object.reload + '&class_key='+ r.result.object.class_key + '&context_key='+ r.result.object.context_key);
```

Ibochkarev (changes requested):

manager/assets/modext/widgets/resource/modx.panel.resource.js
theboxer: ?

Ibochkarev: Extra tab added. Not critical =)

Ibochkarev (approved):
It works!

manager/assets/modext/widgets/resource/modx.panel.resource.js
theboxer: oh, I read the suggested change wrong. It seemed like you suggested to add an extra tab :) Fixed, thanks for noticing ;)

JoshuaLuckers:
Logically, wouldn't it be better to use `false` or `null`?

theboxer:
0 is used when you are creating new resource, so this is using same thing

PR #15109, opened 2020-06-12
Corrected output of resource data in "Overview" form

Ruslan-Aleev:
### What does it do?
Corrected output of resource data in "Overview" form.
Changed the order and added the missing fields.

res-data_2

### Why is it needed?
- Not all fields were shown.
- Some fields did not display anything.

### Related issue(s)/PR(s)
None


JoshuaLuckers:
It would be very helpful to split the changes into multiple commits. 

Commits for:
- Fields not displaying (the right) values.
- Changing the _current_ fields order.
- Adding the new fields.

This makes it easier to see at a glance what has changed and to undo them if needed. Or even to cherry pick a particular change (e.g the fix for the values).

Ruslan-Aleev:
@JoshuaLuckers I corrected PR

JoshuaLuckers (approved):

PR #15086, opened 2020-06-02
Remember the filter properties of the system setting grid in the browser url

Jako:
### What does it do?
Add the selected settings filter properties in the browser url

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url i.e. for support sessions 

### Related issue(s)/PR(s)
#14086

PR #15115, opened 2020-06-15
Remember the filter properties of the system setting grid in the browser url

Jako:
### What does it do?
Global approach for #15086 in MODx.grid.Grid with a replaceState method and a makeUrl method triggered by an urlFilters config array. Since the key request parameter is used in the context window, the filter 'key' is renamed to 'query'

### Why is it needed?
Since #15086 the user settings page was broken.

### Related issue(s)/PR(s)
#15086

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -715,6 +715,39 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
     actionContextMenu: function(record, recordIndex, e) {
         this._showMenu(this, recordIndex, e);
     }
+
+    ,makeUrl: function () {
+        if (typeof this.config.urlFilters !== "undefined" && this.config.urlFilters !== null && this.config.urlFilters.length !== null && this.config.urlFilters.length > 0) {
+            var s = this.getStore();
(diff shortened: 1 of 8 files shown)


manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -715,6 +715,39 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
     actionContextMenu: function(record, recordIndex, e) {
         this._showMenu(this, recordIndex, e);
     }
+
+    ,makeUrl: function () {
+        if (typeof this.config.urlFilters !== "undefined" && this.config.urlFilters !== null && this.config.urlFilters.length !== null && this.config.urlFilters.length > 0) {
```
theboxer: I believe this can be simplified to `if(Array.isArray(this.config.urlFilters) && this.config.urlFilters.length > 0)`

theboxer (changes requested):

manager/assets/modext/widgets/core/modx.grid.js
Jako: You are right: https://caniuse.com/#search=isArray

theboxer (approved):

PR #15117, opened 2020-06-17
Corrected styles in manager panel

Ruslan-Aleev:
### What does it do?
Corrected some styles in manager panel.

**TV, if located below**
Before:
tv_bottom_1

After:
tv_bottom_2

**Checkboxes**
Before:
tv_radios_1

After:
tv_radios_2

**Number fields in grid**
Before:
num_field_1

After:
num_field_2

**Added hover styles for files in file browser**
file_browser

Previously, hover styles coincided with the styles of the selected file, which can be confusing.

**Changed the breadcrumbs icon in the package manager**

bc_icon

### Why is it needed?
UI fix

### Related issue(s)/PR(s)
None

### DIFF
--- a/_build/templates/default/sass/_breadcrumbs.scss
+++ b/_build/templates/default/sass/_breadcrumbs.scss
@@ -36,10 +36,14 @@
           line-height: 34px;
           position: absolute;
           top: 0;
-          left: 6px;
+          left: 0;
           text-align: center;
           text-indent: 0;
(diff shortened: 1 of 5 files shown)


Ibochkarev (approved):

JoshuaLuckers (approved):

alroniks (approved):
GJ!

Issue #14961, opened 2020-01-21
Display dates in manager in one format


Ruslan-Aleev:
### Summary
In the system settings, there are date and time settings:
- `manager_date_format`
- `manager_time_format`

However, these settings are sometimes not used. For example, in the “Edit File form” and in the “Error log”, the date is displayed in a different format (most likely in other sections too). Although the settings are `d/m/Y` and `H:i`:

It is worth bringing the output of dates and times to the format specified in the settings.

**The sections do not take into account time settings, but should:**
- "Who's Online" widget;
- "Created On" field in "User" edit form;
- "Recent documents" tab in "System Info" section.

### Environment
MODX 2.7 >

Jako:
The error log date format can't be changed (on the fly), the other ones should be formatted with one format.

PR #15118, opened 2020-06-17
Changed styles for widgets

Ruslan-Aleev:
### What does it do?
Changed the styles for widgets, basically made the links more visible.
In the current version, the links are not visible at all.
Which is strange, especially for a MODX Security Feed widget.

Before:
widgets_1

After:
widgets_2

### Why is it needed?
UI/UX improvements

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/templates/default/dashboard/rssitem.tpl
+++ b/manager/templates/default/dashboard/rssitem.tpl
@@ -1,5 +1,5 @@
 <div class="news_article">
     <h2><a href="[[+link]]" target="_blank">[[+title]]</a></h2>
     <div class="content">[[+description]]</div>
-    <div class="date_stamp">[[+pubdate:strtotime:date=`%Y-%m-%d`]]</div>
+    <div class="date_stamp">[[+pubdate]]</div>
 </div>
--- a/manager/templates/default/dashboard/rssitem.tpl
+++ b/manager/templates/default/dashboard/rssitem.tpl
@@ -1,5 +1,5 @@
 <div class="news_article">
     <h2><a href="[[+link]]" target="_blank">[[+title]]</a></h2>
-    <span class="content">[[+description]]</span>
-    <span class="date_stamp">[[+pubdate]]</span>
+    <div class="content">[[+description]]</div>
+    <div class="date_stamp">[[+pubdate:strtotime:date=`%Y-%m-%d`]]</div>
 </div>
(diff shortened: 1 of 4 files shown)


manager/templates/default/dashboard/rssitem.tpl
```diff
@@ -1,5 +1,5 @@
 <div class="news_article">
     <h2><a href="[[+link]]" target="_blank">[[+title]]</a></h2>
-    <span class="content">[[+description]]</span>
-    <span class="date_stamp">[[+pubdate]]</span>
+    <div class="content">[[+description]]</div>
+    <div class="date_stamp">[[+pubdate:strtotime:date=`%Y-%m-%d`]]</div>
```
JoshuaLuckers: There is a system setting `manager_date_format` for customizing date formats in the manager. The format is set without `%`, we need to look how we can apply this setting here.

Jako:
manager_date_format is JavaScript based, the one used here is PHP based.

JoshuaLuckers:
> manager_date_format is JavaScript based, the one used here is PHP based.
> 

I'm confused:
https://github.com/modxcms/revolution/search?l=PHP&q=manager_date_format&type=

Jako:
The format is JavaScript/ExtJS date based and not php strftime based. Maybe it is somehow compatible with php date.

Ruslan-Aleev:
I displayed the date as in the system settings, but did not understand how to apply the `manager_date_format` settings in widget.
There is already a issue with the date output (https://github.com/modxcms/revolution/issues/14961), we can fix it in a separate PR in the future.
This PR is not about the date, but about the styles :)

JoshuaLuckers:
> This PR is not about the date, but about the styles :)

The date is changed in this PR 😋

JoshuaLuckers (approved):

Jako:
Maybe we could work with `date($this->modx->getOption('manager_date_format'))` here to display date with the JavaScript/ExtJS based `manager_date_format`. But it has some issues: It is not using any locale and  maybe not all options of http://cdn.sencha.com/ext/gpl/3.4.1.1/docs/#!/api/Date are usable.

I suggest to output a script tag that formats a placeholder with

```
"<script>
var dateString = new Date(" . date('c', $yourDateTime) . ").format(MODx.config.manager_date_format);
</script>"
```

JoshuaLuckers (approved):

Jako:
I moved the date formatting to ExtJS.

PR #15119, opened 2020-06-18
Removed allowNegative parameter for TV with number type

Ruslan-Aleev:
### What does it do?
Removed `allowNegative` parameter for TV with number type. `allowNegative` setting is superfluous, `minValue` setting is enough.
**Parameters may conflict, which may cause a non-obvious error.**

This is the current behavior:
1. If `Allow Negatives` is set to `No` and `Min Value` is `empty`:
- The input field doesn't accept the `-` character (_great_);
- If the value retrieved from the database is negative it *will be converted* to positive (so -10 becomes 10) without a warning/error (_to me, as a user, unexpected behavior_);
- The minimum value you can insert is `0`.

2. If `Allow Negatives` is set to `No` and `Min Value` is `> 0`:
- Same behavior as above except;
- Gives an error if the value is less then `Min Value`

3. If `Allow Negatives` is set to `Yes` and `Min Value` to `>= 0`:
- The input field accepts the `-` character (negative values can be inserted);
- If the value retrieved from the database is less then `Min Value` it's *not converted*;
- Gives an error if the value is less then `Min Value`. 

I agree with @Ruslan-Aleev, to me as a user,`Allow Negatives` is superfluous.

_Originally posted by @JoshuaLuckers in https://github.com/modxcms/revolution/pull/15042#issuecomment-633927489_

### Why is it needed?
UI / UX Improvement

### Related issue(s)/PR(s)


theboxer:
Would be nice to have an upgrade script when going from 2.x to 3.x for this. For cases where users used `allowNegative` option.

Ruslan-Aleev:
@theboxer Can you add this update script to this PR? I don’t understand how it should be.

theboxer:
@Ruslan-Aleev take a look here: https://github.com/modxcms/revolution/tree/2.x/setup/includes/upgrades you can find examples of upgrade scripts there

theboxer (approved):

smg6511:
I would respectfully argue that the setting is not without use. While, yes, you can rely on the minValue for validation purposes the allowNegative setting removes the end user's ability to type in an invalid dash when set to 'No.' Thus a negative value should not be able to be passed to the database anyway.

I'd say the bigger issue is the lack of effective validation of the input and output options for TVs. You're allowed in many instances to _save_ invalid values (for instance a minValue that's greater than a maxValue). I've been working up a pretty major proposed change for how that works, which would allow for more robust validation in that area.

Ibochkarev (approved):

Jako (approved):

Mark-H (approved):

PR #15120, opened 2020-06-19
Changed xtype for description in grids

Ruslan-Aleev:
### What does it do?
Changed xtype for description in **sources, widgets & providers grids**
For all grids, for example, contexts, settings, etc., the description basically has `xtype: textfield`, for sources, widgets & providers grids for some reason `xtype: textarea` - it looks strange.

A grid of sources, for example. Before:
sources_grid_1

After:
sources_grid_2

A grid of contexts, for example:
sources_grid_3

p.s. `xtype: textarea` is also set for the lexicons grid, but this is appropriate here.

### Why is it needed?
UI / UX Improvement

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/assets/modext/widgets/source/modx.panel.sources.js
+++ b/manager/assets/modext/widgets/source/modx.panel.sources.js
@@ -97,7 +97,7 @@ MODx.grid.Sources = function(config) {
             ,dataIndex: 'description'
             ,width: 300
             ,sortable: false
-            ,editor: { xtype: 'textarea' }
+            ,editor: { xtype: 'textfield' }
             ,renderer: Ext.util.Format.htmlEncode
         }]
         ,tbar: [{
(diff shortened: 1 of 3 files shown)


Jako:
Could a description contain multiple lines or not? The database type is text (or tinytext) there. So I would say that this change is not right.

Ruslan-Aleev:
Well, this is just the output of the description in the grid. It does not affect the description itself. It does not seem to me that someone wrote a lot of text in the description.
In general, it seems to me that `tinytext` is enough for description, `text` is superfluous.

Jako:
Please check, if the new line signs are removed with your change, if you edit the value in the grid.

Ruslan-Aleev:
> Please check, if the new line signs are removed with your change, if you edit the value in the grid.

No, everything is fine:

source_desc

theboxer:
I'm against this change. If the description is a `textarea` in the panel, it should be a `textarea` in the grid as well.

Ruslan-Aleev:
> I'm against this change. If the description is a textarea in the panel, it should be a textarea in the grid as well.

Yes, that would be right if other grids had this behavior.
However, this is not observed either in contexts, or in access policies, or in form customization, etc.

theboxer:
so imho it'd make sense to do it the other way around than you are proposing. Instead of changing textareas to textfields, I'd vote for changing textfields to textareas.

Issue #14966, opened 2020-01-23
Double load for TV media sources list


ant-tar:
### Summary
User should click twice on source selectore while changing media source for TV

### Step to reproduce
1. Create second media source (additionally to default one).
2. Open any TV and try to change appropriate media source via Media Sources tab 

### Observed behavior
Please check this:

![media_source_double_load](https://user-images.githubusercontent.com/2815640/72981617-435e8480-3dee-11ea-95c8-f560f12eb5cc.gif)

### Expected behavior
Better UX  - list should be loaded only once after click OR preloaded initially somehow. Single click should allow to open the list and select another source anyway. 

By the way, there are no problems in "System settings" with the choice of the sources.

### Environment
MODX 2.7.2

Ruslan-Aleev:
Duplicate https://github.com/modxcms/revolution/issues/14760
Although it’s written more competently here, I’ll close my.

Ruslan-Aleev:
If we delete this line https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/element/modx.panel.tv.js#L669, then there is no double loading, but the source name is not displayed (only its id):

![tv-sources](https://user-images.githubusercontent.com/12523676/81921558-7ae27e00-95e3-11ea-9f83-a17f9a973ba4.gif)

In principle, in the system settings and other elements in manager grids only id is displayed.
For the UI, hiding the name is a minus, but the lack of double loading is a huge plus :)

Can we leave id output @JoshuaLuckers @Jako what do you think?

Ruslan-Aleev:
If we delete this line https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/element/modx.panel.tv.js#L669, then there is no double loading, but the source name is not displayed (only its id):

![tv-sources](https://user-images.githubusercontent.com/12523676/81921558-7ae27e00-95e3-11ea-9f83-a17f9a973ba4.gif)

In the system settings and other elements in manager grids only id is displayed.
For the UI, hiding the name is a minus, but the lack of double loading is a huge plus :)

Can we leave id output @theboxer , what do you think?

p.s. Most likely, many will not notice the changes, but the double loading bug is extremely annoying.

Jako:
@Ruslan-Aleev It is better to detect the bug than disabling an option.

Ruslan-Aleev:
@Jako Thank you for helping with the bug!
[...]

PR #15122, opened 2020-06-19
Fix combocolumn grid refresh

Jako:
### What does it do?
Fix a second refresh of the grid when opening the combo box reloads the store.

### Why is it needed?
During editing a combo box in the grid the combobox currently has to be opened twice. This happens because the 'load' handler was missing the scope property and the options property was used as scope and the option single: true was not used.

### Related issue(s)/PR(s)
#14966


Ruslan-Aleev (approved):
Thank you so much!

theboxer (approved):

PR #15124, opened 2020-06-22
Changing old documentation links to current ones

Ibochkarev:
### What does it do?
Changing old documentation links to current ones

### Why is it needed?
For relevance of information

### Related issue(s)/PR(s)
NA


theboxer:
@Ibochkarev Is the `/current/` in the URL a correct way for doing this? I suppose after 3.x will go out as PL, the `current` in docs will reference the 3.x release and all these links will be invalid.

Ibochkarev:
> @Ibochkarev Is the `/current/` in the URL a correct way for doing this? I suppose after 3.x will go out as PL, the `current` in docs will reference the 3.x release and all these links will be invalid.

You can make similar changes for version 3.

theboxer:
No, my point is, that when 3.x will go PL, these links in 2.x won't be correct and it will require another PR to fix them

Mark-H:
Structure of 2.x and 3.x docs are the same, at least as far as these particular links are concerned, so docs wise I'd say current is fine.

Mark-H (approved):

PR #15126, opened 2020-06-23
Changed xtype for description in grids

Ruslan-Aleev:
### What does it do?
Changed xtype for description in grids.

A large text can be specified in the description (the field has a text or tinytext in the database), so that it is more convenient to edit it in the grid, I changed `xtype` to `textarea`.

For some grids, xtype is already set to textarea: sources, widgets & providers.
p.s. This is just the output of the description in the grid. It does not affect the description itself. 

A grid of contexts, for example:
Before:
sources_grid_3

After:
grid_desc_2

### Why is it needed?
UI / UX Improvement

### Related issue(s)/PR(s)
None

### DIFF
--- a/manager/assets/modext/widgets/fc/modx.grid.fcprofile.js
+++ b/manager/assets/modext/widgets/fc/modx.grid.fcprofile.js
@@ -78,7 +78,7 @@ MODx.grid.FCProfile = function(config) {
             ,dataIndex: 'description'
             ,width: 250
             ,sortable: true
-            ,editor: { xtype: 'textfield' }
+            ,editor: { xtype: 'textarea' }
         },{
             header: _('usergroups')
             ,dataIndex: 'usergroups'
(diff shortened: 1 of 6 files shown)


Ibochkarev (approved):

theboxer (approved):

JoshuaLuckers:
FYI: you can press shift+enter to "save"

JoshuaLuckers:
The descriptions in the grid are still displayed on 1 line. Would it be possible to show the newlines as well?

Ruslan-Aleev:
> Would it be possible to show the newlines as well?

I don’t even know, didn’t think about it :)

Ruslan-Aleev:
Newlines is shown during editing and inside the panel (in the "description" field).

xtypde_desc

But in the grid it will not work to show newlines, how will it look?

theboxer:
I believe this is a correct behaviour. Can't imagine seeing the multiline description in the grid :) Can be achieved with expandable columns if needed somewhere.

JoshuaLuckers:
I'm confused as to what the purpose of the textarea is in this case?

Ruslan-Aleev:
> I'm confused as to what the purpose of the textarea is in this case?

I did this for consistency + in the description field you can specify a large text (text or tinytext in database) and such a text would not be convenient to edit in textfield xtype.
p.s. In a lexicon grid where there is a lot of text, xtype textarea is most useful :)

PR #15134, opened 2020-06-25
Prevent path traversal when sending registry message

opengeek:
### What does it do?
Prevents a carefully crafted message key provided to the modFileRegister->send() method from writing files above the registry directory location if the message key contains path traversal syntax.

### Why is it needed?
modFileRegister will write a file above the registry directory when a message key is provided that includes path traversal syntax, i.e. `../`. This can result in PHP files written anywhere on a file system that is permissible by the PHP process. Though it is not likely this could be exploited for anything beyond annoyance, it still needs to be prevented.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/_build/test/Tests/Model/Registry/modFileRegisterTest.php
+++ b/_build/test/Tests/Model/Registry/modFileRegisterTest.php
@@ -124,6 +124,12 @@ class modFileRegisterTest extends MODxTestCase {
                 array('a' => 1, 'b' => 2.0, 'c' => 3.25, 'd' => 4.1390, 'e' => 5),
                 array()
             ),
+            array(
+                false,
+                '/topic5/',
+                array('../../invalidMessageKey' => 'invalid'),
+                array()
+            ),
         );
     }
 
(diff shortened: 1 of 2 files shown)


Mark-H (approved):

Jako (approved):

JoshuaLuckers (commented):
What do we do during read if such a carefully crafted message already exists? Do we allow it or do we prevent it from being read and clean it up?

theboxer (approved):

opengeek:
> What do we do during read if such a carefully crafted message already exists? Do we allow it or do we prevent it from being read and clean it up?

Reading the messages is not the issue; I see no problem allowing the reading of such a message. If someone was letting unfiltered user input be put into the message itself, which is an entirely different problem, and not one we can solve at the modRegister level, there could be major implications. But the problem here was simply that someone could cause the message file to be written just about anywhere in the path that was permissible by the PHP process, and this was being allowed unless specifically configured in a way to prevent it, by several extras packages.

Issue #15133, opened 2020-06-25
XSS in template name when rendered in TVs grid of template access 


Mark-H:
### Summary

The template name is not properly sanitised in the TVs "Template Access" grid 

### Step to reproduce

Follow the steps to reproduce #15132, then go to a template variable and notice it's executing the code in the template access grid.

### Observed behavior

Exploited XSS for users with the permission to edit TVs.

### Expected behavior

Safely handle user input, even for administrative users

### Environment

2.7.3-dev @ 3c7706e7ed97b3b188f0bd32bacbf0df4bdb14bd

Mark-H:
To note, this should not affect 3.x because that has globally applied a default renderer for grids.

theboxer:
@Mark-H that's not true in all cases. Change the name of Filesystem mediasource or create new one with XSS `Test"><img src="#" onerror="alert(1)" />`. Open existing TV/create new TV and you'll see the alert, if the alert won't show up, go to the media sources tab for that TV and use the inline editor to change the media source to the one with XSS. Right after you select the source, alert will show up. After save and reload it'll start popping as well.

Mark-H:
Any case in 3.x that isn't true should be addressed. ;)

smg6511:
This effect can be replicated in various places in 3.x alpha-3. I'm not sure why, though, there'd be any reason to allow tags of any kind in the non-content, non-textarea (name, description, etc.) fields of elements, sources, contexts, files, etc.—the exception being certain resource fields where limited formatting tags should be allowed. In those fields where it could be agreed upon that _no_ tags be allowed, it might make sense to:
1. Assess whether there's a compelling reason to allow the <> characters (or the = sign for that matter). If not, simply failing validation with these characters should do the trick.
2. If it's determined those characters need to be usable, then some pretty specific regex might be needed to invalidate input containing possible code injection (looking for onerror and other events that can invoke a script, perhaps taking into account the particular tags they can be attached to).
3. Make sure whatever validation is applied on the js side is applied on the php side of the equation.

For resource text (not textarea) inputs (pagetitle, longtitle, menutitle, attributes), only allow certain tags. Those tags could be made configurable via a system setting. You could extend that to establishing character limits via system setting too.

alroniks:
Is it still possible to create a Template name with such tags? I tried on 2.8.1 and failed, it prevents my attempts.
@Mark-H Could you clarify it a bit?

Mark-H:
[...]

PR #15137, opened 2020-06-26
[WIP] Prevent XSS on Template's name

### LINKED ISSUE #15132: XSS in edit template panel + quick update title
### Summary

The template name is not encoded in the heading "Template: <templatename>"

### Step to reproduce

1. Edit a template and set its name to `">`, refresh page to trigger the XSS in the heading
2. Attempt quick editing the template to trigger the XSS in the quick update title

### Observed behavior

Exploited XSS for users with the permission to edit templates.

### Expected behavior

Safely handle user input, even for administrative users

### Environment

2.7.3-dev @ 3c7706e7ed97b3b188f0bd32bacbf0df4bdb14bd

theboxer:
### What does it do?
HTML encodes title when showing in quick update window, or template panel.

### Why is it needed?
XSS

### Related issue(s)/PR(s)
Closes #15132

### DIFF
--- a/manager/assets/modext/widgets/element/modx.panel.template.js
+++ b/manager/assets/modext/widgets/element/modx.panel.template.js
@@ -315,7 +315,7 @@ Ext.extend(MODx.panel.Template,MODx.FormPanel,{
         if (this.initialized) { this.clearDirty(); return true; }
         this.getForm().setValues(this.config.record);
         if (!Ext.isEmpty(this.config.record.templatename)) {
-            var title = _('template')+': '+this.config.record.templatename;
+            var title = _('template') + ': ' + Ext.util.Format.htmlEncode(this.config.record.templatename);
             if (MODx.perm.tree_show_element_ids === 1) {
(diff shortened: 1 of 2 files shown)


theboxer:
So after looking at #15133 this is more problematic to handle. The name isn't the only issue, there's the same problem with description and with other objects like resource group, media source, etc.

Instead of using `Ext.util.Format.htmlEncode` when outputting the value as a primary fix, I'd suggest fixing this on the model side - not allowing to store these characters (same way modCategory handles it). And then using the `htmlEncode` for live updates of the name in template panel.

@Mark-H / @opengeek let me know if you agree with this approach or not.

Mark-H:
There's nothing inherintly wrong about having some special characters in the name of a template, so no, I don't agree with limiting it on the model.

This is indeed a bit of a recurring theme in extjs, but will need to be fixed in the places that uses it like you did here. Many similar ones have been sorted out already.

3.x also applies a catch-all default to encode grid values, but that can't be done in these types of cases, so this fix is correct.

Mark-H (approved):

Issue #15116, opened 2020-06-16
Make the tab active if there are errors in the form


Ruslan-Aleev:
### Summary
@Alroniks made extremely useful functionality that catches the tab with errors and makes it active - https://github.com/modxcms/revolution/pull/14235

**But, as it turned out, tab does not always become active.**

_For example, if the name TV contains invalid characters:_

![tab_error_1](https://user-images.githubusercontent.com/12523676/84759977-1c3c6580-afd0-11ea-9606-bf279acc76f9.gif)

_Although if the name TV is empty, the tab is activated, everything is fine:_

![tab_error_2](https://user-images.githubusercontent.com/12523676/84759985-1d6d9280-afd0-11ea-9290-d24b5b9f6285.gif)

### Expected behavior
Make the tab active if there are ANY errors in the form

### Environment
MODX 2.x >

Ruslan-Aleev:
Most likely related with https://github.com/modxcms/revolution/issues/15088

smg6511:
I have a solution for this and will be submitting a PR in the next few days. The reason this happens (in the other element form panels as well) is the failure method that ultimately fires the failureSubmit event is missing. The reason it works when the name field is blank is that the native extjs allowBlank validation is triggering the failure whereas invalid field value errors are reported in the submission failure response.

Ruslan-Aleev:
> I have a solution for this and will be submitting a PR in the next few days.

It would be great! If you have a desire, look at this issue - https://github.com/modxcms/revolution/issues/15088 they are related.

smg6511:
Almost finished on this issue and #15088 that you mentioned. Trying to cover as many scenarios as I can think of ... the last of which (for now) is handing the behavior when tvs are moved into custom panels via manager customization rules. I should have the PR very soon...

JoshuaLuckers:
> I should have the PR very soon...

You can always create a PR and mark it as draft! We really appreciate the effort!

Mark-H:
Fixed in 2.8

Issue #15088, opened 2020-06-04
[2.7.3] Required field not being filled causes js error, without showing user they need to fix a required field


Mark-H:
### Summary

A required field when editing a resource not being filled causes a javascript error, and prevents the "Please correct the errors in your form before submitting" error from being shown to the user.

The specific error is as follows:

```
modx.panel.resource.js:251 Uncaught TypeError: Cannot read property 'hidden' of undefined
    at MODx.panel.Resource.failureSubmit (modx.panel.resource.js:251)
    at h.Event.fire (ext-all.js:21)
    at MODx.panel.Resource.fireEvent (ext-all.js:21)
    at MODx.toolbar.ActionButtons.handleClick (modx.component.js:298)
    at S.onClick (ext-all.js:21)
    at HTMLSpanElement.I (ext-all.js:21)
```

Also discussed [on the forum](https://community.modx.com/t/save-button-for-resources-not-functioning/2654) and modmore ticket #22940. 

### Step to reproduce

I've not reduced the identified behavior to a very narrow use case, but I suspect this happens when you leave any TV that is marked as required (allowBlank = no) empty. In this particular case, it was a checkbox tv that required a value and had an invalid default value (default value not in the options) and therefore always invalid. 

It seems this may also require the TVs section at the bottom of the page instead of in the tab (`tvs_below_content` setting enabled), but I've not verified that yet. See below.

Trying to save will mark the field in red to indicate it's required, however if that's on a different tab or way at the bottom of the page, that's near impossible to spot. The lack of a notice is definitely troubling.

The problem seems to be in the `failureSubmit` method on `MODx.panel.Resource`. 

``` js
        // The next check needs the tabs index value to check if it is hidden or not
        if (tabs && tabs.items && tabs.items.keys) {
            tabs_index = tabs.items.keys.indexOf(tab_name);
        }

        // We are already on a tab that has an invalid state. No need to switch
        if (!tabs.items.items[tabs_index].hidden)  { // <--- breaks here on accessing .hidden on undefined
            return;
        }
```

It appears that `tabs_index` is -1 (i.e. the current tab, `tab_name = "modx-panel-resource-tv"`, does not exist in the tab panel). 

Because it breaks on what would normally be the tvs tab, I strongly suspect this only happens when tvs_below_content is enabled as it then no longer is a tab, but a separate panel at the bottom. 

### Possible fix

[...]

Issue #14743, opened 2019-09-19
Scroll screen to required TV


Ruslan-Aleev:
### Summary
If the TVs are located under the content and they are required, then saving the resource does not seem to work, because it is not visible that you need to fill in the TV field.

![tv-r](https://user-images.githubusercontent.com/12523676/65239895-078a7a00-daf1-11e9-9dd4-15293eafcfb9.gif)

p.s. Misunderstanding often happens with customers because of this.

### Suggested solution(s)
Move the screen to the required fields.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14235
https://github.com/modxcms/revolution/issues/11112

Mark-H:
Fixed in 2.8

PR #15146, opened 2020-07-02
Fix Various Issues with Resource/Element Validation and Field Display

smg6511:
#Made several changes to gain expected validation behavior when editing resources, templates, tvs, snippets, chunks, and plugins. Now, the first tab with an invalid field is activated when not currently visible. Also small fixes to validation regex rules as well as message display were made. Lastly, some inconsistencies with field description labelling is addressed (i.e., when inline help is specified in system settings, all element editing fields now show their descriptions below instead of in tooltips). This labelling inconsistency should IMO be fixed in resource editing pages as well, but is not addressed here. (Please forgive all the minute space diffs; most of them are not intentional.)

### What does it do?

- Fixed regex in modx schema and class maps for element names (end of rules incorrectly had a negative _lookahead_ for space rather than negative _lookbehind_).
- Removed failureSubmit and traverseNode methods in MODx.panel.Resource. The showErroredTab and detectErrors methods in the parent MODx.FormPanel class, with adjustments, serve the same purpose. The failureSubmit listener of each element panel subclass now calls the showErroredTab method to evaluate field errors.
- Added helper method in MODx.FormPanel to convert an array of component itemIds into their associated component ids. This to ensure a proper value to pass to getCmp when a component's id is created on the fly.
- Added failure listener to element panels. Without it, the failureSubmit _event_ never gets triggered.
- Fixed element create and update processors to handle the name field for templates correctly. Without this fix, it was possible to save templates with duplicate names.
- Some general streamlining of defaults in element panel configs.
- The first-found (and now navigated to) field error, when being returned from the server side, would disappear as soon as the alert message appeared. This was due to the default behavior of native (Ext) validation triggering on blur. If the field in question is valid as far as the front end config rules go, its invalid state gets cleared once focus is lost even though the back end validation may still not be passing. Thus the default validation event was set to 'change' instead.

### Why is it needed?

To ensure better evaluation and display of, as well as dynamic navigation to, found field errors. At present, it can be confusing when encountering an error that is not in the currently active tab.

### Related issue(s)/PR(s)

Addresses issues #14743 , #15116 ,  and #15088


manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -105,7 +121,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                 this.rteLoaded = false;
             }
         }
-
+        // console.log("MODx.panel.Resource::setup, isDirty?", this.isDirty());
```
Ruslan-Aleev: This line should be deleted :)

smg6511: Thought I'd tracked down all those logging lines! Fixed.

Ruslan-Aleev (approved):
**Great PR, thanks a lot!**
Solves a lot of problems in UX :)

The active sub-category of the TV tabs opens, and the screen "scrolls" to the error (when the TV position is below the content or in a separate tab):
tv_req_1

Messages are displayed and the tab is activated:
tv_req_2

Ruslan-Aleev:
Guys, please test this PR, it significantly improves the current behavior for validation in forms.
It would be great to have this in the 2.8 branch :)

Mark-H:
It looks our current js build workflow fails when using `let` and `const` constructs (in the uglifyjs bit) so I'm reverting those back to plain ol' `var`. Probably a good idea to sort out that workflow at some point, but for now I'm doing it this way to get your improvements into 2.8. Thanks @smg6511 for the great work.

Mark-H (approved):

Ruslan-Aleev:
@Mark-H what about the 3.x branch? The 3.x branch also contains these bugs.

Mark-H:
I'll try to cherry pick this (and the other things I merged the other day) into 3.x later. Couldn't get that to do what I wanted last sunday around midnight so figured that was a good one for another day. ;)

smg6511:
Mark - If you want I can integrate the changes into a new PR for 3.x. Also, regarding your comment on the workflow you may want to take a look at Terser in place of Uglifyjs. I think ES6 has been out long enough and support is broad enough for us to be integrating its features into our coding.

rthrash:

https://community.modx.com/t/proposal-to-move-element-code-into-its-own-tab/4562/1

Issue #15148, opened 2020-07-04
Simplify the “User Groups” section: display groups as a grid, not a tree


Ruslan-Aleev:
### Summary
In the current version, it’s almost impossible to understand where to edit a specific “User Group”, because the editing link is hidden in the context menu.
And when you click on a group, it’s not the group’s settings appearing (what you expect), but adding a user to the group.

![users_groups_1](https://user-images.githubusercontent.com/12523676/86515911-3cb84c80-be25-11ea-9237-5ef740395169.gif)

![users_groups_2](https://user-images.githubusercontent.com/12523676/86515912-3d50e300-be25-11ea-9bd1-4801618befa8.gif)

**I suggest making the display of "User Group" in the form of a grid instead of a tree.** And the user can be added to the group inside the group editing (this already works, see gif above).
This will be much clearer, especially with the new link on name and gear context menu.

It will look something like this:

![users_groups_3](https://user-images.githubusercontent.com/12523676/86516279-e4cf1500-be27-11ea-8714-7fd46e7d31bc.png)

It will be possible to display useful information in the "User Group" grid, such as users in a group, access to sections (contexts, groups, etc.)

### Why is it needed?
Simplification in UI / UX

### Related issue(s)/PR(s)
N/A

sergant210:
In addition
1. Rename the "User groups & users" tab to "User groups".
2. Simplify the copy logic of Access Policies .

SnowCreative:
I agree that it's confusing that way it's set up now.

sebastian-marinescu:
A "Update User group" or "Update group settings" button above the already loaded grid would also make it much clearer.

Ruslan-Aleev:
> A "Update User group" or "Update group settings" button above the already loaded grid would also make it much clearer.

I do not quite understand.
Now all actions are located in the gear menu on the right, so an additional button is not needed.

sebastian-marinescu:
Your suggestion and UI proposal look very good, assuming someone would provide a PR for that. 
I'm just saying that it might also be possible to add another button, like this: 

![image](https://user-images.githubusercontent.com/3798871/87715105-21e2c200-c7ad-11ea-923b-dfd6bc3ac5a0.png)

Ruslan-Aleev:
Now it's clearer :) Yes, we can do this, and it's easier.
But personally, I am in favor of rewriting the section completely.
p.s. Probably for branch 2.8 the button solution would be appropriate.

[...]

PR #15149, opened 2020-07-04
Reordered the navigation tabs in the "User Group"

Ruslan-Aleev:
### What does it do?
1) Reordered the navigation tabs in the "User Group".

Made them more logical:
- You first create a group
- Then configure permissions for group
- Then add users to the group, if it need
- Other settings

**After:**
users_groups_tab_2

Before:
users_groups_tab_1

2) Changed lexicons to more specific

users_groups_lexicons

### Why is it needed?
UI/UX fix


JoshuaLuckers (approved):

rthrash:
This is a really nice improvement and _much_ more logical. Thanks! For those looking to find/reproduce this, it's done by right-clicking on an existing user group in the ACL list: Settings (cog) at the lower left > Access Control Lists > right-click Administrator (or other ACL group name). 

I would suggest, though, we rename the Permissions top tab to Access Permissions. The vertical tabs wrap awkwardly on that tab, and "Access" is redundant here (and inconsistently capitalized). As a general rule, anywhere "Here you can…" is encountered, as in the screenshot below, it can be written more directly/clearly, usually by simply eliminating "here you can": 

> Set the Contexts this User Group can access. 

User Group: Administrator | MODX Revolution 2020-07-23 22-49-41


While the following potentially expands the scope of this PR from its original intention … I also wonder if the parent menu could use some UX love itself. There's a lot of extra bling here that distracts from the UI, and a lot of hidden functionality. 

Access Control Lists | MODX Revolution 2020-07-23 22-38-30

First, suggested new title/description so people will know what to expect and how to use this section without having to visit the documentation:

> Click the group names to manage membership and the gear-icon to update its settings. (User counts in parenthesis. Anonymous means any visitor without a login.)

If we adopt this suggested lexicon change, the "(anonymous)" group should have the parenthesis removed from around it. If anything, it should have an infinity symbol or asterisk in parenthesis beside it, and possibly re-labeled "Unauthenticated". Since "Administrator" is capitalized, I vote for also capitalizing "Anonymous" for consistency.

> Anonymous (*)

Next, the drop-down triangle in the header doesn't function; the up/down arrows above it that do are a strange deviation from most conventions. But … why would you want to collapse a single list in the first place? 

The title above the groups is redundant with the tab itself. 

I wonder what state would ever require using the refresh tree icon, though it appears to work as advertised. I vote for eliminating it. People can ctrl-r to refresh if needed.

Losing five icons/buttons/action items/distractions, and clarifying the language should make working in this complex part of Revo a little bit simple.

With losing all the redundant header stuff, the "new user group" button could be moved to the description row/top-right.

Finally—and most importantly IMO—each User Group row should have a clearly visible edit icon. Otherwise, it's hard to figure out how to edit existing user groups because it's hidden behind a right-click _which doesn't exist on mobiles/tablets_. And while not shown on the markup screenshot below, do we need to collapse the sidebar also? I vote to eliminate its collapsibility.

Access Control Lists | MODX Revolution 2020-07-23 22-43-55

Ruslan-Aleev:
@rthrash Thanks for the detailed comment. You are right, for "User Groups" I will correct the tab names and descriptions later.

As for renaming `(anonymous)` to something else, then the question arises: How to call it more clearly? Maybe `(unauthorized)`? The name `Unauthenticated` is not very clear to me. Also, I would leave parentheses and lowercase letters (to draw attention), since the MODX manager has similar entities that are displayed in a similar way, for example, an empty resource template - `(empty)`.

As for the output of groups in the form of a tree, we already had a discussion and even some fixes, see https://github.com/modxcms/revolution/issues/15148 and https://github.com/modxcms/revolution/pull/15159 :)

opengeek:
I've attempted to rebase these changes on the latest and there are too many other changes to this js for me to work out easily. If you have time, can you take a look @Ruslan-Aleev and push it if you can work it out?

Ruslan-Aleev:
Yes, I'll deal with it later. But here the strange behavior of github, I just move the tab in the file, at first everything is fine, but after a while github sees these changes as conflicting.
I have already redone twice, why the changes are marked as conflicts is not clear for me.

Ruslan-Aleev:
@opengeek Done

rthrash (commented):
Adding a second review based on my comments here: https://github.com/modxcms/revolution/pull/15149#issuecomment-663341168 (and ultimately for a how to do a review article) … All in all I think it's a net positive and can go.

PR #15151, opened 2020-07-06
Change the https rule to a variable one

Jako:
### What does it do?
Change the example.com based https rule to a variable based rule

### Why is it needed?
To make enabling a bit more comfortable

### Related issue(s)/PR(s)
none

### DIFF
--- a/ht.access
+++ b/ht.access
@@ -35,12 +35,9 @@ RewriteRule "/\.|^\.(?!well-known/)" - [F]
 #RewriteRule ^(.*)$ https://www.%1/$1 [R=301,L] .
 
 
-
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
Maybe it makes sense to move the working directives higher in file? For example - https://github.com/modxcms/revolution/pull/15151/files#diff-531da0d53410c851304b5c842ccd2dadL54
Otherwise, we may not notice them under a large list of not working directives (#).

opengeek:
@Jako can you please push these PR branches to your fork instead of the upstream repository?

Jako:
@opengeek I could recreate them on my fork after my vacation next week.

PR #15152, opened 2020-07-06
Changed the output template for menu items

Ruslan-Aleev:
### What does it do?
I deleted the descriptions for 2 menu items: `media_desc` and `about_desc`:
- because for all the main items except the `site` and `about` there are no descriptions in the lexicons
- now the `topmenu_show_descriptions` setting works for main menu items and the user can add a description, if necessary

In the current version of the menu, the icons for some menu items are set not through the `icon` field, such as for the "User" or "About" items, _but through styles, which is not correct_.

Also, if you move the menu item to another menu area, then the styles break.

menu_icons_1

This PR fixes this.

**Also added icons for submenu items.**
Often the components menu has icons, I think it will be useful to display it.

submenu_icons

### Why is it needed?
Uniformity in the menu, avoiding possible bugs in the future

### Related issue(s)/PR(s)
N/A


JoshuaLuckers:
You are right, having the icons hardcoded in the style is wrong but to me it makes sense to use styles for the icons. The icon entry should work the same as the icon for templates (icon value is used as class name). 

It's odd to have HTML markup stored in the menu entry to display an icon. 
So I propose to store values like `icon-cube icon icon-large` instead of `<i class="icon-file-text-o icon icon-large"></i>`.

Ruslan-Aleev:
> So I propose to store values like `icon-cube icon icon-large` instead of `<i class="icon-file-text-o icon icon-large"></i>`.

I thought about this, but there are drawbacks to your variant:
1) Defining just a class is less clear than `<i class="..."></i>`
2) In the field, not only the icon can be set, for example, https://github.com/modxcms/revolution/pull/15152/files#diff-6062460821700417bc6dd9d70456715fR16
3) It is also necessary to take into account that in packages icons will be set via `<i class="..."></i>`.

It probably makes sense to do this, but in a different PR.

_build/data/transport.core.menus.php
```diff
@@ -122,10 +123,11 @@
 $topNavMenus[1]->fromArray([
   'menuindex' => 1,
   'text' => 'media',
-  'description' => 'media_desc',
+  'description' => '',
```
JoshuaLuckers, line 126: Please keep the description.

core/lexicon/en/topmenu.inc.php
```diff
@@ -7,7 +7,6 @@
  * @subpackage lexicon
  */
 $_lang['about'] = 'About';
-$_lang['about_desc'] = 'Learn more and get help with MODX';
```
JoshuaLuckers, line 10: Please keep the lexicons.

core/lexicon/en/topmenu.inc.php
```diff
@@ -60,7 +59,6 @@
 $_lang['logout_desc'] = 'Log out of the Manager';
 $_lang['manage'] = 'Manage';
 $_lang['media'] = 'Media';
-$_lang['media_desc'] = 'Update Media and Media Sources';
```
JoshuaLuckers: Please keep the lexicons.

JoshuaLuckers (commented):
Please don't remove the description from the menu entries. We need to hide the descriptions when they are the topmost menu entry. If we, for example, move the menu entry to be a child of "Content" the description is still being used and displayed in the dropdown.

Correct me if I'm wrong.

Ruslan-Aleev:
@JoshuaLuckers I adjusted the styles for the description, they now work correctly.
For the main menu items, I deleted the descriptions because for all the main items except the `site` and `about` there are no descriptions in the lexicons, and it is displayed in menu only for `site`.
Why did I decide not to hide the descriptions through the styles? - There is a setting `topmenu_show_descriptions` and descriptions are shown for all but not for the main ones, which can confuse users.

Those, the descriptions work, and they can be displayed if necessary.
And for the main items, I deleted them, because they were lexicons only for 2 items: `site` and `about` .

Ruslan-Aleev:
Maybe you’re right, and we need to hide the descriptions for the main items through the styles, but I still don’t see the point of saving the descriptions of `site_desc` and `about_desc`.

JoshuaLuckers:
> Maybe you’re right, and we need to hide the descriptions for the main items through the styles, but I still don’t see the point of saving the descriptions of `site_desc` and `about_desc`.

They are useful when the menu item is moved to a location where the description is visible. They can also be useful for other purposes like screenreader (even if they are hidden).

Maybe we can display the descriptions if we "hover" over the menu item (title attribute)?

Ruslan-Aleev:
I understand you, but then it’s worth adding descriptions for all main menu items. **I think, is for another PR**.
I am not against the descriptions, I repaired their styles, and they can be displayed (or with title attribute). But only for 2 main menu items of 7 have a description. Why leave a description for 2, but ignore 5?

JoshuaLuckers:
It's better to focus a PR on one thing. 
The PR is to fix the icons, not the removal or addition of descriptions. 
Otherwise we could have merged this already 😎. 

I didn't notice the others that are missing so nice catch!

Ruslan-Aleev:
I am for uniformity and logic. I added styles for the description, and adjusted the output template - now the description for the main items is displayed depending on the `topmenu_show_descriptions` system setting. And removed description for 2 menu items, because for 5 there were none. Those the setting works, and the user can add a description, if necessary - this is logical.

You suggest leaving a description for 2 menu items, but then I have to remove the description output from the menu template, leaving only the output in the title attr. As a result, the description disappears, regardless of the settings of `topmenu_show_descriptions` and title is shown not for all menu items - which is not logical at all.

p.s. I changed the name PR to make it clearer.
I had to edit not only the icons, because elements are located side by side (label, description).
I also don’t think that anyone (except us) will notice these changes :)

JoshuaLuckers (approved):

Ibochkarev (approved):
I tested it. Works!

PR #15153, opened 2020-07-07
Fix of lack of translation for anonymous

sergant210:
### What does it do?
Fix of lack of translation for the anonymous user group in the context interface.

### Why is it needed?
Anonimous user group in the context interface is not translated. And this group is different from the group in User Groups list.

### Related issue(s)/PR(s)
No.


JoshuaLuckers:
To me it would make more sense to target 3.x instead of 2.x

JoshuaLuckers (approved):

theboxer (approved):

theboxer:
Should go to 3.x as well

theboxer:
@cla-bot check

Issue #14307, opened 2019-01-24
[enhancement] New default media source


Ruslan-Aleev:
### Summary
In MODX, the FIlesystem source is set by default, but any user can see the system files in the file viewer, which is not correct (if user access is not set).
And also, if the sources for TV are incorrectly set, then all uploaded files will go to the root folder, cluttering up the root, and here you can, by mistake, delete the system files with the uploaded ones.

**I suggest creating a second media source and linking it to the /assets/ folder (may be /assets/uploads/) and set it as default**. This folder /assets/ already in MODX and it is intended for non-system files. For example, it can also upload static files to the new source.

Ibochkarev:
@digitalpenguin I ask you to Express your point of view on how to better organize and set up a new media source. What more opinions.

muzzwood:
Considering that MODX is installed with a single admin/sudo user, I would suggest setting up an assets media source like @Ruslan-Aleev mentioned but make it the default media source rather than the filesystem root.
At the moment IIRC new user groups automatically have access to the file system root unless manually configured otherwise. Perhaps by default new user groups could have access to the Assets media source but not the filesystem root. 
( I could be wrong about that, I haven't had time to double check).

PR #15154, opened 2020-07-07
Created a second media source to /assets/ folder and set it by default

Ruslan-Aleev:
### What does it do?
Created a second media source and linked it to the `/assets/` folder and set it by default.
This is relevant for the new MODX3 installation.

### Why is it needed?
In MODX, the FIlesystem source is set by default, but any user can see the system files in the file viewer, which is not correct (if user access is not set).
And also, if the sources for TV are incorrectly set, then all uploaded files will go to the root folder, cluttering up the root, and here you can, by mistake, delete the system files with the uploaded ones.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14307

### DIFF
--- a/_build/data/transport.core.media_sources.php
+++ b/_build/data/transport.core.media_sources.php
@@ -14,3 +14,28 @@ $collection[1]->fromArray([
   'class_key' => 'MODX\Revolution\Sources\modFileMediaSource',
   'properties' => [],
 ], '', true, true);
+$collection[2]= $xpdo->newObject(modMediaSource::class);
+$collection[2]->fromArray([
+  'id' => 2,
+  'name' => 'Assets',
+  'description' => '',
+  'class_key' => 'MODX\Revolution\Sources\modFileMediaSource',
+  'properties' => [
+    'basePath' => [
+        'name' => 'basePath',
+        'desc' => 'prop_file.basePath_desc',
+        'type' => 'textfield',
+        'options' => '',
+        'value' => '[[++assets_url]]',
+        'lexicon' => 'core:source',
+    ],
+    'baseUrl' => [
+        'name' => 'baseUrl',
+        'desc' => 'prop_file.baseUrl_desc',
+        'type' => 'textfield',
+        'options' => '',
+        'value' => '[[++assets_url]]',
+        'lexicon' => 'core:source',
+    ],
+  ],
+], '', true, true);
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -418,7 +418,7 @@ $settings['default_duplicate_publish_option']->fromArray([
 $settings['default_media_source']= $xpdo->newObject(modSystemSetting::class);
 $settings['default_media_source']->fromArray([
   'key' => 'default_media_source',
-  'value' => 1,
+  'value' => 2,
   'xtype' => 'modx-combo-source',
   'namespace' => 'core',
   'area' => 'manager',


JoshuaLuckers:
Nice! I'm not sure if we should set it as default. E.g: The `default_media_source` is also used as default Media Source when creating new static elements (template, chunk, snippet, plugin).

Ruslan-Aleev:
If we do not set it by default, then all the meaning is lost :)
Maybe for static elements we need to create a separate setting?

JoshuaLuckers:
Such a setting already exists: `static_elements_default_mediasource` but you need to enable automated static elements for it (`static_elements_automate_*`).

JoshuaLuckers:
For this Media Source the values for `basePath` and `baseUrl` should be without any preceding slashes: `assets/` instead of `/assets/`.

If you use `[[++assets_url]]` it does have a preceding slash.

Ruslan-Aleev:
Yes you are right. But I tested with `[[++assets_url]]` - visually everything works.
You cannot manually specify the `assets/` folder, because it can change during the installation process.
Perhaps there is another solution - I did not find it.

JoshuaLuckers:
In the media browser it will cause a double slash to appear.
Screen Shot 2020-07-12 at 12 53 47

I'm not sure if it will cause issues when MODX is installed in a subdirectory, that's something I'll look into another time :)

Ruslan-Aleev:
Alternatively, we can create a new setting `[[++assets_source_url]` in `config.inc.php` during the installation process:
```
if (!defined('MODX_ASSETS_PATH')) {
    $modx_assets_path= '/home/modx3/public_html/assets/';
    $modx_assets_url= '/assets/';
    $modx_assets_source_url= 'assets/';
    define('MODX_ASSETS_PATH', $modx_assets_path);
    define('MODX_ASSETS_URL', $modx_assets_url);
    define('MODX_ASSETS_SOURCE_URL', $modx_assets_source_url);
}
```
But I'm not sure about this...

Ruslan-Aleev:
We can specify the value like this:
```
[[++assets_url:stripString=`/`]]/
```
then there will be no problems with extra `/` and visually everything works. But it looks strange :)

Although I'm not sure if it is possible to set a nested category in assets_url, like: `parent/assets`

Ibochkarev (approved):
LGTM 👍 

https://user-images.githubusercontent.com/2138260/111724815-20afea80-8890-11eb-806b-4169ff609c21.mov

Ruslan-Aleev:
@alroniks Any ideas? And will there be a problem with paths with a slash at the beginning?

Ruslan-Aleev:
@JoshuaLuckers Any thoughts on this PR?

JoshuaLuckers:
> We can specify the value like this:
> 
> ```
> [[++assets_url:stripString=`/`]]/
> ```
> 
> then there will be no problems with extra `/` and visually everything works. But it looks strange :)
> 
> Although I'm not sure if it is possible to set a nested category in assets_url, like: `parent/assets`

stripString will remove all occurrences in the string, if we want to use the value from `assets_url` we need to use [ltrim](https://www.php.net/manual/en/function.ltrim.php) and specify `/` as character we want to trim. 


I think the development required to get this second Media Source benefit the larger amount of users doesn't outweigh the effort it takes for users to create a second Media Source manually.

Ruslan-Aleev:
> I think the development required to get this second Media Source benefit the larger amount of users doesn't outweigh the effort it takes for users to create a second Media Source manually.

I agree, but in my opinion, initially there is no problem that you indicated earlier https://github.com/modxcms/revolution/pull/15154#issuecomment-655543687 And you were going to look in the future - https://github.com/modxcms/revolution/pull/15154#issuecomment-657207268.
I mean, maybe we don't need any effort and it already works as it is? =)

JoshuaLuckers:
I've looked into it and the paths we define as value for basePath and baseUrl should be relative paths (so without the preceding slash).

PR #15155, opened 2020-07-08
Changed "remove" to "delete", "update" to "edit" in lexicons

Ruslan-Aleev:
### What does it do?
Changed "remove" to "delete", "update" to "edit" in lexicons.  PR only changes lexicons entries.

p.s. In the future, it is necessary to simplify the lexicons entries (key: value) and create a standard for writing them.

Often there are duplicates:
https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/snippet.inc.php#L20
https://github.com/modxcms/revolution/blob/3.x/core/lexicon/en/snippet.inc.php#L28

### Why is it needed?

### Related issue(s)/PR(s)


JoshuaLuckers:
This made me wonder: what's the difference between "remove" and "delete"?

> Delete and remove are defined quite similarly, but the main difference between them is that delete means erase (i.e. rendered nonexistent or nonrecoverable), while remove connotes take away and set aside (but kept in existence).

Source: https://english.stackexchange.com/a/52509

Ruslan-Aleev:
In MODX, with the exception of Resources, everything is permanently deleted, it seems to me. Therefore, the term "Delete" is more applicable (although it is better to use it for resources too, so that there is no confusion).

core/lexicon/en/default.inc.php
```diff
@@ -138,10 +138,10 @@
...
 $_lang['elements'] = 'Elements';
 $_lang['email'] = 'Email';
-$_lang['empty_recycle_bin'] = 'Remove Deleted Resources';
-$_lang['empty_recycle_bin_confirm'] = 'Are you sure you want to completely and permanently remove all deleted Resources? This is irreversible.';
-$_lang['empty_recycle_bin_empty'] = 'There are no deleted Resources to remove.';
-$_lang['empty_recycle_bin_emptied'] = 'All deleted Resources have been permanently removed.';
+$_lang['empty_recycle_bin'] = 'Delete Deleted Resources';
+$_lang['empty_recycle_bin_confirm'] = 'Are you sure you want to completely and permanently delete all deleted Resources? This is irreversible.';
+$_lang['empty_recycle_bin_empty'] = 'There are no deleted Resources to delete.';
+$_lang['empty_recycle_bin_emptied'] = 'All deleted Resources have been permanently deleted.';
```
JoshuaLuckers: English isn't my first language but I would use "Erase" instead of "Delete" when talking about deleting/removing `Deleted Resources`. 
A native English speaker's view on this would be great!

JoshuaLuckers (approved):

Ibochkarev (approved):

alroniks (approved):

PR #15158, opened 2020-07-16
Made quick search more convenient by adding system settings

Ruslan-Aleev:
### What does it do?
Made quick search more convenient by adding system settings:
- Corrected search styles.
- Fixed display of name and description. Did not work for Templates (name) and TV (description).
- **Added new system settings - result limit and search in content.**

### Related issue(s)/PR(s)
N/A


core/src/Revolution/Processors/Search/Search.php
```diff
@@ -37,9 +37,6 @@ class Search extends Processor
     const TYPE_USER = 'user';
     const TYPE_RESOURCE = 'resource';
 
-    /** @deprecated todo: move hardcoded value to settings */
-    public $maxResults = 5;
-
```
JoshuaLuckers, line 42: 🎉

setup/includes/upgrades/common/3.0.0-create-system-settings.php
```diff
@@ -0,0 +1,44 @@
...
+use MODX\Revolution\modSystemSetting;
+
+$keys = ['fast_search_in_content','fast_search_result_max'];
+
+foreach ($keys as $key) {
+
+    /** @var modSystemSetting $setting */
+    $setting = $modx->getObject(modSystemSetting::class, ['key' => $key]);
+
+    if (!$setting) {
```
alroniks: I would invert this condition to avoid nesting level of the code.
```php
if ($setting !== null) {
  continue;
}

... other code
```

setup/includes/upgrades/common/3.0.0-create-system-settings.php
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Common upgrade script for System Setting
+ *
+ * @var modX $modx
+ * @package setup
+ */
+
+use MODX\Revolution\modSystemSetting;
+
+$keys = ['fast_search_in_content','fast_search_result_max'];
```
alroniks:
```suggestion
$keys = [
    'fast_search_in_content',
    'fast_search_result_max'
];
```

Because options are pretty long, easier to read them when they in the column.

core/src/Revolution/Processors/Search/Search.php
```diff
@@ -149,15 +157,20 @@ protected function searchResources()
...
     {
         $c = $this->modx->newQuery($class);
-        $c->where([
+        $querySearch = [
             $nameField . ':LIKE' => '%' . $this->query . '%',
             'OR:' . $descriptionField . ':LIKE' => '%' . $this->query .'%',
-            'OR:id:=' => $this->query,
-        ]);
+        ];
+        if ($this->searchInContent()) {
```
alroniks: To be safer, need to check that `$contentField` is not empty too. Else it will trigger an error on database level.

core/src/Revolution/Processors/Search/Search.php
```diff
@@ -58,7 +55,15 @@ public function checkPermissions()
      */
     protected function getMaxResults()
     {
-        return $this->maxResults;
+        return intval($this->modx->getOption('fast_search_result_max', null, 10));
```
alroniks: Typecast via `(int)` works faster than function call. So, better to adjust it.

```suggestion
        return (int)$this->modx->getOption('fast_search_result_max', null, 10);
```

core/lexicon/en/setting.inc.php
```diff
@@ -266,6 +266,12 @@
 $_lang['setting_failed_login_attempts'] = 'Failed Login Attempts';
 $_lang['setting_failed_login_attempts_desc'] = 'The number of failed login attempts a User is allowed before becoming \'blocked\'.';
 
+$_lang['setting_fast_search_in_content'] = 'Allow search in content';
+$_lang['setting_fast_search_in_content_desc'] = 'If \'Yes\', then the content of the element (resource, template, chunk, etc.) will also be available for fast search.';
```
alroniks: Could we use emphasizing via bolding the text instead of quotes? To get "If **Yes**, then ..."

alroniks (changes requested):
Looks good, but I would adjust it a bit.

setup/includes/upgrades/common/3.0.0-create-system-settings.php
JoshuaLuckers: Do we even need a upgrade script for this? Are new settings not created automatically?

alroniks: Good question. I know that they can not be updated, but new ones should be created automatically. Need to check.

core/lexicon/en/setting.inc.php
Ruslan-Aleev: I think this is for another PR, because in lexicons, `Yes / No` is indicated everywhere through quotation marks.

Ruslan-Aleev:
I just thought that `fast_search` is not a very correct name, I think `quick_search` would be more competent. What do you think?

core/lexicon/en/setting.inc.php
alroniks: Yes, fixing all such texts is the other task, but if we can do it easier now, we shouldn't keep the technical debt in any case.

alroniks:
> I just thought that `fast_search` is not a very correct name, I think `quick_search` would be more competent. What do you think?

Yes, 'quick' more suitable, because it mostly about time than the ability to move fast.

core/lexicon/en/setting.inc.php
Ruslan-Aleev: I do not agree here. Because if the task (changing the quotes to bold) will take months, or will be forgotten, then we will get an inconsistency in the interface. Which is very likely...

alroniks: In this case, I agree with creating issue about it, just to not forget about it.
I don't think that changing the quotes can take months, it is mostly about the agreement, but technically it is easy.

setup/includes/upgrades/common/3.0.0-create-system-settings.php
Ruslan-Aleev: > Good question. I know that they can not be updated, but new ones should be created automatically. Need to check.

Any information on this? I thought that the update script is always needed, although I doubt it now :)

JoshuaLuckers: Existing system settings are not updated (`xPDOTransport::UPDATE_OBJECT => false`), new system settings are automatically created if they are defined in `_build/data/transport.core.system_settings.php`.
https://github.com/modxcms/revolution/blob/803ebb4303af0a925c8cdde53a25340f53904dc8/_build/transport.core.php#L241-L253

In this instance there is no need for an upgrade script.

Jako (approved):

Ibochkarev (approved):
I tested it. Works!

image

alroniks (approved):

PR #15159, opened 2020-07-17
Improve Navigation for Access Control Lists > User Groups & Users

Ruslan-Aleev:
### What does it do?
Added a button "Update User Group" for "ACL" section.
As suggested in https://github.com/modxcms/revolution/issues/15148#issuecomment-659629172 added a button, this option is easier.
After discussion and merging into the 3.0 branch I will add for the 2.0 branch as well.

And in the future, it is worth completely rewriting the section in the form of a grid.

Before:
user_groups_1

After:
users_groups_2

### Why is it needed?
In the current version, it’s almost impossible to understand where to edit a specific “User Group”, because the editing link is hidden in the context menu.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15148

### DIFF
--- a/core/docs/changelog.txt
+++ b/core/docs/changelog.txt
@@ -2,6 +2,7 @@
 This file shows the changes in recent releases of MODX. The most current release is usually the
 development release, and is only shown to give an idea of what's currently in the pipeline.
 
+- Improve Navigation for Access Control Lists > User Groups & Users [#15159]
 - Various quick search improvements [#15158]
 - Improve layout of additional options for TV List Box (Single Select) type [#15150]
 - Remove allowNegative parameter for TV with number type [#15119]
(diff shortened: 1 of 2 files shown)


alroniks:
It is a good improvement, indeed. But two primary buttons brake the interface drastically. So, I would leave one button as a primary and make the other button more neutral.

Ruslan-Aleev:
I personally like it.
Because firstly, this is the section "User Groups" and "Users" (this is the name of the tab), i.e. two buttons are appropriate here.
Secondly, there were initially more than one primary buttons (the "New User Group" & "Add User to Group" buttons).

Ruslan-Aleev:
And if we make only one primary button, then which one? This is the section "User Groups", but at the same time the list of "Users". Dilemma :)

JoshuaLuckers:
Wouldn't it make more sense to have it display a + icon like how it's done in the elements tree?

Ruslan-Aleev:
If you are talking about "Creating a group", then it will be less clear, in my opinion, although closer to the design.
If we make it an icon, then the rest of the trees will need to be corrected. But this task for a separate PR.

Ruslan-Aleev:
Any comments? In my opinion, this PR can be approved.

Ibochkarev (approved):
I tested it. Works!

image

Mark-H (approved):

PR #15160, opened 2020-07-17
Register woff2 format in system settings

Ibochkarev:
### What does it do?
Registers the woff2 format in the default settings to allow downloads.

### Why is it needed?
Registration of woff2 format in the Types of upload files setting upload_files for working with formats out of the "box"

### Related issue(s)/PR(s)
NA

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1904,7 +1904,7 @@ $settings['unauthorized_page']->fromArray(array (
 $settings['upload_files']= $xpdo->newObject('modSystemSetting');
 $settings['upload_files']->fromArray(array (
   'key' => 'upload_files',
-  'value' => 'txt,html,htm,xml,js,css,zip,gz,rar,z,tgz,tar,mp3,mp4,aac,wav,au,wmv,avi,mpg,mpeg,pdf,doc,docx,xls,xlsx,ppt,pptx,jpg,jpeg,png,tiff,svg,svgz,gif,psd,ico,bmp,odt,ods,odp,odb,odg,odf,md,ttf,woff,eot,scss,less,css.map,webp,woff2',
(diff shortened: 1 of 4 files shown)


_build/data/transport.core.system_settings.php
```diff
@@ -1904,7 +1904,7 @@
 $settings['upload_files']= $xpdo->newObject('modSystemSetting');
 $settings['upload_files']->fromArray(array (
   'key' => 'upload_files',
-  'value' => 'txt,html,htm,xml,js,css,zip,gz,rar,z,tgz,tar,mp3,mp4,aac,wav,au,wmv,avi,mpg,mpeg,pdf,doc,docx,xls,xlsx,ppt,pptx,jpg,jpeg,png,tiff,svg,svgz,gif,psd,ico,bmp,odt,ods,odp,odb,odg,odf,md,ttf,woff,eot,scss,less,css.map,webp',
+  'value' => 'txt,html,htm,xml,js,css,zip,gz,rar,z,tgz,tar,mp3,mp4,aac,wav,au,wmv,avi,mpg,mpeg,pdf,doc,docx,xls,xlsx,ppt,pptx,jpg,jpeg,png,tiff,svg,svgz,gif,psd,ico,bmp,odt,ods,odp,odb,odg,odf,md,ttf,woff,eot,scss,less,css.map,webp,woff2',
```
Jako: I would sort the list by type (scss, less, css.map after css; webp after bmp and woff2 after woff).

js.map is missing or css.map should be reduced to map.

Jako (changes requested):

JoshuaLuckers (approved):
This also needs to be done in 3.x

Jako (approved):

Ruslan-Aleev:
@Jako maybe it's better to arrange the groups of files in the order of frequency of use? For example, images are more often downloaded than archives. And dock files are more often than video, etc.

Ruslan-Aleev:
For example:
`'value' => 'html,htm,xml,js,js.map,css,scss,less,css.map,jpg,jpeg,png,tiff,svg,svgz,gif,psd,ico,bmp,webp,txt,pdf,doc,docx,xls,xlsx,ppt,pptx,odt,ods,odp,odb,odg,odf,md,zip,gz,rar,z,tgz,tar,mp3,mp4,aac,wav,au,wmv,avi,mpg,mpeg,ttf,woff,woff2,eot',`

Jako:
I just grouped them, because not wanted groups are easier to remove from the system setting then.

Issue #15001, opened 2020-02-08
Fatal error when using runProcessor()


wuuti:
### Summary
When I tried to use run-processor in an extra connector with 3.x, it fails with a fatal error. The possible reason is that the error handler is never added to the services container, but checked for in the runprocessor method.

### Step to reproduce
Try to call this in a connector:
```php
$response = $this->modx->runProcessor( 'workspace/packages/getlist', $scriptProperties );
```       

### Observed behavior
The following fatal error is thrown and logged:
```log
[Sat Feb 08 10:41:38.893028 2020] [proxy_fcgi:error] [pid 13097] AH01071: Got error 'PHP message: 
PHP Fatal error:  Uncaught MODX\\Revolution\\Services\\NotFoundException: 
Dependency not found with key error. in /vagrant/www/core/src/Revolution/Services/Container.php:28\n
Stack trace:\n
#0 /vagrant/www/core/src/Revolution/modX.php(1733): MODX\\Revolution\\Services\\Container->get('error')\n
#1 /vagrant/www/core/components/updater/model/updater.class.php(395): MODX\\Revolution\\modX->runProcessor('workspace/packages/getlist', Array)\n
#2 /vagrant/www/assets/components/mycomponent/connectors/myconnector.refresh.php(28): Class->myMethod()\n
#3 {main}\n  thrown in /vagrant/www/core/src/Revolution/Services/Container.php on line 28'
```

### Expected behavior
That error should not been thrown, and the error "service" should either be not checked in the runprocessor _or_ initialize it itself if its not there instead of just throwing the exception.

### Possible solution
This can by bypassed if something, e.g. the error handler is added to the container in modX.php:
```php
/**
     * Loads the error handler for this instance.
     *
     * @param array|null $options An array of options for the errorHandler.
     */
    protected function _initErrorHandler($options = null) {
        if ($this->errorHandler == null || !is_object($this->errorHandler)) {
            if ($ehClass = $this->getOption('error_handler_class', $options, modErrorHandler::class, true)) {
                $ehPath = $this->getOption('error_handler_path', $options, '', true);
                if ($ehClass = $this->loadClass($ehClass, $ehPath, false, true)) {
                    if ($this->errorHandler = new $ehClass($this)) {
                        $result = set_error_handler([$this->errorHandler, 'handleError'], $this->getOption('error_handler_types', $options, error_reporting(), true));
                        if ($result === false) {
[...]

PR #15162, opened 2020-07-18
Make the error_handler_class available as service

JoshuaLuckers:
### What does it do?
- Deprecates the public $errorHandler property
- Loads the error handler as a service 
- Use a anonymous function as callback for the PHP error handler using the new services container
- Use the error handler provided by MODX as fallback.

### Why is it needed?
Makes it easier to load a custom PHP error handler without the need to define it as system setting.
It also solves an issue that would break MODX when the error handler defined in the system settings was unable to load.

### Related issue(s)/PR(s)
Inspired by issue #15001

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -192,6 +192,7 @@ class modX extends xPDO {
     public $virtualDir;
     /**
      * @var modErrorHandler|object An error_handler for the modX instance.
+     * @deprecated Use the service/DI container to access the `error_handler` service.
      */
     public $errorHandler= null;
     /**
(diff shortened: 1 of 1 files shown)


opengeek:
Is there a reason it could not just have been converted to `errorHandler` instead of `error_handler`?

JoshuaLuckers:
@opengeek I decided for the snake case variant because it appears to be the default naming convention for Pimple. 
It also makes more sense in terms of what it's used for: set_error_handler

opengeek:
But the service has always been called `errorHandler` inside MODX. MODX service naming conventions have always been camel case, and I don't see why we would want to abandon it here. `set_error_handler` is just the PHP function—not the service class inside of MODX.

JoshuaLuckers:
You should have told me immediately 🙈. I'm wrong in the naming of the service and I'll change it in the next 24 hours.

JoshuaLuckers:
@opengeek I changed the name of the service.

Jako (approved):

JoshuaLuckers:
@opengeek ping

PR #15163, opened 2020-07-18
Fix a fatal error when calling runProcessor in a connector

### LINKED ISSUE #15001: Fatal error when using runProcessor()

JoshuaLuckers:
### What does it do?
Use the deprecated `getService` to load `modError`.

### Why is it needed?
To fix a fatal error when calling runProcessor in a connector.

### Related issue(s)/PR(s)
Fixes #15001

### Note
We definitely need to fix it the proper way, however doing it this way it will give us a bit more time without blocking a 3.0 release.

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1730,7 +1730,7 @@ class modX extends xPDO {
             $this->services->get('lexicon');
         }
         if (!$this->error) {
-            $this->services->get('error');
+            $this->getService('error', 'error.modError');
         }
 
         // First check if the processor can be found directly as a class name provided to $action
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1727,10 +1727,16 @@ class modX extends xPDO {
 
         // Make sure the required services are loaded before initialising a processor
         if (!$this->lexicon) {
-            $this->services->get('lexicon');
+            if (!$this->services->has('lexicon')) {
+                $this->services->add('lexicon', new modLexicon($this));
+            }
+            $this->lexicon = $this->services->get('lexicon');
         }
         if (!$this->error) {
-            $this->getService('error', 'error.modError');
+            if (!$this->services->has('error')) {
+                $this->services->add('error', new modError($this));
+            }
+            $this->error = $this->services->get('error');
         }
 
         // First check if the processor can be found directly as a class name provided to $action


JoshuaLuckers:
@wuuti would you be so kind to test this PR? 😄

theboxer:
I believe that using a deprecated method is not a proper fix :)

JoshuaLuckers:
> I believe that using a deprecated method is not a proper fix :)

Unless someone provides a PR with a proper fix this is a good alternative.

Ruslan-Aleev (approved):

Jako (approved):
Better than nothing.

PR #15131, opened 2020-06-25
Changed hiding/showing "Input Option Values" and added allowBlank check

Ruslan-Aleev:
### What does it do?
Changed hiding/showing "Input Option Values" and added allowBlank check.
- Now "Input Option Values" check in one place, no code in TV renders.
- Added allowBlank check for "Input Option Values". Previously, it was possible to create, for example, checkboxes, but not set "Input Option Values" and there was an empty TV in resource form.

tv_input_options

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15016

PR #15164, opened 2020-07-20
TV Input/Output Options, Form Enhancements

smg6511:
### What does it do?

This proposed enhancement changes the way that TV input and output properties are incorporated into the TV editing form, generating these dynamically-loaded fields as Ext components (as the rest of the form does) instead of a rendered DOM element. This allows a level of manipulation of those fields (and native Ext tracking of their dirty states) that's not currently possible, as well as fully-functional validation.

To preserve compatibility with community-developed custom TVs, the original renders and templates were left untouched and a new directory "configs" was created alongside the original "renders" one. The structure within is essentially the same, though the field configs are set up directly in php arrays instead of loading a Smarty-parsed tpl file. These configs are subsequently iterated in javascript and inserted as items into a new fieldset component.

Also, the Default Value and Input Option Values fields are only shown when appropriate. Note that the Default Value field's labelling changes based on the chosen TV type, and its description has been tweaked to hopefully be more representative of how the field is used.

I submit this in draft mode to get a read on what those active in modx development think of the idea and whether it's worthwhile to continue in the direction I'm heading. I believe it is, and helps refine the backend area of creating/updating elements.

More to come...

### Why is it needed?

To solve problems with validation and refine the control over/display of element form fields (particularly that of TVs).

### Related issue(s)/PR(s)

This builds on my earlier pull request #15146. Also, this addresses in a different way the fix proposed in #15131.


Mark-H:
This certainly sounds interesting, but should probably target 3.x for starters and the diff is a bit confusing with the other PR included here as well. 

Concept-wise, separating the definition and rendering of the input/output props sounds like a solid idea. Have you considered the route of adding it into the TV's Input class instead of with magical file names loading raw PHP before landing on this approach?

smg6511:
This has been further developed and exclusively targets 3.x now. Closing for 2.x.

Issue #14802, opened 2019-10-21
Context settings are not appearing in the javascript MODx.config variable in the manager


pmacswebteam:
### Summary
Context settings are not appearing in the javascript MODx.config variable in the manager

### Step to reproduce
When doing `console.log(MODx.config);`, the (custom) context setting `ckeditor.wctx` is not available/listed in MODX 2.7.1, but it available/listed in MODX 2.6.5

### Observed behavior
![Screen Shot 2019-10-21 at 2 16 48 PM](https://user-images.githubusercontent.com/12412403/67231834-1759ee80-f40e-11e9-9990-0dfcc8149f08.png)

### Expected behavior
![Screen Shot 2019-10-21 at 2 17 14 PM](https://user-images.githubusercontent.com/12412403/67231855-1fb22980-f40e-11e9-89c6-0948334f7c12.png)

### Environment
MODX 2.7.1

Mark-H:
As far as I know, `MODx.config` always only included system settings. The manager after all is its own context with its own context settings, and in many views (basically all views but editing a resource) there is no relevant working context to use. 

Are you sure this is not a CKEditor feature/bug to dynamically add that?

pmacswebteam:
@Mark-H We don't believe it's an issue specific to CKEditor as other context settings are displaying the same behavior. Each setting listed below is a setting that we override per context. In the MODX 2.6.5 examples, the overridden values are displaying properly for each context. For the MODX 2.7.1 examples, the overridden values are being ignored and displaying the system setting values instead:

**base_url in MODX 2.6.5** - contains context setting value, which has the name of context "simple"
<img width="277" alt="Screen Shot 2019-10-22 at 7 55 09 AM" src="https://user-images.githubusercontent.com/12412403/67283520-d3b0c480-f4a1-11e9-91a5-f2e16a8ddb53.png">

**base_url in MODX 2.7.1** - contains the default system setting value
<img width="188" alt="Screen Shot 2019-10-22 at 7 55 24 AM" src="https://user-images.githubusercontent.com/12412403/67283768-47eb6800-f4a2-11e9-8e84-1677c961b6c9.png">

---

**site_url in MODX 2.6.5** - contains context setting value, which has the name of context "simple"
<img width="274" alt="Screen Shot 2019-10-22 at 7 56 20 AM" src="https://user-images.githubusercontent.com/12412403/67283827-68b3bd80-f4a2-11e9-8667-8d32283446c5.png">

[...]

PR #15165, opened 2020-07-21
Fix url parameter for the config.js processor

Jako:
### What does it do?
Get the url parameter with $this->getProperty/$this->getProperties instead of using the not existing $scriptProperties array.

### Why is it needed?
During wrapping the code in the class in #14161, the $scriptProperties variable was not changed. It does not exist in the method process.

### Related issue(s)/PR(s)
#14161, #14802

### DIFF
--- a/core/model/modx/processors/system/config.js.class.php
+++ b/core/model/modx/processors/system/config.js.class.php
@@ -11,13 +11,9 @@
 /**
  * Outputs the $modx->config to JSON
  *
- * @param string $action If set with context, will output the context info for a
- * custom context by the action
- * @param string $context If set with action, will output the context info for a
- * custom context by its action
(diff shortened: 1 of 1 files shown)


opengeek:
Thanks for taking care of this one!

opengeek:
FWIW, this could not be applied to 3.x. If you have time and can prepare a version of this on 3.x, that'd be great. If not, I'll try and get to it later this week.

Jako:
This was already done in 2247e4d52f393598286077e63484eee118d40e69

opengeek:
> This was already done in [2247e4d](https://github.com/modxcms/revolution/commit/2247e4d52f393598286077e63484eee118d40e69)

Oh. Great! Thanks!

PR #15166, opened 2020-07-26
Fix fatal error when deleting resources in trash manager

raffy99:
### What does it do?
Correct parameters for the invokeEvent calls: OnBeforeEmptyTrash and OnEmptyTrash as stated in the docs https://docs.modx.com/3.x/en/extending-modx/plugins/system-events/onemptytrash

### Why is it needed?
OnBeforeEmptyTrash with old parameters resulted in: 
PHP Fatal error: Uncaught Exception: Serialization of 'Closure' is not allowed in /core/src/Revolution/modElement.php on line 278

### DIFF
--- a/core/src/Revolution/Processors/Resource/Trash/Purge.php
+++ b/core/src/Revolution/Processors/Resource/Trash/Purge.php
@@ -115,7 +115,6 @@ class Purge extends Processor
         // fire before empty trash event
         $this->modx->invokeEvent('OnBeforeEmptyTrash', [
             'ids' => &$this->ids,
-            'resources' => &$this->resources,
         ]);
 
         // we track success and failure independently, as we don't want
@@ -162,9 +161,8 @@ class Purge extends Processor
         }
 
         $this->modx->invokeEvent('OnEmptyTrash', [
+	        'ids' => &$success,
             'num_deleted' => count($success),
-            'resources' => &$this->resources,
-            'ids' => &$success,
         ]);
 
         $this->modx->logManagerAction('empty_trash', modResource::class, implode(',', $success));


JoshuaLuckers:
It would be nice to have a small test case to reproduce this issue.

raffy99:
I got the error while collections was installed. It is not reproducable in a fresh install.
But it is reproducable if you create a empty plugin with just return and add at least one of the two events OnBeforeEmptyTrash or OnEmptyTrash. So this might be an error or not, i am no more sure about this. Compared to the docs (https://docs.modx.com/3.x/en/extending-modx/plugins/system-events/onbeforeemptytrash) the "resources" parameter is unnecessary.

Jako:
But the resources parameter exists in Revolution 2.7.x. So there should be a different reason for this issue.

raffy99:
> But the resources parameter exists in Revolution 2.7.x. So there should be a different reason for this issue.

Yes, the reason seems to be a change in modElement: "Generate sortable tag signature with serialized array values [#14689]"

opengeek:
> > But the resources parameter exists in Revolution 2.7.x. So there should be a different reason for this issue.
> 
> Yes, the reason seems to be a change in modElement: "Generate sortable tag signature with serialized array values [#14689]"

I'm sorry, but I don't see how this is related at all. Can you please elaborate?

raffy99:
> > > But the resources parameter exists in Revolution 2.7.x. So there should be a different reason for this issue.
> > 
> > 
> > Yes, the reason seems to be a change in modElement: "Generate sortable tag signature with serialized array values [#14689]"
> 
> I'm sorry, but I don't see how this is related at all. Can you please elaborate?

I will try. The trash purge processor invokes events "OnBeforeEmptyTrash" and "OnEmptyTrash". if a plugin is triggered by the event, the parameters will pass through as properties to the getTag-function of modElement. My example plugin only returns empty string. "invokeEvent" function of modx calls the process function of modScript which calls parent process function and thats modElement. modElement process function calls the getTag function, which raises the error:
At this point i don't know if the parameters should be changed or not. Are the referenced docs valid?

opengeek:
> > > > But the resources parameter exists in Revolution 2.7.x. So there should be a different reason for this issue.
> > > 
> > > 
> > > Yes, the reason seems to be a change in modElement: "Generate sortable tag signature with serialized array values [#14689]"
> > 
> > 
> > I'm sorry, but I don't see how this is related at all. Can you please elaborate?
> 
> I will try. The trash purge processor invokes events "OnBeforeEmptyTrash" and "OnEmptyTrash". if a plugin is triggered by the event, the parameters will pass through as properties to the getTag-function of modElement. My example plugin only returns empty string. "invokeEvent" function of modx calls the process function of modScript which calls parent process function and thats modElement. modElement process function calls the getTag function, which raises the error:
> PHP Fatal error: Uncaught PDOException: You cannot serialize or unserialize PDO instances
> At this point i don't know if the parameters should be changed or not. Are the referenced docs valid?

Wow. Okay, thanks for that. I do indeed see the problem now, but this is not the fix for it. There will need to be something put in place to resolve the problem introduced, as you indicated, by the changes in #14689. I appreciate you taking the time to explain.

raffy99:
Remains the question regarding the docs:
https://docs.modx.com/3.x/en/extending-modx/plugins/system-events/onbeforeemptytrash and
https://docs.modx.com/3.x/en/extending-modx/plugins/system-events/onemptytrash

opengeek:
> Remains the question regarding the docs:
> https://docs.modx.com/3.x/en/extending-modx/plugins/system-events/onbeforeemptytrash and
> https://docs.modx.com/3.x/en/extending-modx/plugins/system-events/onemptytrash

I don't know. I'm assuming someone just forgot to add the new parameter to the documentation.

raffy99:
Still a question in my head, for what the reference to the resource-iterator is needed. And what happens after the resources are deleted in the db and the resource-parameter is still passed in the OnEmptyTrash event?

Jako:
The resource iterator could be used to delete connected data (i.e. images) of the deleted resources in OnBeforeEmptyTrash. The resource iterator could contain not deleted rows (for whatever reason) in OnEmptyTrash.

JoshuaLuckers:
@raffy99 thanks again for the PR and info! Is the info below part of a bug report? If not, do you have the time to create one?

> I will try. The trash purge processor invokes events "OnBeforeEmptyTrash" and "OnEmptyTrash". if a plugin is triggered by the event, the parameters will pass through as properties to the getTag-function of modElement. My example plugin only returns empty string. "invokeEvent" function of modx calls the process function of modScript which calls parent process function and thats modElement. modElement process function calls the getTag function, which raises the error:
> PHP Fatal error: Uncaught PDOException: You cannot serialize or unserialize PDO instances
> At this point i don't know if the parameters should be changed or not. Are the referenced docs valid?

Jako:
In my opinions the docs just have to be updated or am I wrong with that?

PR #15171, opened 2020-07-29
Fix bug in the resource trash interface

sergant210:
### What does it do?
1. Fix SQL query bug if there are no deleted resources. This query causes an error "Encountered empty IN condition with key id".
2. Remove JOIN subquery because it's redundant.

### Why is it needed?
A quick way to check
1. Move to "manager/index.php?a=resource/trash" page.
2. Look at the MODX error log.

### DIFF
--- a/core/model/modx/processors/resource/trash/getlist.class.php
+++ b/core/model/modx/processors/resource/trash/getlist.class.php
@@ -59,7 +59,11 @@ class modResourceTrashGetListProcessor extends modObjectGetListProcessor
         if (!empty($context)) {
             $c->where(array('modResource.context_key' => $context));
         }
-        $c->where(array('modResource.id:IN' => $this->getDeleted()));
+        if ($deleted = $this->getDeleted()) {
+            $c->where(array('modResource.id:IN' => $deleted));
+        } else {
+            $c->where(array('modResource.deleted' => true));
+        }
 
         return $c;
     }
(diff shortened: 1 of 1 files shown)


core/model/modx/processors/resource/trash/getlist.class.php
```diff
@@ -59,15 +59,18 @@ public function prepareQueryBeforeCount(xPDOQuery $c)
         if (!empty($context)) {
             $c->where(array('modResource.context_key' => $context));
         }
-        $c->where(array('modResource.id:IN' => $this->getDeleted()));
+        if ($deleted = $this->getDeleted()) {
+            $c->where(array('modResource.id:IN' => $deleted));
+        } else {
+            $c->where(array('modResource.deleted' => true));
```
Jako: This fix should get an empty result a different way. I would suggest:

```
$c->where(array('modResource.id' => 0));
```

The query in getDeleted could get changed later on to something, that does not only check the deleted state but also the deletedby or the usergroups of the deletedby user.

Jako (changes requested):

core/model/modx/processors/resource/trash/getlist.class.php
sergant210: Done.

Jako (approved):

core/model/modx/processors/resource/trash/getlist.class.php
Jako: Thanks!

PR #15173, opened 2020-07-29
Fixed lexicons for consistency

Ruslan-Aleev:
### What does it do?
- Changed "folder" to "directory" in lexicons.

PR only changes lexicons entries.

### Why is it needed?

### Related issue(s)/PR(s)


opengeek:
I would vote for the opposite. Why not change any instances of folder to directory for consistency?

JoshuaLuckers:
It depends on the context. See: https://en.m.wikipedia.org/wiki/Directory_(computing)

Ruslan-Aleev:
> It depends on the context. See: https://en.m.wikipedia.org/wiki/Directory_(computing)

On many computers, directories are known as folders. :)

Ruslan-Aleev:
> I would vote for the opposite. Why not change any instances of folder to directory for consistency?

Changing `folder` to `directory` changes more lines. So I chose this option.

Ruslan-Aleev:
In general, a `directory` is a broader concept, and a `folder` refers specifically to the file structure, as it seems to me (https://en.m.wikipedia.org/wiki/Directory_(computing)#Folder_metaphor).

>  but if accessed through a graphical file manager, users may sometimes call it a folder.

And most of the lexicon lines affect exactly the files structure.

ghost:
Folders are directories, it's the same thing. They are an organizational unit of a hierarchical file system. And yes, there are other types of file systems. We can thank Windows and its pretty pictures of file folders for directories being called folders.

opengeek:
No. We are not changing extras and add-ons to packages / components.

opengeek (dismissed):
Folder needs to become directory, not the other way around. And MODX "components" have always been called extras, and are called that elsewhere in the ecosystem. Also, let's stop making multiple sets of changes in a single PR.

Ruslan-Aleev:
> Also, let's stop making multiple sets of changes in a single PR.

It's better to do something than not to do it at all, right? These are lexicon edits, I don't see any point in doing separate PRs. There is no code here.

> Folder needs to become directory, not the other way around.

I see no reason to use "directory" instead of "folders".

> And MODX "components" have always been called extras, and are called that elsewhere in the ecosystem.

You can search the lexicons for entries. "Packages" and "components" are mentioned many times more often than "extras". What ecosystem are you talking about?

opengeek:
> It's better to do something than not to do it at all, right? These are lexicon edits, I don't see any point in doing separate PRs. There is no code here.

In this case, I am willing to let consensus drive the decision to make the change for consistency on the directory / folder issue. But I'm not likely to agree to change extras. Putting changes like this together in one commit makes it harder to make progress and get consensus by combining the debate of unrelated changes and making them dependent on one another.

> What ecosystem are you talking about?

https://modx.com/extras/
https://modmore.com/

Ruslan-Aleev:
> Putting changes like this together in one commit makes it harder to make progress and get consensus by combining the debate of unrelated changes and making them dependent on one another.

There are two commits, not one - this is the firstly. Secondly, it is impossible to know in advance whether there will be debates on commit or not. Otherwise, all PRs would have to be created with one commit :)

> https://modx.com/extras/ & https://modmore.com/

The MODX lexicons have the number of entries:
- extras - 9
- components - 20
- packages - 120

MODX, for which these "extensions" are created, uses one terminology, and the ecosystem is completely different. This is strange.
How to do it right, in the end? Rename "packages" and "components" to "extras", or only "packages" or what?

opengeek:
> There are two commits, not one - this is the firstly. Secondly, it is impossible to know in advance whether there will be debates on commit or not. Otherwise, all PRs would have to be created with one commit :)

This is one of many reasons to limit the scope of each PR as much as possible. Also, where possible, all PRs should be squashed to one commit anyway.

> How to do it right, in the end? Rename "packages" and "components" to "extras", or only "packages" or what?

All extras come in packages, but not all packages are extras. I think we are trying to oversimplify this. Components should probably be extras, but otherwise, I do not see a need to convert everything to one concept because they are two related but still individual concepts.

Bruno17:
there might also be 'Extras' which don't come in a package. And not all Extras are 'Components'.
An Extra could just be an Element, like just a snippet, plugin, template, chunk, whatever....

Ruslan-Aleev:
> An Extra could just be an Element, like just a snippet, plugin, template, chunk, whatever....

The goal is to simplify entities. Now there will be 2 left:
- Package - a mechanism for installing something ("Extras", "Plugin") in the manager panel;
- Extra - an entity installed from a package. A plugin can also be installed, but there is a section called Plugins.

> there might also be 'Extras' which don't come in a package. And not all Extras are 'Components'.

What do you mean? Can you give examples?

JoshuaLuckers:
I agree with @opengeek, folder needs to become directory.

Ruslan-Aleev:
- Changed "folder" to "directory" in lexicons.
- Changed "components", "add-ons" to "packages / extras" in lexicons.

alroniks:
I agree with @opengeek to have separated pull requests for different topics to discuss. Atomic commits and even pull requests easier to validate and discuss. There is a nice example when the discussion started from the folder/directory question but finished about components and extras. And it is a bad example by the meaning.

JoshuaLuckers (approved):

Ibochkarev (approved):

Ruslan-Aleev:
@opengeek When you have time, check the PR

Mark-H (approved):

core/lexicon/en/package_builder.inc.php
```diff
@@ -46,7 +46,7 @@
 $_lang['resolver_name_desc'] = 'The name of the resolver. Used for organizational purposes.';
 $_lang['resolver_remove'] = 'Delete Resolver';
 $_lang['resolver_remove_confirm'] = 'Are you sure you want to delete this resolver?';
-$_lang['resolver_source_desc'] = 'The absolute path of the source of the resolver. If it is a file resolver, select the directory of the files you\'d like to copy. If a PHP Script, specify the script. Example: <br /><br />/public_html/modx/_build/components/demo/';
+$_lang['resolver_source_desc'] = 'The absolute path of the source of the resolver. If it is a file resolver, select the directory of the files you\'d like to copy. If a PHP Script, specify the script. Example: <br /><br />/public_html/modx/_build/components/demoextra/';
```
opengeek: I don't think this is necessary.

core/lexicon/en/topmenu.inc.php
```diff
@@ -14,7 +14,7 @@
 $_lang['admin'] = 'Admin';
 $_lang['api_docs'] = 'API Documentation';
 $_lang['api_docs_desc'] = 'Complete API documentation for MODX Revolution.';
-$_lang['components'] = 'Extras';
+$_lang['components'] = 'Packages';
 $_lang['content_types'] = 'Content Types';
```
opengeek: This should remain extras.

core/lexicon/en/topmenu.inc.php
```diff
@@ -50,7 +50,7 @@
 $_lang['import_site'] = 'Import HTML';
 $_lang['import_site_desc'] = 'Import HTML files to Resources';
 $_lang['installer'] = 'Installer';
-$_lang['installer_desc'] = 'Manage Add-ons and Distributions';
+$_lang['installer_desc'] = 'Manage Packages and Distributions';
```
opengeek: I suggest this be changed to "Manage Extras"

core/lexicon/en/workspace.inc.php
```diff
@@ -23,14 +23,14 @@
...
-$_lang['dir_err_core_comp'] = 'Your core/components/ directory was not able to be created at: [[+path]] <br />You will need to create this directory and make it writable if you want to use Package Management or 3rd-Party Components.';
+$_lang['dir_err_assets'] = 'Your assets/ directory was not able to be created at: [[+path]] <br />You will need to create this directory and make it writable if you want to use Package Management or 3rd-Party Extras.';
+$_lang['dir_err_assets_comp'] = 'Your assets/components/ directory was not able to be created at: [[+path]] <br />You will need to create this directory and make it writable if you want to use Package Management or 3rd-Party Extras.';
+$_lang['dir_err_core_comp'] = 'Your core/components/ directory was not able to be created at: [[+path]] <br />You will need to create this directory and make it writable if you want to use Package Management or 3rd-Party Extras.';
 $_lang['download'] = 'Download';
-$_lang['download_success'] = 'The Extra was downloaded successfully!';
+$_lang['download_success'] = 'The Package was downloaded successfully!';
```
opengeek: Why do we need this to change? Extras imply a package. This is not necessary.

core/lexicon/en/workspace.inc.php
```diff
@@ -23,14 +23,14 @@
...
 $_lang['downloaded'] = 'Downloaded';
 $_lang['downloading'] = 'Downloading...';
-$_lang['download_extras'] = 'Download Extras';
+$_lang['download_extras'] = 'Download Packages';
```
opengeek: Same as above.

core/lexicon/en/workspace.inc.php
```diff
@@ -152,14 +152,14 @@
 $_lang['packages'] = 'Packages';
 $_lang['packages_in_category'] = 'Below is a list of the packages available in the selected Category.';
 $_lang['packages_browse_msg'] = 'Click details on a package to get more information. Click Download to download the package to your MODX Revolution install.';
-$_lang['packages_desc'] = 'Packages are collections of MODX components that can be used within MODX in a variety of ways. Here you can manage them, update their installation status, disable or enable installed packages, or download new ones from remote providers.';
+$_lang['packages_desc'] = 'Packages are collections of MODX extras that can be used within MODX in a variety of ways. Here you can manage them, update their installation status, disable or enable installed packages, or download new ones from remote providers.';
```
opengeek: This is confusing the issue. Packages are collections of MODX components that form an extra. If anything this should be changed to "Extras are packages of MODX components…" This is making my head hurt.

core/lexicon/en/workspace.inc.php
```diff
@@ -184,14 +184,14 @@
 $_lang['provider_err_remove'] = 'An error occurred while trying to delete the Provider.';
 $_lang['provider_err_save'] = 'An error occurred while trying to save the Provider.';
 $_lang['provider_home_msg'] = 'Choose a Category from the left to begin browsing this Provider.';
-$_lang['provider_home_title'] = 'Download Extras';
+$_lang['provider_home_title'] = 'Download Packages';
```
opengeek: Unnecessary change, again.

opengeek (changes requested):
I would prefer to remove the extras/packages/components-related changes from this PR so that can be dealt with separately. Then we could get the other merged.

Ruslan-Aleev:
@opengeek I tried to designate such logic, I indicated it in the description:

>  Everything that is downloaded, updated, assembled and archived - is packages. Already installed and Configurable - Extra*

*I changed it to Extra at your request, although Components is more familiar, it seems to me.

In chats, 2 entities are more often used: packages (what is in the archive) and components. Why are you creating additional entities?

core/lexicon/en/workspace.inc.php
Ruslan-Aleev: It looks weird and unfinished. When 1 entity in neighboring buttons is called differently.

package

opengeek:
They have been referred to as Extras for quite some time. An Extra is synonymous with a component. Packages are what Extras come in. A package could potentially contain more than one Extra, or even other things beyond Extras. So technically they are Extras packages, or packages containing an Extra. So I could be alright with changing the lexicons for the "Package Management" area to all refer to packages, because that technically manages packages. But, Extras is what should be referred to in the top menu and is the name of a single "component" that extends the CMS.

Ruslan-Aleev:
Hahah, In general, I did not understand much from your message, where 60% are the words "Extra", "Packages" and "Components" :)

I'll remove the "extra/packages" commit, let someone else fix it in the future, if it needs to be fixed at all.

smg6511:
I'm way late to this party and won't weigh in on this specifically, but I'd encourage everyone to try to remove themselves from the programmer's perspective when making these decisions on naming conventions (and other descriptive texts) —leaning more toward natural language that non-computer science, content manager type end users would more likely relate to.

PR #15174, opened 2020-07-29
Language switcher now remembers the page

Ruslan-Aleev:
### What does it do?
When changing the language, the language switch opened the manager home page, which is inconvenient. The language switcher now remembers the page.

**Unfortunately, caching does not update the `&page` (switch language page) parameter and I have no idea how to fix this problem.**

If you comment out this line - https://github.com/Ruslan-Aleev/revolution/blob/20eb1adb899a43d52fa8a7888e3bec564ee1b60f/core/src/Revolution/modMenu.php#L69, then the switch works as desired.

switch_lang

p.s. Most likely there is a more competent solution, without using `_SERVER`.

### Why is it needed?
Now the language switcher remembers the page on which the language was changed.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14046

### DIFF
--- a/core/src/Revolution/modMenu.php
+++ b/core/src/Revolution/modMenu.php
@@ -94,6 +94,11 @@ class modMenu extends modAccessibleObject
 
         $this->xpdo->lexicon->load('core:languages');
 
+        $currentPage = '';
+        if (!empty($_SERVER['REQUEST_URI'])) {
+            $currentPage = '&page=' . urlencode($_SERVER['REQUEST_URI']);
+        }
+
         foreach ($languages as $code => &$language) {
             $language = [
                 'id' => $code,
@@ -104,7 +109,7 @@ class modMenu extends modAccessibleObject
                 ),
                 'parent' => 'language',
                 'action' => 'language',
-                'params' => '&switch=' . $code,
+                'params' => '&switch=' . $code . $currentPage,
                 'namespace' => 'core',
                 'permissions' => ''
             ];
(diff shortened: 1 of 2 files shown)


Ruslan-Aleev:
@Jako @JoshuaLuckers @Alroniks Any ideas on how to fix the caching problem?
p.s. I understand that the user will not switch the language often, but the way the switch works now is not very convenient. Especially when testing PR.

core/src/Revolution/modMenu.php
```diff
@@ -94,6 +94,11 @@ protected function getLanguageMenu()
 
         $this->xpdo->lexicon->load('core:languages');
 
+        $currentPage = '';
+        if (!empty($_SERVER['QUERY_STRING'])) {
```
Jako: $_SERVER['QUERY_STRING'] could contain more than the page request parameter.

Maybe you could use `$params = $this->modx->request->getParameters` to access the values, add the switch parameter and `str_replace('%2F', '/', $this->modx->toQueryString($params))` as params value. Then a CMP would be open after the switch etc.

Jako (changes requested):

theboxer:
@Ruslan-Aleev you still planning to work on this as it's still marked as draft?

Ruslan-Aleev:
@theboxer I'm not sure if this can be fixed with this menu implementation.
I can close PR, maybe you have a smart solution.

theboxer:
@Ruslan-Aleev I don't have any solutions, just going through PRs that are left in void.

alroniks:
@Ruslan-Aleev I've created a proper solution in the #15895 and this one I am closing if you don't mind.

PR #15179, opened 2020-08-03
Changed the lexicon to be more specific

Ruslan-Aleev:
### What does it do?
In the "Contexts", "User groups" and "Users" sections there is a button for creating a setting. However, it is more correct to specify a more specific lexicon, which is also used in the "System settings" section, especially since the `create_new` lexicon is no longer used in the code.

create_setting_1

create_setting_2

### Why is it needed?
- Changed the lexicon to be more specific
- Removed unnecessary lexicon

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/core/lexicon/en/default.inc.php
+++ b/core/lexicon/en/default.inc.php
@@ -92,7 +92,6 @@ $_lang['create'] = 'Create';
 $_lang['create_document_here'] = 'Create document here';
 $_lang['create_document_inside'] = 'Create document inside';
 $_lang['create_folder_here'] = 'Create folder here';
-$_lang['create_new'] = 'Create New';
 $_lang['create_user_group'] = 'Create User Group';
 $_lang['created'] = 'Created';
 $_lang['createdon'] = 'Creation date';
--- a/manager/assets/modext/widgets/security/modx.grid.user.group.settings.js
+++ b/manager/assets/modext/widgets/security/modx.grid.user.group.settings.js
@@ -22,7 +22,7 @@ MODx.grid.GroupSettings = function(config) {
         ,save_action: 'Security/Group/Setting/UpdateFromGrid'
         ,fk: config.group
         ,tbar: [{
-            text: _('create_new')
+            text: _('setting_create')
             ,cls:'primary-button'
             ,scope: this
             ,handler: {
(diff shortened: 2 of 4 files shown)


Ibochkarev (commented):
Is there a setting_create lexicon key?

Ruslan-Aleev:
> Is there a setting_create lexicon key?

Good question. There is a lexicon, but in another file - https://github.com/modxcms/revolution/blob/fa199b846b45b01f5ee7e10458e03ec9ee24fa57/core/lexicon/en/setting.inc.php#L39, but everything works. @JoshuaLuckers is this how it should work?
What then is the point of the `default.inc.php` file if the lines from other lexicon files are perfectly visible, there is no error? Or am I confusing something?

JoshuaLuckers:
I like where this is going! 

> https://github.com/modxcms/revolution/blob/fa199b846b45b01f5ee7e10458e03ec9ee24fa57/core/lexicon/en/setting.inc.php#L39
> 
> , but everything works. @JoshuaLuckers is this how it should work?
> What then is the point of the `default.inc.php` file if the lines from other lexicon files are perfectly visible, there is no error? Or am I confusing something?

MODX, in most cases, **_always loads the default core lexicons_** (`default.inc.php`). 

When interacting with the **manager** `modManagerRequest` is the entry point and is responsible for handling the request and preparing a response. The default core lexicons are loaded here:
https://github.com/modxcms/revolution/blob/e991c4f15aa92a3c3c4f06514e932093800643d0/core/src/Revolution/modManagerRequest.php#L113

The response and output we get is the responsibility of `modManagerResponse`, who in turn loads an instance of `modManagerController` to render the output. [See the documentation for more details about controllers](https://docs.modx.com/current/en/extending-modx/custom-manager-pages#adding-a-base-controller).

Before loading the `modManagerController` **_additional lexicon topics_** are loaded:
https://github.com/modxcms/revolution/blob/021277ca192354088b45fec37f03e387c5b8fd41/core/src/Revolution/modManagerResponse.php#L72

The controller **_also loads additional lexicon topics_** specified by in the `getLanguageTopics` function. E.g: for the manager page to update a user group: https://github.com/modxcms/revolution/blob/803ebb4303af0a925c8cdde53a25340f53904dc8/manager/controllers/default/security/usergroup/update.class.php#L94-L100

Ruslan-Aleev:
@JoshuaLuckers Thanks for the detailed answer. So everything works as it should :)

JoshuaLuckers:
> [..] So everything works as it should :)

More or less, yes.

Ibochkarev (approved):

Jako (approved):

PR #15180, opened 2020-08-04
Added character limit for 'site_name' in header link

Ruslan-Aleev:
### What does it do?
Added character limit for `site_name` and **icon** for link in header.

**Before:**
site_name_1

**After:**
site_name_2

### Why is it needed?
UI/UX fixes

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -907,6 +907,14 @@
     &:focus {
       color: $colorSplash;
     }
+
+    &:after {
+      @extend %pseudo-font;
+      content: fa-content($fa-var-eye);
+      margin-left: 5px;
+      font-size: 14px;
+      opacity: 0.5;
+    }
   }
 
   img+a {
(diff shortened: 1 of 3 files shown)


theboxer (approved):

GulomovCreative:
@Ruslan-Aleev Ext has own method ellipsis, [use it](https://docs.sencha.com/extjs/3.4.0/#!/api/Ext.util.Format-method-ellipsis)

`Ext.util.Format.ellipsis(MODx.config.site_name, 45, true);`

Ruslan-Aleev:
Thank you, I didn't know, I'll fix it.
Although `subst()` is also used in files, the question is not important :)

Jako:
I would remove the icon. It is no external link, so it is misleading. The ellipsis indicates, that it is truncated.

Ruslan-Aleev:
I can set another icon, for example, the same as in the widget. I placed the icon because it is not clear that there is a link here, there was a discussion on the forum about this and that the link is missing.

icon

Ruslan-Aleev:
Changed icon and ellipsis method

PR #15193, opened 2020-08-12
Changed the width of fields for "Package" panel

Ruslan-Aleev:
### What does it do?
- Changed the width of fields for "Package" panel. Made the width 100%.
- Moved the "Changelog" area higher because this information is more important than general package information and license.

Before:
package_panel_1

After:
package_panel_3

### Why is it needed?
For unification in design.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/manager/assets/modext/workspace/package/package.panel.js
+++ b/manager/assets/modext/workspace/package/package.panel.js
@@ -1,11 +1,18 @@
-
+/**
+ * The package info container
+ *
+ * @class MODx.panel.Package
+ * @extends MODx.Panel
+ * @param {Object} config An object of options.
+ * @xtype modx-panel-package
+ */
 MODx.panel.Package = function(config) {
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

Ruslan-Aleev:
In the process, I discovered an interesting bug - the content of the first textarea is scrolled down.

Jako (approved):

Jako:
The scrolling does not occur on the latest Chrome on Mac OS. Which browser do you use?

Ruslan-Aleev:
As it turned out, this problem is observed in Windows 10, in the Google Chrome browser latest (and in inconite mode), although there is no such bug in the Edge browser.
p.s. For MODX 2.x this is also observed.

theboxer (approved):

PR #15196, opened 2020-08-13
Changed "Template Variable" to "TV" in lexicons

Ruslan-Aleev:
### What does it do?
In the manager, Template Variable is sometimes called in full version, sometimes in an abbreviated version.
I brought the naming to the following logic:

When creating a TV, the description contains the full naming and abbreviated name in brackets, for clarification, in other variants the abbreviated form will be used.

Before:
tv_1

After:
tv_2

Eexcept the tree and tabs (there are only 3 such sections): Resource, Template and Form Customization (otherwise, it will be difficult to navigate with the cursor).

tv_4

### Why is it needed?

### Related issue(s)/PR(s)


Ibochkarev (approved):

core/lexicon/en/tv.inc.php
```diff
@@ -10,56 +10,56 @@
...
-$_lang['tv_management_msg'] = 'Manage additional custom content fields (Template Variables) for your documents.';
-$_lang['tv_msg'] = 'Here you can create/edit Template Variable. Template Variables must be assigned to templates in order to access them from snippets and documents.';
-$_lang['tv_name'] = 'Variable Name';
-$_lang['tv_new'] = 'Create Template Variable';
-$_lang['tv_novars'] = 'No Template Variables found';
+$_lang['tv_input_options_msg'] = '<p>Here you can edit the input options for the TV, specific to the type of input render that you select.</p>';
+$_lang['tv_lock'] = 'Lock TV for editing';
+$_lang['tv_lock_msg'] = 'Users must have the edit_locked attribute in order to edit this TV.';
+$_lang['tv_management_msg'] = 'Manage additional custom TVs for your documents.';
+$_lang['tv_msg'] = 'Here you can create/edit Template Variable (TV). TVs must be assigned to templates in order to access them from snippets and documents.';
```
JoshuaLuckers: Where possible we should use the [abbreviation element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr):
```suggestion
$_lang['tv_msg'] = 'Here you can create/edit <dfn>Template Variable</dfn> (<abbr title="Template Variable">TV</abbr>). TVs must be assigned to templates in order to access them from snippets and documents.';
```

Ruslan-Aleev: In my opinion, it is sufficient to add the `<abbr>` tag only on this `tv_msg` line, as you suggested.
For the rest of the lines, this is not very useful:
- The familiar TV is already used and users have no questions;
- Bugs may appear, as in https://github.com/modxcms/revolution/issues/14437

Ibochkarev (approved):

core/lexicon/en/template.inc.php
```diff
@@ -41,14 +41,14 @@
...
-$_lang['template_no_tv'] = 'No template variables have been assigned to this template yet.';
+$_lang['template_no_tv'] = 'No TVs have been assigned to this template yet.';
 $_lang['template_properties'] = 'Default Properties';
 $_lang['template_reset_all'] = 'Reset all pages to use Default template';
 $_lang['template_reset_specific'] = 'Reset only \'%s\' pages';
 $_lang['template_title'] = 'Create/edit template';
-$_lang['template_tv_edit'] = 'Edit the sort order of the template variables';
-$_lang['template_tv_msg'] = 'The template variables assigned to this template are listed below.';
+$_lang['template_tv_edit'] = 'Edit the sort order of the TVs';
+$_lang['template_tv_msg'] = 'The TVs assigned to this template are listed below.';
```
JoshuaLuckers:
```suggestion
$_lang['template_tv_msg'] = 'The <abbr title="Template Variables">TVs</abbr> assigned to this template are listed below.';
```

core/lexicon/en/tv.inc.php
JoshuaLuckers: In the panel descriptions it's useful, furthermore I agree.

JoshuaLuckers (commented):
I'm fine with using the abbreviation, however I think it makes sense to have the heading for the create/edit mode use the full definition.

core/lexicon/en/tv.inc.php
JoshuaLuckers: In the panel descriptions it's useful, furthermore I agree.

Ruslan-Aleev:
> I'm fine with using the abbreviation, however I think it makes sense to have the heading for the create/edit mode use the full definition.

The line in the title of the TV creation form is also used in the element tree. If we leave it full, it will look like in the screenshot (links in tree):
tv_1

JoshuaLuckers (approved):

Jako (approved):

PR #15199, opened 2020-08-14
Fix PHP error with absent array index

sergant210:
### What does it do?
Fix PHP error with absent array index in the category tree.
```
(.../processors/mgr/element/getnodes.class.php : 298) PHP notice: Undefined offset: 1
```

### Why is it needed?
If expand the category list (only the first level) in the element tree, `$map` contains only one element with index 0.  And `$map[1]` is not set. To see this error the system setting "log_level"must be set to 2.

### DIFF
--- a/core/model/modx/processors/element/getnodes.class.php
+++ b/core/model/modx/processors/element/getnodes.class.php
@@ -294,8 +294,10 @@ class modElementGetNodesProcessor extends modProcessor {
             );
         }
 
-        foreach (array_keys($this->actionMap) as $type) {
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

Jako:
Maybe https://github.com/modxcms/revolution/blob/05b5e582a5514acb3fcf398516e1e71ddc195a92/core/model/modx/processors/element/getnodes.class.php#L238 could be changed to the same check. But please check before, if they are triggered by the same situation.

PR #15205, opened 2020-08-19
Fix of checking existence of <?php tag

sergant210:
### What does it do?
Fix the check of starting php tag existence.

### Why is it needed?
if snippet or plugin code contains the "<?php" tag somewhere, it will not be added to the beginning of the code. This causes some problems.

### Way to reproduce the problem
1. Create a snippet with code
```
<?php
$var1 = 'String';
$var2 = '<?php';
return $var2 . $var1;
```
2. Save it.
3. Wait for the page to reload.
4. Look at the code - 
```
$var1 = 'String';
$var2 = '<?php';
return $var2 . $var1;
```

### DIFF
--- a/manager/controllers/default/element/plugin/update.class.php
+++ b/manager/controllers/default/element/plugin/update.class.php
@@ -100,7 +100,7 @@ class ElementPluginUpdateManagerController extends modManagerController {
         $this->pluginArray = $this->plugin->toArray();
         $this->pluginArray['properties'] = $data;
         $this->pluginArray['plugincode'] = $this->plugin->getContent();
-        if (strpos($this->pluginArray['plugincode'],'<?php') === false) {
+        if (strpos(ltrim($this->pluginArray['plugincode']),'<?php') !== 0) {
             $this->pluginArray['plugincode'] = "<?php\n".$this->pluginArray['plugincode'];
         }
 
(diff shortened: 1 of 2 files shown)


Jako (approved):

JoshuaLuckers (approved):

JoshuaLuckers:
This also needs to be done for 3.x

opengeek:
I'm confused. There is not supposed to be an opening PHP tag in a snippet or plugin's code.

opengeek:
I think the bug is that the `<?php` is being put into the field automatically. That is irrelevant. The tag should always be stripped out before saving.

sergant210:
> I'm confused. There is not supposed to be an opening PHP tag in a snippet or plugin's code.

Even a check like this?
```
if (strpos($fileContent, '<?php') === false) {
```
Anyway, the check is incorrect. It checks for the existence of a <?php tag somewhere in the code, not just at the beginning.

Jako:
It is not irrelevant for i.e. Ace. The <?php tag is added during editing the plugin code. It is stripped away during saving the plugin code.

opengeek:
I don't understand. This is not in the database at all. There should NEVER be PHP tags in MODX code at all.

opengeek:
What is the purpose of this tag being added?

Jako:
It is shown in the editor field in the backend. Because of adding this, the code highlighting of Ace and other backend code editors is executed. Thats the reason for adding the <?php tag. See the screenshots:

Bildschirmfoto 2020-08-25 um 16 05 10
Bildschirmfoto 2020-08-25 um 16 05 26

opengeek:
Thanks, Jako. I didn't realize that was needed for those extensions. Makes more sense now.

PR #15217, opened 2020-08-29
Changed id for options for TV output with html-tag type

Ruslan-Aleev:
### What does it do?
- Changed id for options for TV output with html-tag type: for `id` and `attributes` options.
- Removed unnecessary `tag_id` lexicon, replacing it with `id` lexicon.

### Why is it needed?
Some option ids are the same and some are not.
Identical option ids are more convenient, because when you change the output type, the values are set in, you do not need to enter them again.

tv-tag-output

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15212
https://github.com/modxcms/revolution/pull/15208

### DIFF
--- a/core/lexicon/en/tv_widget.inc.php
+++ b/core/lexicon/en/tv_widget.inc.php
@@ -118,7 +118,6 @@ $_lang['start_day_desc'] = 'Day index at which the week should begin, 0-based (d
 $_lang['string'] = 'String';
 $_lang['string_format'] = 'String Format';
 $_lang['style'] = 'Style';
-$_lang['tag_id'] = 'Tag ID';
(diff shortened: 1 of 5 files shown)


Jako (changes requested):
You have to write a converter for pre 3.x installations, if you change the property keys of a template variable.

Ruslan-Aleev:
> You have to write a converter for pre 3.x installations, if you change the property keys of a template variable.

Are you talking about the update script? Yes, I'll finish it a little later. Although I do not believe in the upgrade from 2.x to 3.x, useless work :)

Jako (approved):
Thanks, looks good.

PR #15218, opened 2020-08-29
Changed options for TV output with url type

Ruslan-Aleev:
### Upd
Because here the option id also changes (`attrib` -> `atributes`), then **this PR needs to be merged after** https://github.com/modxcms/revolution/pull/15217 Because the update script is the same.

### What does it do?
- Changed id for options for TV output with **url type** `attributes` option.
- Added `ID` option: for all tags (img, html-tag) the output has an id, but for the link there is no, it will not be superfluous.

### Why is it needed?
Some option ids are the same and some are not.
Identical option ids are more convenient, because when you change the output type, the values are set in, you do not need to enter them again.

Example with  setting of option values:
tv-tag-output

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15217
https://github.com/modxcms/revolution/pull/15212
https://github.com/modxcms/revolution/pull/15208


Ruslan-Aleev:
Because here the option id also changes (`attrib` -> `atributes`), then **this PR needs to be merged after** https://github.com/modxcms/revolution/pull/15217 Because the update script is the same.

PR #15221, opened 2020-08-30
Corrected lexicon for 'use_context_resource_table' setting

Ruslan-Aleev:
### What does it do?
Corrected lexicon for `use_context_resource_table` setting.
With the previous lexicons, it seemed that the setting did not belong to the "Caching" area.

### Related issue(s)/PR(s)
N/A


opengeek:
This is actually a cache-related setting, as it determines if the resource caches are built using this table. I'm not sure this is necessary.

Ruslan-Aleev:
If so, then we should indicate this in the name of the setting and the description so that there are no questions in the future.
Can you write a more correct text indicating that this setting related to caching? I will then close this PR and create another one that corrects the lexicons.

opengeek:
Add an explicit 'cache' between 'context' and 'refreshes' in the description. The phrase 'context refreshes' implies a cache refresh.

> When enabled, context cache refreshes use the context_resource table. This enables you to programmatically have one resource in multiple contexts. If you do not use those multiple resource contexts via the API, you can set this to false. On large sites, you will get a potential performance boost in the manager then.

Ruslan-Aleev:
@opengeek I corrected the PR. Please check.

Issue #14400, opened 2019-02-17
Media manager sometimes uploads files to the wrong folder


SnowCreative:
### Summary
The media manager doesn't always upload files to the expected folder.

### Step to reproduce
1. Set up a TV (file or image) and link it to a specic media source. In that media source, create folders (like "images/newsletter/" for example). 
2. In a resource that uses that TV, link the TV to a file INSIDE a subfolder. Save the resource. 
3. Edit a different resource, then go back to the original resource.
4. Edit the TV. In the middle pane, the media browser will show the files inside the folder corresponding to the TV setting you saved.
5. Click the upload button in the upper left of the media browser and upload a file. 

### Observed behavior
Following the steps above, the file will be uploaded into the ROOT folder of the media source, NOT the folder whose contents are displayed in the main panel.

### Expected behavior
As it is now, a folder on the left has to be clicked on first, before clicking the upload button, to make files go into that folder. When the media browser first opens, if the contents of a subfolder are displayed in the main panel, that folder should ALSO be selected in the left list of folders/files. Otherwise, even though a folder is not selected on the left, users think that uploading a file will go into the folder that is being viewed in the main panel. 

### Environment
All versions of MODX.

SnowCreative:
Put another way, when opening the media browser from a TV field that has a file entered, make the media browser tree have the folder containing that file selected.

JoshuaLuckers:
Thanks for adding additional information!

SnowCreative:
Any update on this? I continually have to educate new clients about this bug in the media manager so they can upload files where they want them to go.

SnowCreative:
Any update?

JoshuaLuckers:
Sorry, I don’t have any update on this. My time is also a bit limited due to my baby girl being born! 

Maybe @Ruslan-Aleev is able to help fix this?

Ruslan-Aleev:
Sorry, but I have no time and no opportunity.

Maybe @Jako is able to help fix this? :)

SnowCreative:
> Sorry, I don’t have any update on this. My time is also a bit limited due to my baby girl being born!
>

Hey, that's exciting! Congrats!

Ruslan-Aleev:
Related with https://github.com/modxcms/revolution/issues/5320

Ruslan-Aleev:
> Any update?

[...]

Issue #5320, opened 2011-07-24
Suggestion: Should be possible to set File Manager to always remember last location


modxbot:
pumkinweb created Redmine issue ID 5320

I'm running into this issue a lot with end-users recently, where in situations such as managing galleries or lists of images (often using an extra such as MIGX), the process of adding images or other files, one-by-one becomes lengthy when the user has to open the correct folder from the root each time. For example, if they have a set of files 4-folders deep, each time they add a new file into a TV field, they have to navigate to the correct folder from the base path each time.

The file manager does appear to remember the last folder location when replacing a file in a TV, but I'd like to suggest that either the file manager should always remember the last location, or it should be possible to add a setting to achieve the same.

Gallenkamp:
Guys, you closed both the issues with reference to each other.

#6455 AND #5320. 

Please reopen one or them.

rtripault:
@Gallenkamp this one appears open ;) (the reference before your comment could be confusing tho, since it displays the status of the other issue, which is closed)

Gallenkamp:
Ah! Ok :) Thank you for your quick response! I will always scroll up to the top from now on :D

sebastian-marinescu:
I'd also like to see this feature

+1

Peeet93:
Are there any news about this feature?

ilyautkin:
There are many cases when users will "stumble" on this feature. Example:

1. User select image for TV `one` from `/assets/images/bg/` - directory `assets` is open in tree;
2. Then user want to select file for TV `two` from `/docs/insctuctions/machine`. But files tree open `/assets/images/bg/`;
3. User have to close `assets` and open `docs`.

Any TV can be assotiated with its own Media Source. If you will use it, you can set different directories for different TVs.

I think this feature is not needed in core.

JoshuaLuckers:
Maybe it's an idea to add a function that has a list of most recent directories that have been opened. Clicking one of the items in the list will automatically navigate to that directory in the media browser.  
The list should contain only directories for the Media Source that applies to the resource that is being edited.

Ibochkarev:
Fixed #15227

@modxbot close

PR #15227, opened 2020-09-06
Fixed incorrect path setting when uploading files

Ruslan-Aleev:
### What does it do?
- Added reset of active node when changing source.

If you select directory in one media source, then select another source and immediately try to upload the file, the downloader will try to upload to the previous source (active node).

**How the bug looks:**
sources_bug_1

- Added saving previous path to upload file

If there are several directories in the source, then when editing the TV the file will be uploaded into the root of the source, although the viewer shows the files of the correct directory.

**How the bug looks:**
sources_bug_2

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14400
https://github.com/modxcms/revolution/issues/5320

### DIFF
--- a/manager/assets/modext/widgets/system/modx.tree.directory.js
+++ b/manager/assets/modext/widgets/system/modx.tree.directory.js
@@ -254,6 +254,7 @@ Ext.extend(MODx.tree.Directory,MODx.tree.Tree,{
     }
 
     ,changeSource: function(sel) {
+        this.cm.activeNode = '';
         var s = sel.getValue();
         var rn = this.getRootNode();
         if (rn) { rn.setText(sel.getRawValue()); }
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
@rainbowtiger Please check.

opengeek:
Is this already solved in 3.x?

Ruslan-Aleev:
@opengeek No, the problem is still relevant for 3.x

matdave:
Nice catch. Tested and works.

theboxer (approved):

opengeek:
In the future, please submit these PRs with a single commit when possible. Having three commits for a few lines changed in one file unnecessarily complicates the git history for the project.

Ruslan-Aleev:
@opengeek Got it. I think, the correct to do is to merge the commits into one after the discussion ends and after the reviewers have added the "ready for merging" label. It's just that the reviewers are asking for split commits for ease of testing.

opengeek:
If you are submitting a draft for collaboration, I can see how having separate commits you can remove if necessary could be necessary. But when you submit it for final review, it should be squashed. I would squash it myself before merging, but I cannot do that when manually merging (so I can do the grunt build and/or add changelog entries) or it will not show the PR as merged.

Issue #13234, opened 2017-01-04
Filemanager doesn't mark selected file


himurovich:
### Summary
Filemanager doesn't mark selected file, "Filesystem" folder remains closed.

### Step to reproduce
In the backend sidebar click on "Files" then open "Filesystem" and select any file. 

![template variable_ new tv _ modx revolution - mozilla firefox 2017-01-04 12 32 14](https://cloud.githubusercontent.com/assets/2815640/21657302/a7f39334-d2d2-11e6-95be-26d07e465d4e.png)

### Observed behavior
"Filesystem" still closed, file wasn't marked as selected. 

![edit file_ ht access _ modx revolution - mozilla firefox 2017-01-04 12 33 29](https://cloud.githubusercontent.com/assets/2815640/21657299/a7f1ca22-d2d2-11e6-9bac-4dae312c284e.png)

### Expected behavior

It would be nice if filemanager shows what file is currently opened, moreover "Filesystem" would also be active (open)

Here are some helpful screenshots where you can find element's classes which could solve the issue.

![firebug - edit file_ ht access _ modx revolution 2017-01-04 12 35 08](https://cloud.githubusercontent.com/assets/2815640/21657300/a7f23246-d2d2-11e6-97b1-97213140dc80.png)

![firebug - edit file_ ht access _ modx revolution 2017-01-04 12 46 12](https://cloud.githubusercontent.com/assets/2815640/21657303/a7f59a08-d2d2-11e6-931d-380992bd6cf1.png)

It is important to say that the same (current object selection within tree) works correctly for "Resources" and "Elements" tabs entities.
Also I've noticed that 2.5.1 version was able to mark as "opened" Filesystem folder icon, unfortunately file was also non-selected there. 

![edit file_ styles css - google chrome 2017-01-04 22 45 29](https://cloud.githubusercontent.com/assets/2815640/21657301/a7f331e6-d2d2-11e6-9e99-b482c3a9abd5.png)

### Environment
MODX Revolution 2.5.4-pl on MODXCloud

guyinpv:
I observed this behavior but only when no subfolders were expanded. (on 2.5.6-pl)

TEST: 
With NO subfolders expanded, click some file in the root folder like htaccess or index.php. 
Now simply expand ANY folder in the sidebar, just expand it, don't even select a file under it. Now click on a root-level file again. This time the sidebar remains open.

It seems to be collapsed when a file is opened, only when no subfolders are expanded. Strange.

[...]

PR #15228, opened 2020-09-07
Added indication of selected file & expanding active source in tree

Ruslan-Aleev:
### What does it do?
Added indication of selected file & expanding active source in tree.

In the current version, the sources in the tree are collapsed when you edit the file (until you expand any source folder).
And the file is not marked as active in the tree.

**This PR fixes that. How it looks now:**

sources_files

~p.s. `URLSearchParams` is not IE compatible, do I need to rewrite the code or can I leave it as it is?~

### Why is it needed?
UX fixes

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13234

### DIFF
--- a/manager/assets/modext/widgets/core/tree/modx.tree.js
+++ b/manager/assets/modext/widgets/core/tree/modx.tree.js
@@ -257,8 +257,20 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
      * @param node
      */
     ,prepareNodes: function(node) {
+        var params = {};
+        if (location.search) {
+            var parts = location.search.substring(1).split('&');
+
+            for (var i = 0; i < parts.length; i++) {
+                var nv = parts[i].split('=');
+                if (!nv[0]) continue;
+                params[nv[0]] = nv[1] || true;
+            }
+        }
+        var activeFile = params.file;
+
         Ext.each(node.childNodes,function (node) {
-            if (node.attributes.selected) {
+            if (node.attributes.selected || node.id == activeFile) {
                 //node.select();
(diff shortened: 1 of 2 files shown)


matdave:
Just tested this out. Works great!

Ruslan-Aleev:
@matdave My solution does not work for IE (I indicated in the description), do I need to redo it? On the one hand, IE is no longer supported, but on the other hand, many still use it.

opengeek:
> My solution does not work for IE (I indicated in the description), do I need to redo it? On the one hand, IE is no longer supported, but on the other hand, many still use it.

@Ruslan-Aleev 2.x officially supports IE 11 for the manager. Does it just not do what is expected in IE 11, or does it break the UX completely in IE 11?

Ibochkarev:
> > My solution does not work for IE (I indicated in the description), do I need to redo it? On the one hand, IE is no longer supported, but on the other hand, many still use it.
> 
> @Ruslan-Aleev 2.x officially supports IE 11 for the manager. Does it just not do what is expected in IE 11, or does it break the UX completely in IE 11?

whether we need IE support in 2020 is a rhetorical question.

Ruslan-Aleev:
@opengeek This solution does not work for IE, i.e. for IE everything will be as it was, without active class and expanding the tree.
I will now redo it so that there are no questions.

opengeek:
> @opengeek This solution does not work for IE, i.e. for IE everything will be as it was, without active class and expanding the tree.
> I will now redo it so that there are no questions.

I was only concerned if it broke the UI. If it still is operational, this is fine as is.

Ruslan-Aleev:
I redid PR, now works for old browsers and IE too.

PR #15231, opened 2020-09-10
Displayed caption instead of description in the template TV grid

Ruslan-Aleev:
### What does it do?
When creating a TV, the description is set less often than the caption field. Therefore, it is more logical to display caption.

Displayed `caption` instead of `description` in TV grid:

tpl-tvs

This is useful if there are many TVs in the grid.

### Why is it needed?
UI/UX fix

### Related issue(s)/PR(s)
N/A


Mark-H (approved):

Jako (approved):

Jako:
This has to go into 3.x, too.

Issue #14641, opened 2019-07-12
Improve mediabrowser handling for existing files


sdrenth:
### Summary
The mediabrowser handling can be improved when an image is uploaded which already exists. 

### Why is it needed?
This will improve user experience when uploading files which already exists.

### Suggested solution(s)
The best solution would be to ask the user what action to take when a file with the same name exists:

- Cancel the upload
- Overwrite the old file
- Keep both and generate a unique name for this file

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14566
https://github.com/modxcms/revolution/issues/8371

Ibochkarev:
@sdrenth Thanks

Added #15232 
@modxbot close

PR #15232, opened 2020-09-11
Added a check for the existence of a file when uploading it

Ruslan-Aleev:
### What does it do?
Added a check for the existence of a file when uploading it.
In the current version, **if the file names match, then the file will be overwritten.**

**This PR fixes that:**
file_exist

p.s. For Amazon source, we can probably add a file existence check too, in line
https://github.com/modxcms/revolution/blob/2.x/core/model/modx/sources/mods3mediasource.class.php#L823

add
```
if (file_exists($container.$file['name'])) {
    $this->addError('path',sprintf($this->xpdo->lexicon('file_err_ae'),$file['name']));
    continue;
}
```
but I cannot test it as I don't use sources from Amazon.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/8371
https://github.com/modxcms/revolution/issues/14641

### DIFF
--- a/core/model/modx/sources/modfilemediasource.class.php
+++ b/core/model/modx/sources/modfilemediasource.class.php
@@ -593,6 +593,25 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
     }
 
     /**
+     * Check that an object (directory, file) exists
+     *
+     * @param string $objectPath The object path to check
+     * @param string $objectName The object name displayed in the error message
+     * @return bool
+     */
+    protected function checkObjectExist($objectPath, $objectName) {
+        if (file_exists($objectPath)) {
+            if (is_dir($objectPath)) {
+                $this->addError('name', $this->xpdo->lexicon('file_folder_err_ae'));
+                return true;
+            }
+            $this->addError('name', sprintf($this->xpdo->lexicon('file_err_ae'), $objectName));
+            return true;
+        }
+        return false;
+    }
+
+    /**
      * @param string $oldPath
      * @param string $newName
      * @return bool
@@ -622,13 +641,7 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
         $newPath = $this->fileHandler->sanitizePath($newName);
         $newPath = dirname($oldPath).'/'.$newPath;
 
-        /* check to see if the new resource already exists */
-        if (file_exists($newPath)) {
-            if (is_dir($newPath)) {
-                $this->addError('name',$this->xpdo->lexicon('file_folder_err_ae'));
-                return false;
-            }
-            $this->addError('name',sprintf($this->xpdo->lexicon('file_err_ae'),$newName));
+        if ($this->checkObjectExist($newPath,$newName)) {
             return false;
         }
 
@@ -874,6 +887,10 @@ class modFileMediaSource extends modMediaSource implements modMediaSourceInterfa
 
             $newPath = $this->fileHandler->sanitizePath($file['name']);
             $newPath = $directory->getPath().$newPath;
+
+            if ($this->checkObjectExist($newPath,$file['name'])) {
+                return false;
+            }
 
             if (!move_uploaded_file($file['tmp_name'],$newPath)) {
                 $this->addError('path',$this->xpdo->lexicon('file_err_upload'));


Ruslan-Aleev:
It was only after I submitted the PR that I noticed that this had already been fixed but not merged - https://github.com/modxcms/revolution/pull/14514/commits/6b4cf38f9493d8b3032bf4feb3f7558c3fbebbdc ¯\_(ツ)_/¯
Should I redo my PR?

Jako:
The code of @webinmd checks the renaming of files too. Thats better. But the name of the new method is somehow strange.

Ruslan-Aleev:
@Jako I redid the PR, check when you have time.

matdave:
Would it make sense to have a system (or media source) setting to allow the original overwrite behavior? I personally use the overwrite method quite a bit to replace files, so this would add an extra step.

Ruslan-Aleev:
I added a check for the file because the same check already exists when renaming the file or category. And it is strange that it was not there when the file was uploaded. Replacing a file without alerting - the behavior is not normal.
Probably, in the future, we can add settings of this kind, but it seems to me that this is unnecessary, although you use this :)

Ruslan-Aleev:
There were other suggestions for fileloader behavior - https://github.com/modxcms/revolution/pull/14566
Perhaps in the future it is worth considering all the wishes and expanding the functionality, but, again, I think this is superfluous.

matdave:
I guess this is an improvement over silently overwriting. Would be ideal to later add a checkbox or setting to allow the previous behavior, but I can see how it could be perceived as a bug (instead of a feature, hah).

Jako (approved):

core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -592,6 +592,25 @@ public function checkFiletype($filename) {
...
     }
 
+    /**
+     * Check that an object (directory, file) exists
+     *
+     * @param string $objectPath The object path to check
+     * @param string $objectName The object name displayed in the error message
+     * @return bool
+     */
+    public function checkObjectExist($objectPath, $objectName) {
```
JoshuaLuckers: If I see a method like this I expect it to return `true` if the object already exists.

theboxer: I agree with Joshua, this should definitely return true if the object exists.

core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -622,13 +641,7 @@ public function renameObject($oldPath,$newName) {
...
         $newPath = dirname($oldPath).'/'.$newPath;
 
-        /* check to see if the new resource already exists */
-        if (file_exists($newPath)) {
-            if (is_dir($newPath)) {
-                $this->addError('name',$this->xpdo->lexicon('file_folder_err_ae'));
-                return false;
-            }
-            $this->addError('name',sprintf($this->xpdo->lexicon('file_err_ae'),$newName));
+        if (!$this->checkObjectExist($newPath,$newName)) {
```
theboxer: After the you adjust the `checkObjectExist` to return `true` if the object exists, this will need to change as well - removing the `!`.

core/model/modx/sources/modfilemediasource.class.php
```diff
@@ -875,6 +888,10 @@ public function uploadObjectsToContainer($container,array $objects = array()) {
             $newPath = $this->fileHandler->sanitizePath($file['name']);
             $newPath = $directory->getPath().$newPath;
 
+            if (!$this->checkObjectExist($newPath,$file['name'])) {
```
theboxer: Same as above.

theboxer (changes requested):

Ruslan-Aleev:
@JoshuaLuckers @theboxer I redid PR, please check.

core/model/modx/sources/modfilemediasource.class.php
Jako: I don't see the issue: It returns true, if the file or the path exists.

JoshuaLuckers: That's because @Ruslan-Aleev force-pushed his changes after my comment 😋

theboxer (approved):

opengeek (commented):
I'm concerned that this is an incomplete solution to the problem since it only resolves the bug if working with one of many potential media source implementations. But in order to resolve it appropriately, we would need to add this new method to the interface and expect all implementations to be updated to implement this method. Then we would find ourselves with a change that breaks backwards compatibility. I think this needs to be discussed and considered.

opengeek:
To clarify, I have two concerns here. One is that if this is just changing the behavior of a single media source, the extracted method here should not be public in scope. The second is regarding the consistency of behavior across all core and non-core media sources. If the default file media source does not automatically overwrite existing files in this UX, then shouldn't this be consistent for all media sources?

Jako:
@opengeek All your concerns are valid, but they can't be solved with 2.x. How many different media source implementations do currently exist? The S3 media source implementation in the core is outdated and does not work anymore (at least in Europe).

Ruslan-Aleev:
It improves the situation anyway, doesn't it?
For example, if other types of file sources use `core/model/modx/sources/modfilemediasource.class.php` in their base, then there are no problems and the file is not rewritten silently, and if these types of sources have their own logic, then it is probably already implemented there check of this kind (for example, miniShop2 has its own check). And if there is no check, then the situation for such sources does not change.

opengeek (changes requested):
Make this a protected method and I'll merge it then.

Ruslan-Aleev:
@opengeek I redid PR

opengeek (approved):
Looks good. Thanks, @Ruslan-Aleev!

SnowCreative:
I'm not sure what's happening here. Are you changing the behavior so that overwriting existing files is impossible? If so, I don't think we should change this behavior until it can be done properly, offering an option to replace or rename files on uploading. All my clients are used to updating existing files (like PDFs) with new versions without renaming so that links to those files throughout their sites will not get broken.

opengeek:
> I'm not sure what's happening here. Are you changing the behavior so that overwriting existing files is impossible? If so, I don't think we should change this behavior until it can be done properly, offering an option to replace or rename files on uploading. All my clients are used to updating existing files (like PDFs) with new versions without renaming so that links to those files throughout their sites will not get broken.

This is certainly a valid concern if there is currently no way to replace the files.

Ruslan-Aleev:
> All my clients are used to updating existing files (like PDFs) with new versions without renaming so that links to those files throughout their sites will not get broken.

Many (if not all) systems indicate this when uploading existing files rather than overwrite them. By the way, clients point to this too.
This behavior is better for UX than accidentally overwriting an important file. In your case, clients can always delete and download the file again, although this requires more steps.

Ruslan-Aleev:
Other options have already been discussed https://github.com/modxcms/revolution/pull/14566, but many users chose the warning option, if the file exist.
By the way, if we do some kind of setting, then we need to remember about renaming the file, and not just replacing it when uploading.

opengeek:
If we are changing behavior that existing clients are used to, then we need to account for that. #14566 was discussed for 3.x because it is a significant behavior change. In either case, IMO, this really needs to have an option to warn users before overwriting but allow them to be overwritten if they choose.

SnowCreative:
I see this was pushed through anyway and incorporated into the new 2.8 update. This is disastrous for all my clients, and I will NOT be upgrading until this is corrected properly. As opengeek said, we need to have an option to overwrite if desired, not just block it entirely.

Mark-H:
A PR was merged after discussion, and points raised **after** that merge did not lead to anyone providing a better pull request that matches the expectation of certain users before the release. Nothing's being "pushed through" - there has simply not been a better PR for review before the release. A full revert would not make it better overall. 

Users that want to overwrite existing files will first need to delete them explicitly rather than that happening magically and without any form of user feedback or interaction. It's not impossible to do so. 

I certainly understand some of your users may be accustomed to this and would welcome additional improvements that make it an optional checkbox of sorts in the upload window. Or a system setting that toggles it, perhaps, as proposed earlier in the thread.

Ruslan-Aleev:
@rainbowtiger 

> I see this was pushed through anyway...

Here you are substituting concepts. This PR was merged 7 days before you wrote your comment (look at the dates).

> This is disastrous for all my clients...

Is it really for all, and is it really a disaster? Clients will make more clicks, yes, but no more.

> I will NOT be upgrading until this is corrected properly.

You decide. But this now works **properly**. The fact that it was possible to overwrite the file without informing the user is just incorrect behavior and it has been fixed.
You can make a PR for the 2.x or 3.x branch based on https://github.com/modxcms/revolution/pull/14566 and create a system setting, and wrap the file check condition.

Mark-H:
@rainbowtiger How about this: if you plant some trees (let's say 100) at https://ecologi.com/modmore I'll prepare a pull request today to add a setting to disable the overwrite behavior and do what I can to get that into 2.8.1 for you. 🌴

Ruslan-Aleev:
@Mark-H if you are going to do PR, then I would leave 3 options for setting:
- Check existing files (used by default);
- Overwrite files;
- Create copies like in https://github.com/modxcms/revolution/pull/14566.

And the setting should also affect renaming existing files.

rthrash:

https://community.modx.com/t/version-2-8-is-out-i-wont-be-upgrading/3166/2

SnowCreative:
> Here you are substituting concepts. This PR was merged 7 days before you wrote your comment (look at the dates).
> 
I'm not sure what you mean by "substituting concepts".

> > This is disastrous for all my clients...
> 
> Is it really for all, and is it really a disaster? Clients will make more clicks, yes, but no more.

Yes, it is. Most of my clients are NOT very tech-savvy, so anything creating difficulties for them is not good.

> 
> > I will NOT be upgrading until this is corrected properly.
> 
> You decide. But this now works **properly**. The fact that it was possible to overwrite the file without informing the user is just incorrect behavior and it has been fixed.

I disagree. "Properly" would be having the option to overwrite, just like all file systems do. This change is only half of a "proper" fix. No doubt, it's better for some people. My point is that it's not better for ALL people. It gets rid of one deficiency, but creates another. 

> You can make a PR for the 2.x or 3.x branch based on #14566 and create a system setting, and wrap the file check condition.

Unfortunately, I'm not enough of a programmer to create pull requests.

SnowCreative:
> @rainbowtiger How about this: if you plant some trees (let's say 100) at https://ecologi.com/modmore I'll prepare a pull request today to add a setting to disable the overwrite behavior and do what I can to get that into 2.8.1 for you. 🌴

Great idea! 100 trees have been purchased! (modx@snowcreative.com)

SnowCreative:
> A PR was merged after discussion, and points raised **after** that merge did not lead to anyone providing a better pull request that matches the expectation of certain users before the release. Nothing's being "pushed through" - there has simply not been a better PR for review before the release. A full revert would not make it better overall.
> 
> Users that want to overwrite existing files will first need to delete them explicitly rather than that happening magically and without any form of user feedback or interaction. It's not impossible to do so.
> 
> I certainly understand some of your users may be accustomed to this and would welcome additional improvements that make it an optional checkbox of sorts in the upload window. Or a system setting that toggles it, perhaps, as proposed earlier in the thread.

A system toggle is so easy to add, I just wonder why that wasn't included for this pull request. There was plenty of feedback on other threads relating to this topic before this final request was merged. 

I do hope this topic continues and gets included in 3.0 with all the finished functionality. I'm not sure if there are still open issues around this slated for 3.0, or whether they were all closed because of this merge.

JoshuaLuckers:
> @rainbowtiger How about this: if you plant some trees (let's say 100) at https://ecologi.com/modmore I'll prepare a pull request today to add a setting to disable the overwrite behavior and do what I can to get that into 2.8.1 for you. 🌴 

I planted 100 trees!

Mark-H:
I'll get coding in about an hour or two. ;)

Mark-H:
Tadaa: https://github.com/modxcms/revolution/pull/15285

Thanks for all the trees, folks. Good day for the forest.

SnowCreative:
You've got to tell me how you signed up to have a modmore Ecologi page. I'd love to do that as well.

SnowCreative:
And now, I'm interpreting "upload a file through the tree" in a completely different way!

Mark-H:
It's a business profile: https://ecologi.com/business - only takes a few minutes to set up. The basic idea is you subscribe based on the number of employees to get a number of trees planted (and CO2 offsets) monthly based on that. Renews automatically, and honestly is pretty cheap as far as "business costs" go. It's fun to see the forest grow and read about the projects they're supporting too - definitely recommend it.

They also have personal profiles which I think work mostly the same as business profiles, but are more suited for individual people. 

Plus, I like asking for trees as an alternative to monetary donations for.. well, things like this ;)

Ibochkarev:
@Mark-H Nice project and needed. Well, now you can do a good job and develop MODX!

SnowCreative:
I signed up! Planted 72 more trees.

JoshuaLuckers:
> It's a business profile: https://ecologi.com/business - only takes a few minutes to set up. The basic idea is you subscribe based on the number of employees to get a number of trees planted (and CO2 offsets) monthly based on that. Renews automatically, and honestly is pretty cheap as far as "business costs" go. It's fun to see the forest grow and read about the projects they're supporting too - definitely recommend it.
> 
> 
> 
> They also have personal profiles which I think work mostly the same as business profiles, but are more suited for individual people. 
> 
> 
> 
> Plus, I like asking for trees as an alternative to monetary donations for.. well, things like this ;) 

Slightly off-topic, but my bank also plants trees for every 100 euro I spend using one of my cards. I love planting trees! 🌳

PR #15238, opened 2020-09-16
Prevent potential XXE vulnerability in modRestService by disabling libxml entity loader

### LINKED ISSUE #15237: XXE Vulnerability
snipped

Mark-H:
### What does it do?

Disables the libxml entity loader when reading an XML string in a request to a modRestService API. This is enabled by default, but can be disabled with a service option if custom DTDs are necessary in an API.

### Why is it needed?

Makes modRestService implementations more secure by default, fixing the report in #15237. The MODX core does not use this particular service, so only third party implementations would be affected. 

### Related issue(s)/PR(s)
 
Fixes #15237

### DIFF
--- a/core/model/modx/rest/modrestservice.class.php
+++ b/core/model/modx/rest/modrestservice.class.php
@@ -59,6 +59,7 @@ class modRestService {
             'responseSuccessKey' => 'success',
             'trimParameters' => false,
             'xmlRootNode' => 'response',
+            'xmlDisableEntityLoader' => true,
 		),$config);
 		$this->modx->getService('lexicon','modLexicon');
         if ($this->modx->lexicon) {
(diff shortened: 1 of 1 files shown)


Jako (commented):
What about this: https://php.watch/versions/8.0/libxml_disable_entity_loader-deprecation

BTW libxml 2.9.0 should be available since 2012.

opengeek:
> What about this: https://php.watch/versions/8.0/libxml_disable_entity_loader-deprecation
> 
> BTW libxml 2.9.0 should be available since 2012.

Interesting, @Jako. We should probably add one of those recommended checks in there to avoid future deprecated warning issues.

Mark-H:
Nice catch, I hadn't seen that yet. Updated to check the `LIBXML_VERSION` constant, which was recommended in the PR on php-src that deprecated it as it's supposedly possible for builds to already use a newer version that's safe by default pre-PHP8.

Jako (approved):

core/model/modx/rest/modrestservice.class.php
```diff
@@ -59,6 +59,7 @@ public function __construct(modX &$modx,array $config = array()) {
             'responseSuccessKey' => 'success',
             'trimParameters' => false,
             'xmlRootNode' => 'response',
+            'xmlDisableEntityLoader' => true,
```
JoshuaLuckers, line 62: Since the setting only works for a specific version range of the library, do we need this as default setting?

PR #15242, opened 2020-09-17
Located "Publish" and "Delete" checkboxes in one line

Ruslan-Aleev:
### What does it do?
Completes https://github.com/modxcms/revolution/pull/14545 There was a discussion, but then we could not fix it.

Before:
del_1

After:
del_2

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14545

### DIFF
--- a/_build/templates/default/sass/_xtheme-modx.scss
+++ b/_build/templates/default/sass/_xtheme-modx.scss
@@ -1189,7 +1189,6 @@ td.x-grid3-hd-menu-open .x-grid3-hd-inner {
   >.x-column {
     margin-right: 0;
     overflow: visible;
-
   }
 
   /* distance to the right for all following columns, this makes the first one 15px wider than the rest*/
@@ -1197,6 +1196,11 @@ td.x-grid3-hd-menu-open .x-grid3-hd-inner {
   >.x-column:not(.x-hide-display)~.x-column {
     margin-right: 0;
     margin-left: 15px;
+
+    &.resource-deleted-wrap {
+      margin-right: 0;
+      margin-left: 5px;
+    }
   }
 }
 
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -1124,7 +1124,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                 ,id: 'ta'
                 ,fieldLabel: _('resource_content')
                 ,anchor: '100%'
-                ,height: 468
+                ,height: 438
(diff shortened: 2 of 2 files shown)


Ibochkarev (approved):

JoshuaLuckers:
What happens if the screen is a bit smaller?

Ruslan-Aleev:
Yes, there is a problem. Although objectively speaking, it looks crooked with the current chechbox display (although a little better).

del_3

del_4

Ruslan-Aleev:
We can hide (text-indent: -999em) the text from the label for this screen size. There is a pop-up there, but the labels will not be displayed, it will be like this:

chekboxes_2

JoshuaLuckers:
Then there is still the problem of the elements below not having enough space. I looked into this a while ago but didn’t find a good solution without moving the elements to above/below the main section.

Jako:
Displaying the whole element below, in a new tab or appended to 'settings' could be an option (only for small screens).

Ruslan-Aleev:
It is necessary to reduce both the menu on the left in width and the tree on the left, because even at significantly width (1280px), the interface looks skinny. And for small screens, yes - we need to move it down or add a new tab.

Ruslan-Aleev:
I moved the discussion to another issue - https://github.com/modxcms/revolution/issues/15247
Because here is a PR about the delete checkbox, and it's easier to solve the issue with it.

Jako:
This could be merged, but we have to check the backend later on for responsiveness.

Jako (approved):

Ruslan-Aleev:
I'll adjust the styles a little later so that the label is hidden at a certain width, otherwise it looks crooked. https://github.com/modxcms/revolution/pull/15242#issuecomment-694736944

Jako:
But that's bad UX and should be solved different.

Ruslan-Aleev:
As part of this PR, I would hide the labels, or I would not merge this PR at all, I see no other solution.
If we solve the issue of adaptive globally, then yes, there is no need to hide labels, but this is another PR question.

Jako:
I don't see a responsive solution for moving fields/blocks in ExtJS. But an adaptive solution that changes only after a reload could be possible.

opengeek:
Is there any consensus on merging this or not?

ghost:
Having those two buttons close like that will be a problem for me, for what it's worth. I've learned the hard way to give "delete" buttons a great deal of respect and keep them away from more commonly used buttons whenever possible. Of course, this delete button is not as final as many others are in other environments. With Fred, in the element's context menu, the delete option is just below the duplicate option, and more than once I hit the wrong button, with rather different results than I was expecting.

PR #15246, opened 2020-09-18
Fixed display of the pagination toolbar

gvozdb:
Fixed bug with display of the pagination toolbar. If the request to the processor was delayed, the `expand` function had time to work because of this piece of code:
```
if (this.mode == 'remote' && !this.loaded && this.tries < 4) {
    // Store not yet loaded, let's wait a little bit
    this.tries += 1;
    Ext.defer(this.expand, 250, this);
    return false;
}
```
And when the answer came from the processor, the `expand' function was not started anymore. We had to make a small crutch that instantly collapse the list and expanded it again.

### What does it do?
Describe the technical changes you did.

### Why is it needed?

### Related issue(s)/PR(s)

### DIFF
--- a/manager/assets/modext/widgets/core/modx.combo.js
+++ b/manager/assets/modext/widgets/core/modx.combo.js
@@ -108,7 +108,7 @@ MODx.combo.ComboBox = function(config,getStore) {
         this.loaded = true;
         
         // Show the pagination panel if it didn't show up earlier
-        if (this.pageSize < this.store.getTotalCount() && this.pageTb.hidden === true) {
+        if (this.isExpanded() && this.pageSize < this.store.getTotalCount() && this.pageTb.hidden === true) {
             this.collapse();
             this.expand();
         }
--- a/manager/assets/modext/widgets/core/modx.combo.js
+++ b/manager/assets/modext/widgets/core/modx.combo.js
@@ -106,6 +106,12 @@ MODx.combo.ComboBox = function(config,getStore) {
         // Workaround to let the combobox know the store is loaded (to help hide/display the pagination if required)
         this.fireEvent('loaded', this);
         this.loaded = true;
+        
+        // Show the pagination panel if it didn't show up earlier
+        if (this.pageSize < this.store.getTotalCount() && this.pageTb.hidden === true) {
+            this.collapse();
+            this.expand();
+        }
     }, this, {
         single: true
     });
@@ -135,6 +141,7 @@ Ext.extend(MODx.combo.ComboBox,Ext.form.ComboBox, {
             }
             if(this.pageSize < this.store.getTotalCount()){
                 this.assetHeight += this.footer.getHeight();
+                this.pageTb.show();
             } else {
                 this.list.setHeight(this.list.getHeight() - this.footer.getHeight());
                 this.pageTb.hide();


Jako:
What is the reason for the delayed answer of the processor? It makes more sense to solve that issue than patching the ExtJS combo code.

gvozdb:
> What is the reason for the delayed answer of the processor? It makes more sense to solve that issue than patching the ExtJS combo code.

It does not depend on the code, the problem to the server. If the server is processing the request for a long time, this bug happens.

Jako:
> If the server is processing the request for a long time, this bug happens.

What is the reason for this? It is better to solve that reason than patching the combo code.

Jako:
If a processor does not load the result in 1 second, the issue is not an issue of the combo.

gvozdb:
> > If the server is processing the request for a long time, this bug happens.
> 
> What is the reason for this? It is better to solve that reason than patching the combo code.

What does this have to do with anything? There is a problem at the code level - a long answer from the server causes the problem - it does not show the paging toolbar. I have fixed this problem. What are you talking about, anyway? Why do you look at how long or short the server replies? What does the long answer from the server have to do with the correct MODX operation?!

gvozdb:
> If a processor does not load the result in 1 second, the issue is not an issue of the combo.

Are you kidding me, or what?! What is this nonsense? What do you care how long the server response takes? MODX must work correctly in any case!

Mark-H:
Please be nice @gvozdb, we should be able to disagree, but still be civil in the way we communicate. 

This seems like a fine improvement to make a very important core component more resilient for failure. I also agree with @Jako that if a specific processor is slow that we should look at improving that too, but that does not mean this fix is not a welcome improvement. Thanks for working on that. 

@gvozdb Can you please follow the instructions from the cla-bot? The CLA needs to be signed in order for the MODX project can accept anyone's contributions.

Mark-H (approved):

gvozdb:
@Mark-H Ok, I'm sorry if I was rude. The thing is, this is a very old bug that has been showing itself for a long time. Today I have found time to analyze and correct it. And in the end I am told that I need to optimize my server. No, it's not about the server, the bug can manifest itself on your server too.

I signed the CLA 1-2 hours ago.

Mark-H:
@cla-bot check

Mark-H:
Awesome, thanks. :) 

I think @Jako meant optimising the MODX processor code, not a server. But yeah, even if loading the data takes longer for whatever reason, a combobox should still work.

Jako:
@gvozdb Almost all issues with combobox getlist processors I have faced in the MODX core and in some extras are caused by not filtering the results to something that generates a short list. 

The processors return a many hundred or thousand items list on initial load and let the combo box code do the filtering. Thats causing the load issues and patching the combo code to solve those issues is the wrong way.

Thats the reason, why I am arguing this way.

JoshuaLuckers:
Cant we add a “load”/“success” listener to the store that calls the expand method?

Mark-H:
That's literally what this PR does @JoshuaLuckers.

JoshuaLuckers:
> That's literally what this PR does @JoshuaLuckers. 

My mistake, must be the sleep deprivation 👶🙈

Jako:
I am not sure, if this could lead to an unwanted issue. When the store is loaded (which could happen at two times: when the combo dropdown it is opened and when an initial value is assigned/displayed), the combo dropdown will be opened. So it has to be checked, what happens with an initial value.

gvozdb:
> I am not sure, if this could lead to an unwanted issue. When the store is loaded (which could happen at two times: when the combo dropdown it is opened and when an initial value is assigned/displayed), the combo dropdown will be opened. So it has to be checked, what happens with an initial value.

Yes, you are right! I added a commit to fix it.

Jako (approved):

Ibochkarev (approved):

PR #15258, opened 2020-09-21
Adjust getParents method to work with deleted resources

### LINKED ISSUE #15243: If a resource is deleted, then its parent is not shown in the breadcrumbs of the manager
### Summary
If a resource is deleted, then **its parent is not shown** in the breadcrumbs of the manager.
Quite an interesting bug :)

### Observed behavior
bc_bug

### Environment
MODX 3-alpha3

theboxer:
### What does it do?
I moved away from using resource map in the getParents method for getting the ID of all parents. The resource map doesn't contain deleted resources, so the parent ids were always empty for deleted resource. It also seemed unnecessary as for every ID the modResource object had to be fetched from DB. The new code is traversing the `Parent` relation until there's no parent or we hit the limit set by `resource_panel_max_crumbs` system setting.

### Why is it needed?
Breadcrumbs in edit resource panel don't show the parent path if the current resource is deleted.

### Related issue(s)/PR(s)
Resolves #15243

### DIFF
--- a/manager/controllers/default/resource/resource.class.php
+++ b/manager/controllers/default/resource/resource.class.php
@@ -583,8 +583,13 @@ abstract class ResourceManagerController extends modManagerController
         $ctx = null;
         $height = $this->modx->getOption('resource_panel_max_crumbs', null, 3);
         $columns = ['id', 'pagetitle', 'published', 'hidemenu', 'parent'];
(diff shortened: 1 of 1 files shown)


opengeek (approved):
Solid improvement.

opengeek:
Original PR branch is gone… superceded by #15427

PR #15260, opened 2020-09-21
Fix downloading files

### LINKED ISSUE #15245: Error downloading files from the manager
### Summary
For some reason, downloading the file gives an error, see gif:

download_bug

This happens in both the file tree and the media browser.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15129

### Environment
MODX 3-alpha3

theboxer:
### What does it do?
Not sure why, but the download process was sending XHR request Download processor to get the url of the file and after success response the same processor was used to initiate the download... I skipped the the xhr request as we don't need the URL for download and passed directly the `relativePath` to the Download processor and it seems to be working :) (tested on assets media source and as well on s3 media source).

### Why is it needed?
Downloading files from any media source except Filesystem is broken.

### Related issue(s)/PR(s)
Resolves #15245

### DIFF
--- a/core/src/Revolution/Processors/Browser/File/Download.php
+++ b/core/src/Revolution/Processors/Browser/File/Download.php
@@ -45,7 +45,6 @@ class Download extends Browser
 
         // Download file
         @session_write_close();
-        $file = $this->sanitize($this->getProperty('file'));
         try {
             if ($data = $this->source->getObjectContents($file)) {
                 $name = preg_replace('#[^\w-.]#ui', '_', $data['basename']);
--- a/manager/assets/modext/widgets/system/modx.tree.directory.js
+++ b/manager/assets/modext/widgets/system/modx.tree.directory.js
@@ -683,22 +683,7 @@ Ext.extend(MODx.tree.Directory,MODx.tree.Tree,{
 
     ,downloadFile: function(item,e) {
         var node = this.cm.activeNode;
-        MODx.Ajax.request({
-            url: MODx.config.connector_url
-            ,params: {
-                action: 'Browser/File/Download'
-                ,file: node.attributes.pathRelative
-                ,wctx: MODx.ctx || ''
-                ,source: this.getSource()
-            }
-            ,listeners: {
-                'success':{fn:function(r) {
-                    if (!Ext.isEmpty(r.object.url)) {
-                        location.href = MODx.config.connector_url+'?action=Browser/File/Download&download=1&file='+r.object.url+'&HTTP_MODAUTH='+MODx.siteId+'&source='+this.getSource()+'&wctx='+MODx.ctx;
-                    }
-                },scope:this}
-            }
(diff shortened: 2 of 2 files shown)


JoshuaLuckers:
What happens if the processor returns a failure message?

theboxer:
I'd say same thing you can see in the original issue: https://github.com/modxcms/revolution/issues/15245

core/src/Revolution/Processors/Browser/File/Download.php
```diff
@@ -45,7 +45,6 @@ public function process()
 
         // Download file
         @session_write_close();
-        $file = $this->sanitize($this->getProperty('file'));
```
Mark-H, line 48: Were you having issues with this applied? Seems like a good idea to not blankly accept whatever file parameter is provided, and the sanitize method probably helps with that.

theboxer: @Mark-H it's a double call of the `sanitize` method. It's already called on line 36.

Mark-H: Ahh, gotcha 👍

Mark-H (approved):

JoshuaLuckers:
Do we still need to following if we don't use the XHR request to retrieve the URL?
https://github.com/modxcms/revolution/blob/8d054b9d4a9c253a9a3a8ee8da8b07c57018ea7f/core/src/Revolution/Processors/Browser/File/Download.php#L42-L44

And I'm still having a hard time understanding how this is a fix. Shouldn't we use the XHR to check if we can download the file and return a success or error response?

opengeek:
@theboxer — could you respond to @JoshuaLuckers question so I am confident there is no confusion remaining here before merging this?

theboxer:
@JoshuaLuckers the xhr is not there to verify if the file exists. If you look at the processor it always returns a success. Even if you pass a file that doesn't exist.

IMHO the processor is a relic from 2.x before media sources got changed in 3.x. Right now the processor translates a relative to media source root file path to a relative path to modx's base. This new path is then passed as prop to download, but the download happens via the media source method, that expects relative path to the media source (like in the first step). But it gets different and it doesn't exist in the media source -> error.

I believe there's not a method on the media source to check if the file is valid.

JoshuaLuckers:
> IMHO the processor is a relic from 2.x before media sources got changed in 3.x. Right now the processor translates a relative to media source root file path to a relative path to modx's base. This new path is then passed as prop to download, but the download happens via the media source method, that expects relative path to the media source (like in the first step). But it gets different and it doesn't exist in the media source -> error.

@theboxer @opengeek Now I understand, thanks!

Jako (approved):

opengeek:
Superseded by #15385 because the original PR branch was deleted by the author.

Issue #15244, opened 2020-09-17
Incorrect display of paths for files


Ruslan-Aleev:
### Summary
After merging the PR https://github.com/modxcms/revolution/pull/15227 into the 3.x branch, I noticed that the category does not become active in the file browser.

This is due to condition not being triggered:

```
if (n.id == this.config.openTo) {
    n.select();
}
```

Because the decoded path is set in `n.id`, for example `images%2Fproducts%2Ffile.txt`, and the unencoded path is set in `this.config.openTo` - `images/product/file.txt`.

Of course, we can just decode the `this.config.openTo` in the condition, but we need to be sure that there are no more problems of this kind.

**We need to bring it to a single format.**

There has already been a mention of this in previous issues, but I didn't find it.

![path](https://user-images.githubusercontent.com/12523676/93507062-cc01d600-f925-11ea-9030-6dbd28918ecf.png)

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15227
https://github.com/modxcms/revolution/pull/15228

### Environment
MODX 3-alpha3

Ibochkarev:
![изображение](https://user-images.githubusercontent.com/2138260/111323952-86d01e00-8694-11eb-8e8e-c2d1fb3e5f5c.png)

The problem is relevant in the installation from github from 16-03-2021

Ibochkarev:
Still happens on 3.x. The check from 16 March 2021 still matches the code.

A directory node id in `modMediaSource` is `rawurlencode()` of the relative path plus a trailing slash, so `images/products/` is stored as `images%2Fproducts%2F`. `getOpenTo()` returns the plain path with a slash, `images/products/`. `MODx.tree.Directory._saveState()` still compares those two strings directly:

```js
if (n.id == this.config.openTo) {
    n.select();
}
```

That comparison does not run for any nested folder. `quickCreateFile()` already calls `decodeURIComponent(node.attributes.id)` before it uses the path. `attributes.path` on the same node is the plain path. The id stays encoded because `editFile()` and `createFile()` put it in the query string.

`_initExpand()` does not use that id check. It calls `selectPath('/' + _('files') + '/' + this.config.openTo, 'text')`. The root node text is the media source name (`rootName` starts as `Filesystem` and is replaced when the source changes), not the Files lexicon, so that path does not point at the folder either.

[...]

PR #15262, opened 2020-09-21
[Security] [Requires rebuild] Fix XSS in file upload and file tree

Mark-H:
### What does it do?

Prevent XSS vulnerabilities in the manager by users with file upload permission.

### Why is it needed?

Specially crafted filenames can result in a self-xss in the file upload dialog and a stored XSS by the file tree. This sanitises that.

This also, inadvertently, fixes crlf line endings in the s3 media source file which were incorrect. [Diff without line ending change](https://github.com/modxcms/revolution/pull/15262/files?diff=unified&w=1)

### Related issue(s)/PR(s)

Responsibly disclosed by Nguyen Dang Toan from VSEC RED TEAM, thanks.

### DIFF
--- a/core/model/modx/processors/browser/directory/getlist.class.php
+++ b/core/model/modx/processors/browser/directory/getlist.class.php
@@ -59,6 +59,10 @@ class modBrowserFolderGetListProcessor extends modProcessor {
         $this->source->initialize();
 
         $list = $this->source->getContainerList($this->getProperty('dir'));
+        foreach ($list as &$item) {
+            // Make sure the id is HTML-safe as it will be inserted into an attribute
+            $item['id'] = htmlentities($item['id'], ENT_QUOTES, 'UTF-8');
+        }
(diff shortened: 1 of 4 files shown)


Ruslan-Aleev:
Unfortunately, after the implementation of this PR, the PR https://github.com/modxcms/revolution/pull/15228 does not work, because node id and url path no longer match (and class is not applied to the node), more details for 3.x - https://github.com/modxcms/revolution/issues/15244

JoshuaLuckers:
Can you create a new issue for this?

Ruslan-Aleev:
@JoshuaLuckers Added a new issue https://github.com/modxcms/revolution/issues/15302.
It turns out that https://github.com/modxcms/revolution/pull/15227 works, but https://github.com/modxcms/revolution/pull/15228 does not work for 2.x. And for 3.x everything is exactly the opposite.

PR #15264, opened 2020-09-22
Add check that the file exists when uploading it

Ruslan-Aleev:
### What does it do?
Same as in https://github.com/modxcms/revolution/pull/15232 but for the 3.x branch.

Added a check for the existence of a file when uploading it.
In the current version, **if the file names match, then the file will be overwritten.**

files_check

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15232
https://github.com/modxcms/revolution/issues/8371
https://github.com/modxcms/revolution/issues/14641


Ibochkarev (approved):

theboxer (approved):

JoshuaLuckers (changes requested):
Please add the system setting that's introduced in PR #15285.

Ruslan-Aleev:
@JoshuaLuckers I forgot about PR completely, added a setting, check :)

_build/data/transport.core.system_settings.php
```diff
@@ -1730,6 +1730,15 @@
...
 ], '', true, true);
+$settings['upload_file_exists']= $xpdo->newObject('modSystemSetting');
+$settings['upload_file_exists']->fromArray(array (
+  'key' => 'upload_file_exists',
+  'value' => '1',
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'file',
+  'editedon' => null,
+), '', true, true);
```
JoshuaLuckers:
```suggestion
$settings['upload_file_exists']->fromArray([
  'key' => 'upload_file_exists',
  'value' => true,
  'xtype' => 'combo-boolean',
  'namespace' => 'core',
  'area' => 'file',
  'editedon' => null,
], '', true, true);
```

JoshuaLuckers (dismissed):

JoshuaLuckers:
I believe it's better to have the markup be part of the lexicon.

Issue #15263, opened 2020-09-22
JavaScript error while creating a new folder in the root of a media source


Oetzie:
### Summary
When you you create a new folder in the root of the media source you will get a JavaScript error.

TypeError: Cannot read property 'constructor' of undefined
 /manager/assets/modext/widgets/system/modx.tree.directory.js:593

### Step to reproduce
Open de the media browser and create a folder (do not select any parent folder), you will get a JavaScript error.

### Fix
Change the line 593 of modx.tree.directory.js:
`if (this.cm.activeNode.constructor.name === 'constructor' || parent === '' || parent === '/') {`
To:
`if ((this.cm.activeNode && this.cm.activeNode.constructor.name === 'constructor') || parent === '' || parent === '/') {`

### Environment
MODX 3 and at least MODX 2.7.3

Ibochkarev:
I confirmed bug in 2.7.3
![image](https://user-images.githubusercontent.com/2138260/93872026-629d1100-fcf1-11ea-95d1-64ee09f78659.png)

Ibochkarev:
The proposed solution works.

Mark-H:
Fixed in 2.8

PR #15265, opened 2020-09-22
Fix JavaScript error when create a new folder in the root

Ibochkarev:
### What does it do?
Fix JavaScript error when create a new folder in the root

```
TypeError: Cannot read property 'constructor' of undefined
/manager/assets/modext/widgets/system/modx.tree.directory.js:593
```

### Why is it needed?
When we create a new folder at the root of the media source, we get a JavaScript error.

### Related issue(s)/PR(s)
#15263


theboxer (approved):

Jako (approved):

Mark-H:
I can't seem to reproduce this but trust ya'lls judgement and it's a harmless fix.

PR #15273, opened 2020-10-04
Fix assorted stored XSS issues in the manager

Mark-H:
### What does it do?

Resolves a couple of stored XSS issues (one per commit for clarity) in the manager. All of these require manager access, with either access to edit resources or to upload files. I'm considering each of these low level issues. 

I expect these to cleanly cherry-pick into 3.x. If not, I can recreate them. 

One of these (resource pagetitle in duplicate/delete action from the tree) was responsibly disclosed by Nguyen Dang Toan  
from VSEC RED TEAM. The others I just stumbled across while going through things. 

### Why is it needed?

Tightens manager security - even if the impact of a XSS issue by a manager user is very low given the (expected) trusted nature of manager users, this still makes MODX 2.8 a little bit more secure than it was before. 

### Related issue(s)/PR(s)

Responsible disclosure 3103

Rebuild of assets is required due to the fix in the resource tree.


Mark-H:
Ported to 3.x in prevenhttps://github.com/modxcms/revolution/pull/15843

PR #15275, opened 2020-10-05
Normalize error message

Jako:
### What does it do?
Normalize the `tv_err_invalid_name` error message.

### Why is it needed?
Every other lexicon entry for `*_err_invalid_name` is using `<type> name is invalid`.

### Related issue(s)/PR(s)
none


JoshuaLuckers:
To me it would make sense if the error message would tell why the value is invalid.

Jako:
But the previous message did not tell which characters are invalid. So the degression is negligible for me.

We could translate the regular expression to something understandable - but then for every MODX element name.

JoshuaLuckers (approved):

JoshuaLuckers:
> But the previous message did not tell which characters are invalid. So the degression is negligible for me.

Fair enough!

Ibochkarev (approved):

PR #15279, opened 2020-10-07
Fix URL TV

theboxer:
### What does it do?
Use `{literal}` wrapper around everything that doesn't need twig tags.

### Why is it needed?
When URL TV is assigned to resource, the edit resource panel breaks. It's caused by having `}}` in the input's tpl as twig is trying to parse that as twig tag.

### Related issue(s)/PR(s)
https://community.modx.com/t/bug-in-modx-2-8/3164

### DIFF
--- a/manager/templates/default/element/tv/renders/input/url.tpl
+++ b/manager/templates/default/element/tv/renders/input/url.tpl
@@ -26,7 +26,12 @@ Ext.onReady(function() {
         ,typeAhead: false
         ,forceSelection: false
         ,msgTarget: 'under'
-        ,listeners: { 'select': { fn:MODx.fireResourceFormChange, scope:this}}
+        ,listeners: {
+            'select': {
+                fn:MODx.fireResourceFormChange,
+                scope:this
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

rthrash:

https://community.modx.com/t/bug-in-modx-2-8/3164/9

Ruslan-Aleev:
@theboxer please check the other TV types, just in case. I tested - everything seems to work, but you never know :)

Mark-H:
@Ruslan-Aleev I've created an instance of all core input types to test. Unless there's an issue with specific input properties, there's no other fatal errors there.

Ruslan-Aleev:
@Mark-H Thanks for the info :)

rthrash:

https://community.modx.com/t/bug-in-modx-2-8/3164/14

rthrash:

https://community.modx.com/t/upgrading-to-2-8-0-is-it-worth-it/3185/5

PR #15280, opened 2020-10-07
Fix XSS in resource group name at TV > Access Permissions + XSS in resource list TV

Mark-H:
### What does it do?

Fixes a stored XSS similar to one fixed in 2.8.0. Exploitable only by admin users with permission to edit resource groups.

Also fixes XSS in the resourcelist TV, if a resource pagetitle contains markup.

### Why is it needed?

Because while admin users are trusted users and this requires permissions not typically given to editors, technically still counts as security issue.

### Related issue(s)/PR(s)

#15273

### DIFF
--- a/manager/assets/modext/widgets/element/modx.grid.tv.security.js
+++ b/manager/assets/modext/widgets/element/modx.grid.tv.security.js
@@ -1,6 +1,6 @@
 /**
- * Loads a grid of resource groups assigned to a resource. 
- * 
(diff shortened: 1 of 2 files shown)


Mark-H:
Ported to 3.x in https://github.com/modxcms/revolution/pull/15843

Note I've not ported the renderer added to the grid, as that's already done by default on all grids in 3.0.

PR #15282, opened 2020-10-07
Fix JS TypeError if image/file TV fails validation, preventing error popup from showing [Rebuild]

Mark-H:
### What does it do?

Prevent JS error if an image/file tv fails validation.

### Why is it needed?

On 2.8.0 saving an image/file tv that is required (allow blank = no) results in:

```
modx.panel.js:405 Uncaught TypeError: Cannot read property 'hidden' of undefined
    at MODx.panel.Resource.showErroredTab (modx.panel.js:405)
    at MODx.panel.Resource.fn (modx.panel.resource.js:25)
    at h.Event.fire (ext-all.js:21)
    at MODx.panel.Resource.fireEvent (ext-all.js:21)
    at MODx.toolbar.ActionButtons.handleClick (modx.component.js:298)
    at S.onClick (ext-all.js:21)
    at HTMLSpanElement.I (ext-all.js:21)
 ```

This avoids the js error, allowing the popup to be shown. 

### Related issue(s)/PR(s)

#15146

### DIFF
--- a/manager/assets/modext/widgets/core/modx.panel.js
+++ b/manager/assets/modext/widgets/core/modx.panel.js
@@ -402,7 +402,7 @@ Ext.extend(MODx.FormPanel,Ext.FormPanel,{
                     ;
                 if (tvTabs && tvTabs.items && tvTabs.items.keys) {
                     var tvTabIndex = tvTabs.items.keys.indexOf(errFldPanelId);
-                    if (tvTabs.items.items[tvTabIndex].hidden)  {
+                    if (tvTabs.items.items[tvTabIndex] && tvTabs.items.items[tvTabIndex].hidden)  {
                         tvTabs.activate(errFldPanelId);
                     }
                 }


smg6511:
Not disputing this, but I want to replicate it to get a better understanding of the case. Can you detail how you make this error happen?

Mark-H:
If you're not seeing it when a tv is required and uncategorized, I'll have to double-check.

smg6511:
Yeah, I've tried with a bunch of scenarios (tvs below, above, only uncategorized, categorized and uncategorized) but can't seem to get the error to pop up.

Mark-H:
Okay, it looks to be specific to the **image and file TV types** and the categorisation isn't relevant - with an image/file TV that has allow blank disabled, the error does not popup and it does not open the right TV (sub)tab, wether that's uncategorized or something else. 

Changing the TV to a text/number/date/richtext input while keeping allowBlank disabled makes it work as expected.  

TVs are in the standard tab.

PR #15285, opened 2020-10-07
Add `upload_check_exists` system setting to allow overwriting files with same name

Mark-H:
## What does it do?

Adds a new setting, upload_check_exists, to allow reverting to the pre-2.8.0 behavior of quietly overwriting files with the same name on upgrade. 

Also fixes a potential self-xss due to insufficient escaping of the filename.

### Why is it needed?

As discussed in https://community.modx.com/t/version-2-8-is-out-easy-replacement-of-files-pdf-images-is-axed/3166 and the discussion in #15232, some people are used to being able of replacing files simply by uploading them with the same name. That behaviour was removed in #15232 (v2.8.0-pl) to prefer a stricter check where users are presented with an error when a file already exists with the same name.

The new setting, enabled by default, gives users the choice to return to that old behaviour and bypassing the check on upload by disabling the setting. 

(The check is still in use, and valid, when renaming files. This is because the renameObject method on the media source will also fail if the target file already exists, so it makes sense to keep that there with the nice error.)

This specifically does **not** introduce new behaviour, like creating different file names if a file already exists, transliteration, etc. There's definitely room for improvement, and extras can also fill that gap by hooking into `OnFileManagerBeforeUpload` to rename the file/change the target directory/etc. It would be nice to incorporate some more standard behaviour in 3.0, tho given the amount of discussion and conflicting views surrounding the topic it may be desirable to leave that to extras for more flexibility. 

### How to test

- Re-run the build (`php _build/transport.core.php`) and run setup to create the new setting.
- Upload a file through the tree/browser so that it exists.
- Toggle the `upload_check_exists` setting between yes and no and identify that it properly refuses or allows the same file to be uploaded. 

### Related issue(s)/PR(s)

Discussion in #15232 and https://community.modx.com/t/version-2-8-is-out-easy-replacement-of-files-pdf-images-is-axed/3166/12

### Special thanks ❤️ 

Thanks for the [400 trees](https://ecologi.com/modmore) 🌴

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1901,6 +1901,15 @@ $settings['unauthorized_page']->fromArray(array (
   'area' => 'site',
   'editedon' => null,
 ), '', true, true);
+$settings['upload_check_exists']= $xpdo->newObject('modSystemSetting');
+$settings['upload_check_exists']->fromArray(array (
+  'key' => 'upload_check_exists',
+  'value' => '1',
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'file',
+  'editedon' => null,
+), '', true, true);
 $settings['upload_files']= $xpdo->newObject('modSystemSetting');
 $settings['upload_files']->fromArray(array (
   'key' => 'upload_files',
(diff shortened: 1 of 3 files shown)


rthrash:

https://community.modx.com/t/version-2-8-is-out-easy-replacement-of-files-pdf-images-is-axed/3166/17

Issue #15283, opened 2020-10-07
Setting the resource [[*alias_visible]] is disabled if you change the template when creating a resource before saving the resource


Ruslan-Aleev:
### Summary
The resource setting `[[*alias_visible]]` is disabled **if you change the template when creating a resource BEFORE saving the resource**.

### Step to reproduce
1) Create a resource, but **not save the resource**.
2) Change the resource template and save the resource.
3) Open the "Settings" tab and look at `[[*alias_visible]]`.

### Observed behavior
![res_alias](https://user-images.githubusercontent.com/12523676/95343234-f7db0080-08c0-11eb-97af-2757ec5956dc.gif)

### Expected behavior
Setting `[[*alias_visible]]` shouldn't change.

### Environment
MODX 2.x, 3.x

sergant210:
The same for "empty cache" checkbox.

Ibochkarev:
Fixed #15289 

@modxbot close

PR #15289, opened 2020-10-08
Fixed incorrect behavior for resource form checkboxes "alias_visible" and "syncsite"

sergant210:
### What does it do?
Fixed incorrect behavior for resource form checkboxes "alias_visible" and "syncsite"

### Why is it needed?
#15283.

### Related issue(s)/PR(s)
#15283.

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -117,6 +117,7 @@ class ResourceCreateManagerController extends ResourceManagerController {
                 'searchable' => $this->context->getOption('search_default', 1, $this->modx->_userConfig),
                 'cacheable' => $this->context->getOption('cache_default', 1, $this->modx->_userConfig),
                 'syncsite' => $this->context->getOption('syncsite_default', 1, $this->modx->_userConfig),
+                'alias_visible' => $this->context->getOption('alias_visible_default', 1, $this->modx->_userConfig),
             ));
 
             // Allow certain fields to be prefilled from the OnDocFormRender plugin event
@@ -140,7 +141,6 @@ class ResourceCreateManagerController extends ResourceManagerController {
         } else {
             $this->resourceArray = array_merge($this->resourceArray, $reloadData);
             $this->resourceArray['resourceGroups'] = array();
-            $this->resourceArray['syncsite'] = true;
             $this->resourceArray['resource_groups'] = $this->modx->getOption('resource_groups',
                 $this->resourceArray, array());
             $this->resourceArray['resource_groups'] = is_array($this->resourceArray['resource_groups']) ?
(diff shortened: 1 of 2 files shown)


JoshuaLuckers (commented):
Please don't reformat the code (in the 2.x branch), it creates unnecessary noice in the diffs.

sergant210:
@JoshuaLuckers Got it.

JoshuaLuckers:
> @JoshuaLuckers Got it.

When I read it back now, it comes across as a bit rude. I think it's good to have a code base that's formatted properly based on modern standards.
To keep it consistent across the code base it's better to keep the current formatting for 2.x.

opengeek (changes requested):
This still contains the reformatted code. If you can submit just the changes necessary to correct the behavior for these checkboxes without reformatting and without unrelated changes, we can get this incorporated for the next patch release.

manager/controllers/default/resource/create.class.php
```diff
@@ -214,18 +222,10 @@ public function getDefaultTemplate() {
...
-                        if (!empty($siblings)) {
-                            foreach ($siblings as $sibling){
-                                $defaultTemplate = $sibling->get('template');
-                            }
-                        }else{
-                            if (!empty($this->parent->id))
-                                $defaultTemplate = $this->parent->get('template');
-                        }
+                        $lastSibling = $this->modx->getObject('modResource', $c);
+                        $defaultTemplate = !empty($lastSibling) ? $lastSibling->get('template') : $this->parent->get('template');
```
Jako: You have missed the `!empty($this->parent->id)` check in the else branch. Is it safe to remove that?

manager/controllers/default/resource/update.class.php
```diff
@@ -114,21 +123,23 @@ public function process(array $scriptProperties = array()) {
...
-        $this->resourceArray['hidemenu'] = intval($this->resourceArray['hidemenu']) == 1 ? true : false;
-        $this->resourceArray['isfolder'] = intval($this->resourceArray['isfolder']) == 1 ? true : false;
-        $this->resourceArray['richtext'] = intval($this->resourceArray['richtext']) == 1 ? true : false;
-        $this->resourceArray['searchable'] = intval($this->resourceArray['searchable']) == 1 ? true : false;
-        $this->resourceArray['cacheable'] = intval($this->resourceArray['cacheable']) == 1 ? true : false;
-        $this->resourceArray['deleted'] = intval($this->resourceArray['deleted']) == 1 ? true : false;
-        $this->resourceArray['uri_override'] = intval($this->resourceArray['uri_override']) == 1 ? true : false;
+        $this->resourceArray = array_merge($this->resourceArray, $overridden);
+
+        $this->resourceArray['published'] = (int)$this->resourceArray['published'] == 1;
```
Jako: I am not sure what would be executed first in that code: the int casting or the comparison. So i would prefer:
`$this->resourceArray['syncsite'] = intval($this->resourceArray['xxx']) == 1`

Jako (changes requested):

manager/controllers/default/resource/create.class.php
sergant210: @Jako Look at the parent `if` block and everything will become clear.

manager/controllers/default/resource/update.class.php
sergant210: What do you mean? Of course int casting! Besides `(int)` is up to 50% faster then `intval()`.

manager/controllers/default/resource/create.class.php
Jako: Then there is a bug or a glitch in the code. In root an automatic template assignment is not possible for siblings.

sergant210: That's right. 
If you look at the controller code more carefully you will have a lot of questions.

The most harmless question
```
// Default id is 1
$this->modx->getOption('default_template', null, 1)
...
// But here the default id is already 0
$this->context->getOption('default_template', 0, $this->modx->_userConfig);
```
The code of the `getDefaultTemplate` method is strange. What is the first line for? And so on.

sergant210:
PR is changed.

manager/controllers/default/resource/update.class.php
JoshuaLuckers: I would go for a stricter check but casting to int this way looks good to me.

JoshuaLuckers (approved):

Jako (approved):

PR #15294, opened 2020-10-08
Update MODX logo and icon

opengeek:
### What does it do?
Updates the MODX logo and icon with those from the 2020 refresh.

NOTE: This removes the old png versions of the icon and logo along with the corresponding references in the stylesheets, as I believe these are no longer needed.

### Why is it needed?
To bring the logo and icon up to date.

### How to test

- Run `grunt build` in `_build/templates/default/`
- Clear browser caches
- View the manager UI to see the new icon
- View the help page in the manager UI to see the new logo

### Related issue(s)/PR(s)
n/a


README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://modx.com/">
-    <img alt="MODX Revolution" src="https://modx.com/assets/design/css/images/modx-logo-color.svg" width="180" />
+    <img alt="MODX Revolution" src="https://modx.com/assets/themes/manana/images/logo.svg" width="180" />
```
JoshuaLuckers: Would it be possible to have the logo available in the root like the favicon? 
This way we don't need to change the URL whenever the theme changes.

Ruslan-Aleev:
@opengeek can you add a favicon as well? It was also useful to have a new favicon :)
How was it done here - https://github.com/modxcms/revolution/pull/14324 But there the update script needs to be added...

README.md
alroniks: It will be a mess. And if it is in the root, it may be deleted accidentally by users. I don't like solve such kind of issues. And another case – people may remove it but to blame the MODX.

alroniks: Hm, it is a path on the modx.com website. 
What if in this case just use a link to the file in the repo? Like this?
https://raw.githubusercontent.com/modxcms/revolution/2.x/manager/templates/default/images/modx-icon-color.svg -->

opengeek: I went with the link to the file in the GitHub repository.

opengeek:
> @opengeek can you add a favicon as well? It was also useful to have a new favicon :)
> How was it done here - #14324 But there the update script needs to be added...

There is no favicon in the 2.x branch and I don't want to add it with this PR which is intended to simply update the existing logo and icon.

alroniks (approved):

Jako (approved):

Jako:
We have to change the favicon for 3.x (https://github.com/modxcms/revolution/blob/3.x/manager/favicon.ico)

Ibochkarev (approved):

Ibochkarev:
> https://github.com/modxcms/revolution/blob/3.x/manager/favicon.ico

@opengeek You can prepare favicon 32*32 px?

rthrash:
I'll do that and attach here. Do we want to do the favicon "right" including windows metro UI version, etc?

opengeek:
> We have to change the favicon for 3.x (https://github.com/modxcms/revolution/blob/3.x/manager/favicon.ico)

This is not for 3.x — this is only for 2.x.

opengeek:
> I'll do that and attach here. Do we want to do the favicon "right" including windows metro UI version, etc?

This is not the right place for this discussion. I will prepare a separate PR for 3.x and that can be included there. Please send the favicon to me.

Ruslan-Aleev:
@opengeek You forgot about the logo in the "Base Template" for MODX-extras:

logo_old

PR #15297, opened 2020-10-12
Fix readme logo to full logo, not just icon

rthrash:
### What does it do?
Keeps the logo in the readme.md file consistent with previous releases, only updated.

### Why is it needed?
The logo was updated from the previous version to the icon-only.

### How to test
View: https://github.com/rthrash/revolution

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://modx.com/">
-    <img alt="MODX Revolution" src="https://raw.githubusercontent.com/modxcms/revolution/2.x/manager/templates/default/images/modx-icon-color.svg" width="180" />
+    <img alt="MODX Revolution" src="https://raw.githubusercontent.com/modxcms/revolution/2.x/manager/templates/default/images/modx-logo-color.svg" width="180" />
   </a>
 </p>
 <h1 align="center">


opengeek:
I made this change on purpose. We don't need the full logo as it is redundant with the title below it.

rthrash:
Then I would suggest to shrink it down, a lot.

opengeek:
> Then I would suggest to shrink it down, a lot.

I kind of like it that prominent. What size do you think it should be? 120? 100? 80?

rthrash:
LOL … I clearly think it should be the full logo. But if it must be icon-only, about 100 wouldn't look so jarringly huge.

JoshuaLuckers:
Icon only looks much better

README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://modx.com/">
-    <img alt="MODX Revolution" src="https://raw.githubusercontent.com/modxcms/revolution/2.x/manager/templates/default/images/modx-icon-color.svg" width="180" />
+    <img alt="MODX Revolution" src="https://raw.githubusercontent.com/modxcms/revolution/2.x/manager/templates/default/images/modx-logo-color.svg" width="180" />
```
JoshuaLuckers, line 3:
```suggestion
    <img alt="MODX Revolution" src="https://raw.githubusercontent.com/modxcms/revolution/2.x/manager/templates/default/images/modx-icon-color.svg" width="100" />
```

PR #15312, opened 2020-10-28
Remove flash based xtypes

Jako:
### What does it do?
Remove flash based xtypes

### Why is it needed?
It could be possible, that sites running any Flash will be actively blocked by Chrome.

### How to test
The manager works with this changes and the xtypes removed are not used in the core.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13199#issuecomment-717908136


Jako:
This PR could be backported/copied to 2.x

JoshuaLuckers:
What if an extra relies on these?

Jako:
Then the extras have a different issue, because browser based flash is EOL. 

Maybe we could add the removed parts to an extra file.

Ruslan-Aleev:
Maybe in the future it makes sense to transfer/remove other components that are not used?
If there are many of them, then by removing them we will lighten the size of the distribution, we can avoid potential errors in the future (with flash, for example) and even update extJS components separately.

Jako:
http://swfobject.googlecode.com/svn/trunk/swfobject/expressInstall.swf and http://yui.yahooapis.com/2.8.2/build/charts/assets/charts.swf don't exist anymore. So the Ext.chart things could be removed. They can't be used anymore and should currently throw an error.

Jako:
The FileApi code has a fallback to flash, but the referenced files are not available anymore. It could be disabled quite easy.

JoshuaLuckers (approved):

manager/assets/modext/util/utilities.js
```diff
@@ -421,28 +421,8 @@ Ext.applyIf(Ext.form.Field,{
...
-                }
-                document.getElementById(flashcopier).innerHTML = '';
-                var divinfo = '<embed src="' + MODx.config.manager_url
-                    + 'assets/modext/_clipboard.swf" FlashVars="clipboard='
-                    + MODx.util.Clipboard.escape(text)
-                    + '" width="0" height="0" type="application/x-shockwave-flash"></embed>';
-                document.getElementById(flashcopier).innerHTML = divinfo;
-            }
+        copy: function(text){
+            console.log('MODx.util.Clipboard was deactivated!');
```
JoshuaLuckers: I suggest removing this component completely. If there are good reasons to keep it we should log an error instead.

JoshuaLuckers (changes requested):

manager/assets/modext/util/utilities.js
Jako: A github code search for MODx.util.Clipboard shows 459 results, most in the utilities.js file or in some cache files. So I think that the component is no longer in use.

Ibochkarev (approved):

JoshuaLuckers (approved):

PR #15315, opened 2020-11-03
Fix uninstalling package error message

sergant210:
### What does it do?
Fixes not passing property to a lexicon string & an error when uninstall fails.

### Why is it needed?
If package uninstall fails, user gets a message in the console:
```
Error uninstalling package with signature: [[+signature]]
```

### DIFF
--- a/core/model/modx/processors/workspace/packages/uninstall.class.php
+++ b/core/model/modx/processors/workspace/packages/uninstall.class.php
@@ -62,7 +62,7 @@ class modPackageUninstallProcessor extends modProcessor {
         );
 
         if ($this->package->uninstall($options) == false) {
-            return $this->failure(sprintf($this->modx->lexicon('package_err_uninstall'),$this->package->getPrimaryKey()));
+            return $this->failure(sprintf($this->modx->lexicon('package_err_uninstall',  array('signature' => $this->package->get('signature'))), $this->package->getPrimaryKey()));
         }
 
         $this->modx->log(modX::LOG_LEVEL_WARN,$this->modx->lexicon('package_uninstall_info_success',array('signature' => $this->package->get('signature'))));


core/model/modx/processors/workspace/packages/uninstall.class.php
```diff
@@ -62,7 +62,7 @@ public function process() {
         );
 
         if ($this->package->uninstall($options) == false) {
-            return $this->failure(sprintf($this->modx->lexicon('package_err_uninstall'),$this->package->getPrimaryKey()));
+            return $this->failure(sprintf($this->modx->lexicon('package_err_uninstall',  array('signature' => $this->package->get('signature'))), $this->package->getPrimaryKey()));
```
alroniks, line 65: Let's use short array syntax for new code at least.

alroniks (changes requested):

core/model/modx/processors/workspace/packages/uninstall.class.php
sergant210: You know, this is a controversial moment. Last time they explained to me that I need to use the current syntax. And I had to redo my changes.

opengeek: Let's just match the current syntax used in that file and we can worry about short array syntax in the 3.x branch.

opengeek (approved):

Ruslan-Aleev:
@opengeek @Mark-H @JoshuaLuckers @Jako @Alroniks Guys there needs a label `PR required for 3.x` for MODX 3, otherwise some of the changes will be lost and have to be done from scratch for the 3.x branch again.
Does it bother only me?
As soon as the PR is merged for 3.x as well, we just remove that label from the PR and that's it, the fixes won't be lost.

PR #15147, opened 2020-07-02
Added missing lexicon for 'refresh_media_sources'

Ruslan-Aleev:
### What does it do?
Added the missing lexicon for `refresh_media_sources` in the console when clearing the cache.

refresh_ms

### Related issue(s)/PR(s)
N/A

PR #15336, opened 2020-11-20
Revise cache clearing lexicons

Ruslan-Aleev:
### What does it do?
The same as in PR https://github.com/modxcms/revolution/pull/15147

Fixed lexicons for clearing cache:
- Brought to one format;
- Added missing lexicon for `refresh_media_sources`.

p.s. For 3.x branch, these lexicons should also be taken from 2.x to avoid unnecessary duplication.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15147


Ibochkarev (approved):

opengeek:
These absolutely should not be all changed to "clearing". This is why the method is now called refresh, not clear, because the cache is not cleared, which implies things are just deleted. The cache is refreshed. In some cases, this does lead to the deletion (or clearing) of certain cache partitions, but in many cases, especially regarding contexts, the cache is regenerated and replaced—never cleared.

Ruslan-Aleev:
In my opinion, with regard to the cache, "clearing" and "regenerating" are the same thing. The cache will not be empty (otherwise it makes no sense), therefore "clearing the cache" means deleting data + filling with new ones = "regenerating".

But, so that there are no questions, maybe it is worth replacing the lexicons of the cache with `refreshing` instead of "clearing" and "regenerating"?

opengeek:
Personally, I prefer to have concise words used. Regenerating is perfect when they are regenerated and replaced. Cleared is perfect when they are just deleted and new entries added on demand.

rthrash:
Refreshing is how it makes a person feel when something good happens or something tasty is consumed.  Or if it has a bit of cleaning or updating, it's refreshed, but the original base remains with some extra paint and polish. 

Regenerate is a much more accurate word. To regenerate is to cause it to rebuild from the original context/source. 

To clear is to remove something.

JoshuaLuckers (approved):

PR #15346, opened 2020-12-04
JsonGrid fixes

Jako:
### What does it do?
- Change the record value of the currently edited field on keyup
- Don't use a menu in the action column for deleting a record.
- Remove the column menu for the grid.

### Why is it needed?
- When the user clicks on save before the currently edited field loses the focus, the changes in the current field are not saved.
- The menu in the action column does not make sense, since there is only one delete entry in that menu.
- The column menu does not make sense, since the columns are not sortable (the grid is sortable by drag & drop).

### How to test
Create a new system setting with the field type JSON Grid and save that setting. The JSON grid is visible during editing the the setting again in the edit window at the bottom.

### Related issue(s)/PR(s)
None known


JoshuaLuckers (approved):

theboxer (changes requested):
I still see the context menu on the JsonGrid. I believe you'll have to remove the getMenu method to nuke it.

Jako:
Thanks, the description of the fix was a bit wrong. I meant the menu in the action column, not the context menu.

theboxer (approved):

Issue #15323, opened 2020-11-06
Overly Agressive Template & Plugin Name Validation


matdave:
### Summary
The template & plugin name validation introduced in #15146 appears to be overly aggressive and causing issues on older sites. The validation appears to be copy/pasted from chunk and snippet name validation. However, I believe that validation was specifically designed because those elements are "callable" for the templating engine. It is preventing saving of templates on existing sites that use common characters such as colons or apostrophe's. 

This also critically breaks Teleport as a template or plugin with a previously valid name can no longer be injected into a site. 

### Step to reproduce
If you edit a site in the SQL to change the name to something like "Categegory: Interior" you will not be able to save the template within MODX without changing the name. 

### Observed behavior
This creates confusing behavior as existing Templates are suddenly not capable of being saved without changing the name. It also breaks the ability to inject sites using Teleport if these characters currently exist on their templates or plugins. 

### Expected behavior
Standard characters that do not pose an issue to XSS or break the site in anyway should be allowed in Template Names and Plugins.

### Related issue(s)/PR(s)
Introduced in https://github.com/modxcms/revolution/pull/15146

### Environment
MODX 2.8.1

garryn:
Took a look at the validation regex, I'm posting an amendment below for discussion that expands it to include common punctuation I've encountered in Template names (this would also be fine in my opinion to be used for Plugin names):

```
^(?!\s)[a-zA-Z0-9\x23-\x2f\x3a\x5b-\x5d\x7f-\xff-_\s]+(?<!\s)$
```

(To see what is included in the various `\x` ranges, this site is useful: https://www.codetable.net/asciikeycodes)

JoshuaLuckers:
> If you edit a site in the SQL to change the name to something like "Categegory: Interior" you will not be able to save the template within MODX without changing the name.

A workaround that's not intended to be one: if the last character is a whitespace you are able to save the template just fine.

PR #15348, opened 2020-12-11
Adding more character options to plugins and templates

matdave:
### What does it do?
Expands the qualifying character patterns for Templates and Plugins names to prevent breaking sites during upgrade. 

### Why is it needed?
PR #15146 reduced the available characters in these non-callable element names. This breaks older sites that used common character patterns in their names. 

### How to test
Attempt to add additional strings to a plugin or template names. It still blocks callable code, but allows commonly used special characters. 

### Related issue(s)/PR(s)
issue #15323

_Edited: clarified these are referring to the names of the elements_

### DIFF
--- a/core/model/modx/mysql/modplugin.map.inc.php
+++ b/core/model/modx/mysql/modplugin.map.inc.php
@@ -184,7 +184,7 @@ $xpdo_meta_map['modPlugin']= array (
         'invalid' =>
         array (
           'type' => 'preg_match',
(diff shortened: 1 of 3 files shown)


opengeek (changes requested):
Changes to the map files must be made in the schema and generated to the map files. These changes will otherwise be overwritten the next time the schema is parsed.

matdave:
D'oh... done

PR #15350, opened 2020-12-12
Wrong term used

Jako:
### What does it do?
Change the term document to resource in the lexicon.

### Why is it needed?
The term documents is wrong in this lexicon. TVs are assigned to resources in MODX.

See the Basic Concepts page in the docs: https://docs.modx.com/current/en/getting-started/what-is-modx#basic-concepts

### How to test
Not needed

### Related issue(s)/PR(s)
None known.


Ibochkarev (approved):

Jako:
This seems to be not the only place where the term 'document' is used instead of 'resource'.

Jako:
Sorry, i should not edit the files in this repository. Reopened in #15351

Issue #15374, opened 2021-02-05
MODX 3 modUser class joinGroup() crashes is $_SESSION is not set


BobRay:
### Summary
Quick summary what's this issue about.

Calling $user->joinGroup() with no $_SESSION set crashes.

Here's the offending code (line 716 of core/model/modx/moduser.class.php):

```
                     unset($_SESSION["modx.user.{$this->get('id')}.userGroupNames"],
                     $_SESSION["modx.user.{$this->get('id')}.userGroups"]);
```

This breaks acceptance tests that run in the MODX Manager.

### Environment
MODX version, MODX 2.8.1

BobRay:
I would propose the fix of adding this line just above the unset call:
```
$_SESSION = isset($_SESSION)? $_SESSION : array();
```

sergant210:
`$_SESSION["modx.user.{$this->get('id')}.userGroupNames"` - this code is legacy. It must be removed. All permission logic must be removed from the session.

Ibochkarev:
Fixed #15376

PR #15376, opened 2021-02-06
Update modUser.php

BobRay:
Prevent PHP error when $_SESSION is not set (e.g., in a unit or acceptance test).

### What does it do?
Sanity check for $_SESSION

### Why is it needed?
Original code throws a PHP fatal error if $_SESSION is not set.

### Related issue(s)/PR(s)
Issue: #15374

### DIFF
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -166,7 +166,7 @@ class modUser extends modPrincipal
         }
         if ($this->_attributes === null || $reload) {
             $this->_attributes = [];
-            $_SESSION = isset($_SESSION)? $_SESSION : [];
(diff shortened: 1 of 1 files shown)


BobRay:
@cla-bot check

core/src/Revolution/modUser.php
```diff
@@ -166,6 +166,7 @@ public function loadAttributes($target, $context = '', $reload = false)
         }
         if ($this->_attributes === null || $reload) {
             $this->_attributes = [];
+            $_SESSION = isset($_SESSION)? $_SESSION : [];
```
JoshuaLuckers:
```suggestion
            $_SESSION = isset($_SESSION) ? $_SESSION : [];
```

JoshuaLuckers (approved):

Issue #15373, opened 2021-02-05
MODX 3 lexicon->load() crashes $_SESSION not set


BobRay:
### Summary
I no $_SESSION array is set, lexicon->load crashes.

### Step to reproduce
Call lexicon->load() with no $_SESSION set and $modx->context set to 'mgr'.

Here is the offending code (around line 189 of core/src/Revolution/modLexicon.php)

```
if ($this->modx->context && $this->modx->context->get('key') == 'mgr') {
            $defaultLanguage = $this->modx->getOption('manager_language', $_SESSION,
                $this->modx->getOption('cultureKey', null, 'en'));
        } else {
            $defaultLanguage = $this->modx->getOption('cultureKey', null, 'en');
        }
```

This will break testing code. (i.e. Unit, Integration, or Acceptance test) that call `$lexicon->load();`. 

There are a number ways to fix this. This is one (just above the first getOption() call), but I'm not sure it's the best:

```
if (! isset($_SESSION)) {
    $_SESSION = null;
}
```

### Environment
MODX 3, Windows 10,  apache

Mark-H:
What do you mean exactly by "crashes"? Is it causing an error or unexpected behavior downstream?

BobRay:
Sorry, attempting to access a member of an array that doesn't exist throws a PHP error: Unknown variable `$_SESSION`.

BobRay:
I think a better solution is to put this above the first getOption() call:

```
$_SESSION = isset($_SESSION)? $_SESSION : array();
```

Ibochkarev:
Fixed #15377

PR #15377, opened 2021-02-06
Update modLexicon.php

BobRay:
Sanity check to prevent PHP error if $_SESSION is not set

### What does it do?
Make sure $_SESSION variable exists before it's referenced

### Why is it needed?
Existing code breaks automated acceptance tests (and possibly unit tests) where $_SESSION isn't necessarily set.

### Related issue(s)/PR(s)
Issue: #15373

### DIFF
--- a/core/src/Revolution/modLexicon.php
+++ b/core/src/Revolution/modLexicon.php
@@ -186,8 +186,7 @@ class modLexicon
         $topics = func_get_args(); /* allow for dynamic number of lexicons to load */
 
         if ($this->modx->context && $this->modx->context->get('key') == 'mgr') {
-            $defaultLanguage = $this->modx->getOption('manager_language', $_SESSION,
-                $this->modx->getOption('cultureKey', null, 'en'));
+            $defaultLanguage = $this->modx->getOption('manager_language', $_SESSION ?? [], $this->modx->getOption('cultureKey', null, 'en'));
         } else {
             $defaultLanguage = $this->modx->getOption('cultureKey', null, 'en');
         }


BobRay:
@cla-bot check

JoshuaLuckers (approved):

JoshuaLuckers (approved):

core/src/Revolution/modLexicon.php
```diff
@@ -186,6 +186,7 @@ public function load()
         $topics = func_get_args(); /* allow for dynamic number of lexicons to load */
 
         if ($this->modx->context && $this->modx->context->get('key') == 'mgr') {
+            $_SESSION = isset($_SESSION) ? $_SESSION : [];
```
alroniks: As we support PHP 7.0+ for MODX 3, I suggest using null coalescing operator.

core/src/Revolution/modLexicon.php
```diff
@@ -186,6 +186,7 @@ public function load()
...
             $defaultLanguage = $this->modx->getOption('manager_language', $_SESSION,
```
alroniks:
```suggestion
            $defaultLanguage = $this->modx->getOption('manager_language', $_SESSION ?? [],
```

And check above is not needed.

alroniks (dismissed):

BobRay:
Using the null coalescing operator is fine with me.

opengeek:
I modified this to use the null coalescing operator rather than setting the global variable $_SESSION. Though chances are slim this would ever have a real-world consequence, setting the value of a PHP superglobal to avoid an error seemed like the wrong approach.

BobRay:
> I modified this to use the null coalescing operator rather than setting the global variable $_SESSION. Though chances are slim this would ever have a real-world consequence, setting the value of a PHP superglobal to avoid an error seemed like the wrong approach.

It regularly crashed Codeception acceptance tests of the MODX Manager, so I'm glad it's fixed.  Using the null coalescing operator is a much better approach than mine.

PR #15387, opened 2021-02-10
Revert "Fixed work with menu items in the main menu"

### LINKED ISSUE #15386: MODX3: Menu item sorting does not work
### Summary
In https://github.com/modxcms/revolution/pull/15230 I added a parameter `ddAppendOnly: true` to the script, this prevented the menu items from being dragged behind the main menu areas (Main Navigation and User Navigation).

It was necessary because when dragging, the menu item could disappear from the main menu, which is not correct.

menu_dd_1

**But this parameter, apparently, also participates in the sorting of menu items.**

### Step to reproduce
Try changing the sorting in the menu settings.

### Expected behavior
In general, a solution is needed that would prohibit dragging the menu item outside of the main areas (Main Navigation and User Navigation), but would also enable drag-and-drop sorting.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15230

### Environment
MODX 3

opengeek:
Reverts #15230 and resolves #15386

### DIFF
--- a/manager/assets/modext/widgets/system/modx.tree.menu.js
+++ b/manager/assets/modext/widgets/system/modx.tree.menu.js
@@ -12,11 +12,10 @@ MODx.tree.Menu = function(config) {
         rootIconCls: 'icon-navicon'
         ,rootId: 'n_'
         ,rootName: _('menu_top')
-        ,rootVisible: false
+        ,rootVisible: true
         ,expandFirst: true
         ,enableDrag: true
         ,enableDrop: true
-        ,ddAppendOnly: true
         ,url: MODx.config.connector_url
         ,action: 'System/Menu/GetNodes'
         ,sortAction: 'System/Menu/Sort'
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
Here it is worth removing only this line - `ddAppendOnly: true`, the rest of the fixes should be left.

manager/assets/modext/widgets/system/modx.tree.menu.js
```diff
@@ -12,11 +12,10 @@ MODx.tree.Menu = function(config) {
         rootIconCls: 'icon-navicon'
         ,rootId: 'n_'
         ,rootName: _('menu_top')
-        ,rootVisible: false
+        ,rootVisible: true
         ,expandFirst: true
         ,enableDrag: true
         ,enableDrop: true
-        ,ddAppendOnly: true
```
Ruslan-Aleev, line 19: Only this line should be removed here.

Ruslan-Aleev (changes requested):

Issue #14521, opened 2019-04-02
Incorrect behavior of the menu tree in the "Menu" section


Ruslan-Aleev:
### Summary
1) If I move the item to the main section "Top Menu", the item will simply disappear from the menu in the manager panel. You can make a mistake and do not understand why the item disappeared, but the item is present in the menu section (the nesting of the items is not clear until all the menus are minimized).

![menu-bug](https://user-images.githubusercontent.com/12523676/55397391-2ef78580-5557-11e9-9973-695e7a7eea27.gif)

This section "Top Menu" should be renamed to "Main Menu" and **prohibit dragging items into it or create an menu item with an empty parent**, as it seems to me.

There is another related bug. **If you delete the main categories (Top Navigation and User Navigation), then the new created items can no longer be displayed in the manager panel**.

2) There is also missing icon in the tree, by analogy with another tree, for example, the resource tree.

![top-menu-1](https://user-images.githubusercontent.com/12523676/55397884-5438c380-5558-11e9-941b-b7d35a1e814b.png)

---

It may be worth hiding the root category (Top menu) and leaving only the menu categories (Top navigation and User navigation) so that the menu item cannot be added by dragging to the root category. But then it is worth removing the empty parent (root category) when creating an item.

![top-menu-2](https://user-images.githubusercontent.com/12523676/55497417-626b0a80-5652-11e9-9106-5e1ad8bcf21f.png)

### Environment
MODX 2.7.1.

JoshuaLuckers:
Related issue: #14220

Ruslan-Aleev:
Also, if you move menu items from "Main Navigation" to "User Navigation", then the icon item is not displayed.

![menu_bug_3](https://user-images.githubusercontent.com/12523676/86520679-cd0b8700-be4f-11ea-8bc9-3c67f94f36e8.png)

JoshuaLuckers:
The description of the menu item shouldn't be displayed when it's a direct child of the main navigation.
<img width="448" alt="Screen Shot 2020-07-07 at 09 43 02" src="https://user-images.githubusercontent.com/1268085/86739568-44027480-c036-11ea-8968-9d3f7bb7afab.png">

Ruslan-Aleev:
> The description of the menu item shouldn't be displayed when it's a direct child of the main navigation.
[...]

PR #15388, opened 2021-02-11
Menu improvements

### LINKED ISSUE #15386: MODX3: Menu item sorting does not work

theboxer:
### What does it do?
- Hides the menu root (Main Menu) - we don't really need it
- Prevents dropping menu items outside of `Main Navigation` or `User Navigation`
- Prevents dragging/editing/deleting `Main Navigation` and `User Navigation`
- Adds context menu `Create Menu` on all nodes

### Why is it needed?
To prevent menu items from disappearing.

### How to test
Play with menus ;)

### Related issue(s)/PR(s)
#15386, #15230, #14521

### DIFF
--- a/core/src/Revolution/Processors/System/Menu/GetNodes.php
+++ b/core/src/Revolution/Processors/System/Menu/GetNodes.php
@@ -102,6 +102,11 @@ class GetNodes extends GetListProcessor
             'data' => $object->toArray(),
             'qtip' => strip_tags($desc),
         ];
+
+        if (empty($object->parent)) {
+            $objectArray['draggable'] = false;
+        }
+
         if ($object->get('childrenCount') < 1) {
             // Workaround for leaf record not to display "arrows"
             $objectArray['loaded'] = true;
--- a/core/src/Revolution/Processors/System/Menu/GetNodes.php
+++ b/core/src/Revolution/Processors/System/Menu/GetNodes.php
@@ -101,11 +101,8 @@ class GetNodes extends GetListProcessor
             'leaf' => false,
             'data' => $object->toArray(),
             'qtip' => strip_tags($desc),
+            'draggable' => !empty($object->parent),
         ];
-
-        if (empty($object->parent)) {
-            $objectArray['draggable'] = false;
-        }
 
         if ($object->get('childrenCount') < 1) {
             // Workaround for leaf record not to display "arrows"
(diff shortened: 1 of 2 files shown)


Ruslan-Aleev (approved):
Thanks, it works!

mm

core/src/Revolution/Processors/System/Menu/GetNodes.php
```diff
@@ -102,6 +102,11 @@ public function prepareRow(xPDOObject $object)
             'data' => $object->toArray(),
             'qtip' => strip_tags($desc),
         ];
+
+        if (empty($object->parent)) {
+            $objectArray['draggable'] = false;
+        }
```
mishanthrop: Instead of these three lines, you can write above like this::
```php
'draggable' => !empty($object->parent),
```

manager/assets/modext/widgets/system/modx.tree.menu.js
```diff
@@ -35,6 +34,24 @@ MODx.tree.Menu = function(config) {
 Ext.extend(MODx.tree.Menu, MODx.tree.Tree, {
     windows: {}
 
+    ,_handleDrop: function (dropEvent) {
+        var node = dropEvent.target;
+        if (node.isRoot) return false;
+
+        if ((dropEvent.point === 'above') || dropEvent.point === 'below') {
```
mishanthrop: unnecessary parentheses

```javascript
if (dropEvent.point === 'above' || dropEvent.point === 'below') {
```

Ibochkarev (approved):
I confirm - it works. Thank you =)

PR #15389, opened 2021-02-11
User extended properties - fix rendering value

### LINKED ISSUE #15356: In extended user fields, the <i> tag is added in the attribute tree
### Summary
When editing a user, any attribute can be added on the "Extended Fields" tab, however, an `<i>` tag is added around the attribute value in the attribute tree:

user_fields

### Related issue(s)/PR(s)
N/A

### Environment
MODX 2.x, 3.x

theboxer:
### What does it do?
Add `renderItemText` to the `MODx.orm.Tree` that is returning encoded value. No encoding is needed in the method as it's already encoded before setting the `text` property on the node.

### Why is it needed?
`<i></i>` tags are visible around the value.

### How to test
Add new extended field to the user. The value should show in italic.

### Related issue(s)/PR(s)
#15356

### DIFF
--- a/manager/assets/modext/widgets/core/modx.orm.js
+++ b/manager/assets/modext/widgets/core/modx.orm.js
@@ -143,7 +143,7 @@ Ext.extend(MODx.orm.Tree,Ext.tree.TreePanel,{
                         }
 
                         var n = new Ext.tree.TreeNode({
-                            text: r.name
+                            text: Ext.util.Format.htmlEncode(r.name)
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev (approved):
Thanks, it works!

uf

Ibochkarev (approved):
it works!


https://user-images.githubusercontent.com/2138260/108910337-eae06300-764f-11eb-88e1-934bb1eae030.mov

PR #15395, opened 2021-02-17
Fix chunk/snippet blank page with <script><!--

### LINKED ISSUE #15222: If you create a snippet containing <script><!-- then the snippet editing form stops working
### Summary
If you create a snippet containing `<script><!--` then the snippet editing form stops working.

Of course it is unlikely that such a snippet is needed, **but there may be a more serious problem here**.

snippet-bug

Errors in browser console:

```
ext-all.js:21 Uncaught TypeError: Cannot read property 'dom' of null
    at S.applyToMarkup (ext-all.js:21)
    at S.Ext.Component [as constructor] (ext-all.js:21)
    at new S (ext-base.js:21)
    at Object.create (ext-all.js:21)
    at MODx.Layout.Default.createComponent (ext-all.js:21)
    at MODx.Layout.Default.lookupComponent (ext-all.js:21)
    at MODx.Layout.Default.add (ext-all.js:21)
    at MODx.Layout.Default.<anonymous> (ext-all.js:21)
    at Object.each (ext-base.js:21)
    at MODx.Layout.Default.add (ext-all.js:21)
```

### Step to reproduce
Create snippet containing:
`<script><!--`

### Environment
MODX 2.x >

theboxer:
### What does it do?
It insert's `"-->"` at the end of the `snippetArray`/`chunkArray`.

The tricky question here is why it works. My humble opinion is that this is some kind of a browser issue. The edit snippet/chunk page's controller converts the object to JSON and passes it to `MODx.load` as `record` prop. The JSON itself is valid, but when it's placed in the source code of the page, it somehow breaks. Seems like the browser don't understand that the `<script><!--` is a part of a string and it stops processing any following JS/HTML code.

The issue also only happens when there is more than 1 occurrence of the `<script><!--` code. You can try this by a simple HTML page:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
    <script type="text/javascript">
        var test = {
            "a": "<script><!--",
            "b": "<script><!--",
        };
    </script>
</head>
<body>
HELLO
</body>
</html>
```

This page above won't render anything. If you remove the `b` property from the `test` object, it starts working.

Why it's twice in the snippet/chunk edit page, when we're putting it once in the content field? Snippet has alias for the `snippet` field which holds the value - `content` and they are both dumped to the JSON.

At first I tried removing the `content` from the JSON, as it's not used by the edit page anyways, but then I was able to break it 
with: 
```
<script><!--
<script><!--
````

For a strange reason, appending `-->` to the JSON, makes the browser think that everything is OK and it renders the page correctly. The JS is able to read the whole JSON without an issue.

### Why is it needed?
For unknown reason people tried to save a snippet or chunk with `<script><!--` which is obviously invalid for snippets and for html as well. So if you do this twisted thing, the edit snippet/chunk page will blank after reload.

### How to test
Described here: #15222 + try in chunk as well

### Final note
I'm not sure if we want to merge this, it seems like it fixes the related issue, but it's weird. I've spent so much time on this and I'm not willing to explore more. Feel free to submit other ideas. Resources and templates break as well, but the double script is needed:
```
<script><!--
<script><!--
````

### Related issue(s)/PR(s)
#15222

### DIFF
--- a/manager/controllers/default/element/chunk/update.class.php
+++ b/manager/controllers/default/element/chunk/update.class.php
@@ -107,6 +107,9 @@ class ElementChunkUpdateManagerController extends modManagerController {
         $this->chunkArray['properties'] = $data;
         $this->chunkArray['snippet'] = $this->chunk->getContent();
 
+        /* Don't remove the following line or the edit chunk page blanks when saving <script><!-- GitHub issue: #15222 */
+        $this->chunkArray['__fix_for_15222'] = '-->';
+
         $this->prepareElement();
 
         /* invoke OnRichTextEditorInit event */
--- a/manager/controllers/default/element/snippet/update.class.php
+++ b/manager/controllers/default/element/snippet/update.class.php
@@ -8,7 +8,9 @@
  * files found in the top-level directory of this distribution.
  */
 
+use MODX\Revolution\modCategory;
 use MODX\Revolution\modManagerController;
+use MODX\Revolution\modSnippet;
 use MODX\Revolution\modSystemEvent;
 
 /**
@@ -103,6 +105,9 @@ class ElementSnippetUpdateManagerController extends modManagerController {
         if (strpos(ltrim($this->snippetArray['snippet']),'<?php') !== 0) {
             $this->snippetArray['snippet'] = "<?php\n".$this->snippetArray['snippet'];
         }
+
+        /* Don't remove the following line or the edit snippet page blanks when saving <script><!-- GitHub issue: #15222 */
+        $this->snippetArray['__fix_for_15222'] = '-->';
 
         $this->prepareElement();
 

### CODE REVIEW

**Comment by Jako** [comment]
Would it help to wrap that part in `//<![CDATA[
    ...code...
//]]>
`?

**Comment by theboxer** [comment]
It's already there
```php
$this->addHtml('
        <script type="text/javascript">
        // <![CDATA[
        MODx.onSnipFormRender = "'.$this->onSnipFormRender.'";
        Ext.onReady(function() {
            MODx.load({
                xtype: "modx-page-snippet-update"
                ,id: "'.$this->snippetArray['id'].'"
                ,record: '.$this->modx->toJSON($this->snippetArray).'
            });
        });
        // ]]>
        </script>
');
```

Also by guys from https://stackoverflow.com/questions/7092236/what-is-cdata-in-html it seems like CDATA has no meaning in HTML.

Jako:
If quickedit of those elements still works, I would not merge this.

theboxer:
Quick edit works, because it doesn't load data through controller, but directly via XHR request. But that doesn't mean that blank page is a correct state :), even though  you can go back in history and use quick edit to fix it.

Jako:
For me it would be better to disallow those invalid tags during saving the element and/or create a FAQ entry with:
Blank page on editing, try quickedit and remove `<script><!--`

Ruslan-Aleev:
Thanks for your participation! Interesting solution :)
I would merge it if this fix doesn't add bugs.

Jako:
Yes, thats the issue. You can't be sure, that it does not create a bug. It closes something that in most cases does not have to be closed.

theboxer:
It's not closing anything real, it's adding a string `"-->"` to the JSON data :) It probably just tricks the browser

JoshuaLuckers:
> It's not closing anything real, it's adding a string `"-->"` to the JSON data :) It probably just tricks the browser

The text inside the HTML comment must not start with the string ">", nor start with the string "->", nor contain the strings "", or "--!>", nor end with the string "<!-".


Source: https://html.spec.whatwg.org/multipage/syntax.html#comments

JoshuaLuckers:
I agree with @Jako.

JoshuaLuckers:
At first I agreed with @Jako but then I looked into the details, it doesn't add any data to the chunk itself. It ads a new property `__fix_for_15222` with the value `-->` to the JSON.

That looks harmless to me. But still an interesting browser bug/feature. 

This is the generated javascript:
```javascript
// <![CDATA[
Ext.onReady(function() {
    MODx.load({
        xtype: "modx-page-chunk-update"
        ,chunk: "1"
        ,record: {"id":1,"source":1,"property_preprocess":false,"name":"TestChunk","description":"","editor_type":0,"category":0,"cache_type":0,"snippet":"<script><!-- \n\nThis is a test that breaks without this fix","locked":false,"properties":[],"static":false,"static_file":"","content":"<script><!-- \n\nThis is a test that breaks without this fix","__fix_for_15222":"-->","openTo":"\/"}
    });
});
MODx.onChunkFormRender = "";
MODx.perm.tree_show_element_ids = 1;
MODx.perm.unlock_element_properties = 1;
// ]]>
```

Jako:
@JoshuaLuckers Thanks for the clarification. It looks a bit like this browser issue could be misused to create a blank page everywhere. Can't the code be changed to load the record value by XHR?

JoshuaLuckers:
> @JoshuaLuckers Thanks for the clarification. It looks a bit like this browser issue could be misused to create a blank page everywhere. Can't the code be changed to load the record value by XHR?

I don't think it's that big of an issue to change the way we load our values everywhere this particular input will trigger a parsing error in the browser. 
As soon as the output contains two script tags with the comment open elements the browser will show an empty page. This also includes the front-end. 

If you edit a resource (without rich text) and use the example code as value it will also break the manager and front-end page.

JoshuaLuckers (approved):

Jako (approved):

Mark-H (changes requested):
Sorry, but this is so hacky.. can't this be added into the script tags instead of whatever array may be encoded to JSON?

Mark-H:
Maybe something like this in the addHtml?

```
// <![CDATA[ <!--

// --> ]]> 
```

sottwell:
It already is https://github.com/modxcms/revolution/pull/15395#issuecomment-780636392

Jako:
No, Mark wants to insert additional html comments inside of the CDATA section.

theboxer:
@Mark-H https://github.com/modxcms/revolution/pull/15395#issuecomment-780636392

JoshuaLuckers:
> Maybe something like this in the addHtml?
> 
> ```
> // <![CDATA[  ]]> 
> ```

That means we can't add any HTML comments using `addHtml`. Nested comments are not possible and will probably break the parser.

After thinking about it, the current solution will also break parsers or cause unintended behavior since the browser will still parse the JSON as text and end any HTML comment that was started when encountering the `-->` added in the JSON string.

I hope it makes sense what I try to say here haha.

theboxer:
@JoshuaLuckers no, the proposed solution doesn't break the parser. The JSON is valid, it's not closing not-existing comment. Like I mentioned, it seems to only trick the browser to render. I'm fine closing the PR as it really is a weird & hacky solution.

theboxer:
I'm going to close this one, it's a hacky solution and didn't gain enough support :)

Issue #15339, opened 2020-11-24
New character limit of 191 doesn't match Manager enforcement on descriptions, also breaks some Extra installs


jaredfhealy:
### Summary
I first noticed this issue when going to install QuickApi 1.0.0-beta2. I received an error during the installation.

### Step to reproduce
1. Install the referenced version from the store, or another known Extra with a Description field over 191.
2. See the error displayed

**Note:** I just published a new version that has reduced Description field lengths and some other changes. Depending on timing you may need to download the previous version and install manually.

### Steps to reproduce - related manager issue
1. Go to create a snippet, or any element with a description field
2. Populate any description that is greater than 191 characters but less than the previously allowed 255.
3. Save the record
4. See a modal error: 

![image](https://user-images.githubusercontent.com/9794500/100109084-ba0e4900-2e28-11eb-9f54-4393a78eb5d4.png)

5. Put over 255 characters into the description
6. Manager displays error message based on old character limit of 255

![image](https://user-images.githubusercontent.com/9794500/100109545-2ee18300-2e29-11eb-8427-9aebcee30620.png)

It appears that when the field length change was made, the enforced lengths in the ExtJS components were not updated so hints are displayed with pervious values.

### Observed behavior
When installing any Extra that has over 191 character description on any element, an error is displayed that it is unable to insert.

```
[...]

Issue #15319, opened 2020-11-06
Unable to save snippet with a description that is "too long"


muzzwood:
### Summary
Saving a snippet with a description that is "too long" will fail without explaining why.

### Step to reproduce
Create or update a snippet in the manager with a description that is more than a couple of lines.

### Observed behavior
The snippet does not save and a pop-up appears saying "An error occurred while saving the snippet."
The MODX error log shows the following:
```
/core/xpdo/om/xpdoobject.class.php : 1452) Error 22001 executing statement:
INSERT INTO `modx_site_snippets` (`source`, `property_preprocess`, `name`, `description`, `editor_type`, `category`, `cache_type`, `snippet`, `locked`, `moduleguid`, `static`, `static_file`) VALUES (0, 0, 'snippetName', 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.', 0, 2, 0, 'return \'\';', 0, '', 0, '')
Array
(
    [0] => 22001
    [1] => 1406
    [2] => Data too long for column 'description' at row 1
)
```
and the same error for an UPDATE request.

### Expected behavior
Possible fixes:
- The snippet should save and just truncate the description.
- There should be a more informative error message and form validation highlighting showing where the issue is.
- If the description field isn't indexed, it could be made longer that 191 chars and still be compatible with UTF8mb4 InnoDB.

### Environment
MODX version 2.8.1

sergant210:
Temporary solution - make a plugin for the OnBeforeDocFormSave event and validate the form data.

alroniks:
Fixed

PR #15398, opened 2021-02-19
Switch non-index fields back to 255 char limit

theboxer:
### What does it do?
Switch non-index fields back to 255 char limit (from 191). Adjust JS validations to the same limit as in schema.

### Why is it needed?
JS validations stayed at 255 chars + there's no reason to limit non-index fields to 191 chars.

### Related issue(s)/PR(s)
#15339 #15319

### DIFF
--- a/core/src/Revolution/Registry/Db/sqlsrv/modDbRegisterMessage.php
+++ b/core/src/Revolution/Registry/Db/sqlsrv/modDbRegisterMessage.php
@@ -13,7 +13,7 @@ class modDbRegisterMessage extends \MODX\Revolution\Registry\Db\modDbRegisterMes
         'version' => '3.0',
         'table' => 'register_messages',
         'extends' => 'MODX\\Revolution\\xPDOObject',
-        'fields' =>
+        'fields' => 
         array (
             'topic' => NULL,
             'id' => NULL,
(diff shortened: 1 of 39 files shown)


core/src/Revolution/Registry/Db/sqlsrv/modDbRegisterMessage.php
```diff
@@ -25,171 +25,171 @@ class modDbRegisterMessage extends \MODX\Revolution\Registry\Db\modDbRegisterMes
             'payload' => NULL,
             'kill' => 0,
         ),
-        'fieldMeta' =>
+        'fieldMeta' =>
```
JoshuaLuckers, line 28: Are those spaces generated by xPDO's parse schema script?

theboxer: Most likely, I didn't touch any of these files

opengeek: Yeah, this is how var_export always does this. It appears at some point someone edited these files by hand and their IDE likely trimmed the whitespace automatically (or they ran it through some kind of code formatting).

Ibochkarev (approved):
изображение

But in 3.X we do not display an error to the short description that the maximum length is 255

core/src/Revolution/Registry/Db/sqlsrv/modDbRegisterMessage.php
Ibochkarev: After adding the editorconfig linter, there will be no such cases.

alroniks (approved):

PR #15335, opened 2020-11-18
PHP 8 Compatibility

opengeek:
### What does it do?
* Resolves the change to PDO's default error mode, which is now ERRMODE_EXCEPTION in PHP 8
* Conditionally executes the get_magic_quotes_gpc() function on PHP < 7.4 since it was deprecated in that version and removed in PHP 8
* Fixes errors when merging PDO driver options in xPDOConnection objects as well as when loading modX configuration options

### Why is it needed?
To make 2.x compatible with PHP 8.

### How to test
Build, run, and test on PHP 8.

### Related issue(s)/PR(s)
n/a

PR #15401, opened 2021-02-20
Fix fatal error: "Uncaught Error: Call to undefined function get_magic_quotes_gpc()" on PHP8

Mark-H:
### What does it do?

Fixes an issue on PHP8 related to calling a method that no longer exists, and was deprecated in PHP 7.4.

### Why is it needed?

PHP 8 compatibility.

### How to test

Run on PHP 8, see it breaks without this patch.

### Related issue(s)/PR(s)

This change was originally included in #15335, but somehow didn't end up in the 2.x branch. Perhaps @opengeek wants to re-do that merge but at any rate without this fix PHP8 compatibility is impossible so I figured I'd dust off my ability to create PRs.

### DIFF
--- a/core/model/modx/modx.class.php
+++ b/core/model/modx/modx.class.php
@@ -339,7 +339,7 @@ class modX extends xPDO {
                         $iteration++;
                     }
                 }
-                if (get_magic_quotes_gpc()) {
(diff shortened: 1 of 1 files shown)


opengeek:
The merge was really messy and confusing for some reason. Thanks for cleaning that up for me.

Issue #15344, opened 2020-12-02
Encoding error in json response for ACL messages for non-english languages


Ruslan-Aleev:
### Summary
ACL can give an incorrect error message, for example, this.

![acl_error_en](https://user-images.githubusercontent.com/12523676/100891630-84bdc700-34ca-11eb-988d-1d496fc8d6b0.png)

It is already not very clear to the end user what exactly the ACL error is.

But if the manager panel is not in english, then json displays messages in incorrect encoding, i.e. it is not even clear that the error is related to ACL.

![acl_error_ru](https://user-images.githubusercontent.com/12523676/100891632-85565d80-34ca-11eb-81b1-d2ab79a144dd.png)

It also makes sense to display specific information in window, for example, the permission that is required (`view_template`, `list_template` etc).

### Potential solution
Use second parameter for `json_encode` (available as of php 5.4) to avoid encoding problems:
`json_encode($data, JSON_UNESCAPED_UNICODE);`

### Step to reproduce
Change the language in the manager panel, for example, to russian. Create a user with access policy with `view_template` disabled in the policy template and try to edit resource.

### Expected behavior
Display readable messages.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14506 (Error Messages)

### Environment
MODX 2.x>

Ruslan-Aleev:
Fixed in https://github.com/modxcms/revolution/pull/15402

Issue #13243, opened 2017-01-06
2.5.4 requires view_template permission for listing templates, triggers permission denied popup ("200 OK error")


funkyduke:
### Summary
When logging in as a user with permissions the famous 200 OK error is back after upgrade to 2.5.4.  (https://forums.modx.com/thread/?thread=92206)

### Solution
The solution I found was to check the permission 'view_template'

### Step to reproduce
Set permissions to a user and log in. Admins with all rights have no problem.

### Observed behavior
You get the 200 OK error

### Expected behavior
No 200 OK error

### Environment
Modx 2.5.4. / PHP 5.5 or higher

Mark-H:
If the user doesn't have permission to view templates, they will get an error they don't have permission to access a template when for example listing templates. While the error is not very descriptive, this does not seem like a bug?

jaygilmore:
If it returns a mystery error message it's a UX bug. If it's showing that old dialogue with 200 and the okay button, that's not just obtuse it gives the impression of something being broken or that the user may have done something wrong. 

It should have error state microcopy associated with it. Whenever someone attempts to access something they are not permitted to access. Something like "You do not have permissions to access this [OBJECTNAME], please contact your site administrator"

funkyduke:
In previous versions of Modx (< 2.5.4) there were no problems. After upgrading to 2.5.4. the error 200 OK message is back when the user is going to a resource to update. See the forum for more info about this error, many had this problem in the past, I also. When checking 'view_template' the error is gone. To me that seems like a bug.

jaygilmore:
Are you suggesting that the Admin or SUDO user doesn't have View Template access by default or is this after creating a new user with a specific set of user permissions?

Mark-H:
It sounds more like you restricted the user account in the past, but never logged out and back in to update the permissions cached in the session. Then along comes 2.5.4, which wipes the session table, and now a simple configuration issue is a bug in a new version? :P 

Improving the error message is tracked in #12568.

funkyduke:
For the Admin or SUDO everything is fine. It's after creating a new user who has a set of user permissions. Or in the case of Modx 2.5.4. after upgrading to 2.5.4. the existing user permissions of that policy.

jaygilmore:
Right, @Mark-H, and thanks for sharing the error message task. @funkyduke, I think we need to understand the nature of the user in question for sure.

jaygilmore:
[...]

Issue #12568, opened 2015-08-19
"Permission denied" alerts for limited access policy Users


pixelchutes:
### Summary

Everything functioning as desired in MODX 2.3.5 with a custom "customer service" user role, tied to a custom access policy with limited permissions _(enough to change basic user profile fields / passwords, etc)_, but `settings` permission is **not** granted, as User Settings / System Settings access is not desired.

Upgrading to MODX 2.4.0-pl, and now a "Permission denied." alert is shown on each page load when attempting to update a user.
### Step to reproduce

Create a User, assigned to a custom role tied to a custom policy _without_ `settings` permission access. Login as the new user and attempt to edit another User in the system.
### Observed behavior

Permission Denied alert is now prompted on every page load when editing a user once the following processor is called:
- `system/settings/getAreas`

> Code: 200 OK
> {"success":false,"message":"Permission denied!","total":0,"data":[],"object":[]}

![alert](https://cloud.githubusercontent.com/assets/352182/9387892/d673804c-471f-11e5-9dbd-bea17e7935c4.jpg)
### Expected behavior

Prior to the 2.4.0 upgrade, these processors did not _alert_ the permission denied error each time. This was just fine since the access policy did not grant `settings` permission, and "failed silently", allowing the customer service user to administer users without issue, to the extent their access allowed. User settings grid was empty (by design).

Now a confusing "permission denied" prompt is alerted by simply loading the page...

Hopefully there is a way to suppress this warning / error message, as it is important to be able to provide limited access controls in the Manager (e.g. to _only_ manage users) without having to grant access to `settings` (System Settings!) to do so.
### Environment

MODX 2.4.0-pl, Chrome 46

pixelchutes:
Ok, I've finally traced it down to  0ba100587a5e5a23c232a6d5c74fa9cb54e8fe07 which modified **manager/assets/modext/widgets/core/modx.combo.js** in #12354

Specifically, it is stemming from [this fireEvent](https://github.com/modxcms/revolution/commit/0ba100587a5e5a23c232a6d5c74fa9cb54e8fe07#diff-3ccc2c503407319d1f6a255ba9bdb48dR649) call when the settings panel is empty (per missing `settings` permission)

The combo doesn't load (e.g. "fails" silently), and works as before...until the `fireEvent` is invoked, clicking on a combo with no data store.

@theboxer any ideas on how to account for both scenarios here?

pixelchutes:
[...]

PR #15402, opened 2021-02-20
Improve the processor permission response error message

Jako:
### What does it do?
- Add a lexicon entry for the processor permission message.
- Output the permission and the processor action in the message (Moving the permission class property to the modProcessor class was needed).
- Show the new error message in the MODx.combo.ComboBox loadexception

### Why is it needed?
It is not easy to detect, which permission is needed for a combobox processor permission error.

Before:

Bildschirmfoto 2021-02-20 um 21 13 52

After:

Bildschirmfoto 2021-02-20 um 21 14 42

### How to test
Add the following code to i.e. core/model/modx/processors/workspace/namespace/getlist.class.php and open the MODX system settings.

```
    public function checkPermissions() {
        return false;
    }
```

In a grid, the message is shown in the empty grid content.

### Related issue(s)/PR(s)
#13243, #12568, #15344


Ruslan-Aleev:
Also related with - https://github.com/modxcms/revolution/issues/15344
Powerful PR! Will there be no problem with encoding in json here? As in the issue above.

Jako:
Yes, the JSON will be decoded and the encoded unicode chars will be transformed to readable text.

Mark-H:
I don't typically want to engage in nitpicking textual stuff, but would "Permission 'foo bar' is **required** for 'action'" be better English? 

I think "permission denied 'foo' for 'bar'" may give users the impression they **have** permission 'foo' but that it's not a valid role of sorts.

Ruslan-Aleev (approved):
Thanks for the PR! Everything works visually.

res_changes
template_views

Jako:
@Mark-H You are right. Thanks!

theboxer (approved):

PR #15406, opened 2021-02-22
Changed the default gravatar type

Ruslan-Aleev:
### What does it do?
Changed the default gravatar type.

If you don't have a gravatar, then by default a picture with the `mm` type was displayed:
gravatar_0

But other types of gravatar (for example, `retro`) allow you to generate unique images by email hash:
email 1
gravatar_1

email 2
gravatar_2

And you can immediately understand under which account you are in the manager panel, **provided that the gravatar is enabled in the system settings** (disabled by default).

### Why is it needed?
UI / UX improvements

### DIFF
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -1041,7 +1041,7 @@ class modUser extends modPrincipal
      *
      * @return string The Gravatar photo URL
      */
-    public function getGravatar($size = 128, $default = 'mm')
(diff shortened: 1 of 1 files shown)


theboxer (approved):

Ibochkarev (approved):

Ruslan-Aleev:
@opengeek If you have the desire and time, you can merge this commit into the 2.x branch, since in the 2nd branch gravatar is enabled by default. For the 2.x branch this will be more useful.

Issue #14732, opened 2019-09-10
MODX3: The country name in the user section is not always displayed


Ruslan-Aleev:
### Summary
Sometimes the name of the country is not loaded, for what reason this happens - it is not clear. The consoles are silent.

On the gif, I just refresh the page:

![country](https://user-images.githubusercontent.com/12523676/64611519-aec92c00-d3e2-11e9-95f9-d11688f14bee.gif)

The bug is found in both https://github.com/modxcms/revolution/pull/14731 and https://github.com/modxcms/revolution/pull/14420

And in my opinion, this bug **has always been.**

This is how it looks in .js - https://github.com/modxcms/revolution/pull/14731/files#diff-28553f69fb263eee4808c9f7d481ef28R255

### Environment
MODX 3
Windows 7
Google Chrome 76.0.3809.132, (64)

Jako:
Is this the normal MODX combo? If yes, there has to be an empty or late response of the connector. Please look for XHR requests with an `system/country/getlist` action.

The processor is called, when the combo is displayed and when the combo is opened. When the combo is displayed, it is called with an 'iso' request property, that should filter the result to the selected country. The processor does not cover this case and creates a response with all countries everytime. So the patch has to go in manager/assets/modext/widgets/core/modx.combo.js

Jako:
Do you see the flickering value just before the blank value occurs? It shows `At`, but the keys i.e. in core/lexicon/country/en.inc.php are lowercase. Is that user created in MODX? How did you get the `At` value in that field?

Ruslan-Aleev:
This is pure MODX3 and the administrator that is created during installation, I did not make any additional changes. **For the appearance of this bug, I just refresh the page several times and that's it**.

Ruslan-Aleev:
By the way, I described this behavior in february - https://github.com/modxcms/revolution/pull/14420#issuecomment-466306163
What a strange bug :)

Jako:
Could you look for the database values in the user country? Are they uppercase or titlecase? If not, are they transformed during the Ajax Request?

Ruslan-Aleev:
The value with uppercase is written to the database (lexicon files have an ISO code in lowercase)

![country_db](https://user-images.githubusercontent.com/12523676/64657763-064cb380-d446-11e9-872c-703e120af229.png)

As I understand it, this line is responsible for this https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/Processors/System/Country/GetList.php#L39

[...]

PR #15413, opened 2021-02-24
Fix country name in the user section is not always displayed

Jako:
### What does it do?
- Add an empty value to the getlist processor, when the processor properties have set `combo=1`.
- Remove the default values in the combo and the panel.

### Why is it needed?
The country name in the user section is not always displayed after a reload of the page and stays empty even with a value.

### How to test
It should be fixed now, but the field value is flickering after load.

### Related issue(s)/PR(s)
#14732


Ruslan-Aleev:
~Unfortunately, the problem not disappear. The logic behind the bug is not clear. In the gif below, I just refresh the page in the browser:~

country

Jako:
Please make sure that you are using the changed javascript files in the PR.

Ruslan-Aleev:
Yes, sorry, I forgot to disable js compression in the settings. Everything now works, as it seems to me:

country_2

Ruslan-Aleev (approved):

Ibochkarev (approved):

theboxer (approved):

PR #15418, opened 2021-02-26
Strip base_url only from beginning of relative url

Jako:
### What does it do?
Use a regular expression to replace MODx.config.base_url only at the beginning of the submitted relative url.

### Why is it needed?
MODx.config.base_url should only occur at the beginning of the beginning of the submitted relative url. Currently it is stripped everywhere in the string.

### How to test
Set the value of the base_url system setting to i.e. '/test/' and create/edit a static resource. Browse for the static resource file and edit the path at the bottom manually afterwards and insert the string '/test/' somewhere in that path. Before it will be replaced everywhere, afterwards it is only replaced at the start of the string.

### Related issue(s)/PR(s)
None known.

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.static.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.static.js
@@ -47,7 +47,8 @@ Ext.extend(MODx.panel.Static,MODx.panel.Resource,{
                 'select':{fn:function(data) {
                     var str = data.fullRelativeUrl;
                     if (MODx.config.base_url != '/') {
-                        str = str.replace(MODx.config.base_url,'');
+                        var regex = new RegExp('^' + MODx.config.base_url + '(.*)');
+                        str = str.replace(regex, '/$1');
                     }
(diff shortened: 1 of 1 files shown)


alroniks:
@Jako I can not reproduce it. I tried to find such a system setting, but it does not exist. Also, I tried changing it directly in `config.inc.php`, but also all works as expected. The same results with adjusting baseUrl in the media source.

Jako:
The setting is used i.e. in LangRouter/xRouting + Babel sites as context setting with a different value than `/`. And I my client had issues with not found static PDF files because some parts of the path were stripped away after the file was selected in the MODX browser.

Jako:
Why did you add that label?

opengeek:
@Jako — does this issue apply to 3.x, as well?

Jako:
Maybe, the JavaScript can be the same.

Jako:
Yes, https://github.com/modxcms/revolution/blob/3.x/manager/assets/modext/widgets/resource/modx.panel.resource.static.js uses the same code and it should be fixed for 3.x too.

PR #15420, opened 2021-02-27
Simplified lexicons for grids

Ruslan-Aleev:
### What does it do?
MODX has specific lexicons for almost EVERY grid **that are not very useful** (see pictures below).
There are a lot of lines / translations, extra code.

ul_1

I bring the lexicons of the grid to a common format: instead of "Create Media Source" -> "Create", "Edit Media Source" -> "Edit", etc.

ul_2

In the future, I will also fix other lines, for example, "Delete multiple", "Import / Export", "Search by key", etc.

### Why is it needed?
Less code, interface unification, less lines / translations.

### How to test
Disable these system settings before testing:
- `cache_lang_js`
- `compress_js`


alroniks:
In general, I agree. Need carefully review changes, but lexicons require simplifying.

theboxer (approved):

theboxer:
Good job on this one :)

Ibochkarev (approved):
👍

alroniks:
Do you remember guys, that we had an opposite discussion in the past when people asked about consistency and add the full text to the buttons for clarity? I remember something, but I can not find the links.

I agree that we have to simplify our lexicons and the way is proper, but what if we can combine two lexicons in one place?
For example:

```
name: __('create') + ' ' __('chunk');
```

Ruslan-Aleev:
> I agree that we have to simplify our lexicons and the way is proper, but what if we can combine two lexicons in one place?
`name: __('create') + ' ' __('chunk');`

I'm not sure, but in my opinion not all languages support such a structure.

Mark-H:
What @Ruslan-Aleev said - that wouldn't be great in Dutch where the object would be first ("Chunk toevoegen") or even in the middle ("Nieuwe Chunk maken"). 

Something like this could work though: 
```
_('create', {what: _('chunk')})
```

Lexicon could put [[+what]] where it needs to go.

alroniks:
I understand it, and I know more powerful tools for it - [ICU Project](http://site.icu-project.org/) and [Intl PHP extension](https://www.php.net/manual/en/book.intl.php), but it is too complicated. :)
But I am about the case in general, should we simplify it such drastically or maybe we can find a way how to solve it technically? Maybe some advice from UX people needed about button names.

Ruslan-Aleev:
Well, for this PR there is no problem, because these lexicons are in the grids of a section, and a section usually has a title and a description.
I think it makes sense to create an issue and think over options or add your thoughts to https://github.com/modxcms/revolution/issues/14512.
In general, it is worth starting by removing unused lexicons - there will already be a huge benefit, I'm talking about - https://github.com/modxcms/revolution/issues/14512#issuecomment-650345391

alroniks:
Ok, let's then go with it as is, but I hope @rthrash will help with setup Epics to have a roadmap, and we will have the ability to collect all ideas on how we can improve translations (lexicons) under that Epic.

@Ruslan-Aleev Could you resolve conflicts, please?

Ruslan-Aleev:
Yes, I will fix the conflicts in the coming days.

Ruslan-Aleev:
@alroniks I resolved conflicts

PR #15424, opened 2021-03-02
Changed old class name to FCQN in the class_key for resources in the tree

### LINKED ISSUE #15423: [3.x] When creating a new resource from the "plus" button above the tree, Resource Type field shows "modDocument"
### Summary
On a resource under the settings tab, the default value is shown as `modDocument`. This only occurs when using the plus button above the resource tree as far as I can tell. If you use a context menu by right-clicking on the tree, it correctly shows "Document".

Screenshot from 2021-03-02 13-19-30

### Step to reproduce
Click the plus button above the resource tree to create a new resource then click the settings tab.

### Observed behavior
The dropdown shows `Document`, it's just the placeholder that shows `modDocument`

### Expected behavior
It should show `Document` in every situation.

### Environment
MODX 3 latest git

muzzwood:
…when building the create resource URL.

### What does it do?
When creating a new resource from the plus button on the resource tree, the create URL was being built incorrectly.
This PR updates it to use `MODX\Revolution\modResource` instead of just `modResource`.

### Why is it needed?
The incorrect class name (or lexicon value) was showing in the class type field when creating a new resource via the plus button.

### How to test
Try creating a new resource via the plus button on the resource tree. I assume this will need the minified JS file to be updated, so you might not see the fix without that.

### Related issue(s)/PR(s)
Fix #15423


Ibochkarev (approved):

JoshuaLuckers:
There are various other JavaScript files that use `modDocument` as reference instead of `'MODX\\Revolution\\modDocument'`.

muzzwood:
> There are various other JavaScript files that use `modDocument` as reference instead of `'MODX\\Revolution\\modDocument'`.

I'll have a look into this further shortly. Thanks!

muzzwood:
Converted other cases to MODX\Revolution\modDocument

There's also some of the other resource types that could be converted too but will need to pick that up later.

JoshuaLuckers (approved):

manager/assets/modext/widgets/resource/modx.tree.resource.js
```diff
@@ -657,7 +657,7 @@ Ext.extend(MODx.tree.Resource,MODx.tree.Tree,{
         if (Ext.isObject(o)) {
             Ext.apply(types,o);
         }
-        var coreTypes = ['modDocument','modWebLink','modSymLink','modStaticResource'];
+        var coreTypes = ['MODX\\Revolution\\modDocument','modWebLink','modSymLink','modStaticResource'];
```
alroniks: Other classes also should be with namespaces.

muzzwood: Thanks @alroniks , I ran out of time unfortunately. I'm aiming to get this completed today! :)

muzzwood: Done!

theboxer (approved):

PR #15442, opened 2021-03-05
Fix user block checks on login.

sergant210:
### What does it do?
Fixed logic errors and incorrect field name of the "user_attributes" table. 
The logic is next - the "blocked" checkbox is an indicator of the user's current state. It can be switched  manually or automatically if the user is blocked after time ("Blocked After") or the block time ("Blocked Unfil") is expired.
If the user is blocked because of the "Blocked After" time, then switching off the "blocked" checkbox will clear the "Blocked After" field.
The "Blocked Until" check will work if the "blocked" checkbox is On.

blockuser

### Why is it needed?
Some user block state checks do not work correctly. Now you will see a message about too many attempts to login.

### How to test
1. Create a test user.
2. Play around with the block options. 
3. Open manager login page in the incognito browser mode. 
4. Then try to login.

### Related issue(s)/PR(s)
[External issue](https://github.com/Sterc/revolution/issues/27).

### DIFF
--- a/core/src/Revolution/Processors/Security/Login.php
+++ b/core/src/Revolution/Processors/Security/Login.php
@@ -187,7 +187,9 @@ class Login extends Processor
                 $profile->save();
             }
 
-            return $this->modx->lexicon('login_blocked_admin');
+            if ($profile->get('blocked')) {
+                return $this->modx->lexicon('login_blocked_admin');
+            }
         }
 
         if ($profile->get('blockedafter') > 0 && $profile->get('blockedafter') < time()) {
--- a/core/src/Revolution/Processors/Security/Login.php
+++ b/core/src/Revolution/Processors/Security/Login.php
@@ -178,16 +178,15 @@ class Login extends Processor
             return $this->modx->lexicon('login_blocked_too_many_attempts');
         }
 
-        if ($profile->get('blockeduntil') != 0 && $profile->get('blockeduntil') < time()) {
-            $profile->set('failedlogincount', 0);
-            $profile->set('blocked', 0);
-            $profile->save();
-        }
-
         if ($profile->get('blocked')) {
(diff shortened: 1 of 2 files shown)


core/src/Revolution/Processors/Security/Login.php
```diff
@@ -170,29 +170,31 @@ public function checkIsBlocked()
         /** @var modUserProfile $profile */
         $profile = $this->user->Profile;
 
-        if ($profile->get('failed_logins') >= $this->modx->getOption('failed_login_attempts') && $profile->get('blockeduntil') > time()) {
-            return $this->modx->lexicon('login_blocked_too_many_attempts');
-        }
-
         if ($profile->get('failedlogincount') >= $this->modx->getOption('failed_login_attempts')) {
```
Mark-H, line 173: By removing the check above and instead letting it fall back to this `if {}` block, the blocked time will be extended each time the user tries to login while they're still blocked - is that intentional?

JoshuaLuckers: Not necessarily a requirement for this PR to be approved but It would be really great to have a Unit Test for this.

sergant210: > By removing the check above and instead letting it fall back to this `if {}` block, the blocked time will be extended each time the user tries to login while they're still blocked - is that intentional?

This will never happen, because if the user is blocked, the count of login attempts will stop because it happens after this check, which fails. By the way, this condition will always be false. And not just because the "failed_logins" field doesn't exist. I have never received a message about too many attempts. I bet you can't either. )

Mark-H (approved):

Mark-H:
Not sure why this didn't autoclose, but it's merged. Thanks!

PR #15444, opened 2021-03-05
Update MODX logo and icon

Ibochkarev:
### What does it do?
Update MODX logo and icon

### Why is it needed?
Updates to the current state

### How to test
Pick up the changes in this PR and test =)

Before:

2021-03-05 12-36-44 MODX Revolution 3 0 0-alpha3 » Установить - Google Chrome

After:

### Related issue(s)/PR(s)
NA


Ruslan-Aleev:
Edits from https://github.com/modxcms/revolution/pull/15405 need to be added too :)

Ruslan-Aleev:
By the way, it makes sense for separate PRs to change the green color, because now it does not match the logo, although it used to be.
@rthrash What do you think?

green_1

green_2

Jako:
> By the way, it makes sense for separate PRs to change the green color, because now it does not match the logo, although it used to be.

The new color is more blue than green, so it could stay at that state.

Ruslan-Aleev:
> The new color is more blue than green, so it could stay at that state.

On the spectrum, the new color is more green, hehehe:

green_3_0

But this is the color I would not use - the contrast drops, but I would make it darker to match the logo.

green_3

These are only thoughts so far.

JoshuaLuckers:
It's best to keep the discussion in this PR related to the new logo and to discuss the color changes in a new issue/PR.

Jako:
But that's not a bug.

Ruslan-Aleev:
I will create a feature request, after the merge, to discuss in the future.

JoshuaLuckers (changes requested):
The setup is using the old favicon `setup/favicon.ico`.

Ibochkarev:
> The setup is using the old favicon `setup/favicon.ico`.

Can't get updated in ico format

Jako:
Use the ones from https://modx.com/

Jako (changes requested):
Please use this https://modx.com/favicon.ico

Jako:
The favicon is not the right one. Please use https://modx.com/favicon.ico.

Ibochkarev:
> The favicon is not the right one. Please use https://modx.com/favicon.ico.

Thank you, I will send corrections today.

Ibochkarev:
@Jako done

JoshuaLuckers (approved):

theboxer (approved):

Jako (approved):

PR #15454, opened 2021-03-05
Add snippet type binding for Template Variables

### LINKED ISSUE #14074: Add @SNIPPET binding for generating list of TV options
### Summary
With PR https://github.com/modxcms/revolution/pull/13865 `@EVAL` has been removed for security reasons. If you want to use a snippet for generating a list of TV options we normally use `@EVAL`. The suggested fix for this issue was to use a chunk which calls a snippet (as mentioned in the comments of the PR).

I think `@SNIPPET` should be added as a binding option, else you need to add a chunk and a snippet instead of just having the one snippet. It's more a work-around then a proper fix in my opinion.

Also I'd like to see support for `@FILE`, and make it possible to do something like this:
```
@FILE [[++core_path]]/components/mycomponent/elements/snippets/getlist.php
```

Which makes it possible to also work filebased without any reference to the database.

JoshuaLuckers:
### What does it do?
Adds a new binding type (@SNIPPET) to Template Variables. 

### Why is it needed?
To make it possible to generate a dynamic list of input values. 
See #14074 for more info.

### How to test
Create a new snippet that returns a valid Input Options string for TV's:
``` 
<?php
$options = [
    'Option 1',
    'Option 2',
];

return implode('||', $options);
```

Create a new select TV with Input Type "Listbox" (single or multi).
In Input Option Values you can now use the `@SNIPPET` binding:
```
@SNIPPET snippetname
```

### Related issue(s)/PR(s)
Closes #14074

### DIFF
--- a/core/src/Revolution/modTemplateVar.php
+++ b/core/src/Revolution/modTemplateVar.php
@@ -52,6 +52,7 @@ class modTemplateVar extends modElement
     public $bindings = [
         'FILE',
         'CHUNK',
+        'SNIPPET',
         'DOCUMENT',
         'RESOURCE',
         'SELECT',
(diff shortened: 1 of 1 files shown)


Bruno17:
would be nice, if we could pass snippet-properties to the snippet somehow

JoshuaLuckers:
@Bruno17 I would consider that a new feature, it would be useful for the `@CHUNK` binding as well

PR #15457, opened 2021-03-05
Add transliteration of files when uploading to a media source #modxbughunt

ilyautkin:
### What does it do?
Add system setting `upload_translit`. If this setting is `yes` (default), filename of any uploaded file will be transliterated according global translitirating rules (setting `friendly_alias_translit`)

### Why is it needed?
There are many problems a file can give on server, for example, with a russian name.

### How to test
You should to rebuild transport core - `php _build/transport.core.php`, remove `/setup/.locked/` and run setup. Then you can install `Translit` extra from MODX repo or any other, for example `yTranslit` from `modstore.pro`. Set setting `friendly_alias_translit` to `russian` or any other. Upload any files with non latin symbols. Name of uploaded file should be transliterated.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15007

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -736,7 +736,7 @@ $_lang['setting_upload_media'] = 'Uploadable Media Types';
 $_lang['setting_upload_media_desc'] = 'Here you can enter a list of files that can be uploaded into \'assets/media/\' using the Resource Manager. Please enter the extensions for the media types, separated by commas.';
 
 $_lang['setting_upload_translit'] = 'Transliterate names of uploading files?';
-$_lang['setting_upload_translit_desc'] = 'If <b>yes</b> name of any uploading file will be transliterated by global translitiration rules.';
+$_lang['setting_upload_translit_desc'] = 'If \'Yes\' name of any uploading file will be transliterated by global translitiration rules.';
 
 $_lang['setting_use_alias_path'] = 'Use Friendly Alias Path';
 $_lang['setting_use_alias_path_desc'] = 'Setting this option to \'yes\' will display the full path to the Resource if the Resource has an alias. For example, if a Resource with an alias called \'child\' is located inside a container Resource with an alias called \'parent\', then the full alias path to the Resource will be displayed as \'/parent/child.html\'.<br /><strong>NOTE: When setting this option to \'Yes\' (turning on alias paths), reference items (such as images, CSS, JavaScripts, etc.) use the absolute path, e.g., \'/assets/images\' as opposed to \'assets/images\'. By doing so you will prevent the browser (or web server) from appending the relative path to the alias path.</strong>';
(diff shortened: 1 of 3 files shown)


Ibochkarev (changes requested):
Need a script to upgrade from previous versions

core/lexicon/en/setting.inc.php
```diff
@@ -735,6 +735,9 @@
 $_lang['setting_upload_media'] = 'Uploadable Media Types';
 $_lang['setting_upload_media_desc'] = 'Here you can enter a list of files that can be uploaded into \'assets/media/\' using the Resource Manager. Please enter the extensions for the media types, separated by commas.';
 
+$_lang['setting_upload_translit'] = 'Transliterate names of uploading files?';
+$_lang['setting_upload_translit_desc'] = 'If <b>yes</b> name of any uploading file will be transliterated by global translitiration rules.';
```
Ruslan-Aleev: Here it is worth writing as in the lexicons below, instead of `<b>yes</b>` -> `\'yes\'`

ilyautkin: Yes, thank you. This fixed

Mark-H:
@Ibochkarev why do you say this needs an upgrade script, am I overlooking something?

Mark-H (approved):

Mark-H:
Note for implementation testing, please make sure disabling the setting allows the functionality to actually be disabled. I've seen cases where the value being set to a string `no` causes cast-to-boolean checks like this to not work correctly, but I don't immediately recall if that's still a problem in todays MODX.

Ibochkarev (approved):
Installed add-on **translit**, indicated `friendly_alias_translit` for the system setting in `russian`

Transliteration works fine

изображение

If files have already been uploaded before enabling the upload_translit system setting



And we are trying to upload the same file

Ibochkarev:
> @Ibochkarev why do you say this needs an upgrade script, am I overlooking something?

I was wrong here. All OK

core/lexicon/en/setting.inc.php
```diff
@@ -735,6 +735,9 @@
 $_lang['setting_upload_media'] = 'Uploadable Media Types';
 $_lang['setting_upload_media_desc'] = 'Here you can enter a list of files that can be uploaded into \'assets/media/\' using the Resource Manager. Please enter the extensions for the media types, separated by commas.';
 
+$_lang['setting_upload_translit'] = 'Transliterate names of uploading files?';
```
alroniks, line 738: Not sure that here `?` is needed. It should just describe the setting. And if there is a question, it should be rewritten according to grammar.

ilyautkin: This is boolean setting. We have questions in this settings, for example: https://github.com/modxcms/revolution/blob/dffa281d8bb2470cca7c06912b0a3417863cf1ad/core/lexicon/en/setting.inc.php#L673-L686

alroniks (approved):

Issue #13973, opened 2018-07-04
MODX 3: Users - Password notification method missing. Bug or by design?


jonleverrier:
### Summary
There is a large gap where the Password notification method use to be. Also, the Password notification method is missing from a users profile - is this a bug or by design?

When "Let MODX generate a password" or "Let me specify the password" is selected on save, the password is presented on screen - no email is sent.

![screen shot 2018-07-04 at 21 15 24](https://user-images.githubusercontent.com/2373940/42292901-45e081d6-7fd6-11e8-95a2-a5272e45ec15.png)

### Step to reproduce
1. Edit a users profile
2. Go to the new password area
3. Try to specify a password and email the user

### Observed behavior
Password notification method missing

### Expected behavior
To show Password notification method options

![screen shot 2018-07-04 at 21 15 15](https://user-images.githubusercontent.com/2373940/42292998-f7f14342-7fd6-11e8-90a4-92fc19ea590f.png)

### Environment
MODX3 Alpha v 3.0.0-dev
From https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99

Mark-H:
I believe that to be by design; doesn't it send the user an email to change their pass always now? Maybe the text needs updating.

jonleverrier:
I thought that also @Mark-H. But I didn't receive the email when I selected "Let MODX generate a password" and hit save. Plus there is a gap in the design where the password notifications use to be.

jonleverrier:
Somethings not right - I stand by my original issue.

We're missing the ability to either send an email or see the new password on screen.

When I used the new password methods it had actually displayed the new password on screen, I just didn't spot it

<img width="804" alt="screen shot 2018-07-04 at 22 32 19" src="https://user-images.githubusercontent.com/2373940/42293425-65e83fb0-7fda-11e8-9fe3-21e561b28f64.png">

JoshuaLuckers:
There are a lot of times I don't want an email to be send when a (new) password is made for a user.
I really don't hope this will be the default behaviour.

Mark-H:
I may be wrong and mixing up the manager reset password feature and changing the password from the back-end. Stuff happened in that department, but I don't immediately have it clear what did and didn't change, so this issue would warrant double checking at the least.

OptimusCrime:
Actually, we should consider removing the option for sending the password over email. It is _very_ insecure and considered a terrible practice.

[...]

PR #15461, opened 2021-03-05
Let user specify a new password (#13973)

sdrenth:
### What does it do?
This adds the possibility to let a user specify their own MODX password, using the same flow as the current manager Forgot password flow by sending a hashed link in the email which allows users to setup their new password.

Credits to @jonleverrier for coming up with this solution.

New password options:
Screenshot 2021-03-05 at 11 35 16

Set up password mail:
Screenshot 2021-03-05 at 10 20 48

### Why is it needed?
There used to be an option to send an email to the MODX user which contains their MODX account password. This has been removed (for security purposes). This basically readds that functionality but in a saver way.

### Related issue(s)/PR(s)
https://github.com/Sterc/revolution/issues/22
https://github.com/modxcms/revolution/issues/13973
https://github.com/Sterc/revolution/pull/31


Ibochkarev:
@sdrenth please change target branch

sdrenth:
@JoshuaLuckers Thanks! :)

core/src/Revolution/Processors/Security/User/Create.php
```diff
@@ -232,6 +232,52 @@ public function sendNotificationEmail() {
                 'html' => true,
             ]);
         }
+
+        if ($this->getProperty('passwordgenmethod') === 'user_email_specify') {
+            $activationHash = md5(uniqid(md5($this->object->get('email') . '/' . $this->object->get('id')), true));
```
JoshuaLuckers: I propose to use the `hashing` service to generate the hash. 

For reference: https://github.com/modxcms/revolution/blob/2ee2c45cda11e578595990874407fbfa668f4ff4/core/src/Revolution/modUser.php#L67

Jako:
Maybe the sendNotificationEmail method can be moved to a helper class instead of duplicating the code.

sdrenth:
@Jako I agree with your comment but think that should be a separate issue, because the duplicate code was already in here (and there's more). Also where should the helper class be placed?

JoshuaLuckers:
The user needs access to the manager context, otherwise an access denied error will show up (only after you try to set your own password).

sdrenth:
@JoshuaLuckers Is that just an FYI? 

I would think that for frontend users you would have a registration form and a forgot password form. You don't want to guide those users to the manager part of the application. 

I could also rename that option to "Let the user choose their own password via email (Needs manager access)" to clarify this?

What do you think?

JoshuaLuckers:
First of all, I like this new feature! 

I'm having some doubts about the flow of this feature. If a user doesn't have access to the manager it still receives the email and the option to set the password. Only after setting your password you get the access denied message.

JoshuaLuckers (approved):

core/lexicon/en/user.inc.php
```diff
@@ -200,3 +201,5 @@
 $_lang['users'] = 'Users';
 $_lang['user_createdon'] = 'Created On';
 $_lang['user_createdon_desc'] = 'The date the user was created.';
+$_lang['user_password_email_subject'] = 'Set up your password';
+$_lang['user_password_email'] = '<h2>Set up your password</h2><p>We received a request to set up your MODX Revolution password. You can set up your password by clicking the button below and following the instructions on screen.</p><p class="center"><a href="[[+url_scheme]][[+http_host]][[+manager_url]]?modhash=[[+hash]]" class="btn">Set up my password</a></p><p class="small">If you did not send this request, please ignore this email.</p>';
```
alroniks, line 205: Let's avoid using HTML tags inside lexicons. It becomes a hell while translating such stuff.

core/src/Revolution/Processors/Security/User/Create.php
```diff
@@ -232,6 +233,52 @@ public function sendNotificationEmail() {
                 'html' => true,
             ]);
         }
+
+        if ($this->getProperty('passwordgenmethod') === 'user_email_specify' && $this->modx->getService('hashing', modHashing::class)) {
```
alroniks, line 237: It should be refactored, to move this implementation into a method to avoid such long if statement.

alroniks (changes requested):
I like the feature, but the code should be improved before merging. 
I will do it if you don't mind.

core/src/Revolution/Processors/Security/User/Create.php
Ibochkarev: @sdrenth Please make changes to promote PR

alroniks:
Looks like a new feature for including to the beta after the feature freeze. I suggest moving to 3.1
WDT @Mark-H, @opengeek?

core/src/Revolution/Processors/Security/User/Create.php
```diff
@@ -232,6 +233,52 @@ public function sendNotificationEmail() {
...
+            $activationHash = $this->modx->hashing->getHash('md5', 'hashing.modMD5', [])->hash($this->object->get('email') . '/' . $this->object->get('id'));
```
Mark-H, line 238: This is **not** a secure hash/tokens and introduces a potential vulnerability. The scope would be limited because the "tokens" have a limited lifetime after manually being created by an admin, however it makes the hash _very_ predictable.

In 3.x we can use `random_bytes`, so I would suggest:

```` 
$activationHash = bin2hex(random_bytes(32));
```` 

for a 64-character secure random token.

Mark-H (changes requested):

Mark-H:
As this PR has gone stale but this is still something that would be great to have, I've replaced it with a new one over at #16519.

Issue #13065, opened 2016-07-07
Enhance modal dialog for deleting/removing a resource, directory, file or category


pepebe:
### Summary

Add name of resource/directory/category or file to modal dialog. Somebody added this to elements (templates, chunks, snippets,plugins,tvs) in the past (god bless him/her), but it is still not included for resources or files.
### Step to reproduce

Delete/Remove any resource, file or directory.
### Observed behavior

No feedback inside the modal about the object that will be affected by the current action. 

![screen shot 2016-07-07 at 09 30 21](https://cloud.githubusercontent.com/assets/667315/16645718/1f8a6800-4426-11e6-9910-45ebb71423f3.png)

![screen shot 2016-07-07 at 09 30 36](https://cloud.githubusercontent.com/assets/667315/16645720/24e1ebfc-4426-11e6-82fc-80de43abf3f7.png)

![screen shot 2016-07-07 at 09 27 49](https://cloud.githubusercontent.com/assets/667315/16645727/31dac0b8-4426-11e6-93c7-8fdd6d64bb2e.png)

![screen shot 2016-07-07 at 09 28 33](https://cloud.githubusercontent.com/assets/667315/16645724/2a85b9da-4426-11e6-97c5-4ff61179290a.png)
### Expected behavior

The modal dialog shows the name of the object to be deleted/removed. 

This is how its done right:

![screen shot 2016-07-07 at 09 28 19](https://cloud.githubusercontent.com/assets/667315/16645740/473c870c-4426-11e6-8fb6-8e05eb652c7b.png)
### Environment

MODX 2.5.0

Cheers,

patrick

mindeffects:
+10
I wish I had this a couple if weeks ago when I delete the wrong stuff.
Yes, point fingers at me for not having a backup, but I rather prefer a backup AND a propper user expetience. ;-)

Ruslan-Aleev:
Working on it, #BugHunt2021

Ibochkarev:
Added #15463

Ruslan-Aleev:
Not all delete windows have been fixed, for example, windows in grid still need to be fixed. I'll open it for now :)

PR #15463, opened 2021-03-05
Added display of the name of elements in delete messages

Ruslan-Aleev:
### What does it do?
Added displaying the names of elements in delete messages, BUT only for trees, for grids I will deal with in the future (there need to simplify a lot of lexicons first).

Before:
del_1

After:
del_2

### Why is it needed?
You may not understand what is being deleted and delete the wrong one :)

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/13065


Ibochkarev (approved):
image

JoshuaLuckers:
For some of those confirm messages we should look into using the lexicon `remove_this_confirm`. It accepts a `type` and `name` property.

Can be used like this:
```
_('remove_this_confirm',{
    type: 'Resource Group'
    ,name: 'The Name Of The Resource Group'
})
```

This is also used for the warning dialog when deleting elements.

Ruslan-Aleev:
> For some of those confirm messages we should look into using the lexicon `remove_this_confirm`. It accepts a `type` and `name` property.

The current variant is the most optimal. Only for "User Group" there is a `type`, but there is no lexicon for user group type (and if add a lexicon, it will be confusing, because lexicons will be loaded from `default.inc.php`, but for the "User Group" section with an existing `user.inc.php`).

JoshuaLuckers (approved):

PR #15464, opened 2021-03-05
Add next/prev/up buttons to the resource panel

wax100:
### What does it do?
Add buttons next \ prev \ up to resources

### Why is it needed?
~Enhancement #12~

### Related issue(s)/PR(s)
~Enhancement #12 https://github.com/modxorg/DocsApp/issues/12~

### DIFF
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1610,10 +1610,10 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
                 }
             }
 
-            $arr = array(
+            $arr = [
                 'left' => array_reverse($left),
                 'right' => $right,
-            );
+            ];
         }
 
         return $arr;
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1601,7 +1601,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
             $ids = $q->stmt->fetchAll(PDO::FETCH_COLUMN);
             $current = array_search($this->id, $ids);
 
-            $right = $left = array();
+            $right = $left = [];
             foreach ($ids as $k => $v) {
                 if ($k > $current) {
                     $right[] = $v;
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1584,4 +1584,38 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
             $this->xpdo->invokeEvent('OnResourceCacheUpdate', ['id' => $this->get('id')]);
         }
     }
+
+    /**
+     * Returns array with all neighborhood resources
+     *
+     * @return array $arr Array with neighborhood from left and right
+     */
+    public function getNeighborhood()
+    {
+        $arr = array();
+
+        $q = $this->xpdo->newQuery('modResource', array('parent' => $this->parent));
+        $q->sortby('menuindex', 'ASC');
+        $q->select('id');
+        if ($q->prepare() && $q->stmt->execute()) {
+            $ids = $q->stmt->fetchAll(PDO::FETCH_COLUMN);
+            $current = array_search($this->id, $ids);
+
+            $right = $left = array();
+            foreach ($ids as $k => $v) {
+                if ($k > $current) {
+                    $right[] = $v;
+                } elseif ($k < $current) {
+                    $left[] = $v;
+                }
+            }
+
+            $arr = array(
+                'left' => array_reverse($left),
+                'right' => $right,
+            );
+        }
+
+        return $arr;
+    }
 }
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1594,7 +1594,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
     {
         $arr = [];
 
-        $q = $this->xpdo->newQuery('modResource', array('parent' => $this->parent));
+        $q = $this->xpdo->newQuery(modResource::class, ['parent' => $this->parent]);
         $q->sortby('menuindex', 'ASC');
         $q->select('id');
         if ($q->prepare() && $q->stmt->execute()) {
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1592,7 +1592,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
      */
     public function getNeighborhood()
     {
-        $arr = array();
+        $arr = [];
 
         $q = $this->xpdo->newQuery('modResource', array('parent' => $this->parent));
         $q->sortby('menuindex', 'ASC');
--- a/manager/assets/modext/sections/resource/update.js
+++ b/manager/assets/modext/sections/resource/update.js
@@ -227,6 +227,8 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
             ,scope: this
         });
 
+        btns.push(this.getAdditionalButtons(config));
+
         btns.push({
             process: 'Resource/Update'
             ,text: _('save') + ' <i class="icon icon-check"></i>'
@@ -241,6 +243,49 @@ Ext.extend(MODx.page.UpdateResource,MODx.Component,{
         });
 
         return btns;
+    }
+
+    ,getAdditionalButtons: function (config) {
+        return [{
+            text: '<i class="icon icon-arrow-left"></i>',
+            handler: this.prevPage,
+            disabled: !config['prev_page'],
+            scope: this,
+            tooltip: _('prev_page'),
+            keys: [{key: 37, alt: true, scope: this, fn: this.prevPage}]
+        }, {
+            text: '<i class="icon icon-arrow-up"></i>',
+            handler: this.upPage,
+            scope: this,
+            disabled: !config['up_page'],
+            tooltip: _('up_page'),
+            keys: [{key: 38, alt: true, scope: this, fn: this.upPage}]
+        }, {
+            text: '<i class="icon icon-arrow-right"></i>',
+            handler: this.nextPage,
+            disabled: !config['next_page'],
+            scope: this,
+            tooltip: _('next_page'),
+            keys: [{key: 39, alt: true, scope: this, fn: this.nextPage}]
+        }];
+    }
+
+    ,prevPage: function () {
+        if (this.config['prev_page'] > 0) {
+            MODx.loadPage('resource/update', 'id=' + this.config['prev_page'])
+        }
+    }
+
+    ,nextPage: function () {
+        if (this.config['next_page'] > 0) {
+            MODx.loadPage('resource/update', 'id=' + this.config['next_page'])
+        }
+    }
+
+    ,upPage: function () {
+        if (this.config['up_page'] > 0) {
+            MODx.loadPage('resource/update', 'id=' + this.config['up_page'])
+        }
     }
 });
 Ext.reg('modx-page-resource-update',MODx.page.UpdateResource);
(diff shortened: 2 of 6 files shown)


Mark-H:
That issue was about adding prev/next links to the documentation, not to MODX 😆 

This looks cool as well though!

wax100:
> That issue was about adding prev/next links to the documentation, not to MODX 😆

this is feature, not bug =)  Sorry, don't know how to work with Docs

core/src/Revolution/modResource.php
```diff
@@ -1584,4 +1584,38 @@ public function clearCache($context = '')
...
     }
+
+    /**
+     * Returns array with all neighborhood resources
+     *
+     * @return array $arr Array with neighborhood from left and right
+     */
+    public function getNeighborhood()
+    {
+        $arr = array();
```
JoshuaLuckers:
```suggestion
        $arr = [];
```

core/src/Revolution/modResource.php
```diff
@@ -1584,4 +1584,38 @@ public function clearCache($context = '')
...
+
+        $q = $this->xpdo->newQuery('modResource', array('parent' => $this->parent));
```
JoshuaLuckers:
```suggestion
        $q = $this->xpdo->newQuery(modResource::class, ['parent' => $this->parent]);
```

core/src/Revolution/modResource.php
```diff
@@ -1584,4 +1584,38 @@ public function clearCache($context = '')
...
+        $q->sortby('menuindex', 'ASC');
+        $q->select('id');
+        if ($q->prepare() && $q->stmt->execute()) {
+            $ids = $q->stmt->fetchAll(PDO::FETCH_COLUMN);
+            $current = array_search($this->id, $ids);
+
+            $right = $left = array();
```
JoshuaLuckers:
```suggestion
            $right = $left = [];
```

core/src/Revolution/modResource.php
```diff
@@ -1584,4 +1584,38 @@ public function clearCache($context = '')
...
+                    $right[] = $v;
+                } elseif ($k < $current) {
+                    $left[] = $v;
+                }
+            }
+
+            $arr = array(
+                'left' => array_reverse($left),
+                'right' => $right,
+            );
```
JoshuaLuckers:
```suggestion
            $arr = [
                'left' => array_reverse($left),
                'right' => $right,
            ];
```

core/src/Revolution/modResource.php
```diff
@@ -1584,4 +1584,38 @@ public function clearCache($context = '')
...
+    /**
+     * Returns array with all neighborhood resources
+     *
+     * @return array $arr Array with neighborhood from left and right
+     */
+    public function getNeighborhood()
+    {
+        $arr = [];
+
+        $q = $this->xpdo->newQuery(modResource::class, ['parent' => $this->parent]);
```
theboxer, line 1597: I'm not sure I like this approach. On large sites, this basically means that everytime you open resource to edit you'll fetch all resources under the same parent, which could be hundreds or even thousands of records in some cases, just to get 2 specific IDs.

Wouldn't be better to query directly for the resource under and above the current one?

wax100: It was idea for modx docs, not for modx CMS) my mistake

Ruslan-Aleev (approved):
An interesting option, convenient for a news list, for example :)

res_nav

core/src/Revolution/modResource.php
Mark-H: Here's one way to get the prev/next without having to get all resources: add a `where` that matches the sortby.

For the next:

```
$qNext->where([
     'id:!=' => $this->get('id'),
     'AND:menuindex:<=' => $this->get('menuindex'),
]);
$qNext->limit(1);
$qNext->sortby('menuindex', 'ASC');
```

For the previous:
```
$qPrevious->where([
     'id:!=' => $this->get('id'),
     'AND:menuindex:>=' => $this->get('menuindex'),
]);
$qPrevious->limit(1);
$qPrevious->sortby('menuindex', 'ASC');
```

The "equals or larger/smaller" makes sure that items with the same menuindex still appear. 

It means two queries, but will be much more optimized than loading all siblings, especially on large sites.

Mark-H (changes requested):
Needs to use more optimized queries.

Mark-H:
Even if the initial request wasn't meant for the core, this could be a nifty feature to add to 3.1. 

Maybe there's something to be said for optimising the UI a bit more, to make it take up less space in an already busy toolbar? Perhaps a buttongroup instead of multiple buttons? Not, sure, just ideas.

smg6511:
Even if just the keyboard actions were implemented and there were no buttons added this would be a handy feature. That said, the suggestion of a button group would make sense to keep the UI as simple as possible in this area.

@wax100 - Would you like someone else to pick up where you left off on this PR?

theboxer:
Is moving to the next/prev resource really something that's used that much on all resources, that it deserves the spot in the already packed action bar?

I can see this useful in some specific scenarios for resources under specific parent, where it **may** be handy to quickly move to the next/prev one. But in general space, I don't think this is needed.

Ibochkarev:
## Summary

Adds prev/next/up navigation buttons to the resource edit panel, allowing users to move between sibling resources (by menuindex) and to the parent resource. Uses `modResource::getNeighborhood()` to fetch sibling IDs and passes `prev_page`, `next_page`, `up_page` to the frontend. Keyboard shortcuts: Alt+Left, Alt+Right, Alt+Up.

**Scope:** 6 files, ~111 lines (modResource, update.js, 4 resource update controllers).

## Reviewer Feedback Addressed / Outstanding

### 1. Query optimization (theboxer, Mark-H)

**Feedback:** theboxer — "On large sites, this basically means that everytime you open resource to edit you'll fetch all resources under the same parent, which could be hundreds or even thousands of records in some cases, just to get 2 specific IDs.

Wouldn't be better to query directly for the resource under and above the current one?"

**Mark-H (Jan 2023):** Suggested optimized queries using `menuindex` with `where`:

- For next: `'AND:menuindex:<=' => $this->get('menuindex')`, `'id:!=' => $this->get('id')`, sort by menuindex ASC, limit 1
- For prev: `'AND:menuindex:>=' => $this->get('menuindex')`, `'id:!=' => $this->get('id')`, sort by menuindex ASC, limit 1

**Current state:** Not addressed. The PR still loads all sibling IDs with `$q->select('id')` and then filters in PHP. **Mark-H requested changes** — "Needs to use more optimized queries."

**Recommendation:** Implement Mark-H's approach: two queries instead of loading all siblings. Example:

```php
public function getNeighborhood()
{
    $arr = ['left' => [], 'right' => []];
    $id = (int)$this->get('id');
    $parent = (int)$this->get('parent');
    $menuindex = (int)$this->get('menuindex');

    if ($parent < 0) {
        return $arr;
    }

    $qNext = $this->xpdo->newQuery(modResource::class, ['parent' => $parent]);
    $qNext->where(['id:!=' => $id, 'AND:menuindex:>=' => $menuindex]);
    $qNext->sortby('menuindex', 'ASC');
    $qNext->sortby('id', 'ASC');
    $qNext->limit(1);
    $qNext->select('id');
    if ($next = $this->xpdo->getValue($qNext->prepare())) {
        $arr['right'] = [(int)$next];
    }

    $qPrev = $this->xpdo->newQuery(modResource::class, ['parent' => $parent]);
    $qPrev->where(['id:!=' => $id, 'AND:menuindex:<=' => $menuindex]);
    $qPrev->sortby('menuindex', 'DESC');
    $qPrev->sortby('id', 'DESC');
    $qPrev->limit(1);
    $qPrev->select('id');
    if ($prev = $this->xpdo->getValue($qPrev->prepare())) {
        $arr['left'] = [(int)$prev];
    }

    return $arr;
}
```

Adjust logic for correct prev/next (menuindex ordering may need refinement). The controllers expect `$neighborhood['right'][0]` and `$neighborhood['left'][0]`, so the structure can stay.

### 2. UI placement (theboxer, Mark-H)

**theboxer (Nov 2024):** "Is moving to the next/prev resource really something that's used that much on all resources, that it deserves the spot in the already packed action bar? I can see this useful in some specific scenarios for resources under specific parent, where it **may** be handy to quickly move to the next/prev one. But in general space, I don't think this is needed."

**Mark-H:** "Perhaps a buttongroup instead of multiple buttons? Not sure, just ideas."

**smg6511:** "Even if just the keyboard actions were implemented and there were no buttons added this would be handy."

**Recommendation:** Consider:
- A single button group (prev | up | next) to reduce toolbar clutter
- Or keyboard shortcuts only, with a configurable system setting to show/hide buttons
- Or showing the buttons only when there are siblings (e.g. when `prev_page` or `next_page` > 0)

### 3. Missing lexicon entries

The JS uses `_('prev_page')`, `_('next_page')`, `_('up_page')` for tooltips. The PR does not add these lexicon keys. Without them, tooltips will show the raw key or fallback.

**Recommendation:** Add to `core/lexicon/en/resource.inc.php` (or equivalent):

```php
$_lang['prev_page'] = 'Previous resource';
$_lang['next_page'] = 'Next resource';
$_lang['up_page'] = 'Parent resource';
```

## Code Quality Issues

### 1. `getNeighborhood()` — parent = 0

When `parent` is 0 (root), the query `['parent' => $this->parent]` will return all root siblings. On large sites this can be hundreds of resources. The optimization above addresses this.

### 2. Duplication in controllers

Each of the four update controllers (document, static, symlink, weblink) repeats the same logic:

```php
$neighborhood = $this->resource->getNeighborhood();
// ...
'next_page' => !empty($neighborhood['right'][0]) ? $neighborhood['right'][0] : 0,
'prev_page' => !empty($neighborhood['left'][0]) ? $neighborhood['left'][0] : 0,
'up_page' => $this->resource->parent
```

**Recommendation:** Move this to the base `ResourceUpdateManagerController` so it is set once and inherited by document, static, symlink, weblink. If the base controller already prepares `$data`, add the neighborhood there.

### 3. `$this->parent` vs `$this->get('parent')`

`modResource::getNeighborhood()` uses `$this->parent`. In xPDO, `$this->parent` may be the raw field value. Prefer `$this->get('parent')` for consistency with the rest of the codebase.

---

### 4. `array_search` with strict types

`array_search($this->id, $ids)` — `$ids` from `fetchAll(PDO::FETCH_COLUMN)` are strings. `$this->id` may be int. Use `array_search((string)$this->id, $ids)` or strict comparison to avoid false matches.

## Assessment

### Strengths

- Useful feature for editing sequential resources (e.g. news, blog).
- Keyboard shortcuts (Alt+Left/Right/Up) improve power-user workflow.
- Clean separation: `getNeighborhood()` in model, controller passes data, JS handles UI.
- Works for all resource types (document, static, symlink, weblink).

### Concerns

- **Performance:** Loads all siblings on every resource edit; must be optimized per Mark-H.
- **UI:** Toolbar placement disputed; may need button group or optional display.
- **Missing lexicons:** Tooltips will show raw keys.
- **Duplication:** Neighborhood logic repeated in four controllers.

## Verdict

**Request changes** — must address the following before merge:

1. **Optimize `getNeighborhood()`** — use two targeted queries instead of loading all siblings (per Mark-H's suggestion).
2. **Add lexicon entries** for `prev_page`, `next_page`, `up_page`.
3. **Consider refactoring** — move neighborhood data preparation to base `ResourceUpdateManagerController` to avoid duplication.
4. **Consider UI** — button group or optional display to address theboxer's feedback on toolbar space.

The feature is valuable; the main blocker is the query optimization that Mark-H requested.

smg6511:
Closing due to inactivity. This proposal has had little to no movement in 5 years and may be better suited to being handled by an Extra. While some site structures might have a use case for this feature, many (if not a majority) will not.

Ibochkarev:
@smg6511 I can take on and complete this improvement. Is that okay?

PR #15467, opened 2021-03-05
Fix modElement getTag error (#15172)

### LINKED ISSUE #15172: modElement getTag function raises fatal error
### Step to reproduce
create plugin which is invoked on "OnBeforeEmptyTrash" and "OnEmptyTrash" and returns empty string. now delete one or more resources and purge them in the trash manager.

### Observed behavior
The trash purge processor invokes events "OnBeforeEmptyTrash" and "OnEmptyTrash". if a plugin is triggered by the event, the parameters will pass through as properties to the getTag-function of modElement. My example plugin only returns empty string. "invokeEvent" function of modx calls the process function of modScript which calls parent process function and thats modElement. modElement process function calls the getTag function, which raises the error:

### Expected behavior
no errors to expect

### Related issue(s)/PR(s)
#14689

### Environment
MODX 3.x

wax100:
### What does it do?
Fix issue #15172 . Check if value is_object and set is to empty string

### Why is it needed?
Because earlier this cause fatal error 

### How to test
 
Step to reproduce
create plugin which is invoked on "OnBeforeEmptyTrash" and "OnEmptyTrash" and returns empty string. now delete one or more resources and purge them in the trash manager.

Observed behavior
The trash purge processor invokes events "OnBeforeEmptyTrash" and "OnEmptyTrash". if a plugin is triggered by the event, the parameters will pass through as properties to the getTag-function of modElement. My example plugin only returns empty string. "invokeEvent" function of modx calls the process function of modScript which calls parent process function and thats modElement. modElement process function calls the getTag function, which raises the error:

### Related issue(s)/PR(s)
#15172

### DIFF
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -275,14 +275,13 @@ class modElement extends modAccessibleSimpleObject
                     if (is_scalar($value)) {
                         $propTemp[$key] = $key . '=`' . $value . '`';
                     } else {
-                        if(is_array($value)){
-                            foreach($value as $v){
-                                if(is_object($v)){
-                                    $value = '';
-                                }
-                            }
+                        try {
+                            $propTemp[$key] = $key . '=`' . (is_array($value) ? md5(serialize($value))
+                                    : md5(uniqid(rand(), true))) . '`';
                         }
-                        $propTemp[$key] = $key . '=`' . (is_array($value) ? md5(serialize($value)) : md5(uniqid(rand(), true))) . '`';
+                        catch (\Throwable $handlerException) {
+                            $propTemp[$key] = $key . '=`' . md5(uniqid(rand(), true)) . '`';
+                        }
                     }
                 }
                 if (!empty($propTemp)) {
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -275,6 +275,13 @@ class modElement extends modAccessibleSimpleObject
                     if (is_scalar($value)) {
                         $propTemp[$key] = $key . '=`' . $value . '`';
                     } else {
+                        if(is_array($value)){
+                            foreach($value as $v){
+                                if(is_object($v)){
+                                    $value = '';
+                                }
+                            }
+                        }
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
@opengeek I would like your opinion on this fix.

core/src/Revolution/modElement.php
```diff
@@ -275,6 +275,13 @@ public function getTag()
...
                         $propTemp[$key] = $key . '=`' . $value . '`';
                     } else {
+                        if(is_array($value)){
+                            foreach($value as $v){
+                                if(is_object($v)){
+                                    $value = '';
+                                }
+                            }
+                        }
                         $propTemp[$key] = $key . '=`' . (is_array($value) ? md5(serialize($value)) : md5(uniqid(rand(), true))) . '`';
```
theboxer: Wouldn't it be better to wrap this line with `try/catch` and use the `md5(uniqid(rand(), true)))` in the catch block?

opengeek:
This sounds like a bug in a plugin, not in modx. Or what am I missing? Why would you be passing a PDO object?!

core/src/Revolution/modElement.php
opengeek: I'm going to concur with @theboxer here.

wax100: Please check my last commit, I test it and look like no 500 error appear

theboxer (approved):

Mark-H:
@opengeek If I understand correctly, there's a `modX`/`xPDO` reference being passed around which causes the error, it just needs a custom plugin to listen to the event to find it. The plugin itself doesn't have to do or provide anything.

opengeek:
> @opengeek If I understand correctly, there's a `modX`/`xPDO` reference being passed around which causes the error, it just needs a custom plugin to listen to the event to find it. The plugin itself doesn't have to do or provide anything.

That sounds like a mistake in the event call then.

opengeek:
In any case, if it is a database class that is being passed through this, we should consider checking if it is an xPDOObject derivative and using the toArray() method (without triggering lazy loading) to run through the serialize call.

JoshuaLuckers:
Improved fix for this is merged  PR #15809

PR #15470, opened 2021-03-05
Allow MODX Setup work in cases when instance running behind the proxy

alroniks:
### What does it do?
Allows MODX Setup work in cases when instance running behind the proxy. 

### Why is it needed?
The problem in detecting the security of the request from server variables. When used proxy, usually proxy (or load balancer) does work about resolving certificates, and after it, they send traffic over HTTP internally. In this case, browser loads the website over HTTPS, but MODX does know nothing about it, and sets `base_url` with `http` in the beginning. 

These days almost all proxies and load balancers pass additional headers under `HTTP_X_FORWARDED_` namespace, so need to check them as well, to be sure, that request passed via a secured channel.

In my case, I use docker to running services, I have php-fpm, nginx, and MySQL containers, which works under one docker-composer configuration, but do not have a nightmare with ports, I use Traefik as a load balancer, which detects hosts automatically from docker-compose files and pass traffic to the proper container. 

I guess additional context would be good.

The issue starts [here](https://github.com/modxcms/revolution/blob/3.x/setup/includes/request/modinstallrequest.class.php#L62), when it set `base_url` by the constant `MODX_SETUP_URL`. And contact defined in [bootstrap.php](https://github.com/modxcms/revolution/blob/3.x/setup/provisioner/bootstrap.php#L27), where I added this additional checks.

### How to test
Run MODX on any server via HTTP, and setup over it another server with HTTPS. The best way is docker and Traefik, also the case is using AWS ALB/ELB.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/setup/provisioner/bootstrap.php
+++ b/setup/provisioner/bootstrap.php
@@ -9,24 +9,35 @@
  */
 
 define('MODX_SETUP_INTERFACE_IS_CLI', (PHP_SAPI === 'cli'));
-define('MODX_SETUP_PHP_VERSION', phpversion());
 
-$setupPath= strtr(realpath(dirname(dirname(__FILE__))), '\\', '/') . '/';
+$setupPath = str_replace('\\', '/', realpath(dirname(__FILE__, 2))) . '/';
 define('MODX_SETUP_PATH', $setupPath);
-$installPath= strtr(realpath(dirname(dirname(__DIR__))), '\\', '/') . '/';
+$installPath = str_replace('\\', '/', realpath(dirname(__DIR__, 2))) . '/';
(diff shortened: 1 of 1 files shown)


Mark-H (approved):

hugopeek:
@alroniks This breaks an upgrade on my localhost over http (from 2.7.2, don't know if that matters..)

Uncaught TypeError: Argument 1 passed to checkServerSecurity() must be of the type string, null given, called in /home/hugo/Localhost/dev/ff-next/setup/provisioner/bootstrap.php on line 27 and defined in /home/hugo/Localhost/dev/ff-next/setup/provisioner/bootstrap.php:18

PR #15479, opened 2021-03-05
Skip files with specified extensions

wax100:
### What does it do?
Do not show files with extensions that are set in **skipExtensions** property in media Source

### Why is it needed?
If You  automatically convert files to webp on you server You can get such problem: 
Screenshot_8

And with this setting You will see this
Screenshot_9

### How to test
Upload .webp to mediabrowser and set skipExtensions to 'webp'


meshkov:
disabling the webp is not a solution to the problem with the webp format

Ruslan-Aleev:
@meshkov As I understand it, this PR solves the problem of SKIPING EXTENSIONS, and webp is indicated here as an example.
The webp display problem is not relevant to this PR.

wax100:
Yes, U can use or not use this setting

core/src/Revolution/Sources/modFileMediaSource.php
```diff
@@ -154,6 +154,14 @@ public function getDefaultProperties()
                 'value' => '.svn,.git,_notes,nbproject,.idea,.DS_Store',
                 'lexicon' => 'core:source',
             ],
+            'skipExtensions' => [
+                'name' => 'skipExtensions',
+                'desc' => 'prop_file.skipExtensions_desc',
+                'type' => 'textfield',
+                'options' => '',
+                'value' => 'webp',
```
JoshuaLuckers: This will make media sources hide `webp` files by default.

JoshuaLuckers (changes requested):
I think `skipExtensions` is a good addition. Some notes:
- Skipping `webp` by default isn't something I prefer.
- We should add `skipExtensions` to `modFTPMediaSource` and `modS3MediaSource`.

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -1979,6 +1982,25 @@ protected function getAllowedExtensionsArray($properties = [])
...
 
+    /**
+     * @param array $properties
+     *
+     * @return array|mixed|string
+     */
+    protected function getSkipExtensionsArray($properties = [])
+    {
+        $skipExtensions = $this->getOption('skipExtensions', $properties, '');
+        if (is_string($skipExtensions)) {
```
alroniks: Options always stored as strings, so no need to check the type here. And after the following condition can be simplified.

JoshuaLuckers (approved):

Ibochkarev (changes requested):
изображение

It is necessary to add a lexicon to describe the parameter

alroniks (dismissed):
I agree that sometimes it may be useful to skip just one extension instead of enumerating all allowed in `allowedFileTypes`, but it all becomes too complicated in case several rules applied. Also, there is a `skipFiles` setting, which can be improved and empowered to handle the same request about skipping.

But my main change request is to make setting name correlating with `allowedFileTypes`, for example, name it as `deniedFileTypes` or `skippedFileTypes`.

core/src/Revolution/Sources/modMediaSource.php
wax100: Denied means that You can't upload it - but it is just skipped (hided)

alroniks: I agree with denied, but still prefer FileTypes against Extensions, because allowedFileTypes also stores extensions, let's do it in a consistent way.

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -1979,6 +1982,23 @@ protected function getAllowedExtensionsArray($properties = [])
...
+    protected function getSkipExtensionsArray($properties = [])
+    {
+        $skipExtensions = $this->getOption('skipExtensions', $properties, '');
+        if (empty($skipExtensions)) {
+            $skipExtensions = [];
+        } else {
+            $skipExtensions = explode(',', $skipExtensions);
+        }
+
+        return $skipExtensions;
```
alroniks:
```suggestion
        return !empty($skipExtensions) ? explode(',', $skipExtensions) : [];
```

And remove `if` statement.

smg6511:
Folks - I was taking a look at this and am curious: What is the use-case for allowing certain file types to be _uploaded_, but then have some subset _hidden_ by a skip rule? Or conversely _shown_ by an allow rule? Why would you not just create Sources with corresponding ACLs to control visibility (and ability to upload) for different users/groups? I guess what I'm getting at is if one needs all these settings in a Media Source, perhaps it points to possible design flaws on the end-user/developer side. 

IMO there's a lot of needlessly confusing behavior and ambiguous naming of settings going on here (not referring @wax100's submission here, but the Media Source and file handling area of modx in general).

Mark-H (approved):

opengeek:
Can we resolve any unresolved conversations before marking them as ready-for-merging?

core/src/Revolution/Sources/modMediaSource.php
Mark-H: Is this naming important to resolve in your opinion @alroniks?

alroniks: I still prefer to use proper names. But in this case, I am OK.

Issue #14712, opened 2019-09-03
Add context switch mechanism in core


sdrenth:
### Summary
The power of MODX lies in the ability to have multiple contexts in the same CMS. But isn't it weird that the core does not provide a way to switch between those multiple contexts?  

In my opinion this context switch mechanism should be included in the core without having to do some extra coding. 

### Why is it needed?
I think it should be included in the core because the power of MODX lies in the support of multiple contexts and that it would be a nice enhancement.

### Suggested solution(s)
I think the context switching mechanism could be added in the index in the webroot and perhaps include a config file (`/core/config/config.inc.php`) which could contain a domain name => context mapping.

Mark-H:
What exactly would a core context router do, technically?

Given we have routers available as extras based on domains, domain aliases, base url, domains and base urls, browser language, and maybe even more.. giving the responsibility for determining the context to extras is one of the more flexible parts of MODX.

What would a core implementation be able of accomplishing that an extra can't?

sdrenth:
@Mark-H At least provide a default mechanism to switch between contexts based on hosts/base url's. In the ideal situation you would also be able to customize the code and return the context key which you want to initialize, so you could also do stuff based on browser language etc.

It's not about what an extra can't, but I think it should be part of the core because its a basic and vital part of MODX.

I'd like to hear some more opinions about this.

JoshuaLuckers:
Related to and/or duplicate of issue #14462

Mark-H:
> In the ideal situation you would also be able to customize the code and return the context key which you want to initialize, so you could also do stuff based on browser language etc.

Do you mean you want to make it possible for an extra to handle the logic for what context to initialise... like the current situation? ;) 

With [some MODX principles in mind](https://community.modx.com/t/mab-2019-meeting-2-notes-may-2-2019/731) I don't see what principle would benefit from turning this into a core feature. 

I do appreciate some people's argument to want to put all all things into the core, but we benefit much more from the flexibility of extras. I like the analogy to rich text editors. 97% of sites use a RTE (SiteDash data, just now), but if a decade ago TinyMCE was built into the core, we may not have had Redactor, TinyMCE RTE, TinyMCE Wrapper, CKEditor, etc. If ContextRouter (which does the domain > context routing you propose should be possible in the config.inc.php) was built into the core 5 years ago, xRouting may not have been made as a more powerful improvement.

[...]

Issue #14364, opened 2019-02-11
MODX 3 - Include auto-configuration options in the installer: multilingual site, blog etc.


muzzwood:
## Feature request
Creating a new issue for this feature request that began while discussing issue #13876

### Summary
It would be great if MODX 3 could include configuration options in the installer. These options would provide a richer starting point for developing a project with MODX. For example, if selected, a multi-lingual site could be configured automatically while MODX is installed. A similar concept to what the MultiLangSetup extra does. Or if a blog is needed, it could be configured automatically using the appropriate extras which would be downloaded during setup.

### Why is it needed?
There is no one size fits all and that's especially true with how customizable MODX is. This is certainly one of it's main strengths. However, it can be daunting for newcomers to be presented with a blank slate. In addition, for MODX veterans it would be incredibly useful to have time-saving functions such as these.
I believe this would help to grow the community by making the first few steps more user friendly.

### Suggested solution(s)
I propose this is added via configuration scripts in the MODX installer. Required extras for the various setups would be downloaded on demand. For example, a multilingual site might need Babel and LangRouter installed and then contexts/settings configured.
This change to the installer only adds extra options and if not selected, a blank MODX site would be installed exactly the way it is currently.

- What do you, the community, think about the proposition?
- If for it, what configuration setups should be included?
- Which extras should be used for each option?

### Related issue(s)/PR(s)
Related to issue #13876

Semdevmaster:
A very unambiguous topic, for newbies it may be relevant, but MODX as a whole is not for beginners.
For example, I do installation scripts on the MODX API for different tasks, one click and everything on a clean system is everything that is needed and configured as needed. If you make such a functional, then each preset must be carried out in the form of an interactive sequence of steps, as during the installation of the system, so that any user understands what will be changed and what will be installed, with the possibility to make adjustments at each step. Somewhere to disable the system settings that he does not need, somewhere to disable the package he does not need. And of course there must be a point - I'll set everything up myself.

muzzwood:
@arsentver I think that's a bit more ambitious than what I was initially envisioning. Not to say I'm not open to the idea of an interactive questionnaire though.. 

My original thought was just to have basic presets that can be used or not. If you wanted to alter the configuration, you could do that post installation or just not not use the options at all.

Semdevmaster:
The idea of presets is good.
[...]

PR #15480, opened 2021-03-05
Add automatic detect context to initialize #modxbughunt

ilyautkin:
### What does it do?
Add context switch mechanism in core

### Why is it needed?
MODX have ability to create multiple contexts in the same CMS. But everyone needs to switch contexts by himself. Now MODX will switch context if you add context setting `site_url` or/and `base_url`

### How to test
Create new context, add context setting `site_url` or/and `base_url`, add domains to website. To handle context switching by `base_url` you can use this nginx rule `rewrite ^/(en|nl)/(.*)$ /?cultureKey=$1&q=$2 break;`

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14712, https://github.com/modxcms/revolution/issues/14364

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -82,6 +82,15 @@ $settings['archive_with']->fromArray([
   'area' => 'system',
   'editedon' => null,
 ], '', true, true);
+$settings['auto_detect_context']= $xpdo->newObject(modSystemSetting::class);
+$settings['auto_detect_context']->fromArray([
+  'key' => 'auto_detect_context',
+  'value' => true,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'site',
+  'editedon' => null,
+], '', true, true);
 $settings['auto_menuindex']= $xpdo->newObject(modSystemSetting::class);
 $settings['auto_menuindex']->fromArray([
   'key' => 'auto_menuindex',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -85,7 +85,7 @@ $settings['archive_with']->fromArray([
 $settings['auto_detect_context']= $xpdo->newObject(modSystemSetting::class);
 $settings['auto_detect_context']->fromArray([
   'key' => 'auto_detect_context',
-  'value' => true,
+  'value' => false,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'site',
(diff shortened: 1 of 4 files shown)


_build/data/transport.core.system_settings.php
```diff
@@ -82,6 +82,15 @@
   'area' => 'system',
   'editedon' => null,
 ], '', true, true);
+$settings['auto_detect_context']= $xpdo->newObject(modSystemSetting::class);
+$settings['auto_detect_context']->fromArray([
+  'key' => 'auto_detect_context',
+  'value' => true,
```
Jako: It has to default to false.

Jako (dismissed):

_build/data/transport.core.system_settings.php
ilyautkin: Ok, thank you

Bruno17:
does this work without the need of a nginx rule or rewrite rule in .htaccess, like it is working with xRouting?

ilyautkin:
> does this work without the need of a nginx rule or rewrite rule in .htaccess, like it is working with xRouting?

If your contexts should be in subdirectory - `site.com/en/` and `site.com/nl/`, you need rules. If you use contexts on different domains, you do not need rules.

Bruno17:
with Babel and the langRouter plugin, we don't need any custom rules. Would it be possible to add that functionality, too?
Or would we still need a plugin, if we don't want to mess with rewrite rules?

_build/data/transport.core.system_settings.php
ilyautkin: @Jako Hi, can you see this code and mark change request as completed?

Mark-H:
I still strongly believe this should not be in core as previously discussed in #14712. The current options available in extras are much more powerful and flexible, with different solutions covering different uses cases. 

This is setting us up for having to support every possible use case in the core under the guise of "it already does x... adding y isn't so much extra work or code". 

Just look at the very first request to make it more flexible - that was already commented here before it even landed. That's going to keep happening if this is merged, and a very cursory comparison with xrouting and langrouter already leaves enough to be desired about this core implementation, from the most basic caching so MODX isn't preparing every single context on every single request, to supporting domain aliases, www/no-www, automatic routing based on browser language, etc.  

The worst part about having a formal vote in reviewing contributions from people who made a real effort in good faith, is having to say no to an idea. Luckily those cases are rare, but I just don't see anything that could persuade me to accept a context router into core, no matter how basic _or_ powerful it is. Plugins to handle this aspect of MODX is just much more flexible regardless of how it gets implemented. Sorry. 

I am open to discussing ways of better surfacing the different options, including picking up the idea for distributions or installing recommended extras from the setup to cover cases like this. That's also discussed in other issues. 

I'm also disappointed that despite these concerns having been raised before, making it clear the chances of this getting accepted would be _very_ slim, a company still decided to incentivize people to spend their time on building it with a bounty and forcing me in the position of having to be the bad guy. I do hope you still get the bounty for your effort, @ilyautkin.

PR #15481, opened 2021-03-05
#modxbughunt Fix displaying big SVG images in media browser

wax100:
### What does it do?
Fix displaying big SVG images (by big I mean width \ height that are set in svg)

### Why is it needed?
This allow to display svg correct
Before fix
Screenshot_10
After Fix
Screenshot_12

### How to test
Upload svg with big viewBox (like 512*512)

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2264,8 +2264,8 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
                 }
                 $filemanager_thumb_height = $this->ctx->getOption('filemanager_thumb_height', 80);
                 if($height > $filemanager_thumb_height){
-                    $height = $filemanager_thumb_height;
                     $width = ($filemanager_thumb_height / $height) * $width;
+                    $height = $filemanager_thumb_height;
                 }
                 $file_size = [
                     'width' => $width,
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2238,6 +2238,10 @@ protected function getImageDimensions($path, $ext)
                 } else {
                     return false;
                 }
+                if($height > 80){
```
JoshuaLuckers: It's better to use `$this->ctx->getOption('filemanager_thumb_height', 80)` instead of a hardcoded value of 80.

wax100: should I also add **filemanager_thumb_height**  system setting by default or not?

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2238,6 +2238,11 @@ protected function getImageDimensions($path, $ext)
                 } else {
                     return false;
                 }
+                $filemanager_thumb_height = $this->ctx->getOption('filemanager_thumb_height', 80);
+                if($height > $filemanager_thumb_height){
+                    $height = $filemanager_thumb_height;
+                    $width = ($filemanager_thumb_height / $height) * $width;
```
JoshuaLuckers, line 2267: `$filemanager_thumb_height` and `$height` have the same value here, dividing them will always results in 1`1. Did you really mean to overwrite the `$height` value with `$filemanager_thumb_height`?

wax100: Should be better? 
```
$width = ($filemanager_thumb_height / $height) * $width;
$height = $filemanager_thumb_height;
```

JoshuaLuckers: I believe so!

Ibochkarev (approved):
Worked

изображение

Ruslan-Aleev (approved):

Issue #14337, opened 2019-02-06
Trash manager context menu only working for sudo user


travisbotello:
### Summary
Context menu to "Purge" or "Restore" single resource inside trash manager is not working when user doesn't have the sudo flag. 

### Step to reproduce
- Create a new user group
- Assign Administrator Policy access policy with all permissions enabled
- Assign new user but make sure to uncheck sudo flag
- Delete resource
- Switch to trash manager
- Try right click on a deleted resource

### Observed behavior
Context menu is not showing up via right click.

### Expected behavior
Context menu should be visible via right click.

### Environment
MODX 2.7.0 advanced

- When adding the sudo flag to this user, the context menu is showing up as expected.

wax100:
The problem is (as I understand) in line 130 , file **\core\model\modx\processors\resource\trash\getlist.class.php**
`$canEdit = $this->modx->hasPermission('edit');`
I can't find permission **EDIT**, we have only **SAVE** 
Thats why clients do not see context menu on _manager/?a=resource/trash_ page

sdrenth:
I will fix this

Ibochkarev:
Fixed  #15482

PR #15482, opened 2021-03-05
Fix trash page permission issues

sdrenth:
### What does it do?
This fixes the issues with the permission checks related to the trash manager. The context menu was missing for any non-sudo users. Also added the menu_trash permission to the Administrator access policy by default.

Screenshot 2021-03-05 at 15 35 42

### Why is it needed?
By default a non-sudo user would not have the context menu on the resource trash manager page. Also the Administrator access policy was lacking the menu_trash permission by default, so non sudo administrator users would not be able to visit the trash manager.

### How to test

1. Create a new user (non sudo)
2. Assign Administrator Policy access policy
4. Delete a resource
5. Switch to trash manager, you should be able to access this page now
6. Try right click on a deleted resource, the context menu should now be shown

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14337


Jako (approved):

Ibochkarev (approved):
изображение

PR #15484, opened 2021-03-05
Fix image rotation in media browser #modxbughunt

ilyautkin:
### What does it do?
Fixing images rotates in media browser

### Why is it needed?
If image have EXIF data and was made on rotated devise, it displays as rotated in mediabrowser. For example - [this image](https://user-images.githubusercontent.com/79689999/109482539-6b52f980-7a7e-11eb-876a-4eb0c59c3ae1.jpg). Now this images will be displayed as needed.

### How to test
Upload [this image](https://user-images.githubusercontent.com/79689999/109482539-6b52f980-7a7e-11eb-876a-4eb0c59c3ae1.jpg) in media browser.

### Related issue(s)/PR(s)
https://github.com/Sterc/revolution/issues/23

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2219,6 +2219,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             ]);
             $image = $this->ctx->getOption('connectors_url', MODX_CONNECTORS_URL) . 'system/phpthumb.php?' . $imageQuery;
         }
+
         return [
             'src' => $image,
             'width' => $width,
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2184,13 +2184,15 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             } else {
                 $absolute_path = $path;
             }
-            $exif = @exif_read_data($absolute_path);
-            if (!empty($exif) && $exif['Orientation'] >= 5) {
-                // This image was rotated
-                $new_width = $size['height'];
-                $new_height = $size['width'];
-                $size['width'] = $new_width;
-                $size['height'] = $new_height;
+            if (function_exists('exif_read_data')) {
+                $exif = exif_read_data($absolute_path);
+                if (!empty($exif) && $exif['Orientation'] >= 5) {
+                    // This image was rotated
+                    $new_width = $size['height'];
+                    $new_height = $size['width'];
+                    $size['width'] = $new_width;
+                    $size['height'] = $new_height;
+                }
             }
             // get original image size for proportional scaling
             if ($size['width'] > $size['height']) {
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2178,6 +2178,18 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             } catch (Exception $E) {
                 $timestamp = 0;
             }
+
+            if ($bases['pathIsRelative']) {
+                $absolute_path = $bases['pathAbsolute'] . $path;
+            }
+            $exif = exif_read_data($absolute_path);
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2178,6 +2178,18 @@ protected function buildManagerImagePreview($path, $ext, $width, $height, $bases
             } catch (Exception $E) {
                 $timestamp = 0;
             }
+
+            if ($bases['pathIsRelative']) {
+                $absolute_path = $bases['pathAbsolute'] . $path;
+            }
+            $exif = exif_read_data($absolute_path);
```
JoshuaLuckers: This requires the Exif extension to be enabled.

ilyautkin: if `$exif` will be empty, nothing will broke

JoshuaLuckers: Does the function `exif_read_data` exist when the Exif extension is disabled?

ilyautkin: Is this will be enough? `$exif = @exif_read_data($absolute_path);` Or should I check `function_exists`?

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2178,6 +2178,20 @@ protected function buildManagerImagePreview($path, $ext, $width, $height, $bases
             } catch (Exception $E) {
                 $timestamp = 0;
             }
+
+            if ($bases['pathIsRelative']) {
+                $absolute_path = $bases['pathAbsolute'] . $path;
+            } else {
+                $absolute_path = $path;
+            }
+            $exif = @exif_read_data($absolute_path);
```
JoshuaLuckers: What do you think about adding a `function_exists('exif_read_data')`?

ilyautkin: Yes, right, thank you =))

JoshuaLuckers: Thank you too!

Ibochkarev (approved):
https://user-images.githubusercontent.com/2138260/110201080-76959300-7e8b-11eb-97ec-9b724e370cc4.mov

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2203,6 +2219,7 @@ protected function buildManagerImagePreview($path, $ext, $width, $height, $bases
                 'wctx' => $this->ctx->get('key'),
                 'source' => $this->get('id'),
                 't' => $timestamp,
+                'ar' => 'x'
```
Jako, line 2222: What happens when exif_read_data is not available?

ilyautkin: Then we will see original croped image - in this case server will not know about rotating.

Jako: The issue I see is:
- exif_read_data does not exist. So the height and width are not swapped.
- later on the thumbnail generation is auto rotated with ar = x by what ever information.

How is this auto rotation detected in phpThumb? Can that ar property be set only if exif_read_data exists? Or better can the image be rotated directly with when $exif['Orientation'] >= 5 to avoid different behaviour there?

ilyautkin:
```
  ar = Auto Rotate: set to "x" to use EXIF orientation
       stored by camera. Can also be set to "l" or "L"
       for landscape, or "p" or "P" for portrait. "l"
       and "P" rotate the image clockwise, "L" and "p"
       rotate the image counter-clockwise.
```

If EXIF data does not exists, height and width will not swapped and phpThumb also will not rotate image.

ilyautkin: https://github.com/JamesHeinrich/phpThumb/blob/master/phpthumb.class.php#L2416-L2477

JoshuaLuckers (approved):

alroniks (approved):

PR #15487, opened 2021-03-05
Fix HTTPS check on setup bootstrap

Mark-H:
### What does it do?

Fixes the setup

### Why is it needed?

Setup was broken since merging https://github.com/modxcms/revolution/pull/15470 due to an incorrect check.

### How to test

Running the setup and seeing if it uses https or http properly.

Ideally test this on as many different environments as possible. I'm pretty sure this simplified check covers the different scenarios but testing is clearly important for changes here. 

### Related issue(s)/PR(s)

#15470

### DIFF
--- a/setup/provisioner/bootstrap.php
+++ b/setup/provisioner/bootstrap.php
@@ -15,17 +15,9 @@ define('MODX_SETUP_PATH', $setupPath);
 $installPath = str_replace('\\', '/', realpath(dirname(__DIR__, 2))) . '/';
 define('MODX_INSTALL_PATH', $installPath);
 
-function checkServerSecurity(string $param) {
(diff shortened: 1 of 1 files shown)


hugopeek:
This fixes it in my scenario (2.7.2 > latest 3.x, Nginx, PHP 7.3, http).

alroniks (approved):
Works pretty fine. Sorry for bothering by my such changes.

JoshuaLuckers:
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto for more info.

Ibochkarev (approved):
I confirm. Fixes an error during installation.

alroniks:
I am tending to merge it as it may block others.

PR #15488, opened 2021-03-05
properties for @CHUNK and @SNIPPET binding

### LINKED ISSUE #14074: Add @SNIPPET binding for generating list of TV options

Bruno17:
### What does it do?
Adds passing properties to snippets or chunks in @CHUNK and @SNIPPET bindings

### Why is it needed?
@CHUNK and @SNIPPET bindings doesn't support passing properties to snippets or chunks

### How to test
Could be tested with a binding like that: 
```@SNIPPET createOptions {"classname":"modResource","field":"pagetitle"}```

snippetcode:
```
$classname = $modx->getOption('classname',$scriptProperties,'');
$field = $modx->getOption('field',$scriptProperties,'');

$items = [];
if ($collection = $modx->getCollection($classname)){
    foreach ($collection as $object){
        $items[] = $object->get($field) . '==' . $object->get('id');
    }
}

return implode('||',$items);
```

### Related issue(s)/PR(s)
could close #14074
addition to #15454

### DIFF
--- a/core/src/Revolution/modTemplateVar.php
+++ b/core/src/Revolution/modTemplateVar.php
@@ -52,6 +52,7 @@ class modTemplateVar extends modElement
     public $bindings = [
         'FILE',
         'CHUNK',
+        'SNIPPET',
         'DOCUMENT',
         'RESOURCE',
         'SELECT',
@@ -833,11 +834,11 @@ class modTemplateVar extends modElement
         $cmd = false;
         $param = '';
         if (substr($nvalue, 0, 1) == '@') {
-            list($cmd, $param) = $this->parseBinding($nvalue);
+            list($cmd, $param, $properties) = $this->parseBinding($nvalue);
             $cmd = trim($cmd);
         }
 
-        return ['cmd' => $cmd, 'param' => $param];
+        return ['cmd' => $cmd, 'param' => $param, 'properties' => $properties];
     }
 
     /**
@@ -870,6 +871,7 @@ class modTemplateVar extends modElement
         }
         $cmd = $bdata['cmd'];
         $param = !empty($bdata['param']) ? $bdata['param'] : null;
+        $properties = !empty($bdata['properties']) ? $bdata['properties'] : [];
         switch ($cmd) {
             case 'FILE':
                 if ($preProcess) {
@@ -879,9 +881,15 @@ class modTemplateVar extends modElement
 
             case 'CHUNK': /* retrieve a chunk and process it's content */
                 if ($preProcess) {
-                    $output = $this->xpdo->getChunk($param);
+                    $output = $this->xpdo->getChunk($param, $properties);
                 }
                 break;
+                
+            case 'SNIPPET':
+                if ($preProcess) {
+                    $output = $this->xpdo->runSnippet($param, $properties);
+                }
+                break;                
 
             case 'RESOURCE':
             case 'DOCUMENT': /* retrieve a document and process it's content */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Mark-H** [comment]
Seems like the tests aren't too happy about this, from a quick glance because the properties may be null.

**Comment by Bruno17** [comment]
@Mark-H yeah, have seen that.
before trying to fix that, do have as described here:
https://docs.modx.com/current/en/contribute/code/contributors-guide#5-on-to-the-next-one
todo something like that
```
git fetch upstream 3.x
git fetch origin 3.x
git checkout 3.x
git pull upstream 3.x
```
or can I just commit and push?

**Comment by Mark-H** [comment]
You can just commit and push to your binding-properties branch to add to this PR. :)

**Comment by Jako** [comment]
You need to be locally on the `Bruno17:binding-properties` branch and add your commits there and push them to origin.

File: core/src/Revolution/modTemplateVar.php [hunk 1]
```diff
@@ -986,11 +995,27 @@ public function parseBinding($binding_string)
...
         $binding_string = trim($binding_string);
         $regexp = '/@(' . implode('|', $this->bindings) . ')\s*(.*)/is'; /* Split binding on whitespace */
+        
         if (preg_match($regexp, $binding_string, $match)) {
             /* We can't return the match array directly because the first element is the whole string */
+            
+            $regexp2 = '/(\S+)\s+(.+)/is'; /* Split binding on second whitespace to get properties */
+            
+            $properties = [];
+            if (preg_match($regexp2, $match[2] , $match2)) {
```
Jako, line 1006: Please initialize $match2.

Jako (changes requested):

Jako (approved):

core/src/Revolution/modTemplateVar.php
```diff
@@ -984,13 +993,30 @@ public function processBindings($value = '', $resourceId = 0, $preProcess = true
...
+                if (isset($match2[2])) {
+                    $props = json_decode($match2[2],true);
+                    if (is_array($props)){
```
Jako: Maybe it is useful to check for json_last_error instead and log the issue somehow.

See this method from https://stackoverflow.com/questions/6041741/fastest-way-to-check-if-a-string-is-json-in-php

```
function json_validate($string)
{
    // decode the JSON data
    $result = json_decode($string);

    // switch and check possible JSON errors
    switch (json_last_error()) {
        case JSON_ERROR_NONE:
            $error = ''; // JSON is valid // No error has occurred
            break;
        case JSON_ERROR_DEPTH:
            $error = 'The maximum stack depth has been exceeded.';
            break;
        case JSON_ERROR_STATE_MISMATCH:
            $error = 'Invalid or malformed JSON.';
            break;
        case JSON_ERROR_CTRL_CHAR:
            $error = 'Control character error, possibly incorrectly encoded.';
            break;
        case JSON_ERROR_SYNTAX:
            $error = 'Syntax error, malformed JSON.';
            break;
        // PHP >= 5.3.3
        case JSON_ERROR_UTF8:
            $error = 'Malformed UTF-8 characters, possibly incorrectly encoded.';
            break;
        // PHP >= 5.5.0
        case JSON_ERROR_RECURSION:
            $error = 'One or more recursive references in the value to be encoded.';
            break;
        // PHP >= 5.5.0
        case JSON_ERROR_INF_OR_NAN:
            $error = 'One or more NAN or INF values in the value to be encoded.';
            break;
        case JSON_ERROR_UNSUPPORTED_TYPE:
            $error = 'A value of a type that cannot be encoded was given.';
            break;
        default:
            $error = 'Unknown JSON error occured.';
            break;
    }

    if ($error !== '') {
        // throw the Exception or exit // or whatever :)
        exit($error);
    }

    // everything is OK
    return $result;
}
```

Jako: Such a validation method can go into in a global (helper) class.

Bruno17: yeah, testing and logging all possible mistakes. like valid json or not existing snippets, chunks...  could be an additional feature request

Jako: Until then you can check for json_last_error and log a debug message with 'invalid JSON'.

Bruno17: check added

Jako (approved):

JoshuaLuckers (approved):
It would be great if we could use the property sets functionality, but for now LGTM!

PR #15489, opened 2021-03-05
Removed TV with "list-multiple-legacy" type

Ruslan-Aleev:
### What does it do?
This `list-multiple-legacy` type TV has been deprecated since the first versions of the 2.x branch, it's time to remove it :)

### Related issue(s)/PR(s)
N/A


Ibochkarev (approved):

JoshuaLuckers:
Please make sure the deleted files are also deleted when upgrading from a previous version.
See: https://github.com/modxcms/revolution/blob/aba62d85d1bd8728534fc9810a8a822d92dfd721/setup/includes/upgrades/common/3.0-cleanup-files.php

Ruslan-Aleev:
Yes, I didn't think about it, thanks.

setup/includes/upgrades/common/3.0.0-update-tvs-list-legacy.php
```diff
@@ -0,0 +1,21 @@
...
+$tvs_list_legacy = $modx->getCollection(modTemplateVar::class, ['type' => 'list-multiple-legacy']);
+
+if ($tvs_list_legacy) {
+
+    foreach ($tvs_list_legacy as $tv_list_legacy) {
+        $tv_list_legacy->set('type', 'listbox-multiple');
+        $tv_list_legacy->save();
+    }
+
+}
```
alroniks: new line missed

setup/includes/upgrades/common/3.0.0-update-tvs-list-legacy.php
```diff
@@ -0,0 +1,21 @@
...
+ * @var modX $modx
+ * @package setup
+ */
+
+use MODX\Revolution\modTemplateVar;
+
+/** @var modTemplateVar $tvs_list_legacy */
+$tvs_list_legacy = $modx->getCollection(modTemplateVar::class, ['type' => 'list-multiple-legacy']);
+
+if ($tvs_list_legacy) {
```
alroniks: Make sense to check it first by `$modx->getCount(modTemplateVar::class, $condition)` and after do replacing. Just to avoid wrapping code in the if statement.

setup/includes/upgrades/common/3.0.0-update-tvs-list-legacy.php
```diff
@@ -0,0 +1,21 @@
...
+ * Common upgrade script for modify modTemplateVar with 'list-multiple-legacy' input type
+ *
+ * @var modX $modx
+ * @package setup
+ */
+
+use MODX\Revolution\modTemplateVar;
+
+/** @var modTemplateVar $tvs_list_legacy */
+$tvs_list_legacy = $modx->getCollection(modTemplateVar::class, ['type' => 'list-multiple-legacy']);
```
alroniks: Let's use PSR-12 for a new code. 
```suggestion
$legacyMultipleList = $modx->getCollection(modTemplateVar::class, ['type' => 'list-multiple-legacy']);
```

alroniks (changes requested):

setup/includes/upgrades/common/3.0.0-update-tvs-list-legacy.php
theboxer: That doesn't make sense at all.
1. replacing `if` with a query to db which will require another if to see if there's something seems like a lot more work to do
2. getCollection will return an array here, so the `if` is just unnecessary :)

JoshuaLuckers (approved):

alroniks (changes requested):
Also need to remove templates with legacy list.

https://github.com/modxcms/revolution/blob/3.x/manager/templates/default/element/tv/renders/input/list-multiple-legacy.tpl
https://github.com/modxcms/revolution/blob/3.x/manager/templates/default/element/tv/renders/inputproperties/list-multiple-legacy.tpl

alroniks (approved):

Issue #15415, opened 2021-02-24
[PHP8, 2.x] modLexicon->parse() throws fatal error if $params contains nested array


Mark-H:
### Summary

If the lexicon $params contains an error, PHP 8 throws:

```
Exception: str_replace(): Argument #2 ($replace) must be of type string when argument #1 ($search) is a string
```

### Step to reproduce

For now I haven't found a core place where this goes wrong (stumbled across it testing Commerce), but it's reproducible with the following in a snippet:

```
$modx->lexicon->load('core:default');
return $modx->lexicon('resource', ['key' => 'value', 'properties' => ['more' => 'values']]);
``` 

The key has to be valid and loaded for it to kick into the str_replace; it does not need to have [[+placeholders]] in the string.

I've gotten used to this method passing in an entire object into a lexicon (e.g. `$modx->lexicon('key', $object->toArray())`) to allow the lexicon key to access whatever is in there to provide a contextualised piece of text, however in PHP 8 that breaks if the object has any embedded arrays.  

### Observed behavior

```
Exception: str_replace(): Argument #2 ($replace) must be of type string when argument #1 ($search) is a string
```

### Expected behavior

It should return the lexicon key with any accessed values replaced. Ideally it should work the same as elements in parsing nested arrays (e.g. `[[+properties.foo.bar]]`), but as a stopgap it should make sure the value is scalar and skip arrays.

### Related issue(s)/PR(s)

N/a.

### Environment
MODX 2.8.2-dev (2.x branch as of Feb 22)

wax100:
I create snippet and place on page. But there is no errors in **error.log**
```
<?php
$modx->lexicon->load('core:default');
return $modx->lexicon('resource', ['key' => 'value', 'properties' => ['more' => 'values']]);
```

Jako:
The issue should be logged in the PHP error log and not in the MODX error log.

PR #15490, opened 2021-03-05
Flatten the lexicon parameters by dot notation

Jako:
### What does it do?
Flatten the lexicon parameters by dot notation

### Why is it needed?
modLexicon->parse() throws fatal error if $params contains a nested array

### How to test

Use the following code.

```
$this->modx->lexicon->load('core:default');
$test = $this->modx->lexicon('resource', ['key' => 'value', 'properties' => ['more' => 'values']]);
```

It will break on PHP 8 with: 

```
Exception: str_replace(): Argument #2 ($replace) must be of type string when argument #1 ($search) is a string
```

After the change the following array is sent to str_replace:

```
Array
(
    [key] => value
    [properties.more] => values
)
```

### Related issue(s)/PR(s)
#15415

### DIFF
--- a/core/model/modx/modlexicon.class.php
+++ b/core/model/modx/modlexicon.class.php
@@ -457,10 +457,33 @@ class modLexicon {
         if (!$str) return '';
         if (empty($params)) return $str;
 
+        $params = $this->_flatten($params);
         foreach ($params as $k => $v) {
             $str = str_replace('[[+'.$k.']]',$v,$str);
         }
         return $str;
+    }
+
+    /**
+     * Flattens an array by dot notation
+     *
+     * @access private
+     * @param array $array The array to flatten
+     * @return array The processed array
+     */
+    private function _flatten($array) {
+        $result = array();
+        if (is_array($array)) {
+            $iterator = new RecursiveIteratorIterator(new RecursiveArrayIterator($array));
+            foreach ($iterator as $value) {
+                $keys = array();
+                foreach (range(0, $iterator->getDepth()) as $depth) {
+                    $keys[] = $iterator->getSubIterator($depth)->key();
+                }
+                $result[ join('.', $keys) ] = $value;
+            }
+        }
+        return $result;
     }
 
     /**


core/model/modx/modlexicon.class.php
```diff
@@ -457,12 +457,35 @@ private function _parse($str,$params) {
...
 
+    /**
+     * Flattens an array by dot notation
+     *
+     * @access private
+     * @param array $array The array to flatten
+     * @return array The processed array
+     */
+    private function _flatten($array) {
+        $result = array();
```
JoshuaLuckers, line 475:
```suggestion
        $result = [];
```

Jako: Hey, we are on 2.x and not on 3.x

JoshuaLuckers: Oops, my mistake!

core/model/modx/modlexicon.class.php
```diff
@@ -457,12 +457,35 @@ private function _parse($str,$params) {
...
     }
 
+    /**
+     * Flattens an array by dot notation
+     *
+     * @access private
+     * @param array $array The array to flatten
+     * @return array The processed array
+     */
+    private function _flatten($array) {
```
alroniks, line 474: It does not make sense to use underscore to mark a method as private when we have a special keyword in the language. It would be nice also to have configured symbols, like a dot, but as it only for lexicons and private, it is not required.

Jako: I have adopted that underscore notation by the invoking method name _parse. We can add that flatten method to a core helper class, if that is needed. There we can make the separator configurable.

alroniks: It would be a nice feature for 3.x I guess, but for 2.x I am fine with the current solution.

alroniks (approved):
It works as expected. Tested in Docker over PHP 8.

I created the lexicon entry with the following content "This special entry with [[+key]] and [[+properties.more]]." and returned it in the snippet `echo $modx->lexicon('entry', ['key' => 'value', 'properties' => ['more' => 'values']]);`.

Result is:
```text
PHP: 8.0.2 // I added PHP version to the output as well
This special entry with value and values.
```

No fatal errors.

opengeek:
Is this not a breaking change?

opengeek:
It seems to me it would be better to solve this by creating a simple function to recursively process nested property arrays. But perhaps I'm missing something here.

Jako:
Only if avoiding a fatal error is a breaking change. That flatten function works recursive …

opengeek:
What caused this error then? What was passing this nested array? And how was it expected to be processed before this change? I'm a little lost here.

Jako:
It was detected and described well by @Mark-H in the linked issue.

opengeek:
Right, but what actual usage caused this? Is there an extra that does this? Is there a core call that passes a nested array? When the nested array was passed before, and there was no code to turn it to dot notation, what was the result. Or was this just an arbitrary pen test to try and create a fatal error?

opengeek:
Ok, I see where he stumbled upon this in Commerce, but technically, this is a new feature. Or am I wrong?

Jako:
The nested parsing is somehow a feature. Is this nested parsing not working the same as in normal chunks?

opengeek:
The point is, it's a new feature, and technically cannot go into a patch release. The reason for this is that an extra could depend on it and you cannot specify extras as supporting a specific minimum patch release.

Jako:
Hmm, then this has to be made configurable.

opengeek:
Or leave it as a feature for 3.x and just adjust this to skip non-scalar properties for 2.8.2.

Ibochkarev (approved):

opengeek:
@Jako Are we going to move forward with just skipping non-scalar properties so we can include this in 2.8.x?

PR #15492, opened 2021-03-06
Flatten the lexicon parameters by dot notation

Ibochkarev:
### What does it do?
Flatten the lexicon parameters by dot notation

### Why is it needed?
`modLexicon->parse() throws fatal error if $params contains a nested array`

### How to test

Use the following code.

```php
$this->modx->lexicon->load('core:default');
$test = $this->modx->lexicon('resource', ['key' => 'value', 'properties' => ['more' => 'values']]);
```

It will break on PHP 8 with:

`Exception: str_replace(): Argument #2 ($replace) must be of type string when argument #1 ($search) is a string`

After the change the following array is sent to str_replace:

``` php
Array
(
    [key] => value
    [properties.more] => values
)
```

### Related issue(s)/PR(s)
#15415 https://github.com/modxcms/revolution/pull/15490

### DIFF
--- a/core/src/Revolution/modLexicon.php
+++ b/core/src/Revolution/modLexicon.php
@@ -597,7 +597,7 @@ class modLexicon
      * @return array The processed array
      */
     private function _flatten($array) {
-        $result = [];
+        $result = array();
 
         if (is_array($array)) {
             $iterator = new RecursiveIteratorIterator(new RecursiveArrayIterator($array));
--- a/core/src/Revolution/modLexicon.php
+++ b/core/src/Revolution/modLexicon.php
@@ -597,12 +597,12 @@ class modLexicon
      * @return array The processed array
      */
     private function _flatten($array) {
-        $result = array();
+        $result = [];
 
         if (is_array($array)) {
             $iterator = new RecursiveIteratorIterator(new RecursiveArrayIterator($array));
             foreach ($iterator as $value) {
-                $keys = array();
+                $keys = [];
                 foreach (range(0, $iterator->getDepth()) as $depth) {
                     $keys[] = $iterator->getSubIterator($depth)->key();
                 }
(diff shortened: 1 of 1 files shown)


Jako:
Thanks for porting this.

core/src/Revolution/modLexicon.php
```diff
@@ -581,13 +583,37 @@ private function _parse($str, $params)
...
     }
 
+    /**
+     * Flattens an array by dot notation
+     *
+     * @access private
+     * @param array $array The array to flatten
+     * @return array The processed array
+     */
+    private function _flatten($array) {
```
alroniks, line 601: I agree with such kind of solution for 2.x, but for 3.x it should be done as a feature that allows making array a flat. And for this case need to reuse such a helper.

alroniks (changes requested):

core/src/Revolution/modLexicon.php
Ibochkarev: @Jako Can you help with this?

Ibochkarev: @alroniks Can you do this in a separate PR? Since I myself will not be able to implement it.

sergant210: @alroniks I think it's enough to remove the underscore from the beginning of the name. Where to move this function? To the MODX class? He's  already a God class. We need a specified service for these helpers. But it hasn't been created yet.

JoshuaLuckers: I'm in favor of @sergant210's proposal for now, this is blocking a bug and we can always add a helper function down the road. What do you think @alroniks ?

Jako: @Ibochkarev Can you please just remove the underscore from the method name at both places?

Mark-H (approved):

Issue #15328, opened 2020-11-10
File browser not opening folders with special chars


ezrarieben:
### Summary
When trying to open a folder within MODX's file browser that contains one of the following special chars:
* ä
* ö
* ü

The folder does not expand and there is no hover effect on mouseover.

### Step to reproduce
1. Create a folder named "Porträt"  using the MODX file manager .
2. Place a test file in the folder using FTP or another type of directory access (not using MODX).
3. Refresh the file manager in MODX and try to expand the folder "Porträt".

### Observed behavior
Any folder with a special char (out of the list above) in it's name does not expand and has no hover effect on mouseover.

### Expected behavior
The folder should expand, the files within the folder should be visible and the hover effect should be visible.

### Related issue(s)/PR(s)
N/A

### Environment
* MODX: 2.8.1
* Apache: 2.4
* MySQL: 5.7
* Browser: Edge (Version 86.0.622.51) / Chrome (Version 86.0.4240.193)

Askalab:
I have exactly the same issue with French chars : "éèàêâùô"...

So : any "non-english" char !

Ruslan-Aleev:
Related with https://github.com/modxcms/revolution/issues/15271 & https://github.com/modxcms/revolution/issues/14264 & https://github.com/modxcms/revolution/issues/12932 & https://github.com/modxcms/revolution/issues/13412
Posibly related with https://github.com/modxcms/revolution/issues/14320 & https://github.com/modxcms/revolution/issues/13855 & https://github.com/modxcms/revolution/issues/15682 & https://github.com/modxcms/revolution/issues/14508

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/filemanager-cant-load-folders-with-umlauts-in-2-8/3698/1

Ibochkarev:
Fixed https://github.com/modxcms/revolution/pull/15505

frischnetz:
I don't think, this is resolved in 2.8.3, yet. I still have the same issues with umlauts in file/folder names. 

In https://community.modx.com/t/filemanager-cant-load-folders-with-umlauts-in-2-8/3698/1 it was suggested to remove line 65 from `core/model/modx/processors/browser/directory/getlist.class.php`:

`$item['id'] = htmlentities($item['id'], ENT_QUOTES, 'UTF-8');`

When I try this for testing purposes, this issue is indeed resolved. But I doubt that this is the best solution.

[...]

Issue #15271, opened 2020-10-03
Incorrect display of files or directories in the file tree, if the name contains +


Ruslan-Aleev:
### Summary
Incorrect display of files or directories in the file tree if the directory or file name contains `+`:
- Files are no longer displayed in the file tree if they are inside a directory with a “+” in the name (although the file is in the directory).
- And the preview for the picture also ceases to be displayed if there is a “+” in the name of the picture.

However, there are no such problems in the MODX file browser.
There is no bug in MODX3 either.

More information here -https://community.modx.com/t/creating-a-directory-which-ends-with-a-doesnt-load-in-modx-dashboard/3135

Possible explanation:
> Maybe it has to do with the fact, that a plus sign in the query string of a url represents a space.
If the manager makes a call to .../connectors/index.php?action=browser/directory/getList&id=gl+/&type=dir, the id-parameter is possibly interpreted as gl plus a space.

### Environment
MODX 2.7.3

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/creating-a-directory-which-ends-with-a-doesnt-load-in-modx-dashboard/3135/5

Ruslan-Aleev:
@rthrash Yes, I indicated this in the description of the issue :)

Mark-H:
Pssst, @rthrash is a bot. Don't tell anyone tho.

pixelstuff:
Possibly related. If you create a folder named [[++holidaytheme]] in the file system the MODx folder tree (left column) will create an infinitely nested loop of those folders.

![image](https://user-images.githubusercontent.com/428855/97478165-f41e2500-1926-11eb-9b73-c505512e39f6.png)

https://community.modx.com/t/file-folder-looping-or-something-bug/3253

alroniks:
It fixed by merged PR

PR #15505, opened 2021-03-06
Fix bug with special chars in the directory or file names

sergant210:
### What does it do?
Fixed bug with incorrect display of files or directories in the file tree if the name contains special chars and "+" symbol.

### Why is it needed?
The file manager works incorrect.

### How to test
1. Open file manager.
2. Create a folder named "Porträt" or "+Portrait" using the MODX file manager .
3. Place a test file in the folder using FTP or another type of directory access (not using MODX).
4. Refresh the file manager in MODX and try to expand the folder "Porträt".

### Related issue(s)/PR(s)
#15271, #15328, #15508


alroniks (approved):
It works.

Issue #15500, opened 2021-03-06
User not found. when trying to create a message


Ibochkarev:
### Summary

User not found. when trying to create a message

### Step to reproduce
Go to the User menu, then select Messages. Create new message

### Observed behavior

![image](https://user-images.githubusercontent.com/2138260/110199352-3847a600-7e82-11eb-8f27-d6597acde0ac.png)

![image](https://user-images.githubusercontent.com/2138260/110199333-282fc680-7e82-11eb-904d-f8cbe3146c6a.png)

### Expected behavior
The message must be sent

### Related issue(s)/PR(s)
NA

### Environment
MODX Revolution 3.0.0-alpha3 (git), PHP Version 7.4.12

sergant210:
I'll take it.

sergant210:
The same for role.

alroniks:
Fixed

PR #15511, opened 2021-03-07
Fix issue when sending an internal message to the user

sergant210:
### What does it do?
Fixed a bug with the inability to send a message to the user. Added the option to send messages to user group.

### Why is it needed?
Error "User not found" occurs when trying to send a message to a user or role ("Role not found").

### How to test
Open Messages interface and try to create a message to any user or role.

### Related issue(s)/PR(s)
#15500.

### DIFF
--- a/core/src/Revolution/Processors/Security/Message/Create.php
+++ b/core/src/Revolution/Processors/Security/Message/Create.php
@@ -74,15 +74,12 @@ class Create extends ModelProcessor
     public function getMailingObject($primaryKey, $objectClass, $objectType)
     {
         $id = (int)$this->getProperty($primaryKey);
-        if (!$id) {
+
+        if (!$id || !is_numeric($this->getProperty($primaryKey))) {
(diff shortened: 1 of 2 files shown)


Ibochkarev (approved):
https://user-images.githubusercontent.com/2138260/110237273-4de1cc00-7f65-11eb-8cd2-dbead244a44f.mp4

alroniks:
I suggest applying such a fix in 2.x because it is also broken there. And just remove messages in 3.x at all. What do you think @opengeek?

sergant210:
@alroniks There is no problem in 2.x.

Ibochkarev:
> I suggest applying such a fix in 2.x because it is also broken there. And just remove messages in 3.x at all. What do you think @opengeek?

I propose to add this fix, and then in the future it will be possible to take out as a separate component

alroniks (approved):

PR #15460, opened 2021-03-05
Fixing media browser issues (#15129)

wax100:
### What does it do?
fix downloding file name (earlier was index)
fix upload file window (destroy and create new instead of creating, creating, creating...)
fix deleting more then 1 file

### Related issue(s)/PR(s)
issue #15129

PR #15512, opened 2021-03-07
Incorrect regular expression for the downloaded file.

sergant210:
### What does it do?
Fixed incorrect regular expression for a downloaded file.

### Why is it needed?
When trying to download a file an error occurs 
```
[2021-03-07 13:38:00] (ERROR @ ...\core\src\Revolution\Processors\Browser\File\Download.php : 50) PHP warning: preg_replace(): Compilation failed: invalid range in character class at offset 4
```

### How to test
Open the media manager and try to download any file. Then open the log error.


Ruslan-Aleev:
Related with https://github.com/modxcms/revolution/pull/15460

Jako (approved):

JoshuaLuckers (approved):

PR #15517, opened 2021-03-09
Decode the node id before passing it as directory to quick create file

### LINKED ISSUE #15513: Incorrect display of the file creation path
### Summary

Incorrect display of the file creation path 

### Step to reproduce

Go to the Files tab, right-click on the assets folder and select Quick Create File from the menu.

https://user-images.githubusercontent.com/2138260/110241473-2bf34400-7f7b-11eb-83aa-a8bd4de1ae75.mp4

### Observed behavior
image

### Expected behavior
Should be `assets/` instead of `assets%2F`

### Related issue(s)/PR(s)
NA

### Environment

theboxer:
### What does it do?
Decode the path before passing it to the window.

### Why is it needed?
Directory path appears url encoded in the quick create file window.

### How to test
Described in #15513

### Related issue(s)/PR(s)
#15513


Ibochkarev (approved):
изображение

JoshuaLuckers (approved):

PR #15523, opened 2021-03-09
Declare modX as final and so it cannot be extended

JoshuaLuckers:
### What does it do?
Mark modX as final 

### Why is it needed?
It prevents extending the modX class. 

With this PR I want to re-open the discussion that started in PR #14574 and gather community input.

### Related issue(s)/PR(s)
PR #14574

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -39,7 +39,7 @@ use xPDO\Om\xPDOObject;
  *
  * @package modx
  */
-class modX extends xPDO {
+final class modX extends xPDO {
     /**
      * The parameter for when a session state is not able to be accessed
      * @const SESSION_STATE_UNAVAILABLE


JoshuaLuckers:
> Here's an interesting read about `final`: https://matthiasnoback.nl/2018/09/final-classes-by-default-why/

opengeek:
I don't think this is a good idea, at all. With a more modern codebase, it makes sense. But I do not think it does in this case.

JoshuaLuckers:
> I don't think this is a good idea, at all. With a more modern codebase, it makes sense. But I do not think it does in this case.

See this as a opportunity to share how we can accomplish this 😁

opengeek:
> > I don't think this is a good idea, at all. With a more modern codebase, it makes sense. But I do not think it does in this case.
> 
> See this as a opportunity to share how we can accomplish this 😁

I understand the reasoning, but people have already come to work with modX by extending it and in some cases, it is the only way to change any behavior internal to modX or xPDO. Once we can undo that mess in a properly architected system, then this principle will be essential. That's all I'm getting at.

alroniks:
I agree with Jason. Now it possible to change something only by extending. I like the idea, but not now.

JoshuaLuckers:
What I try to say is, what are the steps we need to take so we can actually do this? I know it isn't something short term but doing it one step at a time is great.

_I almost hit the squash and merge button by accident 🙈_

Ibochkarev:
> What I try to say is, what are the steps we need to take so we can actually do this? I know it isn't something short term but doing it one step at a time is great.

We ourselves feel that everything we do is a drop in the ocean. But the ocean will be smaller without this drop.
© Mother Teresa

This means that in small steps we are closer to the goal.

PR #15525, opened 2021-03-09
New keep a changelog format

Ibochkarev:
### What does it do?
New keep a changelog format

### Why is it needed?
https://keepachangelog.com/en/1.0.0/

### Related issue(s)/PR(s)
NA


opengeek:
@Ibochkarev — I appreciate you taking a stab at this. I will probably make some additional changes to this before merging in.

Jako:
Thanks! The `### Removed` entries are moved to the section above (if I'm not wrong).

Ibochkarev:
> Thanks! The `### Removed` entries are moved to the section above (if I'm not wrong).

Example from the source site:

```
## [0.0.4] - 2014-08-09
### Added
- Better explanation of the difference between the file ("CHANGELOG")
and its function "the change log".

### Changed
- Refer to a "change log" instead of a "CHANGELOG" throughout the site
to differentiate between the file and the purpose of the file — the
logging of changes.

### Removed
- Remove empty sections from CHANGELOG, they occupy too much space and
create too much noise in the file. People will have to assume that the
missing sections were intentionally left out because they contained no
notable changes.
```

alroniks:
It would be nice to use https://www.conventionalcommits.org/en/v1.0.0/ and have autogenerated changelog and version but seems not in the nearest time.

sergant210:
@alroniks Not in this life ))

alroniks:
I am wondering if we change the changelog format to use some kind of markdown syntax, does it make sense to rename the file to changelog.md (need to update some build scripts as well), to be able to read the file directly from GitHub with proper formatting?

JoshuaLuckers:
> @Ibochkarev — I appreciate you taking a stab at this. I will probably make some additional changes to this before merging in.

@opengeek is this still on your radar?

opengeek:
It is, I just have been swamped dealing with other issues and PRs that are a bit higher priority. I'll circle back around, hopefully before final 3.0 release or immediately thereafter.

JoshuaLuckers:
ping @opengeek

opengeek:
While I'm okay with changing the format to markdown, I don't see doing that as creating much value. I am not keen on how much friction this specific format would create when preparing the changelog for each release. If I'm being honest, we don't need any more friction in the process right now.

Jako:
Separating the release notes into `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security` is a big advantage of this normalized format. This can be prepared with tagging of the PRs in the future too. So I don't think that this format creates much friction in the current process.

opengeek:
> Separating the release notes into `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security` is a big advantage of this normalized format. This can be prepared with tagging of the PRs in the future too. So I don't think that this format creates much friction in the current process.

Is there a way to output git log by those labels? If not, then yes, it is too much friction currently.

Jako:
There is no git command to retrieve the labels of a GitHub repository. So there are two options now:

1. Put the label into the commit message like in this [proposal](https://mokkapps.de/blog/how-to-automatically-generate-a-helpful-changelog-from-your-git-commit-messages/)
2. Retrieve the labels of each GitHub PR of a changelog generated by `git log` with the [GitHub REST API](https://docs.github.com/de/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request) and sort the changelog afterwards by these labels. This has to go into a shell script.

I can help with the second option, if you tell me how the current changelog is created.

opengeek:
Currently, I pull the git log since the last release date, e.g.:

``` git log --since="April 28, 2022" --pretty=format:%s | cat```

…copy and paste the output into a text editor, remove the entries that are irrelevant, fix the wording as appropriate, then use a regex to add the ` -` in front of each entry. Then I copy and paste that into the changelog and add the release date to the version header.

I would prefer not to have these labels in the commit subject. It is hard enough squeezing in everything we need for a concise description into the length of it and including the PR number when it is from a PR.

And honestly, I can't get people to review the PRs as it is. Not to mention we have a lot of labels that are already poorly maintained as it is. There would also need to be manual intervention for any commits that are not from PRs.

Jako:
Is it ok to write a small shell- or php-script that runs after the `git log` command? It will load the result of `git log`, retrieve the labels of each line and orders them on base of some labels into the `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed` or `Security` section.

Ibochkarev:
@opengeek Since no one is particularly interested in this, I close

Issue #15502, opened 2021-03-06
Unable to view thumbs of .jpg or .png in media browser or file tree


muzzwood:
### Summary

Using the latest git version of 3.x just now, thumbs are not showing for .jpg or .png files in either the file tree or media browser.
.svg thumbs work.

I'm also getting errors in the log: 
```
[2021-03-06 15:52:18] (ERROR @ /var/www/html/core/src/Revolution/modPhpThumb.php : 312) PHP warning: Cannot modify header information - headers already sent by (output started at /var/www/html/core/src/Revolution/Sources/modMediaSource.php:1499)
```

![Screenshot from 2021-03-05 22-54-58](https://user-images.githubusercontent.com/5160368/110199731-ccbb0400-7e94-11eb-95b2-146095c706ae.png)

![Screenshot from 2021-03-06 15-56-18](https://user-images.githubusercontent.com/5160368/110199737-d3e21200-7e94-11eb-9f90-b8b230da21e1.png)

I only have the default filesystem media source. No others were created.

### Step to reproduce
Hover over an image in the file tree.

### Observed behavior
Thumbs are broken and error displays in log

### Expected behavior
Thumbs shouldn't be broken :P

### Environment
MODX 3.x latest git, nginx, php7.4, mysql8, Firefox

Ibochkarev:
Possibly related: https://github.com/modxcms/revolution/pull/15481

Ibochkarev:
@digitalpenguin Please do not use [3.x] in the title

Mark-H:
Can you please check the output of that phpthumb URL? The warning in the log indicates something was already in the output, which is most likely a php notice or warning indicating another problem.

Looking at the referenced line (modMediaSource line 1499), I'm guessing it's a notice for `url doesn't exist in array` or something like that:

```
        $properties = $this->getPropertyList();
        if ($properties['url']) {
            $src = $properties['url'] . DIRECTORY_SEPARATOR . ltrim($src, DIRECTORY_SEPARATOR);
        }
```

... which begs the question if the url is supposed to be there or how to handle that if not.

muzzwood:
`$properties` array at line 1499:

```
[2021-03-07 11:55:30] (ERROR @ /var/www/html/core/src/Revolution/Sources/modMediaSource.php : 1499) Array
(
    [basePath] => 
    [basePathRelative] => 1
    [baseUrl] => 
    [baseUrlRelative] => 1
    [allowedFileTypes] => 
[...]

PR #15526, opened 2021-03-10
Improved check to see if the key "url" exists in media source $properties array. This fixes the preview thumbnail issue.

muzzwood:
### What does it do?
Improves the check to see if the key `url` exists in the media source properties array when attempting to build a preview thumbnail.

### Why is it needed?
It fixes the problem in issue #15502 where .png and .jpg images had broken preview thumbnails in the tree and the media browser.

### How to test
Hover over an image file in the file tree or look at the thumbs in the media browser. You'll see they are no longer broken.

### Related issue(s)/PR(s)
issue #15502

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -1496,7 +1496,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
         }
 
         $properties = $this->getPropertyList();
-        if (array_key_exists('url',$properties)) {
+        if (isset($properties['url'])) {
             $src = $properties['url'] . DIRECTORY_SEPARATOR . ltrim($src, DIRECTORY_SEPARATOR);
         }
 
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -1496,7 +1496,7 @@ public function prepareSrcForThumb($src)
         }
 
         $properties = $this->getPropertyList();
-        if ($properties['url']) {
+        if (array_key_exists('url',$properties)) {
```
theboxer: Wouldn't be better to use `isset` or even `empty` check? `array_key_exists` returns `true` even when the value is `null`, which won't be a valid URL.

theboxer (changes requested):

core/src/Revolution/Sources/modMediaSource.php
muzzwood: Ah yes... updated! :)

JoshuaLuckers (approved):

theboxer (approved):

PR #13905, opened 2018-05-18
Template picker

Oetzie:
### What does it do?
The new template picker feature. Before you create a new resource you will get a window where you can choose the 'pagetitle' and preview the 'template' (if screenshot is available). This PR also includes the new 'parent field combo' and 'template field combo'.

PR #15535, opened 2021-03-11
Refactor of template-picker

muzzwood:
Only the refactoring and a few extra features is my work, everything else is credited to the original author.

### What does it do?
This is a refactor of the template picker feature by @Oetzie 

Screenshot_2021-03-16 Template BaseTemplate MODX Revolution(1)

### Why is it needed?
The original PR #13905 is dated due to the core code base being refactored.

### How to test
**Requires assets to be rebuilt
**Requires an added column `preview_file` added to the modTemplateVar `modx_site_tmplvars` table.
**Requires system setting `enable_template_picker_in_tree` (create as a yes/no boolean)

1. First edit the template in the manager, set a media source for the template and select an image to use.
2. Make sure the system setting `enable_template_picker_in_tree` is set to Yes. (create setting if not there)
2. Click on create a new resource in the resource tree. 
3. Try using the template picker when it appears.

Screenshot_2021-03-12 Template BaseTemplate MODX Revolution

In the original PR there was some modification of the parent resource picker which was applied to the settings tab of the resource edit screen. This was causing display issues of the forms so I removed it. (Unsure of what the intention behind it was)

### Related issue(s)/PR(s)
PR #13905


Ibochkarev:
@digitalpenguin Looking forward to being ready for testing. Great job. Thank!

JoshuaLuckers:
You're really doing some amazing work @digitalpenguin !

core/src/Revolution/modElement.php
```diff
@@ -1157,6 +1157,36 @@ public function isStaticSourceValidPath()
...
+    /**
+     * Get the absolute path to the preview file for this instance.
+     *
+     * @return string
+     */
+    public function getPreviewUrl() {
+        if (!empty($this->get('preview_file'))) {
+            $previewfile = $this->get('preview_file');
+
+            if ($this->get('source') > 0) {
```
digitalpenguin, line 1169: Turns out a media source needs to be assigned to the template for the preview thumb to be displayed.

muzzwood:
OK it's in a working state!
(minus the improvements suggested by Mark in the original PR)

The only thing I can't figure out at the moment is where this error is coming from:

```
[2021-03-11 17:10:04] (ERROR @ /var/www/html/core/vendor/xpdo/xpdo/src/xPDO/Om/xPDOObject.php : 227) Error 42S22 executing statement: 
Array
(
    [0] => 42S22
    [1] => 1054
    [2] => Unknown column 'MODX\Revolution\modResource.id' in 'field list'
)
```

If anyone has an idea I'd be very much appreciative. 

It doesn't seem to prevent any functionality from what I can see so far.

Jako:
The issue is created here:

```
        $c->select($this->modx->getSelectColumns($this->classKey, $this->classKey));
```

Jako:
Can it be caused by the parent ID combo?

muzzwood:
Thanks @Jako 
The second param of the `getSelectColumns()` call needed to be an alias. Due to the namespaced class names the old code wouldn't work anymore. I fixed it by exploding the class name and using the last element.

If there's a better way to get the object alias, I'm happy to change it.

Ruslan-Aleev:
Thanks for the feature. There are several comments and a question :)

1) It is better to replace the file icon with a picture icon (in template section)
2) You changed the formatting in js components a lot (indentation before `:`, comma at the end). The code looks strange now.
3) The main question. How will such resources be created (with a template preview) in manager? As I understand it, there is a "Create resource" link on the widget panel, I understand correctly or not?

muzzwood:
> Thanks for the feature. There are several comments and a question :)
> 
> 1) It is better to replace the file icon with a picture icon (in template section)
> 2) You changed the formatting in js components a lot (indentation before `:`, comma at the end). The code looks strange now.
> 3) The main question. How will such resources be created (with a template preview) in manager? As I understand it, there is a "Create resource" link on the widget panel, I understand correctly or not?

Hi @Ruslan-Aleev ,

Yeah good suggestion about the icon.

In regards to the formatting I assume that was the original author's preference. I've just copy and pasted it as is with some tweaks to get it working.
I agree it looks a bit odd in comparison to the rest of the code though. I'll have a look at reformatting it to match the rest.

As for how it functions, it opens a modal when you attempt to create a new resource via the tree. Once you've picked your settings it opens the resource edit screen with those parts done.

Jako:
Can’t you use the parent combo of the resource edit page?

muzzwood:
@Jako Yeah I thought about that and did try, but encountered some difficulties. When using the parent combo from the resource edit page the resource is just created according to where you click on the tree and ignores the value. The combo included with this PR appears to take that into account and works nicely.

Ruslan-Aleev:
Some bugs:
1) I added a picture, but it is not displayed for me, only an empty picture.
tpl_0

2) The resource type is displayed incorrectly, although there is no such problem in the drop-down list of types.
tpl_1

3) In the description under "Parent resource" the ID is indicated, but in the drop-down list the resource pagetitle are indicated, not the ID. Probably it is worth removing the description about ID.
By the way, the parent's name "Root" is not clear. When editing a resource, you can specify the parent "0" or the context "Website", but not "Root", i.e. confusion in the naming.

Suggestions:
1) Add a window with a preview of the template only by clicking on "Create resource" in the widget, not in the tree.
tpl_3

Or add a setting that allows you to turn off the template preview window.
In the current version, this is not convenient for the UX, unnecessary actions appear.

2) If the preview picture is not added for any template, then it should disable the window with the template preview.

_build/templates/default/sass/_forms.scss
```diff
@@ -1548,3 +1551,12 @@ td.x-date-mp-sep {
     background-color: rgba(0, 0, 0, .8);
   }
 }
+
+// Template preview field that opens the file browser
+#x-form-el-modx-template-preview-file {
```
Ruslan-Aleev: This line can be moved higher, and use variables, as in https://github.com/modxcms/revolution/blob/3.x/_build/templates/default/sass/_forms.scss#L479

muzzwood:
Thanks for the feedback @Ruslan-Aleev 

1. Do you have a media source set on the template? That's currently a requirement, though I'd like to make it so it's not.
2. I believe that will be fixed by this PR: https://github.com/modxcms/revolution/pull/15424
3. Good point about that text, I'll change it. Do you think "No Parent" would be better than "Root"? I feel that the context name might be a bit confusing for an end user.

Yeah a setting to turn it off/on in the tree is a good idea, I'll attempt to add that. 
I'm not sure how the check to automatically show the modal or not would work though, because even if the default/parent template didn't have an image you still might want to pick a different template when creating one, right?

Besides the visual "niceness", I think my favourite part of this feature is that because you select a template in the modal, you don't have to have the page reload when changing a template. i.e. moving from the create to update forms.
Everything related to the template is there on the resource on the initial load.

_build/data/transport.core.system_settings.php
```diff
@@ -1009,6 +1009,15 @@
   'area' => 'manager',
   'editedon' => null,
 ], '', true, true);
+$settings['enable_template_picker_in_tree']= $xpdo->newObject(modSystemSetting::class);
+$settings['enable_template_picker_in_tree']->fromArray([
+    'key' => 'enable_template_picker_in_tree',
+    'value' => true,
```
muzzwood, line 1024: I made the default value `true` but we can change that if the consensus is that it shouldn't be enabled on install.

JoshuaLuckers: It's a great feature to have enabled by default!

muzzwood: I agree, but didn't want to presume ;)

Ruslan-Aleev:
Thank you for adding the setting, although I still think that the template preview should only work in the widget :)

> I'm not sure how the check to automatically show the modal or not would work though, because even if the default/parent template didn't have an image you still might want to pick a different template when creating one, right?

Here I meant that if there are no pictures in ANY TEMPLATE, then turn off the preview window. If you are talking about a quick selection of a template, then I will write about this below.

> Besides the visual "niceness", I think my favourite part of this feature is that because you select a template in the modal, you don't have to have the page reload when changing a template. i.e. moving from the create to update forms.

But all this is already in the quick creation of the resource :) Moreover, there you can even set the content in quick form. And the problem of changing the template after adding the automatic_template_assignment setting is not so important.

Therefore, for a widget, this is the best option - it is easier to explain to the manager that if you need a preview of the template, click on this button in the widget. And there will be no confusion for the rest users, because now we have 2 similar forms of quick creation of a resource, but one has a preview of the template, and the other has a field for content, which is strange.

muzzwood:
You make good points!  :)
I took up the refactoring of this PR as I had thought this to be a sought after feature. (It was one of the bounties on the bug hunt)
I might leave this for now and see what everyone else thinks before doing any more with it.

JoshuaLuckers:
Can you solve the merge conflicts so we can give it a good test? It's a great thing to ship in the beta and gather feedback that way as well!

muzzwood:
Sure thing, will do!

Mark-H:
@gpsietzema @Oetzie I believe you've been using this in your fork, has any useful feedback come from that which should be taken into consideration?

muzzwood:
OK conflicts have been resolved.

core/src/Revolution/Processors/Element/Template/GetList.php
```diff
@@ -39,39 +40,57 @@ public function prepareQueryBeforeCount(xPDOQuery $c)
...
+                'h'             => 236,
+                'HTTP_MODAUTH'  => $this->modx->user->getUserToken($this->modx->context->get('key')),
+                'zc'            => 1
+            ));
+
+            $preview = $this->modx->getOption('connectors_url', MODX_CONNECTORS_URL) . 'system/phpthumb.php?' . urldecode($imageQuery);
+        }
+
+
+        $array  = array(
```
theboxer: The variable is unnecessary, you can just return the array, instead of storing it to variable and then returning it.

core/src/Revolution/Processors/Resource/GetList.php
```diff
@@ -31,11 +32,54 @@ class GetList extends GetListProcessor
     public $defaultSortField = 'pagetitle';
     public $permission = 'view';
 
+    public function prepareQueryBeforeCount(xPDOQuery $c) {
+        $c->leftJoin('modContext', 'Context');
```
theboxer: `'modContext'` should be `\MODX\Revolution\modContext::class`

core/src/Revolution/Processors/Resource/GetToolbar.php
```diff
@@ -31,33 +31,54 @@ public function getLanguageTopics()
...
-                'tooltip' => $this->modx->lexicon('document_new'),
-                'handler' => 'new Function("this.redirect(\"?a=resource/create' . $context . '\");");',
+                'cls'       => 'tree-new-resource',
+                'tooltip'   => $this->modx->lexicon('document_new'),
+                'handler'   => 'new Function("MODx.createResource(' . $record . ');");'
             ];
         }
         if ($this->modx->hasPermission('new_weblink')) {
+            $record = json_encode([
+                'class_key'     => 'MODX\\Revolution\\modWebLink',
```
theboxer: the `::class` construct is preferred, instead of typing the class name on your own. So it should be `\MODX\Revolution\modWebLink::class` or a combination with `use` statement and `modWebLink::class`

core/src/Revolution/Processors/Resource/GetToolbar.php
```diff
@@ -31,33 +31,54 @@ public function getLanguageTopics()
...
-                'tooltip' => $this->modx->lexicon('add_weblink'),
-                'handler' => 'new Function("this.redirect(\"?a=resource/create&class_key=modWebLink' . $context . '\");");',
+                'cls'       => 'tree-new-weblink',
+                'tooltip'   => $this->modx->lexicon('add_weblink'),
+                'handler'   => 'new Function("MODx.createResource(' . $record . ');");'
             ];
         }
         if ($this->modx->hasPermission('new_symlink')) {
+            $record = json_encode([
+                'class_key'     => 'MODX\\Revolution\\modSymLink',
```
theboxer: Same as modWebLink

core/src/Revolution/Processors/Resource/GetToolbar.php
```diff
@@ -31,33 +31,54 @@ public function getLanguageTopics()
...
-                'tooltip' => $this->modx->lexicon('add_symlink'),
-                'handler' => 'new Function("this.redirect(\"?a=resource/create&class_key=modSymLink' . $context . '\");");',
+                'cls'       => 'tree-new-symlink',
+                'tooltip'   => $this->modx->lexicon('add_symlink'),
+                'handler'   => 'new Function("MODx.createResource(' . $record . ');");'
             ];
         }
         if ($this->modx->hasPermission('new_static_resource')) {
+            $record = json_encode([
+                'class_key'     =>  'MODX\\Revolution\\modStaticResource',
```
theboxer: Same as modWebLink

manager/assets/modext/widgets/element/modx.panel.template.js
```diff
@@ -111,7 +111,7 @@ MODx.panel.Template = function(config) {
                         ,source: config.record.source != null ? config.record.source : MODx.config.default_media_source
                         ,openTo: config.record.openTo || ''
                         ,id: 'modx-template-static-file'
-                        ,triggerClass: 'x-form-code-trigger'
+                        //,triggerClass: 'x-form-code-trigger'
```
theboxer: If it's not needed, delete the line.

manager/assets/modext/widgets/element/modx.panel.template.js
```diff
@@ -389,7 +406,7 @@ Ext.extend(MODx.panel.Template,MODx.FormPanel,{
         }
     }
     ,toggleStaticFile: function(cb) {
-        var flds = ['modx-template-static-file','modx-template-static-source'];
+        /*var flds = ['modx-template-static-file','modx-template-static-source'];
```
theboxer: If it's not needed, delete the code.

manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -825,20 +825,11 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,items: [{
                 xtype: 'modx-combo-template'
                 ,fieldLabel: _('resource_template')
-                ,hideLabel: true
+                //,hideLabel: true
```
theboxer: If it's not needed, delete the line

manager/assets/modext/widgets/element/modx.panel.template.js
```diff
@@ -419,6 +436,27 @@ Ext.extend(MODx.panel.Template,MODx.FormPanel,{
                     }
                 }
             }
+        }*/
+        var fields = [
+            'modx-template-static-file',
+            'modx-template-static-file-help',
+            //'modx-template-static-source',
```
theboxer: Adding new code where half of the array is commented out doesn't make sense :) Include only what's needed.

manager/assets/modext/widgets/resource/modx.window.resource.js
```diff
@@ -0,0 +1,303 @@
...
+    MODx.window.CreateResource.superclass.constructor.call(this, config);
+};
+
+Ext.extend(MODx.window.CreateResource, MODx.Window);
+
+Ext.reg('modx-window-create-resource', MODx.window.CreateResource);
+
+MODx.panel.TemplatePicker = function(config) {
+    config = config || {};
+    //console.log(config);
```
theboxer: Delete the line

manager/assets/modext/widgets/resource/modx.window.resource.js
```diff
@@ -0,0 +1,303 @@
...
+        cls: 'x-form-template-preview'
+    });
+
+    MODx.panel.TemplatePreview.superclass.constructor.call(this, config);
+};
+
+Ext.extend(MODx.panel.TemplatePreview, Ext.Panel, {
+    setPreview: function(record) {
+        this.removeAll();
+        //console.log(record);
```
theboxer: delete the line

theboxer (dismissed):

muzzwood:
Thanks for the review @theboxer !
I've made the changes requested.

muzzwood:
> Can’t you use the parent combo of the resource edit page?

@Jako Got this working

Ibochkarev (commented):
@muzzwood  All OK. But - for some reason, there are no changes from this PR https://github.com/modxcms/revolution/pull/15593. Therefore, I had to add it manually for testing.

Jako (approved):
After adding modx.window.resource.js to modx.jsgrps-min.js everything works fine.

Jako:
@cla-bot check

setup/includes/upgrades/common/3.0.0-template-preview.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @var modX $modx
+ * @package setup
+ */
+
+/* add modTemplate.preview_file field */
+$class = 'modTemplate';
+$table = $modx->getTableName($class);
+
+$description = $this->install->lexicon('add_column',array('column' => 'preview_file','table' => $table));
```
Ibochkarev:
```suggestion
$description = $this->install->lexicon('add_column', array('column' => 'preview_file','table' => $table));
```

setup/includes/upgrades/common/3.0.0-template-preview.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
Ibochkarev, line 1:
```suggestion
<?php

```

Ibochkarev (changes requested):

Ibochkarev:
@cla-bot check

muzzwood:
I signed the CLA both as my old username digitalpenguin and my new username muzzwood. (digitalpenguin is now an organisation)
I changed my username after I made this PR though, perhaps that's why the bot is having trouble with it?

Mark-H:
@cla-bot check

core/src/Revolution/Processors/Resource/GetList.php
```diff
@@ -31,11 +32,54 @@ class GetList extends GetListProcessor
...
+                'OR:longtitle:LIKE' => '%' . $query . '%'
+            ));
+        }
+
+        $ignore = $this->getProperty('ignore');
+
+        if (!empty($ignore)) {
+            $c->where(array(
+                'id:NOT IN' => explode(',', $ignore)
+            ));
```
JoshuaLuckers:
```suggestion
            $c->where([
                'id:NOT IN' => explode(',', $ignore)
            ]);
```

core/src/Revolution/Processors/Resource/GetList.php
```diff
@@ -31,11 +32,54 @@ class GetList extends GetListProcessor
...
+    public function prepareQueryBeforeCount(xPDOQuery $c) {
+        $c->leftJoin(\MODX\Revolution\modContext::class, 'Context');
+
+        $query = $this->getProperty('query');
+
+        if (!empty($query)) {
+            $c->where(array(
+                'pagetitle:LIKE'    => '%' . $query . '%',
+                'OR:longtitle:LIKE' => '%' . $query . '%'
+            ));
```
JoshuaLuckers:
```suggestion
            $c->where([
                'pagetitle:LIKE'    => '%' . $query . '%',
                'OR:longtitle:LIKE' => '%' . $query . '%'
            ]);
```

core/src/Revolution/Processors/Element/Template/GetList.php
```diff
@@ -39,39 +40,54 @@ public function prepareQueryBeforeCount(xPDOQuery $c)
...
+        $preview = $object->getPreviewUrl();
+
+        if (!empty($preview)) {
+            $imageQuery = http_build_query(array(
+                'src'           => $preview,
+                'w'             => 335,
+                'h'             => 236,
+                'HTTP_MODAUTH'  => $this->modx->user->getUserToken($this->modx->context->get('key')),
+                'zc'            => 1
+            ));
```
JoshuaLuckers:
```suggestion
            $imageQuery = http_build_query([
                'src'           => $preview,
                'w'             => 335,
                'h'             => 236,
                'HTTP_MODAUTH'  => $this->modx->user->getUserToken($this->modx->context->get('key')),
                'zc'            => 1
            ]);
```

core/src/Revolution/Processors/Element/Template/GetList.php
```diff
@@ -39,39 +40,54 @@ public function prepareQueryBeforeCount(xPDOQuery $c)
...
 
-        return $objectArray;
+        return array(
+            'id'            => $object->get('id'),
+            'templatename'  => $object->get('templatename'),
+            'description'   => $object->get('description'),
+            'category_name' => $object->get('category_name'),
+            'preview'       => $preview,
+            'time'          => time()
+        );
```
JoshuaLuckers:
```suggestion
        return [
            'id'            => $object->get('id'),
            'templatename'  => $object->get('templatename'),
            'description'   => $object->get('description'),
            'category_name' => $object->get('category_name'),
            'preview'       => $preview,
            'time'          => time()
        ];
```

core/src/Revolution/Processors/Resource/GetList.php
JoshuaLuckers:
```suggestion
            $c->where([
                'pagetitle:LIKE'    => '%' . $query . '%',
                'OR:longtitle:LIKE' => '%' . $query . '%'
            ]);
```

core/src/Revolution/Processors/Resource/GetList.php
JoshuaLuckers:
```suggestion
            $c->where([
                'id:NOT IN' => explode(',', $ignore)
            ]);
```

muzzwood:
Thanks @JoshuaLuckers, I've made your suggested changes.

_build/templates/default/sass/_forms.scss
muzzwood: Resolved this by adding a triggerClass in the JS instead of adding new CSS. Thanks!

Mark-H (approved):

Mark-H:
Great work everyone -- especially @Oetzie and @muzzwood but also the many reviewers -- on getting this ready to go. I love this feature in MODX 3. 

If any issues arise with the template picker now it's merged, please make sure to open new issues/PRs. Thanks!

SnowCreative:
Got a question about this one. Will there be some way to override the template selector? For example, when setting up Collections, I often want the specify the template used for the Collection children, and don't want the user to be able to select their own template.

Mark-H:
I believe Collections sends the user to `/manager/?a=resource/create&parent=123&template=321` along with other options passed through the URL parameters, so that should be unaffected.

PR #15537, opened 2021-03-11
Add possibility to translate message about site unavailability

Ruslan-Aleev:
### What does it do?
Corrected untranslated lines:
- Removed untranslated description from Base Template
The Base Template description says "Template", but it is not useful.
- Added lexicon for `site_unavailable_message` setting
The lexicon will be replaced with the correct string depending on the `cultureKey` setting.

We can also display other settings through lexicons, for example, `phpthumb_nohotlink_text_message`, but these are technical messages and translation is not particularly needed.

### Why is it needed?
For unification in translation

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/6092

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1598,7 +1598,7 @@ $settings['site_status']->fromArray([
 $settings['site_unavailable_message']= $xpdo->newObject(modSystemSetting::class);
 $settings['site_unavailable_message']->fromArray([
   'key' => 'site_unavailable_message',
-  'value' => 'The site is currently unavailable',
+  'value' => '[[%site_unavailable_message]]',
   'xtype' => 'textfield',
   'namespace' => 'core',
   'area' => 'site',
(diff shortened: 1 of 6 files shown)


Ibochkarev (approved):
LGTM 👍 

изображение

изображение

JoshuaLuckers (approved):

alroniks:
Technically it OK, but what I see, that having a system setting with just lexicon entry does not make a lot of sense. 
I would like to remove such a setting at all and replace it with lexicon directly. 
I can imagine that somebody can use the different messages for different contexts or maybe even user groups or users (as the system setting allows), but it also does not make a lot of sense. This message is a technical notice that something wrong with a site and should be nothing more.

What do you think @opengeek and @Mark-H. Maybe I missed important use cases and I am wrong.

opengeek (changes requested):
I'd prefer not to automatically update this setting. Existing users may have customized this setting already and this would undo their customization. New users can have the lexicon value on installation as a default, but changing existing values should not be done, in my opinion.

Ruslan-Aleev:
@opengeek Yes, I agree. I corrected the PR.

alroniks (approved):

PR #15539, opened 2021-03-12
Corrected fields in the User / Account sections

Ruslan-Aleev:
### What does it do?
Corrected fields in the "User" / "Account" sections:
1) Changed icon for "photo" field
2) Made "fullname" field not required
In the "User" section, the "fullname" field is not required, but in the "Account" section, it is required. You can leave the field blank in the "User" section and save. And when you go to the "Account" section, there will be an error in the form.
3) Made "username" field required
It is required by default, but in the form it is not indicated (* not specified).
4) Displayed "username" field in "Account" section
But made it uneditable. The "username" field cannot be saved in the "Account" section. _And it is not clear whether this is a bug or it should be so (the rest of the fields are saved normally)._
5) Changed order of "fullname" field in "Account" section
In the same order as in the "User" section.

### Why is it needed?
UI/UX fix

### Related issue(s)/PR(s)
N/A


Ibochkarev (approved):
изображение

изображение

alroniks (approved):

PR #15545, opened 2021-03-13
Adjust PR template to encourage using right format to connect issues

Mark-H:
### What does it do?

Adjusts template for new PRs to suggest "Resolves #foo" format of issues.

### Why is it needed?

Allows issues to be correctly linked, which doesn't work if people only provide the number. https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

### How to test
N/a

### Related issue(s)/PR(s)
Resolves _absolutely nothing_ (but touches on conversation in governance/integrators about the project board)

(Yes I created a branch on the official repo again for a quick edit made via the GitHub UI because my fork was out of date so I couldn't easily do it a better way.)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,4 +8,4 @@ Describe the issue you are solving.
 Describe how to test the changes you made.
 
 ### Related issue(s)/PR(s)
-Let us know if this is related to any issue/pull request (see https://github.com/blog/1506-closing-issues-via-pull-requests)
+Provide any issue that's addressed by this change in the format "Resolves #<issue number>", and mention other issues/pull requests with relevant information. 


Ibochkarev (approved):

opengeek:
Thanks for this. Just a note on this, though. Adding this to the template and having it in a PR will close issues, but it will not automatically link the issue unless they are on the default branch. Once we switch the default branch to 3.x this will be less of an annoyance, but still worth noting.

opengeek (approved):

Ibochkarev:
> Thanks for this. Just a note on this, though. Adding this to the template and having it in a PR will close issues, but it will not automatically link the issue unless they are on the default branch. Once we switch the default branch to 3.x this will be less of an annoyance, but still worth noting.

For example https://github.com/marketplace/actions/auto-close-issues

Mark-H:
Ahh, I thought it did get automatically added even for 3.x but it looks like those were added manually after the fact.

JoshuaLuckers (approved):

theboxer (approved):

PR #15552, opened 2021-03-16
Set controller property in controller's config

### LINKED ISSUE #15544: MODX3 form customization not working at all
### Summary
MODX3 form customization not working at all.

### Step to reproduce
Enter form customization for any template.

### Observed behavior
None of the changes entered in form customization show up when editing resources with that template.

Is anyone else experiencing this? I tried on a converted site, but then did a clean install of alpha3-20210309, and it didn't work there either.

### Environment
MODX alpha3-20210309.

theboxer:
### What does it do?
Sets `controller` property in a config passed to the controller's class constructor. This property is checked in FC rules and was empty (unset). There might be other places where this property is used.

### Why is it needed?
FC rules don't work

### How to test
#15544

### Related issue(s)/PR(s)
Resolves #15544

### DIFF
--- a/core/src/Revolution/modManagerResponse.php
+++ b/core/src/Revolution/modManagerResponse.php
@@ -102,6 +102,7 @@ class modManagerResponse extends modResponse
                 'namespace' => $this->namespace,
                 'namespace_path' => $this->namespacePath,
                 'action' => $this->route,
+                'controller' => $this->route,
             ]);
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):
LGTM 👍 

изображение
изображение

alroniks (approved):

PR #15555, opened 2021-03-16
Improve the lexicon entry for mail_smtp_prefix

Ibochkarev:
### What does it do?

Change the lexicon entry for `mail_smtp_prefix`

### Why is it needed?

The current text is a bit misleading. This setting sets the value of SMTPSecure in PHPMailer 6. The description in PHPMailer is What kind of encryption to use on the SMTP connection.

### Related issue(s)/PR(s)

#15527

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -381,8 +381,8 @@ $_lang['setting_mail_smtp_pass_desc'] = 'The password to authenticate to SMTP ag
 $_lang['setting_mail_smtp_port'] = 'SMTP Port';
 $_lang['setting_mail_smtp_port_desc'] = 'Sets the default SMTP server port.';
 
-$_lang['setting_mail_smtp_prefix'] = 'SMTP Connection Prefix';
-$_lang['setting_mail_smtp_prefix_desc'] = 'Sets connection prefix. Options are "", "ssl" or "tls"';
+$_lang['setting_mail_smtp_prefix'] = 'SMTP Encryption';
+$_lang['setting_mail_smtp_prefix_desc'] = 'Sets the encryption of the SMTP connection. Options are "", "ssl" or "tls"';
 
 $_lang['setting_mail_smtp_autotls'] = 'SMTP Auto TLS';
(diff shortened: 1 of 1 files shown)


alroniks (approved):

core/lexicon/en/setting.inc.php
```diff
@@ -381,8 +381,8 @@
 $_lang['setting_mail_smtp_port'] = 'SMTP Port';
 $_lang['setting_mail_smtp_port_desc'] = 'Sets the default SMTP server port.';
 
-$_lang['setting_mail_smtp_prefix'] = 'SMTP Connection Prefix';
-$_lang['setting_mail_smtp_prefix_desc'] = 'Sets connection prefix. Options are "", "ssl" or "tls"';
+$_lang['setting_mail_smtp_prefix'] = 'SMTP Encryption';
```
sergant210, line 384: I would suggest keeping the old name for a logical connection to the setting name.

Jako: The setting key should be changed too. The setting that is changed with this is SMTPSecure in PHPMailer 6.

sergant210: @Jako Are you talking about the "setting_mail_smtp_prefix" key or the "SMTP Connection Prefix" name?

Jako: This MODX system setting changes the value for the PHPMailer 6 property named SMTPSecure. The text of the MODX setting is misleading. Maybe 'SMTP Connection Prefix' was the right text somewhere in the past. Now the PHPMailer property changes the encryption of the SMTP connection.

alroniks (approved):

core/lexicon/en/setting.inc.php
rthrash: I lean towards updating the Lexicon key also, and would suggest `setting_mail_smtp_secure` to match PHP Mailer. This will lead to a little less confusing for future contributors. The language of the setting is clear to me.

This (obviously) would require changing the key in the the relevant Manager page, also.

opengeek: I concur with changing the key of the lexicon and altering the setting key itself from mail_smtp_prefix to mail_smtp_secure.

opengeek (changes requested):

Jako:
This PR is superceded with #16062.

PR #15538, opened 2021-03-12
Adding linting styles

Ibochkarev:
### What does it do?
Adds basic config for stylelint with Prettier rules, AirBnB code guide and rational order.

### Why is it needed?
For style linting that not only catches errors, but also helps you adhere to conventions on coding style and practices. Also added a Git hook to run the linter only for changed files when creating a commit.

Example report:

``` shell
sass/components/_primary-button.scss
  2:3  ✖  Replace "··transition:·background-color·" with "transition:·background-color·0"   prettier/prettier     
  3:1  ✖  Delete "··"                                                                       prettier/prettier     
  3:5  ✖  Expected "background" to come before "transition" in group "Visual"               order/properties-order
  4:3  ✖  Delete "··"                                                                       prettier/prettier     
  5:1  ✖  Delete "··"                                                                       prettier/prettier     
  5:5  ✖  Expected "color" to come before "box-shadow" in group "Typography"                order/properties-order
  7:1  ✖  Replace "····" with "··"                                                          prettier/prettier     
  8:1  ✖  Delete "··"                                                                       prettier/prettier     
  9:1  ✖  Delete "··"                                                                       prettier/prettier     
 10:1  ✖  Delete "····"                                                                     prettier/prettier     
 11:5  ✖  Delete "····"                                                                     prettier/prettier     
 12:1  ✖  Delete "····"                                                                     prettier/prettier     
 12:9  ✖  Expected "color" to come before "box-shadow" in group "Typography"                order/properties-order
 13:1  ✖  Delete "··"                                                                       prettier/prettier     
 15:3  ✖  Delete "··"                                                                       prettier/prettier     
 16:1  ✖  Delete "··"                                                                       prettier/prettier     
 17:1  ✖  Delete "····"                                                                     prettier/prettier     
 18:1  ✖  Replace "········" with "····"                                                    prettier/prettier     
 19:1  ✖  Delete "····"                                                                     prettier/prettier     
 19:9  ✖  Expected "color" to come before "box-shadow" in group "Typography"                order/properties-order
 20:3  ✖  Delete "··"                                                                       prettier/prettier     
 22:1  ✖  Replace "····" with "··"                                                          prettier/prettier     
 23:1  ✖  Delete "··"                                                                       prettier/prettier     
 24:1  ✖  Delete "··"                                                                       prettier/prettier     
 25:1  ✖  Delete "····"                                                                     prettier/prettier     
 26:5  ✖  Delete "····"                                                                     prettier/prettier     
[...]

PR #15558, opened 2021-03-17
Added GitHub action editorconfig

Ibochkarev:
### What does it do?
Added GitHub action editorconfig checker

### Why is it needed?
To check compliance with the editorconfig rules when sending PR

https://github.com/editorconfig-checker/editorconfig-checker

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15538 https://github.com/modxcms/revolution/issues/14311


JoshuaLuckers:
Can't we add this to the existing workflow?

Ibochkarev:
> Can't we add this to the existing workflow?

What's wrong with a single action?

JoshuaLuckers:
A waste of resources, also two places to maintain the OS etc.

Ibochkarev:
> Can't we add this to the existing workflow?

Done

.ecrc
```diff
@@ -0,0 +1,3 @@
+{
+    "exclude": ["\\.swf$", "\\.css$"]
```
JoshuaLuckers, line 2: If there are any `.swf` files in the core we should remove it.

theboxer (approved):

Mark-H (approved):

Mark-H (commented):
Funnily enough this PR itself is preventing it from getting merged 🙈

opengeek:
Is this still needed? Isn't this covered by the phpcs checks?

Ibochkarev:
> Is this still needed? Isn't this covered by the phpcs checks?

phpcs only checks the php format, it checks the other for compliance. But after adding phpcs, it may not be so relevant already

PR #15574, opened 2021-03-22
Improved UX for action buttons

Ruslan-Aleev:
### What does it do?
Improved UX for action buttons:
1) Removed the drop-down menu, because this increased the number of clicks, bad for UX
2) Left only icons for the "Delete" and "Help" buttons (text + icon don't look very good)
These "Help" button not often used (saving space), and it will be useful for "Delete" button, there will be less mistaken clicks (you need to aim)
3) Brought buttons to one look
Previously, some forms had a drop-down menu with icons (chunks, TV), and somewhere ordinary text buttons (contexts, media sources)
4) Changed the order of the buttons
Brought to standard, sorted by frequency of use

**Before:**
buttons_before

**After:**
buttons_after

### Why is it needed?
Improved UX

### Related issue(s)/PR(s)
Partially https://github.com/modxcms/revolution/issues/14006


SnowCreative:
Hmm. Other than the fact that you have to click on the three dots, I've come to quite like having this dropdown so that the long row of buttons doesn't cover up the page title, especially now that MODX 3 is displaying path information before the title, making it even more likely that the actual title, or part of it, will get hidden:


Screen Shot 2021-03-23 at 5 29 40 PM

Is there a way to keep the dropdown but not have to click on the button for it to appear?

alroniks:
> Is there a way to keep the dropdown but not have to click on the button for it to appear?

On Desktop I would also preferably have a hovered block. It may an issue on mobile, but the current manager isn't mobile-ready, so who cares? :-D

sergant210:
About buttons. The icon is usually placed in front of the text. For example look at the Github buttons.

sergant210:
The button block overlaps the scroll bar.
Plugin test1  MODX Revolution — Mozilla Firefox

Ibochkarev:
> The button block overlaps the scroll bar.
> Plugin test1 MODX Revolution — Mozilla Firefox

It needs to be fixed

Ruslan-Aleev:
> Is there a way to keep the dropdown but not have to click on the button for it to appear?

UX also deteriorates here, because you will often miss the dot button :) (not to mention mobile devices)


> making it even more likely that the actual title, or part of it, will get hidden

Pagetitle is visible in the form below :) But yes, it can partially hide, there is a separate issue about this, unfortunately I did not find it. It might be worth adding the title attribute to the title.


> The icon is usually placed in front of the text. For example look at the Github buttons.

Well GithHub has money for a designer, it looks ok in any position :)
But, even if you move the icons to the left, then there are no icons on other sections (such as Contexts, Media Sources), will you fix this? It looks crooked now.


> The button block overlaps the scroll bar.

Create a separate issue to discuss.

Ibochkarev (approved):
Worked!

https://user-images.githubusercontent.com/2138260/114098353-0e4f2c80-98e3-11eb-9f25-607280132a1b.mp4

rthrash:
This is a really great move to consistency … which is excellent UX IMO. Kudos @Ruslan-Aleev for the effort! 

I did see a few places where the old "Help!" was being used as opposed to the circle-? icon: Package Management (around 0:37), System Settings (1:05), and Media Sources (and any sub-sections) (1:22-1:37) … those should probably be consistent.

Ruslan-Aleev:
@rthrash Thanks for the kind words :)
Yes, I missed some of the buttons in the sections at the beginning, but in the last edits (a couple of weeks ago) I corrected it.

theboxer (approved):

Jako (approved):

opengeek (approved):

Issue #15056, opened 2020-04-09
Returned values from plugins resets before each new plugin


mvoevodskiy:
### Summary
$modx->event->returnedValues resets at each new plugin. Values don't returned from plugin if next plugin will run.

### Step to reproduce
Create 2 plugins at one event, assign some value to $modx->event->returnedValues at first plugin.

### Observed behavior
Try to get this values in a code generated event.

### Expected behavior
Values must be accessible after all plugins if plugin don't reset it.

### Environment
MODX 2+

EMDM45:
I have a problem with it too

touol:
When will you fix it?
It is necessary to comment out line 2875 in core\model\modx\modx.class.php
'$this->returnedValues = null;
in function resetEventObject(){

Ibochkarev:
Open fix: https://github.com/modxcms/revolution/pull/16877 - keeps `returnedValues` across the plugin loop in `invokeEvent` instead of wiping them per plugin.

That matches the report and @touol's workaround. On current `3.x`, `modX::invokeEvent()` still calls `resetEventObject()` at the start of each plugin (`core/src/Revolution/modX.php`), and `modSystemEvent::resetEventObject()` sets `$this->returnedValues = null` (`core/src/Revolution/modSystemEvent.php`). So values from plugin A vanish before plugin B (and before anything that reads the event after the loop). Closed attempts #15057 / #15577 did not land this.

I have not run a two-plugin event harness here.

Keep open until #16877 merges (or an equivalent change).

PR #15577, opened 2021-03-23
Refactoring the invokeEvent method and the modSystemEvent class

sergant210:
### What does it do?
Refactoring the invokeEvent method and the modSystemEvent class.

### Why is it needed?
This makes the code clearer and more consistent.

### How to test
1. Make 2 plugins on the OnBeforeDocFormSave event.
Plugin 1.
```
$modx->event->output('Output from plugin 1.');
return 'return from plugin  1.';
```
Plugin 2.
```
$modx->event->output('Output from plugin 2.');
return 'return from plugin  2.';
```
2. Open any resource and click the Save button. 

### Expected result
- Get a popup window with the message "Output from plugin 2. Output from plugin 1."
- There are 2 records in the error log
```
[OnBeforeDocFormSave] return from test2.
[OnBeforeDocFormSave] return from test1.
```

In addition, you can pass values in the `$modx->event->returnedValues` property between plugins of the same event.

### Related issue(s)/PR(s)
#15056, #15057

### DIFF
--- a/core/src/Revolution/modSystemEvent.php
+++ b/core/src/Revolution/modSystemEvent.php
@@ -31,22 +31,12 @@ class modSystemEvent {
      */
     public $name = '';
     /**
-     * The name of the active plugin being invoked
-     * @var string $activePlugin
-     * @deprecated
-     */
-    public $activePlugin = '';
-    /**
      * A reference/instance of the currently processed modPlugin object
      *
      * @var modPlugin|null
-     */
-    public $plugin = null;
-    /**
-     * @var string The name of the active property set for the invoked Event
      * @deprecated
      */
-    public $propertySet = '';
+    public $plugin = null;
     /**
      * Whether or not to allow further execution of Plugins for this event
      * @var boolean $_propagate
@@ -60,6 +50,7 @@ class modSystemEvent {
     /**
      * Whether or not this event has been activated
      * @var boolean
+     * @deprecated
      */
     public $activated;
     /**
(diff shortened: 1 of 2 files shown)


core/src/Revolution/modSystemEvent.php
```diff
@@ -74,23 +65,22 @@ class modSystemEvent {
...
-     * context.  Use {@link modX::_log()} in the meantime.
-     * @param string $msg The message to display.
+     * modSystemEvent constructor.
+     * @param string $name Event name
      */
-    public function alert($msg) {}
+    public function __construct($name = null)
+    {
+        if ($name) {
+            $this->name = $name;
```
alroniks, line 74: Why not cast null to empty string like this: `$this->name = (string)$name` and avoid condition?

sergant210: For the same reason why I didn't do so `__construct(string $name = null)`. Soft check.

alroniks: The method signature is clear, it will always require a passing string, but type casting just will convert it to a string.

sergant210: I don't like this way. `(string)$array == 'Array'`.

Ibochkarev (approved):
Worked 👍 

https://user-images.githubusercontent.com/2138260/113500646-81eae580-9541-11eb-8abc-ffae4e0dcd02.mp4

opengeek:
@sergant210 — why was this closed?

Issue #14509, opened 2019-03-27
Unpublish for the "site_start" resource


Ruslan-Aleev:
### Summary
The resource that is specified in the system settings as "site_start" cannot be unpublished from the context menu in tree - an error will appear, see the picture below.

![site_start](https://user-images.githubusercontent.com/12523676/55071533-57d4d200-50a2-11e9-9167-beb2b49960c4.png)

But at the same time, the editing of this resource can be unpublished trough chekbox, and the site will open with 404, see the picture below.

![site_start_error](https://user-images.githubusercontent.com/12523676/55071534-586d6880-50a2-11e9-81dc-8a216758ef79.png)

By the way, a similar inconsistency for **deletion "site_start" resource** (the context menu in the tree / resource editing form)

### Environment
MODX 2.7.1
https://github.com/modxcms/revolution/issues/9774 - Discussed earlier, but now it works in opposite way, as it seems to me.

JoshuaLuckers:
While it's not recommended I do think it should be possible to mark any resource as unpublished if you really want to.

Ruslan-Aleev:
This is logical, yes. Then in the context menu should activate this possibility.

devlanda:
The behaviour should be consistent.

JoshuaLuckers:
I agree.

ghost:
More than once I've accidentally clicked the "unpublish" menu option, but I have never accidentally unchecked the "published" checkbox.

Ruslan-Aleev:
By the way, a similar inconsistency for **deletion "site_start" resource** (the context menu in the tree / resource editing form)

Ibochkarev:
Fixed #15579

PR #15579, opened 2021-03-23
Added checks for pages specified in 'error_page' and 'site_unavailable_page' settings

Ruslan-Aleev:
### What does it do?
The resource that is specified in the system settings as `site_start` cannot be unpublished from the context menu in tree - an error will appear. But at the same time, this resource can be unpublished trough chekbox in resource form.

The resources specified in `error_page` and `site_unavailable_page` could be deleted / unpublished both through the tree and through the resource form.

**In general, I added checks for these resources.**

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14509


Ruslan-Aleev:
In the process of working on PR, strange settings key are annoying! :)

I would like to have more logical names:
```
site_index (instead of site_start);
site_404 (instead of error_page);
site_503 (instead of site_unavailable_page).
```

ghost:
What about resource fields and database field names, like "introtext" and "summary"? I nagged about that one in the beginning until (introtext) was added next to the field label.

Ruslan-Aleev:
@sottwell Yes, it looks strange too.

Ibochkarev (approved):
изображение

изображение

изображение

изображение

core/src/Revolution/Processors/Resource/Update.php
```diff
@@ -479,18 +477,48 @@ public function checkPublishingPermissions()
      *
      * @return boolean
      */
-    public function checkForUnPublishOnSiteStart()
+    public function checkForUnPublishOnSiteSettingPages()
     {
-        $passed = true;
+        $result = true;
```
theboxer: Unnecessary variable

core/src/Revolution/Processors/Resource/Update.php
```diff
@@ -479,18 +477,48 @@ public function checkPublishingPermissions()
...
+            $result = $this->modx->lexicon('resource_err_unpublish_errorpage_dates');
+        } else if ($isSiteUnavailablePage && $deleted == 1) {
+            $result = $this->modx->lexicon('resource_err_delete_siteunavailable');
+        } else if ($isSiteUnavailablePage && $published == 0) {
+            $result = $this->modx->lexicon('resource_err_unpublish_siteunavailable');
+        } else if ($isSiteUnavailablePage && (!empty($publishDate) || !empty($unPublishDate))) {
+            $result = $this->modx->lexicon('resource_err_unpublish_siteunavailable_dates');
+        }
+
+        return $result;
```
theboxer: Change to `return true;`

core/src/Revolution/Processors/Resource/Update.php
```diff
@@ -479,18 +477,48 @@ public function checkPublishingPermissions()
...
+        }
+        if ($this->getProperty('id') == $workingContext->getOption('error_page') || $this->getProperty('id') == $this->modx->getOption('error_page')) {
+            $isSiteErrorPage = true;
+        }
+        if ($this->getProperty('id') == $workingContext->getOption('site_unavailable_page') || $this->getProperty('id') == $this->modx->getOption('site_unavailable_page')) {
+            $isSiteUnavailablePage = true;
+        }
+
+        if ($isSiteStart && $deleted == 1) {
+            $result = $this->modx->lexicon('resource_err_delete_sitestart');
```
theboxer: Follow return early pattern.

replace with `return $this->modx...`

core/src/Revolution/Processors/Resource/Update.php
```diff
@@ -479,18 +477,48 @@ public function checkPublishingPermissions()
...
+        } else if ($isSiteStart && $published == 0) {
```
theboxer: Replace all `else if` with a regular `if` and follow return early pattern.

`return $this->modx...`

core/src/Revolution/Processors/Resource/Delete.php
```diff
@@ -120,6 +123,36 @@ public function process()
...
+
+    /**
+     * Checks if the given resource is set as site_unavailable_page
+     * @return bool
+     */
+    public function isSiteUnavailablePage()
+    {
+        $workingContext = $this->modx->getContext($this->getProperty('context_key', $this->resource->get('context_key') ? $this->resource->get('context_key') : 'web'));
+        return ($this->resource->get('id') == $workingContext->getOption('site_unavailable_page') || $this->resource->get('id') == $this->modx->getOption('site_unavailable_page'));
+    }
```
theboxer: This calls for some abstraction as you are redefining these checks 3 times.

theboxer (changes requested):

Ruslan-Aleev:
@theboxer I have corrected the code. Any thoughts?

Ruslan-Aleev:
@theboxer Please check PR.

theboxer (approved):

theboxer:
Seems good to me, thanks!

opengeek (approved):

PR #15582, opened 2021-03-23
"Add" button in upper right of dashboard overlaps widgets

SnowCreative:
## What does it do?
Reduces the negative margin at the top of the "modx-dashboard" div.

## Why is it needed?
The gray area around the "Add+" button slightly overlaps the widgets:

overlap

### DIFF
--- a/manager/templates/default/css/index.css
+++ b/manager/templates/default/css/index.css
@@ -21732,7 +21732,7 @@ ul.x-tab-strip-bottom {
   display: flex;
   -ms-flex-flow: row wrap;
       flex-flow: row wrap;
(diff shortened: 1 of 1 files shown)


alroniks (approved):

Ibochkarev (approved):

alroniks:
Hm, seems it fixed directly in CSS, while it should be done in sass file.

SnowCreative:
> Hm, seems it fixed directly in CSS, while it should be done in sass file.

Yeah, I was wondering about that. I've updated the SAAS file (see reference below)

Ibochkarev:
Closed in favor #15599

PR #15583, opened 2021-03-24
Improve dashboard a11y

### LINKED ISSUE #13485: Help! Links must have discernible text
### Summary
The 2.x default theme utilizes icon fonts in an inaccessible manor:

```html
<a href="?a=help" title=""><i class="icon-question-circle icon icon-large"></i></a>
```

It would be great to doing a sweep and fix this every where in the manager, but I think it is important for at least the Help! button to be accessible.

### Step to reproduce
Open the Manager

### Observed behavior
Can't access the help link using assistive technology

### Expected behavior
Should be able to access the help link using assistive technology

### Environment
MODX 2.x

### Solution

```html
<a href="?a=help" title="" aria-label="Help!"><i class="icon-question-circle icon icon-large"></i></a>
```

Ibochkarev:
### What does it do?
Improving accessibility in the control panel

### Why is it needed?

- Fix `target="_blank"` links: a security and performance issue in web page
- Fixing the meta tag charset
- Added an attribute aria-label for buttons widgets on the home page
- Correcting color contrast in Who's Online and Recently Edited Resources panels (WCAG 2 AA)
- Added aria-label parameter for the left menu

### Related issue(s)/PR(s)
Resolves #13485


theboxer (approved):

manager/controllers/default/header.php
```diff
@@ -188,11 +188,11 @@ public function buildMenu($name, $placeholder)
                     $menu['action'] .= '&namespace='.$menu['namespace'];
                 }
                 $onclick = (!empty($menu['handler'])) ? ' onclick="'.str_replace('"','\'',$menu['handler']).'"' : '';
-                $menuTpl .= '<a href="?a='.$menu['action'].$menu['params'].'"'.( $top ? ' class="top-link"': '' ).$onclick.$title.'>'.$label.$description.'</a>'."\n";
+                $menuTpl .= '<a href="?a='.$menu['action'].$menu['params'].'"'.' aria-label="'.$menu['text'].'" '.( $top ? ' class="top-link"': '' ).$onclick.$title.'>'.$label.$description.'</a>'."\n";
```
Mark-H, line 191: Is $menu['text'] already encoded somewhere to avoid XSS insertion in the menu text?

Ibochkarev: @Mark-H I think not - how can I fix it?

Mark-H: Just wrapping the variable with htmlentities does the trick.

theboxer (changes requested):
@Ibochkarev will you be able to resolve Mark's request?

manager/controllers/default/header.php
rthrash: @Mark-H would the following work for @Ibochkarev to address the XSS concern (also adding to lines 193, 195 and 328, below)? 

`' aria-label="'.htmlentities($menu['text'], ENT_QUOTES, 'UTF-8').'" '.`

Mark-H: Yeah that's how that would normally be handled. It does look like it's also not being applied to the $title and the $description so this whole thing might be a problem. Or, perhaps it's already handled elsewhere, I haven't gone through the getCache() method reponsible for fetching menus to see if that's sanitising it.

PR #15589, opened 2021-03-25
Increases the basic required version of php

Ibochkarev:
### What does it do?
Increases the basic required version of php

### Why is it needed?

https://docs.modx.com/3.x/en/getting-started/server-requirements

The documentation states that minimum version = 7.1. This fix upgrades the version to the required

https://github.com/modxcms/revolution/blob/3.x/.github/workflows/ci.yml#L28 - Do we need tests for 7.0?

### Related issue(s)/PR(s)
NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['7.0', '7.1', '7.2', '7.3', '7.4', '8.0']
+        php-version: ['7.1', '7.2', '7.3', '7.4', '8.0']
 
     steps:
       - name: Checkout
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "vendor-dir": "core/vendor"
     },
     "require": {
-        "php": ">=7.0",
+        "php": ">=7.1",
         "xpdo/xpdo": "^3.0@dev",
         "league/flysystem": "^1.0",
         "league/flysystem-aws-s3-v3": "^1.0",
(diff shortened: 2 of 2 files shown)


sergant210:
7.1 and 7.2 are not supported now. [Supported versions](https://www.php.net/supported-versions.php).

Ibochkarev:
> 7.1 and 7.2 are not supported now. [Supported versions](https://www.php.net/supported-versions.php).

I can change to 7.3 and remove the old version of php in the ci.yml, but this needs the approval of the integrators

opengeek:
We will not be removing support for 7.1/7.2 at this time. They are still in widespread use regardless of what the core PHP maintainers say.

opengeek:
This needs the check for 7.0 removed.

opengeek (changes requested):
This needs the check for 7.0 removed.

Ibochkarev:
> This needs the check for 7.0 removed.

I will submit a commit with the 7.0 checkout removal in the Github action tests

opengeek (approved):

JoshuaLuckers:
Don't forget to change it here: https://github.com/modxcms/revolution/blob/0acc0d92fcea3fc97ab5692a8d3b55648fd6ef2c/setup/provisioner/requirements.php#L11

opengeek:
Why do we have all of this duplicated there?

opengeek:
What I mean to say is, I wonder if we can find a way to use the composer declarations in setup rather than duplicating it all.

JoshuaLuckers:
> What I mean to say is, I wonder if we can find a way to use the composer declarations in setup rather than duplicating it all.

Yes that would be great, but is the composer.json in the downloadable build?

opengeek:
> > What I mean to say is, I wonder if we can find a way to use the composer declarations in setup rather than duplicating it all.
> 
> Yes that would be great, but is the composer.json in the downloadable build?

It certainly should be.

JoshuaLuckers:
> > > What I mean to say is, I wonder if we can find a way to use the composer declarations in setup rather than duplicating it all.
> > 
> > 
> > Yes that would be great, but is the composer.json in the downloadable build?
> 
> It certainly should be.

I've downloaded the alpha-2 release from [modx.com/download](http://modx.com/download), composer.json is not included.

Issue #15590, opened 2021-03-25
Create and Quick Create Dropdowns 


ghost:
### Summary
The Create and Quick Create dropdowns don't drop down

### Step to reproduce
In the Tree, in the Resources tab, right-click to bring down the Resources menu. 

### Observed behavior
The Create and Quick Create dropdowns don't drop down.

<img width="318" alt="Screen Shot 2021-03-25 at 7 48 47" src="https://user-images.githubusercontent.com/1073780/112425947-ae338480-8d3f-11eb-818c-328fcad4b091.png">

### Expected behavior
The Create and Quick Create submenus should drop down, as they do on earlier installations

<img width="421" alt="Screen Shot 2021-03-25 at 7 56 23" src="https://user-images.githubusercontent.com/1073780/112425987-c5727200-8d3f-11eb-9b7f-b68e2d807958.png">

### Related issue(s)/PR(s)
Let us know if this is related to any issue/pull request.

### Environment
MODX 3.0.0 Alpha 3, updated from git 25 March

Ruslan-Aleev:
@digitalpenguin When you have time, please fix this bug.
After your PR - https://github.com/modxcms/revolution/pull/15424 this condition is no longer triggered:

https://github.com/modxcms/revolution/blob/5169becd127ae0f5fe3eb5794da63797372af779/manager/assets/modext/widgets/resource/modx.tree.resource.js#L665

muzzwood:
Sure, I'll look into that tomorrow morning.

muzzwood:
Fixed in #15593

PR #15593, opened 2021-03-26
Fixed context sub-menu not appearing in resource tree

muzzwood:
### What does it do?
Changed array of core resource class names to use single slashes rather than double.

### Why is it needed?
After a previous PR #15424 , the context sub-menu was no longer being triggered.

### How to test
Use the resource tree context menu (right-click) to create a resource.

### Related issue(s)/PR(s)
Issue: #15590
Previous PR: #15424


ghost:
Still not working

Edited line in the file
Cleared browser cache
Cleared core/cache

muzzwood:
Hi @sottwell, this change would only be in the uncompressed js assets. It's possible your manager is using the compressed versions. Could you check the system setting?

ghost:
That was it. Works fine now. Thank you!

Ibochkarev (approved):
It worked!

Ruslan-Aleev (approved):

PR #15594, opened 2021-03-27
Fixed dummy data for the makeUrl method

sergant210:
### What does it do?
Fixed dummy data for the makeUrl method.

### Why is it needed?
This test is incorrect and cannot be fulfilled. So PR #15593 cannot be tested.

### DIFF
--- a/_build/test/Tests/Cases/Request/MakeUrlTest.php
+++ b/_build/test/Tests/Cases/Request/MakeUrlTest.php
@@ -123,7 +123,7 @@ class MakeUrlTest extends MODxTestCase {
     public function providerSingleParameter() {
         return [
             // Dummy data to pass on first makeUrl
-            [12345, ''],
+            [12349, ''],
             [12345, 'unit-test/'],
             [12346, 'unit-test/child.html'],
         ];


opengeek:
I don't understand. This test passes everywhere. The failure on #15593 seems like it may have been a transient problem because it took too long to execute, and only on PHP 7.3.

sergant210:
@opengeek Could you explain me how the makeUrl method with the same resource id can give different results?
```
[12345, ''],
[12345, 'unit-test/'],
```
This is a deterministic function.

Mark-H:
Rerunning the jobs on #15593 made it pass on 7.3 as well, but I do think that's a very good question @sergant210 😆

opengeek:
> @opengeek Could you explain me how the makeUrl method with the same resource id can give different results?
> 
> ```
> [12345, ''],
> [12345, 'unit-test/'],
> ```
> 
> This is a deterministic function.

If you look at the results, it shows it was a "risky test," which means it took too long to run. I'm betting this was a temporary problem possibly as a result of unusually high utilization on the host machine. If it keeps happening, we can mark the test as `@medium` so it has more time to complete.

Mark-H:
I think what @sergant210 is asking is: why running makeUrl for resource 12345 is expected to return two different values? Shouldn't the first `$modx->makeUrl(12345)` call return the same uri as the second identical `$modx->makeUrl(12345)` call?

opengeek:
As for the test result itself, that IS a very good question. I have no idea, to be honest.

opengeek:
> I think what @sergant210 is asking is: why running makeUrl for resource 12345 is expected to return two different values? Shouldn't the first `$modx->makeUrl(12345)` call return the same uri as the second identical `$modx->makeUrl(12345)` call?

Yeah. How that is passing is beyond my ability to rationalize at the moment.

sergant210:
> If you look at the results, it shows it was a "risky test," which means it took too long to run. I'm betting this was a temporary problem possibly as a result of unusually high utilization on the host machine. If it keeps happening, we can mark the test as `@medium` so it has more time to complete.

Because it's not unit testing, but functional testing. Unit testing does not involve working with a database. This is what mocks are designed for.

opengeek:
> > If you look at the results, it shows it was a "risky test," which means it took too long to run. I'm betting this was a temporary problem possibly as a result of unusually high utilization on the host machine. If it keeps happening, we can mark the test as `@medium` so it has more time to complete.
> 
> Because it's not unit testing, but functional testing. Unit testing does not involve working with a database. This is what mocks are designed for.

Feel free to refactor the entire test suite to use mocks. 😬

sergant210:
> Feel free to refactor the entire test suite to use mocks. 😬

Then I will have to rewrite almost the entire MODX and xPDO. 😂

opengeek:
> > Feel free to refactor the entire test suite to use mocks. 😬
> 
> Then I will have to rewrite almost the entire MODX and xPDO. 😂

Exactly. I'm glad someone gets me and my dry humor! 😂

JoshuaLuckers:
Are we going to rewrite MODX and xPDO or can we approve and merge this PR? 🙈

opengeek:
> Are we going to rewrite MODX and xPDO or can we approve and merge this PR? 🙈

I wish I could use GIFs here, because the little girl shrugging with the "why not both?" caption would be perfect.

PR #15597, opened 2021-03-27
Fixes notification display time

### LINKED ISSUE #15573: New password cannot be copied
### Summary
New password cannot be copied after automatic generation of MODX

### Step to reproduce
Go to Management - Users. Select a user, check the New password checkbox, select automatic password generation and save.

### Observed behavior

https://user-images.githubusercontent.com/2138260/111960771-5cf67b80-8b1a-11eb-92bd-1dbaf5bb4248.mp4

### Expected behavior
After generation, the notification should not close immediately

### Related issue(s)/PR(s)
NA

### Environment

Ibochkarev:
### What does it do?
Fixes notification display time from 1.5 seconds to 10

### Why is it needed?
Fixes an issue with the inability to copy the generated password, but also affects other notifications in other parts of the control panel

### Related issue(s)/PR(s)
Resolves #15573

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -830,9 +830,9 @@ Ext.extend(MODx.Msg,Ext.Component,{
         var fadeOpts = {remove:true,useDisplay:true};
         if (!opt.dontHide) {
             if(!Ext.isIE8) {
-                m.pause(opt.delay || 1.5).ghost("t",fadeOpts);
+                m.pause(opt.delay || 10).ghost("t",fadeOpts);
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
This is not really a solution to the problem, now in other forms there is a delay of 10 seconds.
I think for the "User" section, we need to return the window, as in the 2.x branch:

pass_w

alroniks:
I agree with Ruslan

Ibochkarev:
At least I tried, but the correct solution would be to return the window, as Ruslan suggests.

PR #15598, opened 2021-03-28
Avoid autoloading core processors with a custom processors path

Jako:
### What does it do?
Check for a custom processors path that is not the core processor path

### Why is it needed?
Avoid running autoloaded core processors with a custom processors_path option. This could lead to accidentally execute core processors with a custom connector.

### How to test

Change line 79 of `connectors/index.php` and add a processors_path option with a not existing path.
```
$modx->request->handleRequest(['processors_path' => 'test']);
```

and i.e. refresh the system settings grid. Without the patch, the core processor class is executed.

### Related issue(s)/PR(s)
None known.

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1751,18 +1751,21 @@ class modX extends xPDO {
             return $processor->run();
         }
 
-        // Check if we're dealing with an action like "context/update", which we transform into the 3.0+ class name
-        $legacyAction = 'MODX\\Revolution\\Processors\\' . implode('\\', array_map('ucfirst', explode('/', $action)));
+        // Check for a custom processors path that is not the core processor path to avoid running autoloaded core processors
+        if (!(isset($options['processors_path']) && !empty($options['processors_path']) && $options['processors_path'] != $this->config['processors_path'])) {
+            // Check if we're dealing with an action like "context/update", which we transform into the 3.0+ class name
+            $legacyAction = 'MODX\\Revolution\\Processors\\' . implode('\\', array_map('ucfirst', explode('/', $action)));
 
-        // Special case for TVs, for which the directory is named differently from how it was called in the past
-        if (strpos($legacyAction, '\\Tv\\') !== false) {
-            $legacyAction = str_replace('\\Tv\\', '\\TemplateVar\\', $legacyAction);
-        }
-        if (static::isProcessorClass($legacyAction)) {
-            /** @var Processor $processor */
-            $processor = $legacyAction::getInstance($this, $legacyAction, $scriptProperties);
+            // Special case for TVs, for which the directory is named differently from how it was called in the past
+            if (strpos($legacyAction, '\\Tv\\') !== false) {
+                $legacyAction = str_replace('\\Tv\\', '\\TemplateVar\\', $legacyAction);
+            }
+            if (static::isProcessorClass($legacyAction)) {
+                /** @var Processor $processor */
+                $processor = $legacyAction::getInstance($this, $legacyAction, $scriptProperties);
 
-            return $processor->run();
+                return $processor->run();
+            }
         }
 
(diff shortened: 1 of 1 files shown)


core/src/Revolution/modX.php
```diff
@@ -1751,18 +1751,21 @@ public function runProcessor($action = '', $scriptProperties = [], $options = []
             return $processor->run();
         }
 
-        // Check if we're dealing with an action like "context/update", which we transform into the 3.0+ class name
-        $legacyAction = 'MODX\\Revolution\\Processors\\' . implode('\\', array_map('ucfirst', explode('/', $action)));
+        // Check for a custom processors path that is not the core processor path to avoid running autoloaded core processors
+        if (!(isset($options['processors_path']) && !empty($options['processors_path']) && $options['processors_path'] != $this->config['processors_path'])) {
```
sergant210: I've always wondered - why check `isset` if `!empty` does the same thing?

Mark-H: This nested double negative condition had me squinting for a minute - any chance that can be simplified for future readers?

Mark-H (commented):
While the initial problem reported was about core processors being executable from third-party connectors when provided in the format `resource/update`, earlier on line 1747 checks if the provided action is a fully qualified name (like `\MODX\Revolution\Processors\Resource\Update`) so isn't this additional check just going to get bypassed by providing the full class?

Do we perhaps need to rethink if extras need their own connector at all for 3.0+ extras?

core/src/Revolution/modX.php
opengeek: They are not precisely the same thing. `isset()` checks if the variable (in this case, an array element) exists and is NOT null. The additional check for `!empty()` makes sure it is not an empty string or other value that would basically be invalid but would still pass the `isset()` test.

sergant210: @opengeek Just open the [documentation](https://www.php.net/manual/en/function.empty.php).
> No warning is generated if the variable does not exist. That means empty() is essentially the concise equivalent to !isset($var) || $var == false. 

Accordingly, `isset()` is redundant.

Jako: A `or` condition (which is mostly the same as a `not and` condition) ist not possible without a PHP notice, since PHP stops the check in a `and` chain when the first check is false. The first check avoids the second check, throwing a notice. array_key_exist would be an option, but that’s slower.

Jako: Ok, thanks. If I am right, this was different before (PHP 4 or 5). I hope this will not be changed back.

opengeek: Agreed that `isset()` is indeed redundant here. But they are not the same thing. As @Jako is alluding to, I believe `empty()` used to throw an undefined index notice at some point, probably back around 5.3, so this is just a leftover cargo cult artifact from the early days of Revo.

sergant210: > But they are not the same thing.

I know it. I'm talking about this specific check.

Jako:
> Do we perhaps need to rethink if extras need their own connector at all for 3.0+ extras?

A custom connector is only needed, when the extra processor classes autoloading is not provided by the core. But we can think about custom connectors, which can be restricted to custom namespace prefixes. The core connector can run all processors and the extra connector only the ones with a mandatory prefix property.

core/src/Revolution/modX.php
Jako: It is solved now.

JoshuaLuckers (approved):

Mark-H (changes requested):
I think this needs more discussion before changing.

theboxer:
>  But we can think about custom connectors, which can be restricted to custom namespace prefixes.

In this case, you can add the necessary checks to the custom connector itself? To allow only processors you want to allow. I don't think that this needs to be baked in core.

I'd vote for closing this one.

PR #15605, opened 2021-03-29
Update default value for the systems settings

### LINKED ISSUE #14273: [enhancement] System settings: change default value

Ibochkarev:
### What does it do?
Update default value for the systems settings

### Why is it needed?
Changes the default values for the following system settings to more appropriate ones:
- `hidemenu_default`
- `publish_default`

### Related issue(s)/PR(s)
Resolves #14273

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -706,7 +706,7 @@ $settings['global_duplicate_uri_check']->fromArray([
 $settings['hidemenu_default']= $xpdo->newObject(modSystemSetting::class);
 $settings['hidemenu_default']->fromArray([
   'key' => 'hidemenu_default',
-  'value' => false,
+  'value' => true,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'site',
@@ -1337,7 +1337,7 @@ $settings['phpthumb_zoomcrop']->fromArray([
 $settings['publish_default']= $xpdo->newObject(modSystemSetting::class);
 $settings['publish_default']->fromArray([
   'key' => 'publish_default',
-  'value' => false,
+  'value' => true,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'site',


opengeek:
I would prefer to deal with each of these changes separately, as some of them seem fine and others I am in total disagreement with. In particular, the `friendly_alias_restrict_chars` needs to remain as `pattern` (if the default pattern needs adjusting, then we should do that instead) and the `request_method_strict` should not be `true` by default in my estimation.

opengeek:
Also, this should not have upgrade scripts. We never want to change values for these settings on existing deployments.

Ibochkarev:
@opengeek I made all the changes

sergant210:
> and the `request_method_strict` should not be `true` by default

@opengeek  Why? As a result, we get 2 URLs of one page. What for?

opengeek:
> > and the `request_method_strict` should not be `true` by default
> 
> @opengeek Why? As a result, we get 2 URLs of one page. What for?

I have no idea what you mean.

opengeek:
> > > and the `request_method_strict` should not be `true` by default
> > 
> > 
> > @opengeek Why? As a result, we get 2 URLs of one page. What for?
> 
> I have no idea what you mean.

Oh, my apologies. I was thinking of the `friendly_urls_strict` setting. Nevermind. The `request_method_strict` is fine changing to default `true`.

opengeek:
> @opengeek I made all the changes

I was just providing feedback. We can always discuss these things before you just go changing everything. 😄

sergant210:
> The `request_method_strict` is fine changing to default `true`.

So the test that you call the unit will be broken ;))

Ruslan-Aleev:
> In particular, the friendly_alias_restrict_chars needs to remain as pattern (if the default pattern needs adjusting, then we should do that instead)

@opengeek Why such difficulties? Why such variability at all for this setting, can you give a specific example where it is needed?
In my opinion, 99.9% (and maybe 100%) of all sef-url consist of numbers and letters, or I'm wrong?

_build/data/transport.core.system_settings.php
```diff
@@ -706,7 +706,7 @@
 $settings['hidemenu_default']= $xpdo->newObject(modSystemSetting::class);
 $settings['hidemenu_default']->fromArray([
   'key' => 'hidemenu_default',
-  'value' => false,
+  'value' => true,
```
JoshuaLuckers, line 709: Why do we want to hide resources from the menu's by default?

ant-tar:
Please add label "needs-docs" - I'll update doc settings, default values

_build/data/transport.core.system_settings.php
opengeek: We absolutely do not want that.

opengeek:
This PR has not been accepted as is and is thus not ready for documentation changes, yet.

Ruslan-Aleev:
@opengeek We need to discuss this PR in more detail. For example, there are comments without attention:

> So the test that you call the unit will be broken ;))

See https://github.com/modxcms/revolution/pull/15605#issuecomment-809609275 What can we do about it?

> In particular, the friendly_alias_restrict_chars needs to remain as pattern (if the default pattern needs adjusting, then we should do that instead)

See https://github.com/modxcms/revolution/pull/15605#issuecomment-809676827 In my opinion, the `alphanumeric` option is better.

p.s. Maybe you should move the discussion to the Discussions section?

p.s.s By the way, our approach to the discussion is not entirely correct. Those. at the stage of creating an issue, questions are ignored, and only in PR does the discussion begin. As a result, the PR author remakes the work several times. It is necessary to somehow think over the interaction at the issue stage.

Mark-H:
I'd personally prefer the lead architect isn't dragged into "bikeshedding" over tiny details like this ;) These 2 settings have an impact on the user experience, touching on what people may expect to happen when creating resources on a clean site. Best addressed from a UX perspective.

The only argumentation in favor of these changes I've seen is in the issue:

> In my opinion, resources are published more often than not published. Better to set "yes".

That's very anecdotal. It is my experience that resources tend to start off as unpublished before being published when they're ready to go live. So now we have two conflicting anecdotes - how do you propose we resolve that? By changing the default every other version?

If the underlying argument would be that it's hard for new users to understand a freshly created resource isn't available until it's published (which could be a valid concern, and also have different ways to fix besides this default setting), I'm very confused the same PR proposes hiding resources by default as that then goes against that same idea.  

My personal opinion is that these default values are fine as is and don't need changing.

Ruslan-Aleev:
I agree with `hidemenu_default` and `publish_default`. Although other settings have a more weighty rationale (but maybe not everything is simple with them either) :)
When I created the issue, I wrote in the description of the issue that we need a discussion.

PR #15608, opened 2021-03-29
Ignore l10n_** branches on ci workflow actions

opengeek:
### What does it do?
Prevents CI workflow on l10n_** branches pushed by CrowdIn

### Why is it needed?
When CrowdIn pushes a lot of individual commits for translations, a backlog of workflow is created.

### How to test
n/a

### Related issue(s)/PR(s)
none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,11 @@ name: CI
 
 on:
   push:
+      branches-ignore:
+          - 'l10n_**'
   pull_request:
+      branches-ignore:
+          - 'l10n_**'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


alroniks (approved):

Ibochkarev (approved):

.github/workflows/ci.yml
```diff
@@ -2,7 +2,11 @@ name: CI
 
 on:
   push:
+      branches-ignore:
+          - 'l10n_**'
   pull_request:
+      branches-ignore:
+          - 'l10n_**'
```
Mark-H, line 9: Would it make sense to still keep it active on pull_request? I guess there's not a ton of applicable tests being run on CI related to translations, but that should at least be one instead of _\*checks notes\*_ 685 times?

opengeek: I'm not sure. I wasn't clear on how that works. Is the PR workflow checking the branch-ignores for the target branch or the branch that is pushed?

Ibochkarev: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

opengeek: I read that, @Ibochkarev, but I still do not understand what I mentioned above.

opengeek: If I'm right, the statement…

> For a pull_request event, only branches and tags on the base are evaluated.

…means that this would only ignore pull requests if they targeted a base branch that matches the pattern. Let's try it and we can fix it if that is not the case.

PR #15610, opened 2021-03-30
a11y of setup: using the tab key to select the language

mishanthrop:
### What does it do?
During the setup, the user can select the system language using the Tab key

### Why is it needed?
Because accesibility

### How to test
1. goto http://localhost/setup/?action=language
2. use keyboard to select language

### Related issue(s)/PR(s)
-

modx_install_language_a11y


mishanthrop:
@cla-bot check

Ibochkarev (approved):
LGTM 👍 

https://user-images.githubusercontent.com/2138260/112941665-cd238380-9150-11eb-8047-325f11464ed0.mov

mishanthrop:
@Ibochkarev thanks for how-it-works-video. Borders of focused elements are differs between browsers and OS (sometimes borders are solid black). Anyway, it should be convenient for user.

Mark-H:
FWIW this was closed because of new PR #15823 I believe

PR #15612, opened 2021-03-30
Add return link to package management on a page of single package

Ruslan-Aleev:
### What does it do?
Added breadcrumbs for the "Package" panel:

package_breadcrumbs

### Why is it needed?
UI/UX improvement

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/14890

### DIFF
--- a/manager/assets/modext/workspace/package/package.panel.js
+++ b/manager/assets/modext/workspace/package/package.panel.js
@@ -15,7 +15,7 @@ MODx.panel.Package = function(config) {
         ,cls: 'container'
         ,chunk: ''
         ,bodyStyle: ''
-        ,items: [{
+        ,items: [this.getPageHeader(config),{
             html: _('package')
             ,id: 'modx-package-header'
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):
LGTM 👍 

https://user-images.githubusercontent.com/2138260/113014059-19e17b80-919e-11eb-81e0-c3e0084112fe.mov

Ruslan-Aleev:
@opengeek Here in the PR header there is not "Package Manager", but the "Package" panel :) Probably, it would be more correct to specify this way -  "Package Manager -> Package".

opengeek:
> @opengeek Here in the PR header there is not "Package Manager", but the "Package" panel :) Probably, it would be more correct to specify this way - "Package Manager -> Package".

Is it not all of the Package Management area that you are adding breadcrumbs to? Individual packages are still part of package management.

alroniks (approved):

Issue #10479, opened 2010-08-23
Variable naming in the config.inc.php file


shamblett:
shamblett created Redmine issue ID 10479

Some variables in the config.inc.php file start with '$modx_' but most don't. This is not a problem unless you are using this file to create a MODx object in API mode only, possibly embedding this in a 3rd party application say. When you include the config.inc.php file these generic names such as '$http_host' pollute the environment downstream of the include and can cause difficult to find errors. Prefixing all these variables with '$modx_' would help to alleviate this and provide consistency.

vaibhavbansal23:
config.inc.php is located in which folder ?

OptimusCrime:
core/config I belive

Ruslan-Aleev:
It also makes sense to place related fields side by side, for example, $dbase and $database_user, since the fields often have the same values and it will be easier to edit them visually.
Logically connected with https://github.com/modxcms/revolution/issues/14238

Ruslan-Aleev:
By the way, there are lines in the config:
```
$database_type = 'mysql';
$database_server = 'localhost';
$database_user = 'db_user';
$database_connection_charset = 'utf8';
$dbase = 'db_name';
...
$database_dsn = 'mysql:host=localhost;dbname=db_name;charset=utf8';
```
But why variables ($database_type, $database_server, $dbase and other) are not used in **$database_dsn**, but data needs to be specified again.

sergant210:
`$database_dsn = "mysql:host=$database_server;dbname=$dbase;charset=$database_connection_charset";`

PR #15622, opened 2021-04-01
Changed 'database_dsn' connection setting in config.inc.php

Ruslan-Aleev:
### What does it do?
Changed `database_dsn` connection setting in `config.inc.php`:
1) Changed the order of db connection settings.
For example, `$dbase` and `$database_user`, since the fields often have the same values and it will be easier to edit them visually.

2) ~Changed `database_dsn` setting.
The `database_dsn` setting contained data, although there are variables above in `core/config.inc.php`, it is more logical to use them.
When you move a site to another server, there will be less editing and fewer possible bugs.~

3) Changed the order of variables for paths.

### Why is it needed?
Fewer edits and fewer possible bugs in config.inc.php.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/10479#issuecomment-568462571

### DIFF
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -4,11 +4,11 @@
  */
 $database_type = '{database_type}';
 $database_server = '{database_server}';
+$database_connection_charset = '{database_connection_charset}';
+$table_prefix = '{table_prefix}';
+$dbase = '{dbase}';
 $database_user = '{database_user}';
 $database_password = '{database_password}';
-$database_connection_charset = '{database_connection_charset}';
-$dbase = '{dbase}';
-$table_prefix = '{table_prefix}';
 $database_dsn = '{database_dsn}';
 $config_options = {config_options};
 $driver_options = {driver_options};
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -20,25 +20,11 @@ $site_sessionname = '{site_sessionname}';
 $https_port = '{https_port}';
 $uuid = '{uuid}';
 
-if (!defined('MODX_CORE_PATH')) {
-    $modx_core_path= '{core_path}';
-    define('MODX_CORE_PATH', $modx_core_path);
-}
-if (!defined('MODX_PROCESSORS_PATH')) {
-    $modx_processors_path= '{processors_path}';
-    define('MODX_PROCESSORS_PATH', $modx_processors_path);
-}
-if (!defined('MODX_CONNECTORS_PATH')) {
-    $modx_connectors_path= '{connectors_path}';
-    $modx_connectors_url= '{connectors_url}';
-    define('MODX_CONNECTORS_PATH', $modx_connectors_path);
-    define('MODX_CONNECTORS_URL', $modx_connectors_url);
-}
-if (!defined('MODX_MANAGER_PATH')) {
-    $modx_manager_path= '{mgr_path}';
-    $modx_manager_url= '{mgr_url}';
-    define('MODX_MANAGER_PATH', $modx_manager_path);
-    define('MODX_MANAGER_URL', $modx_manager_url);
+if (!defined('MODX_ASSETS_PATH')) {
+    $modx_assets_path= '{assets_path}';
+    $modx_assets_url= '{assets_url}';
+    define('MODX_ASSETS_PATH', $modx_assets_path);
+    define('MODX_ASSETS_URL', $modx_assets_url);
 }
 if (!defined('MODX_BASE_PATH')) {
     $modx_base_path= '{web_path}';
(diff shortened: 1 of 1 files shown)


core/docs/config.inc.tpl
```diff
@@ -4,12 +4,12 @@
  */
 $database_type = '{database_type}';
 $database_server = '{database_server}';
-$database_user = '{database_user}';
-$database_password = '{database_password}';
 $database_connection_charset = '{database_connection_charset}';
-$dbase = '{dbase}';
 $table_prefix = '{table_prefix}';
-$database_dsn = '{database_dsn}';
+$dbase = '{dbase}';
```
alroniks, line 9:
```suggestion
$database_name = '{database_name}';
```

For consistency. Placeholder probably need to left the same, but still.

JoshuaLuckers: @alroniks will this work when you upgrade MODX?

alroniks: I am not sure, I review it in detail a bit later.

Ruslan-Aleev:
@alroniks Any thoughts?

core/docs/config.inc.tpl
opengeek: FWIW, I would rather not change the variable names. Though it is not good practice, I've seen places where people use the global variables in the config file.

setup/includes/modinstallsettings.class.php
```diff
@@ -42,14 +42,17 @@ protected function rebuildDSN() {
                     $server = explode(':', $this->settings['database_server']);
                     $port = !empty($server[1]) ? "port={$server[1]};" : '';
                     $database_dsn = "{$this->settings['database_type']}:host={$server[0]};{$port}dbname={$this->settings['dbase']};charset={$this->settings['database_connection_charset']}";
+                    $database_dsn_config = '{$database_type}:host={$database_server};dbname={$dbase};charset={$database_connection_charset}';
```
opengeek: Why is this line being added? Wouldn't it be better to just use the variables in the config file to set the existing variable here?

core/docs/config.inc.tpl
Mark-H: Gitify is one of those cases; it grabs db details to make a backup with mysqlbackup.

setup/includes/modinstallsettings.class.php
Ruslan-Aleev: You mean what is better to write in `database_dns` without creating `database_dns_config`, right?

opengeek: Yes, unless there is some purpose for still having `database_dsn` separate that I am missing?

Ruslan-Aleev: In my opinion, there is no need to add a line, later I will check for sure and redo the PR.

It also confuses me that there are separate variables for `host={$server[0]};{$port}` in `database_dsn / server_dsn`. This is due to some requirements, can't we just use `host={$database_server}`?

opengeek: It's like that for a very specific reason. I will take a stab at cleaning this PR up myself and moving forward with it.

JoshuaLuckers: @opengeek do you think you'll be able to do this before the public release of 3.0?

opengeek: I completely forgot about this. Probably not, but we can get it in a patch release.

opengeek:
After re-reviewing this, I don't see the point in changing the order of the variables. The DSN part of this is simply incorrect and should not be changed anyway.

Ruslan-Aleev:
> After re-reviewing this, I don't see the point in changing the order of the variables. The DSN part of this is simply incorrect and should not be changed anyway.

@opengeek I would still discuss this point.
The order of variables should be changed due to the fact that those that have the same values are located in different places. And placing the variables side by side will make it convenient to edit them, for example, when moving the site to another server. Why make things difficult for users?
For example, in WordPress, in the latest versions, this was also noticed and fixed (in the installer, at least).

And about DSN I did not understand what you mean?

opengeek:
I've reworked it without the invalid changes to the DSN. The DSN host and port are separate variables whereas the database_server variable is in the format host:port. These are incompatible and must be handled differently.

Ruslan-Aleev:
Thank you very much, I will test it tomorrow.
Should I remove `database_dsn_config`? Or will you delete it?

core/docs/config.inc.tpl
```diff
@@ -1,87 +1,87 @@
...
+        }
+        $http_host .= ($_SERVER['SERVER_PORT'] == 80 || $isSecureRequest) ? '' : ':' . $_SERVER['SERVER_PORT'];
+        define('MODX_HTTP_HOST', $http_host);
+    }
+}
+if (!defined('MODX_SITE_URL')) {
+    $site_url= $url_scheme . $http_host . MODX_BASE_URL;
+    define('MODX_SITE_URL', $site_url);
+}
+if (!defined('MODX_ASSETS_PATH')) {
```
Ruslan-Aleev: @opengeek I apologize for not paying attention to this.
Maybe the `MODX_ASSETS_PATH` variable should be placed higher, before other path variables, since we are correcting the order? Because other variables with `_PATH` are located above.
Could you sort or group them somehow more competently? You know better the priority of variables and their logic.
p.s. I understand that the PR is not really about this, but it would add convenience and logic...

Ruslan-Aleev:
@opengeek I changed the order of the variables for the paths too.
I also looked for a solution for `database_dsn`, but, apparently, because of the ports, we will have to print the value, and not pass variables.

opengeek:
I do not agree with changing the order of the paths. If you want to move ASSETS_PATH up below the last one, that's fine. But, honestly, this is a distraction and not really useful. We do not need to change the DSN.

Ruslan-Aleev:
I understood about DSN, yes.
About the order, initially why I paid attention to this (both for connecting to the database and for the path), this is because when moving the site to another server, this constantly needs to be edited.
Therefore, I moved the paths above, and arranged them alphabetically (easier to find).
And, for example, MODX_LOG_LEVEL_FATAL hardly needs to be changed, so it's more logical to leave them below.

Ruslan-Aleev:
If it seems to you that we need to sort it differently - write the desired order, I will correct it.
Although, to be honest, I never remembered their order initially, and ordering them alphabetically seemed right to me.

PR #15627, opened 2021-04-02
Improve language selector accessibility on the login page

### LINKED ISSUE #12793: Improve Accessibility of Language Select Boxes
### Summary

As you can notice towards the end of [this screencast](http://j4p.us/0v1I1t17203e) the languages in the language select box aren't read accessibly.
### Step to reproduce

Navigate to the language select box using a screen reader.
### Observed behavior

Languages aren't decipherable. 
### Expected behavior

Languages should be decipherable. 
### Environment

MODX 2.4 Mac OS X Safari with VoiceOver

Ibochkarev:
### What does it do?
Improve Accessibility of Language Select Boxes

### Why is it needed?

Adds date attributes to improve accessibility such as `data-iso-country-code` and `aria-label`

Before:

``` html
<option value="en" lang="en">English</option>
```

After

``` html
<option aria-label="English" value="en" lang="en">English</option>
```

### How to test

### Related issue(s)/PR(s)
Resolves #12793

### DIFF
--- a/manager/templates/default/security/login.tpl
+++ b/manager/templates/default/security/login.tpl
@@ -180,7 +180,7 @@
                 <div class="c-languageselect">
                     <select name="manager_language" id="modx-login-language-select" class="c-languageselect__select" aria-label="{$_config.cultureKey}">
                         {foreach $languages as $language => $native}
-                            <option lang="{$language}" value="{$language}"{if $language == $_config.cultureKey} selected{/if}>{$native|capitalize}</option>
+                            <option lang="{$language}" aria-label="{$native|capitalize}" value="{$language}"{if $language == $_config.cultureKey} selected{/if}>{$native|capitalize}</option>
                         {/foreach}
                     </select>
                     <span class="c-languageselect__arrow"></span>


manager/templates/default/security/login.tpl
```diff
@@ -180,7 +180,7 @@
                 <div class="c-languageselect">
                     <select name="manager_language" id="modx-login-language-select" class="c-languageselect__select" aria-label="{$_config.cultureKey}">
                         {foreach $languages as $language => $native}
-                            <option lang="{$language}" value="{$language}"{if $language == $_config.cultureKey} selected{/if}>{$native|capitalize}</option>
+                            <option lang="{$language}" data-iso-country-code="{$language}" aria-label="{$native|capitalize} selected" value="{$language}"{if $language == $_config.cultureKey} selected{/if}>{$native|capitalize}</option>
```
JoshuaLuckers: Is the aria-label really necessary? And if yes, should it include the "selected" string?

Ibochkarev: Checked in MacOS - the `lang` tag is enough for us, an additional attribute has been added `aria-label`

sergant210 (approved):

alroniks (approved):

PR #15628, opened 2021-04-02
Fix access policy template group name

alroniks:
### What does it do?
It renames access policy template group to proper Administrator instead of short and ugly Admin.

### Why is it needed?
When creating new ACL rules, like adding access to contexts, in popup access policy templates filtered by groups, so that group name goes as a parameter and it tries to filter templates by this group. In case the wrong name, not the full list will be shown.

#### How it looks before (not the length of the list)
User Group_ test _ MODX Revolution (1)

#### How it looks after
User Group_ test _ MODX Revolution

### How to test
Open ACL manager page, edit any user group and on the tab `Access Permissions` try to add a new rule for Context (vertical tab). In the popup in the field  `Access Policy` try to select any item and check how many there are items.

### Related issue(s)/PR(s)
#14009

### DIFF
--- a/manager/assets/modext/widgets/security/modx.grid.user.group.context.js
+++ b/manager/assets/modext/widgets/security/modx.grid.user.group.context.js
@@ -83,7 +83,7 @@ MODx.grid.UserGroupContext = function(config) {
             ,allowBlank: true
             ,baseParams: {
                 action: 'Security/Access/Policy/GetList'
-                ,group: 'Admin'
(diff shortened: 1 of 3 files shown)


alroniks:
Seems it require some rebase because it was started over the previous branch with ACL fixes, but check the latest commit in that case.

sergant210 (approved):

Ruslan-Aleev (approved):
p_l

PR #15629, opened 2021-04-02
Returned the method to display the password in "User" section

Ruslan-Aleev:
### What does it do?
Returned the method to display the password in "User" section to make the password display window work.

pass

In PR - https://github.com/modxcms/revolution/pull/14420 the area for selecting the password display method was hidden.
This was done in the file `manager/assets/modext/widgets/security/modx.panel.user.js`, through the `hidden: true`.

```
{
    xtype: 'radiogroup'
    ,hidden: true
    ,fieldLabel: _('password_method')
    ,columns: 1
    ,items: [...]
}
```

As I understand it, password was not sent to email or it was a security requirement.

But just hiding the area is not a correct option. First, the extjs methods don't work as expected, and second, the password display method is used in other MODX code as well.
Those, in the future, it’s better to rewrite the entire code, and not just hide the block.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15573


Ibochkarev (approved):
LGTM 👍 

https://user-images.githubusercontent.com/2138260/113500279-7d70fd80-953e-11eb-8882-cd2818578dd4.mp4

alroniks:
Now the password showed twice, in the popup and in the notification. And if try to copy from notification it still is failed. It may confuse people, would be nice to have one way.

Ruslan-Aleev:
> Now the password showed twice, in the popup and in the notification. And if try to copy from notification it still is failed. It may confuse people, would be nice to have one way.

Well it always worked like that, you can see it in 2.x. I don't see a big problem in this.

JoshuaLuckers (approved):

Issue #15586, opened 2021-03-24
MODX3 Can't log in unless I'm a sudo user


SnowCreative:
### Summary
Can't log in unless I'm a sudo user.

### Step to reproduce
Install MODX 3 alpha 3 nightly March 23. Assign your user Administrator permissions, super user. Uncheck "sudo" and save.

### Observed behavior
You'll get logged out, and can't log in again. I've tried assigning other user groups and roles, and they all fail unless sudo is checked off.

### Expected behavior
Should log in! I hadn't tested this with previous versions of alpha 3, so I don't know if this is new.

### Environment
MODX 3 alpha 3  20210323

sergant210:
1. I created a test user without the sudo permission.
2. Added him to the Administrator group with the "Super user" role.
3. Logged in to the manager successfully.

Did I do everything right?

SnowCreative:
> 1. I created a test user without the sudo permission.
> 2. Added him to the Administrator group with the "Super user" role.
> 3. Logged in to the manager successfully.
> 
> Did I do everything right?

Did you use the latest nightly from March 23?

sergant210:
I just tested on the latest dev 3.x branch. It works as expected.

Ibochkarev:
Checked on a build installed today using a composer (3.x-dev 0acc0d92fcea3fc97ab5692a8d3b55648fd6ef2c)

https://user-images.githubusercontent.com/2138260/112715700-1767f300-8f0c-11eb-861c-18cf3916c2d8.mov

sergant210:
Did you install from the 3.x branch or as usual?

Ibochkarev:
> Did you install from the 3.x branch or as usual?

`composer create-project modx/revolution www 3.x-dev`

sergant210:
Confirmed.

sergant210:
It seems to me that no one has tested permissions in MODX3. They don't work at all.

Ibochkarev:
> It seems to me that no one has tested permissions in MODX3. They don't work at all.

This is an alarming signal.

ghost:
Updated from github 31 March
Created user in Administrator group, not sudo user.

<img width="250" alt="Screen Shot 2021-03-31 at 8 14 37" src="https://user-images.githubusercontent.com/1073780/113093857-5ee6cb80-91f9-11eb-8684-1ca53c065670.png">

ghost:
[...]

PR #15632, opened 2021-04-03
Fix permissions checking in modAccessibleObject objects

sergant210:
### What does it do?
Fixed permissions cheching in objects of the modAccessibleObject class.

### Why is it needed?
Permissions for Contexts, Namespaces, Resource Groups and Categories don't work at all.

### How to test
For "mgr" context 
1. Create a test user without the sudo permission.
2. Add him to the Administrator group with the "Super user" role.
3. Try to log in in the manager.

### Related issue(s)/PR(s)
#15586.

### DIFF
--- a/core/src/Revolution/modCategory.php
+++ b/core/src/Revolution/modCategory.php
@@ -202,7 +202,7 @@ class modCategory extends modAccessibleSimpleObject
                 $acls = $this->xpdo->getIterator(modAccessCategory::class, $c);
 
                 foreach ($acls as $acl) {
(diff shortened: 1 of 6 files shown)


Ibochkarev (approved):
It worked! 👍 

https://user-images.githubusercontent.com/2138260/113501374-6504e100-9546-11eb-863a-940f482e3a2c.mp4

Ruslan-Aleev (approved):
It works, thanks

alroniks (approved):

PR #15633, opened 2021-04-03
Automatically mark issues by appropriate labels

alroniks:
### What does it do?
Adds a predefined label to issue templates, base on the type.

### Why is it needed?
All newly created issues will be marked by appropriate labels, like `bug` or `feature` and maybe later `rfc`, which will allow to integrators deal with issues because they already will be marked. Also, some automation and collecting statistics processors can be applied.

Documentation is here https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

### How to test
N/A

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Create a report to help us improve MODX Revolution
-
+labels: bug
 ---
 
 ## Bug report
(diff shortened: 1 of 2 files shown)


Ibochkarev (approved):

alroniks:
Seems to need to create backport to 2.x, as it is still the default branch and probably GitHub will take config from it.

Mark-H (approved):

PR #15637, opened 2021-04-05
Fix form customization on static resources

### LINKED ISSUE #13042: Static Resources and Manager/Form Customization and the "Content" field
### Summary

It doesn't seem to be possible to hide the "content" field of a static resource. 
It's also not possible to change the label/title of this field.

Use case: I don't want my users to mess around with the value stored inside "content" because I'm setting the destination of a static resource with a file TV using a dedicated media source (assets/id/{id}/files/).
### Step to reproduce

Create a form customization rule for a template that is used with static resources. Try to hide or rename the content field.
### Observed behavior

The content field remains unaffected by form customization.
### Expected behavior

You should be able to hide this field or to rename it.
### Environment

Latest (2.5) MODX Revo on Skytoaster.

Cheers,

pepebe

alroniks:
### What does it do?
It renames back content field to proper `modx-resource-content` which also used in form customizations rules.

### Why is it needed?
It addresses the issue when impossible to hide the content field for static resources via form customization.

### How to test
- Create FC profile and set.
- Uncheck `modx-resource-content` and save set
- Create a static resource and check the content field. It should be hidden.

### Related issue(s)/PR(s)
It was introduced with this fix #11689 wherein comments said that is only one reason - shadow. In 3.x it looks fine without any shadows, so no reason to have a different name.

Fix #13042

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.static.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.static.js
@@ -19,7 +19,7 @@ Ext.extend(MODx.panel.Static,MODx.panel.Resource,{
     defaultClassKey: 'MODX\\Revolution\\modStaticResource'
     ,classLexiconKey: 'static_resource'
     ,rteElements: false
-    ,contentField: 'modx-resource-content-static'
+    ,contentField: 'modx-resource-content'
 
     ,getContentField: function(config) {
         return {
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):
Worked!

https://user-images.githubusercontent.com/2138260/113628680-0eeb8700-9687-11eb-868c-e69655efbde7.mp4

theboxer (approved):

theboxer (approved):

opengeek (approved):

Issue #15225, opened 2020-09-05
Add explanatory text in the window for creating a directory in the file tree


Ruslan-Aleev:
### Summary
When creating a directory inside a media source - `/` is always specified in the creation window instead of the root folder of the media source.
Those it is not clear where the directory is created, inside the root folder of the media source, or inside the root folder of the site.

![source-path](https://user-images.githubusercontent.com/12523676/92309122-759eaa00-efab-11ea-8547-3e22189c3d85.gif)

We need to add clarifying text below the field.

For example:
![source-path-2](https://user-images.githubusercontent.com/12523676/92309123-77686d80-efab-11ea-9400-7e561d46a8d4.png)

### Environment
MODX 2>

### Why is it needed?
UX improvement

alroniks:
Merged

Issue #8226, opened 2012-07-10
Hints for Media Source paths


jcamachott:
john@pixelstation.com created Redmine issue ID 8226

Just a suggestion for the help text for basePath and baseUrl. For basePath:

To the ed of "The file path to point the Source to." add "Do not include the folder your MODx is installed inside. End your path with a slash. e.g. assets/img/subfolder/ "

Same for baseUrl.

alroniks:
Merged

PR #15652, opened 2021-04-08
Add clarity within creating folders in media sources

Ruslan-Aleev:
### What does it do?
1) Added clarification for `basePath/baseUrl` parameters in media source:
Added an example, and indicated the dependence of the paths on other parameters.

2) Added clarification when creating a folder inside the root of media source
When creating a directory inside a media source - `/` is always specified in the creation window instead of the root folder of the media source. Those it is not clear where the directory is created, inside the root folder of the media source, or inside the root folder of the site.

### Why is it needed?
UX improvements

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/8226
https://github.com/modxcms/revolution/issues/15225


alroniks (approved):
It works as explained.

MODX Browser _ MODX Revolution
Media Source_ Filesystem _ MODX Revolution

Ibochkarev (approved):
изображение

PR #15655, opened 2021-04-08
Add permissions to enforce access to specific resource types [SEC-3811]

Mark-H:
### What does it do?

Adds dedicated permissions for weblinks, symlinks, and static resources to control who can:

- create resources of that type (permission already existed, wasn't enforced)
- edit resources of that type (new permission)
- delete resources of that type (new permission)

Also includes a small tweak to the modManagerController to make sure a missing permission uses the "nice" error page, rather than a blank page.  

Default policies for content editor, admin and developer are also updated to include these new permissions with the upgrade.

### Why is it needed?

SEC3801 reported static resources allowing access to things it shouldn't, which included proof from the more demo site. That demo site is heavily restricted, which led me down a very deep rabbit hole where it turns out the relevant permission (new_static_resource) was not actually enforced. That resulted in this security fix to add missing permissions and making sure they're enforced.  

### How to test

First rebuild and run setup to add the new permissions. 

- On a user with these permissions, confirm the different resource types can be interacted with as usual (create, edit, delete).
- On a user without these permissions, confirm you're presented with a friendly permission denied error. 

### Related issue(s)/PR(s)

https://community.modx.com/t/new-static-resource-permission-not-checked/3811/6


Mark-H:
Can some people please test this to make sure I didn't typo any permission checks and users can still edit things after the upgrade?

alroniks:
I will do Mark, don't worry

Mark-H:
Could I convince someone to help port this to 3.x with a high five from this incredibly cute puppy?

afbeelding

alroniks:
FYI, it is under WIP

alroniks:
Port is added, so I guess the mission completed here :) I removed related label.

PR #15656, opened 2021-04-08
Restrict static resources to predefined path [SEC-3801]

Mark-H:
Details @ https://community.modx.com/t/static-resource-vulnerability/3801/19

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1829,6 +1829,24 @@ $settings['static_elements_basepath']->fromArray(array (
   'area' => 'static_elements',
   'editedon' => null,
 ), '', true, true);
+$settings['resource_static_allow_absolute']= $xpdo->newObject('modSystemSetting');
+$settings['resource_static_allow_absolute']->fromArray(array (
+    'key' => 'resource_static_allow_absolute',
+    'value' => '0',
+    'xtype' => 'combo-boolean',
+    'namespace' => 'core',
+    'area' => 'static_resources',
+    'editedon' => null,
+), '', true, true);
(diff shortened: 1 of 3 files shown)


Mark-H:
Description of how this handles paths (from the private forum):

Note up front: the variable naming "sourcePath" is unrelated to media sources. 

It tries several things to be as compatible as possible, in order:

- L61: if absolute is allowed and a path exists at the absolute path, use that
- L75: if a file is passed that starts with the absolute sourcePath (so for example the path is `{assets_path}` and `/home/user/public_html/assets/foo.pdf` is provided), then remove the absolute path temporarily as that's added back on line 91
- L84-89 (this is the part you're looking at if I understand correctly): 
    - First, grab the root-relative relative source path. That is: if the path is set to `{assets_path}`, see if that is located within the MODX_BASE_PATH, and return only the relative portion of that: `assets/`. 
    - If we have a such a relative source path (i.e. the setting points to a directory with in the base path), and the provided filename starts with that source path (i.e. `assets/foo.pdf`), then remove the relative portion from the filename as the absolute source path will be added at line 91.

So this supports a couple of variations for maximum BC while restricting to the provided directory:

- full absolute paths (`/home/user/protected/foo/bar.pdf`), only if enabled
- full absolute paths within the allowed directory (`/home/user/public_html/assets/foo/bar.pdf`)
- relative paths from the root of the site, as that's what the media browser returns (`assets/foo/bar.pdf`)
- relative paths from the provided directory (`foo/bar.pdf`)

Also worth noting, the path setting has an absolute path, not a relative one, so that can't be matched one-on-one and needs this sort of string comparison for different ways a path may be provided.

PR #15666, opened 2021-04-12
MODX Session Cookie with SameSite

### LINKED ISSUE #15301: session_cookie_secure setting will soon be rejected
### Summary
The `session_cookie_secure` system setting is currently a boolean value, with the default being false. This means that on a secure site the session cookie will get set to not secure resulting in the following warning: 

```
Cookie "PHPSESSID" will be soon rejected because it has the "SameSite" attribute set to "None" or an invalid value, without the "secure" attribute. To know more about the "SameSite" attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
```

### Step to reproduce
Access a default install of MODX on a secure domain and check the console log. 

### Observed behavior
The session cookie is set to not secure.

### Expected behavior
If the setting was not adjusted the session cookie's security should be based on the protocol used by the site.  

### Related issue(s)/PR(s)
None so far

### Environment
Any MODX, I've tested on a few 2.6.5+ sites, but guessing this has been an issue for a while.

raffy99:
### What does it do?
May i refer to this thread? 
https://community.modx.com/t/solved-modx-session-cookie-with-samesite/3927

### Why is it needed?
MODX Session Cookie with SameSite

### How to test
add system_setting session_cookie_samesite with value 'Lax' or 'Strict' and control cookie

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1639,6 +1639,17 @@ $settings['session_cookie_httponly']->fromArray(array (
   'area' => 'session',
   'editedon' => null,
 ), '', true, true);
+$settings['session_cookie_samesite']= $xpdo->newObject('modSystemSetting');
+$settings['session_cookie_samesite']->fromArray(array (
+    'key' => 'session_cookie_samesite',
+    'name' => 'setting_session_cookie_samesite',
+    'description' => 'setting_session_cookie_samesite_desc',
+    'value' => '',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'session',
+    'editedon' => null,
+), '', true, true);
 $settings['session_gc_maxlifetime']= $xpdo->newObject('modSystemSetting');
 $settings['session_gc_maxlifetime']->fromArray(array (
   'key' => 'session_gc_maxlifetime',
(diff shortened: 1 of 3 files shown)


opengeek:
This is going to need the `session_cookie_samesite` system setting added with a default setting of an empty string. If it is empty, i.e. no value is provided, then the `samesite` entry should not be set in the array that is passed to the `session_set_cookie_params()` function.

raffy99:
Maybe a value check for 'Strict, Lax' makes more sense, an empty value means that no SameSite cookie attribute will be set. Default value for samesite in the getOption call is also worth discussing.

opengeek:
> Maybe a value check for 'Strict, Lax' makes more sense, an empty value means that no SameSite cookie attribute will be set. Default value for samesite in the getOption call is also worth discussing.

It should be empty by default, meaning no samesite attribute will be set.

opengeek:
And the default in the getOption call should be an empty string, as well.

Ibochkarev:
Related https://github.com/modxcms/revolution/issues/15301

opengeek (changes requested):
Add the session_cookie_samesite system setting with a default blank value and only set the samesite entry in the cookie params array if not empty.

Ibochkarev (changes requested):
Please add new lexicon entries

opengeek (approved):

opengeek:
Please note that I have approved the requested changes, but have not yet tested the functionality.

PR #15668, opened 2021-04-12
Fix ignoring config phpcs

Ibochkarev:
### What does it do?
Fix ignoring config phpcs

### Why is it needed?
Fix ignoring config phpcs

### Related issue(s)/PR(s)
https://github.com/tinovyatkin/action-php-codesniffer/issues/40#issuecomment-754684481


opengeek:
Awesome!

PR #15670, opened 2021-04-13
Set code of conduct for the repository

alroniks:
### What does it do?
It adds the code of conduct, which published on https://modx.com/community/contributor-covenant-code-of-conduct/ 

### Why is it needed?
[GitHub Community Best Practices](https://opensource.guide/) requires having such agreement in the repository. 

### How to test
N/A

### Related issue(s)/PR(s)
N/A


alroniks:
Hm, it is not from a fork, because I tried using internal tools of GitHub, but could we in this case exceptionally go forward?

PR #15678, opened 2021-04-16
Prevent authorized access to hashed user password, cachepwd, salt, and sessionid [SEC-3837]

Mark-H:
### What does it do?

This addresses a variety of different ways a manager user with permission to manage users might be able to access hashed passwords, the cachepwd, or a users' last sessionid. Follows from report by Solar Security who reported it was included in the security/group/user/getList response, after which I did some more searching and found a few more places that could return some of these.

### Why is it needed?

This is not super interesting from a security perspective, as passwords are hashed and any user that has permission to manage users could change their password and just take over an account that way. But as these fields are not supposed to be readable, good to lock 'm down anyway.

### How to test

Perform the actions and confirm there is no sensitive data in the raw response.

### Related issue(s)/PR(s)

https://community.modx.com/t/new-security-issues-reported-all-in-one-report/3837

### DIFF
--- a/core/model/modx/processors/security/group/user/getlist.class.php
+++ b/core/model/modx/processors/security/group/user/getlist.class.php
@@ -55,7 +55,7 @@ class modUserGroupUserGetListProcessor extends modObjectGetListProcessor {
     }
 
     public function prepareQueryAfterCount(xPDOQuery $c) {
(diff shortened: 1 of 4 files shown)


Mark-H:
Ported to 3.x in https://github.com/modxcms/revolution/pull/15843

PR #15683, opened 2021-04-22
Enable custom number of maximum search results

### LINKED ISSUE #14444: Search results limit isn't enough
### Summary
Quick summary what's this feature request about.

I need more visible results for search, some better UX solution

### Why is it needed?

search_results_length
As you can see here I have nine Home pages, but search returns only five positions, when you have tens and hundred resources this search bar isn't useful.

MODX 2.7.1 

### Suggested solution(s)
Show all matched results or add some system setting as possible limit here (I assume that list can have a lot of items). Or even show just a few and add link "show all...", but we need some additional page for these results obviously. So I'm a bit confused here..does it make sense?

### Related issue(s)/PR(s)
Don't know

SnowCreative:
### What does it do?
Controls the maximum number of search results returned for each element type when using the search function in the manager menu.

### Why is it needed?
The maximum number was hard-coded to 5. This is often not enough to be useful.

### How to test
Install this update, then edit the new system setting "search_max_results". You should now see the number of results you specified when using the search tool.

### Related issue(s)/PR(s)
https://community.modx.com/t/search-in-manager-limited/3960


rthrash:

https://community.modx.com/t/search-in-manager-limited/3960/6

JoshuaLuckers:
Wouldn't it be better to target 3.x instead? Especially when we introduce API changes (like removing a public property and adding a public new method).

ant-tar:
related to https://github.com/modxcms/revolution/issues/14444

SnowCreative:
> Wouldn't it be better to target 3.x instead? Especially when we introduce API changes (like removing a public property and adding a public new method).

Well, getMaxResults() works just as well as a private function, so I could change it to "private function getMaxResults() { . . . ", and we could keep "public $maxResults = 5" if you want, even though that wouldn't do anything in the search any more.  Is the variable $maxResults used anywhere else but in the modSearchProcessor class? It sure would make me and my clients happy to get this in 2.8.2.

SnowCreative:
OK, I've made the getMaxResults function private, and restored "public $maxResults = 5;", even though I don't know if that does anything now.

Ruslan-Aleev:
> Wouldn't it be better to target 3.x instead?

For the 3.x branch, I fixed it, only I did it differently - https://github.com/modxcms/revolution/pull/15158

Mark-H:
Given this is already addressed in 3.x I'm leaning towards leaving this as it is in 2.x.

Mark-H:
Closing as fixed in 3.0, thanks @SnowCreative.

PR #15689, opened 2021-04-28
Fix the translation of the Country South Africa

jdaehne:
### What does it do?
Corrects the translation of the country South Africa


opengeek:
@jdaehne Thank you for contributing to the project. However, translations are maintained through our integration with CrowdIn and cannot be modified directly in the repository. Please see https://docs.modx.com/current/en/contribute/translate for additional information.

PR #15702, opened 2021-05-04
Return content from non-binary static resources

### LINKED ISSUE #15696: Static resources not working correctly in 2.8.2
### Summary
When viewing a static resource (web context) in 2.8.2, only the content field renders. The rest of the template is ignored.
I noticed the issue when upgrading a site from 2.8.1 that relies on static resources.
As a quick fix, I have copied over `modstaticresource.class.php` from 2.8.1 and that works as expected.

### Step to reproduce
Create an HTML file, create a static resource and link to the file. Be sure to use the new system settings `resource_static_path` or `resource_static_allow_absolute`. Assign a template that's not empty and has the `[[*content]]` tag specified.

### Observed behavior
Only the content field renders. The rest of the template isn't there.
If the `[[*content]]` tag is removed from the template, then the template renders fine (obviously without the content).

### Expected behavior
Template should render as normal with the static resource content being added via the `[[*content]]` tag.

### Related issue(s)/PR(s)
Could be something to do with a change made here:
https://github.com/modxcms/revolution/pull/15656

### Environment
MODX version 2.8.2

muzzwood:
In my testing it seems like setting the headers for non-binary types here isn't actually required. I have however included that just in-case there's an edge case I missed.

### What does it do?
Returns the content of non-binary static resources so they can still be used with a template.

### Why is it needed?
In MODX 2.8.2 templates can no longer be rendered for static resources.

### How to test
Check the rendering of static resources with and without templates applied and check the headers.
Try changing the resource content type and test them all. e.g. html, json, pdf etc.

### Related issue(s)/PR(s)
See issue #15696

### DIFF
--- a/core/model/modx/modstaticresource.class.php
+++ b/core/model/modx/modstaticresource.class.php
@@ -145,18 +145,6 @@ class modStaticResource extends modResource implements modResourceInterface {
         if (file_exists($this->_sourceFile) && is_readable($this->_sourceFile)) {
             $content = $this->_sourceFile;
             $streamable = true;
-
-            // For non-binary types, return early to allow for template use.
-            if (!$contentType->get('binary')) {
-                // Set the appropriate content type and charset for non-binary content types
-                $mimeType = $contentType->get('mime_type') ?: 'text/html';
-                $charset = $this->xpdo->getOption('modx_charset',null,'UTF-8');
-
-                $header = 'Content-Type: ' . $mimeType . '; charset=' . $charset;
-                header($header);
-
-                return file_get_contents($content);
-            }
         }
 
         if (empty($content)) {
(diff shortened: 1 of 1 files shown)


Mark-H (changes requested):
This does not seem like the right place to me to add a return, as now all the other static functionality/headers are bypassed. 

I haven't tested this use case yet but I'd expect the final part of the function requiring changed logic. Rather than readfile or echo, readfile if streamable and not binary, and return otherwise?

muzzwood:
My thinking with this one was non-binary content wouldn't need any of the added static functionality/headers, since it would be added to a template in the regular way. 
I'm happy to change it as per your suggestion though. Aiming to get it done later this afternoon!

muzzwood:
Ok done! :)

PR #15715, opened 2021-05-18
Avoid setting headers for modStaticResource when returning content

opengeek:
### What does it do?
Prevents setting headers if the static resource is not streamed directly to the browser.

### Why is it needed?
Incorrect filesize header is set on non-binary content types which are returned and then processed. This ultimately changes the actual content length and causes content to be truncated in most instances.

### How to test
Create a static resource of HTML or another non-binary content type that contains tags that are replaced with content larger than the tag string itself and notice the content is no longer truncated.

### Related issue(s)/PR(s)
Reverts issues inadvertently created by #15702 and #15656

### DIFF
--- a/core/model/modx/modstaticresource.class.php
+++ b/core/model/modx/modstaticresource.class.php
@@ -141,10 +141,8 @@ class modStaticResource extends modResource implements modResourceInterface {
         }
 
         $content = false;
-        $streamable = false;
         if (file_exists($this->_sourceFile) && is_readable($this->_sourceFile)) {
             $content = $this->_sourceFile;
-            $streamable = true;
         }
 
         if (empty($content)) {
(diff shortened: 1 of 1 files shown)


Mark-H:
FWIW, on initial review I thought it should've only skipped the content-length header when it's a binary, however on a closer look: when it is not binary and the content is instead returned, the other headers will be set in modResponse->outputContent, so that's fine.

In
```
        if ($streamable && $contentType->get('binary')) {
```

the check for $streamable and in fact setting $streamable in the lines prior can be removed. That's left over from my original attempt to add media source support (some of which would be streamable, while others would require loading the content through the media source api) but that's not relevant in the current situation. $streamable is always true at that check because if the file can't be streamed, it already breaks out of the function because $content is empty.

To reduce the code complexity, I would recommend flipping the logic from currently:


```
if ($contentType->get('binary')) {
     // set headers and do everything else
     echo $content;
}
else {
     return file_get_contents($content);
}
exit();
```

to:


```
if (!$contentType->get('binary')) {
        return file_get_contents($content);
}

// set headers and do everything else
exit();
```
 
That should make the diff easier to understand too.

opengeek:
@Mark-H better?

Mark-H:
Brilliant, we'll get the spaghetti straightened out some day ;) Haven't pulled in the code locally yet for an actual test but this looks like it should be correct.

PR #15720, opened 2021-05-20
XSS issues reported by Solar Security

Mark-H:
### What does it do?
Fixes a couple of reflected XSS vulnerabilities in the manager. 

### Why is it needed?
Security. Full details in the report available to the security team at: https://community.modx.com/t/new-security-issues-reported-all-in-one-report/3837

### How to test
Verify the PoCs provided are no longer functional and regular functionality is unaffected. .

### Related issue(s)/PR(s)
https://community.modx.com/t/new-security-issues-reported-all-in-one-report/3837

### DIFF
--- a/manager/controllers/default/browser/index.class.php
+++ b/manager/controllers/default/browser/index.class.php
@@ -49,6 +49,12 @@ MODx.ctx = "'.$this->ctx.'";
         $placeholders = array();
 
         $scriptProperties['ctx'] = !empty($scriptProperties['ctx']) ? $scriptProperties['ctx'] : 'web';
+        $context = $this->modx->getObject('modContext', [
+            'key' => (string)$scriptProperties['ctx'],
+        ]);
+        if (!$context) {
+            return $this->failure(sprintf($this->modx->lexicon('context_with_key_not_found'), htmlentities($scriptProperties['ctx'], ENT_QUOTES, 'UTF-8')));
+        }
 
         $rtecallback = $this->modx->invokeEvent('OnRichTextBrowserInit', $scriptProperties);
         if (is_array($rtecallback)) $rtecallback = trim(implode(',',$rtecallback),',');
@@ -87,6 +93,6 @@ MODx.ctx = "'.$this->ctx.'";
      * @return array
      */
     public function getLanguageTopics() {
-        return array('file');
+        return array('file', 'context');
     }
 }
--- a/manager/controllers/default/context/update.class.php
+++ b/manager/controllers/default/context/update.class.php
@@ -81,7 +81,7 @@ class ContextUpdateManagerController extends modManagerController {
      */
     public function process(array $scriptProperties = array()) {
         if (empty($this->context)) {
-            return $this->failure(sprintf($this->modx->lexicon('context_with_key_not_found'), $this->scriptProperties['key']));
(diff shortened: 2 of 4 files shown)


Jako:
Can htmlspecialchars be used in favour of htmlentities?

Mark-H:
I guess, but why? I've always used htmlentities for things like this but if there's a code style convention to use htmlspecialchars I'm not opposed. If you want to propose such a convention and discuss that, please open an issue. 

In the meantime unless there's a reason **not** to use htmlentities I'd rather focus on ensuring this resolves the reported issued (have you had a chance to test it?) and get the release out.

Jako:
According to https://stackoverflow.com/questions/1996122/how-to-prevent-xss-with-html-php?noredirect=1&lq=1 htmlspecialchars is enough to escape the string for XSS-safe HTML.

The other parts of the patch are looking well. Only the `json_encode(['q' => $this->searchQuery])` is unsafe, when q is not escaped later on display.

Mark-H:
Both htmlentities and htmlspecialchars are currently used for this purpose in the core, so like I already said, if you want to propose a certain code style convention for it moving forward please do so in an issue. I'm not going to entertain that discussion here, and will not be changing this PR to htmlspecialchars as that will unnecessarily delay its approval and the 2.8.3 release.

Mark-H:
> The other parts of the patch are looking well. Only the json_encode(['q' => $this->searchQuery]) is unsafe, when q is not escaped later on display.

It's not displayed as-is anywhere that I can find (in `manager/assets/modext/widgets/modx.panel.search.js`), its value is only getting set to the value of input fields which do not require escaping. 

The XSS issue addressed was a line break breaking the way the value was passed to the client-side, allowing it to run code on the page from there. By using an actual json encoder the line breaks are properly escaped and it is not possible to exploit. That was also explained in the commit description and could've been inferred from the report in the private forum. 

If I've overlooked any instances where this does get rendered, please elaborate.

Jako (approved):

sergant210:
As far as I remember, XSS vulnerability  in the resource form fields (pagetitle, longtitle) was closed. In version 2.8.3, if you specify a pagetitle with the code `<script>alert ('XSS')</script>`, it will run when the page is opened.

opengeek:
> As far as I remember, XSS vulnerability in the resource form fields (pagetitle, longtitle) was closed. In version 2.8.3, if you specify a pagetitle with the code `<script>alert ('XSS')</script>`, it will run when the page is opened.

Do you mean when you open it for editing in the manager?

sergant210:
> Do you mean when you open it for editing in the manager?

On the front.

opengeek:
> > Do you mean when you open it for editing in the manager?
> 
> On the front.

It would only be XSS in the manager. We do not prevent markup from being presented on the front end. That would sabotage the entire point of being a CMS.

sergant210:
> It would only be XSS in the manager. We do not prevent markup from being presented on the front end. That would sabotage the entire point of being a CMS.

Even for pagetitle and longtitle? I get it.

Mark-H:
We can try to educate and provide good examples (basetemplate, docs) but I agree frontend is not something we can or should want to do automatically.

Mark-H:
3.x port created in https://github.com/modxcms/revolution/pull/15843

PR #15722, opened 2021-05-20
Configurable extension for static html files

Jako:
### What does it do?
Add a configurable extension for static html files

### Why is it needed?
PhpStorm does not really like the .tpl extension for html content and shows some code check issues. .tpl files are detected as Smarty template file by default. Changing this default means that real Smarty templates showing code checks issues afterwards.

### How to test
Rebuild the MODX assets, create the system setting static_elements_html_extension and set it i.e. to `.html`, enable the static_elements_automate_chunks system setting, edit a chunk and set it to static. The static file extension is set to `.chunk.html`.

### Additional work
The system setting static_elements_html_extension has to be added by a upgrade script.

### Related issue(s)/PR(s)
None known

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2259,4 +2259,13 @@ $settings['error_log_filepath']->fromArray(array (
     'area' => 'system',
     'editedon' => null,
 ), '', true, true);
+$settings['static_elements_html_extension']= $xpdo->newObject('modSystemSetting');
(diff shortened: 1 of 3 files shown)


JoshuaLuckers (approved):

Ibochkarev (approved):

alroniks:
@Jako Could you please create a port for MODX 3 just to not forget about it later? I guess it makes sense to merge two PRs together.

alroniks (approved):

PR #15731, opened 2021-05-29
Fix image TV display via hash

tanzirev:
Если выбранные изображения имеют одинаковые названия, но при этом это разные изображения, то превью таких изображений не обновляются на разных страницах, т.к. браузер кеширует прошлое изображение. 
(If the selected images have the same name, but they are different images, then the previews of such images are not updated on different pages, because the browser caches the previous image.)

### DIFF
--- a/core/model/modx/processors/element/tv/renders/mgr/input/image.class.php
+++ b/core/model/modx/processors/element/tv/renders/mgr/input/image.class.php
@@ -31,20 +31,26 @@ class modTemplateVarInputRenderImage extends modTemplateVarInputRender {
         $this->modx->controller->setPlaceholder('source',$source->get('id'));
         $params = array_merge($source->getPropertyList(),$params);
 
+        $value = $this->tv->get('value');
+        $this->tv->set('relativeValue', $value);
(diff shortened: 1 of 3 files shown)


tanzirev:
@cla-bot check

JoshuaLuckers (approved):

Ibochkarev (approved):

Jako (changes requested):
If I understand the issue correctly, the preview for images with the same path and source is the same on different MODX sites.

I suggest using a hash of the MODX site_id instead of a random string. Then the thumbnail can be cached for the same site. Is this suitable?

Ruslan-Aleev:
As I understand it, here is about the fact that the cache of the old image is displayed in the manager when replacing the image with a new one, but with the same name (or just images with the same names). Those MODX site_id will not help with this issue.

Jako:
I have changed the code to use a file (size + modification time) based hash value (if the file exist). Then the browser can cache the results.

Jako (approved):

JoshuaLuckers (approved):

Issue #14216, opened 2018-12-25
[bug] Additional resource trees


Ruslan-Aleev:
### Summary
In the "Resource Groups", "Import site from HTML" and "Import resources from static files" sections, the additional resource tree does not work correctly:
1) ~There are action icons on the right (update the tree, create a resource, create a child resource), but these icons do not work and **they are not needed there**.~
2) Tips for these icons are displayed incorrectly.

![resource_tree](https://user-images.githubusercontent.com/12523676/50423284-d4220400-086c-11e9-85b9-f689f43aee2c.png)

3) ~In the "Resource Groups" section, you can go to editing the resource by clicking on the resource. In general, **editing is not needed here**, because resources are added to groups by dragging (you can click wrongly).~

### Expected behavior
The functionality of additional trees, in my opinion, should be the same as in the "Top Menu" section, without action icons and tips.

### Environment
MODX 2.7

alroniks:
I am not sure but it seems that this features used rarely. I would remove it from the core at all and replace by a 3rd extra.

Ruslan-Aleev:
@Alroniks Yes, in my opinion, export is not particularly used, but resource groups are useful for permission setting at least.
And how is MODX usage statistics collect? Is statistics collected at all?
May be, for example, enable analytics scripts in final versions 2.x and, with the permission of the site administrator, collect a click map, a conversion map, etc. to understand the functions that are used by MODX-users and are needed in core?

JoshuaLuckers:
I agree with @Ruslan-Aleev, Resource Groups are useful and should, in my opinion, not be removed from the core.

alroniks:
In the issue described 3 points and sometimes it's difficult to answer properly for all of them. I agree that Resource Groups should be in the core, they are part of ACL system. Import can be deleted.

patrickatwsrn:
Point 1 and 2 are still an issue. They are neither correctly displayed nor do they anything at all. 

I would also remove the "edit resource on click" funtionality for elements in the resource tree to the right. It is not needed here and potentially harmful for the user experience.

PR #15746, opened 2021-06-23
Hide broken navigation in trees

Ruslan-Aleev:
### What does it do?
- Hide broken navigation in trees: Resource Groups, Import Static Resources, Import HTML.
- Disabled the ability to go inside editing a resource with an accidental click in Resource Groups section.

Before (on the gif, I click on the icons):
rg_before

After:
rg_after

Navigation is not needed there, and links allow you to leave the section by accidental click, which is inconvenient.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14216

### DIFF
--- a/_build/templates/default/sass/_tree.scss
+++ b/_build/templates/default/sass/_tree.scss
@@ -291,6 +291,10 @@
       color: $green !important;
     }
   }
+
+  #modx-gr-tree-resource &, #modx-ih-resource-tree & {
+    display: none;
+  }
 }
 
 /* Direct create buttons for a tree node */
@@ -314,6 +318,10 @@
         opacity: 1.0;
         color: $green;
       }
+    }
+
+    #modx-gr-tree-resource &, #modx-ih-resource-tree & {
+      display: none;
     }
   }
 
--- a/manager/assets/modext/widgets/security/modx.panel.resource.group.js
+++ b/manager/assets/modext/widgets/security/modx.panel.resource.group.js
@@ -8,7 +8,7 @@ MODx.panel.ResourceGroups = function(config) {
     config = config || {};
     Ext.applyIf(config,{
         id: 'modx-panel-resource-groups'
-		,cls: 'container'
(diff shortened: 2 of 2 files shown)


JoshuaLuckers:
Would it be possible to remove the icons/buttons we don't use instead of hiding them?

Ruslan-Aleev:
I have not found how to do this. PR is open for help :)

Jako:
This happens here https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/core/tree/modx.tree.js#L279

Jako:
At the moment it will be added when the node does not have a pseudoroot attribute and when it has a pnew_modDocument class. Maybe the class can be removed in the processor to avoid the node button.

Ruslan-Aleev:
> This happens here https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/core/tree/modx.tree.js#L279

`.addNodeButtons(node)` adds a icon to create a child resource, but it does not add the icons at the root of the tree (add / refresh).

JoshuaLuckers:
Maybe [this commit](https://github.com/modxcms/revolution/commit/774324d132830d7d00bc1ba8d9f7336cfba8a925) is helpful. Or this [search query](https://github.com/modxcms/revolution/search?q=pnew_modDocument&type=code).

Nonetheless, nice job starting this PR and start a good discussion on how to best solve this!

Jako:
The root icons are created in _onAppend https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/core/tree/modx.tree.js#L797

Jako:
An icon creating/removing attribute has to be added/removed i.e. in the modResourceGroupGetNodesProcessor which will be regarded in the modx.tree.js.

JoshuaLuckers (approved):
Removing the icons could be done in another PR.

theboxer (approved):

PR #15749, opened 2021-06-23
Change order of menu items for Context in tree

Ruslan-Aleev:
### What does it do?
1) Changed order of menu items for Context in tree.
The menu items for the Context in tree were not grouped, they are now.
2) Added Сontext key to menu in tree.
For other elements, the ID is displayed, for Context it is useful to display the key.

Before:
ctx_tree_before

After:
ctx_tree_after

### Why is it needed?
UI/UX

### Related issue(s)/PR(s)
N/A


rthrash:
Great improvement. I think it could be further improved by putting the most used options at the top: Create + Quick Create.

Ruslan-Aleev:
@rthrash Yes, I thought about it. Should I put Context-related menu items after "Sort by"?

Ruslan-Aleev:
I think it is worth leaving as done now, because the rest of the elements in the tree (almost all) have similar visual logic, for example, for TV:

tv_tree

JoshuaLuckers (approved):

rthrash:
I don't have a strong opinion, but I'd probably put the sort options last as I suspect that would be used the least. But I'm about a 4/10 on that front. 9/10 on the create/quick create at the top.

Ruslan-Aleev:
I understand the logic, but for now I will leave it as it is, otherwise I will have to redo the menu for elements, resources, and files. Because visual unity will be lost :) It makes sense to do it as a separate PR, for all menus in the tree at once.

JoshuaLuckers:
It's great to have more consistency! LGTM!

Ibochkarev (approved):
🥇

Issue #15745, opened 2021-06-23
Inconsistency in title when copying an item in grids


Ruslan-Aleev:
### Summary
When copying elements in a grid, the names of copies of elements in different grids are set differently:
- With or without the translated word "copy".
- With a different position of the word "copy" at the beginning / end of the title.

Media Sources
![sources](https://user-images.githubusercontent.com/12523676/123061691-ec167680-d41c-11eb-9a03-ad186e23b8f0.png)

Users
![users](https://user-images.githubusercontent.com/12523676/123061698-ed47a380-d41c-11eb-9d93-b9774378118e.png)

### Expected behavior
I think the word "copy" should be at the end so that the items are next to each other.
And the translation is not particularly needed, in my opinion.

### Environment
MODX 3.0.0-alpha3

Ibochkarev:
Fixed in #15755

PR #15755, opened 2021-06-24
Make naming consistent when duplicating user in grid

Ruslan-Aleev:
### What does it do?
Fixed `getNewName()` when duplicating user in grid.
Because all other grids work according to the same pattern, in this PR I have brought `getNewName()` to the same pattern. It turned out that this function worked this way only for the users grid.

Before
users

After
users_copy

Media Sources, for example
sources

### Why is it needed?
UI/UX

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15745

### DIFF
--- a/core/src/Revolution/Processors/Security/User/Duplicate.php
+++ b/core/src/Revolution/Processors/Security/User/Duplicate.php
@@ -40,8 +40,10 @@ class Duplicate extends DuplicateProcessor
     public function getNewName()
     {
         $name = $this->getProperty('new_username', '');
-        $newName = !empty($name) ? $name : $this->modx->lexicon('duplicate_of',
-            ['name' => $this->object->get('username')]);
+        $newName = !empty($name) ? $name : $this->modx->lexicon(
+            'duplicate_of',
+            ['name' => $this->object->get('username')]
+        );
 
         return $newName;
     }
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Would you be so kind to fix the code style issues?

core/src/Revolution/Processors/Security/User/Duplicate.php
```diff
@@ -33,12 +33,17 @@ class Duplicate extends DuplicateProcessor
...
+     * @return string
      */
     public function getNewName()
     {
         $name = $this->getProperty('new_username', '');
-        return !empty($name) ? $name : $this->object->get('username') . '_copy';
+        $newName = !empty($name) ? $name : $this->modx->lexicon('duplicate_of',
+            ['name' => $this->object->get('username')]);
+
+        return $newName;
```
Ibochkarev:
```suggestion
        $newName = !empty($name) ? $name : $this->modx->lexicon(
            'duplicate_of',
            ['name' => $this->object->get('username')]
        );

        return $newName;
```

Ibochkarev (changes requested):

Ruslan-Aleev:
@JoshuaLuckers Fixed :)

JoshuaLuckers (approved):

Ibochkarev (approved):

Issue #15733, opened 2021-06-03
[Proposal] Remove "manager_direction" setting from manager panel


Ruslan-Aleev:
### Summary
Many people don't guess, but in MODX you can change the direction of the text, through the `manager_direction` setting (`ltr` by default). However, changing the setting to `rtl` makes the manager area almost unusable.

**I think it's easier to remove it completely. It is unlikely that anyone often uses it, and it is unlikely that anyone will fix it.**

![manager_direction_rtl](https://user-images.githubusercontent.com/12523676/120633361-1e1a6580-c47b-11eb-93f3-4f1ca57289dc.gif)

### Step to reproduce
Change the `manager_direction` setting to "rtl".

### Environment
MODX 3.0.0-alpha3

Mark-H:
I disagree with the proposal to remove it and would rather see the biggest issues with it fixed. Just briefly checked 2.x and that seems to work better than 3, although that also has some issues mostly with the tree and action buttons overlapping titles. 

Unfortunately usage of ltr/rtl is not a metric that's tracked in SiteDash so I don't have any data, but I think this is too important for some regions (even if it's not perfect) to remove entirely.

Ruslan-Aleev:
I agree with you, but I don’t think this is **too important**. Otherwise, it would be relevant for other CMS, but we do not see it.

alroniks:
I don't agree about removing RTL functionality, but I would remove it from the system settings and make it rely on language. List of RTL languages is known and limited, so it can be used for determining which text direction to use.

Mark-H:
> I agree with you, but I don’t think this is too important. Otherwise, it would be relevant for other CMS, but we do not see it.

WordPress has it built in based on installed language (plus standardised support for front-end themes), and a very quick google shows Kirby,  Plone, Craft, Joomla, OctoberCMS and others also offer it either out of the box or with a plugin. Seems harder to find examples that don't offer any type of support for it ;) 

> I don't agree about removing RTL functionality, but I would remove it from the system settings and make it rely on language. List of RTL languages is known and limited, so it can be used for determining which text direction to use.

This proposal I would also support, especially now that language switching is done dynamically.

Ruslan-Aleev:
> Seems harder to find examples that don't offer any type of support for it ;)

I was wrong. I was looking for exactly switching the direction of the text, but as it turned out, the direction of the text is connected with the selected language.

JoshuaLuckers:
[...]

PR #15758, opened 2021-06-30
Corrected styles for right-to-left text direction

Ruslan-Aleev:
### What does it do?
1) Corrected styles for right-to-left text direction.
Basically made the styles symmetrical, like ` margin-right: 10px; ->  margin: 0 10px;`

Before:
manager_direction_rtl

After:
after

2) Changed styles of submenu position.
If the `top`, `right`, `bottom` and `left` css-attributes are set, then these attributes are combined into a single `inset` attribute. I don't know if this is the behavior of new browsers or something else.
But with `inset` the menu is not positioned correctly.
Removed `bottom` and `right` attributes, since and without them the menu is positioned normally.

Before:
m_before

After:
m_after

**See the `manager_direction` setting. To test this you can change the system setting `manager_direction` to `rtl`.**
Please test PR carefully, both with the direction of the text from right to left, and in the opposite.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15733


Ruslan-Aleev:
p.s. as it turned out, in the 2nd branch it is still worse with the direction from right to left :)

JoshuaLuckers:
To test this you can change the system setting `manager_direction` to `rtl`.

Ruslan-Aleev:
@Mark-H when you have time, test the PR.
p.s. Because you added this issue to your "MODX3 Final Sprint" table :)

JoshuaLuckers (approved):

Mark-H:
Great work 🙌 

It does seem to have a bug where there's a horizontal scroll on the main panel when it's set to rtl:

afbeelding

Seems to go on for like 10x the window size... doesn't show when set to ltr. I'm not immediately seeing where it's coming from, unfortunately :/ 

(Tested on Firefox 90, win10)

JoshuaLuckers:
@Mark-H @Ruslan-Aleev I pushed a commit with a fix for this issue.

Ruslan-Aleev:
@JoshuaLuckers Thank you, I will test it soon :)

Ruslan-Aleev:
Everything seems to work as it should :)

PR #15759, opened 2021-07-02
Change order of menu items for Directory/File in tree

Ruslan-Aleev:
### What does it do?
Change order of menu items for Directory/File in tree. The menu items in tree were not grouped, they are now.
For visual consistency with other context menus in the tree.

Directory before:
d_before

Directory after:
directory_after

File .png before:
f1_before

File .png after:
file_png_after

File .php before:
f2_before

File .php after:
file_php_after

Also added a line separator for the Delete item in the context menu for "Elements" / "Contexts":

context_delete

elements_delete

### Why is it needed?
UI/UX

### Related issue(s)/PR(s)
Logically related with https://github.com/modxcms/revolution/pull/15749


JoshuaLuckers:
My thoughts:
- I usually want to upload and/or creates files when interacting with directories via the context menu. 
- I think destructive actions (e.g delete directory) should be separated from the other items with a line separator. 

Thanks for taking time making the manager more consistent!

Ruslan-Aleev:
@JoshuaLuckers Can you show it visually, what it would look like? And it is advisable to take into account other context menus (context, resources, elements, etc.).
And I would lead to a common style :)

JoshuaLuckers:
The first beta of MODX 3 is a great opportunity to include a change like this and see what the public thinks about the new menu structure.

muzzwood:
I just had a play around with this - thanks for the effort you put in. I agree with @JoshuaLuckers that the destructive actions should be separated i.e. in the "before" images the delete is at the bottom separated by a line.

Also, I don't think Set Visibility needs to be separated in the directory menu, because it only applies to the directory visibility, not the file visibility.

I think if you were to swap the delete and set visibility around that would be the most intuitive, to me at least.

Ruslan-Aleev:
Yes, I wanted to move the "delete" item below, but I need to redo it for other sections too, I'll deal with it later.

Ruslan-Aleev:
@JoshuaLuckers @muzzwood I moved the items and added a separator, see screenshots.
p.s. For the category, the "Delete" item  I did not move to the bottom, because for the rest of the menus it also looks different (see TVs, contexts, resource).
As I understand it, the logic is as follows: first, actions with this entity (including delete), and then something additional - create a file / category, etc.

Personally, I like this option better: it matches the design, and the "delete" item is always separate.

muzzwood:
@Ruslan-Aleev Looks good to me, nice work! :)

Ibochkarev (approved):
изображение

изображение

изображение

изображение

JoshuaLuckers (approved):

SnowCreative:
I know this is already merged, but I don't understand why "delete" isn't the last item in every list. For example, for TVs, isn't creating a TV or category a far more common action than deleting? It seems to me like those should be higher in the list. Same for directories — the original order makes the most sense to me, with upload second and delete at the bottom.

Ruslan-Aleev:
@SnowCreative Yes, that makes sense. In this PR, I basically lead menu to a single pattern (See elements, context and resource. There I did not change the order, but only added a separator before "Delete").
You can create a new PR, and move the "Delete" below for all menus. Or I will create a new PR later.
Although, as I understood (but maybe I'm wrong), there was such logic in the menu: first, actions with **this entity** (including delete), and then something additional - create a file / category, etc.

wfoojjaec:
I prefer the default context menu template in native OS style. First, we have actions like "view/open/send" (interaction and preview?). Then "cut/copy/paste" (edit?). "Delete" is separate, "rename" is separate. And "properties" along with internal actions, base on item type, go last. So merged variant makes more sense to me, but it is just my personal preference.

What is more important, all menus should follow the same style, no matter which style we pick.

PR #15776, opened 2021-07-29
Hide user settings tab based on permissions

Ibochkarev:
### What does it do?

Same as in https://github.com/modxcms/revolution/pull/15756, but for 3.x branch

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15756

### DIFF
--- a/core/src/Revolution/Processors/System/Settings/GetAreas.php
+++ b/core/src/Revolution/Processors/System/Settings/GetAreas.php
@@ -24,12 +24,14 @@ use xPDO\Om\xPDOQuery;
  */
 class GetAreas extends Processor
 {
+    public $permission = 'settings';
+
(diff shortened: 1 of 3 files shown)


opengeek (approved):

sonicpunk:
@Ibochkarev This PR has introduced a bug in the rendering of the settings tab on the Users Admin UI.

https://github.com/modxcms/revolution/pull/15776/files#diff-04fff8c779aee1e37df12d7f3ac420b03cd447a7acc6acfa3cc94d71f294a3f2R54

The permission is called "namespaces" and not "namespace".

The issue is that currently only sudo users will see the user settings tab. Line 54 always results in 0 because the permission "namespace" does not exist.

PR #15777, opened 2021-07-29
Assorted modRest improvements

Mark-H:
### What does it do?
I started working through upgrading the internal modRestClient usage (in the package manager) to modRest, but ran into various issues. Each fix is a separate commit, with a detailed description on each to explain the what and why.

### Why is it needed?

Make modRest work properly.

### How to test
The package manager would be one way when I finish that, but after a brief chat with Jason I'll be looking at implementing a PSR18 implementation for that instead. 

### Related issue(s)/PR(s)
N/a

For reference, the in-progress package manager update to modRest is here: https://github.com/Mark-H/revolution/compare/3.x-modrest...Mark-H:3.x-modrest-pm?expand=1 - that's functional from browsing to downloading and updating, in case the PSR-18 implementation doesn't make it.


core/lexicon/en/workspace.inc.php
```diff
@@ -77,6 +77,7 @@
 $_lang['package_download_err_ns'] = 'Please select at least one package version to download.';
 $_lang['package_err_create'] = 'Could not create the package.';
 $_lang['package_err_file_read'] = 'Could not open file for reading: [[+source]]';
+$_lang['package_err_caught'] = 'Installed failed with [[+type]] in [[+in]]: [[+message]]';
```
opengeek: Should this be present tense? i.e. "Install failed"…

Mark-H: ... yes

PR #15780, opened 2021-07-30
Leave Element/GetClasses processor alone

Mark-H:
### What does it do?

Remove the Element/GetClasses processor from the list of files to be removed on upgrade.

### Why is it needed?

This was added to the list of files to delete in #15079, however is actually an unrelated processor to the modClassMap. It's part of the property sets UI which alloes selecting what type of element a property (set) can be used for.

### How to test

Make sure Element/GetClasses is not removed on upgrade. (Requires building first.)

### Related issue(s)/PR(s)
#15079


opengeek:
Great catch!

PR #15784, opened 2021-08-04
Css fix

wax100:
### What does it do?
Fix css

### Why is it needed?
Some days  ago @ilyautkin made fixes for mobile optimization. Somehow manager css files were broken. 
I run grunt and generate new css files

### How to test
Install current modx version. Check manager. 
replace CSS files with files from this branch. Check 

 Before:
error
After:


wax100:
Incorrect PR

PR #15757, opened 2021-06-29
Upgrade flysystem to v2

muzzwood:
### What does it do?
Upgrades Flysystem to v2.x which contains a lot of breaking changes so a fair bit of refactoring has been involved.

You can read about some of the changes in v2 here:
https://flysystem.thephpleague.com/v2/docs/what-is-new/

### Why is it needed?
Using the latest version in preparation for the MODX 3 beta.

### How to test
Test file manipulation in the file/media browser.

### Related issue(s)/PR(s)
Related to #15725 though that issue was with v1.

PR #15786, opened 2021-08-04
Reworking the S3 System to return objects

matdave:
### What does it do?
Rewrite on the s3 container to list files and folders with the limitations of an S3 endpoint in mind. 

### Why is it needed?
Previous version was depending on returning "visibility" and "mimeType" functions that were not working with the S3 endpoint, resulting in an empty container. 

### How to test
Connect this to an S3 bucket on MODX

### Related issue(s)/PR(s)
Resolves issues introduced in #15757
Extends fixes of issues originally detected in #15785

## Side note
This is a bit of hacking between what was changed in the #15757 update and some fixes I had been working on prior. I am open to suggestions for cleanup.

### DIFF
--- a/core/src/Revolution/Sources/modS3MediaSource.php
+++ b/core/src/Revolution/Sources/modS3MediaSource.php
@@ -5,7 +5,11 @@ namespace MODX\Revolution\Sources;
 use Aws\S3\S3Client;
 use Exception;
 use League\Flysystem\AwsS3V3\AwsS3V3Adapter;
+use League\Flysystem\DirectoryAttributes;
 use League\Flysystem\FilesystemException;
+use League\Flysystem\UnableToDeleteDirectory;
+use League\Flysystem\UnableToMoveFile;
+use League\Flysystem\UnableToReadFile;
 use League\Flysystem\UnableToRetrieveMetadata;
 use League\Flysystem\Visibility;
 use MODX\Revolution\modX;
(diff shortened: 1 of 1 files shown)


matdave:
ALL CHECKS PASS!!! Didn't expect so many CS issues from copy/pasting code XD

theboxer (approved):

core/src/Revolution/Sources/modS3MediaSource.php
```diff
@@ -292,16 +294,206 @@ public function getObjectUrl($object = '')
...
+        }
+
+        $bases = $this->getBases($path);
+        $imageExtensions = explode(',', $properties['imageExtensions']);
+        $skipFiles = $this->getSkipFilesArray($properties);
+        $allowedExtensions = $this->getAllowedExtensionsArray($properties);
+
+        $directories = $dirNames = $files = $fileNames = [];
+
+        try {
```
JoshuaLuckers, line 345: This is a really big try/catch block. It would be nice if it wasn't this big.

core/src/Revolution/Sources/modS3MediaSource.php
```diff
@@ -292,16 +294,206 @@ public function getObjectUrl($object = '')
...
+            $re = '#^(.*?/|)(' . implode('|', array_map('preg_quote', $skipFiles)) . ')/?$#';
```
JoshuaLuckers: Can we have a comment explaining what's happening here and why?

matdave:
I just went through and found some other broken functions from the manager. I've added a few more to the file. 

I've also opened an issue here: https://github.com/thephpleague/flysystem/issues/1343

core/src/Revolution/Sources/modS3MediaSource.php
matdave: This is the skip files logic from the base modMediaSource->getContainerList() function.

core/src/Revolution/Sources/modS3MediaSource.php
matdave: This is from the base modMediaSource->getContainerList() function. The try was for the flysystem->ListContents(), so I've wrapped it more closely.

PR #15799, opened 2021-08-28
Enable friendly url functionality by default for new installations

JoshuaLuckers:
### What does it do?
Enables `friendly_urls` and `use_alias_path` for new MODX installations.

### Why is it needed?
To make MODX more SEO friendly by default.

### How to test
Do a clean install and open a resource.

### Related issue(s)/PR(s)
Issue #13876

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -670,7 +670,7 @@ $settings['friendly_alias_word_delimiters']->fromArray([
 $settings['friendly_urls'] = $xpdo->newObject(modSystemSetting::class);
 $settings['friendly_urls']->fromArray([
   'key' => 'friendly_urls',
-  'value' => false,
+  'value' => true,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'furls',
@@ -1816,7 +1816,7 @@ $settings['upload_translit']->fromArray([
 $settings['use_alias_path'] = $xpdo->newObject(modSystemSetting::class);
 $settings['use_alias_path']->fromArray([
   'key' => 'use_alias_path',
-  'value' => false,
+  'value' => true,
   'xtype' => 'combo-boolean',
   'namespace' => 'core',
   'area' => 'furls',


Ibochkarev (approved):
изображение

Mark-H:
What about rewrite rules? As-is people are gonna get 404s on their friendly urls. That's why it's never been on by default, the extra step that's needed...

opengeek:
Yeah, this is a non-starter for me. The manual steps required to make this work necessitate this being off by default, in my opinion.

SnowCreative:
> As-is people are gonna get 404s on their friendly urls. That's why it's never been on by default, the extra step that's needed...

Not sure why this got closed. Why not just un-comment the block in .htaccess that turns on the FURLs?

Mark-H:
There is no default `.htaccess` file that's enabled - there's a `ht.access` sample file people can use, but making that part of the distribution as `.htaccess` would cause user changes to get overwritten on upgrades. 

I suppose in theory it could be made that it only creates a .htaccess from the sample when there is no .htaccess file already, _however_ there also is no guarantee a .htaccess file will do _anything_ - MODX can also be used on nginx, IIS, etc, which have their own rewrite mechanisms. 

It does seem to me enabling `use_alias_paths` by default could be beneficial, though.

SnowCreative:
So, Wordpress doesn't work on nginx, for example? If it does, then how does it handle rewrites?

opengeek:
> So, Wordpress doesn't work on nginx, for example? If it does, then how does it handle rewrites?

Rewrites are done through server/virtual host configurations in nginx. There are no .htaccess files.

Mark-H:
https://wordpress.org/support/article/nginx/

SnowCreative:
So, what would be wrong with having furls turned on by default in the .htaccess file, and people using nginx or whatever will have to set it up manually? Since MODX advertises itself as being SEO-friendly right out of the box, this seems like a good idea to me. I mean, if Wordpress can do it . . .

Mark-H:
Does WordPress really have furls turned on by default, though? I'm clearly not a WP expert, but it [sounds to me](https://wordpress.org/support/article/using-permalinks/) like its default option is `/?p=foo` which is not the same as `/foo` - that would be the equivalent of `/?q=foo` that MODX has. That option doesn't require a .htaccess file or rewrite rule at all. 

You could certainly make the argument that ought to be a new default instead of `/?id=5`. 

WordPress does offer the ability to write the .htaccess file (if it has permission to do so) with [very similar contents to ours](https://wordpress.org/support/article/htaccess/). As the average MODX user is going to be more technically equipped than WordPress', implying the odds of that file having been customised by a user in a way that may or may not be compatible with the default rules necessary for furls, I'm really not a big fan of the idea of writing to that automatically.

SnowCreative:
cPanel writes to .htaccess as well, but it doesn't delete any custom content already entered there. Anyway, if FURLs were turned on by default in the stock htaccess file and system settings, then that would only apply to new installations, before anyone has customized .htaccess. Anyone using nginx is most likely going to be versed on how to update the server settings manually. Do people using MODX Cloud have to enter the FURL settings manually for every new cloud, or is turning it on via the system settings all that is required?

Mark-H:
Cloud has the rewrite rules by default, as that has full server control just like cPanel. MODX does not have that full server control on a very significant portion of servers.

Let me clarify that I'm not against automatically enabling FURLs by default if it does so reliably across the wide range of potential web server configurations MODX supports. And leaves it off, when it failed to enable the rewrite rules. To date nobody has provided a proposal that does that but I'm always happy to be proven wrong!

Issue #15584, opened 2021-03-24
Unnecessary tabs for the anonymous user group


sergant210:
### Summary
We have a user group for anonymous users. It is virtual. There are 4 tabs in the group editing interface. 

![User Group anonymous  tabs](https://user-images.githubusercontent.com/4679725/112261984-ed010600-8c7d-11eb-801e-3c0d06d39bc3.jpg)

The last two require a real group, not a virtual one. Therefore, trying to add a user or a setting to the anonymous group will cause an error.

![User Group anonymous](https://user-images.githubusercontent.com/4679725/112262133-2d608400-8c7e-11eb-8eaf-791a5a2c085d.jpg)

### Expected behavior
For the anonymous group these tabs should not be displayed.

### Environment
MODX 2.x, MODX 3.x.

Ruslan-Aleev:
How did you get into the anonymous group edit form? Through direct link? If so, then this is not really a bug :)

sergant210:
> How did you get into the anonymous group edit form? Through direct link?

Magic )

![Access Control Lists  MODX Revolution — Mozilla Firefox](https://user-images.githubusercontent.com/4679725/112278632-19278180-8c94-11eb-8159-f8a01c3cc075.jpg)

Ruslan-Aleev:
In my opinion, it is more logical to turn off the menu for anonymous group. Because anonymous group shouldn't be edited or not?

sergant210:
This group should be able to configure access permissions for contexts and resource groups.

Ibochkarev:
Fixed #15802

PR #15802, opened 2021-08-29
Fix Tab Visibility in User Group Editing Panel

smg6511:
### What does it do?
Added listeners to hide tabs not applicable to the anonymous group (both in the main user group panel and the permissions vertical tab panel). Also ensured user group list is hidden when the logged-in manager user does not have permission to view it.

### Why is it needed?
Corrects tab visibility for user group editing panels

### How to test

1. Test as an admin with full permissions. When editing user groups other than anonymous, you should be able to view and manipulate user lists as expected.
2. Remove the _usergroup_user_list_ permissions for your manager user group and test again, noting that user lists should not be visible for any group in the user group editing panel.
3. When editing the anonymous group (regardless of the _usergroup_user_list_ permissions setting), you should see only the "General Information" and "Access Permissions" tabs. Further, within the Access Permissions tab, only the "Contexts" and "Resource Groups" vertical tabs should be visible.

### Related issue(s)/PR(s)
Fixes issue in #15584

### DIFF
--- a/manager/assets/modext/widgets/security/modx.panel.user.group.js
+++ b/manager/assets/modext/widgets/security/modx.panel.user.group.js
@@ -302,7 +302,7 @@ MODx.panel.UserGroup = function(config) {
                         tabPanel.hideTabStripItem(usersPanelKey);
                         tabPanel.hideTabStripItem(settingsPanelKey);
                     }
-                    if (!MODx.perm.usergroup_user_list ) {
+                    if (!MODx.perm.usergroup_user_list) {
                         tabPanel.hideTabStripItem(usersPanelKey);
                     }
                 }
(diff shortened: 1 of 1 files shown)


opengeek:
I just want to clarify a point and get some clarification in return. The "anonymous" user group has nothing to do with manager users, at all. This is something that exists ONLY to apply permissions in other contexts to visitors who have a session but are NOT logged in as a user. You cannot access the mgr context without being logged in as a user, so I'm a little confused about the usage of the "anonymous" user group in this PR's description.

smg6511:
Correct, and that's why the tabs that get hidden by the changes in this PR should not be shown when editing the "anonymous" group. It's only a matter of semantics - I did say "user group" but that's only because when editing in the manager (anonymous) falls under User Groups. I just went ahead and removed the word "user" in a couple places in the description for clarity. We're on the same page ;-)

Ibochkarev (approved):
image
image

JoshuaLuckers (approved):

PR #15797, opened 2021-08-25
Prevent line lengths over 120 from throwing errors

opengeek:
### What does it do?
Overrides the hard line-length limit on the PHPCS check.

### Why is it needed?
Prevent PHPCS from failing code with lines longer than 120 characters.

### How to test
N/A

### Related issue(s)/PR(s)
N/A

Issue #15790, opened 2021-08-10
Fine-Tune PHPCS Config So PRs Without Errors Pass Code Quality Test


smg6511:
### Summary
PHPCS is problematic in its current configuration, failing entire PRs with as little as one warning. A more defined configuration should be created.

### Why is it needed?
Maintainers are going to get a bad impression of or, at worst, ignore non-passing PRs when the "issues" the linter raises may be trivial.

### Suggested solution(s)
- At the very least, the config option _ignore_warnings_on_exit_ should be set to "1" so warnings are shown but do not fail the code quality test.
- Consider increasing the Generic.Files.LineLength limits. IMO the current 120 char limit is too short and can sometimes result in less readable code due to having to unnaturally break up a line.
- Need to address whether MODX will continue to camelCase class names, as it will require a custom rule and sniffer if so.

smg6511:
Another consideration for the codeowners/maintainers: As I've tackled some issues and new features since the implementation of this linter I've noticed that, because there are a litany of changes necessary to pass (or at least attempt to pass) the code quality tests, spotting the actual _content_ changes to PR files can be quite the chore. Would it be better to drop enforcement of PSR-12 now and dedicate a time period (let's say a week or two) where normal PRs would be on hold and the entire php codebase could be brought up to speed PSR-12-wise. Perhaps you could recruit a group of contributors/maintainers/etc and assign specific areas each could update (I'd certainly be willing to be one).

opengeek:
I agree. We need to disable it until it can be tackled at once as a final step before preparing for beta release. I want to see the changes being made and do not care about the formatting at this point.

smg6511:
This most important problems have been solved here, so closing issue.

PR #15803, opened 2021-08-31
Prevent line lengths over 120 from throwing errors

JoshuaLuckers:
### What does it do?
- Overrides the hard line-length limit on the PHPCS check
- Disable fail on warnings for `thenabeel/action-phpcs`

### Why is it needed?
Prevent PHPCS from failing code with lines longer than 120 characters.

### How to test
No idea what's the best approach to test this. 

### Related issue(s)/PR(s)
Issue #15790
PR #15797

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -20,4 +20,10 @@
 
     <!-- Our base rule: set to PSR12-->
     <rule ref="PSR12" />
+    <rule ref="Generic.Files.LineLength">
+        <properties>
+            <property name="lineLimit" value="120" />
+            <property name="absoluteLineLimit" value="0" />
+        </properties>
+    </rule>
 </ruleset>
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -20,10 +20,4 @@
 
     <!-- Our base rule: set to PSR12-->
     <rule ref="PSR12" />
-    <rule ref="Generic.Files.LineLength">
-        <properties>
-            <property name="lineLimit" value="120" />
-            <property name="absoluteLineLimit" value="0" />
-        </properties>
-    </rule>
 </ruleset>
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
@opengeek I believe this change will do the trick!

Mark-H (approved):

phpcs.xml
```diff
@@ -20,4 +20,10 @@
 
     <!-- Our base rule: set to PSR12-->
     <rule ref="PSR12" />
+    <rule ref="Generic.Files.LineLength">
+        <properties>
+            <property name="lineLimit" value="120" />
+            <property name="absoluteLineLimit" value="0" />
+        </properties>
+    </rule>
```
opengeek: I don't think this override is necessary as this is what PSR-12 defines already.

opengeek (changes requested):

phpcs.xml
smg6511: Right, it's really just the fail_on_warnings change that's needed to keep warnings from failing the entire code quality check. That said, as I mentioned in issue #15790 and @opengeek concurred on, we should just disable this code quality step altogether until a defined time period where updating the whole codebase to be PSR-12 compliant is the sole or at least main focus.

opengeek: Actually, I don't see any reason to disable it. It's not blocking merging as it is not a required check. It's nice to have the information, but we need to make it clear to contributors that they do NOT need to attempt to fix the files they change for now.

smg6511: I don't know, it's pretty difficult spotting actual content changes if someone has gone to the effort of fixing all the formatting, especially in larger php files. The trouble is that many of the code checks cause an error, not a warning.  The one particularly problematic one is the casing of class names, which will have to be either carefully changed or a custom sniff would have to replace the default PSR-12 one to allow the current style. Is there a way to allow the errors and warnings to show but not cause failure of the code check? Anyway, just something to consider ;-)

JoshuaLuckers: If the formatting changes are done in a separate commit it isn't that big of a problem.

opengeek: Until I squash and merge it isn't. 😛

opengeek (approved):

PR #15804, opened 2021-09-03
Lazy load thumbnails in media browser

matdave:
### What does it do?
Add the loading="lazy" attribute to images in the media browser.

### Why is it needed?
Opening large folders in the media browser can cause hundreds of phpthumb requests all at once. 

### How to test
In most modern browsers, the loading="lazy" attribute will prevent requesting an image until it is within the screen. You can test this by opening up the media browser and seeing the image requests as you scroll down a large container source. 

### Related issue(s)/PR(s)
This was an internal annoyance


Ibochkarev (approved):

matdave:
Same thing as #15805 I just added a loading mask indicator for larger folders.

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -98,20 +98,20 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
 
     ,run: function(p) {
         p = p || {};
-        if (p.dir) { this.dir = p.dir; }
         Ext.applyIf(p,{
             action: 'browser/directory/getFiles'
-            ,dir: this.dir
+            ,dir: p.dir ?? this.dir
```
JoshuaLuckers: This will break `grunt build`.

matdave: reverted it

JoshuaLuckers (approved):

PR #15805, opened 2021-09-03
Lazy load thumbnails in media browser

matdave:
### What does it do?
Add the loading="lazy" attribute to images in the media browser.

### Why is it needed?
Opening large folders in the media browser can cause hundreds of phpthumb requests all at once. 

### How to test
In most modern browsers, the loading="lazy" attribute will prevent requesting an image until it is within the screen. You can test this by opening up the media browser and seeing the image requests as you scroll down a large container source. 

### Related issue(s)/PR(s)
related to #15804


Ibochkarev (approved):

matdave:
I've added a "Loading Mask" indicator when selecting a new folder. Previously it would load silently and looked as if nothing happened when a folder had lots of files.

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -58,6 +58,7 @@ MODx.browser.View = function(config) {
         }
         ,tpl: MODx.config.modx_browser_default_viewmode === 'list' ? this.templates.list : this.templates.thumb
         ,itemSelector: MODx.config.modx_browser_default_viewmode === 'list' ? 'div.modx-browser-list-item' : 'div.modx-browser-thumb-wrap'
+        ,loadingText : '<div style="padding:10px;"><h4>'+_('loading')+'</h4></div>'
```
Ruslan-Aleev: Maybe it's better not to add layout with styles to extJS? Either specify the class, or just leave the lexicon as you did for the 2.x branch.

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -101,17 +102,18 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
 
     ,run: function(p) {
         p = p || {};
-        if (p.dir) { this.dir = p.dir; }
         Ext.applyIf(p,{
             action: 'Browser/Directory/GetFiles'
-            ,dir: this.dir
+            ,dir: p.dir ?? this.dir
```
JoshuaLuckers: When building the assets with `grunt build` this throws an error:
```
 [SyntaxError]: Unexpected token: operator «?»

 filename: 'widgets/media/modx.browser.js',
  line: 107,
  col: 25,
  pos: 4294
```

JoshuaLuckers (changes requested):

manager/assets/modext/widgets/media/modx.browser.js
matdave: I'm just keeping it consistent with how the "Error/Empty" message loads. This line could be taken out because I don't think it gets triggered very often anyway.

matdave: I've just removed it

manager/assets/modext/widgets/media/modx.browser.js
matdave: I've reverted it

JoshuaLuckers (approved):

PR #15809, opened 2021-09-05
Fix issues for plugins that pass around xPDOObjects

### LINKED ISSUE #15172: modElement getTag function raises fatal error

### LINKED ISSUE #15657: MODX3 — OnResourceSort doesn't work
### Summary
OnResourceSort doesn't seem to fire.

### Step to reproduce
Create a plugin that does anything as a result of OnResourceSort firing.

### Observed behavior
Nothing happens.

### Environment
MODX 3 Alpha 3 2021-04-06

JoshuaLuckers:
### What does it do?
When serializing the element properties, check if it is an xPDOObject derivative and transform it into an array.

### Why is it needed?
To prevent an exception because we can't serialize xPDOObjects. This happens when objects are passed around.

### How to test
Run the newly added Unit Test or create a plugin that's listening to the `OnResourceSort` event and see it working properly.

### Related issue(s)/PR(s)
Closes issue #15657
Closes issue #15172
Supersedes PR #15467

### DIFF
--- a/_build/test/Tests/Model/Element/modElementTest.php
+++ b/_build/test/Tests/Model/Element/modElementTest.php
@@ -150,4 +150,24 @@ class modElementTest extends MODxTestCase {
             ],
         ];
     }
+
+    /**
+     * Test the modElement->getTag() method with xPDOObjects as values.
+     * E.g the nodes when the event `OnResourceSort` is fired
+     *
+     */
+    public function testGetTagWithXPDOObjects()
+    {
+        /** @var modElement $element */
+        $element = $this->modx->newObject(modElement::class);
+        $element->getProperties([
+            'objects' => [
+                $this->modx->newObject(modElement::class),
+            ],
+            'object' => $this->modx->newObject(modElement::class),
+        ]);
+
+        $tag = $element->getTag();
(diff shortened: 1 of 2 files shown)


opengeek (approved):

theboxer (approved):

Ibochkarev:
@JoshuaLuckers please close linked issues

opengeek:
> @JoshuaLuckers please close linked issues

FWIW, these would have closed automatically if the word "issue" would have been left out between the keyword "closes" and the actual issue number.

Issue #15541, opened 2021-03-12
MODX3 button colors seem wrong


SnowCreative:
### Summary
Buttons in popup windows are gray now instead of blue

### Step to reproduce
For example, delete a resource, go into the trash, click Purge All. 

### Observed behavior
The popup window has yes and no buttons, with the background color of yes being gray and no white

### Expected behavior
Primary button should be blue, not gray. Gray makes it look like the button is deactivated.

MODX 3 popup:

![Screen Shot 2021-03-12 at 2 48 44 PM](https://user-images.githubusercontent.com/2528374/110991722-fff10d80-8342-11eb-842d-a4163c46a17d.png)

Same popup in MODX 2.8:
![Screen Shot 2021-03-12 at 2 56 05 PM](https://user-images.githubusercontent.com/2528374/110991816-18f9be80-8343-11eb-9539-ef291edb054f.png)

### Environment
MODX 3 alpha 3 20210309

### Solution
The "index.css" file has this:

 .x-btn-over.x-btn, .actions button.x-btn-over,
  .x-btn-over.inline-button,
  .x-btn-over.x-superboxselect-item,
  .x-btn-over.x-form-trigger,
  .x-btn-over.x-date-mp-ok,
  .x-btn-over.x-date-mp-cancel, .x-btn:hover, .actions button:hover,
  .inline-button:hover,
  .x-superboxselect-item:hover,
  .x-form-trigger:hover,
  .x-date-mp-ok:hover,
  .x-date-mp-cancel:hover, .x-btn-focus.x-btn, .actions button.x-btn-focus,
  .x-btn-focus.inline-button,
  .x-btn-focus.x-superboxselect-item,
  .x-btn-focus.x-form-trigger,
  .x-btn-focus.x-date-mp-ok,
  .x-btn-focus.x-date-mp-cancel, .x-btn:focus, .actions button:focus,
  .inline-button:focus,
  .x-superboxselect-item:focus,
  .x-form-trigger:focus,
  .x-date-mp-ok:focus,
  .x-date-mp-cancel:focus {
    background-color: #E4E4E4;
    box-shadow: #DCDCDC;
    color: #515151;
    /*button {
      color: inherit;
    }*/ }

The background color should be changed from #E4E4E4 to #3697cd (or whatever the theme color is for primary buttons), like it is in MODX 2.8.

Ruslan-Aleev:
Yes, it makes sense to make the buttons more visible.

smg6511:
[...]

PR #15826, opened 2021-10-02
Improve Button Styling in Modals

smg6511:
### What does it do?
Updates scss to show the correct primary button color in confirm, alert, and other message windows, making these modal windows consistent with the other modx-specific ones (usually involving a create/update action). 

### Why is it needed?
Some affirmative buttons were displaying a grey background instead of the primary button color (green in modx 3).

### How to test
From your terminal app, run `grunt build` from within the _build/templates/default folder. Then take various actions which trigger confirmation/alert dialogs—such as deleting a resource, purging deleted resources, deleting elements such as tvs (particularly one that's assigned to a template, as it will trigger an additional dialog), chunks, etc. Note how the affirmative buttons ("Save", "OK", etc.) should appear in the new primary button color (green).

### Related issue(s)/PR(s)
Addresses issue #15541.

### DIFF
--- a/_build/templates/default/sass/_windows.scss
+++ b/_build/templates/default/sass/_windows.scss
@@ -51,6 +51,10 @@
       padding: 8px;
       text-align: center;
     }
+
+    .x-window-header-text {
+        padding-left: 24px;
+    }
   }
 
   /* styles for a collapsed window */
@@ -110,6 +114,12 @@
       .x-window-body {
         padding-top: 15px;
       }
+    }
+
+    .x-window-tc {
+        .x-window-header-text {
+            padding-left: 64px;
+        }
     }
   }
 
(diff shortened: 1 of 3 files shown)


Mark-H (approved):

_build/templates/default/sass/_windows.scss
```diff
@@ -235,6 +245,29 @@
...
+          .x-toolbar-cell {
+              /*
+                  Note that because the MessageBox class in Extjs 3 is a singleton,
+                  it can not be extended or overriden and is very limited in its
+                  configuration options. Thus the fragile method here of selecting
+                  the button by dom position. Ext always outputs all 4 buttons into
+                  the markup and just hides ones not in use for a particular window type.
+                  For the time-being, the primary button will always be the
+                  first ("ok") or second ("yes", typically our save button) index.
+              */
```
JoshuaLuckers, line 261: Thanks for explaining why it's done this way!

JoshuaLuckers (approved):

_build/templates/default/sass/_windows.scss
smg6511: No problem ;-) As a side note, in Extjs 4 Sencha totally refactored the MessageBox class making it highly configurable and extendable. Should we be able and choose to upgrade Ext (to 7) in the future, this would be one of many advantages.

PR #15827, opened 2021-10-05
Fix error: Cannot find module 'sass' when running Compile Assets workflow

JoshuaLuckers:
### What does it do?
When testing the workflow it worked fine before.. until I merged it..
~It now installs sass, but there is still something wrong with the grunt build task. Therefore it's doing a `grunt build --force`.~

~This won't compile the assets but it doesn't fail the CI workflow. I hope to figure out what's causing this issue soon..~

Everything works as expected now!

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -28,4 +28,4 @@ jobs:
                 run: cd _build/templates/default && npm install
 
             -   name: Run Grunt build
-                run: cd _build/templates/default && grunt build --force
+                run: cd _build/templates/default && grunt build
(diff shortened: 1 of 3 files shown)


opengeek (commented):
I'm confused. It works fine here without this PR.

opengeek:
Wait. Is this related to your PR that you merged that adds a GitHub action? I'm still confused about why that was even needed.

opengeek (approved):

JoshuaLuckers:
@opengeek yes, it's related to that PR. Sorry for not being so clear, had to do this quickly.

The GitHub action makes sure we can compile the assets when someone changes them, previously there where some PR's where `grunt build` would fail. 

P.s: it only runs for PR's where assets are changed.

PR #15833, opened 2021-10-11
Fix create button visibility in tree nodes

### LINKED ISSUE #14630: Plus sign for quick create is shown in managing resource groups
### Summary
Since #13774 the plus sign for quick create is visible in managing resource groups.

### Step to reproduce
Visit manager/?a=security/resourcegroup and hover a resource entry on the right.

### Observed behavior
A plus sign will occur on the right with no functionality.

### Expected behavior
No plus sign should be shown.

### Environment
#13774

smg6511:
### What does it do?
Made css changes to hide the create button that appears in the resource tree nodes within the resource groups editing panel.

### Why is it needed?
The ability to create resources from the tree does not apply in the context of the resource groups panel.

### How to test
Clear your caches and remember to run `grunt build` within the __build/templates/default_ directory to compile the change into the main css file. Then navigate to _Content/Resource Groups_ to verify the fix.

### Related issue(s)/PR(s)
Resolves #14630.


Ibochkarev (approved):
It works, the creation icon is not displayed on hover

core/lexicon/en/access.inc.php
```diff
@@ -81,7 +81,7 @@
 $_lang['resource_group_add'] = 'Add Resource Group';
 $_lang['resource_group_access_remove'] = 'Remove Resource from Group';
 $_lang['resource_group_access_remove_confirm'] = 'Are you sure you want to remove this resource from the resource group?';
-$_lang['resource_group_create'] = 'Create Resource Group';
+$_lang['resource_group_create'] = 'Add Group';
```
Ruslan-Aleev: It is better not to change the lexicons for 2.x, firstly, the meaning does not change much here, and secondly, lexicons will appear that will need to be translated again in Crowdin.

smg6511: Ok, will resist lex changes in future for 2.x ;-)

Ruslan-Aleev:
If still relevant - disable the ability to click on resources in the tree, as in https://github.com/modxcms/revolution/pull/15746, I think it will not be superfluous.
And it's worth hiding the icon in the tree, in the "Import resources" section.

PR #15835, opened 2021-10-12
Fix active state for files in main tree

### LINKED ISSUE #15302: Incorrect display of paths for files (path mismatch)
### Summary
After merging PR https://github.com/modxcms/revolution/pull/15262 for an open file, the active class was no longer added to the file tree.

file_class

This happens because the validation condition (`node.id == activeFile`) is not working (see https://github.com/modxcms/revolution/pull/15228).
In `node.id` we have, for example, `images/product/file.txt`, and in `activeFile` (taken from url) we get `images%2Fproducts%2Ffile.txt`

### Step to reproduce
Open any file attached to the folder in the element tree - the active class will not appear. Although, if you open the file at the root, then there is no problem, because there is no `/` in the path and id.

Of course, we can just decode the `node.id` in the condition, but we need to be sure that there are no more problems of this kind. For example, the same kind of problem is relevant for 3.x - https://github.com/modxcms/revolution/issues/15244

**We need to bring it to a single format.**

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15244
https://github.com/modxcms/revolution/pull/15228
https://github.com/modxcms/revolution/pull/15227

### Environment
MODX 2.8.0

smg6511:
### What does it do?
Made a couple tweaks, one of which ensures the active file value (pulled from the query string) in decoded for comparison to the node id.

### Why is it needed?
Active files were not being highlighted in the main tree.

### How to test
Clear your caches and remember to run `grunt build` within the __build/templates/default_ directory to compile the changes into the main js file. Then navigate to the _Files_ tab on the main tree and click to open files at various levels of the hierarchy. Now the file that is open should be highlighted in the tree. Note that the behavior remains correct for Resources and Elements.

### Related issue(s)/PR(s)
Resolves #15302.


Ruslan-Aleev (approved):
Visually works as it should :)

fileActive

opengeek (approved):

Ruslan-Aleev:
@smg6511 Can you test again on your setup? It doesn't work for me for MODX 2.8.4. (Perhaps the same issue is causing https://github.com/modxcms/revolution/issues/16194)

smg6511:
Yes, I can confirm these issues in my 2.8.4-dev branch. The two files changed in this PR are exactly the same in 2.8.4, so something else has changed causing the bug. I'll poke around...

PR #15836, opened 2021-10-14
Form customization front end fixes/improvements

### LINKED ISSUE #15357: A label is cut off after applying form customization rule
### Summary
If you try to rename the label of any default field of the resource and use a comma in the text, it will fail. See images.
Screen Shot 2020-12-23 at 13 09 14
Screen Shot 2020-12-23 at 13 10 23

### Step to reproduce
Create any form of customization set and rule. In the rule, write a new name for any default field, like pagetitle, for example.
Use a comma in the name. After to go to the resource and check what label is shown.

### Observed behavior
Labels are cut off after the rule is applied.

### Expected behavior
Labels should be the same as it was set.

### Related issue(s)/PR(s)
N/A

### Environment
MODX 2.8.1

smg6511:
### What does it do?
Reworks the methods responsible for moving and changing field/tab visibility and labelling. Code optimizations were made and method documentation was added.

### Why is it needed?
Most of these methods were written to receive/process serialized/array data, but there appears to be no reason (any longer at least) to keep them that way—especially given that problems have arisen such as the one referenced by this PR. The changes allow for the use of commas in labels and provides simplification and clarification of the relevant code.

### How to test
Clear your caches and remember to run `grunt build` within the __build/templates/default_ directory to compile the changes into the main js file. Create one or two test TVs, then create a Form Customization profile with at least one rule. Within your rule(s), experiment with re-labelling and hiding various resource fields along with your test TVs. Also create a new tab in Regions tab and verify that assigning your TVs to that tab (or the standard regions) works as expected.

### Related issue(s)/PR(s)
Resolves #15357.

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -238,23 +238,23 @@ Ext.extend(MODx,Ext.Component,{
 			,listeners: {
 				'success': {
 					fn:function() {
-						var tree = Ext.getCmp("modx-resource-tree"); 
-						
+						var tree = Ext.getCmp("modx-resource-tree");
+
 						if (tree && tree.rendered) {
 							tree.refresh();
 						}
 
 						var cmp = Ext.getCmp("modx-panel-resource");
-						
+
 						if (cmp) {
 							Ext.getCmp('modx-abtn-locked').hide();
-							Ext.getCmp('modx-abtn-save').show();	
+							Ext.getCmp('modx-abtn-save').show();
 						}
 					},
 					scope:this
 				}
 			}
-		});  
+		});
     }
 
     ,sleep: function(ms) {
@@ -385,6 +385,7 @@ Ext.extend(MODx,Ext.Component,{
     }
 
     ,helpUrl: false
+
     ,loadHelpPane: function(b) {
         var url = MODx.helpUrl || MODx.config.help_url || '';
         if (!url || !url.length) { return false; }
@@ -403,34 +404,40 @@ Ext.extend(MODx,Ext.Component,{
             ,layout: 'fit'
 			,bodyStyle : 'padding: 0;'
             ,items: [{
-	        	xtype		: 'container',
-				layout		: {
-	            	type		: 'vbox',
-					align		: 'stretch'
+	        	xtype: 'container',
+				layout: {
+	            	type: 'vbox',
+					align: 'stretch'
 				},
-				width		: '100%',
-				height		: '100%',
-				items		:[{
-					autoEl 		: {
-		                tag 		: 'iframe',
-		                src			: url,
-		                width		: '100%',
-						height		: '100%',
+				width: '100%',
+				height: '100%',
+				items: [{
+					autoEl: {
+		                tag: 'iframe',
+		                src: url,
+		                width: '100%',
+						height: '100%',
 						frameBorder	: 0
 					}
 				}]
 			}]
-			//,html: '<iframe src="' + url + '" width="100%" height="100%" frameborder="0"></iframe>'
         });
         MODx.helpWindow.show(b);
         return true;
     }
 
-    ,addTab: function(tbp,opt) {
-        var tabs = Ext.getCmp(tbp);
-        if (tabs) {
-            Ext.applyIf(opt,{
-                id: 'modx-'+Ext.id()+'-tab'
+    /**
+     * Adds a new tab to the specified panel; method called from code inserted via modActionDom (modactiondom.class.php)
+     *
+     * @param {String} panelId - Text id of the tabPanel the new tab will be added to
+     * @param {Object} newTabConfig - The base configuration for the new tab
+     * @return {void}
+     */
+    ,addTab: function(panelId, newTabConfig) {
+        var tabPanel = Ext.getCmp(panelId);
+        if (tabPanel) {
+            Ext.applyIf(newTabConfig,{
+                id: 'modx-' + Ext.id() + '-tab'
                 ,layout: 'form'
                 ,labelAlign: 'top'
                 ,cls: 'modx-resource-tab'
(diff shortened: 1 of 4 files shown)


smg6511:
REVIEWERS: Note that the unit test for this (modActionDomTest.php) will need to be changed for this PR to pass. I'm just not sure _who_ should do that. Basically, the parameters given that are formatted as an array need to be changed to a string. Please advise.

Ibochkarev:
@smg6511 Please correct errors for passing tests

smg6511:
@Ibochkarev OK, the unit test is fixed.

manager/assets/modext/core/modx.js
```diff
@@ -515,45 +553,72 @@ Ext.extend(MODx,Ext.Component,{
...
+
+    /**
+     * Changes the label of the specified field; method called from code inserted via modActionDom (modactiondom.class.php)
+     *
+     * @param {String} containerId - Text id of the field's/container's container
+     * @param {String} fieldId - Text id or name of field/container whose label/title being renamed
+     * @param {String} newLabel - The replacement label text
+     * @return {void}
+     */
+    ,renameLabel: function(containerId, fieldId, newLabel) {
```
theboxer, line 565: This is quite a simple function, but written in unnecessary complex way - a lot of unnecessary nesting. I propose this approach:
```js
,renameLabel: function(containerId, fieldId, newLabel) {
    if (containerId === 'modx-panel-resource' && fieldId.indexOf('modx-resource-content') !== -1) {
        var container = Ext.getCmp('modx-resource-content');
        if (!container) return;
        
        if (container.setTitle) {
            container.setTitle(newLabel);
            return;
        }
        
        if (container.setLabel) {
            container.setLabel(fieldId, newLabel);
            return;
        }

        container.label.update(newLabel);
        return;        
    }
    
    var container = Ext.getCmp(containerId);
    if (!container) return;
  
    container.setLabel(fieldId, newLabel);        
}
```

manager/assets/modext/core/modx.js
```diff
@@ -442,30 +449,45 @@ Ext.extend(MODx,Ext.Component,{
...
+
+    /**
+     * Hides a region or tab; method called from code inserted via modActionDom (modactiondom.class.php)
+     * and from MODX.hideTab (above)
+     *
+     * @param {String} containerId - Text id of the region/tab's container
+     * @param {String} regionId - Text id of the region/tab to hide
+     * @return {void}
+     */
+    ,hideRegion: function(containerId, regionId) {
```
theboxer, line 472: Unnecessary complexity.

```js
,hideRegion: function(containerId, regionId) {
    var tabPanel = Ext.getCmp(containerId);
    if (!tabPanel) return;

    var tabObj = tabPanel.getItem(regionId);
    if (!tabObj) {
        var region = Ext.getCmp(regionId);
        if (!region) return;

        region.hide();
        return;
    }

    var z = tabPanel.hideTabStripItem(regionId);
    var idx = this._getNextActiveTab(tabPanel,regionId);
    
    MODx.hiddenTabs.push(regionId);
    tabPanel.setActiveTab(idx);                    
}
```

manager/assets/modext/core/modx.js
```diff
@@ -477,33 +499,49 @@ Ext.extend(MODx,Ext.Component,{
...
+        if (targetItem) {
+            targetItem.replaceWith(sourceItem);
+        } else {
+            var id = tvId.replace('tv', '');
+            console.warn('Attempted to move the TV named "'
+                + sourceItem.dom.innerText + '" (id #' + id + ') to the panel with the id "'
+                + target.id + '" but the original TV field could not be found. '
+                + 'It is likely a conflicting customization rule has already tried to move this TV.'
+            );
         }
```
theboxer, line 537:
```suggestion
        if (targetItem) {
            targetItem.replaceWith(sourceItem);
            return;
       }

        var id = tvId.replace('tv', '');
        console.warn('Attempted to move the TV named "'
            + sourceItem.dom.innerText + '" (id #' + id + ') to the panel with the id "'
            + target.id + '" but the original TV field could not be found. '
            + 'It is likely a conflicting customization rule has already tried to move this TV.'
        );       
```

theboxer (changes requested):
Hey, thank you for your contribution. This is just a quick code review, I'll be testing this PR later on.

smg6511:
@theboxer - Just a general note as I work on your feedback: The methods I updated were primarily done so with the main purpose in mind (changing the unnecessary array-formatted params to string params), as well as renaming params so they're immediately understandable and providing doc blocks for the methods I worked with. I did not really alter them too much otherwise. As you're pointing out, maybe some re-writing could be done, but that's not what I was looking to address with this PR.

I've found that a lot of reworking needs to be done overall with the element creation and form customization UIs that may not be immediately apparent. See my PR #15773 ... it'll give you a pretty good idea of what I'm referring to. These type of changes are too big to be worthwhile for 2.x and are targeted for 3.x.

theboxer:
@smg6511 Sorry, I kind of missed that this is not a PR for 3.x, priorly knowing that this is for 2.x I wouldn't care about the complexity :))

Feel free to ignore them all.

I'd love to see the overall code quality to improve for 3.x, that's why I was trying to make these suggestions - as it makes the code more readable and easier to follow.

smg6511:
@theboxer - No problem. I'm totally with you on upping the code quality! As I was playing with this I discovered how fragile the form customization is; once you have a conflicting rule (whether in one profile or across multiple ones) unexpected behavior creeps in. That'll be a bigger task for a different day ;-)

smg6511:
@theboxer - There's still a change requested flagged, but I can't see where it is. Not sure if I dismiss that (and if so, how) or you do...

manager/assets/modext/core/modx.js
```diff
@@ -477,33 +499,49 @@ Ext.extend(MODx,Ext.Component,{
...
-        if (!tvp) { return; }
+    /**
+     * Moves a TV to the specified a region or tab; method called from code inserted via modActionDom (modactiondom.class.php)
+     * and from MODX.hideTab (above)
+     *
+     * @param {String} tvId - Text id of the TV to move
+     * @param {String} targetId - Text id of the TV's destination (region/tab)
+     * @return {void}
+     */
+    ,moveTV: function(tvId, targetId) {
```
theboxer, line 510: I'm afraid that this change can't go to 2.x as it technically breaks backwards compatibility (changing the array arg to string).

manager/assets/modext/core/modx.js
```diff
@@ -515,45 +553,72 @@ Ext.extend(MODx,Ext.Component,{
...
+        if (containerId == 'modx-panel-resource' && fieldId.indexOf('modx-resource-content') != -1) {
+            container = Ext.getCmp('modx-resource-content');
+            if (container) {
+                if (container.setTitle) {
+                    container.setTitle(newLabel);
+                } else if (container.setLabel) {
+                    container.setLabel(fieldId, newLabel);
                 } else {
-                    cto.label.update(vals[0]);
+                    container.label.update(newLabel);
```
theboxer, line 575: Same as the previous, breaks backwards compatibility by changing array to string.

manager/assets/modext/widgets/core/modx.panel.js
```diff
@@ -265,15 +265,22 @@ Ext.extend(MODx.FormPanel,Ext.FormPanel,{
...
-            f.hide();
-            var d = f.getEl().up('.x-form-item');
-            if (d) { d.setDisplayed(false); }
+    /**
+     * Called exclusively from MODx.hideField (modx.js) for form customization
+     *
+     * @param {String} fieldId - Text id or name of field whose label is being hidden
+     * @return {void}
+     */
+    ,hideField: function(fieldId) {
```
theboxer, line 274: BC break as well

manager/assets/modext/widgets/core/modx.panel.js
```diff
@@ -290,20 +297,28 @@ Ext.extend(MODx.FormPanel,Ext.FormPanel,{
...
-                f.label.update(v);
-            }
+    /**
+     * Called exclusively from MODx.renameLabel (modx.js) for form customization
+     *
+     * @param {String} fieldId - Text id or name of field whose label is being renamed
+     * @param {String} newLabel - The replacement label text
+     * @return {void}
+     */
+    ,setLabel: function(fieldId, newLabel){
```
theboxer, line 307: BC break

theboxer (changes requested):
Unfortunately the signature changes are not possible for 2.x as they would break a backwards compatibility.

manager/assets/modext/core/modx.js
smg6511: Can you tell me how and at what version this would break it? Same with the others. I'd like to try to fix this issue for 2.x if possible, but it'll likely be unfixable if the FC rules were somehow saved in a serialized way at some point. Thx ;-)

theboxer: Well you changed the signature of the function from accepting array/string and string, to accept string and string. I know you changed all the usages within modx, but if there's an extra out there that's using the array & string signature it'll break. Which is not allowed when following semver.

smg6511: Let me ask you this: Would there be any reason for an extra to interact directly with the FC rules/logic? I might be short-sited on this, but I can't imagine an extra utilizing these methods in any way.

On the breaking change/semver point, I get that and am wholeheartedly for following common and strict protocols. For the sake of debate, is it worth _not_ fixing a bug to maintain protocol when a change technically breaks the protocol but has a slim-to-none practical chance of actually breaking functionality? Arguably, the particular issue this PR addresses is low priority as far as bugs go, but addressing it did lead to discovering some seemingly unnecessarily complicated code.

Anyway, regardless of whether we drop this PR for 2.x, should I assume that the BC we're debating will not be an issue if submitted for 3.x?

opengeek: Is there a specific reason this bug could not be resolved without changing the function signatures?

smg6511: Yes, the issue is that right now FC can't handle a comma in a field label because the current code reads that as a separator. I'm not sure there's a way around it _without_ changing the signature.

smg6511:
Closing this for 2.x. A port to 3.x has been merged.

PR #15839, opened 2021-10-15
Add policy count column to policy templates

### LINKED ISSUE #12492: Deleting a policy template silently removes associated policy
### Problem ###
When you (accidently) delete a policy template, which is still "in use" by a policy, there is no error message. Instead, the associated policies silently disappear form the "Access Policies" tab. They are also deleted from the database. However, the context permission items for a user group with that policy are still there then - but the saved id of the policy won't be found any more - you are left with *ghost entries* there.  

The deletion is irrecoverable and you are forced to rebuild from backup or scratch the permissions then. That's undesirable and should be avoided, especially when we consider the importance but yet complexity of setting-up a well permissioned installation. 

### Solution ###
MODX **should not** allow deleting a policy template which is still in use. 

Prevention of deletion seems already be implemented for AdministratorTemplate, ElementTemplate,ObjectTemplate and ResourceTemplate (these cannot be deleted via context menu, have not tested with bulk actions - I hope not!) Therefore it shouldn't be that hard to implement a "smart" prevention here.

### Enhancement 1 for better overview ###
The "Policy Templates" tab could have an extra column showing the number of policies which are using this template. 

### Enhancement 2 for better overview ###
The "Access Policies" tab could mark (bold text), which policy is actually in use.

theboxer:
### What does it do?
- Adds policy count column to the policy templates grid
- Adjusts delete/delete multiple confirm message, when the template is in use, it adds new bolded paragraph to inform about it

### Why is it needed?
- You can nuke policy template with policies attached to it without a prior warning
- More visibility on what templates are used

### How to test
- Open access policies and check the new column in the policy templates.

### Related issue(s)/PR(s)
Resolves #12492


Mark-H (approved):

alroniks (approved):
Finally! 🥳

Issue #15838, opened 2021-10-14
Improve input formatting options for number TVs


SnowCreative:
### Summary
Some more options for how numbers can be entered and displayed would be very useful, for example when using number fields to enter prices.

### Why is it needed?
You can turn on “Allow Decimals” and set the Decimal Precision to “2”, but if you then type in a number like “12.00”, the zeros and period get stripped and it changes to just “12”. And “12.20” gets changed to “12.2”. That’s not so great when a field is meant to contain prices. 

### Suggested solution(s)
Adding a setting to force the number of decimal places to be displayed would make numbers like prices consistent.
Also, a setting to right-align numbers inside the field would be useful, for example when you use a MIGX grid to list a table of prices in columns.

smg6511:
I've been doing a lot of work on the UI for TV creation and will take a look at this. Probably the best solution would be to create a new output option output type that would allow the specification of various currency formats.

SnowCreative:
More output options would be great, but that won't affect the input fields, which is where this would be most beneficial. Output can already be managed through scripting.

smg6511:
Yes, I see what you mean. I was more focused on the web page output end of it rather than the appearance within the manager. I agree it'd be distracting to see currency values without their full decimal places shown. I just submitted a PR that would satisfy your feature request.

SnowCreative:
Thanks, the PR works great!

smg6511:
Issue fixed in 2.x and 3.x

PR #15840, opened 2021-10-16
Add new config option for number TVs

smg6511:
### What does it do?
Added a new boolean parameter and behavior via and override to Ext.form.NumberField to allow decimal place zero padding for number TVs. Note that this should have no side effects for current installations, as the default value for this new parameter is false (which maintains the current behavior).

### Why is it needed?
Currently it is not possible with the number TV type to save or display a number such as 52.00 or 3.10. Trailing zeros will always be trimmed (yielding 52 and 3.1 respectively). While from a math/technical perspective these are the same value, it's not desirable to lose trailing zeros when working with currency values for example.

### How to test
Clear your caches and remember to run `grunt build` within the __build/templates/default_ directory to compile the changes into the main js file. Create a few TVs with different setting for Allow Decimals, Decimal Precision, and the new parameter Strict Decimal Precision. Create a resource with access to these test TVs and play with saving different values. Observe that these values persist in the editing form and display as expected on the rendered web page.

### Related issue(s)/PR(s)
Provides the requested feature in #15838.


theboxer:
Why would you care about keeping the decimal precision on the input side? You can force the decimal precision on the output by using an output modifier, or maybe a custom output type for the TV.

smg6511:
@theboxer - Yes, that's how I initially responded to the feature request, but it makes sense that you would want to see a currency value with its full decimal precision. It would be inconsistent and understandably irritating to see, for example, $12 or $4.5 when you're expecting to see $12.00 and $4.50 in the manager interface.

Mark-H:
Wouldn't this be better in 3.x considering we're not planning a 2.9 at this point?

smg6511:
I'd say the change is pretty small, thus easily incorporated now.  But I'll submit the change for 3.x as well.

smg6511:
Closing this for 2.x. A port to 3.x has been merged.

smg6511:
Sorry for the confusion ... was meaning to close a different PR for 2.x.

Issue #15844, opened 2021-10-19
PR #15822 Breaks Template Variables


smg6511:
### Summary
The new template preview feature introduced in PR #15822 mistakenly adds the _preview_file_ field to _modTemplateVar_ and removes it from _modTemplate_. 

### Step to reproduce
This PR is only a few hours old, so you'll need to test this against a build of the most recent commit for 3.0.0-alpha3.

### Observed behavior
Try to create a new TV and try to save a preview image in a template. You'll find that neither operation succeeds.

### Related issue(s)/PR(s)
PR #15822.

opengeek:
I'm confused. #15822 simply removed a field. It does not add one.

opengeek:
It looks to me like someone did not add those fields to the correct place in the schema file and did not make those changes to the platform class by running the parse-schema script. We should not be manually changing the platform files without running parse-schema afterward. This is why the whitespace changes from IDE settings were saved to the platform class in that PR, which had to correct by running parse-schema. By doing this, it exposed the problem with this field being improperly defined in the schema from somewhere else. This has nothing to do with #15822—it is just a coincidence that it was exposed by the merge process on that PR.

PR #15847, opened 2021-10-20
Fix preview_file field added to TVs instead of templates, causing iss…

Mark-H:
…ues saving either [#15844]

for @opengeek  - if still open in the morning i'll add more info, now i'm gonna crash

### What does it do?

### Why is it needed?

### How to test

### Related issue(s)/PR(s)

### DIFF
--- a/core/model/schema/modx.mysql.schema.xml
+++ b/core/model/schema/modx.mysql.schema.xml
@@ -1070,6 +1070,7 @@
         <field key="properties" dbtype="text" phptype="array" null="true" />
         <field key="static" dbtype="tinyint" precision="1" attributes="unsigned" phptype="boolean" null="false" default="0" index="index" />
         <field key="static_file" dbtype="varchar" precision="255" phptype="string" null="false" default="" />
+        <field key="preview_file" dbtype="varchar" precision="191" phptype="string" null="false" default="" />
 
         <index alias="templatename" name="templatename" primary="false" unique="true" type="BTREE">
             <column key="templatename" length="" collation="A" null="false" />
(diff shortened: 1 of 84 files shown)


opengeek:
Thanks for fixing the schema up. Let's figure out why your parse-schema run is adding an extra empty line in those files before the class declaration before merging this.

core/model/schema/modx.mysql.schema.xml
```diff
@@ -1070,6 +1070,7 @@
         <field key="properties" dbtype="text" phptype="array" null="true" />
         <field key="static" dbtype="tinyint" precision="1" attributes="unsigned" phptype="boolean" null="false" default="0" index="index" />
         <field key="static_file" dbtype="varchar" precision="255" phptype="string" null="false" default="" />
+        <field key="preview_file" dbtype="varchar" precision="191" phptype="string" null="false" default="" />
```
JoshuaLuckers, line 1073: Shouldn't this have the same precision as `static_file` (or vice versa)?

Mark-H: Hmm... that's a good question.

For new fields, using 191 is a safe bet (and my personal default preference) cause it fits within the restraints of a utf8mb4 innodb index, even if this particular field is not indexed. In reality this could probably have a precision of 100 and work fine, too. 

But we probably don't want to lower the precision for existing non-indexed fields from one fairly arbitrary number to another?

If we value consistency the most, then making preview_file use 255 would indeed be better.

opengeek: Let's take this to an issue if it needs to be one.

opengeek (approved):

PR #15859, opened 2021-10-23
Upgrade Smarty to v4 for PHP 8 support

Mark-H:
### What does it do?

Swaps out v3 with [the new v4 pre-release](https://github.com/smarty-php/smarty/blob/master/CHANGELOG.md#400-rc0---2021-10-13), and makes the necessary adjustments. 

### Why is it needed?

Smarty v3 does not claim to support PHP 8. I've not personally seen any issues with it other than the odd notices, but this seems like a good time to upgrade MODX3. Hopefully while we stabilise MODX3, Smarty will also reach a stable v4. 

It may also be a good idea to backport this to 2.x, however the minimum PHP version in Smarty 4 is PHP 7.1. So I don't think we can. 

### How to test
`composer update` and use the manager. 

### Related issue(s)/PR(s)
This was brought up in the #core-contributors channel on Slack by Ivan.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "league/flysystem": "^2.0",
         "league/flysystem-aws-s3-v3": "^2.0",
         "phpmailer/phpmailer": "^6.0",
-        "smarty/smarty": "^3.1",
+        "smarty/smarty": "^4.0",
         "james-heinrich/phpthumb": "^1.7",
         "erusev/parsedown": "^1.7",
         "inlinestyle/inlinestyle": "^1.2",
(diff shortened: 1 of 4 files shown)


Ibochkarev (approved):

Ibochkarev (approved):
No errors were found when testing this PR. Tested creation/editing (resources, chunks, snippets), output (chunks, templates, TV), changing settings, uploading files.

JoshuaLuckers (approved):

opengeek (approved):

PR #15865, opened 2021-10-25
Skip creation of thumbnails for SVGs

### LINKED ISSUE #13556: Svg support for image TV preview
### Summary
The image TV preview does not support svg.

### Step to reproduce
Add "svg" to the system setting "upload_images"  and your media source's "imageExtensions" setting.
Upload and choose a svg in your image TV. 

### Observed behavior
The image TV preview does not show (a preview of) the svg.

bildschirmfoto 2017-07-18 um 19 09 41

### Expected behavior
The image TV preview should show (a preview of) the svg.

### Environment
MODX 2.5.8-dev
Apache/2.2.29 (Unix)
PHP 7.0.8

verstratenbram:
### What does it do?
Skips the creation of thumbnails for SVG files in the image tv preview and sets max-width and max-height to 400px in the stylesheet.

### Why is it needed?
PhpThumb is not able to convert SVG to PNG resulting in a broken thumbnail in the image tv preview.

### How to test
Create a resource with an image TV.
Upload an SVG image.

### Related issue(s)/PR(s)
Resolves #13556


verstratenbram:
@cla-bot check

Ruslan-Aleev:
Thanks for the PR!
The only thing is, you don't need to edit the compiled `index.css` and `index-min.css` files

verstratenbram:
@Ruslan-Aleev Thank you for pointing this out. The changes to these files have been discarded.

core/model/modx/processors/system/phpthumb.class.php
```diff
@@ -55,6 +55,13 @@ public function process() {
         $src = $this->source->prepareSrcForThumb($src);
         if (empty($src)) return '';
 
+        if (strtolower(substr(strrchr($src, '.'), 1)) === 'svg') {
```
mishanthrop: You don't need to reinvent the wheel: https://www.php.net/manual/en/function.pathinfo.php

verstratenbram:
Thank you for your review @mishantrop. Using `pathinfo` now.

core/model/modx/processors/system/phpthumb.class.php
```diff
@@ -55,6 +55,13 @@ public function process() {
         $src = $this->source->prepareSrcForThumb($src);
         if (empty($src)) return '';
 
+        if (strtolower(pathinfo($src, PATHINFO_EXTENSION)) === 'svg') {
+            /* Skip thumbnail generation for svg and output the file directly */
+            header('Content-Type: image/svg+xml');
+            echo @file_get_contents($src);
```
mishanthrop: Why did it become necessary to silence errors? Can these errors occur?

verstratenbram: > An E_WARNING level error is generated if filename cannot be found, length is less than zero, or if seeking to the specified offset in the stream fails.

I imagine this can happen when using an URL instead of a file from the media browser, or when the local file get deleted. Usually I silence these warnings because the result needs to be validated anyway.

In this case the suppression can be removed because warnings are already handled by the manager.

opengeek (approved):

rthrash:

https://community.modx.com/t/tv-type-image-no-preview-showing-when-source-is-svg-image/5275/2

PR #15871, opened 2021-10-27
Fix accessing $modx->controller from plugins and services

Mark-H:
Please see #core-contributors on slack to discuss this as I'm _super confused_ why my extras are breaking in alpha3 without this. 

### What does it do?

Changes controller initialisation to immediately write to $modx->controller instead of a local variable that is then written to $modx->controller later. 

### Why is it needed?

While testing the alpha3 nightly from last night I kept running into inexplicable errors related to $modx->controller. 

Something as simple as this in a plugin:

    case 'OnManagerPageBeforeRender':
        $assetsUrl = $modx->getOption('commerce.assets_url', null, $modx->getOption('assets_url') . 'components/commerce/');
        $modx->controller->addCss($assetsUrl . 'tvs/products.css');

... would cause a fatal error `Call to undefined method addCss on null`, and a similar line of `$modx->controller->addLexiconTopic(..)` in an OnRichTextEditorInit event would also fail with a fatal error.

I wish I understood better _why_, but this change resolves those problems. 

There may be a relation to #14927 and modExtraManagerController. Previously those would be instantiated with a `getInstanceDeprecated()` method, but that was changed in #14927 to use the generic `getInstance()` method. Another reason to think it may be related to the modExtraManagerController somehow is that I did not see the Redactor error while editing a resource - only when it got loaded on another CMP it seemed to break. 

### How to test

It happened on an alpha3 nightly from yesterday using Redactor and Commerce with both triggering 

### Related issue(s)/PR(s)
N/a

### DIFF
--- a/core/src/Revolution/modManagerResponse.php
+++ b/core/src/Revolution/modManagerResponse.php
@@ -98,26 +98,24 @@ class modManagerResponse extends modResponse
             $className = $this->getControllerClassName($this->route);
 
             /** @var modManagerController $controller */
-            $controller = $className::getInstance($this->modx, $className, [
+            $this->modx->controller = $className::getInstance($this->modx, $className, [
                 'namespace' => $this->namespace,
                 'namespace_path' => $this->namespacePath,
                 'action' => $this->route,
                 'controller' => $this->route,
             ]);
-            $controller->setProperties(array_merge($_GET,$_POST));
-            $controller->initialize();
+            $this->modx->controller->setProperties(array_merge($_GET,$_POST));
+            $this->modx->controller->initialize();
 
-            if (!$controller->checkPermissions()) {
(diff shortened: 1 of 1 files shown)


Mark-H:
Thanks @theboxer for the help.

Turns out it's a rather specific combination of a controller (in this case Commerce's CMP) invoking the rich text editor (in this case Redactor) from its initialize() method. Previously in 2.x $modx->controller would get immediately set when the instance was created, but in 3.x that changed to after permissions and such were checked. And **after** the initialize method() got called. So the editor would get called before $modx->controller was filled and break. 

This specific combination is probably uncommon and just not tested until now.

theboxer (approved):

opengeek (approved):

Issue #15877, opened 2021-10-30
There is no description for the upload_check_exists system setup


Ibochkarev:
### Summary
There is no description for the upload_check_exists system setup

### Step to reproduce
Open the Settings section, File System section

### Observed behavior
![изображение](https://user-images.githubusercontent.com/2138260/139533544-7ac5d4b1-ba0b-42cd-ab7c-f58cc6b3d611.png)

### Expected behavior
There must be a description

### Related issue(s)/PR(s)
NA

### Environment
MODX 3 alpha

Ruslan-Aleev:
I have no `upload_check_exist` key on my MODX 3 installed.
If I'm not mistaken, the check for the existence of a file in MODX 3 occurs through the `upload_file_exist` setting, but in MODX 2 it is `upload_check_exist`. This function was done in parallel, and therefore the keys have different names.
Apparently the key appeared when updating MODX.

Ibochkarev:
> I have no `upload_check_exist` key on my MODX 3 installed. If I'm not mistaken, the check for the existence of a file in MODX 3 occurs through the `upload_file_exist` setting, but in MODX 2 it is `upload_check_exist`. This function was done in parallel, and therefore the keys have different names. Apparently the key appeared when updating MODX.

That's right, I updated the site from 2.8.3 to 3.0 alpha 3

Ibochkarev:
![изображение](https://user-images.githubusercontent.com/2138260/139533601-69d56fd0-3eaf-457b-9f9b-b530b0373fed.png) 

Same problem

JoshuaLuckers:
I guess we need to update the upgrade script that cleans up the system settings:
https://github.com/modxcms/revolution/blob/bd5d09137f1a66c995f488e0257c680f0ea6ceea/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php

PR #15881, opened 2021-10-31
Removed unused system settings

Ibochkarev:
### What does it do?
Removed unused system settings, which still appear after upgrading from 2.x to 3.x.

### Why is it needed?
These settings are outdated or found only in the "System Settings", do not participate in the remaining code:

- `upload_check_exist`
- `websignupemail_message`

### Related issue(s)/PR(s)
#15877

### DIFF
--- a/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
+++ b/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
@@ -22,13 +22,17 @@ $settings = [
     'manager_js_zlib_output_compression',
     'rb_base_dir',
     'rb_base_url',
+    'resource_static_allow_absolute',
+    'resource_static_path',
     'resolve_hostnames',
     'server_protocol',
     'strip_image_paths',
     'udperms_allowroot',
+    'upload_check_exist',
     'upload_flash',
     'use_browser',
     'webpwdreminder_message',
+    'websignupemail_message',
 ];
 
 $messageTemplate = '<p class="%s">%s</p>';
--- a/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
+++ b/setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
@@ -22,8 +22,6 @@ $settings = [
     'manager_js_zlib_output_compression',
     'rb_base_dir',
     'rb_base_url',
-    'resource_static_allow_absolute',
-    'resource_static_path',
     'resolve_hostnames',
     'server_protocol',
     'strip_image_paths',


alroniks:
I am not sure about the two first. There were some security fixes related to it, isn't it @Mark-H?

Ibochkarev:
> I am not sure about the two first. There were some security fixes related to it, isn't it @Mark-H?

https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/modStaticResource.php#L63

Indeed, the setting is in the code and is used. So you need to add lexicons for the name and description of the system setting

JoshuaLuckers:
Hmm my mistake about those resource settings. I didn't see them in my setup 😅

setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
```diff
@@ -22,13 +22,17 @@
     'manager_js_zlib_output_compression',
     'rb_base_dir',
     'rb_base_url',
+    'resource_static_allow_absolute',
+    'resource_static_path',
     'resolve_hostnames',
```
opengeek: Need these taken out before merging.

opengeek (changes requested):

setup/includes/upgrades/common/3.0.0-cleanup-system-settings.php
Ibochkarev: Done

Mark-H:
Ohhh I completely forgot about porting that security fix to 3.x. The idea was to rewrite static resources a bit more extensively for 3.0 so they could actually use media sources. But at the very least we ought to get the fix in from 2.x which indeed uses those two settings.

alroniks (approved):

Ruslan-Aleev:
If the user has a non-default value in the `upload_check_exist` setting in the 2.x branch, then when updating, we need to set the same value for `upload_file_exist`. Or, alternatively, make the key names the same.

JoshuaLuckers:
> Ohhh I completely forgot about porting that security fix to 3.x. The idea was to rewrite static resources a bit more extensively for 3.0 so they could actually use media sources. But at the very least we ought to get the fix in from 2.x which indeed uses those two settings.

@Mark-H Does this mean we need to port something from 2.x to 3.x and keep these system settings?

Mark-H:
Yes that's what it means.

rthrash:
Did the port happen @Mark-H or @JoshuaLuckers ? If so can we close this issue or maybe we close it and open a task to port it (unsure of which issue or I'd link it here).

JoshuaLuckers:
@rthrash not that I'm aware of

PR #15885, opened 2021-11-03
Restrict static resources to predefined path (port of #15656 for 3.x, SEC-3801)

Mark-H:
WIP

### What does it do?

### Why is it needed?

### How to test

### Related issue(s)/PR(s)

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1705,7 +1705,7 @@ $settings['static_elements_basepath']->fromArray([
   'area' => 'static_elements',
   'editedon' => null,
 ], '', true, true);
-$settings['resource_static_allow_absolute']= $xpdo->newObject('modSystemSetting');
+$settings['resource_static_allow_absolute']= $xpdo->newObject(modSystemSetting::class);
 $settings['resource_static_allow_absolute']->fromArray([
     'key' => 'resource_static_allow_absolute',
     'value' => '0',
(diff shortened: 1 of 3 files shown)


Mark-H:
If someone else has time to pick this up over the coming days, please do. I'm otherwise engaged and may not have the time, but it'd be good to get this included in the beta early next week.

muzzwood:
I'm looking in to this at the moment. Though, *so far* all is working as expected without me changing anything.

muzzwood:
I've tested this with the various different content types.

- If `resource_static_allow_absolute` is disabled. Requests always redirect to the `error_page` resource unless within the path specified by the `resource_static_path` setting.
- If `resource_static_allow_absolute` is enabled, the static resource loads the file contents correctly even if outside the path set by `resource_static_path`.
- Full absolute paths work on the resource if `resource_static_allow_absolute` is enabled.
- Full absolute paths work on the resource even if  `resource_static_allow_absolute` is disabled, as long as the path is the one that's set in `resource_static_path`. **NOTE: I'm not *entirely* sure if this is the intended behaviour, but I believe it to be.**
- Working as expected with or without a template applied.
- Tested with different filesystem media sources; no issues found.

opengeek:
@Mark-H — this seems like it is working as expected. Can you verify and go ahead and mark it ready for review if so?

Mark-H:
I've not personally done as extensive testing as @muzzwood but as it's practically copy/pasted from 2.x (which was tested a fair bit!), plus I trust @muzzwood's judgement, go for it. ;)

opengeek (approved):

PR #15888, opened 2021-11-04
Porting adding permissions to enforce access to specific resource types

alroniks:
### What does it do?
It brings fixes of security issues which were applied to 2.x branch.

As the code is different, I applied some refactoring to the original source. Most valuable changes:
- in `_build/data/transport.core.accesspolicies.php`  permissions sorted alphabetically in the touched arrays
- `ActionAccessTrait` trait added to combine common for different classed code. Compositional approach
- `modStaticResource`, `modWebLink` and `modSymLink` updated a bit, there are removed not used parts of code, constructors touched mostly.
- In controllers class aliases are added instead of having two classes in one file (it breaks PSR-4). And extending into an empty class is useless.

### Why is it needed?
To make MODX 3 more secure.

### How to test

First, rebuild and run the setup to add the new permissions.

- On a user with these permissions, confirm the different resource types can be interacted with as usual (create, edit, delete).
- On a user without these permissions, confirm you're presented with a friendly permission denied error.

### Related issue(s)/PR(s)
#15655


opengeek (changes requested):
This introduces a significant number of changes that are not directly related to the port of the security issue. I would prefer we not introduce such changes in a single PR. It also breaks the signature of modParsedManagerController.

alroniks:
To port just changes also it not possible, due to too different code. Issues and signatures I will check and fix, but other things make sense to review and apply. They are cosmetic around the code itself mostly, even if it looks like a big diff.

opengeek:
> To port just changes also it not possible, due to too different code. Issues and signatures I will check and fix, but other things make sense to review and apply. They are cosmetic around the code itself mostly, even if it looks like a big diff.

I re-reviewed the changes, and agreed. If we can fix the `Declaration of MODX\Revolution\modParsedManagerController::render() must be compatible with MODX\Revolution\modManagerController::render()` error and the code quality failures, it's good to go.

alroniks:
I fixed the code and now automatic checks are passed. I also tested the code and it seems that it works. But still, some review is needed, if something should be improved.

Mark-H (approved):
I've done a fair bit of testing with this and all seems in order. The only minor point I came across is detailed in https://github.com/modxcms/revolution/issues/15893 which I'd be happy to see addressed later. Thanks for porting this over!

theboxer (approved):

opengeek (approved):

PR #15889, opened 2021-11-05
[Port to 3.x] Form customization front end fixes/improvements

### LINKED ISSUE #15357: A label is cut off after applying form customization rule

smg6511:
### What does it do?
Reworks the methods responsible for moving and changing field/tab visibility and labelling. Code optimizations were made and method documentation was added. Note this is a port of PR #15836, which will ultimately be withdrawn as it creates breaking changes not desirable for 2.x.

### Why is it needed?
Most of these methods were written to receive/process serialized/array data, but there appears to be no reason (any longer at least) to keep them that way—especially given that problems have arisen such as the one referenced by this PR. The changes allow for the use of commas in labels and provides simplification and clarification of the relevant code.

### How to test
Clear your caches and remember to run `grunt build` within the `_build/templates/default` directory to compile the changes into the main js file. Create one or two test TVs, then create a Form Customization profile with at least one rule. Within your rule(s), experiment with re-labelling and hiding various resource fields along with your test TVs. Also create a new tab in Regions tab and verify that assigning your TVs to that tab (or the standard regions) works as expected.

### Related issue(s)/PR(s)
Resolves #15357.


core/src/Revolution/modActionDom.php
```diff
@@ -32,10 +32,10 @@ class modActionDom extends modAccessibleSimpleObject
      */
     public function apply($objId = '')
     {
+        /*
```
Mark-H: Removing commented-out code would be preferred here.

smg6511: Can't believe I forgot to delete that! Doing so now...

manager/assets/modext/core/modx.js
```diff
@@ -478,33 +486,49 @@ Ext.extend(MODx,Ext.Component,{
...
+        if (tabPanel) {
+            const tabObj = tabPanel.getItem(regionId);
             if (tabObj) {
-                var z = tp.hideTabStripItem(tab);
-                MODx.hiddenTabs.push(tab);
-                var idx = this._getNextActiveTab(tp,tab);
-                tp.setActiveTab(idx);
+                // const z = tabPanel.hideTabStripItem(regionId),
+                //     idx = this._getNextActiveTab(tabPanel,regionId)
+                // ;
```
JoshuaLuckers:
```suggestion
```

Removing code is preferred

manager/assets/modext/core/modx.js
```diff
@@ -478,33 +486,49 @@ Ext.extend(MODx,Ext.Component,{
...
+                tabPanel.hideTabStripItem(regionId);
+                MODx.hiddenTabs.push(regionId);
+                // tabPanel.setActiveTab(idx);
```
JoshuaLuckers:
```suggestion
```

Removing code is preferred

Ibochkarev (commented):
In my test, the customization of forms does not work at all, neither moving through areas, nor changing TV, nor setting the default value.

What am I doing wrong?

https://user-images.githubusercontent.com/2138260/142778757-1960ed35-bc53-42e0-aa93-ff834c4b5351.mp4

smg6511:
@Ibochkarev - The likely solution is to clear your browser cache. I had the same thing happen when moving over to this branch to check out what the problem could be: My browser was using a cached version of the js files.

manager/assets/modext/core/modx.js
smg6511: Fixed.

manager/assets/modext/core/modx.js
smg6511: Fixed.

theboxer (approved):

Issue #15223, opened 2020-09-02
The form for creating a lexicon entry does not work correctly


Ruslan-Aleev:
### Summary
You can create an lexicon entry without set either "Namespace", "Topic", or "Language", because the dropdowns are editable.
The entry will be created visually, but then it cannot be found in the search.

**We need to make these lists required.**
By the way, this form has no required fields at all, which is strange.

![lexicon-bug](https://user-images.githubusercontent.com/12523676/91947140-2553ed80-ed08-11ea-884f-7c0337d7a4d0.gif)

Perhaps there are such bugs in other creation windows too, we need to check.

### Environment
MODX 2.x >

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/logic-behind-structure-of-processors/4583/1

Ibochkarev:
Fixed in #15892

PR #15892, opened 2021-11-06
[3.x] Fix problems in lexicon creation form

smg6511:
### What does it do?
Made changes to lexicon processors to add simple validation logic and changed js form to make certain fields required.

### Why is it needed?
As there is currently no validation of lexicon form values, records without all the required data save to the database but cannot display because of missing information.

### How to test
Test this under two conditions: 
1. Disable the allowBlank properties I added to the js form; in this way you can verify that the basic php validation works when trying to create a lexicon entry with one or more blank values.
2. Re-enable the allowBlank properties and to verify that the front end validation stops the form from submitting when required values are missing.

Note: Now only valid entries will be saved, so they will always be found in the grid when using the filtering toolbar.

### Related issue(s)/PR(s)
Addresses issue #15223.

### DIFF
--- a/core/src/Revolution/Processors/Workspace/Lexicon/Create.php
+++ b/core/src/Revolution/Processors/Workspace/Lexicon/Create.php
@@ -43,14 +43,16 @@ class Create extends Processor
      */
     public function process()
     {
-        if ($this->alreadyExists()) {
+        $data = array_map('trim', $this->getProperties());
+
+        if ($this->alreadyExists($data)) {
             return $this->failure($this->modx->lexicon('entry_err_ae'));
         }
 
         $this->entry = $this->modx->newObject(modLexiconEntry::class);
-        $this->entry->fromArray($this->getProperties());
+        $this->entry->fromArray($data);
         $this->entry->set('editedon', date('Y-m-d h:i:s'));
-
+        
         if ($this->entry->save() === false) {
             return $this->failure($this->modx->lexicon('entry_err_save'));
         }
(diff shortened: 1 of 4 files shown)


core/src/Revolution/Processors/Workspace/Lexicon/Create.php
```diff
@@ -43,12 +44,14 @@ public function getLanguageTopics()
      */
     public function process()
     {
-        if ($this->alreadyExists()) {
+        $data = array_map('trim', $this->getProperties());
```
Mark-H: This may cause issues as it trims values as well. Some lexicons end with a space by design, because they're suffixed (in code) by a value. Some quick examples I found:

```
$_lang['file_err_open'] = 'Cannot open file: ';
$_lang['package_update_to_version'] = 'Please select the version to update this package to: ';
$_lang['error_sending_email_to'] = 'Error while sending mail to ';
```

I'd suggest removing the array_map here and specifically trimming the namespace, topic, and language if needed in alreadyExists/fromArray.

Mark-H (commented):
Tested and seems to work, but it may be worth fixing up the array_map trim.

core/src/Revolution/Processors/Workspace/Lexicon/Create.php
smg6511: OK, yes, I didn't think of that! I'll make the update later today...

mishanthrop: @Mark-H it makes sense to change lexicon items in future:
```php
$_lang['file_err_open'] = 'Cannot open file: ';
```
to
```php
$_lang['file_err_open'] = 'Cannot open file: [[+file]]';
```

Respectively:
```php
$this->modx->lexicon('file_err_open') . $this->getProperty('file');
```
to

```php
$this->modx->lexicon('file_err_open', [ 'file' => $this->getProperty('file') ]);
```

This will allow not to worry about such dangling lexicon items.
Concatenation is bad practice when it comes to localization.

Mark-H: I don't disagree, but it's probably better to leave that for another day ;)

Mark-H (approved):

opengeek (approved):

Issue #15740, opened 2021-06-17
Use strong random number/string generators inside generatePassword()


patrickatwsrn:
### Summary

Looking for a  native solution to create random strings, I stumbled upon the generate Password function today.

I've noticed that the "randomness" of the generated Password relies on mt_rand(), a "Pseudo-Random Number Generator". I guess this is a pice of code that has been active from the old php5 days. PRNG are considered to be a "weak" nowadays.

Basing a random number on a timestamp using srand() is also considered to be a potential problem.

```
    source: core/model/modx/moduser.class.php, line 797
    /**
     * Returns a randomly generated password
     *
     * @param integer $length The length of the password
     * @param array $options
     * @return string The newly generated password
     */
    public function generatePassword($length = null,array $options = array()) {
        if ($length === null) {
            $length = $this->xpdo->getOption('password_generated_length', null, 10, true);
        }
        $passwordMinimumLength = $this->xpdo->getOption('password_min_length', null, 8, true);
        if ($length < $passwordMinimumLength) {
            $length = $passwordMinimumLength;
        }
        $options = array_merge(array(
            'allowable_characters' => 'abcdefghjkmnpqrstuvxyzABCDEFGHJKLMNPQRSTUVWXYZ23456789',
            'srand_seed_multiplier' => 1000000,
        ),$options);

        $ps_len = strlen($options['allowable_characters']);
        srand((double) microtime() * $options['srand_seed_multiplier']);
        $pass = '';
        for ($i = 0; $i < $length; $i++) {
            $pass .= $options['allowable_characters'][mt_rand(0, $ps_len -1)];
        }
        return $pass;
    }
```

### Why is it needed?

Random passwords should be random by todays standards. 

I'm aware that its very unlikely that the login mechanism might be compromised by this "vulnerabilty", but I like to point out that a modx developer might make use of this function to generate a "secure" and "random" strings for other purposes.

### Suggested solution(s)

In php 7 "Cryptographically Secure Pseudo-Random  NUMBER/STRING Generators" are available. These methods are considered to be "strong":

* https://www.php.net/manual/de/function.random-bytes.php (php >= 7.0)
* https://www.php.net/manual/en/function.random-int.php (php >= 7.0)
* https://www.php.net/manual/de/function.openssl-random-pseudo-bytes.php (php >= 5.3.0)

PHP 5.x polyfill for random_bytes() and random_int() 
* https://github.com/paragonie/random_compat

SIDENOTE:
[...]

PR #15894, opened 2021-11-08
Change generatePassword() logic

crystaldaking:
### What does it do?
Updated the password generation function using a truly random string

### Why is it needed?
I am deeply convinced that the more a password looks like a hash, the less it attracts the attention of hackers. Moreover, this implementation uses a truly random string in contrast to the stock implementation

The `options` parameter as part of password generation is a rather useless parameter. It is not safe to specify salt, any other parameters simply do not occur to me. As an option, I left the implementation of the classic `alphabet` password that can be invoked using options

### How to test
Standard tests work, no logic is affected.
You can also test password generation in the user control panel

### Related issue(s)/PR(s)
#15740

### DIFF
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -904,19 +904,14 @@ class modUser extends modPrincipal
         if ($length < $passwordMinimumLength) {
             $length = $passwordMinimumLength;
         }
-        $options = array_merge([
-            'allowable_characters' => 'abcdefghjkmnpqrstuvxyzABCDEFGHJKLMNPQRSTUVWXYZ23456789',
-            'srand_seed_multiplier' => 1000000,
-        ], $options);
 
-        $ps_len = strlen($options['allowable_characters']);
-        srand((double)microtime() * $options['srand_seed_multiplier']);
-        $pass = '';
-        for ($i = 0; $i < $length; $i++) {
-            $pass .= $options['allowable_characters'][mt_rand(0, $ps_len - 1)];
+        if ($options['alphabet']) {
+            $alphabet = array_merge(range('a', 'z'), range('A', 'Z'));
+            shuffle($alphabet);
+            return substr(implode($alphabet),0,$length);
         }
 
-        return $pass;
+        return bin2hex(random_bytes($length));
     }
 
 
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -898,7 +898,7 @@ class modUser extends modPrincipal
     public function generatePassword($length = null, array $options = [])
     {
         if ($length === null) {
-            $length = $this->xpdo->getOption('password_generated_length', null, 10, true);
+                $length = $this->xpdo->getOption('password_generated_length', null, 10, true);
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):
Before:

изображение


After:

изображение

JoshuaLuckers:
@Mark-H I would like your opinion on this change.

core/src/Revolution/modUser.php
```diff
@@ -898,25 +898,20 @@ public function removeLocks(array $options = [])
...
-        $ps_len = strlen($options['allowable_characters']);
-        srand((double)microtime() * $options['srand_seed_multiplier']);
-        $pass = '';
-        for ($i = 0; $i < $length; $i++) {
-            $pass .= $options['allowable_characters'][mt_rand(0, $ps_len - 1)];
+
+        if ($options['alphabet']) {
+            $alphabet = array_merge(range('a', 'z'), range('A', 'Z'));
+            shuffle($alphabet);
+            return substr(implode($alphabet),0,$length);
```
Ibochkarev, line 911:
```suggestion
            return substr(implode($alphabet), 0, $length);
```

core/src/Revolution/modUser.php
```diff
@@ -898,25 +898,20 @@ public function removeLocks(array $options = [])
...
         }
 
-        return $pass;
+        return substr(bin2hex(random_bytes($length)),$length);
```
Ibochkarev, line 914:
```suggestion
        return substr(bin2hex(random_bytes($length)), $length);
```

Ibochkarev (changes requested):

Mark-H:
Other than the phpcs issues I don't have a strong opinion on this - the use of `bin2hex(random_bytes())` is valid for this type of random generation. Ideally people don't keep using these random passwords anyway. 

The `modUserTest::testGeneratePassword` tests do need to be updated it seems.

Mark-H (changes requested):
Tests need to be updated to match new implementation.

JoshuaLuckers:
@crystaldaking would it be possible for you to update the Unit Tests to match your implementation?

core/src/Revolution/modUser.php
```diff
@@ -898,25 +898,20 @@ public function removeLocks(array $options = [])
...
-            'srand_seed_multiplier' => 1000000,
-        ], $options);
-
-        $ps_len = strlen($options['allowable_characters']);
-        srand((double)microtime() * $options['srand_seed_multiplier']);
-        $pass = '';
-        for ($i = 0; $i < $length; $i++) {
-            $pass .= $options['allowable_characters'][mt_rand(0, $ps_len - 1)];
+
+        if ($options['alphabet']) {
```
Mark-H, line 908: This may also require a check as I don't see the `alphabet` key being required elsewhere. So something like `if (!empty($options['alphabet'])) {` to avoid php warnings.

Ibochkarev:
@crystaldaking ping...

Mark-H:
As this PR has gone stale, I recreated it at #16521.

PR #15895, opened 2021-11-08
Add ability to return to the same page after language switching

alroniks:
### What does it do?
It fixes the issue when after switching the language user always returns to the main manager page. Now it respects the previous page (via referer header) and sends a redirect to that page if it exists. Otherwise fallback to the manager base URL. 

### Why is it needed?
Improve user experience during manager usage.

### How to test
Apply changes, go to any page inside the manager, and try to switch the language. You have to stay on the same page after language switching.

### Related issue(s)/PR(s)
#15174
#14046

### DIFF
--- a/core/src/Revolution/modLexicon.php
+++ b/core/src/Revolution/modLexicon.php
@@ -461,7 +461,7 @@ class modLexicon
      *
      * @return array An array of available languages
      */
-    public function getLanguageList($namespace = 'core')
+    public function getLanguageList(string $namespace = 'core'): array
     {
         $corePath = $this->getNamespacePath($namespace);
         $lexPath = str_replace('//', '/', $corePath . '/lexicon/');
--- a/core/src/Revolution/modMenu.php
+++ b/core/src/Revolution/modMenu.php
@@ -84,13 +84,13 @@ class modMenu extends modAccessibleObject
     }
 
     /**
-     * Returns list of available languages in the system with descriptions ans translated names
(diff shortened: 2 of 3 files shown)


Ruslan-Aleev (approved):
Works as it should :)

lang_switcher

Mark-H (changes requested):
This looks like it introduces an [open redirect vulnerability](https://sec.okta.com/articles/2021/02/why-it-so-hard-prevent-open-redirects); the referrer will need to be sanitized or forced to a relative url if we want to use this.

I'd personally feel best if the redirect created the redirect URL itself from a provided &target and &target_namespace and any other properties passed, i.e. something like:

```
$params = $scriptProperties;
$params['a'] = $params['target'] ?? 'welcome';
$params['namespace'] = $params['target_namespace'] ?? 'core';
unset($params['target'], $params['target_namespace']);
$url = MODX_MANAGER_URL . '?' . http_build_query($params);
```

JoshuaLuckers:
@alroniks do you have time to address the changes @Mark-H requested?

Ruslan-Aleev:
Deterioration in UX:
If you change the language to another and exit the manager panel, then the login form will display your system language (and not the one selected before exiting the panel), i.e. now you need to select the language again.

Ruslan-Aleev:
It has been fixed and merged into https://github.com/modxcms/revolution/pull/16069 I will close this PR.

PR #15898, opened 2021-11-10
Improve snippets logging in case error messages with arrays

intersel:
### What does it do?
Fix bad log from modsnippet when there are errors when saving or removing snippet

### Why is it needed?
Better to have a correct log

### How to test
Remove the 'write' rights  of a snippet static file and try to save the snippet in the manager. It should try to log the error. The logged error shows the "snippet_err_save" message with "Array" written after... after fix, will show the content of the array of the snippet ($this->toArray() ) as it was designed...

### Related issue(s)/PR(s)
none


Mark-H (approved):

Jako (approved):

JoshuaLuckers:
@Mark-H in 3.x this is already fixed

alroniks (approved):

PR #15902, opened 2021-11-12
Switch resource_preview to resource_view

muzzwood:
### What does it do?
Changes the lexicon key from `resource_preview` to `resource_view`, and the English words _Preview_ to _View_.
This is used in the context menu of the resource tree.

Screenshot from 2021-11-12 17-28-23

### Why is it needed?
Adds uniformity to action names. The button `View` on the resource action bar does the same thing but they're named differently.

Screenshot 2021-11-12 at 18-04-16 Editing Home MODX Revolution

The term preview may also make a user think they can see changes without saving.

### How to test
Right click on a resource in the tree, or look in recent resources at ?a=security/profile

### Related issue(s)/PR(s)
I remember this being mentioned before but if there's an issue I've missed it.


Ruslan-Aleev (approved):

Ibochkarev (approved):

JoshuaLuckers:
I believe the `resource_preview` lexicon should be deleted in the language files and not renamed (except for `en`). 
@opengeek Am I right?

Mark-H (approved):

opengeek (changes requested):
PRs should never modify lexicon files other than the ones in en.

opengeek:
> I believe the `resource_preview` lexicon should be deleted in the language files and not renamed (except for `en`). @opengeek Am I right?

I don't think the language files should be touched at all. Just modify the en/ source.

Ruslan-Aleev:
> I don't think the language files should be touched at all. Just modify the en/ source.

Why should we keep unused lexicons? By the way, there are so maaany of them, in fact.

opengeek:
> > I don't think the language files should be touched at all. Just modify the en/ source.
> 
> Why should we keep unused lexicons? By the way, there are so maaany of them, in fact.

No one is suggesting keeping unused lexicons. CrowdIn will modify them based on changes to the source (i.e. the en lexicon is the source).

muzzwood:
Sorry for the delay, I'll get this updated tonight/tomorrow. Thanks!

theboxer (approved):

Mark-H:
Setting this to rc2 so there's time for translations to catch up before the release.

PR #15907, opened 2021-11-19
[3.x] Package Manager UX - loading mask

muzzwood:
### What does it do?
Adds a mask with a loading animation to the packages grid on render. A small UX improvement.

Here is the current behaviour. Note that sometimes the wait time can be much longer than what is shown in the video.

https://user-images.githubusercontent.com/5160368/142601114-d6988a18-af79-48e2-aa28-c85f48f66967.mp4

Here is the behaviour with this change applied:

https://user-images.githubusercontent.com/5160368/142601420-9ec563c8-556c-423c-9ae1-4fb0909c3133.mp4

### Why is it needed?
When the cache has been cleared or doesn't exist yet, the main package management grid can take a while to load and currently it just sits there with no indication it's doing anything. 
It can be confusing considering this behaviour (before anything has loaded) is the same as if there are no packages installed.

### How to test
Delete your cache and then go to the package manager.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -149,7 +149,7 @@ MODx.grid.Package = function(config) {
     MODx.grid.Package.superclass.constructor.call(this,config);
     this.on('render',function() {
         this.mask = new Ext.LoadMask(this.body.dom, {
-            msg: _('checking_for_package_updates') + '...'
+            msg: _('checking_for_package_updates')
         });
         if (!this.loaded) {
             this.mask.show();
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -148,7 +148,9 @@ MODx.grid.Package = function(config) {
     });
     MODx.grid.Package.superclass.constructor.call(this,config);
     this.on('render',function() {
-        this.mask = new Ext.LoadMask(this.body.dom);
+        this.mask = new Ext.LoadMask(this.body.dom, {
+            msg: _('loading_packages') + '...'
+        });
         if (!this.loaded) {
             this.mask.show();
         }
--- a/manager/assets/modext/workspace/package.grid.js
+++ b/manager/assets/modext/workspace/package.grid.js
@@ -148,8 +148,17 @@ MODx.grid.Package = function(config) {
     });
     MODx.grid.Package.superclass.constructor.call(this,config);
     this.on('render',function() {
-        this.getView().mainBody.update('<div class="x-grid-empty">' + _('loading') + '</div>');
-    },this);
+        this.mask = new Ext.LoadMask(this.body.dom);
+        if (!this.loaded) {
+            this.mask.show();
+        }
+    }, this);
+    this.getStore().on('load', function(s) {
+        if (this.mask) {
+            this.mask.hide();
+        }
+        this.loaded = true;
+    }, this);
     this.on('afterrender', function () {
         this.uploader = new MODx.util.MultiUploadDialog.Upload({
             url: MODx.config.connector_url,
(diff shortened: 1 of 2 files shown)


Jako:
You can change the 'Loading …' text with http://cdn.sencha.com/ext/gpl/3.4.1.1/docs/#!/api/Ext.LoadMask

```new Ext.LoadMask(this.body.dom, {msg:"Check for package updates."});```

**Comment by muzzwood** [comment]
Nice one @Jako !

**Comment by Ruslan-Aleev** [comment]
@muzzwood @Jako Sorry, didn’t write before. Do we really need an extra lexicon? Isn't it clear with "Loading ..."?

**Comment by Jako** [comment]
@Ruslan-Aleev The loading delay happens because MODX checks for package updates from the package providers in the processor. Thats a lot slower than the other grid processors and the reason for loading the grid that slow can be shown.

@muzzwood The text should show something like 'Check for package updates.' or similar.

**Comment by Ruslan-Aleev** [comment]
@Jako It's clear. But I mean is that we don't need to create a new lexicon with the `loading_packages` key if there is a lexicon with the `loading` key. This is an extra line, and more translations.

**Comment by muzzwood** [comment]
I think differentiating the loading of the package grid from others using `Checking for package updates...` is worth the extra lexicon/translations. It signifies a difference i.e. it may take longer than normal as it needs to check each one.

Screenshot from 2021-11-20 13-51-33


That being said, I don't feel incredibly strongly about it so no worries either way 🤷🏻 🙂

**Review by Ibochkarev — APPROVED** [review]

**Review by Jako — APPROVED** [review]

**Review by JoshuaLuckers — APPROVED** [review]

**Review by Mark-H — APPROVED** [review]

File: manager/assets/modext/workspace/package.grid.js [hunk 1]
```diff
@@ -148,8 +148,19 @@ MODx.grid.Package = function(config) {
     });
     MODx.grid.Package.superclass.constructor.call(this,config);
     this.on('render',function() {
-        this.getView().mainBody.update('<div class="x-grid-empty">' + _('loading') + '</div>');
-    },this);
+        this.mask = new Ext.LoadMask(this.body.dom, {
+            msg: _('checking_for_package_updates') + '...'
```
mishanthrop: Why is the ellipsis separate from the lexicon?

muzzwood: The reasoning is so it doesn't need to be included in each language lexicon file, and theoretically the lexicon could be used elsewhere where an ellipses may not be appropriate. 
I see this is not the standard way though so if people would prefer this changed for uniformity sake, I don't mind! :)

JoshuaLuckers: @muzzwood in some cases it's included in the lexicon, in some it isn't (as far as I've seen). I don't mind the way it is now.

mishanthrop: It is strange that the formation of the text occurs in the code

`> theoretically the lexicon could be used elsewhere where an ellipses may not be appropriate`
Theoretically do you take into account the peculiarities of languages (for example, RTL)?

JoshuaLuckers: > Theoretically do you take into account the peculiarities of languages (for example, RTL)?

Good point, @muzzwood would you be so kind as to include the ... in the lexicon? Thanks!

muzzwood: Sorry for the delay. Very good point @mishantrop , I'll get that changed tonight/tomorrow.

muzzwood:
The failing code quality check is referencing files that aren't related to this PR... 🤔

opengeek:
> The failing code quality check is referencing files that aren't related to this PR... 🤔

I'm not sure what the heck is going on…
Screen Shot 2021-12-02 at 8 34 13 PM

theboxer (approved):

PR #15911, opened 2021-11-21
Fix Zend OPcache API warning when upgrading MODX

### LINKED ISSUE #15897: Zend OPcache API is restricted during upgrade 2.8 - 3.0beta1
### Summary
When trying to upgrade 2.8.3 to 3.0.0 beta1 itt gives me OPcache API error

### Step to reproduce
copy all new files over old ones, 
run setup
get 
Warning: Zend OPcache API is restricted by "restrict_api" configuration directive in /usr/www/users/XXX/setup/includes/modinstallsettings.class.php on line 97 
Warning: Zend OPcache API is restricted by "restrict_api" configuration directive in /usr/www/users/XXX/setup/includes/modinstallsettings.class.php on line 97 
Warning: Cannot modify header information - headers already sent by (output started at /usr/www/users/XXX/setup/includes/modinstallsettings.class.php:97) in /usr/www/users/powerycwff/setup/includes/request/modinstallrequest.class.php on line 232 

cache was manually deleted before upgrade attempt

### Observed behavior
error and can not carry on install

### Expected behavior
should complete install

### Related issue(s)/PR(s)
n/a

### Environment
apache 
php 7.4 and 8.0
mariaDB

JoshuaLuckers:
### What does it do?
Check if we are allowed to call OPcache API functions.

### Why is it needed?
To fix an OPcache API functions warning during upgrades.

### How to test?
See issue #15897 

### Related issue(s)/PR(s)
Closes #15897

### DIFF
--- a/setup/includes/modinstallsettings.class.php
+++ b/setup/includes/modinstallsettings.class.php
@@ -94,7 +94,9 @@ class modInstallSettings {
     public function load() {
         if (file_exists($this->fileName)) {
             if (function_exists('opcache_invalidate')) {
-                opcache_invalidate($this->fileName);
+                if (stripos(__FILE__, ini_get('opcache.restrict_api')) !== false) {
+                    opcache_invalidate($this->fileName);
+                }
             }
             $this->settings = include $this->fileName;
             if (empty($this->settings)) {
--- a/setup/includes/modinstallsettings.class.php
+++ b/setup/includes/modinstallsettings.class.php
@@ -41,11 +41,6 @@ class modInstallSettings
     public function load()
     {
         if (file_exists($this->fileName)) {
-            if (function_exists('opcache_invalidate')) {
-                if (stripos(__FILE__, ini_get('opcache.restrict_api')) !== false) {
-                    opcache_invalidate($this->fileName);
-                }
-            }
             $this->settings = include $this->fileName;
             if (empty($this->settings)) {
                 $this->restart();
--- a/setup/includes/modinstallsettings.class.php
+++ b/setup/includes/modinstallsettings.class.php
@@ -11,31 +11,88 @@
 /**
  * @package setup
  */
-class modInstallSettings {
+class modInstallSettings
+{
     public $install = null;
     public $config = [];
     public $fileName = '';
     public $settings = [];
 
-    function __construct(modInstall &$install,array $config = []) {
+    public function __construct(modInstall &$install, array $config = [])
+    {
         $this->install =& $install;
-        $this->config = array_merge([],$config);
+        $this->config = array_merge([], $config);
 
-        $this->fileName = $this->getCachePath().'settings.cache.php';
+        $this->fileName = $this->getCachePath() . 'settings.cache.php';
         $this->load();
     }
 
     /**
      * @return string
      */
-    protected function getCachePath() {
+    protected function getCachePath()
+    {
         return MODX_CORE_PATH . 'cache/' . (MODX_CONFIG_KEY == 'config' ? '' : MODX_CONFIG_KEY . '/') . 'setup/';
     }
 
     /**
      * @return void
      */
-    protected function rebuildDSN() {
+    public function load()
+    {
+        if (file_exists($this->fileName)) {
+            if (function_exists('opcache_invalidate')) {
+                if (stripos(__FILE__, ini_get('opcache.restrict_api')) !== false) {
+                    opcache_invalidate($this->fileName);
+                }
+            }
+            $this->settings = include $this->fileName;
+            if (empty($this->settings)) {
+                $this->restart();
+            }
+        }
+    }
+
+    /**
+     * @return void
+     */
+    public function restart()
+    {
+        $this->erase();
+        if (empty($this->install->request) && !($this->install->request instanceof modInstallCLIRequest)) {
+            header('Location: ' . MODX_SETUP_URL . '?restarted=1');
+            exit();
+        }
+    }
+
+    /**
+     * @return void
+     */
+    public function erase()
+    {
+        if (file_exists($this->fileName)) {
+            @unlink($this->fileName);
+        }
+    }
+
+    /**
+     * @param string $k
+     * @param mixed $v
+     * @return void
+     */
+    public function set($k, $v)
+    {
+        $this->settings[$k] = $v;
+        if (in_array($k, ['database_type', 'database_server', 'dbase', 'database_connection_charset'])) {
+            $this->rebuildDSN();
+        }
+    }
+
+    /**
+     * @return void
+     */
+    protected function rebuildDSN()
+    {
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
phpcs is failing because of this:
```
Error: <error line="14" column="1" severity="error" message="Each class must be in a namespace of at least one level (a top-level vendor name)" source="PSR1.Classes.ClassDeclaration.MissingNamespace"/>

Error: Each class must be in a namespace of at least one level (a top-level vendor name)
```

setup/includes/modinstallsettings.class.php
```diff
@@ -11,31 +11,88 @@
...
 
     /**
      * @return void
      */
-    protected function rebuildDSN() {
+    public function load()
+    {
+        if (file_exists($this->fileName)) {
+            if (function_exists('opcache_invalidate')) {
+                if (stripos(__FILE__, ini_get('opcache.restrict_api')) !== false) {
```
Jako: I don't get the clue on this patch. You are searching the path of the current file for the value of opcache.restrict_api php setting?

JoshuaLuckers: Yes, maybe I should use $_SERVER['SCRIPT_FILENAME'] instead. 
But see https://serverfault.com/a/998876 for more info.

Something similar is done in Joomla as well: https://github.com/joomla/joomla-cms/blob/ebd801ade267cfc367979972c7876480c6343248/administrator/components/com_joomlaupdate/extract.php#L1791

opengeek:
I'm afraid we are oversimplifying the problem here. If opcache is enabled and calling the function is restricted in setup, I believe we need to warn the user and stop setup. Otherwise, we are going to load stale settings here and problems will arise. Or we need to find another way to make sure opcache is not used for setup.

Omeryl:
My recommendation is to `ini_set()` opcache.enable to false for the set up, which is something that can be done at runtime. 

```
opcache.enable bool
Enables the opcode cache. When disabled, code is not optimised or cached. The setting opcache.enable can not be enabled at runtime through ini_set(), it can only be disabled. Trying to enable it in a script will generate a warning.
```

setup/includes/modinstallsettings.class.php
Jako: Ok. Thanks. I somehow got the wrong description for that setting.

JoshuaLuckers: @Jako To clear the confusion:

The description for the `opcache.restrict_api` setting:
> Allows calling OPcache API functions only from PHP scripts which path is started from specified string. The default "" means no restriction.

We are allowed to use `opcache_` functions if the path defined in `opcache.restrict_api` can be found in the path of the current file/script. 

> stripos(string $haystack, string $needle, int $offset = 0)
> Find the numeric position of the first occurrence of `needle` in the `haystack` string.

JoshuaLuckers:
> My recommendation is to `ini_set()` opcache.enable to false for the set up, which is something that can be done at runtime.
> 
> ```
> opcache.enable bool
> Enables the opcode cache. When disabled, code is not optimised or cached. The setting opcache.enable can not be enabled at runtime through ini_set(), it can only be disabled. Trying to enable it in a script will generate a warning.
> ```

At runtime this will only disable OPCache for the current request (it disables caching of the remaining scripts in your current request). 
Source: https://stackoverflow.com/a/21590668

And if I understand correctly, the files already in cache are not invalidated. Other requests might still use the cached scripts/files. Therefore we need to invalidate it by calling `opcache_invalidate`.

Omeryl:
@JoshuaLuckers As far as I understand, as long as the script file isn't yet included it will bypass the OPcache entirely. You'd definitely want to short circuit it as soon as possible.

JoshuaLuckers:
> I'm afraid we are oversimplifying the problem here. If opcache is enabled and calling the function is restricted in setup, I believe we need to warn the user and stop setup. Otherwise, we are going to load stale settings here and problems will arise. Or we need to find another way to make sure opcache is not used for setup.

I was looking at the [PHP source code](https://github.com/php/php-src/blob/ae5498c929244ccb68d5b1bfa31329707e763430/ext/opcache/ZendAccelerator.c) and if I'm not mistaken, calling `opcache_invalidate` will only mark the cached file as invalid. We might still end up with a cached version of the file [until a certain threshold is met and the cache is refreshed](https://github.com/php/php-src/blob/ae5498c929244ccb68d5b1bfa31329707e763430/ext/opcache/ZendAccelerator.c#L1336-L1355). Correct me if I'm wrong.

opengeek:
> > I'm afraid we are oversimplifying the problem here. If opcache is enabled and calling the function is restricted in setup, I believe we need to warn the user and stop setup. Otherwise, we are going to load stale settings here and problems will arise. Or we need to find another way to make sure opcache is not used for setup.
> 
> I was looking at the [PHP source code](https://github.com/php/php-src/blob/ae5498c929244ccb68d5b1bfa31329707e763430/ext/opcache/ZendAccelerator.c) and if I'm not mistaken, calling `opcache_invalidate` will only mark the cached file as invalid. We might still end up with a cached version of the file [until a certain threshold is met and the cache is refreshed](https://github.com/php/php-src/blob/ae5498c929244ccb68d5b1bfa31329707e763430/ext/opcache/ZendAccelerator.c#L1336-L1355). Correct me if I'm wrong.

We've gone from over-simplifying to over-complicating. 😂 This was added because the cached settings values that were just written to the file were not getting loaded in this exact spot, causing old values to be used during this step of the upgrade. Adding this method fixed that problem, absolutely. If we prevent the settings file from ever being cached by turning this off for all of the setup, we can avoid the problem completely, I believe. This will of course require testing.

JoshuaLuckers:
> If we prevent the settings file from ever being cached by turning this off for all of the setup, we can avoid the problem completely, I believe. This will of course require testing.

I agree, we should turn off OPCache for all of the setup. I propose to always turn off OPCache if it's enabled, even if we are not restricted to call `opcache_invalidate`. 

@Omeryl recommends to ini_set() opcache.enable to false for the set up, which is something that can be done at runtime, at the earliest as possible.

If it's ok with all of you to disable OPCache I will add a check to see if OPCache is enabled and disable it in both [cli-install.php](https://github.com/modxcms/revolution/blob/e59f4e0de39adf5b3091ff0281d3062f4cdadb2f/setup/cli-install.php) and [index.php](https://github.com/modxcms/revolution/blob/e59f4e0de39adf5b3091ff0281d3062f4cdadb2f/setup/index.php).

opengeek:
This has never affected cli-install, to my knowledge. I believe it is unique to web install. Let's start with web and if the issue comes up, we can push the changes to cli.

JoshuaLuckers:
@opengeek done!

Mark-H (approved):

opengeek (approved):

PR #15919, opened 2021-11-25
[3.x] Refine grids code to correct actions access and visibility (revised)

smg6511:
NOTE: This is a simplified re-submission of a PR submitted on 11/23.

### What does it do?

This changes in this PR initially set out to fix issues raised in #14929, which uncovered other problems in the grids UI. There are quite a few loose ends there in terms of reflecting what a user can do according to their group permissions and object-level access policies (when applied [to contexts, media sources, etc.]). Additionally there's a lack of full validation within some grids, including their create forms. Also, as @Ruslan-Aleev points out in #15281, validation response is inconsistent because the proper specification is missing on the js side of things. I've arrived at the solution you'll see by creating a number of methods and properties in the main grid class (modx.grid.js), as well as an updated way to identify/apply permissions in the child grid classes and their respective GetList processors.

**Fixes**

1. First and foremost, the visibility of the actions (gear) menu and its contents matched to permissions
2. Ensures certain core entries are protected from deletion and that certain core entry values can not be duplicated in other user-entered entries (e.g., _Filesystem_ [in sources], _mgr_ and _web_ keys and _Manager_ name [in contexts], _Super User_ and _Member_ name [in roles])
3. Provides in-grid validation where it was missing and more visible feedback in some instances
4. General code consistency and clarity in the associated js and php files
5. Match front end validation messages with back end ones (mostly via adding the `blankText` config to required fields)
6. For checkbox model grids (e.g., the media sources one shown in this PR), adds a new checkbox style for rows that can not be selected for batch processing (protected from deletion in this case). Further, the batch delete toolbar menu item will only be active when one or more deletable grid items is selected. 

**Proposed Enhancements Included**

1. Created a new arbitrary field (creator) to distinguish core entries from user entries. This enables better control the over view (i.e., you can sort on this field) and a mechanism to hide one group or the other via a toolbar control (not implemented here). IMO the ability to hide, for example, the 12 core policy entries when working on your own set of policies would be nice (especially if you're working on a relatively large set of policies).
2. Also set apart core entries by italicizing them and setting a slightly different color (`$colorSplash`)
3. Changed the anchor/link style to a dotted bottom border instead of solid. There many solid rules/borders in the UI, which is fine, but the solid text decoration on links seems over the top to me. IMO the dotted rule is still plenty visible but tones it down a bit.

### How to Test

Note that while I have applied some of my changes to a couple other grids, their implementation may be incomplete. The three grids to test for this draft are: 

- The Roles grid (ACLs/Roles [tab])
- The Contexts grid
- The Media Sources grid

**Process**
1. Run `grunt build` to ensure css is updated.
2. Create a secondary user that you will apply lower-level permissions to; then log in as this user from a different browser)
3. Create a new [Test User Group] and ACL administrator policy to apply to that group; then add your new secondary user to the test group.
4. Experiment with a variety or permissions in this new policy and verify their effect on each grid. The relevant permissions for each grid are:
    - Roles: _save_role_, _edit_role_, _new_role_, and _delete_role_
    - Contexts: _save_context_, _edit_context_, _new_context_, and _delete_context_
    - Sources: _source_save_, _source_edit_, and _source_delete_
5. Create a new context policy using the _ContextTemplate_ template  and a new media source policy using the _MediaSourceTemplate_ template that will be used to further limit permissions for specific test contexts and sources you've created.
6. For contexts and sources, create new user-group specific access rules in ACLs/[Test User Group]/Access Permissions. Experiment with these permissions and verify your secondary user's access is reflected accurately in each grid.
7. In each of the three grids, attempt to rename one of your entries with the same name as one of the core entries (where creator is 'modx'). You should see an alert prompting you to enter another valid value.

As those of you who actually coded these areas of MODX or are very familiar with their inner workings, this is a necessarily highly tedious process. I'd envision doing this across multiple PRs so small logical batches could be tested and put into place over time (in as short a time as possible).

### Screenshots Highlighting Changes Made

Using the Sources grid as an example (since it highlights all of the fixes/proposed features implemented), below are a few annotated visuals to get a quick idea of what to expect.

Screen Shot 2021-11-23 at 11 17 13 PM

Screen Shot 2021-11-23 at 11 38 07 PM

Screen Shot 2021-11-23 at 11 34 44 PM


manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+    }
+
+    ,userCanDeleteRecord: function(record) {
+        const permissions = record.json.permissions;
+        return !Ext.isEmpty(permissions) && !record.json.isProtected && permissions.delete === true ? true : false ;
+    }
+
+    ,userCanDuplicateRecord: function(record) {
+        const permissions = record.json.permissions;
+        return !Ext.isEmpty(permissions) && permissions.duplicate === true ? true : false ;
```
mishanthrop: It confuses me that `permissions` is declared as an array, but at the same time it has such properties as `update`, `delete` and `duplicate`.
The project needs TypeScript...

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+            const modePermission = mode === 'userCanExport' ? true : this[mode] ;
+            if (['userCanCreate', 'userCanEdit'].includes(mode) && modePermission === true) {
+                this.hasSavePermissions = true;
+            }
+            permissions.push(modePermission);
+        });
+        this.showActionsMenu = permissions.length === 0 || permissions.every(permission => permission === false) === true
+            ? false
+            : true
+            ;
```
mishanthrop: nit

Also it would be useful to cover such code with unit-tests.
```suggestion
        this.showActionsMenu = !(!permissions.length || permissions.every(permission => !permission)) ;
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+     * @param {Number} protectedIdentifiers - The record identifiers to be protected (making them non-editable/deletable)
+     *
+     * @return {Boolean}
+     */
+    ,recordIsProtected: function(subject, protectedIdentifiers) {
+        if (Ext.isEmpty(protectedIdentifiers)) {
+            return false;
+        }
+        protectedIdentifiers = protectedIdentifiers.map(identifier => {
+            return typeof identifier === 'string' ? identifier.trim() : identifier ;
```
mishanthrop: nit

The project needs linter...
```suggestion
            return typeof identifier === 'string' ? identifier.trim() : identifier;
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+     */
+    ,valueIsReserved: function(reservedValues, value) {
+        if (!Ext.isArray(reservedValues)) {
+            reservedValues = reservedValues.split(',');
+        }
+        let isReserved = false;
+        reservedValues.forEach(reserved => {
+            if (reserved.toLowerCase() === value.toLowerCase()) {
+                isReserved = true;
+                return;
```
mishanthrop: You cant stop forEach-iteration of `reservedValues` by calling `return` inside callback.
Use `Array.some` to make sure there is an item in the array.

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -665,32 +861,77 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+                // Checking record-level permissions
+                if (record.data.hasOwnProperty('cls')) {
+                    if (Ext.isEmpty(record.data.cls)) {
+                        return;
+                    }
+                }
+                if (record.json.hasOwnProperty('permissions')) {
+                    if (
+                        Ext.isEmpty(record.json.permissions) ||
+                        Object.values(record.json.permissions).every(permission => permission === false) === true
```
mishanthrop:
```suggestion
                        Object.values(record.json.permissions).every(permission => !permission)
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+        permissions = permissions.map(item => item.trim());
+        this.userCanDelete = permissions.every(permission => MODx.perm[permission]);
+    }
+
+
+    /* Record-Level Permissions Checks, for objects with specific policies */
+
+    ,userCanEditRecord: function(record) {
+        const permissions = record.json.permissions;
+        return !Ext.isEmpty(permissions) && permissions.update === true ? true : false ;
```
mishanthrop:
```suggestion
        return !Ext.isEmpty(permissions) && permissions.update;
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+    }
+
+    ,userCanDeleteRecord: function(record) {
+        const permissions = record.json.permissions;
+        return !Ext.isEmpty(permissions) && !record.json.isProtected && permissions.delete === true ? true : false ;
```
mishanthrop:
```suggestion
        return !Ext.isEmpty(permissions) && !record.json.isProtected && permissions.delete;
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+     * @property {Function} valueIsReserved - Wraps a grid value with a real or simulated link — a trigger that appears
+     * like an anchor link, usually to access a dropdown chooser or other control
+     *
+     * @param {Array|String} reservedValues - A set of values that can not be used for a particular object's field
+     * @param {Object} value - The submitted value being tested
+     *
+     * @return {Boolean}
+     */
+    ,valueIsReserved: function(reservedValues, value) {
+        if (!Ext.isArray(reservedValues)) {
```
mishanthrop: Ext.js in 2021...

```suggestion
        if (!Array.isArray(reservedValues)) {
```

mishantrop: Снимок экрана 2021-11-25 в 19 44 12

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -495,14 +515,190 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+        if (selectionList === false) {
+            return false;
+        }
+        selectionList.split(',').forEach(id => {
+            id = itemIdType === 'string' ? id : parseInt(id) ;
+            if (!this.recordIsProtected(id, this.protectedIdentifiers) && !this.nonRemoveableRecords.includes(id)) {
+                removableItems.push(id);
+            }
+        });
+        return removableItems.length > 0 ? removableItems : false ;
```
mishanthrop:
```suggestion
       if (removableItems.length > 0) {
           return removableItems
       };
```

Its okay to return undefined. 
It not semantically correct to return boolean value.

manager/assets/modext/widgets/source/modx.panel.sources.js
```diff
@@ -63,221 +68,377 @@ Ext.reg('modx-panel-sources',MODx.panel.Sources);
...
-        }]
+        }],
+        viewConfig: {
+            forceFit: true,
+            scrollOffset: 0,
+            getRowClass: function(record, index, rowParams, store) {
+                // Adds the returned class to the row container's css classes
+                const disableClasses = record.json.isProtected ? 'modx-protected-row disable-selection' : 'disable-selection' ;
+                return !this.grid.userCanDeleteRecord(record) ? disableClasses : '';
+            }
```
mishanthrop: + more extensible code
+ no duplication

```suggestion
            getRowClass: function(record, index, rowParams, store) {
                // Adds the returned class to the row container's css classes
                if (this.grid.userCanDeleteRecord(record)) {
                    return ''
                }
                const disableClasses = ['disable-selection']
                if (record.json.isProtected) {
                    disableClasses.push('modx-protected-row')
                }
                return disableClasses.join(' ');
            }
```

manager/assets/modext/widgets/source/modx.panel.sources.js
```diff
@@ -63,221 +68,377 @@ Ext.reg('modx-panel-sources',MODx.panel.Sources);
...
+    this.setShowActionsMenu();
+
+    this.on({
+        render: function(grid) {
+            this.setEditableColumnAccess(
+                ['modx-source--name', 'modx-source--description']
+            );
+        },
+        beforeedit: function(e){
+            if (e.record.json.name === 'Filesystem' || !this.userCanEditRecord(e.record)) {
```
mishanthrop: Hardcode

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -169,16 +169,36 @@ MODx.grid.Grid = function(config) {
 
     this.on('click', this.onClickHandler, this);
 };
-Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
+Ext.extend(MODx.grid.Grid, Ext.grid.EditorGridPanel, {
+
     windows: {}
 
+    ,protectedIdentifiers: null
```
mishanthrop: May be to get it from `$coreContexts` in any way?

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Perhaps I could refine the property naming to bring clarity to their differences. 

To explain my reasoning, there are two levels of permissions to evaluate: the primary, class-level ones (whether or not you have the permission to create/edit/delete, say, a _media source_ in general); and the secondary, object-level ones (where a policy, such as _Media Source User_, is applied within a user group's access permissions to further limit access to a particular source). The majority of MODX's grids do not need to evaluate an object-level permission, as those only apply to media sources, contexts, namespaces,  resource groups, and element categories.

Looking at each of the three grid classes I updated, you'll see (using the sources one again as example) this set of statements toward the end of its config:
```
this.gridMenuActions = ['edit', 'delete', 'duplicate'];

this.setUserCanEdit(['source_save', 'source_edit']);
this.setUserCanCreate(['source_save']);
this.setUserCanDelete(['source_delete']);
```
The `gridMenuActions` array indicates the _possible_ actions the user could take on this grid. The next three methods called (`setUserCanX`) check the MODx.perm object for the action's required permissions using the passed permissions keys. The `userCanX` set in the process are globally available in the class. This provides a universal way to define and utilize these primary permissions.

I've revised how secondary permissions get passed on a row-by-row basis, via an associative array on the php/processor end which is converted to an object on the js end.  I felt it was much more clear to have
```
permissions: {
   edit: true
}
```
as opposed to what is currently used in many places
`cls: 'pedit'`

Bottom line: Where these object-level permissions checks apply, they'll only be called if the primary class-level permissions checks pass.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: So I have to say, respectfully, this is a total nit-pick that is unproductive IMO. Honestly this is a personal preference of mine when writing ternary expressions, as it's more pleasing to _my_ eye to see each part of the expression separated with the same spacing. BTW, the php linter I use to apply PSR-12 does not complain about or change this spacing. Prettier does remove the space in js.

smg6511:
@mishantrop - First of all I'd like to express that I appreciate you taking the time to review this PR. It would, however, be much more useful for you to review the functionality and the usability of what is being implemented rather than getting into the weeds with super fine-grained code suggestions and what are, in some instances, personal coding/code-style preferences. 

Does the PR solve the problem it's trying to solve? Does it do it in a way that improves the UI? Does it bring a useful new feature? I'm certainly not looking to ignore your suggestions, and if I and others feel some of them would provide an improvement I'll make the change(s). I'm just saying I think your _initial_ focus is in the wrong place.

manager/assets/modext/widgets/core/modx.grid.js
mishanthrop: Yes, it is nit-pick to this line. But in general it is hard to read code, written in different styles by different people.

mishanthrop: `>  the php linter I use to apply PSR-12`
PSR is for PHP, not for JavaScript

mishanthrop:
@smg6511 - First of all I'd like to express that I appreciate you taking the time to write new features..

`> It would, however, be much more useful for you to review the functionality and the usability of what is being implemented`
Changes to the code can affect functionality. Therefore, first of all, I pay attention to the quality (maintainability, readability, structure etc) of the code.

`> in some instances, personal coding/code-style preferences`
In "some", but not all. Updated my nit-comments.

`> Does the PR solve the problem it's trying to solve?`
I do not question the value of your PR. But I want to take part in the development of MODX and it is important for me that the code is pleasant to work with.

`> I'm just saying I think your initial focus is in the wrong place.`
Don't take my comments personally, this is a common practice in code reviews. Reviewers writes comments, and the author fixes the code or does not fix it. The main thing is to find a compromise.

JoshuaLuckers:
@smg6511 You're on fire! There's a lot to cover in this PR. Please take a look at the technical implementation of PR #14009. Some similar changes have been made relevant to some of yours. 

I'll do my best and find some time in the upcoming days to provide feedback on the functionality and usability of the fixes and new features in this PR.

smg6511:
@JoshuaLuckers - I appreciate the encouragement! Yes, this initial PR is a bit to review, I know ;-) But, once finalized, it will provide an pretty easy framework to update the rest of the grids. I'll take a look at 14009 in just a bit...

OK, I just took a quick look through 14009 and I don't see any conflicts with what I've done. Even so, assuming that my PR in some form gets merged and I move on to other grids (in 14009's case the policy/policy template grids), I'll refer back to 14009 to ensure I don't walk back anything important.

smg6511:
@mishantrop - Don't worry, I'm not taking anything personally. And I, of course, agree that code quality, consistency, and conformity to a standard is also very important. (I also know from comments on other issues/PRs that you and I might not share the same viewpoint on some stylistic and implementation conventions, but that's Ok.) I was just pointing out that I don't think that's the _first_ thing to focus on when you're reviewing a PR.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Yes, what I was referring to is that I use this same ternary style in my php ;-)

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Good catch. I took your suggestion to rewrite using .some.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: I'm going to leave this one as-is, as I find it more readable.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Suggestion taken.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: I had been deferring to Ext's polyfills, but you're probably right that this one's not needed since the native isArray is widely compatible and the Ext version does nothing special (looking at both 3.4 and 7.x).

manager/assets/modext/widgets/core/modx.grid.js
smg6511: I know the truthy check works, but I made a conscious decision to check for strict true. Leaving the way this is done in the three methods for now...

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Leaving for now, explanation above.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: OK, I see what you're saying ... changed it a bit here and where called.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Made a handful of variable name changes to further clarify.

manager/assets/modext/widgets/source/modx.panel.sources.js
smg6511: No, I'm pretty sure this has to be done on-the-fly.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: I'm 99% sure we can not do this. I know what you're getting at and I tried to find a way to not have explicitly state these identifiers in the js, but the event timing poses a problem.

JoshuaLuckers:
> OK, I just took a quick look through 14009 and I don't see any conflicts with what I've done. 

There are no conflicts with PR #14009. I was referring to the technical implementation of certain things, especially things related to _"Ensures certain core entries are protected from deletion"_.
Like the use of constants for some of the things you hardcoded.
E.g: https://github.com/modxcms/revolution/blob/d96badf40a7c6bc550dc90305c802bbdd8d2adae/core/src/Revolution/modAccessPolicy.php#L41-L62

https://github.com/modxcms/revolution/blob/d96badf40a7c6bc550dc90305c802bbdd8d2adae/core/src/Revolution/Processors/Model/GetListProcessor.php#L28-L29

manager/assets/modext/widgets/source/modx.panel.sources.js
smg6511: OK, I updated this in a little different way to avoid duplication (see below). Because this will only be implemented in a handful of places, has to be done on each class where it is implemented, and it's unlikely the class lists for each will need to be added to, I don't think extensibility is really a factor.
```
getRowClass: function(record, index, rowParams, store) {
    // Adds the returned class to the row container's css classes
    if (this.grid.userCanDeleteRecord(record)) {
        return '';
    }
    const rowClasses = 'disable-selection';
    return record.json.isProtected ? `modx-protected-row  ${rowClasses}` : rowClasses ;
}
```

smg6511:
@JoshuaLuckers - Don't know if you withdrew the comments you were making re #14009. (I was about to respond and they disappeared.) Anyway, I see I'll want to carry through how the core items are defined in constants in each object's base class, as they were done in the _modAccessPolicy_ one. 

However, I do think my implementation of the final relay of those permissions via prepareRow of each object class is more straightforward and immediately readable. So when you're reading the record in js, your getting, for example:
```
id: 3,
permissions: {
   create: false,
   update: true,
   delete: false
}
```
instead of:
```
id: 3,
cls: 'pedit'
```

The way I've done it is also very easy to iterate through without any additional fuss.

Note also, I think it makes sense is to use CRUD nomenclature across the back end (in code), which is why I've changed for example _edit_ to _update_ and _remove_ to _delete_. On the front end, it makes more sense (in my mind at least) to use natural language (e.g., _Edit X_ instead of _Update X_). Could be over-thinking it though ;-)

JoshuaLuckers:
Yes I have deleted my comment. I have a lot to share but I still need to find time to properly work it out haha.

But your in the right direction with where I want to go, but you're right about the class names like pedit.

If your on Slack I could share my unstructured notes.

smg6511:
Yes, I am on Slack for another project I work on. How shall we connect?

JoshuaLuckers:
> Yes, I am on Slack for another project I work on. How shall we connect?

There is a MODX Community Slack: https://modx.org/

core/src/Revolution/Processors/Security/Role/GetList.php
```diff
@@ -90,21 +95,28 @@ public function beforeIteration(array $list)
...
+        if ($isCoreRole) {
+            $baseKey = '_role_' . strtolower(str_replace(' ', '', $roleName)) . '_';
+            $roleData['name_trans'] = $this->modx->lexicon($baseKey . 'name');
+            $roleData['description_trans'] = $this->modx->lexicon($baseKey . 'description');
         }
-        $objectArray['perm'] = implode(' ', $perm);
 
-        return $objectArray;
+        $roleData['reserved'] = ['name' => ['Super User', 'Member']];
+        $roleData['isProtected'] = $isCoreRole ? true : false ;
```
mishanthrop: What is the benefit of the ternary condition here?
```suggestion
        $roleData['isProtected'] = $isCoreRole;
```

core/src/Revolution/Processors/Source/GetList.php
```diff
@@ -114,31 +144,32 @@ public function getSortClassKey()
...
+            $baseKey = '_source_' . strtolower(str_replace(' ', '', $sourceKey)) . '_';
+            $sourceData['name_trans'] = $this->modx->lexicon($baseKey . 'name');
+            $sourceData['description_trans'] = $this->modx->lexicon($baseKey . 'description');
         }
 
-        $cls = [];
-        if ($canSave && $canEdit && $object->checkPolicy('save')) {
-            $cls[] = 'pupdate';
+        $sourceData['reserved'] = ['name' => $coreSources];
+        $sourceData['isProtected'] = $isCoreSource ? true : false ;
```
mishanthrop:
```suggestion
        $sourceData['isProtected'] = $isCoreSource;
```

core/src/Revolution/Processors/Security/Role/GetList.php
smg6511: Good catch, changed in all three GetList classes.

smg6511:
@JoshuaLuckers - Ok, I'm on MODX Slack now (Jim Graham).

JoshuaLuckers:
> @JoshuaLuckers - Ok, I'm on MODX Slack now (Jim Graham).

I sent you a direct message!

_build/templates/default/sass/_utility.scss
```diff
@@ -212,3 +211,13 @@
     }
   }
 }
+
+@mixin textLink {
+    color: $colorSplash;
+    text-decoration: none;
+    border-bottom: 1px dotted scale-color($colorSplash, $lightness: 50%);
```
mishanthrop: `border` is an old hack. The only benefit is IE-compatibility.

https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-style
https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color

_build/templates/default/sass/index.scss
```diff
@@ -551,26 +559,28 @@ textarea.x-form-field {
...
-  text-decoration: underline;
-}
-
-a.x-grid-link:hover,
-a.x-grid-link:focus {
-  text-decoration: none;
+.x-grid-link {
+    @include textLink;
+    &.simulated-link {
+        cursor: pointer;
```
mishanthrop, line 567: Why is this line not in the`textLink`?

_build/templates/default/sass/index.scss
```diff
@@ -551,26 +559,28 @@ textarea.x-form-field {
...
+  &:not(.editor-disabled) {
+    cursor: pointer;
+    &:hover,
+    &:focus {
       color: $colorSplash;
+      > div {
+        &::after {
+            @include awesome-font;
+            content: fa-content($fa-var-pen);
+            margin: 0 0.5em;
```
mishanthrop: nit

I prefer to use `rem` instead of `em`, because `rem` is independent of the DOM-hierarchy

JoshuaLuckers:
@mishantrop would you be so kind as to wait reviewing the code until this PR is ready to finalize?

mishanthrop:
@JoshuaLuckers how to know that a PR is ready to finalize? The PR is open, there is no additional information (Draft, WIP etc) about the process.

Besides, my remarks do not concern the already reviewed files.

_build/templates/default/sass/index.scss
smg6511: Yes, I'm with you on this one, but it's something to tackle across all the sass files at some point in a different PR.

_build/templates/default/sass/index.scss
smg6511: It doesn't need to be defined there, only where the "link" is not an anchor tag.

_build/templates/default/sass/_utility.scss
smg6511: Good to know - for some reason that CSS 3 feature slipped by me!

smg6511:
@JoshuaLuckers - Hey, just checking to see if you'd gotten my reply to your comments on Slack.

JoshuaLuckers:
@smg6511 why did you remove the two CSS files?

smg6511:
@JoshuaLuckers - Yeah, I was combing through my open PRs and noticed those two files should not have been included in this one; I was _hoping_ that deleting them from within the PR would just remove them from the PR itself. I get the feeling that's not the case. :-/

core/src/Revolution/Processors/Context/GetList.php
```diff
@@ -33,12 +33,11 @@ class GetList extends GetListProcessor
     public $permission = 'view_context';
     public $languageTopics = ['context'];
     public $defaultSortField = 'key';
-    /** @var boolean $canEdit Determines whether or not the user can edit a Context */
-    public $canEdit = false;
-    /** @var boolean $canRemove Determines whether or not the user can remove a Context */
-    public $canRemove = false;
-    /** @var boolean $canCreate Determines whether or not the user can create a context (/duplicate one) */
-    public $canCreate = false;
```
Mark-H, line 44: While niche, renaming public attributes and making them protected are breaking changes and should be avoided. Someone may have extended the GetList processor. In fact, I think I may have personally done that in 2 of my extras.

Mark-H:
Main problem I have with this PR is that, AFAIK, roles and media sources should not be restricted. Yes they may get pre-installed, but if you want to delete the filesystem media source and only have an assets one in the system, you should be able to. Same with roles.

smg6511:
Revisiting this a bit ... @Mark-H , re your previous comment I've not added new restrictions/protections in this PR; the ones illustrated are and have been present in 2.x and 3.x (and should continue to be IMO)—I've just made it more clear in the UI.

core/src/Revolution/Processors/Context/GetList.php
theboxer: @smg6511 can you please revert these to `public` visibility? As mark pointed out, this is a breaking change.

core/src/Revolution/Processors/Security/Role/GetList.php
```diff
@@ -32,7 +32,11 @@ class GetList extends GetListProcessor
     public $languageTopics = ['user'];
     public $permission = 'view_role';
     public $defaultSortField = 'authority';
-    public $canRemove = false;
+
+    protected $canCreate = false;
+    protected $canUpdate = false;
+    protected $canDelete = false;
+    protected $coreRoles;
```
theboxer: @smg6511 same goes for these, please keep the name and visibility same. And I'd follow the suit with the new variables as well.

core/src/Revolution/Processors/Source/GetList.php
```diff
@@ -36,6 +36,12 @@ class GetList extends GetListProcessor
     /** @param boolean $isGridFilter Indicates the target of this list data is a filter field */
     protected $isGridFilter = false;
 
+    protected $canCreate = false;
+    protected $canUpdate = false;
+    protected $canDelete = false;
```
theboxer: @smg6511 and these

core/src/Revolution/Processors/Workspace/PackageNamespace/GetList.php
```diff
@@ -40,20 +41,66 @@ class GetList extends GetListProcessor
     /** @param boolean $isGridFilter Indicates the target of this list data is a filter field */
     protected $isGridFilter = false;
 
+    protected $canCreate = false;
+    protected $canUpdate = false;
+    protected $canDelete = false;
+
```
theboxer: @smg6511 .

manager/assets/modext/widgets/source/modx.panel.sources.js
```diff
@@ -56,168 +59,313 @@ Ext.reg('modx-panel-sources',MODx.panel.Sources);
...
+                return false;
+            }
+        }
+    });
+
+    this.getStore().on({
+        load: function(store, records, params){
+            records.forEach(record => {
+                if (!this.userCanDeleteRecord(record)) {
+                    this.nonRemoveableRecords.push(record.id);
```
theboxer: @smg6511 this line causes an error as `nonRemovableRecords` is undefined

theboxer (changes requested):

smg6511:
@theboxer - Now that someone's looking at this [ you ;-) ], I'll circle back to it in the next couple days to address comments and get the current conflicts sorted. (I haven't worked on this PR for a while now.) Thanks for checking it out!

core/src/Revolution/Processors/Context/GetList.php
smg6511: Took care of this in recent commit.

core/src/Revolution/Processors/Security/Role/GetList.php
smg6511: Done

core/src/Revolution/Processors/Workspace/PackageNamespace/GetList.php
smg6511: Done

core/src/Revolution/Processors/Source/GetList.php
smg6511: Done.

manager/assets/modext/widgets/source/modx.panel.sources.js
smg6511: Ok, this is fixed. I'd actually gone in a little different direction since submitting the original PR, and this listener is no longer used.

theboxer:
@smg6511 can you please ping me once finishing with updates?

smg6511:
@theboxer Will do. The question now is: In addition to doing any last tweaks/optimizations to what is here now, shall I carry this through the rest of the grids as a part of this PR, or leave that to a supplementary PR? (I don't believe the changes made thus far would negatively affect grids that have not been updated to use the new methods.)

theboxer:
@smg6511 how many affected grid you think there are, that will need an update?

smg6511:
@theboxer Ultimately all editor grids need an update of some sort to ensure the gear icon/menu only shows when applicable. I'll survey which grids still need work—two that come to mind that I'm working on now are the Policy and Policy Templates grids.

theboxer:
@smg6511 I guess it'd be fine to have them all here, but I'm not sure how much of an effort it is, so it'll all get into 3.1.

smg6511:
@theboxer @opengeek - Guys, there's something I hadn't noticed initially about the additional "Creator" column I've added as a feature of this PR: Because the creator data point is not a part of the db data, I'm finding it difficult (probably impossible) to provide an expected secondary sort (on the name column) when sorting the "Creator" column. So, when sorting on  "Creator," the name column is just randomly sorted (which I don't think is great). (I had a little hack going to build the Creator column on the fly, based on extra data generated from various GetList Processors.) 

Anyway, upon finding the secondary sorting issue, I _tried_ to pull another rabbit out of the hat, but I think this ultimately points to the need for one or two more columns in many of the db tables to achieve the desired grouping (and even better, filtering) of objects based on how they were created (i.e., by modx, by an Extra, or by a User). This is probably _not_ a job for this PR. So, I'm thinking of removing that aspect for now.

But I would like to get a read on whether you'd go for making some schema changes to add a general `creator` column and perhaps a couple others[1] to many of the tables (e.g., namespaces, contexts, sources, roles, policies, policy templates, etc.) where "ownership" of the row data can be useful. We could also lose the hard-coded constants in the GetList Processors that indicate protected system-installed data. Right now, the way of identifying protected items (by name in the aforementioned constants) seems a little fragile.

[1] Other columns? 
- `createdby` (specific User id when applicable)
- `protected` (boolean, whether the row can be deleted by User action [as opposed to a system process, such as modx installation/upgrade or an Extra being uninstalled])

opengeek:
Why wouldn't you just sort on the joined table columns for creator?

smg6511:
As it stands, creator doesn't exist in the db; it's derived from other "knowns" that are gathered on the php side of things (e.g., for Contexts, I know which are installed by modx via the hard-coded constants assessed by the `isCoreContext` method). So there's no way via the joins on the sql side (the trick used was sorting via the mysql FIELD() method, which is being fed an explicit list of names [presorted groupings of modx-installed names and, in the case of namespaces, extras-installed namespaces]). Yes, I could do something like create a temporary table to work with that arbitrary data as a "real" column ... but that'd be crazy and you'd have to demote me to a new repo status of "observer only" ;-) .

opengeek:
> As it stands, creator doesn't exist in the db; it's derived from other "knowns" that are gathered on the php side of things (e.g., for Contexts, I know which are installed by modx via the hard-coded constants assessed by the `isCoreContext` method). So there's no way via the joins on the sql side (the trick used was sorting via the mysql FIELD() method, which is being fed an explicit list of names [presorted groupings of modx-installed names and, in the case of namespaces, extras-installed namespaces]). Yes, I could do something like create a temporary table to work with that arbitrary data as a "real" column ... but that'd be crazy and you'd have to demote me to a new repo status of "observer only" ;-) .

Isn't creator just the name fields from user profile? I'm confused. It should absolutely be a simple join to that table from the createdby field.

smg6511:
I'm guessing you're thinking more in the context of Resources, where the so-called creator (createdby) is persisted in the data itself. I believe Resources are the only place that happens. But what I'm working with here doesn't apply to Resources—which are of course always created by a specific User. The "creator" I've added is a more holistic identification (modx, extra, user [in general, not specific id]). In any case, none of the other MODX objects save an identifier in the db as to who or what process created it.

theboxer:
@smg6511 How about you just disable the sort on the `creator` column? We can discuss different approaches later on, that could potentially replace the content of the column as defined in this PR and introduce the sorting/filtering in it as well.

smg6511:
> How about you just disable the sort on the `creator` column?

Yeah, I'll go ahead and do that and remove the sorting-related php. Will ping you when ready...

theboxer:
@smg6511 is this ready for re-review?

smg6511:
@theboxer - Not quite yet, but almost. I had found some details that needed tending to in order to make this change complete in the areas where it's applied so far. Work on those is just about done. 

In working with the more complicated Policies and Policy Templates areas, I've found there's a bit more to do that might justify a second PR (or a bit more time on this one if we choose to wait and submit everything together).

smg6511:
Ok guys, I've concluded that I should pull this apart a little bit, as there are a few distinct issues I've found while working through this PR:
1. The original and main problem — The way the permissions are reflected in manager grids (both in the contextual and gear menus as well as the visual separation of modx (built-in) vs user vs Extras records)
2. The protected, built-in records' names and descriptions should support translation in a better way. The current option of using a Lexicon key in these fields is not great IMO; it's unintuitive and you always have the key showing on the main editing pages (and windows) of objects you can set up in this way. I'm not looking to address this overall, but just for the protected records. In next-gen MODX, perhaps we should consider _separate_ translation key fields (made available in the UI in a tasteful way) for the purpose of user-customizable i13n field naming.
3. The Files [Sources] and Resources [Context] trees, when translation of built-ins is incorporated, should reflect the translated names and descriptions (shown in the tooltips)

I started going down the rabbit hole with fixing all these at once, but realize it's going to be easier to get changes through if each is handled on its own. It's actually good that I dove into them all at once (even as frustrating as it became), as the solution for # 3 informed a change that would be needed for how I was handling # 2.

Anyway, I'm going to submit a new PR, just targeting # 1 and handle the other two items separately...

rthrash:
@smg6511 did you want to close this one in favor of your newer PR on the same areas?

smg6511:
@rthrash - Although this PR is obsolete, I thought it might be helpful to keep it open for reference ... at least until the new PR gets (hopefully) merged.

modxcommunity:

https://community.modx.com/t/help-wanted-testing-a-pr-for-upcoming-revo-release/8528/1

smg6511:
Replacement PR #16653 merged.

PR #15921, opened 2021-11-28
Fix setStaticElementPath method (port of #15912)

Jako:
### What does it do?
Use an Ajax request to retrieve the category name instead of setTimeout to bypass a not set lastSelectionText. Use the select listener instead of a change listener.

### Why is it needed?
The previous implementation used a setTimeout to bypass a not set lastSelectionText. This timeout did not work on several installations with nested categories or a little more categories (~40 in my test) without increasing the timeout. Now the category name is retrieved with an Ajax request and the lastSelectionText and the timeout is no longer needed.

This solves the following issue: A static element is created with a path containing the category. If this static element is edited again, the category in the path is missing. If the static element is saved without temporary modifying the element name, the category path is removed without a notice.

### How to test
Setup an automatic static elements workflow. Create a nested category. Assign this category to a chunk and save that chunk. Reload the manager and look for the path of the static file. Before the category was missing in the path, after the patch it contains the category.

### Related issue(s)/PR(s)
#15912

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -354,13 +354,30 @@ Ext.extend(MODx,Ext.Component,{
         }
 
         if (MODx.config["static_elements_automate_" + typePlural] == 1) {
-            if (Ext.getCmp("modx-" + type + "-category").getValue() > 0) {
-                category = Ext.getCmp("modx-" + type + "-category").lastSelectionText;
+            category = Ext.getCmp("modx-" + type + "-category").getValue();
+            if (category > 0) {
+                Ext.Ajax.request({
+                    url: MODx.config.connector_url,
+                    params: {
+                        action: 'Element/Category/GetList',
+                        id: category,
+                        limit: 0,
+                    },
+                    success: function (response) {
+                        var data = Ext.decode(response.responseText);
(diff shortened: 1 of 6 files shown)


manager/assets/modext/core/modx.js
```diff
@@ -347,13 +347,30 @@ Ext.extend(MODx,Ext.Component,{
...
+                Ext.Ajax.request({
+                    url: MODx.config.connector_url,
+                    params: {
+                        action: 'Element/Category/GetList',
+                        id: category,
+                        limit: 0,
+                    },
+                    success: function (response) {
+                        var data = Ext.decode(response.responseText);
+                        categoryText = (data && data.success && data.results) ? data.results[0].name : '';
```
GulomovCreative, line 368:
```suggestion
                        categoryText = data && data.success && data.results ? data.results[0].name : '';
```

I don't see the point of grouping the left operand

GulomovCreative (changes requested):

manager/assets/modext/core/modx.js
Jako: It's easier to see the condition part of the ternary, if it is surrounded by braces.

theboxer (approved):

Issue #14598, opened 2019-06-05
Category IDs are not consistent


pepebe:
### Summary
The numbers show on the right side of category names listed under templates, tvs, chunks, snippets and plugins are not consistent with the category ids stored in the database

### Step to reproduce
Compare categories listed in the elements tree with the general category list.

#### Category List at bottom of elements:
<img width="567" alt="Screenshot 2019-06-05 10 05 38" src="https://user-images.githubusercontent.com/667315/58940372-9b6e4b00-8779-11e9-902e-8931daebee6e.png">

Example: Schneidscheiben (20) (identical to the value in the database)

#### TV List showing TVs ordered by category
<img width="568" alt="Screenshot 2019-06-05 10 05 21" src="https://user-images.githubusercontent.com/667315/58940373-9c06e180-8779-11e9-9284-79274393aa0e.png">

Example: Schneidscheiben (15) and Produkte doesn't even has an ID

### Expected behavior
The id shown should be euqal to the id used in the database.

### Environment
I guess this effects all versions, tested on 2.7.1

Ibochkarev:
@pepebe The value in brackets next to the TV indicates how many TV parameters are in the folder
![image](https://user-images.githubusercontent.com/2138260/58941513-c8345980-879d-11e9-9f68-26fdb8c9966d.png)

Ruslan-Aleev:
Although @pepebe is right, visually it is confusing, because for the remaining elements in brackets is specified id.
Perhaps it makes sense to **change the brackets for category**, for example:

![cat](https://user-images.githubusercontent.com/12523676/58942643-345b9100-878f-11e9-86f5-82d108552adf.png)

![cat_2](https://user-images.githubusercontent.com/12523676/58942972-d9766980-878f-11e9-9c52-83ff248aa834.png)

Jako:
To make it more obvious we could use `[2 TVs]`, `[2 Elements]`, `[2 Children]`.

pepebe:
Its the number of elements? Wow. I didn't even think about this.

From a consistency point of view I'd expect the number in brackets to be the elements id.

To be honest, I don't care how many elements are inside a category. If anyone does, please give a real world example that explains why this is important information.

Ruslan-Aleev:
@pepebe Maybe to find out how the list will unfold? The information is not very useful, but let it be.
And if you want to know exactly the id categories - they are listed in the list of categories in brackets.

pepebe:
Today I have learned the following things:

[...]

PR #15922, opened 2021-11-30
Added id / names of elements in context menu

Ruslan-Aleev:
### What does it do?
Added id / names of elements in tree. Now it takes less clicks to find out the key / id of an element in the tree + uniformity in the interface.

1) Added a source ID to context menu.

Before:
src_tree_1

After:
source_tree_id (1)

2) Added names of elements (directory and file) to their context menu. Now it is clear for which directory / file the context menu is called.

Before:
dir_tree_1
file_tree_1

After:
dir_tree_2
file_tree_2

### Why is it needed?
UX improvements

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2061,7 +2061,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
 
         $menu = [];
         $menu[] = [
-            'text' => '<b>' . $dirName . '</b>',
+            'text' => '<b>' . htmlentities($dirName) . '</b>',
             'handler' => '',
             'header' => true,
         ];
@@ -2138,7 +2138,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
 
         $menu = [];
         $menu[] = [
-            'text' => '<b>' . $fileName . '</b>',
+            'text' => '<b>' . htmlentities($fileName) . '</b>',
             'handler' => '',
             'header' => true,
         ];
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -418,7 +418,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
                         $directories[$file_name]['visibility'] = $visibility;
                     }
                     $directories[$file_name]['menu'] = [
-                        'items' => $this->getListDirContextMenu(),
+                        'items' => $this->getListDirContextMenu($file_name),
                     ];
 
                 }
@@ -2053,13 +2053,19 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
      *
      * @return array
      */
-    protected function getListDirContextMenu()
+    protected function getListDirContextMenu($dirName)
     {
         $canSave = $this->checkPolicy('save');
         $canRemove = $this->checkPolicy('remove');
         $canCreate = $this->checkPolicy('create');
 
         $menu = [];
+        $menu[] = [
+            'text' => '<b>' . $dirName . '</b>',
+            'handler' => '',
+            'header' => true,
+        ];
+        $menu[] = '-';
         $menu[] = [
             'text' => $this->xpdo->lexicon('directory_refresh'),
             'handler' => 'this.refreshActiveNode',
@@ -2122,6 +2128,8 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
      */
     protected function getListFileContextMenu($path, $editable = true, $data = [])
     {
+        $fileName = basename($path);
+
         $canSave = $this->checkPolicy('save');
         $canRemove = $this->checkPolicy('remove');
         $canView = $this->checkPolicy('view');
@@ -2129,6 +2137,12 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             (empty($data['visibility']) || $data['visibility'] === Visibility::PUBLIC);
 
         $menu = [];
+        $menu[] = [
+            'text' => '<b>' . $fileName . '</b>',
+            'handler' => '',
+            'header' => true,
+        ];
+        $menu[] = '-';
         if ($this->hasPermission('file_update') && $canSave) {
             if ($editable) {
                 $menu[] = [
--- a/manager/assets/modext/widgets/system/modx.tree.directory.js
+++ b/manager/assets/modext/widgets/system/modx.tree.directory.js
@@ -122,6 +122,12 @@ Ext.extend(MODx.tree.Directory,MODx.tree.Tree,{
      */
     ,getRootMenu: function(node) {
         var menu = [];
+        menu.push({
+            text: '<b>'+node.attributes.text+' ('+this.config.source+')</b>'
+            ,handler: function() {return false;}
+            ,header: true
+        });
+        menu.push('-');
         if (MODx.perm.directory_create) {
             menu.push({
                 text: _('file_folder_create')


Jako:
I prefer to not show the context key and the source id in the tree. Maybe it can be switched on by option. For non tech users this is too much information.

Ruslan-Aleev:
Key and id do not complicate the interface, for resources / resource groups, etc. id is shown, let it be there, as it seems to me.
And making a separate setting is clearly unnecessary.

Jako:
For non tech users even the resource Id is too much information. But this information can be used for creating internal links. The context key and the source id can't be used for anything by them.

Ruslan-Aleev:
This is understandable, but I do not see this as a problem. What happens to non-technical users if they see a key or id in tree? Such users will pass out from panic or what? =) And for technical users, the information is useful.

> The context key and the source id can't be used for anything by them.

For context, it is planned to use it through a key, there is such a proposal on github.

Mark-H:
I'm with Thomas, the context key and media source IDs are unnecessary, and would only clutter an already pretty full interface. I would however not suggest a setting to toggle it, just removing it entirely. 

- If the context key is important for editors to know, you can edit your context to add it to the name. 
- Any selection of a media source should not rely on the ID; there are dropdowns for that in the media browser, settings, etc.

In general I think it's better UX to avoid reliance on knowing an ID or other internal identification of an object. If there are places the keys/IDs are needed, those things would be better addressed there. 

I do like the addition of the file/directory to the context menu.

JoshuaLuckers:
I'm actually in favor of showing the context key. To me, as a developer, it's rather useful (same for showing the other primary key values) to have it visible at a glance.

> For non tech users this is too much information.
> In general I think it's better UX to avoid reliance on knowing an ID or other internal identification of an object.

For contexts and Media Sources I agree, the context key and Media Source ID don't need such a prominent place in the UI.

If the context key and Media Source ID are visible in the context menu for these objects it would already be an improvement for me.

Ruslan-Aleev:
I redid the PR, check

manager/assets/modext/widgets/system/modx.tree.directory.js
```diff
@@ -121,7 +121,15 @@ Ext.extend(MODx.tree.Directory,MODx.tree.Tree,{
      * @returns {Array}
      */
     ,getRootMenu: function(node) {
+        var a = node.attributes;
```
mishanthrop: The variable is used only once

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2053,13 +2053,19 @@ protected function getExtJSDirClasses()
...
-    protected function getListDirContextMenu()
+    protected function getListDirContextMenu($dirName)
     {
         $canSave = $this->checkPolicy('save');
         $canRemove = $this->checkPolicy('remove');
         $canCreate = $this->checkPolicy('create');
 
         $menu = [];
+        $menu[] = [
+            'text' => '<b>' . $dirName . '</b>',
```
mishanthrop: I consider that the appearance (including `font-weight`) should be controlled on the frontend and not the server.

Ruslan-Aleev: In an ideal world, yes, but in MODX the titles in context menu is done via `<b></b>`.
It is worth redoing everywhere and from scratch, but this is another task for a separate PR.

theboxer:
I'm against merging this PR.
What's the point of showing the file name above the context menu? It's highlighted in the tree + it's context menu so you kind of know for what it is.

IDs should be hidden for regular users, it's way too confusing for them. And as a developer I don't see a reason to see media source's ID all the time. (IDs are useless for referencing objects anyways).

Ruslan-Aleev:
> What's the point of showing the file name above the context menu? It's highlighted in the tree + it's context menu so you kind of know for what it is.

At least 2 reasons:
1) For consistency in the interface, for other elements (resources, TV, chunks, etc.) the name is shown, I see no reason not to show it here.
2) The file / directory is not always highlighted in the tree menu (I cannot reproduce, but sometimes it happens).
Although the situation is the opposite for directories and subdirectories. Guess which folder the context menu is called for?

folder_context

> IDs should be hidden for regular users, it's way too confusing for them. And as a developer I don't see a reason to see media source's ID all the time.

IDs are hidden, they remain only in the context menu. And I don't see how it could confuse the user. Moreover, for several years of working with MODX, not a single user has asked about resource IDs in the tree. Users either do not care, or they use it in links via [[~id]] if you explain it to them.

core/src/Revolution/Sources/modMediaSource.php
Mark-H: This might need escaping? Granted, XSS from a file/dirname seems pretty difficult...

Ruslan-Aleev: In my opinion, it is not necessary.
1) When creating a directory/file, tags are stripped from the name in MODX.
2) And the OS doesn't allow to a create directory/file with tags, even via the console.

And if do escaping, then `'text' => '<b>' . htmlspecialchars($dirName) . '</b>'` will be enough?

Mark-H: Yeah I struggle to see a practical way to exploit file names as well, hence the question marks ;) and yeah htmlspecialchars/htmlentities is how to avoid it entirely.

Mark-H:
I'm deferring this to rc2 as there are strong disagreements without a clear concensus at this stage, and this is not critical enough to warrant delaying the rc1 release.

JoshuaLuckers:
I think it's best to agree with not showing the ID's since they might be confusing to the average user, however, to me, showing the names of the selected item in the context menu is a improvement in UX since it makes it more clear as to which item the context menu is related.

Ruslan-Aleev:
> I think it's best to agree with not showing the ID's since they might be confusing to the average user...

@JoshuaLuckers  ID is shown now ONLY in the context menu. This will not confuse anyone, but is useful for developers. And you agreed with this earlier https://github.com/modxcms/revolution/pull/15922#issuecomment-983435287 :)

> If the context key and Media Source ID are visible in the context menu for these objects it would already be an improvement for me.

rthrash:
I have no issue with showing IDs in the context menu where they're actually used elsewhere (meaning my vote would be 10/10 for no IDs for media sources, context keys (or categories?)). 

For many, if not most, non-technical users, the interaction won't be via the context menu as it's a right-click thing which isn't a common way to interact with most web apps. Clicks and hovers, yes. Right-clicks, no.

I'd propose a SSOGAF vote of the interested parties to resolve the inclusion or not of IDs. Like Joshua, I'm probably an 8/10 on _including_ the name at the top of what you right-clicked on especially if it receives proper ARIA markup to provide the context of what it is you're about to act on (I have no idea what that should be).

Ruslan-Aleev:
@rthrash 

> I have no issue with showing IDs in the context menu where they're actually used elsewhere (meaning my vote would be 10/10 for no IDs for media sources, context keys (or categories?)).

There is more benefit than harm: the developer will recognize the ID faster, although the average user will be "confused". In my opinion, you underestimate users, I don’t think that they will be confused by the ID in the context menu (and for tree it wouldn’t either) :) **But it's worth voting, yes.**

By the way, there is a clear misunderstanding about categories (even for you). Categories in brackets show the number of nested elements, not ID. It will be necessary to return to this issue - see https://github.com/modxcms/revolution/issues/14598.

Ruslan-Aleev:
Guys, what's wrong with this PR? Tell me, I'll fix it, and we'll merge it, we discussed a lot, but came to nothing.

mishanthrop:
I do not experience difficulties with the current UI. Perhaps this PR will be relevant for mobile devices or for people with ataxia, tremor etc. Is a11y still trending?

Which of the three files is selected? :-)
I don't know how to reproduce, but the selection of files does sometimes work strangely.

Снимок экрана 2022-12-02 в 14 24 08

PR #15926, opened 2021-12-03
[3.x] Fix combobox not expanding if processor response is slow

### LINKED ISSUE #15924: JS Error in MODx.combo.ComboBox if processor response takes too long
### Summary
This occurs when a MODExt combo box is set to `mode: remote`.
If a processor is slow to respond, the console shows this error:
```
Uncaught TypeError: this.pageTb is undefined
```
It references this line: 
https://github.com/modxcms/revolution/blob/2.x/manager/assets/modext/widgets/core/modx.combo.js#L114

### Step to reproduce
Add `sleep(3)` to your processor to replicate the issue.

### Observed behavior
The user will see the first row from the processor response in the combo but nothing else.

### Expected behavior
There should not be an error if the response is slow.

### Related issue(s)/PR(s)
n/a

### Environment
MODX 2.8.3 - I've yet to test it with 3.x but will do so tonight/tomorrow.

muzzwood:
### What does it do?
Prevents checking if the pagination panel is hidden if it's undefined.

### Why is it needed?
Currently remote comboboxes won't expand properly if the ajax response is too slow.

### How to test
Click on a combobox that is set to `mode: remote`.
Slow the processor down for testing by adding `sleep(3)` somewhere.

### Related issue(s)/PR(s)
Closes #15924


muzzwood:
If this gets approved, I'll do the port for 2.x too.

Jako (approved):

Ibochkarev (approved):

Jako:
This looks suitable.

But if I remember right, the answer to the initial AJAX request causes the issue. This request is done with the current value of the combo (mostly in the 'id' property). If the combo processor then tries to create or returns the unfiltered list, which can take longer than a few seconds, the pagination is hidden. The combo processor has to return a single element list with the initial value and the total number in the initial AJAX request, when the combo value is filled. Then the pagination is shown opening the combo. When the combo is opened, the list is filled by a second paginated AJAX request.

It will be less resource consuming, if the combo processor is restricted that way.

muzzwood:
It looks like this is where it was introduced:
https://github.com/modxcms/revolution/commit/ef798e23abebfcef9d0d6e6b5a57326b84cfc3e6

Mark-H (approved):

theboxer (approved):

Issue #15930, opened 2021-12-05
static_resoures lexicon key doesn't exist


sergant210:
### Summary
The lexicon key "static_resoures" for the area combo box is missing.

![static resources](https://user-images.githubusercontent.com/4679725/144732712-98c1dd7d-6c8c-492b-a48b-97f33f00256f.png)

### Environment
MODX 3 Beta 2.

Ibochkarev:
Fixed #15931

PR #15931, opened 2021-12-05
Added missing lexicon key

Ibochkarev:
### What does it do?
Adds a missing lexicon

### Why is it needed?
изображение

### Related issue(s)/PR(s)
#15930

### DIFF
--- a/core/lexicon/bg/setting.inc.php
+++ b/core/lexicon/bg/setting.inc.php
@@ -21,7 +21,6 @@ $_lang['area_manager'] = 'Back-end Manager';
 $_lang['area_phpthumb'] = 'phpThumb';
 $_lang['area_proxy'] = 'Proxy';
 $_lang['area_session'] = 'Session and Cookie';
-$_lang['area_static_resources'] = 'Static Resources';
 $_lang['area_static_elements'] = 'Static Elements';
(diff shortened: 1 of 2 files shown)


opengeek (changes requested):
We should not be modifying any translations by hand. If it is not there, then something is wrong upstream via CrowdIn and needs to be addressed.

Jako:
@Ibochkarev The string is just added in the wrong lexicon.

Ibochkarev:
> We should not be modifying any translations by hand. If it is not there, then something is wrong upstream via CrowdIn and needs to be addressed.

This section was created in this PR https://github.com/modxcms/revolution/pull/15885, but the lexicon for the name of the system settings section is omitted.

Ibochkarev:
> @Ibochkarev The string is just added in the wrong lexicon.

In which file should this lexicon be placed?

opengeek:
The lexicon needs to be added to the en/ lexicon, not the bg/. The en/ lexicons are the source and are the only ones that should be edited directly.

opengeek (approved):

Ibochkarev:
> The lexicon needs to be added to the en/ lexicon, not the bg/. The en/ lexicons are the source and are the only ones that should be edited directly.

Sorry, I saw a mistake

Jako (approved):

Issue #15928, opened 2021-12-04
Logout all users action doesn't work with custom session handlers


sergant210:
### Summary
If you have specified your own session handlers, you will not be able to flush all user sessions.

### Step to reproduce
1. Create a new session handler which extends the modSessionHandler class.
2. Specify it in the system setting "session_handler_class".
3. Try to logout all users in the admin panel (Manage/Logout all users).

### Observed behavior
![flush_sessions](https://user-images.githubusercontent.com/4679725/144698268-8f356a99-2bdd-4e29-802c-ce84971a45a4.png)

There are 2 problems
1. Corresponding lexicon topic is not loaded.
2. Sessions are not deleted.

The problem is in the very strange [check](https://github.com/modxcms/revolution/blob/1014ed8b4ed8a53f6fa46dda97bde31913e2fe9d/core/model/modx/processors/security/flush.class.php#L24). So we have got a system setting for extending the handler class, but we cannot change it.

### Environment
MODX 2.x, MODX 3 Beta 2.

Ibochkarev:
Related https://github.com/modxcms/revolution/issues/15510

PR #15934, opened 2021-12-07
Fix bugs with extending the session handler class

sergant210:
### What does it do?
1. Fixes the bugs from issue #15928.
2. Refactoring inititialization of the session handler.

### How to test
See the issue.

### Related issue(s)/PR(s)
#15928.

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -2746,25 +2746,14 @@ class modX extends xPDO {
         $contextKey= $this->context instanceof modContext ? $this->context->get('key') : null;
         if ($this->getOption('session_enabled', $options, true) || isset($_GET['preview'])) {
             if (!in_array($this->getSessionState(), [modX::SESSION_STATE_INITIALIZED, modX::SESSION_STATE_EXTERNAL, modX::SESSION_STATE_UNAVAILABLE], true)) {
-                $sh = false;
-                if ($sessionHandlerClass = $this->getOption('session_handler_class', $options)) {
-                    if ($shClass = $this->loadClass($sessionHandlerClass, '', false, true)) {
-                        if ($sh = new $shClass($this)) {
-                            session_set_save_handler(
-                                [& $sh, 'open'],
-                                [& $sh, 'close'],
-                                [& $sh, 'read'],
-                                [& $sh, 'write'],
-                                [& $sh, 'destroy'],
-                                [& $sh, 'gc']
-                            );
-                        }
+                $sessionHandlerClass = $this->getOption('session_handler_class', $options);
+                if (is_string($sessionHandlerClass) && !empty($sessionHandlerClass) && class_exists($sessionHandlerClass)) {
+                    $ch = new $sessionHandlerClass($this);
+                    if ($ch instanceof \SessionHandlerInterface) {
+                        $this->services->add('session_handler', $ch);
                     }
                 }
-                if (
-                    (is_string($sessionHandlerClass) && !$sh instanceof $sessionHandlerClass) ||
-                    !is_string($sessionHandlerClass)
-                ) {
+                if (!$this->services->has('session_handler')) {
                     $sessionSavePath = $this->getOption('session_save_path', $options);
                     if ($sessionSavePath && is_writable($sessionSavePath)) {
                         session_save_path($sessionSavePath);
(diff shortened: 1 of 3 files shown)


core/src/Revolution/modX.php
```diff
@@ -2746,25 +2746,14 @@ protected function _initSession($options = null) {
         $contextKey= $this->context instanceof modContext ? $this->context->get('key') : null;
         if ($this->getOption('session_enabled', $options, true) || isset($_GET['preview'])) {
             if (!in_array($this->getSessionState(), [modX::SESSION_STATE_INITIALIZED, modX::SESSION_STATE_EXTERNAL, modX::SESSION_STATE_UNAVAILABLE], true)) {
-                $sh = false;
-                if ($sessionHandlerClass = $this->getOption('session_handler_class', $options)) {
-                    if ($shClass = $this->loadClass($sessionHandlerClass, '', false, true)) {
-                        if ($sh = new $shClass($this)) {
-                            session_set_save_handler(
```
Mark-H, line 2753: This line is removed but not added back anywhere that I can see - so it seems with this patch we're not overriding the session handler anymore

core/src/Revolution/modX.php
```diff
@@ -2746,25 +2746,14 @@ protected function _initSession($options = null) {
...
+                    $ch = new $sessionHandlerClass($this);
+                    if ($ch instanceof \SessionHandlerInterface) {
+                        $this->services->add('session_handler', $ch);
                     }
                 }
-                if (
-                    (is_string($sessionHandlerClass) && !$sh instanceof $sessionHandlerClass) ||
-                    !is_string($sessionHandlerClass)
-                ) {
+                if (!$this->services->has('session_handler')) {
```
opengeek, line 2757: I'm not sure this is correct. You do not need a custom session handler to be able to set the session_save_path. In fact, this is meant to be used with the default PHP session handling to customize the location of the session files.

core/src/Revolution/modX.php
```diff
@@ -2746,25 +2746,14 @@ protected function _initSession($options = null) {
...
-                        if ($sh = new $shClass($this)) {
-                            session_set_save_handler(
-                                [& $sh, 'open'],
-                                [& $sh, 'close'],
-                                [& $sh, 'read'],
-                                [& $sh, 'write'],
-                                [& $sh, 'destroy'],
-                                [& $sh, 'gc']
-                            );
-                        }
```
opengeek, line 2761: Where is the session handler being registered if this is removed?

core/src/Revolution/modX.php
sergant210: Look at line 2749.

Mark-H: I'm not seeing another instance of `session_set_save_handler` anywhere in this diff.

sergant210: Oops, I forgot to move it from my test site. Thanks.

core/src/Revolution/modX.php
sergant210: It's my fault. I forgot to add this `session_set_save_handler($sh);`.

core/src/Revolution/modX.php
sergant210: Can you show me an example to make this condition true?
```
if (
    (is_string($sessionHandlerClass) && !$sh instanceof $sessionHandlerClass) ||
     !is_string($sessionHandlerClass)
)
```

opengeek: If it is not set to a modSessionHandler class or it is empty so it uses what is defined in PHP's ini.

sergant210: 1. There is a deprecated loadClass method.
2. `$sh` object will always be an instance of `$sessionHandlerClass`, so this condition will be true only if `$sessionHandlerClass` is not a string or "".
3. I've moved this condition to line 2750.

core/src/Revolution/modX.php
sergant210: Fixed.

JoshuaLuckers:
@cla-bot check

rthrash:

https://community.modx.com/t/error-flush-sessions-not-supported-when-trying-to-logout-all-users-in-modx-3/6361/2

rthrash:

https://community.modx.com/t/error-flush-sessions-not-supported-when-trying-to-logout-all-users-in-modx-3/6361/7

Mark-H:
Closing in favour of #16522 - please see my explanation there.

PR #15185, opened 2020-08-06
Added remembering filter properties of fcp grid in browser url

Ruslan-Aleev:
### What does it do?
- Added the selected filter properties in the browser url.

fcp_search

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15184
https://github.com/modxcms/revolution/pull/15183
https://github.com/modxcms/revolution/pull/15182
https://github.com/modxcms/revolution/pull/15181
https://github.com/modxcms/revolution/pull/15115
https://github.com/modxcms/revolution/issues/14086

PR #15181, opened 2020-08-06
Remember filter properties of packages grid in browser url 

Ruslan-Aleev:
### What does it do?
Added the selected filter properties in the browser url.

packages_search

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15115
https://github.com/modxcms/revolution/issues/14086

Issue #14086, opened 2018-09-12
Add additional parameters to the url to simplify navigation in the manager panel


Ruslan-Aleev:
From a recent time Google Chrome (maybe other browsers) to save RAM clears tabs in the background, i.e. when you open the background tab - the tab updates content completely. MODX navigation is built on tabs and filters and happens that tabs and filters are reset if you work with several tabs.

It's worth adding parameters to the URL to save the filtering and the active tab, for example:
Settings:
![filter_settings](https://user-images.githubusercontent.com/12523676/45425169-a8a64f80-b6a9-11e8-99f3-02b836f029e8.png)

Resources:
![filter_resource](https://user-images.githubusercontent.com/12523676/45425168-a80db900-b6a9-11e8-9734-9468ae337761.png)

Lixicons:
![filter_lexicon](https://user-images.githubusercontent.com/12523676/45425167-a80db900-b6a9-11e8-8bda-bc2995572d94.png)

The parameters will allow to share specific links with users and improve the documentation for MODX - you can specify a link to the section, for example, "Friendly URL" in "System Settings":
**site.com/manager/?a=system/settings&filter-area=Friendly%20URL**

p.s. Perhaps there were discussions about the parameters, I did not find it in the issues.

mrhaw:
For reference:
[Action List](https://docs.modx.com/revolution/2.x/developing-in-modx/advanced-development/custom-manager-pages/actions-and-menus/action-list)
 [modAction](https://forums.modx.com/thread/89849/getting-the-url-for-a-cmp-or-other-manager-controller#dis-post-535828)

JoshuaLuckers:
Great proposal. Especially the one where it remembers which tab you are on. 

It would be a nice to have this in MODX 3.

GulomovCreative:
Addition: Limit of per page

Ruslan-Aleev:
@theboxer Do you have any idea how to implement this? At least for the "System Settings".

Ibochkarev:
Added #15086
@modxbot close

Ibochkarev:
@modxbot reopen

Ibochkarev:
Done.

Ruslan-Aleev:
Some sections have not been completed yet :) Let the issue be open

smg6511:
[...]

PR #15935, opened 2021-12-08
Added remembering filter properties of plugin events grid in browser url

Ruslan-Aleev:
### What does it do?
- Added the selected filter properties in the browser url.

plugin_events

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15186
https://github.com/modxcms/revolution/pull/15185
https://github.com/modxcms/revolution/pull/15184
https://github.com/modxcms/revolution/pull/15183
https://github.com/modxcms/revolution/pull/15182
https://github.com/modxcms/revolution/pull/15181
https://github.com/modxcms/revolution/pull/15115
https://github.com/modxcms/revolution/issues/14086

### DIFF
--- a/manager/assets/modext/widgets/element/modx.grid.plugin.event.js
+++ b/manager/assets/modext/widgets/element/modx.grid.plugin.event.js
@@ -23,6 +23,8 @@ MODx.grid.PluginEvent = function(config) {
         ,baseParams: {
             action: 'Element/Plugin/Event/GetList'
             ,plugin: config.plugin
+            ,group: MODx.request.group ? decodeURIComponent(MODx.request.group) : ''
+            ,query: MODx.request.query || ''
             ,limit: 0
         }
         ,saveParams: {
@@ -82,25 +84,51 @@ MODx.grid.PluginEvent = function(config) {
             ,id: 'modx-plugin-event-filter-group'
             ,itemId: 'group'
             ,emptyText: _('group')+'...'
+            ,value: MODx.request.group ? decodeURIComponent(MODx.request.group) : ''
             ,width: 200
             ,listeners: {
-                'select': {fn:this.filterGroup,scope:this}
+                'select': {
+                    fn: function (cb, rec, ri) {
+                        if (!MODx.request.query) {
+                            this.filterByGroup(cb, rec, ri);
+                        }
+                    }
+                    ,scope: this
+                }
             }
         },{
             xtype: 'textfield'
-            ,name: 'search'
-            ,id: 'modx-plugin-event-search'
+            ,name: 'query'
+            ,id: 'modx-plugin-event-query'
             ,cls: 'x-form-filter'
             ,emptyText: _('search')
+            ,value: MODx.request.query || ''
             ,listeners: {
-                'change': {fn: this.search, scope: this}
-                ,'render': {fn: function(cmp) {
-                    new Ext.KeyMap(cmp.getEl(), {
-                        key: Ext.EventObject.ENTER
-                        ,fn: this.blur
-                        ,scope: cmp
-                    });
-                },scope:this}
+                'change': {
+                    fn: function (cb, rec, ri) {
+                        this.eventSearch(cb, rec, ri);
+                    }
+                    ,scope: this
+                }
+                ,'afterrender': {
+                    fn: function (cb) {
+                        if (MODx.request.query) {
+                            this.eventSearch(cb, cb.value);
+                            MODx.request.query = '';
+                        }
+                    }
+                    ,scope: this
+                }
+                ,'render': {
+                    fn: function (cmp) {
+                        new Ext.KeyMap(cmp.getEl(), {
+                            key: Ext.EventObject.ENTER
+                            ,fn: this.blur
+                            ,scope: cmp
+                        });
+                    }
+                    ,scope: this
+                }
             }
         },{
             xtype: 'button'
@@ -122,23 +150,27 @@ MODx.grid.PluginEvent = function(config) {
     this.addEvents('updateEvent');
 };
 Ext.extend(MODx.grid.PluginEvent,MODx.grid.Grid,{
-    search: function(tf,newValue) {
+    eventSearch: function (tf, newValue) {
         var nv = newValue || tf;
-        this.getStore().baseParams.query = Ext.isEmpty(nv) || Ext.isObject(nv) ? '' : nv;
-        this.getStore().load();
-        return true;
+        var s = this.getStore();
+        s.baseParams.query = Ext.isEmpty(nv) || Ext.isObject(nv) ? '' : nv;
+        s.load();
+        this.replaceState();
     }
-    ,filterGroup: function (cb,nv,ov) {
-        this.getStore().baseParams.group = Ext.isEmpty(nv) || Ext.isObject(nv) ? cb.getValue() : nv;
-        this.getStore().load();
-        return true;
+    ,filterByGroup: function (cb, newValue, ov) {
+        var nv = Ext.isEmpty(newValue) || Ext.isObject(newValue) ? cb.getValue() : newValue;
+        var s = this.getStore();
+        s.baseParams.group = nv;
+        s.load();
+        this.replaceState();
     }
-    ,clearFilter: function() {
-        delete this.getStore().baseParams.query;
-        delete this.getStore().baseParams.group;
-        Ext.getCmp('modx-plugin-event-search').reset();
-        Ext.getCmp('modx-plugin-event-filter-group').reset();
-        this.getStore().load();
+    ,clearFilter: function () {
+        var s = this.getStore();
+        s.baseParams.group = s.baseParams.query = '';
+        Ext.getCmp('modx-plugin-event-filter-group').setValue('');
+        Ext.getCmp('modx-plugin-event-query').setValue('');
+        s.load();
+        this.replaceState();
     }
     ,updateEvent: function(btn,e) {
         var r = this.menu.record;
--- a/manager/assets/modext/widgets/element/modx.panel.plugin.js
+++ b/manager/assets/modext/widgets/element/modx.panel.plugin.js
@@ -16,7 +16,7 @@ MODx.panel.Plugin = function(config) {
             action: 'Element/Plugin/Get'
         }
         ,id: 'modx-panel-plugin'
-		,cls: 'container form-with-labels'
+        ,cls: 'container form-with-labels'
         ,class_key: 'MODX\\Revolution\\modPlugin'
         ,plugin: ''
         ,bodyStyle: ''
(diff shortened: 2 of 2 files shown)


JoshuaLuckers:
In this case we need to remember the selected tab as well.

Ruslan-Aleev:
I agree, I would add `#tab-id` to url for any LAST clicked tab (or collect an array of active tabs...), but this needs to be done in a separate js-component for MODX (which would work for any panels), and not just for the current panel.  And if there is `#tab-id` when loading the URL, then the tab would become active.

But I don't know how to do it right. Maybe @smg6511 will help. It would be a very useful improvement, see https://github.com/modxcms/revolution/issues/14086.

smg6511:
OK, I'll put this on my plate to take a look at.

manager/assets/modext/widgets/element/modx.grid.plugin.event.js
```diff
@@ -23,6 +23,8 @@ MODx.grid.PluginEvent = function(config) {
         ,baseParams: {
             action: 'Element/Plugin/Event/GetList'
             ,plugin: config.plugin
+            ,group: MODx.request.group ? decodeURIComponent(MODx.request.group) : ''
+            ,query: MODx.request.query ? MODx.request.query : ''
```
GulomovCreative:
```suggestion
            ,query: MODx.request.query || ''
```

manager/assets/modext/widgets/element/modx.grid.plugin.event.js
```diff
@@ -82,25 +84,51 @@ MODx.grid.PluginEvent = function(config) {
...
+                    ,scope: this
+                }
             }
         },{
             xtype: 'textfield'
             ,name: 'search'
             ,id: 'modx-plugin-event-search'
             ,cls: 'x-form-filter'
             ,emptyText: _('search_ellipsis')
+            ,value: MODx.request.query ? MODx.request.query : ''
```
GulomovCreative:
```suggestion
            ,value: MODx.request.query || ''
```

manager/assets/modext/widgets/element/modx.grid.plugin.event.js
```diff
@@ -82,25 +84,51 @@ MODx.grid.PluginEvent = function(config) {
...
-                    });
-                },scope:this}
+                'change': {
+                    fn: function (cb, rec, ri) {
+                        this.eventSearch(cb, rec, ri);
+                    }
+                    ,scope: this
+                }
+                ,'afterrender': {
+                    fn: function (cb){
```
GulomovCreative:
```suggestion
                    fn: function (cb) {
```

manager/assets/modext/widgets/element/modx.grid.plugin.event.js
```diff
@@ -82,25 +84,51 @@ MODx.grid.PluginEvent = function(config) {
...
+                        if (MODx.request.query) {
+                            this.eventSearch(cb, cb.value);
+                            MODx.request.query = '';
+                        }
+                    }
+                    ,scope: this
+                }
+                ,'render': {
+                    fn: function(cmp) {
```
GulomovCreative:
```suggestion
                    fn: function (cmp) {
```

manager/assets/modext/widgets/element/modx.grid.plugin.event.js
```diff
@@ -122,23 +150,28 @@ MODx.grid.PluginEvent = function(config) {
     this.addEvents('updateEvent');
 };
 Ext.extend(MODx.grid.PluginEvent,MODx.grid.Grid,{
-    search: function(tf,newValue) {
+    eventSearch: function(tf,newValue) {
```
GulomovCreative:
```suggestion
    eventSearch: function (tf,newValue) {
```

manager/assets/modext/widgets/element/modx.grid.plugin.event.js
```diff
@@ -122,23 +150,28 @@ MODx.grid.PluginEvent = function(config) {
...
     }
     ,clearFilter: function() {
-        delete this.getStore().baseParams.query;
-        delete this.getStore().baseParams.group;
-        Ext.getCmp('modx-plugin-event-search').reset();
-        Ext.getCmp('modx-plugin-event-filter-group').reset();
-        this.getStore().load();
+        var s = this.getStore();
+        s.baseParams.group = '';
+        s.baseParams.query = '';
```
GulomovCreative:
```suggestion
        s.baseParams.group = s.baseParams.query = '';
```

manager/assets/modext/widgets/element/modx.panel.plugin.js
```diff
@@ -256,19 +256,19 @@ MODx.panel.Plugin = function(config) {
...
-					,fieldLabel: _('plugin_code')
-					,name: 'plugincode'
-					,id: 'modx-plugin-plugincode'
-					,anchor: '100%'
-					,height: 400
-					,value: config.record.plugincode || "<?php\n"
+                xtype: 'panel'
+                ,border: false
+                ,layout: 'form'
+                ,cls:'main-wrapper'
```
GulomovCreative:
```suggestion
                ,cls: 'main-wrapper'
```

manager/assets/modext/widgets/element/modx.panel.plugin.js
```diff
@@ -280,7 +280,8 @@ MODx.panel.Plugin = function(config) {
                 ,xtype: 'modx-description'
             },{
                 xtype: 'modx-grid-plugin-event'
-				,cls:'main-wrapper'
+                ,cls:'main-wrapper'
```
GulomovCreative:
```suggestion
                ,cls: 'main-wrapper'
```

GulomovCreative (changes requested):

smg6511:
@Ruslan-Aleev - OK, I have the solution for activating the correct tab. What's the best way to relay that addition to you for this PR?

Jako:
@smg6511 I won't add that with an anchor in the url, since there can be more than one tab group available i.e. two nested horizontal and vertical tabs groups.

Ruslan-Aleev:
> @Ruslan-Aleev - OK, I have the solution for activating the correct tab. What's the best way to relay that addition to you for this PR?

@smg6511 I mean GLOBAL improvement, not only within this PR. Somehow to keep track of active tabs, and take them into account when reloading the page.

Ruslan-Aleev:
> @smg6511 I won't add that with an anchor in the url, since there can be more than one tab group available i.e. two nested horizontal and vertical tabs groups.

I only meant the last active tab that the user clicked. Because this is most often in the grid, then at first glance this would be enough. Or in general, make this behavior only for grids.

smg6511:
@Ruslan-Aleev @Jako - The functionality to remember the state of a page is already there (by setting and few params in a tab panel’s config; _see_ the `modx.panel.source.js` file [lines 25–32](https://github.com/modxcms/revolution/blob/b3c5d74cd4ea53645e3d26d56a93d11c0c94ee26/manager/assets/modext/widgets/source/modx.panel.source.js#L25) for an example of how this is already implemented). If that's to be done, you'd probably want to give users the choice via a system setting; there are many page types (resource for example) where I'd find it somewhat annoying to land on a tab other than the primary one when opening any given document. To make this global, it would probably be best done within the `MODx.Tabs` class. I can play around with it.

But we’re talking about two different things: For the feature Ruslan’s introduced (of being able to pull up a grid panel’s filter “state” based on the URL’s query params), you need to do it in a way such as I've done for any page where the tab is not the primary one. 

BTW, I'm assuming you saw the addition I made to the plugin panel class within the setup method (I wasn't sure of how to properly attack adding proposed changes to someone else’s PR ... guess I did it correctly?).

GulomovCreative (approved):

smg6511:
> To make this global, it would probably be best done within the `MODx.Tabs` class.

After some initial tinkering with this, it's not as simple to implement as I had thought. One immediately noticeable issue within the elements panels having a code block (chunk, snippet, plugin, template) that is transformed by a code-editing plugin (Ace, CodeMirror, etc.) is that those plugins make certain un-ideal assumptions in their setup — leading to errors when loading the panel on a tab other than the one with that code block (which is always the first one). So, all extras that transform the code field would need to be tested and updated to accommodate this feature; they'd be very small changes, but this feature would be dependent upon those changes being made.

The other factor that comes into play is how the implementation should be designed; should each individual item be stateful or should the state be based on simply the panel’s class (_e.g._, if you were editing a template and the properties tab was open before navigating to another template, any subsequent templates would open to the properties tab until leaving a template page in a different state). How fine-grained should the statefulness be? Should the grids within maintain state.

Anyway, maybe some thought and consensus on exactly what the requirements of this feature should be would be prudent before diving into the coding of it.

There's one item (that I believe another user mentioned) that I would really like to be stateful on a very specific level: The per page limit on grids.

Ruslan-Aleev:
> e.g., if you were editing a template and the properties tab was open before navigating to another template, any subsequent templates would open to the properties tab until leaving a template page in a different state.

I think that each individual page (for example, a template page) has nothing to do with the tabs state of other pages, if I understand what it is about.
Now MODX, it seems to me, keeps the active tab for each page separately. For example, when editing TV, saving on the tab and then when opening TV again - the tab retains its activity (which was when saving).

smg6511:
> Now MODX, it seems to me, keeps the active tab for each page separately.

Yes, as I've observed, _only_ the TV panel maintains the page state of individual TVs. And that's the behavior I was asking about with my templates example. Would we want the state of every individual page (elements and resources) to be remembered, replicating the current behavior of the TV panel? Could it be problematic to do that for resources on sites with a large number of pages?

We may also want/need to look into persisting page state via LocalStorage instead of cookies (Ext 3 only has a cookie provider, not surprising as Ext 3 is over a decade old), especially if we're looking to be able to maintain the state of several items.

Ruslan-Aleev:
Keeping tabs in this way is unnecessary, it is now confusing, as it seems to me (I would remove this behavior, leaving only for the tree) or maybe it is worth adding system settings, I'm not sure.
p.s. @smg6511  If you have additional information, then it is worth discussing it in the issue, because in PR it can get lost.

Ruslan-Aleev:
@JoshuaLuckers @Jako As soon as you have time, please test PR, because even in the current version it is better than what we have now.
The tab can be clicked, and writing a tab processing script for each grid (there are several of them) is wrong, as it seems to me, and this is a task for a global solution.

JoshuaLuckers:
@Ruslan-Aleev sorry for the long wait, can you resolve the conflicts?

Ruslan-Aleev:
@JoshuaLuckers Conflicts resolved, please check :)

JoshuaLuckers (approved):

Mark-H (approved):

Ruslan-Aleev:
There were conflicts in the files, I corrected them.

PR #15936, opened 2021-12-08
[3.x] Restrict tag usage and ensure proper escaping of element name, caption, and description fields

### LINKED ISSUE #15925: Incomplete escaping of characters in TV (XSS)
### Summary
In TV caption you can specify html tags that are not escaped. Maybe for other elements it is relevant too.

### Step to reproduce
Add: `<script>alert('xss');</script>` in TV caption and edit the resource with this TV.

### Environment
MODX 2.x, MODX 3.x

smg6511:
### What does it do?

Strips tags and html encodes the name, caption/label, and description fields for elements—both in the back end (processors) and front end (js forms, grids, and windows). Made separate commits for processors and js in case that makes testing easier.

### Why is it needed?

There should be more appropriate limitations on HTML tag usage within certain fields. The initial submission of this PR is highly restrictive, which I realize may need to be loosened depending on reviewer feedback.  

### How to test
 
1. Note, There are three areas covered that should be tested:
    - The main element create/edit form pages
    - The quick create/edit windows
    - The form customization grid (you'll need to create a new profile and set to test)
2. From your terminal app, run `grunt build` from within the `_build/templates/default` folder and clear your browser cache.
2. Cherry pick just the processors commit to verify tags can not be saved to the database for the _name/templatename_, and only select tags and attributes can be saved for _caption_ (TVs only) and _description_ fields for all elements (set with the 4 new system settings `elements_caption_allowedtags`, `elements_caption_allowedattr`, `elements_description_allowedtags`, and `elements_description_allowedattr`).
3. Add in the javascript commit and verify that tags are prevented in these form fields and special characters are html encoded as well.

### Related issue(s)/PR(s)
Resolves #15925


smg6511:
> That was for rendering in the resource TV panel, as in-context help/guidance. As an example, consider a "hero type" TV with a narrow height image in the description to show what the 3 options in a select/radio tv would look like.

> For the caption (again, rendering in the resource tv panel) I can also imagine adding just a help link to the caption.

> I don't think I have an immediate example at hand as the few sites I'm actively managing these days are not set up for non-technical clients, but I just want my objection that this should not be classified as a XSS vulnerability but a feature noted, before #15936 is considered. #15936 does go beyond just what is reported here and from a cursory review some of that would certainly be good to resolve, but I don't want to be the one to tell people the help text they set up for their clients is no longer possible in 3.

(Just to make this reference to this PR easier to get to, the above is from @Mark-H re this PR.)

One quick question on the matter of XSS, and this is just to get a better understanding: Is it because of the context of these input forms being only accessible via a backend interface that XSS would not be applicable?

Regardless, it seems I should consider opening it up a bit in the following ways:

1. Allow the TV captions to have links (perhaps without scripting access?)
2. Allow the TV descriptions to use either a limited set of specified tags, or all tags other than <script>

Jako:
Disallowing script tags is not enough. Each element can have several `on` attributes that can execute javascript.

smg6511:
@Jako - Yes, it goes without saying that virtually all attributes should be discarded, especially event-based ones. ;-)

Jako:
MODx.util.safeHtml does this and should be quite safe.

core/src/Revolution/Processors/Element/Create.php
```diff
@@ -50,13 +58,34 @@ public function cleanup()
...
-        $name = $this->getProperty('name');
+        $locked = (bool)$this->getProperty('locked', false);
+        $this->object->set('locked', $locked);
+
+        $elementClassName = array_pop(explode('\\', $this->classKey));
+
+        if ($elementClassName === 'modTemplateVar') {
+            if ($caption = $this->getProperty('caption', '')) {
+                $this->object->set('caption', strip_tags($caption));
+            }
```
mishanthrop: 1. It could be written shorter.

2. Why do you prefer assignment within a condition? 

`> Although the programmer's intent could have been to assign b to a and test the value of the result, this usage frequently occurs when the programmer mistakenly used the assignment operator = rather than the equality operator ==.`
(more details: https://wiki.sei.cmu.edu/confluence/display/java/EXP51-J.+Do+not+perform+assignments+in+conditional+expressions)

```suggestion
            $this->object->set('caption', strip_tags($this->getProperty('caption', '')));
```

smg6511: This is done this way because captions only apply to TVs and that property will not be present in the other elements that this processor handles.

mishanthrop: Checking a specific field (`caption`) to determine the entity class (TV in this case) is an unreliable construct. Explicit is better than implicit.
Also this does not answer the question about assignment within a condition.

smg6511: A couple things:
1. To answer your question: I often make an assignment in a condition when the variable is only needed within that condition. I think that should be self-evident. It's one less line and just as easy to understand.
2. I'm not trying to determine the entity class, I was just pointing out that `caption` is not a property of every element (it's currently only a property of TVs).

mishanthrop: Do you еven need the variable `$caption`?

smg6511:
OK all, I've made several updates in the latest commit to address the concerns raised above, as well as to make the processing more complete (on the php side) and flexible by making the tag and attribute restrictions customizable via new system settings.

core/lexicon/en/setting.inc.php
```diff
@@ -240,6 +240,18 @@
...
+$_lang['setting_elements_caption_allowedattr_desc'] = 'When adding an element caption, the html tag attribute(s) provided in this comma-separated list will be preserved. This currently only applies to template variables (TVs).';
+
+$_lang['setting_elements_caption_allowedtags'] = 'Element Captions: Allowed Tags';
+$_lang['setting_elements_caption_allowedtags_desc'] = 'When adding an element caption, the html tag(s) provided in this comma-separated list will be preserved. This currently only applies to template variables (TVs).';
+
+$_lang['setting_elements_description_allowedattr'] = 'Element Descriptions: Allowed Attributes';
+$_lang['setting_elements_description_allowedattr_desc'] = 'When adding an element description, the html tag attribute(s) provided in this comma-separated list will be preserved.';
+
+$_lang['setting_elements_description_allowedtags'] = 'Element Descriptions: Allowed Tags';
+$_lang['setting_elements_description_allowedtags_desc'] = 'When adding an element description, the html tag(s) provided in this comma-separated list will be preserved.';
```
JoshuaLuckers:
```suggestion
$_lang['setting_elements_caption_allowedattr_desc'] = 'When adding an element caption, the HTML tag attribute(s) provided in this comma-separated list will be preserved. This currently only applies to Template Variables (TVs).';

$_lang['setting_elements_caption_allowedtags'] = 'Element Captions: Allowed Tags';
$_lang['setting_elements_caption_allowedtags_desc'] = 'When adding an element caption, the HTML tag(s) provided in this comma-separated list will be preserved. This currently only applies to Template Variables (TVs).';

$_lang['setting_elements_description_allowedattr'] = 'Element Descriptions: Allowed Attributes';
$_lang['setting_elements_description_allowedattr_desc'] = 'When adding an element description, the HTML tag attribute(s) provided in this comma-separated list will be preserved.';

$_lang['setting_elements_description_allowedtags'] = 'Element Descriptions: Allowed Tags';
$_lang['setting_elements_description_allowedtags_desc'] = 'When adding an element description, the HTML tag(s) provided in this comma-separated list will be preserved.';
```

core/src/Revolution/Processors/Element/Create.php
```diff
@@ -50,14 +58,47 @@ public function cleanup()
...
-        if ($this->alreadyExists($name)) {
-            $this->addFieldError($nameField, $this->modx->lexicon($this->objectType . '_err_ae', [
-                'name' => $name,
-            ]));
+        $locked = (bool)$this->getProperty('locked', false);
+        $this->object->set('locked', $locked);
+
+        $elementClassName = array_pop(explode('\\', $this->classKey));
+
+        if ($elementClassName === 'modTemplateVar') {
```
JoshuaLuckers: Is there a reason why you're not using the fully qualified classname here because doing it like this would make it harder to maintain down the road.

core/src/Revolution/modX.php
```diff
@@ -2981,4 +2981,89 @@ public function _postProcess() {
...
     }
+
+    /**
+     * Removes unwanted tags and/or tag attributes from an HTML string
+     *
+     * @param string $htmlSource The html string to clean
+     * @param string|array $allowedTags An array or comma-separated list of tag names to allow
+     * @param string|array $allowedAttr An array or comma-separated list of tag attribute names to allow
+     */
+    public function stripHTML(string $htmlSource, $allowedTags = '', $allowedAttr = ''): string
```
JoshuaLuckers: Would it be possible to add a Unit Test for this with various test cases?

core/src/Revolution/Processors/Element/Create.php
smg6511: Simply that I was not as wise then (when I originally submitted this) as I am today. Changed to FQC here and in the Update class (commit [c6812e4](https://github.com/modxcms/revolution/pull/15936/commits/c6812e424c946107895377f17cfbcf4f0b52c0db).)

core/src/Revolution/modX.php
smg6511: Yes, I will likely need a little guidance creating the tests (at least initially) as it'll be my first time doing so. I've been wanting to get some experience with UTs so I definitely don't want to pass it off ;-) I'll throw some questions to you on Slack via DM...

Mark-H: https://docs.modx.com/3.x/en/contribute/code/running-tests

Mark-H:
I'm kinda worried this will impact existing sites and we'll get a bunch of issues reported that users suddenly can't use whatever fancy captions and descriptions they've used in the past, but as long as the safeHtml gets covered by sufficient tests and these new restrictions get added to the documentation I don't see a technical reason to reject it.

core/src/Revolution/modX.php
smg6511: @JoshuaLuckers @Mark-H - At long last, I added a bunch of tests for this method. Note that some tweaks were made to the method as a byproduct of creating those tests.

smg6511:
@Mark-H @Jako @JoshuaLuckers - This is finally ready for re-review. I came back and added some tests and made a few tweaks. Not too long after Mark commented re concerns of being too limiting on caption/desc formatting, I loosened the restrictions I'd initially set up and made them fully user-customizable.

JoshuaLuckers:
@smg6511 I'm no longer involved with this project and therefore unable to review your work. I'm sorry.

smg6511:
@JoshuaLuckers Hey - thanks for the heads up! Would you be interested in connecting on LinkedIn? I'll send you an invite if so. Don't know if there will be reason for our paths to cross again, but ya never know ;-)

JoshuaLuckers:
> @JoshuaLuckers Hey - thanks for the heads up! Would you be interested in connecting on LinkedIn? I'll send you an invite if so. Don't know if there will be reason for our paths to cross again, but ya never know ;-)

Feel free to connect: https://www.linkedin.com/in/joshualuckers/

Ibochkarev:
## Summary

Restricts HTML tag usage and ensures proper escaping of element **name**, **caption**, and **description** fields (chunks, snippets, plugins, templates, TVs) in both backend (processors) and frontend (ExtJS forms, grids, windows). Introduces system settings `elements_caption_allowedtags`, `elements_caption_allowedattr`, `elements_description_allowedtags`, and `elements_description_allowedattr` for configurable tag/attribute whitelists. Adds `modX::stripHTML()` (DOMDocument-based) on the PHP side and `MODx.util.safeHtml` / `MODx.util.stripAndEncode` on the JS side.

---

## Suggestions

### 1. Processors/Element/Create.php and Update.php — name field not sanitized on backend

In `beforeSave` / `beforeSet`, only `caption` and `description` are sanitized. The `name` / `templatename` field is validated (empty check, duplicate check) but not stripped of tags. When elements are created or updated via API or programmatically, HTML in the name can be stored. The frontend sanitizes the name (stripTags in TV panel, stripAndEncode in quick create), but the backend remains a gap.

**Recommendation:** Sanitize the name before validation, e.g.:

```php
$name = strip_tags(trim($this->getProperty($this->elementNameField, '')));
$this->object->set($this->elementNameField, $name);
```

Then use `$name` in the `empty()` and `alreadyExists()` checks.

---

### 2. modX.php:stripHTML() — handling of empty `$allowedTags`

When `empty($allowedTags)`, the method returns `strip_tags($htmlSource)`. If `$allowedTags` is `null` or an invalid type, `is_string($allowedTags)` is false and `$allowedTags` may not be normalized before `empty()`.

**Recommendation:** Normalize input explicitly:

```php
$allowedTags = is_array($allowedTags) ? $allowedTags : (is_string($allowedTags) ? trim($allowedTags) : '');
```

---

### 3. modX.php:stripHTML() — `$allowScripts` and `javascript:` in attributes

With `$allowScripts = false`, the check uses `preg_replace('/[^a-zA-Z:]+/', '', $nodeValue)` and `stripos($testVal, 'javascript:')`. This can miss obfuscated variants such as `java\x00script:` or `&#106;avascript:`.

**Recommendation:** Use a stricter check (e.g. `preg_match` with HTML entity decoding) or rely on the attribute whitelist and the `on*` ban.

---

### 4. Create.php / Update.php — different handling of description for TV vs non-TV

For TVs: `stripHtml` with configurable settings. For other elements: `strip_tags($description)`. Behavior differs; non-TV elements cannot allow specific tags (e.g. `<a href="...">`) in descriptions.

**Recommendation:** Consider a unified approach (e.g. `stripHtml` for all elements, with an empty whitelist for non-TV so it behaves like `strip_tags`).

---

### 5. Update.php:111 — `else if` vs `elseif`

Uses `} else if (`; PSR-12 and phpcs prefer `elseif` without a space.

---

### 6. utilities.js — MODx.util.stripAndEncode.run

`Ext.util.Format.stripTags` and `Ext.util.Format.htmlEncode` are applied to all fields using stripAndEncode. For the name field this is appropriate. Ensure `htmlEncode` does not cause display issues in grids (e.g. `&` → `&amp;`) when values are rendered as HTML.

---

### 7. Tests and documentation

Codecov reports ~95 lines without coverage. Mark-H stressed the need for tests around `safeHtml` and documentation for the new restrictions. The PR adds `modXTest.php` for `stripHTML`, which is a good start; consider adding cases for event attributes, `javascript:` variants, and edge cases.

---

## Assessment

### Strengths

- Centralized sanitization via `stripHTML` and system settings.
- Explicit blocking of event attributes (`on*`) and `javascript:` in attributes.
- Support for `data-*` via whitelist.
- Clear separation for TV caption/description vs other elements.
- Refactoring with `elementNameField` reduces duplication.
- JS sanitization on `change` reduces XSS risk before submit.

### Concerns

- Name field is not sanitized on the backend.
- Possible bypasses of the `javascript:` check in attributes.
- Different description handling for TV vs non-TV may complicate configuration.
- Risk of breaking existing sites with custom caption/description usage (per Mark-H).

---

## Verdict

**Approve** with a recommendation to add backend sanitization for the name field and to change `else if` to `elseif` before merge. The other points can be tracked as follow-up issues.

core/src/Revolution/Processors/Element/Create.php
```diff
@@ -50,14 +59,47 @@ public function cleanup()
...
+                    $this->modx->getOption('elements_description_allowedattr')
+                )
+                : strip_tags($description)
+                ;
+            $this->object->set('description', $description);
+        }
+
+        /* verify element has a name and that name does not already exist */
+
+        $name = $this->getProperty($this->elementNameField, '');
```
Ibochkarev, line 92: Name is validated here but never stripped. Caption/description get `stripHtml` / `strip_tags` above. Please strip tags on `name` / `templatename` in the shared processor path so a non-manager client cannot persist markup. Treat Ext listeners as UX only.

core/src/Revolution/Processors/Element/Create.php
```diff
@@ -50,14 +59,47 @@ public function cleanup()
...
-            $this->addFieldError($nameField, $this->modx->lexicon($this->objectType . '_err_ae', [
-                'name' => $name,
-            ]));
+        $locked = (bool)$this->getProperty('locked', false);
+        $this->object->set('locked', $locked);
+
+        $isTV = $this->classKey === modTemplateVar::class;
+
+        if ($isTV) {
+            if ($caption = trim($this->getProperty('caption', ''))) {
```
Ibochkarev, line 68: This TV/caption/description block is duplicated in Update. Extract one `sanitizeElementMetaFields()` (or equivalent on the shared element processor parent) and call it from both Create and Update.

Also: `if ($caption = trim(...))` skips empty values, so a clear may never sanitize or update. Always sanitize when the property is present, including `''`.

core/src/Revolution/Processors/Element/Update.php
```diff
@@ -46,18 +57,44 @@ public function beforeSave()
...
+                    $this->modx->getOption('elements_description_allowedattr')
+                )
+                : strip_tags($description)
+                ;
+            $this->object->set('description', $description);
+        }
+
+        /* verify element has a name and that name does not already exist */
+
+        $name = $this->getProperty($this->elementNameField, '');
```
Ibochkarev, line 87: Same gap as Create: `name` / `templatename` are not sanitized server-side. Please share one helper with Create and strip tags here too.

core/src/Revolution/modX.php
```diff
@@ -3032,4 +3032,109 @@ public static function isValidTimestamp($value, $limitStart = false, $limitEnd =
...
+
+    /**
+     * Removes unwanted tags and/or tag attributes from an HTML string
+     *
+     * @param string $htmlSource The html string to clean
+     * @param ?string|array $allowedTags An array or comma-separated list of tag names to allow
+     * @param ?string|array $allowedAttr An array or comma-separated list of tag attribute names to allow
+     * @param bool $allowScripts Whether to allow javascript in html source passed to this method
+     */
+    public function stripHTML(string $htmlSource, $allowedTags = '', $allowedAttr = '', bool $allowScripts = false): string
```
Ibochkarev, line 3044: Please move this into a dedicated sanitizer class and leave a thin `modX::stripHTML()` delegate if you need BC. `modX.php` is already huge; DOM quirks and allowlist parsing belong in one focused type.

While extracting:
- collect XPath nodes first, then mutate (do not rewrite while iterating the live list)
- replace deprecated `mb_convert_encoding(..., 'HTML-ENTITIES', 'UTF-8')` for PHP 8.2+
- drop or narrowly scope `$allowScripts` for element caption/description paths

core/src/Revolution/Processors/Security/Forms/Set/Update.php
```diff
@@ -134,6 +134,11 @@ public function setFieldRules()
                 $this->newRules[] = $rule;
             }
             if (!empty($field['label'])) {
+                $field['label'] = $this->modx->stripHtml(
```
Ibochkarev, line 137: PHP allows caption tags/attrs on field and tab labels. The FC grids run `stripAndEncode` (full strip + `htmlEncode`). Users see one policy in the UI and another after save. Align field/tab labels with the same rule on both sides (TV labels already roughly match).

manager/assets/modext/widgets/windows.js
```diff
@@ -398,6 +412,13 @@ MODx.window.QuickCreateChunk = function(config) {
             ,name: 'name'
             ,fieldLabel: _('name')
             ,anchor: '100%'
+            ,allowBlank: false
+            ,blankText: _('chunk_err_ns_name')
+            ,listeners: {
+                change: {
+                    fn: MODx.util.stripAndEncode.onChange
```
Ibochkarev, line 419: These `change` listeners are copy-pasted across quick-create windows and the element panels, and they push this file over ~1k lines. Add one helper/mixin that applies `plain` vs `safeHtml(caption|description)` from config, then call it once per field.

manager/assets/modext/widgets/fc/modx.panel.fcset.js
```diff
@@ -318,6 +318,14 @@ MODx.grid.FCSetFields = function(config = {}) {
     });
     MODx.grid.FCSetFields.superclass.constructor.call(this, config);
     this.propRecord = Ext.data.Record.create(config.fields);
+    this.on('afteredit', function(e) {
+        if (e.field === 'label') {
+            const value = MODx.util.stripAndEncode.run(e.value);
```
Ibochkarev, line 323: Field/tab labels use `stripAndEncode` here, while `Forms/Set/Update.php` runs `stripHtml` with caption allowlists. Please match the PHP policy (or change PHP to match this), so save does not reintroduce markup the grid just stripped.

Ibochkarev (changes requested):
## Review

Requesting changes.

This PR targets a real XSS/markup problem in element metadata. The direction is right: server sanitization plus system settings plus tests. The implementation still leaves a server hole on names, duplicates policy in Create/Update and across Ext listeners, and grows already-large shared files.

### Blockers
1. **Name is not sanitized in processors.** Caption/description get `stripHtml` / `strip_tags`. `name` / `templatename` only go through empty/duplicate checks. A connector or API call skips every Ext `change` listener and can persist markup in the main identity field.
2. **Create and Update copy the same caption/description block.** Put one `sanitizeElementMetaFields()` on the shared element processor path and call it from both.
3. **`stripHTML` lands in `modX.php`.** That file is already ~3k lines. Move the DOM walker to a dedicated sanitizer class and keep a thin `modX` delegate if you need BC. While moving it: avoid mutating the live DOM during XPath iteration, replace deprecated `HTML-ENTITIES` conversion for PHP 8.2+, and do not expose `$allowScripts` on the element-metadata path.
4. **JS and PHP disagree on Form Customization labels.** Field/tab labels use `stripAndEncode` in the grid and `stripHtml` with caption allowlists in `Forms/Set/Update.php`. TV labels roughly match. Pick one policy table and drive both layers from it.
5. **`windows.js` grows past ~1k lines with copy-pasted listeners.** Same pattern is pasted across panels. One helper/mixin (`plain` vs caption/description `safeHtml`) should wire fields from config.

### Also fix
- `if ($caption = trim(...))` / description skips empty values, so clears may not sanitize or update.
- Store plain text server-side; encode at render time unless you intentionally persist entities.

### What looks good
`elementNameField` cleanup, allowlist system settings, and the `stripHTML` unit cases. Keep those when you extract the sanitizer.

Happy to re-review after the shared processor helper and sanitizer extraction land.

Ibochkarev:
## Test report

- Result: Partial
- Date: 2026-09-05
- macOS; MODX 3.3.0-dev / PHP 8.2 Docker

### Expected
- Tags cannot persist in element **name** / templatename
- Caption (TV) / description respect allowlists from the 4 settings (processors + JS)
- Dangerous tags/attrs (`script`, `onclick`, `onerror`) removed
- FC Set labels sanitized (covered lightly via related processor path)

### Actual

**Processor smoke: 12/12 Pass**
- Settings present with defaults (`caption` tags=`a`, attr=`href`; description richer allowlist)
- `modX::stripHTML()` removes script; keeps `<a href>` for caption allowlist
- Snippet/chunk **description**: tags stripped (`strip_tags`)
- TV caption/description: allowlist applied in DB
- Name containing tags: processor rejects as invalid (`Snippet/TV/Template name is invalid`) — not stripped, but not saved

**Manager JS: Partial / mismatch**
- `MODx.util.safeHtml` expects allowed tags as `<a><b>…>` (angle-bracket form)
- System settings ship comma lists (`a`, `div,p,ul,…`) which PHP `stripHTML` accepts
- With live `MODx.config.elements_*` values, JS strips **all** tags including allowed `<a>` / `<p>` / ``
- Example: caption input with link → UI becomes `See docs alert(1)` instead of keeping `<a href="…">`
- Snippet name/description via `stripAndEncode`: tags removed (UX OK for plain fields)

**Contrast (same payload)**
| Layer | Caption result |
| --- | --- |
| PHP processor | `See <a href="https://example.com">docs</a> x` |
| JS `change` + config | `See docs alert(1)` |

TV create after JS sanitize
Snippet create stripAndEncode
TV after PHP processor allowlist

### Errors / logs
none crashing; functional allowlist mismatch between JS and settings/PHP

### Notes
- Align `MODx.util.safeHtml` tag parsing with PHP (`a,b` / `a` comma lists) **or** ship settings as `<a><p>…>` and document that format
- Until fixed, Manager UI is stricter than processors (over-strips allowed markup); API/processor path behaves as designed

PR #15937, opened 2021-12-11
[3.x] Make validation error messaging consistent on save

### LINKED ISSUE #15281: For different types of TV, the message that TV are required is shown in different ways

smg6511:
### What does it do?

To overcome the problems the related issue raises, the following was done: 
1. Removed unnecessary call to markInvalid() in MODx.Component class and optimized the surrounding code. 
2. Updated all TV renders to ensure premature validation does not occur. Because of the way the TV rendering is currently done, a fairly counterintuitive move had to be made: The `value` property is only set on a TV render’s Ext field object if there is a non-empty value present.
3. Made additional changes to` listbox-multiple` and `resourcelist` to ensure message UI consistency.
4. Made a handful of minor changes to the `xdatetime` class (for the date TV) to provide custom message targets for the date and time blocks when specifying “under” as the `msgTarget` in the render tpl (or any other date config using xdatetime). Presently, a qTip message is shown no matter what the `msgTarget` specification is.
5. Altered the `MarkRequiredFieldPlugin` in modx.js to correctly mark all required fields. Currently, some fields do not get marked because they lack the criteria that's checked for within the plugin. To make this simple and consistent, I added the `itemId` property to all TV renders for the plugin to check against.
6. Lastly, made a few CSS updates to ensure consistent error message appearance.

### Why is it needed?

When saving, the `handleClick` method incorrectly calls `markInvalid()` after `validate()` returns false on a field, causing the default `invalidText` message to display no matter what the specific validation error is. This, along with some configuration issues in the TV renders, is what causes the inconsistency in field validation messaging referenced in the issue resolved by this PR.

### How to test

2. Create each type of TV with differing requirements (allowBlank in particular) and attach to a template.
3. Create a resource and experiment with placing invalid values (including empty values for required fields) in these TVs,  with and without clicking "Save," to verify the messaging remains consistent and relevant to the validation rule(s) that the input values do not satisfy.
4. The changed code affects _all_ MODX components with a form, so experiment with a few other areas such as Users — where the username and email fields should show the expected errors when not filled or filled with an invalid value.

### Related issue(s)/PR(s)
Resolves #15281.


Ruslan-Aleev:
The messages are now the same, but there is an inconsistency in that for some TVs "This field is required" is shown immediately, and for others (for example, TV with `text` input type) - only when saving.
1) Is it possible to fix this for TV to always show? Or did you want to make this a separate PR?

tvReq

2) Because the message has become universal, it is necessary to remove unnecessary lines of lexicons so as not to create translations. Where, for example, is the message "The value in this field is invalid" used now (or is it used somewhere besides TV)?

smg6511:
OK, I'd initially misinterpreted the issue you're pointing out and have since replicated it here. It's a "needle in a haystack" — I've done some initial searching on why some field types are validating before any action has been taken, but am not seeing anything remotely obvious. I'll find it eventually.

However, when a resource is initially being _created_, IMO nothing should be validated until event has occurred on a field with a validation rule or when an attempt to save is made. You don't want to throw a bunch of errors before the user has had a chance to fill out the form. The issue you raise of required fields not being as visible as you'd like is a separate issue (one definitely worthy of exploring an elegant solution), and something to be mocked up before putting together a PR. 

Bottom line: I think a more nuanced validation UX is needed, but I'm not looking to do that here.

On your other question re the lexicon entry: The entry that outputs "The value in this field is invalid" (`ext_invalidfield`) is the default fallback and is assigned to the invalidText for the Field prototype and shouldn't be removed.

smg6511:
@Ruslan-Aleev - Ok, please check this out again. I've made a few new commits to fully address the issue at hand. I tried to make it so they could be cherry picked to test specific changes. If so, you'd want to add them in the order they appear in (renders, scss, then modx.js). These do kinda go all together though, so that may or may not be helpful ;-)

Ruslan-Aleev (approved):
Thank you, visually everything works as it should

JoshuaLuckers (approved):

theboxer (approved):

PR #15948, opened 2021-12-25
Added deleting / undeleting child resources

Ruslan-Aleev:
### What does it do?
Added deleting / undeleting child resources if parent is deleted / undeleted.

If you delete a resource through the resource editing form (the "Deleted" checkbox), then the deletion status of the child resources will not change. **BUT child resources will be shown in the trash and will be deleted!**
Hovewer, if you delete a resource through the context menu in the tree, then the child resources will also be deleted.

The same is true when undeleting a parent resource.

Now there is no problem, and the deletion status of the child resources will also change.

resource_deleting

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14167#issuecomment-535401489

### DIFF
--- a/core/src/Revolution/Processors/Resource/Update.php
+++ b/core/src/Revolution/Processors/Resource/Update.php
@@ -494,6 +494,46 @@ class Update extends UpdateProcessor
     }
 
     /**
+     * Set deleted status for child resources
+     * @return void
+     */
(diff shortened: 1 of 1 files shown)


theboxer (changes requested):
@Ruslan-Aleev I'd do it similarly how Resource\Delete processor does it. Also you're not triggering delete events.

PR #15950, opened 2021-12-28
Fix missing labels and descriptions in TV quick create/edit window

### LINKED ISSUE #15947: Empty fields in the TV quick edit form
### Summary
Empty fields in the TV quick edit form:

tv_qe

### Related issue(s)/PR(s)
Probably https://github.com/modxcms/revolution/pull/15773
https://github.com/modxcms/revolution/issues/15854

### Environment
MODX 3

smg6511:
### What does it do?
Made changes to address referenced issue and to improve the layout of the quick create/edit TV window:
- Moved select TV entries to the default topic to make them available application-wide 
- Updated SCSS to match field styles more closely with the main editing panel
- Changed the window mode to modal and darkened the overlay to better separate the window from the main view 

Below is a screenshot of how the window should appear:

Screen Shot 2021-12-28 at 9 20 41 PM

### Why is it needed?
There are missing labels and descriptions in the quick create/edit (QCE) form when working with a TV (when the main view is anything but the TV panel, as that's the only place where the TV-specific lexicons are loaded). Also, changing the window to a modal allows better focus on the task at hand.

### How to Test
2. Create at least one TV using the contextual quick create feature. Then, use quick edit on that TV (or others you may have created in your test environment). Try the QCE feature from different pages within the manager to verify labelling and descriptions remain visible.

### Special Note
Ultimately, more work will need to be done in a separate PR to bring the tv type-specific field display of the main TV panel to the associated QCE window (for the _default value_ and _input options_ fields). Also, I am aware that changing the window to a modal prevents the ability to drag/drop tree items into the window's fields. IMO, that's not particularly useful in the QCE windows and is problematic from my testing; it's very easy to unintentionally drop an item in the form beneath without knowing it. I've also found it very distracting to have the front window and the main page visually on the same plane. 

### Related issue(s)/PR(s)
Resolves #15947


Ruslan-Aleev:
Please, you don't need to do such PRs, just keep it simple.
Instead of fixing a bug (which appeared after other improvements) - you fix the bug + add MORE improvements that can add more bugs :)

Please divide this PR by 2:
1) just fix the label bug in the quick edit window https://github.com/modxcms/revolution/issues/15947.
2) add improvements to the quick edit window.

It's easier to test and merge. Thank you :)

smg6511:
Guys - Sorry, I wouldn't have normally done it this way, but I had begun working up something that would fix this issue months ago (thus all the commits—many of which I don't think are directly related—which I was surprised to see given only a handful of files were changed), before learning better ways of structuring these PRs. Question is: The only way I can think of providing the PR in the way you ask is to manually copy the changed files into a new PR (or two)... is there a better way?

Ruslan-Aleev:
I would make a 2 branch based on this branch `3.x-issue-15947`, then, via `git rebase -i upstream/3.x`, leave only the necessary commits for each branch. But maybe there is an easier option, I'm not sure.

smg6511:
OK, I'm closing this to restructure the changes in two PRs. The fix for this issue #15947 can now be found in PR #15951. A separate PR will include the proposed UI refinements.

PR #15951, opened 2021-12-30
Fix missing labels and descriptions in TV quick create/edit window (revised)

### LINKED ISSUE #15947: Empty fields in the TV quick edit form

smg6511:
### What does it do?
(Note: This is a revised partial of PR #15950.)

Addresses referenced issue by moving select TV entries to the default topic to make them available application-wide.

### Why is it needed?

There are missing labels and descriptions in the quick create/edit (QCE) form when working with a TV (when the main view is anything but the TV panel, as that's the only place where the TV-specific lexicons are loaded).

### How to Test
1. From your terminal app, run `grunt build` from within the `_build/templates/default` folder and clear your browser cache. (The build step is only needed if you are running the manager with compress_js on.)
2. Create at least one TV using the contextual quick create feature. Then, use quick edit on that TV (or others you may have created in your test environment). Try the QCE feature from different pages within the manager to verify labelling and descriptions remain visible.

### Special Note
Ultimately, more work will need to be done in a separate PR to bring the tv type-specific field display of the main TV panel to the associated QCE window (for the _default value_ and _input options_ fields).

### Related issue(s)/PR(s)
Resolves #15947


core/lexicon/en/default.inc.php
```diff
@@ -559,3 +559,23 @@
...
+    Shared lang entries for elements --
+
+    Necessary to to give quick create/edit panels
+    access when they are opened outside the
+    context of their respective element types
+*/
+// All
+
+// TVs
+$_lang['tv_tab_input_options'] = 'Input Options';
```
Ruslan-Aleev: Why did you move this `tv_tab_input_options` lexicon from `core/lexicon/en/tv.inc.php` if it is only used in the TV editing form?

core/lexicon/en/default.inc.php
```diff
@@ -559,3 +559,23 @@
...
+$_lang['tv_type'] = 'Input Type';
+$_lang['tv_default'] = 'Default Value';
+$_lang['tv_default_desc'] = 'The content this TV will show if user-entered content is not provided.';
+$_lang['tv_caption_desc'] = 'The label shown for this TV in Resource editing pages (can be overridden per template or other criteria using <a href="?a=security/forms" target="_blank">Form Customization</a>).';
+$_lang['tv_category_desc'] = 'Use to group TVs in Resource editing pages and within the Elements tree.';
+$_lang['tv_description_desc'] = 'Usage information for this TV shown next to its caption in Resource editing pages and as a tooltip in the Elements tree.';
```
Ruslan-Aleev, line 578: In my opinion, additional TV lexicons with description (`_desc`) are not needed at all in the quick edit window. There are no such descriptions either for other elements, **or even for resources**, which is logical, since this is a quick editing of already existing fields (and their description is in the edit form).

p.s. In a separate commit (or PR), we can change the order of the fields, for example, it will be like this:
tv_qew

core/lexicon/en/tv_widget.inc.php
```diff
@@ -240,11 +237,25 @@
 $_lang['tv_elements_listbox-multiple_desc'] = $_lang['tv_elements_listbox_desc'];
 $_lang['tv_elements_radio_desc'] = $_lang['tv_elements_option_desc'] = $_lang['tv_elements_desc'];
 $_lang['tv_elements_tag_desc'] = $_lang['tv_elements_desc'];
-$_lang['tv_type'] = 'Input Type';
+$_lang['tv_daterange_elements_desc'] = 'Test options desc for daterange with example ph: [[+ex1]]';
+$_lang['tv_daterange_default_text_desc'] = 'Test default text desc for daterange with example ph: [[+ex1]]';
```
Ruslan-Aleev: These lexicons are not used in the code, they are not needed in this PR.

p.s. I ask you not to add / change lexicons without great need, problems arise:
1) New lexicons need to be translated, additional work (sometimes unnecessary), and there are not enough people. As a result, we have a semi-translated MODX, what looks crooked.
2) When a lexicon entry changes, in crowdin (https://crowdin.com/project/modx-revolution) even the translated strings are overwritten with new ones and wait for translation (because they have changed). As a result, it may turn out that after a few PRs ALL translations will be in English, because there are not enough people to translate.
3) Unused lexicons accumulate over time: extra code, extra distribution weight (see https://github.com/modxcms/revolution/issues/14512#issuecomment-650345391).

Ruslan-Aleev (changes requested):

core/lexicon/en/default.inc.php
smg6511: It's needed for the quick create/edit TV windows.

core/lexicon/en/tv_widget.inc.php
smg6511: Sorry, those were stray lex entries I was using to test something else and they never got deleted. They've now been removed.

core/lexicon/en/default.inc.php
smg6511: Editing is only half the story though. I probably use the quick create more than the edit, mainly to just scaffold fields that will be needed for a particular purpose. IMO, that’s the primary usefulness of the feature. 

I will have another PR coming momentarily that would make what I've done here more logical (as it applies a consistent format to all of the quick create/edit windows). There we can debate the merits of including descriptions. The implementation of the help/descriptions is wildly inconsistent and should really be presented in the same way across the board if we want a truly polished UI.

theboxer (approved):

core/lexicon/en/default.inc.php
Ruslan-Aleev: Can you point out where this lexicon is used in `manager/assets/modext/widgets/windows.js` (or am I confusing something)? I didn't find the line.

core/lexicon/en/default.inc.php
Ruslan-Aleev: > The implementation of the help/descriptions is wildly inconsistent and should really be presented in the same way across the board if we want a truly polished UI.

I agree.
In this PR, can you bring terminology to the same format in lexicons (at least **name -> label**)? Now it is different, which can be confusing, I mean:

tv_qew

core/lexicon/en/default.inc.php
smg6511: Oh, sorry I misspoke. That actually is not needed for the windows at this point; I was thinking at some point that I might propose the TV quick edit window be tabbed, with Input Options being a second tab having the Default Value and Input Option Values fields (in which case this move would have been needed). I'll move that entry back to the TV lex in just a bit...

core/lexicon/en/default.inc.php
smg6511: Ok, yes I missed that. Will change...

core/lexicon/en/default.inc.php
smg6511: Update done.

core/lexicon/en/default.inc.php
smg6511: Update done.

core/lexicon/en/default.inc.php
Ruslan-Aleev: The lexicon has not moved, please check again.

core/lexicon/en/tv.inc.php
```diff
@@ -87,3 +86,16 @@
...
 $_lang['tv_tmpl_access'] = $_lang['tv_tab_tmpl_access'];
 $_lang['tv_tmpl_access_msg'] = $_lang['tv_tab_tmpl_access_desc'];
+
+/*
+    Refer to default.inc.php for the keys below.
+    (Placement in this default file necessary to allow
+    quick create/edit panels access to them when opened
+    outside the context of their respective element types)
+
+    tv_tab_input_options
```
Ruslan-Aleev: By the way, don't forget about the comment :)

Ruslan-Aleev (changes requested):

core/lexicon/en/tv.inc.php
smg6511: I must be losing my mind! ;-) I made the change but managed to only push one of the three changed files. Anyway, you should see the change now.

Ruslan-Aleev (approved):

PR #15952, opened 2022-01-03
Compatibility for custom templates between 2.8.1 and 2.8.2

Jako:
### What does it do?
Make the smarty variable $_authToken optional.

### Why is it needed?
If the installation uses a custom manager template, the line of this patch has to be added to the custom template. After that the custom template is working in 2.8.1 and 2.8.2. Before it won't work in 2.8.1, since HTTP_MODAUTH will be used with an empty value. With the old custom manager template the manager won't work in 2.8.2.

### How to test
Create a custom template with the new header.tpl of 2.8.2 and add it to a 2.8.1 or lower installation. The modx.config.js.php is not loaded and a destroyed manager occurs.

### Related issue(s)/PR(s)
#15644

### DIFF
--- a/manager/templates/default/header.tpl
+++ b/manager/templates/default/header.tpl
@@ -19,7 +19,7 @@
 {/if}
 <script src="{$_config.manager_url}assets/modext/core/modx.js?v={$versionToken}"></script>
 <script src="{$_config.connectors_url}lang.js.php?ctx=mgr&topic=topmenu,file,resource,trash,{$_lang_topics}&action={$smarty.get.a|default|htmlspecialchars}"></script>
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

JoshuaLuckers (approved):

opengeek:
Is the _authToken not something that needs to be there from a security patch?

Jako:
Yep, but the advantage of the patch is that the template works in both versions without any problem. And everyone can compare an old custom template with the patched template, apply the changes and test this in pre and after 2.8.2 versions. It is more a how-to than a patch.

Mark-H:
The auth token is required for it to work, so why not put this in the docs somewhere? [2.8.2 upgrade docs?](https://docs.modx.com/current/en/getting-started/maintenance/upgrading/2.8.2)

Edit: guess what, [it's already there lol](https://docs.modx.com/current/en/getting-started/maintenance/upgrading/2.8.2#custom-manager-themes)

Jako:
Sorry, I never saw that part of the docs. And the one who asked me to fix that on their installation too.

PR #15955, opened 2022-01-06
[3.x] Form Customization: Grid editing erases set data

### LINKED ISSUE #15929: [3.x] Form Customization: Grid editing erases set data
### Summary

When editing the basic settings of an FC set (when you first open an FC profile) via the grid editor (editing description, template, and the constraint fields), all the rules saved within that set get wiped out.

### Step to reproduce

1. Create a new FC profile and one FC set within that profile.
2. Customize a handful of labels, default values, etc. within this new set.
3. Close the set editing page (you will arrive back at the listing of sets for the current profile).
4. Update the set’s description (or any of the other 3 fields) via the grid editor.
5. Use the gear menu (or context menu) to open the main editing page for this set once again and note how the rules you previously defined are gone.

### Related issue(s)/PR(s)
n/a

### Environment
MODX 3.0.0-rc1.

theboxer:
### What does it do?
- Updates the UpdateFromGrid processor to extend directly core UpdateProcessor and not `MODX\Revolution\Processors\Security\Forms\Set\Update` Processor
- Make sure only fields editable from grid will store to db

### Why is it needed?
- Right now the update from grid processor extends the `MODX\Revolution\Processors\Security\Forms\Set\Update` which expects the whole data set to be present. The grid only passes a fraction of the data which causes the wipe of all rules, labels etc. (as described in #15929).

### How to test
Described in the related issue #15929

### Related issue(s)/PR(s)
Resolves #15929

### DIFF
--- a/core/src/Revolution/Processors/Security/Forms/Set/UpdateFromGrid.php
+++ b/core/src/Revolution/Processors/Security/Forms/Set/UpdateFromGrid.php
@@ -10,12 +10,22 @@
 
 namespace MODX\Revolution\Processors\Security\Forms\Set;
 
+use MODX\Revolution\modFormCustomizationSet;
+use MODX\Revolution\Processors\Model\UpdateProcessor;
+
 /**
  * Update a FC Profile from grid
  * @package MODX\Revolution\Processors\Security\Forms\Set
  */
-class UpdateFromGrid extends Update
+class UpdateFromGrid extends UpdateProcessor
 {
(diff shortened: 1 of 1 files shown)


smg6511 (approved):
The implementation makes sense after reviewing the previous and newly-assigned parent processors, and solves the issue. Tested by making changes to each and every grid-editable _Set_ field, as well as the parent _Profile_ grid fields (which shouldn't be relevant to this specific issue, but it doesn't hurt to test).

PR #15968, opened 2022-01-13
Adjust modUserGroup

### LINKED ISSUE #15365: Deleting parent user group does not delete children or move them to top level
### Summary
If in the manager interface and you delete a usergroup that has child user groups, the children user groups do not get deleted and thier parent ID does not get updated so they are no longer accessible from the manager interface but are visible across the manager interface in dropdowns etc.

### Step to reproduce
Delete a parent group that has children

image

### Observed behavior
parent group vanishes along with children (mostly due to having a parent id that no longer exists)

### Expected behavior
Children are deleted or parent id changes to 0 to move to top level

### Related issue(s)/PR(s)
Let us know if this is related to any issue/pull request.

### Environment
MODX version 2.7.3

theboxer:
### What does it do?
- Changes Children to composite relation (deleting modUserGroup will also delete all it's children)
- Adjust the confirm message on resource group remove - to include notice about deleting children
- Fix creating empty user group when dropping Anonymous user group under some other group
- Disable D&D for admin and anonym user group
- Disable changing parent of admin user group
- Improve D&D of user groups (before you couldn't drop to user group without childs)
- Remove sorting users from the processor - users are no longer part of the user groups tree

### Why is it needed?
Overall QoL improvements and fixes.

### How to test
Play with user groups

### Related issue(s)/PR(s)
Resolves #15365


core/src/Revolution/Processors/Security/Group/GetNodes.php
```diff
@@ -147,15 +150,21 @@ public function prepareGroup(modUserGroup $group)
...
             'type' => 'usergroup',
             'qtip' => $group->get('description'),
             'cls' => $cls,
+            'allowDrop' => true,
             'iconCls' => 'icon icon-group',
         ];
-    }
 
+        if (!$itemArray['hasChildren']) {
+            $itemArray['expanded'] = true;
```
JoshuaLuckers, line 165: What's the reason to leave out the `expanded` array key when `false`?

manager/assets/modext/widgets/security/modx.panel.user.group.js
```diff
@@ -105,7 +105,7 @@ MODx.panel.UserGroup = function(config) {
                                 ,fieldLabel: _('user_group_parent')
                                 ,editable: false
                                 ,anchor: '100%'
-                                ,disabled: config.record.id === 0
+                                ,disabled: config.record.id === 0 || config.record.name === 'Administrator'
```
JoshuaLuckers, line 108: Is it "Administrator" in all languages?

core/src/Revolution/Processors/Security/Group/GetNodes.php
theboxer: By default nodes are collapsed, no reason to explicitly set that.

manager/assets/modext/widgets/security/modx.panel.user.group.js
theboxer: Same concept is used on line 69 for disabling changing name.

JoshuaLuckers: Shouldn't we use a lexicon instead? Since you're using it in the `isAdminGroup()` function as well:

https://github.com/modxcms/revolution/blob/92931b6d622a9860db63c23dc6794ef755c3c939/core/src/Revolution/Processors/Security/Group/Update.php#L191

theboxer: Fair, using lexicons makes no sense. Adjusted.

JoshuaLuckers (approved):

Issue #15595, opened 2021-03-27
User name in the manager


sergant210:
### Summary
In the manager it is not clear what user is logged in. 

![username](https://user-images.githubusercontent.com/4679725/112712944-985ec480-8ee3-11eb-9617-d47cf8399f5d.jpg)

### Suggested solution(s)
May be to add a header to the context menu.
![username](https://user-images.githubusercontent.com/4679725/112713104-65690080-8ee4-11eb-9be0-a49dc4641946.jpg)

Ruslan-Aleev:
@opengeek @rthrash And why are labels hidden in the bottom menu? Can we display them in the same way as in the menu above? Now it's just hidden through css. And it will be like this:

![footer_menu](https://user-images.githubusercontent.com/12523676/113025517-b43ac280-9190-11eb-9f5f-9f3c3842835f.png)

opengeek:
@Ruslan-Aleev — @rthrash and I have nothing to do with that as far as I know. I'm not sure why you are @ mentioning us?

sergant210:
I would suggest to rename the Admin menu item to System.

Ruslan-Aleev:
@opengeek Who to ask then? Knowing your conservative attitude towards PRs, it is probably logical to assume that such a display of the menu was done, at least with your approval.

Mark-H:
Jason has never been involved with the UI side of things, we all have our areas of expertise 😅

I suppose this was built this way for simplicity, based on initial designs that came out of the sterc project. That was probably not extensively discussed one way or another, but just one person's concept that was further adapted over time.

Important to note there's not much room (horizontally), 
so people using an email address as username for example would never fit in the sidebar itself. 

For system/help it would fit better if people think that's better. Right now in 2.x there's no label for that either though.

jaygilmore:
I don't think any value is added in finding out who made the specific design choice or who accepted it in the first place. I think our energies would be better spent in working to offer considered solutions to what is likely a valid UX concern. 

The "how" it gets implemented definitely merits discussion. As @Mark-H notes, it could get messy there with longer usernames or lexicon entries for different languages (German words can get pretty long). Without looking at the code recently, there may also be some other considerations around accessibility. 

[...]

PR #15973, opened 2022-01-18
Displayed the username in user menu

Ruslan-Aleev:
### What does it do?
Displayed the username in user menu.
The `{$username}` placeholder works in the menu, which allows to display the username.

Before:
menuBefore

After:
menuAfter

### Why is it needed?
In the manager it is not clear what user is logged in.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15595

### DIFF
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -347,7 +347,7 @@ $children = [];
 $children[0]= $xpdo->newObject(modMenu::class);
 $children[0]->fromArray([
   'menuindex' => 0,
-  'text' => 'profile',
+  'text' => '{$username}',
   'description' => 'profile_desc',
   'parent' => 'user',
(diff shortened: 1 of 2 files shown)


alroniks (changes requested):
Migration created, that's nice, but it missed [here](https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/mysql/3.0.0-pl.php). It will be run never.

Ruslan-Aleev:
@alroniks Here is the line - https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/mysql/3.0.0-pl.php#L21
I didn't create the file, it already existed, I just added the code. Or what are you talking about?

JoshuaLuckers (approved):

alroniks (approved):

PR #15977, opened 2022-01-19
Fix empty files considered non-existant (#15976)

### LINKED ISSUE #15976: MODX does not allow you to create/open an empty file
### Summary
MODX does not allow you to create/open an empty file, even if the file is not created through the CMS.

### Step to reproduce
Try to create an empty txt file and save it.

emptyFile

If you create an empty file on the server, then MODX will display the file in the tree, but will not allow you to open it.

### Environment
MODX 3.0.0-rc1

Mark-H:
### What does it do?
Fixes a logic issue in the media source implementation.

### Why is it needed?
Relying on `read()` limits the functionality to files that can be read. But readable files may still exist and should be editable.

### How to test
Create an empty file and try to edit it. Note MODX shows the "File doesn't exist!" error. Apply patch, try again.

### Related issue(s)/PR(s)
#15976

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -577,10 +577,11 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
     public function getObjectContents($path)
     {
         try {
-            if (!$file = $this->filesystem->read($path)) {
+            if (!$this->filesystem->fileExists($path)) {
                 $this->addError('file', $this->xpdo->lexicon('file_err_nf'));
                 return [];
             }
+            $content = $this->filesystem->read($path);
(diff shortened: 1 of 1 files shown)


opengeek (approved):

Ruslan-Aleev:
@Mark-H @opengeek the problem of opening a file from the element tree remains: an empty file is visible, but does not respond to clicks.

Mark-H:
It's working as expected here with the patch. Is the file readable?

Ruslan-Aleev:
Readable, it seems. I can create a file, it opens when saving. But if I close it, I won't be able to open it again.
See gif (I click on the file):

emptyFile

p.s. I have Windows OS, can this be the case?

PR #15978, opened 2022-01-19
Mark modX::parseChunk as deprecated (#12433)

### LINKED ISSUE #12433: parseChunk does not give precedence to values passed in
### Summary

When passing an associative `array` to `$modx->parseChunk()`, the values parsed in are not given priority processing order, and therefore are not always honored if you have a conflicting, preexisting placeholder set before it.
### Step to reproduce
1. Create a Plugin (`OnWebPageInit`) that sets a placeholder on each request:
   
   ``` php
   $modx->setPlaceholder('example', 'ABC' );
   ```
2. Create a Chunk (example) that outputs the placeholder:
   
   ``` html
   Example: [[+example]]
   ```
3. Create and run a Snippet, invoking `parseChunk`, but with custom values:
   
   ``` php
   return $modx->parseChunk( 'example', array( 'example' => 'XYZ' ) );
   ```
### Observed behavior

Output = Example: **ABC**

The `parseChunk` method calls `getChunk` resulting in the conflicting placeholder being merged _first_, prior to parsing the `array` values provided.
### Expected behavior

Output = Example: **XYZ**
### Environment

MODX 2.3.4-dev

Mark-H:
### What does it do?

Marks modX::parseChunk as deprecated to discourage its use, targeting removal with the next batch of breaking changes, in 4.0.

### Why is it needed?
As discussed in #12433, parseChunk doesn't behave how one would expect it to, and is a lesser alternative to getChunk. That means we're better off without it, however we can't just remove it out-right.

### How to test
Not applicable.

### Related issue(s)/PR(s)
Fixes #12433

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1954,8 +1954,10 @@ class modX extends xPDO {
      * @param string $prefix The placeholder prefix, defaults to [[+.
      * @param string $suffix The placeholder suffix, defaults to ]].
      * @return string The processed chunk with the placeholders replaced.
+     * @deprecated Prefer using $modx->getChunk, to be removed in MODX 4.0.
      */
     public function parseChunk($chunkName, $chunkArr, $prefix='[[+', $suffix=']]') {
+        $this->deprecated('3.0.0', 'Use $modx->getChunk instead');
         $chunk= $this->getChunk($chunkName);
         if (!empty($chunk) || $chunk === '0') {
             if(is_array($chunkArr)) {


alroniks (approved):

core/src/Revolution/modX.php
```diff
@@ -1954,8 +1954,10 @@ public function getChunk($chunkName, array $properties= []) {
      * @param string $prefix The placeholder prefix, defaults to [[+.
      * @param string $suffix The placeholder suffix, defaults to ]].
      * @return string The processed chunk with the placeholders replaced.
+     * @deprecated Prefer using $modx->getChunk, to be removed in MODX 4.0.
```
JoshuaLuckers, line 1957: Shouldn't this mention the "deprecated since release"?

JoshuaLuckers (approved):

PR #15981, opened 2022-01-20
[3.x] Update permissions for viewing and clearing manager actions log

### LINKED ISSUE #15739: [Proposal] Change permission in system logs
### Summary
System logs /manager/?a=system/logs is very useful thing. Now we have one permission 'logs' for two different action 'clear logs' and 'show logs'.

### Why is it needed?
To have more powerfull and flexible permission roles.

Example.
There are 5 editor in online store. 4 editor - is simple editor who can't show logs and 1 chief-editor who can show logs, but not truncate it.

### Suggested solutions
1. Add new permission 'clear_logs'
2. Permission 'logs' rename to 'show_logs'.
Screenshot 2021-06-17 at 12 59 03

muzzwood:
### What does it do?
- Removes the non-specific `logs` permission
- Adds `mgr_log_view` permission
- Adds `mgr_log_erase` permission
- Includes these two new permissions in the Administrator and Developer policies.

I went with these names rather than the ones suggested in the issue so that they are uniform with the error log permissions.

### Why is it needed?
This will allow a user to view the manager actions log but not clear it.

### How to test
1. Re-run setup to get the new permissions installed
2. Log in with users with different access to attempt viewing and clearing the manager access log.

### Related issue(s)/PR(s)
Closes #15739


manager/assets/modext/widgets/system/modx.grid.manager.log.js
```diff
@@ -205,19 +211,22 @@ MODx.grid.ManagerLog = function(config) {
...
-            ,scope: this
-            ,handler: function() {
-                var fp = Ext.getCmp(this.config.formpanel);
-                if (fp) {
-                    fp.clearLog();
+        },'->'];
+        if (MODx.perm.mgr_log_erase === true) {
+            tbar.push({
+                xtype: 'button'
+                , text: _('mgrlog_clear')
```
Ruslan-Aleev: Remove the spaces after the commas, otherwise it looks strange :)

Ruslan-Aleev (changes requested):

manager/assets/modext/widgets/system/modx.grid.manager.log.js
muzzwood: Good old auto-formatting. Done!

Mark-H (approved):

JoshuaLuckers:
Do we need to remove the `logs` permission in an upgrade script?

muzzwood:
Thanks @JoshuaLuckers - done!

opengeek:
Hmm. Does this really need to be held until 3.1? If so, then the upgrade script needs to be targeted at 3.1 instead of 3.0.

Mark-H:
Felt more like a feature than a fix, but I'm certainly more than happy to see it in rc2 as well.

opengeek:
Since I don't believe this would affect any extras, I think it makes more sense to just bring it in for RC2 if no one else has objections.

muzzwood:
Fine with me! :)

opengeek (approved):

Issue #14340, opened 2019-02-07
[bug] Confirmation window changes width


Ruslan-Aleev:
### Summary
In the Deleted Resources Manager, **possibly in other sections**, the confirmation window changes its width.

### Step to reproduce
1) In the Deleted Resources Manager, mark the resources and click "Purge All"
2) On the confirmation window, click "No"
3) Press "Purge All" again - the window has reduced its width.

![trash](https://user-images.githubusercontent.com/12523676/52404292-1aac8480-2ae2-11e9-8f17-40af3a3fd420.gif)

### Environment
MODX 2.7.0

Ibochkarev:
I can reproduce this bug. MODX 2.7.0

ant-tar:
Noticed that too for 2.7.1

Ibochkarev:
The problem also exists in 3.x

Ibochkarev:
Fixed #15984

PR #15984, opened 2022-01-21
Changed the minimum width of popup window

Ruslan-Aleev:
### What does it do?
For some reason the width of popups is sometimes reset to the minimum width.
~Previously, the value was equal to 200px, which is not enough, increased to 400px, so that even when resetting the content is not cut off.~

Set window width via parameter - https://github.com/modxcms/revolution/pull/15984#issuecomment-1020008136

Before (content can be cut off by window):
trash

After:
minWidth

### Why is it needed?
Prevent cut off of window content.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14340

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.grid.trash.js
+++ b/manager/assets/modext/widgets/resource/modx.grid.trash.js
@@ -193,6 +193,7 @@ Ext.extend(MODx.grid.Trash, MODx.grid.Grid, {
 
     purgeResource: function () {
         MODx.msg.confirm({
+            minWidth: 500,
             title: _('trash.purge_confirm_title'),
             text: _('trash.purge_confirm_message', {
                 'list': this.listResources('')
@@ -226,6 +227,7 @@ Ext.extend(MODx.grid.Trash, MODx.grid.Grid, {
             withPublish = '_with_publish';
         }
         MODx.msg.confirm({
+            minWidth: 500,
             title: _('trash.restore_confirm_title'),
             text: _('trash.restore_confirm_message' + withPublish, {
                 'list': this.listResources('')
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

Jako:
Do you have a reproducible click path where this issue occurs? Then I can look for this. Does the alert has an explicit ID?

Ruslan-Aleev:
@Jako This happens in the Deleted Resources Manager, see more details https://github.com/modxcms/revolution/issues/14340

> Does the alert has an explicit ID?

This happens, at least, by clicking on the "Erase All", "Restore All" button (most likely relevant for all windows). And you can click one, close it, and click another - the effect will repeat.

Jako:
This is not a real fix, because it does not solve but just bypass the issue.

Ruslan-Aleev:
Yes, that's what I meant by:

> It is more correct to look for the reason for resetting the width...

And added the "proposal" tag.

Jako:
Please use #15996 and set the minWidth of the confirm boxes by option.

```
MODx.msg.confirm({
    ...
    minWidth: 500,
    ...
```

Ruslan-Aleev:
@Jako Are you sure it works?
Whatever width I set, the default width is taken for the window.
Checked at least in Deleted Resources Manager section.

Jako:
Yes it is working here fine. You have to add the minWidth to all MODx.msg.confirm calls in modx.grid.trash.js (6 times).

Jako:
And maybe you have to force reload the manager to load the changed javascripts.

Ruslan-Aleev:
Yes, thanks, it works, I will correct this PR after your PR https://github.com/modxcms/revolution/pull/15996 merges.

Ruslan-Aleev:
@Jako please check.

Jako (approved):

Ibochkarev (approved):

PR #15985, opened 2022-01-21
Fix empty files not being clickable in the file tree (#15976)

### LINKED ISSUE #15976: MODX does not allow you to create/open an empty file

Mark-H:
### What does it do?

- Adds a debug log statement so we can inspect the mime types of files more easily in the future
- Considers application/x-empty as a non-binary file type
- Fixes `PHP warning: Undefined array key "Orientation"` warning for each image in the tree lacking that Exif header

### Why is it needed?
Fix empty files not being clickable in the file tree, due to unrecognised application/x-empty mime type

### How to test
https://github.com/modxcms/revolution/pull/15977#issuecomment-1018613790

### Related issue(s)/PR(s)
Fixes #15976 (again)

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2331,7 +2331,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
     {
         try {
             $mime = $this->filesystem->mimeType($file);
-            $this->xpdo->log(xPDO::LOG_LEVEL_DEBUG,$file . ' has a mime type of: ' . $mime);
+            $this->xpdo->log(xPDO::LOG_LEVEL_DEBUG, $file . ' has a mime type of: ' . $mime);
 
             // Some mimetypes include a character set, e.g. application/json; charset=utf-8
             // so we filter out the last part to make comparison easier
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2331,6 +2331,7 @@ protected function isFileBinary($file)
     {
         try {
             $mime = $this->filesystem->mimeType($file);
+            $this->xpdo->log(xPDO::LOG_LEVEL_DEBUG,$file . ' has a mime type of: ' . $mime);
```
JoshuaLuckers:
```suggestion
            $this->xpdo->log(xPDO::LOG_LEVEL_DEBUG, $file . ' has a mime type of: ' . $mime);
```

muzzwood:
Just tested this. It works for me 👍🏻

JoshuaLuckers (approved):

PR #15996, opened 2022-01-24
[3.x Port] Configurable MessageBox minWidth

Jako:
### What does it do?
Make the minWidth of MODx.Msg.confirm configurable

### Why is it needed?
The current confirm method of MODx.Msg uses the default Ext.MessageBox.minWidth set a few lines above.

### How to test
Set the minWidth property of MODx.confirm to a value and open+close and reopen the confirm box again. Without the minWidth property, the reopened box is too small, since the size is only calculated on the first display (which is a second issue – but thats inside of Ext JS).

### Related issue(s)/PR(s)
#15984, 3.x Port of #15995

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -911,7 +911,7 @@ Ext.extend(MODx.Msg,Ext.Component,{
               this.addListener(i,l.fn,l.scope || this,l.options || {});
             }
         }
-        Ext.MessageBox.minWidth = (config.minWidth) ? config.minWidth : 200;
+        Ext.MessageBox.minWidth = config.minWidth || 200;
         Ext.Msg.confirm(config.title || _('warning'),config.text,function(e) {
             if (e == 'yes') {
                 MODx.Ajax.request({
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -911,6 +911,7 @@ Ext.extend(MODx.Msg,Ext.Component,{
               this.addListener(i,l.fn,l.scope || this,l.options || {});
             }
         }
+        Ext.MessageBox.minWidth = (config.minWidth) ? config.minWidth : 200;
         Ext.Msg.confirm(config.title || _('warning'),config.text,function(e) {
             if (e == 'yes') {
                 MODx.Ajax.request({


manager/assets/modext/core/modx.js
```diff
@@ -911,6 +911,7 @@ Ext.extend(MODx.Msg,Ext.Component,{
               this.addListener(i,l.fn,l.scope || this,l.options || {});
             }
         }
+        Ext.MessageBox.minWidth = (config.minWidth) ? config.minWidth : 200;
```
Ruslan-Aleev: 200 is not enough, you should set the size bigger, it seems to me.

And it might be shorter:
`Ext.MessageBox.minWidth = config.minWidth || 400;`

Jako: It is the minWidth and not the width. So it does not have to be larger.

Ruslan-Aleev: I understand that it is minWidth. Here, for example, is how it looks at 200px with a window of 1920px (fits only 2 buttons, see second window in gif). If for some window minWidth is applied, then the width will obviously not be enough, isn't it?

trash

Ruslan-Aleev: And line `903` becomes unnecessary, or am I mistaken?
`Ext.MessageBox.minWidth = 200;`

p.s. Can you fix tabs in the whole file with a separate commit? Because there are tabs and spaces, it looks strange...

theboxer (approved):

manager/assets/modext/core/modx.js
opengeek: Can we finish this conversation and resolve it, please?

Jako: Line 903 is required for all other MessageBox types except "confirm".

Mark-H:
Documentation updated pending for when this is merged: https://github.com/modxorg/Docs/pull/385

Tested and confirmed working by running the following in the console:

```
MODx.msg.confirm({
   title: 'Are you sure?',
   text: 'Do you want to delete the world? This is irreversible.',
   minWidth: 500,
   listeners: {
        'success':{fn: function(r) {
             MODx.clearCache(); /* clear cache after world destruction, so we dont have latent data */
        },scope:true}
   }
});
```

Mark-H (approved):

PR #15997, opened 2022-01-24
Fix aliases for resource controllers

theboxer:
### What does it do?
Fixes class aliases for resource controllers introduced in #15888

### Why is it needed?
Create resource pages don't work, when you pass the class key, for example: `/manager/?a=resource/create&parent=1&context_key=web&class_key=MODX\Revolution\modDocument` will throw `Error: Class 'DocumentCreateManagerController' not found`

The class alias is defined before the class itself, which is probably invalid :)

### How to test
Open `/manager/?a=resource/create&parent=1&context_key=web&class_key=MODX\Revolution\modDocument` with and without this fix.

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -17,8 +17,6 @@ use xPDO\Om\xPDOQuery;
 
 require_once __DIR__ . '/resource.class.php';
 
-class_alias(ResourceCreateManagerController::class, 'DocumentCreateManagerController');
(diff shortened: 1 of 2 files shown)


JoshuaLuckers (approved):

Ibochkarev (approved):

rthrash:

https://community.modx.com/t/modx-3-0-0-error-class-document-update-manager-controller-not-found/4914/2

PR #16001, opened 2022-01-25
[3.x] Fix issue where moving resource above another would make the "target" a container.

### LINKED ISSUE #15991: [3.0.0-rc1] container flag is set seemingly randomly after dragdrop in resource tree
### Summary
When moving around resources in the resource tree, the container flag (`isfolder` column) is set on seemingly random resources.

### Step to reproduce
container_bug

Take 3 resources. Drag resource 3 into resource 2. Refresh. Drag resource 3 out of resource 2 above resource 1. Refresh. 

### Observed behavior
Resource 1 is now a container.

### Expected behavior
The container flag of resource 1 remains the same.

### Environment
MODX 3.0.0-rc1

JoshuaLuckers:
### What does it do?
Make sure we only mark the target as folder when the node is appended to another node.
If the point we are dragging/dropping the node is "above/below", the target resource is a sibling and not a parent and therefore it should not become a container/folder.

### Why is it needed?
Fix issue where moving resource above another would make the "target" a container.

### How to test
Drag and drop a resource above another resource that's not a container/folder.
Also see #15991

### Related issue(s)/PR(s)
Resolves #15991

### DIFF
--- a/core/src/Revolution/Processors/Resource/Sort.php
+++ b/core/src/Revolution/Processors/Resource/Sort.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
@@ -9,7 +10,6 @@
  */
 
 namespace MODX\Revolution\Processors\Resource;
-
 
 use MODX\Revolution\modContext;
 use MODX\Revolution\Processors\Processor;
@@ -53,9 +53,13 @@ class Sort extends Processor
 
         // Because of BC
         $data = urldecode($this->getProperty('data', ''));
-        if (empty($data)) $this->failure($this->modx->lexicon('invalid_data'));
+        if (empty($data)) {
+            $this->failure($this->modx->lexicon('invalid_data'));
+        }
         $data = $this->modx->fromJSON($data);
-        if (empty($data)) $this->failure($this->modx->lexicon('invalid_data'));
+        if (empty($data)) {
+            $this->failure($this->modx->lexicon('invalid_data'));
+        }
 
         // Because of BC
         $this->getNodesFormatted($data, $this->getProperty('parent', 0));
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev (approved):
Seems to work as it should

core/src/Revolution/Processors/Resource/Sort.php
```diff
@@ -53,9 +53,13 @@ public function process()
 
         // Because of BC
         $data = urldecode($this->getProperty('data', ''));
-        if (empty($data)) $this->failure($this->modx->lexicon('invalid_data'));
+        if (empty($data)) {
+            $this->failure($this->modx->lexicon('invalid_data'));
```
Mark-H:
```suggestion
            return $this->failure($this->modx->lexicon('invalid_data'));
```

core/src/Revolution/Processors/Resource/Sort.php
```diff
@@ -53,9 +53,13 @@ public function process()
...
+        }
         $data = $this->modx->fromJSON($data);
-        if (empty($data)) $this->failure($this->modx->lexicon('invalid_data'));
+        if (empty($data)) {
+            $this->failure($this->modx->lexicon('invalid_data'));
```
Mark-H:
```suggestion
            return $this->failure($this->modx->lexicon('invalid_data'));
```

Mark-H:
Made two minor code suggestions to make sure the validation actually does something, but I can confirm this resolves the issue with random assignments.

opengeek (approved):

opengeek:
I think a botched trying to add brevity to the title of this merged change. It would be helpful if we can more concisely describe changes like this when submitting. I know it's not easy sometimes.

JoshuaLuckers:
> I think a botched trying to add brevity to the title of this merged change. It would be helpful if we can more concisely describe changes like this when submitting. I know it's not easy sometimes.

What would have been a better title?

PR #16002, opened 2022-01-25
Fix selecting parent in quick create resource

### LINKED ISSUE #15983: Error selecting parent resource in quick resource creation window
### Summary
Error selecting a parent resource in the quick resource creation window if you are editing the that resource you want to select as a parent quick resource creation window.

### Step to reproduce
1) Create a resource (for example, "Category 1") that will be the parent of the new resource.
2) Go inside edit resource "Category 1".
3) Create a new resource (for example, "Child 1") through the quick resource creation window and in the Settings tab, select "Category 1" as the parent.
4) The error "The resource cannot be its own parent."
The quick window sees the new resource "Child 1" as editable "Category 1", which is not correct.
Perhaps this problem is relevant for other fields of the resource.

### Observed behavior

qew_parent

### Expected behavior
There shouldn't be any error.

### Environment
MODX 3.0.0-rc1

theboxer:
### What does it do?
Sends 0 instead of undefined as `currentid` to the parent select.

### Why is it needed?
The parent select component uses `MODx.request.id` as default value if no `currentid` is passed. So the parent select thinks you're trying to set a parent for the currently opened resource, not for the one in the quick create window.

### How to test
Check #15983

### Related issue(s)/PR(s)
Resolves #15983


Ruslan-Aleev (approved):
Seems to work as it should :)

JoshuaLuckers (approved):

PR #16011, opened 2022-01-28
Added icon classes for resources in resource group

Ruslan-Aleev:
### What does it do?
Added icon classes for resources in resource group.
Now the status of the resource, its type and template are clear.

Took as a basis - https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/Processors/Resource/GetNodes.php#L406

Before:
rg_before

After:
rg_after

### Why is it needed?
UX Improvements

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14216 (icons)

### DIFF
--- a/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
+++ b/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
@@ -13,6 +13,7 @@ namespace MODX\Revolution\Processors\Security\ResourceGroup;
 use MODX\Revolution\Processors\Processor;
 use MODX\Revolution\modResource;
 use MODX\Revolution\modResourceGroup;
+use ReflectionClass;
 
 /**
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
@cla-bot check

JoshuaLuckers (approved):
I'm not a big fan of the technical implementation but I like this improvement!

Mark-H:
Perhaps worth introducing a modResource->getIcon() or modResource->getTreeClasses or something like that, if we're going to be basically copying what's already in a different place. Not everything has to be DRY if a bit of copying is easier to maintain but from the looks of it this seems like a good candidate for a minor refactor.

Ruslan-Aleev:
I will be glad if you send commits with edits :)
Although, at first glance, the resource tree is no longer used (only resources and resource groups).

PR #16016, opened 2022-01-31
Add consistent ?mv={version} cache busting suffix to prevent browser cache issues

Mark-H:
### What does it do?
Adds ?mv={version} to registered css and js scripts that are prone to change between releases.

Note I've opted for ?mv instead of ?v because there are extras that already use their own ?v so this reduces the odds of conflicts.

### Why is it needed?
Avoid sticky browser caches like https://community.modx.com/t/modx-3-0-0-package-management-processor-not-found/4908

### How to test
Confirm correct version string gets added and things still work. 

### Related issue(s)/PR(s)
N/a

### DIFF
--- a/manager/templates/default/header.tpl
+++ b/manager/templates/default/header.tpl
@@ -18,7 +18,7 @@
 <script src="{$_config.manager_url}assets/ext3/adapter/ext/ext-base.js"></script>
 <script src="{$_config.manager_url}assets/ext3/ext-all.js"></script>
 {/if}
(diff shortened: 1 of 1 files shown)


rthrash:

https://community.modx.com/t/modx-3-0-0-package-management-processor-not-found/4908/5

JoshuaLuckers (approved):

Ibochkarev (approved):

Mark-H:
Oof, I only just noticed this is missing the key part of the PR. Why did nobody question if that one tweak was all there was to it 😆

PR #13773, opened 2018-02-09
Redesigned manager login

Mark-H:
### What does it do?
This implements a new manager login screen. Designed by Rinze van der Brug, front-end by @christianseel, back-end/js by yours truly.

schermafbeelding 2018-02-09 om 23 49 24

The background and logo images can be changed with settings, and some new lexicons can be used to change the help text. 

### Why is it needed?

Modernize the manager login. 

### Related issue(s)/PR(s)

N/a

PR #16020, opened 2022-02-01
Cleanup unused background image logic (#16014)

### LINKED ISSUE #16014: Finish changing the background for the login
### Summary
1) Add pictures for the seasons for the login form

There is a code in the login that changes the background form depending on the season.
See https://github.com/modxcms/revolution/blob/3.x/manager/controllers/default/security/login.class.php#L85

Don't know why this was done, but if this code is already there, then we just need to add pictures and it will work.
It also makes sense to change backgrounds for major versions of MODX, so that you can tell which version of MODX now from the background (as in Mac OS, for example).

2) Add a setting to set your own backgrounds (both by default and seasonal).

As an option, set the path to the folder with backgrounds.

### Why is it needed?
To add beauty :)

Mark-H:
### What does it do?
Removes code that's not actually used.

### Why is it needed?
#16014 was opened to suggest implementing the remaining aspects of the login background, with seasonal custom backgrounds, however as it's [really easy to do this from plugins and settings](https://docs.modx.com/3.x/en/building-sites/client-proofing/manager-login) it seems better to just remove this code and suggest people to use/build extras like [DailyPhoto](https://modx.com/extras/package/dailyphoto) instead that can be updated and maintained separately.

### How to test
Make sure the backgrounds still work.

### Related issue(s)/PR(s)
Introduced in https://github.com/modxcms/revolution/pull/13773
Fixes #16014

### DIFF
--- a/manager/controllers/default/security/login.class.php
+++ b/manager/controllers/default/security/login.class.php
@@ -84,49 +84,20 @@ class SecurityLoginManagerController extends modManagerController
 
         // Define background
         $managerTheme = $this->modx->getOption('manager_theme', null, 'default');
-        $file_path = MODX_MANAGER_PATH . 'templates/' . $managerTheme . '/images/login/';
-        $file_url = MODX_MANAGER_URL . 'templates/' . $managerTheme . '/images/login/';
-        $this->modx->getVersionData();
-        $version = preg_replace('#-.*$#', '', $this->modx->version['full_version']);
         $background = $this->modx->getOption('login_background_image');
-        $season = '';
-        if (!$background || $background == 'auto') {
-            $month = date('m');
-            if ($month >= 3 && $month <= 5) {
-                $season = 'spring';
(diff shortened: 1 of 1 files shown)


Ruslan-Aleev:
Should it be removed? Maybe just add 4 pictures and let them change every 3 months, it will be more beautiful :) And we can change them for major versions.
And if a `login_background_image` setting is set, then we show only it.

Mark-H:
Yes, I think we should remove this and encourage people to use extras to extend it as desired instead, as mentioned in the PR description.

JoshuaLuckers (approved):

smg6511 (approved):
I agree with @Mark-H that further customization of the login image is best left to add-ins. Tested the change and the default image shows as expected and a user-supplied image via the `login_background_image` setting works as well.

theboxer (approved):

opengeek (approved):

PR #16026, opened 2022-02-02
Add consistent ?mv={version} cache busting suffix to prevent browser cache issues (part two)

Mark-H:
I somehow missed committing these changes for #16016, so here's part two. 😬 

### What does it do?
Adds ?mv={version} to registered css and js scripts that are prone to change between releases.

Note I've opted for ?mv instead of ?v because there are extras that already use their own ?v so this reduces the odds of conflicts.

### Why is it needed?
Avoid sticky browser caches like https://community.modx.com/t/modx-3-0-0-package-management-processor-not-found/4908

### How to test
Confirm correct version string gets added and things still work.

### Related issue(s)/PR(s)
#16016

### DIFF
--- a/core/src/Revolution/modManagerController.php
+++ b/core/src/Revolution/modManagerController.php
@@ -631,9 +631,10 @@ abstract class modManagerController
             }
 
             $o = '';
+            $suffix = '?mv=' . $this->_prepareVersionPostfix();
             // Add script tags for the required javascript
             foreach ($externals as $js) {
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

JoshuaLuckers:
How do we know it's complete now? 🤷‍♂️😋

Mark-H:
Well, at least I no longer have any uncommitted changes! 😆 Also:

> ## How to test

PR #16027, opened 2022-02-02
Changed brackets for category child elements

Ruslan-Aleev:
### What does it do?
In the tree, elements in brackets have IDs, but for categories (except for the list of categories themselves), the number of child elements is indicated. And since the brackets are the same, it can be confusing.

I changed the brackets to others to visually make it clear that these are different entities.
**Because the question is not very important, then just changing the brackets, in my opinion, is enough.**

IDs are indicated in red, the number of child elements is indicated in blue:
categoryElements

### Why is it needed?
UI improvement: it is now visually clearer that not ID is specified

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14598

### DIFF
--- a/core/src/Revolution/Processors/Element/GetNodes.php
+++ b/core/src/Revolution/Processors/Element/GetNodes.php
@@ -399,7 +399,7 @@ class GetNodes extends Processor
                 }
             }
 
-            $cc = ($category->get('elementCount') > 0) ? ' (' . $category->get('elementCount') . ')' : '';
+            $cc = ($category->get('elementCount') > 0) ? ' [' . $category->get('elementCount') . ']' : '';
             $nodes[] = [
                 'text' => strip_tags($category->get('category')) . $cc,
                 'id' => 'n_' . $map[0] . '_category_' . ($category->get('id') != null ? $category->get('id') : 0),
@@ -623,7 +623,7 @@ class GetNodes extends Processor
                 }
             }
 
-            $cc = $elCount > 0 ? ' (' . $elCount . ')' : '';
+            $cc = $elCount > 0 ? ' [' . $elCount . ']' : '';
 
             $nodes[] = [
                 'text' => strip_tags($category->get('category')) . $cc,


manager/assets/modext/widgets/element/modx.tree.element.js
```diff
@@ -451,11 +451,15 @@ Ext.extend(MODx.tree.Element,MODx.tree.Tree,{
 
     ,_getCategoryMenu: function(n) {
         var a = n.attributes;
+        var name = a.text;
+        if (a.data.category && a.data.elementCount > 0) {
+            var name = a.data.category + ' [' + a.data.elementCount + ']';
+        }
```
theboxer: Why is this block needed?

theboxer (changes requested):

manager/assets/modext/widgets/element/modx.tree.element.js
Ruslan-Aleev: I do not understand the question. PR changes brackets for nested category elements. Also in this block.

theboxer: I understand what the PR does, but I can't seem to see any difference when I remove this block and revert line 458/462 to use just `a.text`

Mark-H: isn't that for the right click menu?

theboxer: I see it in the right click menu and also in the tree with square brackets even without this block.

Ruslan-Aleev: Yes you are right. I first corrected the script in the tree, and then the processors (because the categories themselves and the categories of elements are slightly different entities) and did not notice that in `a.text` the value was already passed with the new brackets. I'll fix it now.

theboxer (approved):

JoshuaLuckers (approved):

opengeek:
Just a reminder to use the present tense (e.g. change instead of changed, fix instead of fixed, etc.) in your PR and commit descriptions. Thanks!

PR #15578, opened 2021-03-23
Add the EXTR_REFS flag to the extract function

sergant210:
### What does it do?
Adds the EXTR_REFS flag to the extract function.

### Why is it needed?
Allows you to pass to plugins not only objects but also variables by reference. Then the following code will work as expected
```
// modParser
$this->modx->invokeEvent('OnParseDocument', ['content' => &$content]);
```

### How to test
1. Create a plugin event with name "OnTestPassReference".
2. Create a snippet with content
```
$foo = 'bar';
$modx->invokeEvent('OnTestPassReference', ['foo' => &$foo]);
return $foo;
```
3. Create a plugin on the OnTestPassReference event with content
```
$foo = 'baz';
```
4. Put the created snippet in any resource `$foo = "[[!snippet]]"`.
5. Open this resource.
6. The page should display  `$foo = "baz"`.

PR #16030, opened 2022-02-03
Revert passing script variables by reference

Mark-H:
### What does it do?
https://github.com/modxcms/revolution/pull/15578 added the EXTR_REFS flag to the extract() function in modScript::process with the intention of allowing plugins to manipulated variables passed in to it. 

This PR reverts that back to how it was in 2.x.

### Why is it needed?

Unfortunately per https://community.modx.com/t/spform-how-to-change-language/4917/16 the pass by reference is causing issues when local variables in a snippet share the name of a script property or a copy of $scriptProperties is made. This is causing wildly unexpected results. 

### How to test

Two very simple snippets to reproduce the behavior are provided in the forum, both assuming the snippet is called with ```[[!mySnippet? &my_prop=`A`]]```:

```
<?php
$my_prop = 'B';
return $scriptProperties['my_prop']; // returns 'B'
```

```
<?php
$newArr = $scriptProperties;
$newArr['my_prop'] = 'B';

return $scriptProperties['my_prop']; // returns 'B'
```

The expected result, because there is no explicit change made to $scriptProperties, is `A`. However both return `B`. 

It also goes the other way, though this type of code would be less common:

```
<?php
$scriptProperties['my_prop'] = 'B';
return $my_prop;
```

### Related issue(s)/PR(s)

https://community.modx.com/t/spform-how-to-change-language/4917/15
https://github.com/modxcms/revolution/pull/15578/

### DIFF
--- a/core/src/Revolution/modScript.php
+++ b/core/src/Revolution/modScript.php
@@ -84,7 +84,7 @@ class modScript extends modElement
                 $scriptProperties = $this->xpdo->event->params = $this->_properties; /* store params inside event object */
                 ob_start();
                 unset($properties, $content);
-                extract($scriptProperties, EXTR_SKIP|EXTR_REFS);
+                extract($scriptProperties, EXTR_SKIP);
                 $includeResult = include $this->_scriptFilename;
                 $includeResult = ($includeResult === null ? '' : $includeResult);
                 if (ob_get_length()) {


rthrash:

https://community.modx.com/t/spform-how-to-change-language/4917/19

BobRay:
IMO, this should be applied. Passing $scriptProperties by reference is asking for trouble, since many plugins and snippets use variable names that are in the $sccriptProperties array.  I suspect that it also introduces opportunities for malicious code. since any variable in the $scriptProperties array could be modified.

opengeek:
@BobRay — It is already applied and live in the RC2 download.

BobRay:
Thanks. I should have checked.

Issue #13968, opened 2018-07-04
MODX 3: [Q] Space for content area width reduced by 25%. Why?


jonleverrier:
### Summary
MODX3 has _reduced_ the amount of available space for the content area width by 25% when compared to MODX 2.x.x. In MODX3 the content area now occupies 48% of screen width compared to 72% in MODX 2.x.x

The content area has been reduced by adding "publishing", "template" and "behaviour in menu" panels in view and by grouping the content field within the document summary.

The point of this github issue is to understand why reducing the content area for a content manager is a good idea and the thought process behind it. I'm also trying to keep this issue fact based vs my personal opinion. 

**_New MODX3 resource page:_**

![screen shot 2018-07-04 at 12 12 19](https://user-images.githubusercontent.com/2373940/42277651-64ee0c44-7f98-11e8-8b5a-75a6bbd78893.png)

#### 1. Use form customisation 

We have the ability to hide the new "publishing", "template" and "behaviour in menu" panels using form customisation.

Issues with this solution:
1. The content area stays the same width
2. A content editor will no longer have access to the hidden panels
3. There is no way to re-place the content in "publishing", "template", or "behaviour in menu" in to different regions

**_Publishing, template and behaviour in menu panels hidden with form customisation:_**

![screen shot 2018-07-04 at 12 15 20](https://user-images.githubusercontent.com/2373940/42274630-b3fb40c8-7f8d-11e8-880c-7bb57c0200ba.png)

#### 2. Create a custom plugin to make the resource area go full width
This is an option, but doesn't feel right that you essentially have to "hack" a new product from launch. A custom plugin will have to override the pixel width of the wrapper and field divs that ExtJs sets, into percentage values.

Issues with this solution:
1. A content editor still does not have access to the hidden panels
2. There is no way to re-place the content in "publishing", "template", or "behaviour in menu" in to different regions

#### 3. Close the resource tree
This is also an option and will perhaps just take time to get use to the extra click to browse resources or to train clients to use the arrow icon they have never used before - nothing that can't be overcome! 

Whilst closing the resource tree helps, the content area still occupies only 65% of the screen width compared to 96% in MODX 2.x.x

**_Resource page with the tree closed and panels hidden:_**
![screen shot 2018-07-04 at 12 18 08](https://user-images.githubusercontent.com/2373940/42275977-4c71d732-7f92-11e8-8d7c-e5cfb1a676c9.png)

### Other impact
[...]

PR #16037, opened 2022-02-04
Changed the height of the "Content" field

Ruslan-Aleev:
### What does it do?
Corrected the height of the content in the resource.

Before:
before

After:
after

### Related issue(s)/PR(s)
None


smg6511 (approved):
This works and is a nice fine-tuning when the right side regions are fully expanded. Note that the height will become uneven again if the right side is small enough (width-wise) to cause a line break in the "Behavior in menu" title or the "Hide from Menus" label. Regardless, this improves the view in probably the majority of desktop-size screen cases.

Ruslan-Aleev:
> This works and is a nice fine-tuning when the right side regions are fully expanded.

Yes, I agree, the problem persists on other sizes. In general, there are some issues with these floating areas - https://github.com/modxcms/revolution/issues/13968
This resource panel needs to be redone.

theboxer (approved):

PR #16039, opened 2022-02-04
Enable to save changed attribute name or container name in the user's extended fields

### LINKED ISSUE #16032: Can't Rename Attributes Containers
### Summary
While you can rename the Attributes Container in both user extended fields (going back to MODX 2.8.3) or dashboard widgets in MODX 3, the change doesn't save.

### Step to reproduce
Create attribute container and save.  Then try to rename it and save again.

### Observed behavior
After saving and reloading, the name reverts to its original name.

### Expected behavior
It should have the new name.

### Environment
MODX 2.8+ - MODX 3.0.0-rc2

halftrainedharry:
### What does it do?
Writes the changed name to the _name attribute_ of the tree node.

### Why is it needed?
In the "Extended Fields" tab of the user, you can change the name of an attribute or rename a container, but when you save the user, these changes are not saved.

### How to test
In the "Extended Fields" tab of the user, create an attribute and a container and click "Save". Rename the attribute and the container and click "Save" again. Reload the page to make sure, the changes are saved correctly.

### Related issue(s)/PR(s)
Resolves #16032

### DIFF
--- a/manager/assets/modext/widgets/core/modx.orm.js
+++ b/manager/assets/modext/widgets/core/modx.orm.js
@@ -180,6 +180,7 @@ Ext.extend(MODx.orm.Tree,Ext.tree.TreePanel,{
                         var nd = this.getSelectedNode();
                         nd.setId(r.name);
                         nd.setText(r.name);
+                        nd.attributes.name = r.name;
                         this.markFormPanelDirty();
                     },scope:this}
                 }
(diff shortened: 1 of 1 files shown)


smg6511 (approved):
This fixes the issue at hand. Tested with multiple levels of containers and attributes.

One nice-to-have _after_ this fix would be for the rename container window to show the original name in its name field. I did some logging on this to see if that would be easy to do, but I found there's some wacky stuff going on (at least on first glance) in terms of how this form panel and tree work. As such, that's outside the scope of this PR.

theboxer (approved):

PR #16041, opened 2022-02-07
Enable popper for the submenu too (opens on hover)

Jako:
### What does it do?
Apply Popper on subsubnav elements. Change the CSS to be used with Popper. The nav on System Settings is opened below of the entry on mobile. ~~A CSS rule avoids the subsubnav to grow larger than the viewport. The subsubnav content is scrollable.~~

### Why is it needed?
Fix a position out of the viewport for subsubnav elements. The use of Popper avoids the subsubnav be moved partly out of the viewport. Currently large subsubnavs are cut of at the bottom of the screen. Scrolling can be a future option for this (see https://github.com/modxcms/revolution/pull/16041#issuecomment-1036260889).

### How to test
Before: Reload the dashboard page with a viewport height of 700px. Open the 'manage' menu and look for the subnav of Reports. It is cut off on the bottom.
After: Run `grunt build` and do the same. The subnav of Reports is not cut off on the bottom.

### Related issue(s)/PR(s)
#16029 

### The following desktop issues are fixed

Language before (subsubnav: 86vh):
Language Before
Language after (subsubnav: 100vh - 12px):
Language After
Reports before:
Reports Before
Reports after:
Reports After
Long extras before:
Long Extras Before
Long extras after:
Long Extras After

### The following mobile issues are fixed

Language Before:
Language Before
Language After:
Language After
Long Extras Before:
Long Extras Before
Long Extras After:
Long Extras After

System Settings Before:
System Settings Before
System Settings After:
System Settings After

Issue #16029, opened 2022-02-03
Problems on small screens with left sidebar


deakonizer:
### Summary
On a small Laptop screen you cant see the whole Reports. Only the Site Schedule and you can't scroll down. Also the left Bar takes spaces on small screens. 

### Why is it needed?
You cant see the whole Logs and Reports. Here is a screenshot:

![afbeelding](https://user-images.githubusercontent.com/312944/152353833-df97f5c4-12eb-4145-af97-5e46a39e33df.png)

### Suggested solution(s)
The new Theme looks pretty cool and fresh, but, bring the Sidebar back to top. Maybe the User have the choice per Toggle (left or top), or maybe you have the choice for different Manager Themes (Radio Button)

Mfg
Marco

Jako:
You are right: The sidebar menu needs some additional work. The position of the subsubnav has to be calculated on size of the screen. The leftbar can be hidden with the arrow below the MODX logo.

Ruslan-Aleev:
> Maybe the User have the choice per Toggle (left or top), or maybe you have the choice for different Manager Themes (Radio Button)

This is too global changes to make (I personally don't like the menu on the left either).
If you have time and desire - we will be glad to receive PR from you :)

Mark-H:
Fixing the subsubnav with priority would definitely be useful, but I just want to state for the record that I seriously doubt we'd accept a massive change to the UI at this point in the release cycle. We're two weeks from the planned stable release, just not worth the risk breaking more things so short to the deadline.

Jako:
The subsubnav position is a bug. And there is another issue that has to be fixed too working on the bug: the navigation bar has to be reloaded/recalculated after resizing the window.

smg6511:
As @Jako said, there's definitely a bug there in the calculations that needs to be fixed. As to the left menu in general, I've never been a fan of it (from a usability standpoint), but have also known that as @Mark-H said it would be a major undertaking to change it — one that would _almost_ put us back at "square one" UI-wise. If one had a lot of time on their hands however, an optional manager theme could be developed, which would be a first and be great but a whole lot of work ;-)

Jako:
I am working on this. It is fine on large screens but the menu currently has issues on mobile screens (between 640px and 960px width).

PR #16042, opened 2022-02-07
Enable popper for the submenu too (opens on click)

Jako:
Same as #16041 but with click instead of hover.

The click menu has an issue, when the existing menu entry (which contains the submenu) has an action or a javascript handler attached. Then opening the submenu with a click will also run the action or the javascript.

If this change is accepted, the menu generating script has to be changed and existing menu entries with a submenu and an action/javascript have to be modified/moved with an update script.

Maybe this change is better moved to 3.1, but it is somehow breaking and for this reason it can be better added in 3.0.

Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/16029

### DIFF
--- a/_build/templates/default/sass/_navbar.scss
+++ b/_build/templates/default/sass/_navbar.scss
@@ -268,40 +268,43 @@
       }
     }
 
-    & li:hover ul ul,
-    & ul li:hover ul ul,
-    & ul ul li:hover ul ul {
-      display: none;
-    }
-
-    & li:hover ul,
-    & ul li:hover ul,
-    & ul ul li:hover ul,
-    & ul ul ul li:hover ul {
-      display: block;
-    }
-
     &.active {
       opacity: 1;
       visibility: visible;
     }
 
     .modx-subsubnav {
+      max-height: 99vh;
+      overflow-y: auto;
       border: 1px solid $navbarBorder;
       box-shadow: $boxShadowBig;
       border-radius: $borderRadius;
       background: $white;
-      display: none;
+      opacity: 0;
+      visibility: hidden;
       list-style: none;
       position: absolute;
-      left: 295px;
-      top: 0;
       z-index: 24;
-    }
-    &.modx-subnav-usernav {
-      .modx-subsubnav {
-        top: initial;
-        bottom: 0;
+      &.active {
+        opacity: 1;
+        visibility: visible;
+      }
+      &-arrow {
+        border: 8px solid transparent;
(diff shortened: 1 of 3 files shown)


Ruslan-Aleev:
The click menu does not work conveniently; nested items are difficult to open (you need to aim directly at the triangle).
Try clicking **"Refresh URLs"** in Manage -> Clear Cache.

Jako:
I have added some information into the PR description, which were only posted in the core-contributers channel on slack.

Ruslan-Aleev:
Another option is to make the triangle more noticeable, make it like a button: add a background, indent, to make it easier to click on it (although it’s not entirely clear how this will work on mobile devices).

Jako:
I will think about that triangle solution. It has to be separated by a vertical line then.

Ruslan-Aleev:
Thanks!
However, I'm not sure, but is it worth wasting your time? In my opinion, the current hover option is already familiar, maybe we should choose it? And clicks add unnecessary actions, which is worse for UX.

Jako:
Hover is not very usable on touch systems. And touch screens are getting bigger and bigger. So limiting it to mobile sizes is kind of bad for the future. That's the issue solved by a clickable menu.

Jako:
But it is nice, that you have detected the clickable region in the triangle, that does not create a click event on the menu entry itself. Hopefully this region can be made larger.

Ruslan-Aleev:
> Hover is not very usable on touch systems. And touch screens are getting bigger and bigger. So limiting it to mobile sizes is kind of bad for the future. That's the issue solved by a clickable menu.

Well, working in MODX on touch screens or on mobile devices is a strange idea (at least for now) :)
And mobile devices understand hover by long pressing, but for desktop clicks, in my opinion, are not convenient.
Anyway, these are just my thoughts...

Jako:
Think about editing a MODX backend with a tablet. Currently the menu is switching from an indented menu to the hover menu at 960px.

ExtJS 3.4 is old. I am not sure, if hover with a long click is implemented inside (if that's not a system thing).

Maybe touch and hover can be switched with a system setting. The CSS is currently the same. The scripts differ in two sections.

Ruslan-Aleev:
> ExtJS 3.4 is old. I am not sure, if hover with a long click is implemented inside (if that's not a system thing).

In my opinion, this is system-browser behavior, does not depend on extJS.

> Maybe touch and hover can be switched with a system setting. The CSS is currently the same. The scripts differ in two sections.

Yes, good option.

Issue #15927, opened 2021-12-04
А static element can't be saved


sergant210:
### Summary
А static element can't be saved.

### Step to reproduce
Create a snippet, check the "Is static" checkbox and try to save it.

### Observed behavior
![static element](https://user-images.githubusercontent.com/4679725/144697723-7d68f73b-f9f0-49df-b13d-2d5f04335aaf.png)

And I got an error message in the log
> [2021-12-04 07:50:33] (ERROR @ C:\Projects\modx3.local\core\src\Revolution\Sources\modMediaSource.php : 562) Unable to retrieve the last_modified for file at location: C:/Projects/modx3.local/core/components/pdotools/elements/snippets/snippet.pdoresources.php. filemtime(): stat failed for C:\Projects\modx3.local\C:/Projects/modx3.local/core/components/pdotools/elements/snippets/snippet.pdoresources.php

There is invalid file path - `C:\Projects\modx3.local\C:/Projects/modx3.local/core/components/pdotools/elements/snippets/snippet.pdoresources.php`.

### Environment
MODX3 Beta 2.
Windows 10.

Ibochkarev:
**1) If the media source is not selected and you are trying to save, then we get an error**

![изображение](https://user-images.githubusercontent.com/2138260/144734857-4b2b6f52-d131-4abb-8cd2-471178622923.png)

I also get an error when trying to save a text snippet

**2) If you tried to save and you don't have the source selected, and then you select it, we get this error**

![2021-12-05 14 05 52](https://user-images.githubusercontent.com/2138260/144738772-5a800aba-8017-4a72-ac92-e59dd100bac6.jpg)


``` shell
[2021-12-05 10:55:45] (ERROR @ /home/s28269/www/core/src/Revolution/Sources/modMediaSource.php : 585) Unable to read file from location: home/s28269/www/assets/static/snippets/test.snippet.php. file_get_contents(/home/s28269/www/home/s28269/www/assets/static/snippets/test.snippet.php): failed to open stream: No such file or directory
[2021-12-05 10:55:45] (ERROR @ /home/s28269/www/core/src/Revolution/Sources/modMediaSource.php : 562) Unable to retrieve the last_modified for file at location: home/s28269/www/assets/static/snippets/test.snippet.php. filemtime(): stat failed for /home/s28269/www/home/s28269/www/assets/static/snippets/test.snippet.php
[...]

PR #16043, opened 2022-02-07
Fix Static File Creation Issue

### LINKED ISSUE #16005: MODX 3.0.0-rc1 install, now getting filemtime() errors
### Summary
I upgraded my MODX 3 test site to the new release candidate. I have snippets set up as static files. Running these scripts now fails and pages don't finish loading.

### Step to reproduce
Install rc1. No idea what else might be involved here

### Observed behavior
The error log fills up with statements like this for each snippet:
Unable to retrieve the last_modified for file at location: [path to static files]/windowTitle.php. filemtime(): stat failed for  [path to static files]/windowTitle.php

Opening and resaving the snippets, either through the static file itself or through the Manager, has no effect.

### Environment
MODX 3.0.0-rc1, PHP 7.3, MySQL 5.7.36-cll-lve

smg6511:
### What does it do?

Corrected the file path reference used to create new static element files.

### Why is it needed?

Two connected problems are resolved here:
1. The logging issue referenced in issue #16005
2. Without this fix, a duplicate file (which is empty) is produced within the selected media source (but having the full server path). For example, if choosing the "Filesystem" media source and creating a static chunk at "assets/chunks/static-chunk-1.html", two files will be created (assuming an example server root path of "/var/www/site"):
-  assets/chunks/static-chunk-1.html (as expected)
-  assets/chunks/var/www/site/assets/chunks/static-chunk-1.html (empty duplicate at wrong location)

### How to Test

1. To verify the second problem above, create a static element or two on any other install/branch of 3.x. Take a look at your logs and the directories/files that get created.
2. Load _this_ PR branch and create one of each element with a static file source specified. For each, be sure to enter a little text/code into the element's code field to verify its content is actually saved to the target file. 
3. Monitor your manager error log to verify that entries do not get created when a static element is successfully created and saved. 
4. Verify in your file structure that there is no duplication of the static file in another area of the file system (see item 2 above).
5. Although probably not necessary (because I don't believe this code intersects with the parsing end of things), create a page that makes use of each static element to ensure the output is as expected.

### Additional Note

This PR does not seek to fix either of these two issues:
1. If you specify a file type that is not listed in the `upload_files` system setting  you will be unable to save the static element. A resolution for this, perhaps by implementing a new system setting for allowable static types and/or the ability to specify allowable types on a media source basis, will be proposed in a separate PR.
2. Selecting "None" as the media source currently results in unexpected behavior and absolute paths do not work (see also #15927). This will be addressed in a separate PR as well.

### Related issue(s)/PR(s)
Resolves #16005


core/lexicon/en/file.inc.php
```diff
@@ -73,6 +73,7 @@
 $_lang['file_folder_remove'] = 'Delete Directory';
 $_lang['file_folder_remove_confirm'] = 'Are you sure you want to delete the directory: "[[+directory]]"?<br />This could potentially break your website.';
 $_lang['file_folder_rename'] = 'Rename Directory';
+$_lang['file_info_metadata_new'] = "Unable to access metaData for the file “[[+file]]” because it is for a new element and has not been created yet. An attempt will be made to create the file automatically.";
```
Ruslan-Aleev: You are using the wrong quotes (both for the whole line and for the placeholder [[+file]]).
I don't think it's important, but it's better to keep them in the same style (since crowdin sometimes behaves strangely) :)

smg6511: Yeah, I'm picky about using single quotes everywhere possible ... missed that here, will change! On the other quotes, they are proper quotation marks for text that I've used elsewhere (using the utf-8 curly quotation marks instead of html entities).

Ruslan-Aleev: It is better to lead to one style.

smg6511: Made changes to the lex file. Should be all consistent now.

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -545,21 +545,25 @@ public function getObjectsInContainer($path)
...
             $data['size'] = $this->filesystem->fileSize($path);
             $data['mimetype'] = $this->filesystem->mimeType($path);
             $data['type'] = $data['mimetype'] === 'directory' ? 'dir' : 'file';
-
         } catch (FilesystemException | UnableToRetrieveMetadata $e) {
-            $this->xpdo->log(modX::LOG_LEVEL_ERROR, $e->getMessage());
+            if (!$isNew) {
+                $this->xpdo->log(modX::LOG_LEVEL_ERROR, $e->getMessage());
+            } else {
+                $this->xpdo->log(modX::LOG_LEVEL_ERROR, "\r\n\t" . $this->xpdo->lexicon('file_info_metadata_new', ['file' => $path]));
```
Jako: Why do yo create a new line and a tab indent here?

core/lexicon/en/file.inc.php
```diff
@@ -71,8 +71,9 @@
 $_lang['file_folder_parent'] = 'Parent Directory';
 $_lang['file_folder_parent_desc'] = 'Relative to the basePath of the media source';
 $_lang['file_folder_remove'] = 'Delete Directory';
-$_lang['file_folder_remove_confirm'] = 'Are you sure you want to delete the directory: "[[+directory]]"?<br />This could potentially break your website.';
+$_lang['file_folder_remove_confirm'] = 'Are you sure you want to delete the directory: “[[+directory]]”?<br>This could potentially break your website.';
```
Jako: I am fine using the right typography here, but the whole lexicons are different with double quotes.

core/lexicon/en/file.inc.php
```diff
@@ -117,7 +118,7 @@
 $_lang['upf_close'] = 'Close';
 $_lang['upf_close_desc'] = 'Close the window.';
 $_lang['upf_err_failed'] = 'Server is unavailable or internal server error occured.';
-$_lang['upf_err_filetype'] = 'Selected file extension isn\'t permitted.<br/>Please select files with following extensions: {1}';
+$_lang['upf_err_filetype'] = 'Selected file extension isn’t permitted.<br>Please select files with following extensions: {1}';
```
Jako: I am fine using the right typography here, but the whole lexicons are different with single quotes.

core/lexicon/en/file.inc.php
smg6511: Yes, I know. But you've got to start somewhere. I can address the other places in a different PR. I don't see it as a problem to update them over time, as it's not a highly-visible change.

core/src/Revolution/Sources/modMediaSource.php
smg6511: It's just easier to read the part of the output that's most important IMO.

core/lexicon/en/file.inc.php
smg6511: Resolving here, as we can discuss this above re double quotes.

Ruslan-Aleev: It is better to return it as it was, because all changed strings will overwrite translations in Crowdin.
After a few of these iterations, MODX will lose all translations.
p.s. And for the new line, leave it in the old style: firstly, it will look in one style, and secondly, there will be no confusion with quotes.

smg6511:
@Jako @Ruslan-Aleev - Ok guys, I walked back the small formatting changes and will let that go _for now_. Please take a look at this PR for its functionality, soon if you can ;-) 

This is the first of three (independent) PRs I'm working on the static elements issues that I think are pretty critical to the upcoming public release (as @JoshuaLuckers had indicated recently on Slack). The next will tackle the specification of static files that don't use a media source (and rely on an absolute path); this simply does not work currently and is a little more involved because of the move to Flysystem (for file handling). The last addresses the duplicate functionality (which has a bunch of problems as well, though maybe less critical).

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -545,21 +545,25 @@ public function getObjectsInContainer($path)
...
             $data['size'] = $this->filesystem->fileSize($path);
             $data['mimetype'] = $this->filesystem->mimeType($path);
             $data['type'] = $data['mimetype'] === 'directory' ? 'dir' : 'file';
-
         } catch (FilesystemException | UnableToRetrieveMetadata $e) {
-            $this->xpdo->log(modX::LOG_LEVEL_ERROR, $e->getMessage());
+            if (!$isNew) {
+                $this->xpdo->log(modX::LOG_LEVEL_ERROR, $e->getMessage());
+            } else {
+                $this->xpdo->log(modX::LOG_LEVEL_ERROR, $this->xpdo->lexicon('file_info_metadata_new', ['file' => $path]));
```
JoshuaLuckers: Isn't the message we log debug information instead of an error?

> Unable to access metaData for the file “test.php” because it is for a new element and has not been created yet. An attempt will be made to create the file automatically.

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -545,21 +545,25 @@ public function getObjectsInContainer($path)
...
 
     /**
-     * @param $path
+     * @param string $path
+     * @param bool $isNew
      *
      * @return array|bool
      */
-    public function getMetaData($path)
+    public function getMetaData($path, $isNew = false)
```
Mark-H: Changing the signature of a public method of a non-final class is a technical breaking change (as derivative classes could have extended the method). There are no derivatives of this method in core and it was only added in 3.0, so perhaps we can choose to ignore introducing a breaking change at this point in the release, or we could just re-evaluate why we're adding this and go with the easiest approach of just lowering the error level on the message?

Mark-H:
It took me 30 minutes to parse through all the discussion here and elsewhere and changes being proposed, but the key fix here is the one line change from `$sourceFile = $this->getSourceFile();` to `$sourceFile = $this->getStaticFileName();`. 

That prevents doubling up the path in isStaticSourceMutable, which fixes saving static elements. That seems like a logical patch, though incomplete for all currently known issues as indicated in the PR.

The error message is a red herring and IMO best just muted by lowering its level to WARN to avoid introducing a breaking signature change.

core/src/Revolution/Sources/modMediaSource.php
smg6511: It is, yes. A couple things here: I thought it might be helpful for it to be noticable for testing. I did have in mind knocking it down to INFO in the end. Also, I see this disappearing at some point, hopefully soon. I looked very carefully at how the file handling process works now and find it rather convoluted. Why is getMetaData _any_ part of checking whether a file exists or not? This method should strictly be used for what it implies its use is: Fetching the metadata of an existing file.

core/src/Revolution/Sources/modMediaSource.php
smg6511: I'd argue that's completely a technicality, as a default value is given for the new property and it's _only_ used to determine whether the thrown error is to be expected or not.

smg6511:
@Mark-H - Yes, quite incomplete. Just note that I'm being pushed (for better or worse) to execute changes in smaller chunks to just be able to get them reviewed, which is why this PR addresses only a _part_ of the problem.

core/src/Revolution/Sources/modMediaSource.php
opengeek: I'm with Mark on this one. I'd rather see the error level reduced here. The isNew parameter just doesn't even feel right on the getMetaData() method.

opengeek (changes requested):

core/src/Revolution/Sources/modMediaSource.php
smg6511: I agree, It's not right, nor is how this method is leaned on to figure out the writability of a file. Given that I see _this_ PR's fix as a temporary bridge to the real and hopefully much more complete fix I'm working on, it's not a big deal changing this the way you're requesting. Will update later today...

smg6511: OK, the requested change was made. All - please be aware that this PR _does not_ fix the problems with saving non-media source based files.

JoshuaLuckers (approved):

Jako (approved):

core/lexicon/en/file.inc.php
```diff
@@ -73,6 +73,7 @@
 $_lang['file_folder_remove'] = 'Delete Directory';
 $_lang['file_folder_remove_confirm'] = 'Are you sure you want to delete the directory: "[[+directory]]"?<br />This could potentially break your website.';
 $_lang['file_folder_rename'] = 'Rename Directory';
+$_lang['file_info_metadata_new'] = 'Unable to access metaData for the file “[[+file]]” because it is for a new element and has not been created yet. An attempt will be made to create the file automatically.';
```
opengeek: This isn't still necessary since the isNew param was removed, correct?

opengeek (changes requested):

core/lexicon/en/file.inc.php
smg6511: Yes, forgot that! Just removed it in new commit.

opengeek (approved):

PR #16045, opened 2022-02-08
Optionally determine context from config.context.php

opengeek:
### What does it do?
Allows the `/index.php` file to look for an optional `/config.context.php` file in the same directory which returns the context key to use when initializing the modX class.

### Why is it needed?
This will allow a site to initialize a default context other than `web` based on simple logic determined in the optional `/config.context.php` file.

### How to test
Add a `/config.context.php` file in the base_path where the MODX /index.php file is which returns a string representing the context key to use when calling `modX::initialize()`. Ensure that this initializes the context you expect.

### Related issue(s)/PR(s)
Related to #14962
Alternative to #15944

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ web.config
 # ignore all config.core.php
 config.core.php
 
+# ignore all config.context.php
+config.context.php
+
 # ignore concatenated js
 modext.js
 
--- a/index.php
+++ b/index.php
@@ -47,8 +47,12 @@ if (!is_object($modx) || !($modx instanceof \MODX\Revolution\modX)) {
 /* Set the actual start time */
 $modx->startTime= $tstart;
 
-/* Initialize the default 'web' context */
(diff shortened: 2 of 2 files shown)


theboxer (approved):

JoshuaLuckers (approved):
Works as intended however if the context that's defined doesn't exist it will throw a Fatal error and exception. Maybe we should consider using the web context by default in that case? Or show the styled error page instead.

opengeek:
> Works as intended however if the context that's defined doesn't exist it will throw a Fatal error and exception. Maybe we should consider using the web context by default in that case? Or show the styled error page instead.

This sounds like something that would need to be addressed inside initialize rather than in the index.php. Let's address that separately as it is the same issue regardless of how the wrong context is sent to the method.

modxcommunity:

https://community.modx.com/t/context-routing-in-modx-3/8008/1

PR #16050, opened 2022-02-09
Add explanatory text to create category

Ruslan-Aleev:
### What does it do?

1) Removed unnecessary lexicons

2) Added explanatory text in the category creation window

If you create a category inside the element tree (for example, in chunks), then after creation the entire tree is updated and collapsed. Moreover, if you expand the chunks, then the new category will not appear there (because it is empty) and it seems that the category has not been created. Because the category tree is collapsed, it is not at all obvious whether the category has appeared. Added a hint text in the category creation window, now it's clearer.

cat_create

### Why is it needed?
UX improvement

### Related issue(s)/PR(s)
N/A


core/lexicon/en/default.inc.php
```diff
@@ -52,6 +52,7 @@
 $_lang['categories'] = 'Categories';
 $_lang['category'] = 'Category';
 $_lang['category_create'] = 'Create Category';
+$_lang['category_create_desc'] = 'A category without elements is displayed only in the list of all Categories.';
```
smg6511:
```suggestion
$_lang['category_create_desc'] = 'This new category can be used to organize Chunks, Plugins, Property Sets, Snippets, Templates, and Template Variables. This category will only appear within, for example, the Chunks folder when at least one Chunk is assigned to it.';
```

smg6511: I agree that clarity is needed here and generally like what you've done. Consider this text edit, as I think it's even more clear. Because one would assume a category would get listed in the Categories folder, I did not mention it in my version of the description.

A small suggestion too would be to organize these moved entries (from category to default) at the bottom of the default topic, as I'd begun to do. The reason I did that is ultimately I think it's possible we might move them back at some point if we opt to load a wider array of lexicon topics in the various processors.

manager/assets/modext/widgets/windows.js
```diff
@@ -258,6 +258,10 @@ MODx.window.CreateCategory = function(config) {
             ,id: 'modx-'+this.ident+'-rank'
             ,xtype: 'numberfield'
             ,anchor: '100%'
+        },{
```
smg6511: I would suggest moving the description (what is now lines 261-265) to the top, making it the first `fields` object, with the following changed config:
```
{
    xtype: 'modx-description'
    ,html: _('category_create_desc')
}
```
This follows the established pattern of where the form panel descriptions go. Also, there's been a well-established _misuse_ of the `xtype: 'label'` from the beginning of MODx's Ext implementation that needs to be undone over time and at least discontinued.

Note: Sorry, the reference code showing above this comment isn't appearing correctly -- couldn't seem to get it selected in the right spot!

smg6511 (changes requested):
Looking good @Ruslan-Aleev, just a couple suggestions to improve upon what you've done.

core/lexicon/en/default.inc.php
Ruslan-Aleev: In my opinion, you complicate the text (you actually described the definition of the category) :)
For user it is clear what a category is and how they work, because the user has ALREADY called the category creation form.
The problem is that the new category appears only in the list of Categories when it is created, which is exactly what I was trying to make clear (maybe I should write a little more clearly, but I'm not a native speaker).

> A small suggestion too would be to organize these moved entries (from category to default) at the bottom of the default topic, as I'd begun to do. The reason I did that is ultimately I think it's possible we might move them back at some point if we opt to load a wider array of lexicon topics in the various processors.

Yes, a good idea for a separate PR, I also thought about it. There are many problems with lexicons: duplicates, unused, missing, etc.

manager/assets/modext/widgets/windows.js
Ruslan-Aleev: Thanks, I'll test this option.

core/lexicon/en/default.inc.php
Ruslan-Aleev: Here is a slightly more detailed version:

`A category for elements will only be displayed if at least one element is assigned to it (by default, all categories are displayed in the Categories section).`

smg6511: Yes, I thought about the fact that I was more fully defining what a Category is as well, which I elected to keep in there only because these pop up windows are the only place you can define them (as opposed to the other elements that have their own panels where each is defined at the top). Re the second sentence in my edit, although  I come at it from a different angle, I think it illustrates your point (just in a different way). If it still doesn't seem right to you, what about:

"Note that, until they have been assigned to one or more Elements, you will only see the Categories you create listed in the Categories folder."

manager/assets/modext/widgets/windows.js
Ruslan-Aleev: Yes, it looks interesting.
Old version with `xtype: 'label'`:

category_label

New version with `xtype: 'modx-description'`:

category_desc

smg6511: I think you're good to go with this.

jaygilmore: Sorry to chime in late here. Not that we have an official style guide written anywhere (besides the actual Lexicon files themselves) but Element(s) is a proper noun in MODX and should be capitalized when referenced as we do with all other MODX-specific terms such as Snippet, Chunk, Template, etc.

Avoiding the passive voice a variation and an edit: 

```
Categories will appear in the Categories folder once one or more Elements are assigned to that Category.
```

```
Note that, until they are assigned to one or more Elements, you will only see the Categories you create listed in the Categories folder.
```

I’m not sure the “Note that” is required here. I lean on being as direct as possible. But, I’m a 3/10 on that.

smg6511: @jaygilmore - The first variation isn't quite right, it should be:
`"A Category will appear in an Element’s folder once one or more of its Elements are assigned to that Category."`
The lean toward adding the word “only” (e.g., “will only appear”) comes from the desire to be really clear about the required condition, even though it's technically not needed.

Ruslan-Aleev: @smg6511 @jaygilmore I want to choose this variant:

`A Category will appear in an Element’s folder once one or more of its Elements are assigned to it (by default, all Categories, even empty ones, are displayed in the Categories section).`

Is this okay with everyone?

p.s. @jaygilmore You do not indicate that all categories, even empty ones, are in the Categories section, and this is important, because the user can create a bunch of unnecessary categories without knowing it. I added this.

smg6511: That sounds good to me!

jaygilmore: @Ruslan-Aleev @smg6511 good points there. Remind me what the `Element’s folder` refers to in the UI?

smg6511: Chunks, Templates, TVs, Plugins, and Snippet folders. It would also apply to the Property Sets tree, as PSs are technically considered Elements as well.

Ruslan-Aleev: @jaygilmore I didn’t quite understand the question, here the `Element’s folder` means “Any sections of elements” (templates, chunks, TV, etc.) Or are you asking about something else?

To be honest, I like the variant earlier more, it is clearer, in my opinion. With the last comments it will be like this:

`A Category will only be displayed if at least one Element is assigned to it (by default, all Categories, even empty ones, are displayed in the Categories tree).`

p.s. It's not clear what to call these "sections" of elements. Are they tabs, sections, folders, tree or what?

smg6511 (approved):
Looks good to me!

Jako (approved):

Issue #14226, opened 2018-12-30
[bug] Form Customization: order of the fields in "Set Information and Fields" tab, id field


Ruslan-Aleev:
### Summary
1) In the "Set Information and Fields" tab (Edit set) of the "Form Customization" section, the order of the fields must be corrected in accordance with the resource editing form:
**modx-resource-main-left
modx-resource-main-right
modx-resource-content
modx-page-settings-left
modx-page-settings-right
modx-page-settings-right-box-left
modx-page-settings-right-box-right**

2) In the "modx-resource-main-left field" there is an "id" field, but it doesn't work. It seemed to me that it is responsible for displaying id in the title of the edited resource, but this is not so.
The "id" field must either be used in the title, or the id field must be removed.

![form_custom](https://user-images.githubusercontent.com/12523676/50547556-3c019180-0c55-11e9-88fe-bacdbd0ef38d.png)

### Environment
MODX 2.7.0

alroniks:
To show id in the title or nor managed by special system setting. So too many places of configuring that behavior sound worth for me than just delete it from FC. But, id field always exists in the resource forms as a hidden field (editing) and FC allows on not this hidden field. So point 2 unrelated I guess, because FC describes forms, not labels or titles of the page.

Ruslan-Aleev:
Fixed in https://github.com/modxcms/revolution/pull/16052

PR #16052, opened 2022-02-11
Changed the order of the fields in Form Customization

Ruslan-Aleev:
### What does it do?
1) Changed the order of the fields in Form Customization as in the resource form.

Before:
fcf_b

After:
fc_group

2) Set the headings of the regions in the same style.
The headings were called somewhere just "ID", somewhere "Region", which was confusing.
Now the "Region (ID)" is specified - now it is clear that this is a region, and its ID is involved in the rendering of forms.

### Why is it needed?
Now clearer when working with regions in FC

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14226

### DIFF
--- a/manager/assets/modext/widgets/fc/modx.panel.fcset.js
+++ b/manager/assets/modext/widgets/fc/modx.panel.fcset.js
@@ -240,14 +240,14 @@ MODx.grid.FCSetFields = function(config) {
         ,fields: ['id','action','name','tab','tab_rank','other','rank','visible','label','default_value']
         ,autoHeight: true
         ,grouping: true
-        ,groupBy: 'tab'
+        ,groupBy: 'tab_rank'
         ,plugins: [this.vcb]
         ,stateful: false
         ,remoteSort: false
         ,sortBy: 'rank'
         ,sortDir: 'ASC'
         ,hideGroupedColumn: true
-        ,groupTextTpl: '{group} ({[values.rs.length]} {[values.rs.length > 1 ? "'+_('fields')+'" : "'+_('field')+'"]})'
+        ,groupTextTpl: '{[values.rs[0].data.tab]} ({[values.rs.length]} {[values.rs.length > 1 ? "'+_('fields')+'" : "'+_('field')+'"]})'
         ,columns: [{
             header: _('name')
             ,dataIndex: 'name'
@@ -255,6 +255,10 @@ MODx.grid.FCSetFields = function(config) {
         },{
             header: _('region')+' ('+_('tab_id')+')'
             ,dataIndex: 'tab'
+            ,width: 100
+        },{
+            header: _('tab_rank')
+            ,dataIndex: 'tab_rank'
             ,width: 100
         },this.vcb,{
             header: _('label')
--- a/manager/assets/modext/widgets/fc/modx.panel.fcset.js
+++ b/manager/assets/modext/widgets/fc/modx.panel.fcset.js
@@ -253,7 +253,7 @@ MODx.grid.FCSetFields = function(config) {
             ,dataIndex: 'name'
             ,width: 200
         },{
-            header: _('region')
+            header: _('region')+' ('+_('tab_id')+')'
             ,dataIndex: 'tab'
             ,width: 100
         },this.vcb,{
@@ -310,7 +310,7 @@ MODx.grid.FCSetTabs = function(config) {
         ,plugins: [this.vcb]
         ,stateful: false
         ,columns: [{
-            header: _('tab_id')
+            header: _('region')+' ('+_('tab_id')+')'
             ,dataIndex: 'name'
             ,width: 100
         },this.vcb,{
@@ -433,7 +433,7 @@ MODx.grid.FCSetTVs = function(config) {
             ,dataIndex: 'default_text'
             ,editable: false
         },{
-            header: _('region')
+            header: _('region')+' ('+_('tab_id')+')'
             ,dataIndex: 'tab'
             ,width: 100
             ,editor: { xtype: 'textfield' }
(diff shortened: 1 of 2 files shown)


manager/assets/modext/widgets/fc/modx.panel.fcset.js
```diff
@@ -240,22 +240,26 @@ MODx.grid.FCSetFields = function(config) {
...
-        ,groupBy: 'tab'
+        ,groupBy: 'tab_rank'
         ,plugins: [this.vcb]
         ,stateful: false
         ,remoteSort: false
         ,sortBy: 'rank'
         ,sortDir: 'ASC'
         ,hideGroupedColumn: true
-        ,groupTextTpl: '{group} ({[values.rs.length]} {[values.rs.length > 1 ? "'+_('fields')+'" : "'+_('field')+'"]})'
+        ,groupTextTpl: '{[values.rs.length]} {[values.rs.length > 1 ? "'+_('fields')+'" : "'+_('field')+'"]}'
```
JoshuaLuckers: To me it doesn't make sense to remove the group name here. It now says "X fields" but it doesn't give any context about what it's referring to.

Ruslan-Aleev: Good point, but I couldn't display the old `tab` group there (not sure if that's possible), because fields are now grouped by `tab_rank`, not `tab`. Those if you leave `{group}` there, then `tab_rank` will be displayed.

Jako: Why did you change the groupBy then? If you want to sort primary by the tab rank, you have to change the sorting in the processor.

Ruslan-Aleev: @Jako I couldn't do it, because here the question is not sorting fields, but sorting groups in extJS.
Those in whatever sort order the data is coming, the grouping sets its sorting by the group name (or I didn’t figure it out).

alroniks:
Adding region to the table maybe is a good idea, it became more clear. But at the same time grouping by label `1 Field`,  `5 Fields` looks weird and I don't like how it looks.

Ruslan-Aleev:
I figured out how to display the region in the group title:

```
,groupTextTpl: '{[values.rs[0].data.tab]} ({[values.rs.length]} {[values.rs.length > 1 ? "'+_('fields')+'" : "'+_('field')+'"]})'
```

But this, in my opinion, is a strange solution, but it works :)

manager/assets/modext/widgets/fc/modx.panel.fcset.js
Jako: If you sort in the processor before https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/Processors/Model/GetListProcessor.php#L153 in         prepareQueryBeforeCount or prepareQueryAfterCount, you can sort the result by the tab rank.

Ruslan-Aleev: @Jako I will not argue with you, because my knowledge is not enough, but in my opinion, you are confusing sorting and sorting groups in extJS.
Send a commit if it's easier, my version also works (and tab can be set in the group header), I'll redo my version a little later.

Jako: Sorry, I thought it was a processor based grid. The query is created here https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/modFormCustomizationSet.php#L32

I am not absolutely sure, but if you want to have the groups sorted right, you have to sort the lines by the groups before sorting the lines itself.

Jako: According to the code there, it is sorted by Tab.rank first, so it is the LocalGrid that sorts it wrong. It is possible to have multiple sorts in sortBy? Is it possible to remove the sorting there? Because the data is already sorted right.

Ruslan-Aleev: As I understand it, in order to sort groups in extJS, you need to create a function that would override the default sorting, see https://stackoverflow.com/questions/28690480/extjs-5-grid-group-sorting
But in my opinion it's not worth it, my version with grouping by `tab_rank` works as it should.

Ruslan-Aleev:
Guys, I corrected the group label, and also left the regions in the grid (this is consistent with the setting of regions for TV and is clearer, as @alroniks pointed out).

**Now it looks like this:**
fc_group

p.s. But I didn't understand why `modx-resource-content` is always first (0 in `tab_rank`). And the position in the `core/model/schema/modx.action.fields.schema.xml` is not taken into account.

Jako (approved):
Works fine here. When is this xml file loaded into the database?

alroniks (approved):
Works for me as well.

manager/assets/modext/widgets/fc/modx.panel.fcset.js
Ruslan-Aleev: @Jako But, in general, it would be useful to know how to sort groups, it would be useful for other sections. Do you have any ideas?

Ruslan-Aleev:
> When is this xml file loaded into the database?

During the installation process, probably (not sure), the fields have moved, exactly as they are set in xml. BUT not `modx-resource-content` tab.

Here is the condition - https://github.com/modxcms/revolution/blob/3.x/_build/resolvers/resolve.actionfields.php#L32

manager/assets/modext/widgets/fc/modx.panel.fcset.js
opengeek: Can this discussion be resolved?

Ruslan-Aleev: @opengeek Done

PR #16060, opened 2022-02-16
Modify delete and undelete processors to respect syncsite setting

intersel:
### What does it do?
Makes resource delete, undelete, publish and unpublish processors to use syncsite property and the syncsite_default system setting.

### Why is it needed?
'syncsite' is a property that add the possibility to enable/disable the clearing of the cache in the resource create/update processors.
But it is not taken into account in the resource delete/undelete/publish/unpublish processors that would always clear the cache when called...
It would be now with this PR.

On the other hand, the system setting 'syncsite_default' seems to not be used to default syncsite property in the create/update  processors that is hard coded to false ...
This PR uses this system setting to default syncsite property in the processors.

### How to test
* change system setting "syncsite_default" to false and see how it now behaves when saving/changing resources with the cache...
* call the processors with "syncsite" property with true or false value 

### Related issue(s)/PR(s)
* #11980 about delete/undelete processors
* #12593

### DIFF
--- a/core/model/modx/processors/resource/delete.class.php
+++ b/core/model/modx/processors/resource/delete.class.php
@@ -253,7 +253,7 @@ class modResourceDeleteProcessor extends modProcessor {
 
     /**
      * Clear the site cache
-     * @return boolean site cache cleared status
+     * @return void
      */
     public function clearCache() {
 	$clear = $this->getProperty('syncsite',$this->modx->getOption('syncsite_default')) || $this->getProperty('clearCache',false);
@@ -265,7 +265,6 @@ class modResourceDeleteProcessor extends modProcessor {
                 'resource' => array('contexts' => array($this->workingContext->get('key'))),
             ));
         }
-        return $clear;
     }
 
     /**
--- a/core/model/modx/processors/resource/delete.class.php
+++ b/core/model/modx/processors/resource/delete.class.php
@@ -253,10 +253,10 @@ class modResourceDeleteProcessor extends modProcessor {
 
     /**
      * Clear the site cache
-     * @return void
+     * @return boolean site cache cleared status
      */
     public function clearCache() {
-		$clear = $this->getProperty('syncsite',$this->modx->getOption('syncsite_default')) || $this->getProperty('clearCache',false);
+	$clear = $this->getProperty('syncsite',$this->modx->getOption('syncsite_default')) || $this->getProperty('clearCache',false);
         if ($clear) {
             $this->modx->cacheManager->refresh(array(
                 'db' => array(),
--- a/core/model/modx/processors/resource/delete.class.php
+++ b/core/model/modx/processors/resource/delete.class.php
@@ -256,12 +256,16 @@ class modResourceDeleteProcessor extends modProcessor {
      * @return void
      */
     public function clearCache() {
-        $this->modx->cacheManager->refresh(array(
-            'db' => array(),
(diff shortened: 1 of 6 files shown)


core/model/modx/processors/resource/delete.class.php
```diff
@@ -256,12 +256,16 @@ public function fireAfterDelete(array $childrenIds = array()) {
...
+		$clear = $this->getProperty('syncsite',$this->modx->getOption('syncsite_default')) || $this->getProperty('clearCache',false);
+        if ($clear) {
+            $this->modx->cacheManager->refresh(array(
+                'db' => array(),
+                'auto_publish' => array('contexts' => array($this->workingContext->get('key'))),
+                'context_settings' => array('contexts' => array($this->workingContext->get('key'))),
+                'resource' => array('contexts' => array($this->workingContext->get('key'))),
+            ));
+        }
+        return $clear;
```
JoshuaLuckers: According to the docblock comment for this function it should return void. Either the docblock needs to be adjusted or this needs to be changed to return void.

intersel: well yes! perhaps usefull to have the status returned... then I'll change the comment...

JoshuaLuckers:
Oh wait this is a breaking change so we can't add the return in 2.x

JoshuaLuckers:
@intersel thanks!

Mark-H (approved):

opengeek:
I'm wondering if there are any potential side effects by not clearing the cache when deleted and published statuses are changed. If there are non-cacheable resources and/or elements involved, would there be no consequences? 🤔

opengeek:
Though I suppose that is moot since these could be changed via the update processor.

Mark-H:
My understanding is that disabling `syncsite` would mean you already need alternative caching/cache clearing mechanisms in place that warrant it not being done automatically by MODX. So while this would certainly have consequences, I don't think they'd be unexpected, and indeed isn't necessarily different from making the same changes via a standard update. Just makes it consistent.

opengeek (approved):

core/model/modx/processors/resource/delete.class.php
```diff
@@ -256,12 +256,15 @@ public function fireAfterDelete(array $childrenIds = array()) {
...
-            'context_settings' => array('contexts' => array($this->resource->get('context_key'))),
-            'resource' => array('contexts' => array($this->resource->get('context_key'))),
-        ));
+	$clear = $this->getProperty('syncsite',$this->modx->getOption('syncsite_default')) || $this->getProperty('clearCache',false);
+        if ($clear) {
+            $this->modx->cacheManager->refresh(array(
+                'db' => array(),
+                'auto_publish' => array('contexts' => array($this->workingContext->get('key'))),
+                'context_settings' => array('contexts' => array($this->workingContext->get('key'))),
+                'resource' => array('contexts' => array($this->workingContext->get('key'))),
```
opengeek, line 265: While in the process of porting this to 3.x, I noticed these lines were changed from the resource context_key to the working context. I think this was a mistake and an oversight. Does anyone else agree?

Mark-H: Oh, good catch. I definitely overlooked that, and don't see any definition of $this->workingContext anywhere either?

opengeek: Yeah. I'll fix that up and just push it directly.

Issue #14373, opened 2019-02-13
[enhancement] Remove unnecessary windows when creating new objects


Ruslan-Aleev:
### Summary
As @sottwell correctly pointed out in media sources (https://github.com/modxcms/revolution/issues/12853), there is an unobvious step in creating the source, **but the problem is more widespread**.

In MODX, when creating object, a quick creation form first pops up (usually with a name, description), for example, for contexts, see the picture below.

![create_1](https://user-images.githubusercontent.com/12523676/52708368-5d1f0700-2fa3-11e9-8fdc-23beca1169ff.png)

Then, to fully configure the context, you have to edit it in an expanded form through the context menu, see the picture below.

![create_2](https://user-images.githubusercontent.com/12523676/52708378-627c5180-2fa3-11e9-8e4a-334419ed51eb.png)

**Question: Why do I need a quick creation form, if all its fields (name, description, etc.) I can fill in an extended form?**

Quick create form creates unnecessary steps and clicks. And for users it is not clear what to do next, because an object appeared in the list, but how to configure it? There are no icons, there are no links, we already discussed this earlier - https://github.com/modxcms/revolution/issues/14125

This behavior with a quick creation form is in media sources, contexts, access settings, and most likely in other sections.
The logical behavior of the button (without a quick form) can be seen in the "Users" section - /manager/?a=security/user

ant-tar:
Good catch, as far as I remember it was discussed twice some time ago on forums.modx.com too.
I am absolutely convinced that the quick window here is redundant and can be removed. When you work with a lot of contexts inside one website this becomes noticeable and waste time.

petyagrill:
I think that the most optimal way is to redirect to the edit page after creating a media source or context in this window.

Ruslan-Aleev:
@petyagrill In my opinion this is not a very good option, because it will not be clear where the extended form came from, and the behavior is not normal and it will take longer.

petyagrill:
@Ruslan-Aleev  The method I proposed is simpler to implement, because in the case of a redirect, we will have an id of the entity and the rest of the functionality is already working. But if you click on the button to open an extended form (which you first need to create), then you will still need to redirect the first time you save it. What you offer is very similar to the creation of resources, but I believe that to solve this problem there is no need for such large changes.

Ruslan-Aleev:
@petyagrill Unfortunately, I do not know the technical limitations for opening an extended form, perhaps you are right and this is not easy to implement.
[...]

Issue #12853, opened 2016-01-15
UX Enhancement: Creating Media Sources not intuitive


sottwell:
### Summary
Creating and configuring a new Media Source requires two non-obvious steps.

### Step to reproduce
Create a new Media Source.

### Observed behavior
You get a simple form with three fields. Of itself this does nothing at all. You need to right-click and update to get the fields that actually define the Source.

### Expected behavior
There could at least be a "next" button to open the second form.

### Environment
All environments.

sottwell:
This does get a bit tricky, as other Sources such as the Amazon S3 source does load a completely different secondary form. So the Media Source has to be saved before the second form can be loaded.

sottwell:
Also, if changing the Source Type in the second form the actual form doesn't change until the page is refreshed. So an automatic change of the form needs to be triggered, much in the same way a change in TVs is triggered by changing a resource's Template.

PR #16061, opened 2022-02-17
Add media source creation section

Ruslan-Aleev:
### What does it do?
In MODX, when creating source, a quick creation form first pops up (with a name, description and source type). This pop-up fotm itself does nothing at all. You need to right-click and update to get the fields that actually define the source.

Now there is no such problem for the source, in the future I will correct other sections (creating a context, for example).

source_create

**p.s. Please test everything carefully, I did rely on other sections of MODX, but I might have missed something.**

### Why is it needed?
Significantly improves UX, unnecessary actions disappear.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/12853
https://github.com/modxcms/revolution/issues/14373


smg6511:
I see what you're getting at, but I think you could make it better. From your screen capture, my first thought was that something went wrong when you end up on the new source page _without_ the properties grid showing (until clicking save). IMO, if you're going to change it, the ideal workflow solution would be:

1. In the sources grid, add a _Quick Create_ button that triggers the popup and have the _Create_ button trigger a full editing page.
2. For the QC popup, provide a couple more key fields (`basePath` and `basePathRelative`) so that it could at least be useful in quickly scaffolding one's back end sources structure.
3. Further, in the popup, to match the pattern of the (duplicate) Element QCs I'd consider adding a switch to redirect to new source. This could be something to consider adding for the QC create windows too. This would really be more of a nice-to-have offered to the user for edge cases rather than a feature one would typically use or need.

If it was still decided that (even with additional fields) a QC for media sources was not useful, I'd go further and ensure the Create button leads to the full editing panel with properties grid.

Ruslan-Aleev:
My goal is to remove the extra window, and make the section consistent with other MODX sections (see creating a "User", "Widget Panel", etc).
You propose global changes that should be discussed in a separate issue (not sure if the source section needs to be changed so much).

smg6511:
I understand your motivation to get rid of the window, and that's up to you and other reviewers whether to keep or get rid of it. My main request is to remove the extra step that's still there, which is having to save the source before seeing its properties grid.

PR #15946, opened 2021-12-24
Added remembering filter properties of deleted resources grid in browser url

Ruslan-Aleev:
### What does it do?
- Added the selected filter properties in the browser url.
- Fixed processor displaying deleted resources in the trash.
In the trash, it was possible to find an **undeleted** resource (any resource from the tree) through a search :)
- Added a parent to the resource grid.
The relationship between the deleted parent and the undeleted child resources will be a little clearer. Because many resources are not deleted but appear in the grid and will be deleted due to the parent being deleted.
- Removed extra padding for heading in trash.

trash

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15942
https://github.com/modxcms/revolution/pull/15935
https://github.com/modxcms/revolution/pull/15186
https://github.com/modxcms/revolution/pull/15185
https://github.com/modxcms/revolution/pull/15184
https://github.com/modxcms/revolution/pull/15183
https://github.com/modxcms/revolution/pull/15182
https://github.com/modxcms/revolution/pull/15181
https://github.com/modxcms/revolution/pull/15115
https://github.com/modxcms/revolution/issues/14086

PR #15942, opened 2021-12-19
Added remembering filter properties of lexicons grid in browser url

Ruslan-Aleev:
### What does it do?
- Added the selected filter properties in the browser url.
- Now the language in lexicons is equal to the current language of the manger panel, and not en by default.
- Fixed the error loading the list of languages when resetting the filter.
If you chose a different namespace (where there are fewer languages than in the manager), and then reset the filter, then even for namespace == core the list of languages will be from the previous namespace. This is true for 2.x and now.
- Code fixes (pagination, comments).

lexicons

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15935
https://github.com/modxcms/revolution/pull/15186
https://github.com/modxcms/revolution/pull/15185
https://github.com/modxcms/revolution/pull/15184
https://github.com/modxcms/revolution/pull/15183
https://github.com/modxcms/revolution/pull/15182
https://github.com/modxcms/revolution/pull/15181
https://github.com/modxcms/revolution/pull/15115
https://github.com/modxcms/revolution/issues/14086

PR #16063, opened 2022-02-18
Add remembering filter properties of template TVs grid in browser url

Ruslan-Aleev:
### What does it do?
- Added the selected filter properties in the browser url.

tpl_tv_url

### Why is it needed?
Allow to copy/paste the state of the filter with the browser url.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15946
https://github.com/modxcms/revolution/pull/15942
https://github.com/modxcms/revolution/pull/15935
https://github.com/modxcms/revolution/pull/15186
https://github.com/modxcms/revolution/pull/15185
https://github.com/modxcms/revolution/pull/15184
https://github.com/modxcms/revolution/pull/15183
https://github.com/modxcms/revolution/pull/15182
https://github.com/modxcms/revolution/pull/15181
https://github.com/modxcms/revolution/pull/15115
https://github.com/modxcms/revolution/issues/14086

### DIFF
--- a/core/src/Revolution/Processors/Element/Template/TemplateVar/GetList.php
+++ b/core/src/Revolution/Processors/Element/Template/TemplateVar/GetList.php
@@ -47,14 +47,16 @@ class GetList extends GetListProcessor
 
         $category = (integer)$this->getProperty('category', 0);
         if ($category) {
-            $conditions['category'] = $category;
+            $conditions[] = ['category' => $category];
         }
 
-        $search = $this->getProperty('search', '');
-        if (!empty($search)) {
-            $conditions['name:LIKE'] = '%' . $search . '%';
-            $conditions['OR:description:LIKE'] = '%' . $search . '%';
-            $conditions['OR:caption:LIKE'] = '%' . $search . '%';
+        $query = $this->getProperty('query', '');
+        if (!empty($query)) {
+            $conditions[] = [
+                'name:LIKE' => '%' . $query . '%',
+                'OR:caption:LIKE' => '%' . $query . '%',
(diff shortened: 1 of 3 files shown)


smg6511:
@Ruslan-Aleev - Sorry it has taken so long for me to look at these as promised a while ago. I'm attaching a couple files with a more consolidated strategy for you to consider for this and the other grid filtering PRs you've been working on. I'm attaching instead of trying to comment on them inline here because it'll probably be much easier to see what I'm suggesting by just doing a diff on your end. Anyway, the main thing I'm doing here is taking advantage of class inheritance and the fact that the filtering and clearing of filters are doing essentially the same task over and over across all the grids. We don't need a method in every grid for these tasks, so I've created a couple methods in the base grid class. Take a look and let me know what you think...

[grid-files.zip](https://github.com/modxcms/revolution/files/8193708/grid-files.zip)

Also, haven't forgotten that the other important thing you were wanting help on was providing the ability to activate the correct tab easily and in a global manner. I'll be able to focus on that a bit over the next couple weeks.

UPDATE: I actually was able to develop something to handle the tabs as well, and am just working out the more difficult area where there are both horizontal and vertical tabs (I think the ACL user group area is the only place this happens). Anyway, hold off on incorporating the methods I created in the main grid file (attached above) as I've built on those a little bit.

Ruslan-Aleev:
@smg6511 Thank you for your work! I'm sorry I didn't write earlier.
Could you send your code as a separate PR? So that those who understand better in js could also comment.
Unfortunately, I don't have the opportunity to test it yet.

PR #16062, opened 2022-02-18
Rename the lexicon key and text of mail_smtp_prefix

Jako:
### What does it do?
Change the lexicon key and text for `mail_smtp_prefix`

### Why is it needed?
The current text is a bit misleading. This setting sets the value of `SMTPSecure` in PHPMailer 6. The description in PHPMailer is `What kind of encryption to use on the SMTP connection`.

### Related issue(s)/PR(s)
Improved version of #15527 (renaming the key too). Supercedes #15555 (see the discussion there).

PR #16065, opened 2022-02-18
Move the existing value of mail_smtp_prefix to mail_smtp_secure

Jako:
### What does it do?
Copy the existing value of mail_smtp_prefix to mail_smtp_secure if the new mail_smtp_secure key exist and remove the mail_smtp_prefix setting on success.

### Why is it needed?
The system setting mail_smtp_secure can be created during the update before the upgrade script is executed.

### Related issue(s)/PR(s)
#16062

### DIFF
--- a/setup/includes/upgrades/mysql/3.0.0-pl.php
+++ b/setup/includes/upgrades/mysql/3.0.0-pl.php
@@ -35,3 +35,4 @@ include dirname(__DIR__) . '/common/3.0.0-non-index-field-length.php';
 include dirname(__DIR__) . '/common/3.0.0-template-preview.php';
 include dirname(__DIR__) . '/common/3.0.0-remove-content-type-field.php';
 include dirname(__DIR__) . '/common/3.0.0-remove-logs-permission.php';
+include dirname(__DIR__) . '/common/3.0.0-update-smtp-system-settings.php';
--- a/setup/includes/upgrades/common/3.0.0-update-smtp-system-settings.php
+++ b/setup/includes/upgrades/common/3.0.0-update-smtp-system-settings.php
@@ -19,22 +19,41 @@ $settingKeysMap = [
 $messageTemplate = '<p class="%s">%s</p>';
 
 foreach ($settingKeysMap as $setting) {
-    /** @var modSystemSetting $systemSetting */
-    $systemSetting = $modx->getObject(modSystemSetting::class, [
+    /** @var modSystemSetting $existingSetting */
+    $existingSetting = $modx->getObject(modSystemSetting::class, [
         'key' => $setting['old_key']
     ]);
-    if ($systemSetting instanceof modSystemSetting) {
-        $systemSetting->set('key', $setting['new_key']);
-        if ($systemSetting->save()) {
-            $this->runner->addResult(
-                modInstallRunner::RESULT_SUCCESS,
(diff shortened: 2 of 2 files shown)


Ruslan-Aleev:
Maybe the script should be called wider (`3.0.0-update-system-settings-keys.php`, for example) so that in the future it can be used for other system settings?

setup/includes/upgrades/mysql/3.0.0-pl.php
```diff
@@ -35,3 +35,4 @@
 include dirname(__DIR__) . '/common/3.0.0-template-preview.php';
 include dirname(__DIR__) . '/common/3.0.0-remove-content-type-field.php';
 include dirname(__DIR__) . '/common/3.0.0-remove-logs-permission.php';
+include dirname(__DIR__) . '/common/3.0.0-update-smtp-system-settings';
```
Ruslan-Aleev: Here, probably, .php needs to be specified.

opengeek: > Maybe the script should be called wider (`3.0.0-update-system-settings-keys.php`, for example) so that in the future it can be used for other system settings?

No. These are specific to a change and are thus atomic to that change.

opengeek:
> Maybe the script should be called wider (`3.0.0-update-system-settings-keys.php`, for example) so that in the future it can be used for other system settings?

No. These are specific to a change and are thus atomic to that change. They should not be reused.

setup/includes/upgrades/mysql/3.0.0-pl.php
opengeek: Slight oversight. I'm guessing this was not tested before being submitted. (squints at @Jako)

Ruslan-Aleev:
> No. These are specific to a change and are thus atomic to that change. They should not be reused.

I understand, yes, but, for example, there are already too many atomic scripts for updating system settings (see the current state).
Wouldn't it be easier to create a common script (one for updating settings, another for deleting settings, a third for updating keys), and add sys keys in this files in the future? And for future versions, just copy the structure of these scripts (but without the code).
It's just a suggestion, maybe it's overkill.

setup/includes/upgrades/mysql/3.0.0-pl.php
Jako: Sorry, but those update scripts are not easy to test.

opengeek: Welcome to my nightmare!

opengeek:
> I understand, yes, but, for example, there are already too many atomic scripts for updating system settings (see the current state). Wouldn't it be easier to create a common script (one for updating settings, another for deleting settings, a third for updating keys), and add sys keys in this files in the future? And for future versions, just copy the structure of these scripts (but without the code). It's just a suggestion, maybe it's overkill.

Actually, the only reason these are even broken out is because we used to have to have separate scripts for MySQL and sqlsrv and these files served as a way to share certain specific parts that were "common" to both. This is no longer the case. So technically, moving forward, we can put all of the upgrade logic directly into the single upgrade file.

alroniks:
> So technically, moving forward, we can put all of the upgrade logic directly into the single upgrade file.

Better to move to approach with migrations, and every migration should be atomic and revertable (usually rollback is rare case, but ideally it would be nice to have such ability).

Issue #15893, opened 2021-11-08
Don't list page types the user does not have permission to create


Mark-H:
### Summary

After #15888 is applied, the new create resource modal with template preview still lists page types the user does not have permission to create. 

### Step to reproduce

Apply #15888 or wait until that's been merged.

Restrict access to create a certain type of resource, eg static resource or weblink, with the permissions. 

Click the create resource button and try to create the restricted type.

### Observed behavior

Selecting the type is allowed (but it still fails server side).

### Expected behavior

Disallowed types should not be offered to the user.

### Related issue(s)/PR(s)

#15888 

### Environment
Any

JoshuaLuckers:
FYI: PR #15888 has been merged.

PR #16067, opened 2022-02-20
Have the template picker respect new resource action permissions

muzzwood:
### What does it do?
Adds an optional `restrict_actions` parameter to the `System/Derivatives/GetList` processor. Allowed values are `new`,`edit` or `delete`. This allows for checking newly introduced resource action permissions.

### Why is it needed?
The resource type dropdown in the template picker was not respecting these new permissions.

### How to test
Create a user that does not have one or more of these permissions: 
`new_weblink`, `new_symlink` or `new_static_resource`
Then open the template picker and check the resource type list.

### Related issue(s)/PR(s)
#15893

### DIFF
--- a/core/src/Revolution/Processors/System/Derivatives/GetList.php
+++ b/core/src/Revolution/Processors/System/Derivatives/GetList.php
@@ -20,6 +20,8 @@ use xPDO\Om\xPDOObject;
  */
 class GetList extends Processor
 {
+    const RESTRICT_ACTIONS = ['new', 'edit', 'delete'];
+
     /**
      * @return bool
      */
@@ -35,8 +37,43 @@ class GetList extends Processor
     {
         $this->setDefaultProperties([
             'class' => '',
+            'restrict_actions' => '',
         ]);
         return true;
+    }
+
+    /**
+     * @param array $descendents
+     * @return array
+     */
+    protected function checkRestrictions(array $descendents): array
+    {
+        $skip = explode(',', $this->getProperty('skip'));
+        $actions = explode(',', $this->getProperty('restrict_actions'));
+
+        $map = [
+            'MODX\Revolution\modWebLink' => 'weblink',
+            'MODX\Revolution\modSymLink' => 'symlink',
+            'MODX\Revolution\modStaticResource' => 'static_resource',
+        ];
+
+        foreach ($actions as $action) {
+            if (!in_array($action, self::RESTRICT_ACTIONS)) {
+                continue;
+            }
+
+            foreach ($descendents as $descendent) {
+                if ($descendent === 'MODX\Revolution\modDocument') {
+                    continue;
+                }
+
+                if (!$this->modx->hasPermission($action . '_' . $map[$descendent])) {
+                    $skip[] = $descendent;
+                }
+            }
+        }
+
+        return $skip;
     }
 
     /**
@@ -49,8 +86,8 @@ class GetList extends Processor
             $this->failure($this->modx->lexicon('class_err_ns'));
         }
 
-        $skip = explode(',', $this->getProperty('skip'));
         $descendants = $this->modx->getDescendants($class);
+        $skip = $this->checkRestrictions($descendants);
 
         $list = [];
         foreach ($descendants as $descendant) {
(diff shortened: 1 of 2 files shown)


core/src/Revolution/Processors/System/Derivatives/GetList.php
```diff
@@ -35,10 +37,44 @@ public function initialize()
...
+    protected function checkRestrictions(array $descendents): array
+    {
+        $skip = explode(',', $this->getProperty('skip'));
+        $actions = explode(',', $this->getProperty('restrict_actions'));
+
+        $map = [
+            'MODX\Revolution\modWebLink' => 'weblink',
+            'MODX\Revolution\modSymLink' => 'symlink',
+            'MODX\Revolution\modStaticResource' => 'static_resource',
+        ];
```
JoshuaLuckers: It's better to use `::class` to get the FQC instead of using a string.

core/src/Revolution/Processors/System/Derivatives/GetList.php
```diff
@@ -35,10 +37,44 @@ public function initialize()
...
+
+        foreach ($actions as $action) {
+            if (!in_array($action, self::RESTRICT_ACTIONS)) {
+                continue;
+            }
+
+            foreach ($descendents as $descendent) {
+                if ($descendent === 'MODX\Revolution\modDocument') {
```
JoshuaLuckers: It's better to use `::class` to get the FQC instead of using a string.

Ibochkarev:
@muzzwood Please make the suggested changes.

core/src/Revolution/Processors/System/Derivatives/GetList.php
muzzwood: Thanks, done!

core/src/Revolution/Processors/System/Derivatives/GetList.php
muzzwood: Thanks, done!

Mark-H (dismissed):

opengeek:
@muzzwood — any chance you could take a stab at rebasing this on top of the latest 3.x branch? I attempted to but it was not obvious to me how to. If you have any questions or issues on how to rebase, please don't hesitate to ask.

muzzwood:
@opengeek Sure, I'll try and look at that tomorrow 👍🏻

muzzwood:
Rebased to current 3.x, albeit a few days later than promised. 🙈

smg6511:
I'm beginning to take a look at this and note that, to fully address the issue, there are two other places where the derivatives' visibility needs to be controlled:
- The combo under Resource > Settings
- The Resource tree's context menu; I see you may already be on to this, since your proposed new constant and method take edit and delete perms into account.

smg6511:
~~There's one critical issue I just detected that may demand this be done a little differently: Only MODX-native derivatives show in the window combo. Third-party ones, such as Collections, do not show.~~
_Update on the above: It turns out this bug is not caused by this PR's changes, it's another issue in the same js file (in the picker's listeners)._

It also found that this change causes an error when editing a Resource (perhaps this only happens when non-native derivatives are present, I'm not sure), because the combo definition, at the least, needs to be updated in the Resource Settings (as mentioned above).

Sorry to be the bearer of bad news :-(  You probably don't need it, but I can lend a hand if you want.

When tackling this again, be aware that a recently-merged change to the template picker window will require you to rebase again. When you do, you'll want to just accept the _current_ change and re-add your `baseParams` config in the correct place (up near the top of the window definition).

smg6511:
@muzzwood - Hey, not meaning to step on your toes here with my alternate PR I just submitted. It's just that I found the way to do what you wanted was totally different than how you initially approached it and seemed simpler to just submit that solution rather than relay it to this PR somehow.

muzzwood:
Hi @smg6511 ,
No problem at all. 
If it's part of a bigger issue, I'm happy for you to take care of it.
Closing! :)

smg6511:
Ok, great ;-) If you have a few moments in the near future, give it a review to verify all's working. Cheers!

PR #16069, opened 2022-02-20
Stay on the current page after language switch

Jako:
### What does it do?
It adds a new handler method for the language menu entries that collects the request parameter of the current page (in MODx.request) and includes them into the language redirect. Otherwise fallback to the manager base URL.

### Why is it needed?
Fix the issue that the user always returns to the manager base URL after switching the language.

### How to test
Since modx.js is changed you have to run `grunt build`. Afterwards change to any manager page i.e. manager/?a=system/settings and switch the language with the menu. The manager loads the system settings page with the switched language.

### Related issue(s)/PR(s)
Improve #15895 without the open redirect vulnerability, #15174, #14046

### DIFF
--- a/core/src/Revolution/modMenu.php
+++ b/core/src/Revolution/modMenu.php
@@ -103,9 +103,7 @@ class modMenu extends modAccessibleObject
                     strtoupper($code)
                 ),
                 'parent' => 'language',
-                'action' => 'language',
-                'params' => '&switch=' . $code,
-                'namespace' => 'core',
+                'handler' => 'MODx.switchLanguage("' . $code . '"); return false;',
                 'permissions' => ''
             ];
         }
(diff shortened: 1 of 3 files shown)


Ruslan-Aleev:
Thanks, It works!
But the issue remains with resetting the selected language when exiting the manager panel.
Can this be fixed within the current PR?

Jako:
What do you mean with the remaining issue when exiting the manager panel?

Ruslan-Aleev:
My current system language is "Russian", I choose, for example, "German", I work with the manager in "German", everything is fine. But then I exit the manager, and when I enter the manager, on the login form, everything will be in "Russian" again, and I again need to select "German".

Jako:
When you log out from the manger, the session is destroyed and the manager_language setting is gone. That's the issue here.

One option is to save this in a cookie or in the local storage, but thats not part of this fix and has to be done elsewhere. So please open an issue for this, if it does not exist.

Or you can create a user setting.

Ruslan-Aleev (approved):

opengeek (approved):

PR #16070, opened 2022-02-21
Refactor Static Element File Handling

smg6511:
### What does it do?

Made changes and additions to modElement, select intermediate Element processors, and modMediaSource, as well as added a new system setting. These proposed changes ... 
- Improve static file handling and makes non-media source-based statics work again
- Provide the same capability of automatically creating a typed-in file name (where non-existing intermediate directories are created on-the-fly) for non-media source files (user selects "None" as source) as is available for files with an assigned media source. Note that this behavior was introduced with the implementation of the Flysystem file handling library (3.x only), which only applies to media sources.
- Create a new system setting (`static_elements_filetypes`) to enable file types specifically for static elements (ones that you wouldn't want available globally, such as php).
- Ensure that a media source’s `allowedFileTypes` property is honored (it currently is not because its value is called in the wrong way).
- Ensure that only the proper file type is allowed for Plugins and Snippets (php), as well as Chunks and Templates (anything but php, for now).
- Ensure that if a user enters a file path cannot be written or is otherwise invalid that the Element is not saved. Currently, under this condition, an error is shown regarding the invalid file but the Element gets saved anyway (yet the user is left on the create form).
- Ensure that when a static Element is duplicated, its content is duplicated as well.
- Consolidate duplicate logic into new methods in modElement.

### Why is it needed?

There have been assorted issues that have cropped up as a result of the implementation of the Flysystem file handling library that need to be fixed.

### How to test

1. Ensure all caches are completely cleared.
2. Create a few _non_ static Elements to verify correct behavior.
3. Create a variety of static Elements, both by assigning a media source and by selecting "None" as media source, by typing in a path (that does not yet exist) to verify the file gets written and is written to the proper place. You should be able to create a blank file, as well as one in which you've entered content into the Element's code field (which should save to the file).
4. Repeat step 3 above, but this time create some files (either manually or via the file browser) to verify that pointing a static element to an existing file works as expected. 
5. When running steps 3 and 4 above, at least one media source with a `basePath` set outside of the webroot should be verified as well.
6. Duplicate a few Elements to verify correct behavior.
7. Try to make the operations above fail as well, by entering invalid or duplicate data, to verify the process is stopped and validation messages are shown as expected.
8. Change the `static_elements_filetypes` setting to verify it affects the ability to use specified file types as expected.
9. Change the `allowedFileTypes` setting in one or more media sources to verify that it affects the ability to use specified file types as expected. This is the setting that will take top precedence. If blank, the global `static_elements_filetypes` setting applies and, if that setting is blank, the fallback of `upload_files` setting will be honored.

### Related issue(s)/PR(s)

Prompted by issue #15927 and related to PR #16043.


core/src/Revolution/Processors/Element/Update.php
```diff
@@ -29,6 +29,61 @@ abstract class Update extends UpdateProcessor
     /** @var modElement $object */
     public $object;
 
+    protected $hasStaticFile = false;
+
+    public function initialize()
+    {
+        // Intitializing parent first, as we need the Element object created before moving forward
+        if (parent::initialize()) {
+            $className = array_pop(explode('\\', $this->classKey));
```
JoshuaLuckers: Only variables can be passed by reference. You need to introduce a variable here and pass that to `array_pop`.

core/src/Revolution/Processors/Element/Get.php
```diff
@@ -22,6 +22,61 @@
  */
 abstract class Get extends GetProcessor
 {
+    protected $hasStaticFile = false;
+
+    public function initialize()
+    {
+        // Intitializing parent first, as we need the Element object created before moving forward
+        if (parent::initialize()) {
+            $className = array_pop(explode('\\', $this->classKey));
```
JoshuaLuckers: Only variables can be passed by reference. You need to introduce a variable here and pass that to `array_pop`.

core/src/Revolution/Processors/Element/Create.php
```diff
@@ -29,6 +29,61 @@ abstract class Create extends CreateProcessor
     /** @var modElement $object */
     public $object;
 
+    protected $hasStaticFile = false;
+
+    public function initialize()
+    {
+        // Intitializing parent first, as we need the Element object created before moving forward
+        if (parent::initialize()) {
+            $className = array_pop(explode('\\', $this->classKey));
```
JoshuaLuckers: Only variables can be passed by reference. You need to introduce a variable here and pass that to `array_pop`.

core/src/Revolution/modElement.php
```diff
@@ -1221,4 +1431,145 @@ private function getStaticFileName()
...
+                    //     modX::LOG_LEVEL_ERROR,
+                    //     "\rmodElement::validateStaticFile dir not writable!
+                    //     dir: " . $fileDirectory
+                    // );
+                    return ['msgLexKey' => 'file_folder_err_perms'];
+                }
+                if (!$this->_source->checkFileType($fileName)) {
+                    return [
+                        'msgLexKey' => 'file_err_ext_not_allowed',
+                        'msgData' => [ 'ext' => array_pop(explode('.', $fileName)) ]
```
JoshuaLuckers: Only variables can be passed by reference. You need to introduce a variable here and pass that to `array_pop`.

JoshuaLuckers (changes requested):

core/src/Revolution/Processors/Element/Create.php
```diff
@@ -29,6 +29,61 @@ abstract class Create extends CreateProcessor
...
+                    $this->object->relayStaticPropertiesToMediaSource([
+                        'isStaticElementFile',
+                        'ignoreMediaSource',
+                        'staticPathIsAbsolute'
+                    ]);
+                }
+            }
+            return true;
+        }
+    }
```
JoshuaLuckers, line 40: You need to `return false;` here.

core/src/Revolution/Processors/Element/Update.php
smg6511: May I ask why? Not to challenge necessarily, just trying to see if there's something I need to learn here. It _does_ work this way.

core/src/Revolution/Processors/Element/Create.php
smg6511: Just curious why...what does returning true out of initialize do that we don't want. This would be helpful for my general knowledge.

core/src/Revolution/Processors/Element/Update.php
smg6511: Nevermind ... I just found an explanation. I see it's not so much that it won't work, but that it might trigger an E_NOTICE.

core/src/Revolution/Processors/Element/Create.php
smg6511: Ok, figured this one out too. It's weird and un-intuitive though ;-)

core/src/Revolution/modElement.php
smg6511: Got it, this will be changing in another way anyway.

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -1762,40 +1761,114 @@ protected function loadFlySystem(FilesystemAdapter $adapter)
...
-                ? explode(',', $this->xpdo->getOption('upload_media')) : [];
-            $allowedFileTypes = array_unique(array_merge($allowedFiles, $allowedImages, $allowedMedia));
-            $allowedFileTypes = array_map('trim', $allowedFileTypes);
-            $this->setOption('allowedFileTypes', $allowedFileTypes);
+            $uploadFileTypes = [];
         }
 
+        $allowedStaticFileTypes = !empty($staticElementFileTypes) ? $staticElementFileTypes : $uploadFileTypes ;
+
+        if ($this->isStaticElementFile && $this->ignoreMediaSource) {
```
opengeek, line 1808: A media source should have no knowledge of static elements. This is a collision of separate concerns. This needs to be handled in some other way.

smg6511: I'm totally on board with maintaining SOC, but I would argue the way this is designed already goes against that in a major way by mixing the file handling in with the media source. One way or another, we need to be able to tell `checkFileType` to use the static element file extensions setting when applicable. I think this way is cleaner than the alternative, which would be to add a parameter to `checkFileType` to be able to pass in a custom list of file types. In that scenario you end up having logic involved in file extension checking in two places, which is not ideal.

smg6511:
Some of the moves I've made in this PR have been questioned and, whether they end up being the best moves or not, I wanted to outline my motivations for putting this together the way I did and address some of the items @Mark-H and @opengeek have taken issue with. One thing to be aware of: Although I don't agree with the structure of this area (the fact that file handling is done within the media source classes in particular) and of the app-wide relationship of processors to the objects they work upon, I thought it best to try to stick within what's been established as best I could.

Note that this is by no means a declaration of "I'm right, y'all are wrong." It's just a verbose explanation of why I did what I did and additional background for the questions that have been raised.

### File Type Checking
Aside from the obvious need to ensure static element files work across the board, the main goal was also to introduce control over the acceptable file types separate from the core `upload_files`, as we didn't want to just allow php (and perhaps other file types such as js) to be acceptable system wide just to make static elements work again. Thus, a new setting was introduced here: `static_elements_filetypes`. Additionally, it was found that the media source `allowedFileTypes` property was having no effect at all, so I fixed that.

The file checking precedence is set like this, from lowest to highest priority:

* Use `upload_files` as the base specification
  * If _not_ working with a static element, `upload_images` and `upload_media` must also be taken into account
* If working with a static element, use `static_elements_filetypes` as the _final_ specification if the media source is set to "None" or as the _base_ specification if a media source is being used
* When a media source is assigned, its `allowedFileTypes` setting overrides the above and offers the finest-grained control

I believe that's pretty logical and produces the behavior expected and explains the complexity of the updated `checkFileType` method. @Mark-H  had mentioned preferring to see an early return from within the if/else structure that builds the final allowable types list. To do this, another method would need to be added though, separating the building of that list from the checking against it (I'd be fine with that; I was just trying to limit my changes to the class as much as I could). Achieving the above behavior necessitates the `modMediaSource` class have knowledge of what context the file is being written in (whether it's a static element or not). This has been a sticking point, but I argue that the un-ideal solution of setting the property `$ignoreMediaSource` on the class is dictated by the established structure. Taking a bird's eye view, this may seem unjustified; but look at how all of this goes together _very_ closely (regardless of my proposed changes) and you may think differently. Without a complete redesign and disentanglement of the file handling from the media source, I just don't see how else this can be achieved.

### Overuse of Getters
Left and right, MODx leans on get methods (often on the same data) instead of establishing object properties _once_ via its constructor or by other means early on. I'd understand this if the app was predominantly ajax powered and objects were changing over time during the same session. That's not the case though. AFAIK every CRUD action is discrete. As an example, in modElement, how many times is `getSource` called? Many. Why? Anyway, I raise this because the following line I introduced was questioned (it appears in the Create and Update processors):
`$this->object->staticFileAbsolutePath = $this->object->getStaticFileAbsolutePath();`

I don't disagree this is weird at first glance. I tried to find a way to set this in a more expected way but, again, the structure makes it very difficult. The reason I set it in this way is because the value it holds is referred to 11 times in `modElement`. I'd find it ridiculous to call `getStaticFileAbsolutePath` in that many places when I can just establish that value once. 

### The Hurdle of Field Errors Being Set by Processors, not the Data Objects in Question
Another move I made was to pass the processor into my `validateStaticFile` method. I did this because, to keep the ability to set placeholders on an error's message (lexicon), you can't simply return a boolean from the validation method — you'd have to return an array with a message and associated placeholders. I'd initially done it that way, but it was more involved and less clean.

### Pervasive Duplication
Lastly, I really dislike all the duplication of virtually identical functionality that can be found throughout the app. As such, any time that I find opportunity to consolidate functionality I do. The introduction of the `validateStaticFile` method mentioned above is an example of trying to get away from that duplication.

Mark-H:
Closing as my understanding is the original problem was fixed in #16098.

Issue #14254, opened 2019-01-10
[enhancement] System Settings: Sort for areas and settings


Ruslan-Aleev:
### Summary
System Settings is the largest and "intimidating" element in MODX for new users.

Need to make it a clearer and more logical:
1) **Need change the sorting for settings areas in order of logical connection and frequency of use.**
For example (see picture below), if I look for an area in the list, I see "phpThumb", "Proxy", but I do not see the area "Site", although it should be in the first place, in my opinion.
_Need to discuss the area and change their order._
It is also worth **adding an item ("idx" column) to the "Back-end Manager" area in which you can set the order of the areas yourself.**

![areas_sort](https://user-images.githubusercontent.com/12523676/50987069-f7fa7080-1521-11e9-9386-90eb920673ce.png)

2) **Need add a new "idx" column (or name it differently) to sort the settings for a specific area, again in the order of logical connection and frequency of use.**
Now the settings are sorted by "key", but at the same time logically related elements may not be close in the list, see the picture below.
_Need to discuss the settings and change their order._

![settings_sort](https://user-images.githubusercontent.com/12523676/50987071-f8930700-1521-11e9-919d-6bfa28f4ffb4.png)

PR #16076, opened 2022-02-23
Add a setting to set the order of System Settings Areas

Ruslan-Aleev:
### What does it do?
Add a setting to set the order of System Settings Areas.

Areas in System Settings are arranged alphabetically by their aliases. At the same time, their text names (and in translation) are displayed in the list of filters or in the grid, i.e. it is difficult to understand by what logic they are sorted.

And, for example, areas that should be visible to the user immediately and are often used, for example, "Site" are located at the bottom, which again is not convenient.

Added a setting that allows you to set your order of areas.

sys_setting_order

It might be worth discussing/thinking more about the order of areas that are set by default.

### Why is it needed?
UI/UX improvement.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14254

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -920,6 +920,15 @@ $settings['manager_favicon_url']->fromArray([
   'area' => 'manager',
   'editedon' => null,
 ], '', true, true);
+$settings['manager_settings_areas_order'] = $xpdo->newObject(modSystemSetting::class);
+$settings['manager_settings_areas_order']->fromArray([
+  'key' => 'manager_settings_areas_order',
+  'value' => 'site, editor, furls, gateway, file, caching, mail, manager, language, static_resources, static_elements, authentication, proxy, session, phpthumb, system',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'manager',
+  'editedon' => null,
+], '', true, true);
 $settings['manager_time_format'] = $xpdo->newObject(modSystemSetting::class);
 $settings['manager_time_format']->fromArray([
   'key' => 'manager_time_format',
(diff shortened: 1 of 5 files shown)


Jako:
Nice idea, but I would prefer a draggable grid in an extra tab in the system/settings page. That way, custom areas can be sortable too.

Ruslan-Aleev:
> ... but I would prefer a draggable grid in an extra tab in the system/settings page.

Yes, an interesting option, I'll see how it is implemented in TV for the template, maybe I can do it.
Or you send your commit if you know how to implement it :)

Jako:
In short:
- Create a system_area table (id(?), area, rank - index on area)
- Build an area dragging grid for a new tab in the system/settings table (getlist, sortindex processors are necessary).
- Getlist processor has to iterate over the system settings and has to be joined with the system_area table. Unknown areas (not yet created in the system_area table) have to be sorted at the end alphabetically.
- Sortindex processor has first to sort all areas like the getlist processor. Unknown areas have to be created in the system_area table with a rank value at that point. After that it has to update the rank value for the dragged area (and the other areas between the start and end of dragging).
- The areas in the system setting grid have to be sorted like in the getlist processor.
- Find a way to remove not anymore existing areas in the system_setting table from the system_area table.

smg6511:
Guys, IMO you're overthinking it. The combo should simply show the relevant areas in alpha (label) order. I addressed this as a part of one of my open PRs (#16089 - see commit [5410971](https://github.com/modxcms/revolution/pull/16089/commits/54109716423f0f205e2fc3e9e1615518f3fc3b14)). The combo has typeahead on, so long lists can be easily navigated in that way.

Ruslan-Aleev:
> The combo should simply show the relevant areas in alpha (label) order. I addressed this as a part of one of my open PRs (https://github.com/modxcms/revolution/pull/16089 - see commit [5410971](https://github.com/modxcms/revolution/pull/16089/commits/54109716423f0f205e2fc3e9e1615518f3fc3b14)).

@smg6511 Yes, you are right, your variant is easier to implement and understandable in the interface.
I'll test your PR and close this one then.

Jako:
I am fine with an alphabetic sorting. However, if a sorting index had been introduced, it should have been configurable and changeable via drag and drop.

Ruslan-Aleev:
>  However, if a sorting index had been introduced, it should have been configurable and changeable via drag and drop.

Yes, fair. Although for "System Settings", in my opinion, dragging and dropping is an unnecessary thing.

PR #16077, opened 2022-02-23
Update PHP versions in setup tests

opengeek:
### What does it do?
Updates the required and recommended PHP versions in the setup version check.

### Why is it needed?
The version checks were outdated, allowing the installation of a downloaded distribution on versions that will not work.

### How to test
Try to install a distribution (not from git) on any PHP version prior to 7.2.5 and confirm it fails.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/setup/includes/test/modinstalltest.class.php
+++ b/setup/includes/test/modinstalltest.class.php
@@ -66,8 +66,8 @@ abstract class modInstallTest
         $this->title('php_version', $this->install->lexicon('test_php_version_start') . ' ');
         $phpVersion = phpversion();
 
-        $recommended_version = "7.2";
-        $required_version = "7.0";
+        $recommended_version = "7.3";
+        $required_version = "7.2.5";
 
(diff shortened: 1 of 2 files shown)


Ibochkarev (approved):

theboxer (approved):

JoshuaLuckers (changes requested):
The required PHP version needs to be updated here as well:  https://github.com/modxcms/revolution/blob/beed7f890ff7e6d77c2d616fffa0d1d372f9c969/setup/provisioner/requirements.php#L12

opengeek:
> The required PHP version needs to be updated here as well:
> 
> https://github.com/modxcms/revolution/blob/beed7f890ff7e6d77c2d616fffa0d1d372f9c969/setup/provisioner/requirements.php#L12

Ugh. Why is this stuff duplicated in so many places? (rhetorical—don't answer!)

PR #16078, opened 2022-02-24
Fix static elements in 3.0

Mark-H:
In progress - will add to this tomorrow. Alternative to #16070 that's focused on the current problems rather than refactoring.

Any local filesystem based static source should write to the absolute path returned in getSourceFile without checking extensions.. media sources internally still restrict extensions for non-local elements.

### What does it do?

### Why is it needed?

### How to test

### Related issue(s)/PR(s)

### DIFF
--- a/core/src/Revolution/Processors/Element/Create.php
+++ b/core/src/Revolution/Processors/Element/Create.php
@@ -79,17 +79,10 @@ abstract class Create extends CreateProcessor
 
         if ($this->object->staticSourceChanged() || $this->object->staticContentChanged()) {
             if ($this->object->get('content') !== '' && !$this->object->isStaticSourceMutable()) {
-                $source = $this->object->getSource();
-                if ($source && $source->hasErrors()) {
-                    $this->addFieldError('static_file', reset($source->getErrors()));
-                } else {
-                    $this->addFieldError('static_file', $this->modx->lexicon('element_static_source_immutable'));
-                }
-            } else {
-                if (!$this->object->isStaticSourceValidPath()) {
-                    $this->addFieldError('static_file',
-                        $this->modx->lexicon('element_static_source_protected_invalid'));
-                }
+                $this->addFieldError('static_file', $this->modx->lexicon('element_static_source_immutable'));
+            }
+            elseif (!$this->object->isStaticSourceValidPath()) {
+                $this->addFieldError('static_file', $this->modx->lexicon('element_static_source_protected_invalid'));
             }
         }
 
--- a/core/src/Revolution/Processors/Element/Create.php
+++ b/core/src/Revolution/Processors/Element/Create.php
@@ -77,13 +77,11 @@ abstract class Create extends CreateProcessor
         $this->setElementProperties();
         $this->validateElement();
 
-        if ($this->object->staticSourceChanged() || $this->object->staticContentChanged()) {
-            if ($this->object->get('content') !== '' && !$this->object->isStaticSourceMutable()) {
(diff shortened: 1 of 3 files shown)


JoshuaLuckers:
If I create a static element with no media source set it creates the file in the wrong path.
E.g if I select a file using the Media Browser like `assets/test.chunk.tpl` and save the element it creates a new file in `manager/assets/test.chunk.tpl` and removes the content.

If I select a Media Source and save it returns an error and the content of the element (in the manager) is emptied and shows `false` as content when refreshed.

opengeek:
The media browser cannot be used when no media source is set. That needs to be enforced by the UX.

smg6511:
Yes, while the UI does need to be updated to use a plain textfield when the media source is set to "None," it has no bearing on the actual issues with creating and getting the file specified in the field. The UI has been this way for a long time though (maybe always).

opengeek:
> Yes, while the UI does need to be updated to use a plain textfield when the media source is set to "None," it has no bearing on the actual issues with creating and getting the file specified in the field. The UI has been this way for a long time though (maybe always).

I didn't say it did. I was responding to @JoshuaLuckers experience testing this. But it is a fairly significant UX failure, to be honest. Can the media browser be used without a media source? If so, then in the context of static elements, it needs to navigate based on base_path (or element_static_path, if set).

Mark-H:
I'll work on this again tonight. 

IIRC, the media browsers select event also includes the media source ID, so perhaps I can use that to make the media source dropdown match what was selected. That way the interface can stay the same without people having to manually change the source if they happen to choose a file from a different one.

Mark-H:
Comments from Jim's testing he sent to me in a DM. 

```
Hey Mark. Ok, I've put my testing cap on with PR #16078. ( Don't be mad at me! :wink: )Under two conditions, statics are working:

    The file being referred to is pre-existing, and
    A media source is assigned OR "None" is chosen with a correctly-formatted absolute path

Beyond that, I have a list of issues that you can decide whether they are of the blocking variety or not:Media Source "None"
Typed-in paths:

    relative path (e.g., "assets/dir1/file.html"): file ends up in "connectors/assets/dir1/file.html" and its content is not saved (both on create and update)
    absolute path (e.g., "/server/home/assets/dir1/file.html"): file gets created but its content is not saved (both on create and update)
    duplicating element with absolute path does not copy its content to new element

Paths with pre-existing files in place:

    relative path behaves same as #1 above (except gets empty file [the one duplicated in connectors dir]; and that's what the db refers to)
    absolute path (e.g., "/server/home/assets/dir1/file.html") displays file content but can not save new/edited content from the element's form

Media Source Used
Typed-in paths:

    relative path (e.g., "assets/dir1/file.html") using default Filesystem MS: Element saves but unable to save file (no real useful info in the log). This also illustrates the issue caused by saving the element before confirming the file is saved (or exists); you end up with the element saved, but have an error that keeps you on the create page.
    absolute path (e.g., "/assets/dir1/file.html") — same issue as in #1 above
    media source with an absolute basepath outside webroot: same issue as #1 above

Paths with pre-existing files in place:

    file is read fine, but can not be updated via the element's editing panel
    media source with an absolute basepath outside webroot: same issue as #1 above

General

    I think static_elements_basepath (and actually all the settings in the Static Elements area) was really intended for the automation mode and should not be used as the basis for new paths when not in that mode (especially when duplicating)
    Path shown when duplicating an element should start off with that of the source element (the one being duplicated) (see #1 above)
    Trying to create php file via media browser or Quick Create/Edit window fails (extension error). The browser doesn't pick up the allowedFileTypes setting in any given media source because its implementation is broken.
````

PR #16082, opened 2022-02-25
Update Element Panels' Static File/Preview Image UI

smg6511:
### What does it do?

Added new rendering methods for `static_file` and `preview_file` (templates only) fields so that when "None" is selected as the media source, the media browser is disabled and a regular textfield is shown for input.

### Why is it needed?

When a media source is set to "None," the media browser can not account for situations where one may typically use "None" as the source (such as referring to files outside the webroot via an absolute path).

### How to test

1. Ensure your browser cache is cleared and, if you have `compress_js` enabled, rebuild the manager template via `grunt build`.
2. Create one of each element and switch between "None" and your defined media sources to verify the field changes and saving and loading behaves as expected.

### Related issue(s)/PR(s)

Indirectly related to efforts in #16078 and #16070 to shore up the overall behavior of creating and managing static elements.


theboxer:
@smg6511 Why it's only for templates? The other elements will have same issue for the static_file input.

smg6511:
@theboxer It should be working for _all_ elements. I'll double-check later today, but please also be sure it's not a cache issue on your end.

theboxer:
bah, ok, it was cache or something, sorry about that

theboxer (approved):

manager/assets/modext/widgets/core/modx.panel.js
```diff
@@ -415,6 +415,177 @@ Ext.extend(MODx.FormPanel,Ext.FormPanel,{
...
+                    openTo: staticDir
+                }
+        ;
+        let newStaticFile,
+            changeFieldType = false,
+            updateFieldSource = false
+        ;
+        if (elType === 'template') {
+            // need these in method's global scope, so define with var instead of const/let
+            var   staticPreviewFileField = Ext.getCmp(`modx-${elType}-preview-file`),
```
theboxer, line 448: Either stick with var or const/let

manager/assets/modext/widgets/core/modx.panel.js
```diff
@@ -415,6 +415,177 @@ Ext.extend(MODx.FormPanel,Ext.FormPanel,{
...
+    /**
+     * @property {Function} onChangeStaticSource - Updates the static file field based
+     * on the chosen source.
+     *
+     * @param {Object} cmp - The media source field's Ext.Component object
+     * @param {String} elType - The MODX element type (i.e., tv, chunk, or snippet)
+     */
+    ,onChangeStaticSource: function(cmp, elType) {
+
+        const   staticFileField = Ext.getCmp(`modx-${elType}-static-file`),
```
theboxer, line 428: Why there are so many spaces between const and name? Also I'm more in favour not grouping variables this way (but it might be just personal preference)

theboxer (changes requested):

manager/assets/modext/widgets/core/modx.panel.js
smg6511: This is one place where it has to be var because of how block scoping works with const/let.

manager/assets/modext/widgets/core/modx.panel.js
smg6511: This is a preference of mine, as I find it a bit easier to read when all the variable names in a block are left aligned.

manager/assets/modext/widgets/core/modx.panel.js
theboxer: ugh, ok, this is not ideal approach, but if I consider the rest of the codebase... meh

theboxer (approved):

Jako (approved):

PR #16091, opened 2022-03-18
Fix saving static elements

muzzwood:
### What does it do?
Allows static elements to save regardless of the file extension if on a filesystem media source.

Beyond the changes Mark made, this essentially adds an "ignore file type mode" to `modFileMediaSource` that allows skipping the file extension checks. When the mode is not active, any saving to a media source is still subject to those checks.

### Why is it needed?
See issue https://github.com/modxcms/revolution/issues/15927

This is a blocker for MODX 3.0 being released. This is meant to solve just this issue and be kept simple.

### How to test
Try saving static elements with and without a media source.

### Related issue(s)/PR(s)
This is an extension of the draft PR #16078 by @Mark-H . The first commit is his work, I just copied it. ;)
There's also a more in-depth approach by @smg6511 #16070 but that also involved solving other issues.

### DIFF
--- a/core/src/Revolution/Processors/Element/Create.php
+++ b/core/src/Revolution/Processors/Element/Create.php
@@ -77,20 +77,11 @@ abstract class Create extends CreateProcessor
         $this->setElementProperties();
         $this->validateElement();
 
-        if ($this->object->staticSourceChanged() || $this->object->staticContentChanged()) {
-            if ($this->object->get('content') !== '' && !$this->object->isStaticSourceMutable()) {
-                $source = $this->object->getSource();
-                if ($source && $source->hasErrors()) {
-                    $this->addFieldError('static_file', reset($source->getErrors()));
-                } else {
-                    $this->addFieldError('static_file', $this->modx->lexicon('element_static_source_immutable'));
-                }
-            } else {
-                if (!$this->object->isStaticSourceValidPath()) {
-                    $this->addFieldError('static_file',
-                        $this->modx->lexicon('element_static_source_protected_invalid'));
-                }
-            }
+        if (
+            ($this->object->staticSourceChanged() || $this->object->staticContentChanged())
+            && !$this->object->isStaticSourceValidPath()
+        ) {
+            $this->addFieldError('static_file', $this->modx->lexicon('element_static_source_protected_invalid'));
         }
 
         return !$this->hasErrors();
--- a/core/src/Revolution/Processors/Element/Update.php
+++ b/core/src/Revolution/Processors/Element/Update.php
@@ -70,20 +70,11 @@ abstract class Update extends UpdateProcessor
         }
 
         /* can't change content if static source is not writable */
-        if ($this->object->staticSourceChanged() || $this->object->staticContentChanged()) {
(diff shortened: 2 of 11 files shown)


muzzwood:
Ugh. phpcs seems to be finding more issues each time I fix the reported ones.
I'm a tad worried that what was meant to be a straightforward PR is going to be obfuscated be all the syntax alterations. 🥲

Edit:
Turns out I had to set trim trailing white space to false in `.editorconfig` as it was clashing with phpcs
https://github.com/modxcms/revolution/blob/3.x/.editorconfig#L8


For anyone reviewing, just look at the first two commits to see the approach without all the syntax changes.

opengeek (dismissed):
I'm still of the opinion that a media source should have no concept of static elements or static resources or anything else internally. Could this be done as a generic option to disable the file type restrictions instead?

muzzwood:
Hi @opengeek, yeah that should be easy enough. Mark mentioned he'd prefer there be no mode at all, so I've got another approach in testing that does things the same way as 2.x. It only uses the media source to get the base path, concatenates, then writes via the cache manager. This way, there are no changes to media source behaviour, they're just being side-stepped.
I might push that up as a separate PR so everything's clear.

muzzwood:
Alternative PR is here: #16095

muzzwood:
Updated to make the mode naming generic, and remove any mention of static elements within the media source. 👍🏻

opengeek (commented):
Okay, this is getting close I think. Another question, though. Why would the ignore_file_type code be limited to modFileMediaSource? Shouldn't this be part of all media sources? Am I missing something obvious?

muzzwood:
The intention there was so it only worked with a local filesystem media source. But if I misunderstood and it should work for any kind, then moving that to modMediaSource shouldn't break anything.

opengeek:
> The intention there was so it only worked with a local filesystem media source. But if I misunderstood and it should work for any kind, then moving that to modMediaSource shouldn't break anything.

Unless the selection of media sources for static elements is limited to modFileMediaSources, then I think it must work for any modMediaSource.

muzzwood:
That sounds fair! I've just restricted the media source combobox on the element panels to only list media sources with the `class_key` of `MODX\Revolution\Sources\modFileMediaSource`.
This was done by adding an optional `filesystemOnly` parameter to the get list request and adding a conditional `where`. e.g.
```
if ($this->getProperty('filesystemOnly')) {
    $c->where([
        'modMediaSource.class_key' => modFileMediaSource::class
    ]);
}
```

opengeek:
It would be nice to see a couple of reviews/approvals so we can get it merged and release 3.0 ASAP.

opengeek:
Even the most basic save of a static element with no media source is failing for me.

JoshuaLuckers:
Is this PR still relevant?

muzzwood:
> Is this PR still relevant?

Sorry for the delay @JoshuaLuckers ! This and the alternative PR can probably be closed now. Right @opengeek ?

muzzwood:
No longer needed. Closing!

Issue #16034, opened 2022-02-04
MODX 3 - Incorrect word wrapping in the manager panel


Semdevmaster:
### Summary
![admin_styles](https://user-images.githubusercontent.com/2706508/152480628-0d2740dc-a5c6-4e6f-9361-06a1ce9aac91.gif)

### Environment
MODX Revolution RC2

PR #16093, opened 2022-03-18
Fix incorrect word wrapping in manager panel

matdave:
### What does it do?
Adjusts the break in vertical tabs to be less awkward.

### Why is it needed?
Currently vertical tabs will break on the letter instead of on words

### How to test
Described in issue #16034 

### Related issue(s)/PR(s)
#16034


matdave:
lol, going to try that again.

PR #16094, opened 2022-03-18
Fix incorrect word wrapping in manager panel

matdave:
### What does it do?
Adjusts the break in vertical tabs to be less awkward.

### Why is it needed?
Currently vertical tabs will break on the letter instead of on words

### How to test
Described in issue #16034 

### Related issue(s)/PR(s)
#16034


_build/templates/default/sass/_tabs.scss
```diff
@@ -321,7 +321,7 @@ ul.x-tab-strip-bottom {
           .x-tab-strip-text {
             line-height: 1.4;
             padding: 2px 0 2px 0;
-            word-break: break-all;
+            // word-break: break-all;
```
JoshuaLuckers:
```suggestion
```

JoshuaLuckers (changes requested):
To prevent conflicts it's recommended to not commit the build asset files.

JoshuaLuckers (approved):

PR #16095, opened 2022-03-19
Alternative fix for saving static elements in 3.0

muzzwood:
### What does it do?
This is an extension of #16078, and an alternative fix to #16091 and #16070 

### Why is it needed?
This fixes the issues described in #15927 

This is a different approach to the others in that it doesn't touch media sources except to get the base path.
Rather than attempt to write files through media source functions (and thus needing to check file extensions), it handles it the same way that 2.x does and writes static element files directly. 

### How to test
Create and save static elements, with and without a media source. Try doing it with files that are both writable and not. 

### Related issue(s)/PR(s)
See above.

### DIFF
--- a/core/src/Revolution/Processors/Element/Create.php
+++ b/core/src/Revolution/Processors/Element/Create.php
@@ -77,20 +77,11 @@ abstract class Create extends CreateProcessor
         $this->setElementProperties();
         $this->validateElement();
 
-        if ($this->object->staticSourceChanged() || $this->object->staticContentChanged()) {
-            if ($this->object->get('content') !== '' && !$this->object->isStaticSourceMutable()) {
-                $source = $this->object->getSource();
-                if ($source && $source->hasErrors()) {
-                    $this->addFieldError('static_file', reset($source->getErrors()));
-                } else {
-                    $this->addFieldError('static_file', $this->modx->lexicon('element_static_source_immutable'));
-                }
-            } else {
-                if (!$this->object->isStaticSourceValidPath()) {
-                    $this->addFieldError('static_file',
-                        $this->modx->lexicon('element_static_source_protected_invalid'));
-                }
-            }
+        if (
+            ($this->object->staticSourceChanged() || $this->object->staticContentChanged())
+            && !$this->object->isStaticSourceValidPath()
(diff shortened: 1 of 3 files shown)


opengeek:
I'm a little lost. If you define a static element with a selected media source, it would have to be written by the media source. If you do NOT select a media source, then it would be written like this. But this would NOT work if the media source you select is not specifically a local file media source, e.g. S3. While I personally do not understand why anyone would do that (and thus do not think using media sources for static elements is valid anyway), does this really fix the issue, or just create others?

muzzwood:
What you're saying definitely makes sense.
This approach attempts to keep things working along the same lines as 2.x, where it's writing the file outside of the media source (in modElement). 
To have it write via the media source, then either https://github.com/modxcms/revolution/pull/16091 or https://github.com/modxcms/revolution/pull/16070 would be preferable.
Anyway, thought I'd put this here as an option.

JoshuaLuckers:
> This approach attempts to keep things working along the same lines as 2.x, where it's writing the file outside of the media source (in modElement).

@opengeek if it's in line with how it works in 2.x I think this would be a good candidate for a temporary fix so we can release 3.0. We could then focus on PR #16070

opengeek:
> > This approach attempts to keep things working along the same lines as 2.x, where it's writing the file outside of the media source (in modElement).
> 
> @opengeek if it's in line with how it works in 2.x I think this would be a good candidate for a temporary fix so we can release 3.0. We could then focus on PR #16070

AFAIK, this is NOT how it works in 2.x. If a media source is selected, the media source is responsible for saving it. If it one is not selected, then it is saved directly.

muzzwood:
I'll be the first to admit it's quite possible I misread how it works. They're some rather heavy classes to work through.
So if that's the case, apologies for creating any confusion!  ;)
Though, my understanding for 2.x (at the time of creating this PR at least) was that if no media source is selected, it uses the full path that's manually typed in, and if a media source is selected, it takes the base path from it then appends the rest within modElement.
I've just made an update to the other PR, so it's ready to go if approved.

muzzwood:
No longer needed. Closing!

PR #16098, opened 2022-03-22
Revert static element logic to match 2.x branch

opengeek:
### What does it do?
Reverts all logic related to static elements made in the 3.x branch to match the logic in the 2.x branch.

### Why is it needed?
This attempts to resolve all of the static element issues in 3.x.

### How to test
Test static element creation and updating with static elements that are specified with and without media sources.

### Related issue(s)/PR(s)
This is an alternative solution to #16095, #16091, #16078, and #16070
It addresses issue #15927


smg6511:
There's just one problem I've detected with this after testing (but it is a buggy behavior already present in 2.x): If you type in an absolute path to what would be a new file (does not already exist) when media source = "None", the file gets created at `core/components/[the absolute path you entered]` instead of `[the absolute path you entered]`. My guess is that we'd just leave that be for now. If so, this is good to go at least as an interim fix.

opengeek:
> There's just one problem I've detected with this after testing (but it is a buggy behavior already present in 2.x): If you type in an absolute path to what would be a new file (does not already exist) when media source = "None", the file gets created at `core/components/[the absolute path you entered]` instead of `[the absolute path you entered]`. My guess is that we'd just leave that be for now. If so, this is good to go at least as an interim fix.

This is what I've been trying to convey for a while now. It has never wanted an absolute path. It has always expected a relative path to a `components_path` setting that defaults to `MODX_CORE_PATH . 'components/'` that could be overridden by an `element_static_path` setting. So this is a feature, not a bug.

smg6511:
That intent may be moot though, as I'm sure I'm not the only user who has used "None" for the express purpose of specifying an absolute path to a static file. That said, as before one can specify an absolute path to an already-existing file and it works.

smg6511 (approved):
This reversion works from my testing.

opengeek:
> That intent may be moot though, as I'm sure I'm not the only user who has used "None" for the express purpose of specifying an absolute path to a static file.

I'm sure you aren't. But a lot of people have used it as it stands expecting the files to be written in core/components. We can certainly look into adding the ability to specify an absolute path if it is needed, though I would personally rather see some improved UX around selecting/showing the path that will be prefixed.

smg6511:
> We can certainly look into adding the ability to specify an absolute path if it is needed, though I would personally rather see some improved UX around selecting/showing the path that will be prefixed.

OK, well, let's loop back around to this at some point after launch.

theboxer:
Works fine, even with gpm 🚀

theboxer (approved):

JoshuaLuckers:
What's the reason for the removal of `_build/test/Tests/Model/Element/modStaticElementTest.php`?

JoshuaLuckers (approved):
In the tests I've done it works similar to 2.x!

opengeek:
> What's the reason for the removal of `_build/test/Tests/Model/Element/modStaticElementTest.php`?

I didn't specifically decide to remove it. It was added in PR #14993, which was reverted as a whole.

PR #16099, opened 2022-03-23
Resource icons in resource groups tree

theboxer:
### What does it do?
Abstract logic for getting icon classes from `Resource\GetNodes` processor to the `modResource::getIconClasses`.
Use the new method in `Resource\GetNodes` and `ResourceGroup\GetNodes`.

Lock icon is not part of this - as it's specific for resource tree and has no real benefit in resource group tree, same with classes around permissions.

### Why is it needed?
Fixes the BUG that resource group's tree doesn't show resource icons. (100% viable for 3.0)

### How to test
Create resource group, add resource, check if it has an icon :)

### Related issue(s)/PR(s)
Different approach in #16011

### DIFF
--- a/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
+++ b/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
@@ -92,6 +92,7 @@ class GetNodes extends Processor
                             'leaf' => true,
                             'type' => modResource::class,
                             'iconCls' => implode(' ', $icons),
+                            'cls' => implode(" ", $resource->getStatusClasses()),
                         ];
                     }
                 }
--- a/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
+++ b/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
@@ -85,13 +85,13 @@ class GetNodes extends Processor
                     $resources = $resourceGroup->getResources();
                     /** @var modResource $resource */
                     foreach ($resources as $resource) {
+                        $icons = $resource->getIconClasses();
                         $list[] = [
                             'text' => $resource->get('pagetitle') . ' (' . $resource->get('id') . ')',
                             'id' => 'n_' . $resource->get('id') . '_' . $resourceGroup->get('id'),
                             'leaf' => true,
                             'type' => modResource::class,
-                            'cls' => 'icon-' . $resource->get('class_key'),
-                            'iconCls' => 'icon-file',
+                            'iconCls' => implode(' ', $icons),
                         ];
                     }
                 }
--- a/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
+++ b/core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
@@ -85,13 +85,12 @@ class GetNodes extends Processor
                     $resources = $resourceGroup->getResources();
                     /** @var modResource $resource */
                     foreach ($resources as $resource) {
-                        $icons = $resource->getIconClasses();
                         $list[] = [
                             'text' => $resource->get('pagetitle') . ' (' . $resource->get('id') . ')',
                             'id' => 'n_' . $resource->get('id') . '_' . $resourceGroup->get('id'),
                             'leaf' => true,
                             'type' => modResource::class,
-                            'iconCls' => implode(' ', $icons),
+                            'iconCls' => implode(' ', $resource->getIconClasses()),
                             'cls' => implode(" ", $resource->getStatusClasses()),
                         ];
                     }
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1607,9 +1607,17 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
     {
         $classes = [];
 
-        if (!$this->get('published')) $classes[] = 'unpublished';
-        if ($this->get('deleted')) $classes[] = 'deleted';
-        if ($this->get('hidemenu')) $classes[] = 'hidemenu';
+        if (!$this->get('published')) {
+            $classes[] = 'unpublished';
+        }
+
+        if ($this->get('deleted')) {
+            $classes[] = 'deleted';
+        }
+
+        if ($this->get('hidemenu')) {
+            $classes[] = 'hidemenu';
+        }
 
         return $classes;
     }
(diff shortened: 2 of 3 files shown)


Ruslan-Aleev:
Unfortunately, the resource status is not displayed in the group tree (publishing, deleting and displaying in the menu).

deleted_res

smg6511 (commented):
I've tested this and it fixes the missing icons issue. As @Ruslan-Aleev pointed out, the resource status is not shown. That's easy to add in here, but it occurred to me that a case can be made for not showing it; it potentially adds a bunch of visual clutter not necessarily needed for the function of this area (organizing the resources into groups). 

Note that @theboxer's solution here is better than #16011, as the abstraction into a new function is more lean and Ruslan's PR adds a whole bunch of unneeded css classes used in the main resource tree for permissions. (The only operations available in the groups tree is to add or remove a resource from a group.)

I can go either way here. We could introduce a system setting and leave it up to users whether statuses are shown in resource groups.

core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
```diff
@@ -85,13 +85,13 @@ public function process()
                     $resources = $resourceGroup->getResources();
                     /** @var modResource $resource */
                     foreach ($resources as $resource) {
+                        $icons = $resource->getIconClasses();
```
smg6511, line 88: If electing to add statuses:
```suggestion
                        $icons = $resource->getIconClasses();
                        $cls = [];
                        if (!$resource->get('published')) {
                            $cls[] = 'unpublished';
                        }
                        if ($resource->get('deleted')) {
                            $cls[] = 'deleted';
                        }
                        if ($resource->get('hidemenu')) {
                            $cls[] = 'hidemenu';
                        }
```

core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
```diff
@@ -85,13 +85,13 @@ public function process()
...
                         $list[] = [
                             'text' => $resource->get('pagetitle') . ' (' . $resource->get('id') . ')',
                             'id' => 'n_' . $resource->get('id') . '_' . $resourceGroup->get('id'),
                             'leaf' => true,
                             'type' => modResource::class,
-                            'cls' => 'icon-' . $resource->get('class_key'),
-                            'iconCls' => 'icon-file',
+                            'iconCls' => implode(' ', $icons),
```
smg6511, line 94: If adding statuses:
```suggestion
                            'iconCls' => implode(' ', $icons),
                            'cls' => implode(' ', $cls)
```

core/src/Revolution/modResource.php
```diff
@@ -1537,4 +1539,70 @@ public function clearCache($context = '')
...
+
+        if (substr($classKey, 0, 3) === 'mod') {
+            $classKey = substr($classKey, 3);
+        }
+
+        $iconCls = [];
+
+        $contentType = $this->getOne('ContentType');
+        if ($contentType && $contentType->get('icon')) {
+            $iconCls = [$contentType->get('icon')];
```
Mark-H: Shouldn't this (and the same logic on line 1572) be `$iconCls[] = $contentType->get('icon')` or is the point that we can only have a single icon between the content type and template?

This does seem to be the same in the original code so happy to let it slide if it's just the way things used to be.

core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
Mark-H: Would be better to put this logic in the `getIconClasses()` method, no?

Mark-H: Actually now that I think about it longer than 5 seconds, no, because `cls` is applied somewhere else than `iconCls`. So support for this suggestion.

smg6511:
@theboxer - bump - wondering if you're still interested in this or want someone else to pick it up?

smg6511:
@theboxer @Mark-H - Guys, what's happening with this one? I'm happy to help...

core/src/Revolution/modResource.php
theboxer: Adjusted, a relic from past

theboxer:
extracted the status classes as well

core/src/Revolution/modResource.php
```diff
@@ -1534,4 +1536,81 @@ public function clearCache($context = '')
...
+        return $iconCls;
+    }
+
+    public function getStatusClasses()
+    {
+        $classes = [];
+
+        if (!$this->get('published')) $classes[] = 'unpublished';
+        if ($this->get('deleted')) $classes[] = 'deleted';
+        if ($this->get('hidemenu')) $classes[] = 'hidemenu';
```
smg6511: A formatting nitpick, but we should conform to our own PSR rules:
```suggestion
        if (!$this->get('published')) {
            $classes[] = 'unpublished';
        }
        if ($this->get('deleted')) {
            $classes[] = 'deleted';
        }
        if ($this->get('hidemenu')) {
            $classes[] = 'hidemenu';
        }
```

core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
```diff
@@ -85,13 +85,14 @@ public function process()
                     $resources = $resourceGroup->getResources();
                     /** @var modResource $resource */
                     foreach ($resources as $resource) {
+                        $icons = $resource->getIconClasses();
```
smg6511: This and the next change: Another super small thing, but let's make consistent (no need for var declaration)
```suggestion
```

core/src/Revolution/Processors/Security/ResourceGroup/GetNodes.php
```diff
@@ -85,13 +85,14 @@ public function process()
...
                         $list[] = [
                             'text' => $resource->get('pagetitle') . ' (' . $resource->get('id') . ')',
                             'id' => 'n_' . $resource->get('id') . '_' . $resourceGroup->get('id'),
                             'leaf' => true,
                             'type' => modResource::class,
-                            'cls' => 'icon-' . $resource->get('class_key'),
-                            'iconCls' => 'icon-file',
+                            'iconCls' => implode(' ', $icons),
```
smg6511:
```suggestion
                            'iconCls' => implode(' ', $resource->getIconClasses()),
```

smg6511 (approved):
Looking good! I just have two really minor requests that have no bearing on the PR's functionality. Otherwise this is good to go for me and I'm approving as-is, but I'd personally like to see everyone conform to the formatting rules we have set out, applying them at the very least to the code they are touching.

opengeek (approved):

jaygilmore (approved):

PR #16102, opened 2022-03-25
[3.x] Fix static file field visibility in legacy Element panel UIs

### LINKED ISSUE #16096: [3.x] Static file field showing by default when creating an element.
### Summary
When opening the create element page (applies to all types), the static file field is visible even before the "is static" checkbox is checked.

Screenshot 2022-03-19 at 13-41-31 Create Snippet MODX Revolution

### Step to reproduce
Open a create new snippet/chunk/plugin page.

### Observed behavior
Toggling the is static checkbox behaves correctly, it's just the initial page load. I assume it's not set to hidden by default.

### Expected behavior
It should be hidden until the is static checkbox is clicked.

### Environment
Latest 3.x branch

smg6511:
### What does it do?
Made a few small tweaks to ensure static file field is not visible when "Is Static" is not checked. Also, fixed the following small issues found while testing:
- When creating a new static Element and the default media source is "None," the static file field was showing the file browser instead of a plain text field
- When working with a non-static Template, the static file field would appear when changing the Media Source
- When working with a Template with no source assigned, the Media Source displayed a blank value instead of "None"

### Why is it needed?
To correct previously undetected bug introduced in #16082.

### How to test
Create and edit each element type and
- toggle from static to non-static to verify static file visibility is as-expected
- toggle between Media Sources to verify the file browser only becomes available for statics _not_ set to "None"

### Related issue(s)/PR(s)
Resolves #16096.


JoshuaLuckers (approved):

Ruslan-Aleev (approved):
Visually everything works as it should :)

PR #16119, opened 2022-04-05
[Port to 3.x] Fix Various Issues with Form Panel Validation and Field Display - Part 3, Snippet Panel

smg6511:
_NOTE: This is part 3 of 6 of a revised submission of #15887. This should be reviewed after part 2 is merged (there are two small changes needed from that PR to make the locked switch appear correctly)._

### What does it do?

Ports enhancements to the Snippet form panel from PR #15146 to 3.x (including subsequent fixes by others). 
Changes include:
1. Updated panel JS to make form layout consistent with the layout introduced in the current TV panel, and utilize new shared methods introduced in the main FormPanel via part 1 of this PR set (already merged).
2. A handful of minor lexicon updates.

### Why is it needed?

Makes needed UI and consistency improvements to the Snippet panel.

### How to test

2. Create and update a few Snippets to verify behavior and appearance is as expected.

### Related issue(s)/PR(s)
Partial port of #15146


theboxer:
@smg6511 Same issue as on the part for Chunks (https://github.com/modxcms/revolution/pull/16128#issuecomment-1103996357) happens in snippets view as well. I wonder if the placement of the code copy should be adjusted, or if we'll just ignore the issue, as it happens only in specific case (and in different languages the case will be different)

PR #16126, opened 2022-04-06
Fix inability to create/update user password

### LINKED ISSUE #16124: Unable to specify new password for users
### Summary
Unable to specify new password for users, modx sends the password confirmation as specifiedpassword.
Generated html observed in Firefox and Edge:
grafik
modx.panel.user.js sent by server:
grafik

Manually changing the name of the second input from specifiedpassword to confirmpassword makes it possible to create a user with specified password.

### Step to reproduce
Install new MODX 3.0.0-pl.
Try to create a user and specify a password.

### Observed behavior
Submitting the user creation fails without error message.

### Expected behavior
User should be created with the specified password.

### Environment
MODX 3.0.0-pl, PHP 8.0.17, Database 10.5.15-MariaDB, fresh installation

smg6511:
### What does it do?
Implemented change made in PR #16105 in a different way so it doesn't break the ability to specify a user's password.

### Why is it needed?
User passwords cannot currently be specified. Saving silently fails.

### How to test
Create and update at least one user to verify the record saves as expected when choosing to specify their password. Also verify that the password fields do not autocomplete.

### Related issue(s)/PR(s)
Resolves #16124.

### DIFF
--- a/manager/assets/modext/widgets/security/modx.panel.user.js
+++ b/manager/assets/modext/widgets/security/modx.panel.user.js
@@ -34,7 +34,11 @@
         }
     });
     MODx.panel.User.superclass.constructor.call(this,config);
-    Ext.getCmp('modx-user-panel-newpassword').getEl().dom.style.display = 'none';
+    const passwordPanel = Ext.getCmp('modx-user-panel-newpassword');
+    passwordPanel.items.each(item => {
+        item.getEl().set({autocomplete: 'new-password'});
+    });
+    passwordPanel.getEl().dom.style.display = 'none';
     Ext.getCmp('modx-user-password-genmethod-s').on('check',this.showNewPassword,this);
     Ext.getCmp('modx-extended-form').disable();
 };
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
I'm curious as to why my changes broke that functionality.. and how you figured out how to solve this.

smg6511:
The use of the `autoCreate` config caught my eye; it occurred to me that trying that on a class that already creates an input type (via declaring the xtype and inputType in the defaults) may have caused a conflict of sorts. I first looked closely at the password panel part of the config expecting a typo or something, but there was nothing of the sort and it's otherwise set up correctly.

Another small nuance to note: I called `doLayout()` when the panel is shown — something that always needs to be done for hidden components (or components where their configs are dynamically altered) upon showing them; otherwise, their sizes can not be recalculated properly (as seen before this fix).

Re figuring it out: I'd say the more I work in Ext JS, the more intuitive it becomes (never thought I'd say that!).

opengeek (approved):

PR #16127, opened 2022-04-06
Use random hash in tag if element name is blank

### LINKED ISSUE #16121: Chunk process issue
### Summary
Chunk processing returns wrong results, maybe from cache?

### Step to reproduce
*$nav= $modx->newObject('modChunk')->process($properties, $tpl);*

The above code line is from getPage snippet that got broken when I upgraded to Modx3. It seems that the above line returns the previous result even though the parametre $tpl has changed. Is the result coming from a cache?

### Expected behavior
The chunk should be processed, not returning a result from cache.

opengeek:
### What does it do?
Use a random hash instead of the name property to generate the tag string in modElement::getTag() when the name property is blank.

### Why is it needed?
This will prevent cache collisions if multiple elements are processed with the same property string value and no given name.

### How to test
Process two chunks with the same property value(s) and different content without giving them a name. They should return their own unique content. e.g.

```
$chunk = $modx->newObject(\MODX\Revolution\modChunk::class);
echo $chunk->process(['property' => '1'], "<p>[[+property]]</p>");

$chunk = $modx->newObject(\MODX\Revolution\modChunk::class);
echo $chunk->process(['property' => '1'], "<div>[[+property]]</div>");
```

Should return… 
```
<p>1</p><div>1</div>
```

### Related issue(s)/PR(s)
Resolves #16121

### DIFF
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -300,7 +300,7 @@ class modElement extends modAccessibleSimpleObject
             }
             $tag = '[[';
             $tag .= $this->getToken();
-            $tag .= $this->get('name');
+            $tag .= strlen($this->get('name')) > 0 ? $this->get('name') : md5(uniqid(rand()));
             if (!empty($this->_propertyString)) {
                 $tag .= $this->_propertyString;
             }


smg6511 (approved):
Tested before and after change and works as described.

PR #16128, opened 2022-04-07
[Port to 3.x] Fix Various Issues with Form Panel Validation and Field Display - Part 5, Chunk Panel

smg6511:
_NOTE: This is part 5 of 6 of a revised submission of #15887. This should be reviewed after part 2 is merged (there are two small changes needed from that PR to make the locked switch appear correctly)._

### What does it do?

Ports enhancements to the Chunk form panel from PR #15146 to 3.x (including subsequent fixes by others). 
Changes include:
1. Updated panel JS to make form layout consistent with the layout introduced in the current TV panel, and utilize new shared methods introduced in the main FormPanel via part 1 of this PR set (already merged).
2. A handful of minor lexicon updates.

### Why is it needed?

Makes needed UI and consistency improvements to the Chunk panel.

### How to test

2. Create and update a few Chunks to verify behavior and appearance is as expected.

### Related issue(s)/PR(s)
Partial port of #15146


theboxer:
@smg6511 If the chunk name is short enough, this happens :)

smg6511:
> @smg6511 If the chunk name is short enough, this happens :)
>
It's not the length of the element name, it's where it happened to end up in the description that's causing that to happen. What type of device are you viewing on in this example?

theboxer:
Chrome, macbook, external 24" screen. And it's the length of the element, if you hover the tagname to copy it, it appends the copy icon. If the element name is long enough, the icon will push the whole tag to the next line and as you stop hovering it, the icon disappears.

smg6511:
It's ultimately the description's line length though, it has to be just right for the example tag to fall right at the end. I debated making the fool-proof move of returning the tag to a new line (in which case what you pointed out would never happen), but it's such a waste of space to cover a scenario that would so rarely occur. The only other way would be to not offer the copy icon visual cue on hover. I thought it was important to give that hint though. 

I suppose another way around it would be to give the hint via a popover (QuickTip) instead of the icon.

theboxer:
@smg6511 tbh. I don't think this is a blocker for merging these PRs. Just wanted to point out that this is happening in some specific cases. I'd vote for merging and maybe finding a reasonable solution for next version?

PR #16130, opened 2022-04-08
Treat email contents as UTF-8, fixing mangled characters [#16116]

### LINKED ISSUE #16116: MODX 3.0.0 Wrong charset in body of HTML email
### Summary
When sending an email from MODX3, umlauts in the body do not display properly.

### Step to reproduce
Sending an html-email with umlauts.

```
$modx->getService('mail', 'mail.modPHPMailer');
$modx->mail->set(modMail::MAIL_BODY, 'Some text with umlauts ä, ü, ö, etc.');
$modx->mail->set(modMail::MAIL_FROM, $modx->getOption('emailsender'));
$modx->mail->set(modMail::MAIL_FROM_NAME, $modx->getOption('site_name'));
$modx->mail->set(modMail::MAIL_SENDER, $modx->getOption('emailsender'));
$modx->mail->set(modMail::MAIL_SUBJECT, 'Test');
$modx->mail->address('to', 'whatever@somedomain.com');
$modx->mail->setHTML(true);
if (!$modx->mail->send()) {
    $modx->log(modX::LOG_LEVEL_ERROR, 'Could not send mail.');
}
$modx->mail->reset();
```

### Observed behavior
The umlauts in the mail are converted to `Ã¤, Ã¼, Ã¶`.
The problem seems to happen in the class _InlineStyle_. I think DOMDocument::loadHTML assumes ISO-8859-1 if the charset can't be inferred from the mail body.

### Environment
MODX 3.0.0

### Related thread in the forum
https://community.modx.com/t/characters-in-email-body-messed-up/5145

Mark-H:
### What does it do?
Uses `mb_convert_encoding` to turn UTF-8 characters in email contents into entities before passing it to the InlineStyle class. 

### Why is it needed?
The style inlining added in 3.0 is mangling characters: #16116 

With this tweak, that should be resolved.

### How to test
Try sending an email that contains any sort of unicode characters through the mailer. #16116 has an example to quickly put in a snippet, but even better might be testing actual emails (login, contact forms) with this tweak.   

### Related issue(s)/PR(s)
Fixes #16116 

https://community.modx.com/t/characters-in-email-body-messed-up/5145/17?u=markh

### DIFF
--- a/core/src/Revolution/Mail/modPHPMailer.php
+++ b/core/src/Revolution/Mail/modPHPMailer.php
@@ -222,13 +222,14 @@ class modPHPMailer extends modMail
 
         $sent = false;
         try {
-            if (strpos($this->mailer->ContentType, 'html') !== false) {
-                if (!empty($this->mailer->Body)) {
-                    $html = new InlineStyle($this->mailer->Body);
-                    /** @noinspection PhpParamsInspection */
-                    $html->applyStylesheet($html->extractStylesheets());
-                    $this->mailer->Body = $html->getHTML();
-                }
+            if (!empty($this->mailer->Body) && (strpos($this->mailer->ContentType, 'html') !== false)) {
+                $body = $this->mailer->Body;
+                // Turn UTF-8 characters into entities
+                $body = mb_convert_encoding($body, 'HTML-ENTITIES', 'UTF-8');
+                $html = new InlineStyle($body);
+                /** @noinspection PhpParamsInspection */
+                $html->applyStylesheet($html->extractStylesheets());
+                $this->mailer->Body = $html->getHTML();
             }
             $sent = $this->mailer->send();
         } catch (Exception $e) {


rthrash:

https://community.modx.com/t/characters-in-email-body-messed-up/5145/16

core/src/Revolution/Mail/modPHPMailer.php
```diff
@@ -222,13 +222,14 @@ public function send(array $attributes = [])
...
-                if (!empty($this->mailer->Body)) {
-                    $html = new InlineStyle($this->mailer->Body);
-                    /** @noinspection PhpParamsInspection */
-                    $html->applyStylesheet($html->extractStylesheets());
-                    $this->mailer->Body = $html->getHTML();
-                }
+            if (!empty($this->mailer->Body) && (strpos($this->mailer->ContentType, 'html') !== false)) {
+                $body = $this->mailer->Body;
+                // Turn UTF-8 characters into entities
+                $body = mb_convert_encoding($body, 'HTML-ENTITIES', "UTF-8");
```
opengeek: Can you make this conditional based on the existence of the mbstring extension? Otherwise, this adds a non-default extension to the product requirements and would also need to be added to composer.json.

Mark-H: How do you feel about including the symfony mbstring polyfill instead? Pretty lightweight, would allow simplifying some code.

Jako:
You can call InlineStyle with the property containing a DomDocument object. https://github.com/christiaan/InlineStyle/blob/master/InlineStyle/InlineStyle.php#L49

Then the encoding can be set creating the DomDocument. Maybe that's a better option.

Mark-H:
Tried that first, didn't work.

core/src/Revolution/Mail/modPHPMailer.php
Mark-H: Oh hey, fun fact, we're already including `symfony/polyfill-mbstring`:

```
composer why symfony/polyfill-mbstring
mtdowling/jmespath.php  2.6.1   requires  symfony/polyfill-mbstring (^1.17)  
symfony/console         v5.4.7  requires  symfony/polyfill-mbstring (~1.0)   
symfony/string          v5.4.3  requires  symfony/polyfill-mbstring (~1.0)  
```

Would be nice to make it explicit in the composer.json, but looks like there's no need for function checks on mbstring functions anymore.

Mark-H:
The force push just made sure consistent quote styles were used, noticed the mb_convert_encoding had single and double quotes.

opengeek (approved):

Jako (approved):

opengeek (approved):

PR #16140, opened 2022-04-12
Fix inability to download files from the Media Browser

### LINKED ISSUE #16134: MODX 3.0.0 Can’t download files from Media Browser when basePath & baseURL are set
### Summary
In the 'Media Browser', right click -> "Download File" doesn't work, if the file is in a 'Media Source' with basePath/baseUrl set.

### Step to reproduce
Create a 'Media Source' with basePath/baseUrl set to a value other than empty (e.g. `assets/images/`).
In the 'Media Browser' try to download a file from this 'Media Source' with right click -> "Download File".

### Observed behavior
Blank page with error message "Cannot open file: /assets/images/myimage.jpg".

It seems that the basePath is added to the image path twice: `.../assets/images/assets/images/myimage.jpg`
When the processor “Browser/File/Download” is called, the basePath is already prepended to the image name in the “file” request-parameter and then gets added again in the processor (function `prefixPath()` in `core/vendor/league/flysystem/src/PathPrefixer.php`).

### Environment
MODX 3.0.0

smg6511:
### What does it do?
Removed unnecessary ajax call to the download class, making the Media Browser (MB) download method consistent with the currently-working directory tree one.

### Why is it needed?
To re-enable downloads from the MB.

### How to test
Navigate to the MB and verify that downloading files from various Media Sources (having both static and relative basePaths) operates as expected.

### Related issue(s)/PR(s)
Resolves #16134. Note that this PR picks up on what @halftrainedharry suggested as a solution in the referenced issue.

### DIFF
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -189,25 +189,11 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
         w.show(e.target);
     }
 
-    ,downloadFile: function(item,e) {
-        var node = this.cm.activeNode;
-        var data = this.lookup[node.id];
-        MODx.Ajax.request({
-            url: MODx.config.connector_url
-            ,params: {
-                action: 'Browser/File/Download'
-                ,file: data.pathRelative
-                ,wctx: MODx.ctx || ''
-                ,source: this.config.source
-            }
-            ,listeners: {
-                'success':{fn:function(r) {
-                    if (!Ext.isEmpty(r.object.url)) {
-                        location.href = MODx.config.connector_url+'?action=Browser/File/Download&download=1&file='+r.object.url+'&HTTP_MODAUTH='+MODx.siteId+'&source='+this.config.source+'&wctx='+MODx.ctx;
-                    }
-                },scope:this}
-            }
-        });
+    ,downloadFile: function(item, e) {
+        const node = this.cm.activeNode,
+              data = this.lookup[node.id]
+        ;
+        location.href = `${MODx.config.connector_url}?action=Browser/File/Download&download=1&file=${data.pathRelative}&HTTP_MODAUTH=${MODx.siteId}&source=${this.config.source}&wctx=${MODx.ctx}`;
     }
 
     ,copyRelativePath: function(item,e) {


Jako:
There was a reason, why the download has been split in two requests. I am quite sure that this has to work with two requests for not 'filesystem' media sources.

smg6511:
Are you talking as is using an S3 or FTP source? Regardless, why would the tree download work differently?

Jako:
First we have to dig in the code, why there is this two requests solution. When did it occur for the first time? Is it a leftover of MODX2? Maybe the one who wrote this, can tell more.

smg6511:
That's reasonable, but I'm not sure how to proceed on finding out when and why that was introduced. The ajax call was in the latest 2.x (but the processor is a bit different). What I can tell you is that so far _without_ the ajax call, downloading works for:
- relative and absolute path Media Sources (filesystem type)
- FTP Media Sources

An S3 test is pending — I'm setting up a test bucket today.

opengeek:
FWIW, this seems to be the same fix applied outside the media browser in #15385.

Jako:
It can be this is the reason: If the file is not available the manager can be redirected to a file not found page.

It is better to have a helper class that downloads the file in a generated iFrame.

```
MODx.util.HiddenForm(MODx.config.connector_url, {
    action: 'Browser/File/Download',
    download: 1,
    file: data.pathRelative,
    source: this.config.source,
    wctx: MODx.ctx,
    HTTP_MODAUTH: MODx.siteId
});

MODx.util.HiddenForm = function (url, fields) {
    if (!Ext.isObject(fields))
        return;
    var body = Ext.getBody(),
        frame = body.createChild({
            tag: 'iframe',
            cls: 'x-hidden',
            id: 'hiddenform-iframe',
            name: 'iframe'
        }),
        form = body.createChild({
            tag: 'form',
            cls: 'x-hidden',
            id: 'hiddenform-form',
            action: url,
            target: 'iframe',
            method: 'post'
        });

    Ext.iterate(fields, function (name, value) {
        form.createChild({
            tag: 'input',
            type: 'text',
            cls: 'x-hidden',
            id: 'hiddenform-' + name,
            name: name,
            value: value
        });
    });

    form.dom.submit();

    return frame;
};
```

The code can be improved a lot. The iframe must be destroyed after the download or it must get an own id with each call, a success/failure message can be added about the result of the submit.

smg6511:
Couple questions though:
- How is the file even going to show up in the browser listing (or the tree listing for that matter) if it does not exist?
- If you want to add a feature to the download functionality, shouldn't it be explored in a separate PR? Given that the download is just plain broken right now, I'd say make the simple change now.

I'm not implying it's a bad idea, I'm just asking if it's a necessary feature within the context of the manager.

Jako:
It can be added in a separate PR.

Jako:
> How is the file even going to show up in the browser listing (or the tree listing for that matter) if it does not exist?

The file can be renamed/moved/deleted after loading the tree in a different window, by a different user etc. That hiddenForm solution can be added to all downloads in the manager to display a message, after the file download is finished or the download failed from server side.

smg6511:
> It can be added in a separate PR.

Ok, sounds good. Do you want to work that up? Also, in the meantime, could you give this PR a quick test?

Jako (approved):

Jako:
The separate PR can be tested.

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -189,25 +189,11 @@ Ext.extend(MODx.browser.View,MODx.DataView,{
...
-            ,params: {
-                action: 'Browser/File/Download'
-                ,file: data.pathRelative
-                ,wctx: MODx.ctx || ''
-                ,source: this.config.source
-            }
-            ,listeners: {
-                'success':{fn:function(r) {
-                    if (!Ext.isEmpty(r.object.url)) {
-                        location.href = MODx.config.connector_url+'?action=Browser/File/Download&download=1&file='+r.object.url+'&HTTP_MODAUTH='+MODx.siteId+'&source='+this.config.source+'&wctx='+MODx.ctx;
```
Jako, line 206: A better fix is to get the right value back in r.object.url when the download property is sent to the processor. Then we don't loose the ability to download only existing files and don't redirect the manager to an error page for not existing or otherwise not downloadable files.

Jako (commented):
This fix is somehow a degression. Maybe it is better to implement #16150 instead before 3.1.0.

manager/assets/modext/widgets/media/modx.browser.js
smg6511: It's not a degression, it simply doesn't work the way it was originally written. You have to take into account what's happening in the processor as well to see why. And this same approach was already taken to fix the contextual download via the tree. Again, this PR's intent is to just get the basic expected functionality back _now_, not to provide an ideal solution to the download process. While I do see the advantages of your helper class, the scenario it covers is somewhat edge case and shouldn't hold up at least a patch application of this fix. If 3.0 hadn't gone live already, I'd be more apt to just take more time to explore your approach and forgo what I'd done. That and we seem to be having difficulty getting much review on anything lately.

Jako: The patch is a degression in terms of good error handling. The code previously checked for the possibility of a download and this is now dropped. The functionality is back at the expense of good error handling. This is the case here and avoiding this is the reason for the code review process.

A possible valid fix would be to send `data.pathRelative` in the `file` property in the second request instead of `r.object.url` in the original code. But I did not test that.

halftrainedharry: The return value from the first request `r.object.url` is **not** empty, even if the file doesn't exist anymore.

Jako: But then the success listener is not triggered.

halftrainedharry: When I open the Media Browser, then delete an image file on the file system and then execute right-click -> "Download File" on the deleted file, the first AJAX request is successful and `object.url` filled.
```
{"success":true,"message":"","total":0,"data":[],"object":{"url":"%2Fmodx300%2Fmyimage.jpg"}}
```
The success listener is triggered.

Jako: 😳

Jako: I have prepared a check, but my laptop currently has a composer issue. So I can't update the core from 3.0.0-beta to 3.0.0 to test this: https://github.com/Jako/revolution/tree/fix-download-exist

Jako:
I have added a new PR hopefully fixing the whole mess.

manager/assets/modext/widgets/media/modx.browser.js
Jako: @smg6511 Please do not think that your code change was not correct, but it somehow went in the wrong direction. I first had to understand why the previous code ran with two requests and what's really going wrong with that code.

From the current point of view, the additional FileDownload Helper method suggested at the beginning of this thread is no longer strictly necessary. But it provides the only way to tell the user that the download has been completed successfully. This could be displayed with an info bubble on the right side of the browser viewport.

smg6511: @Jako No worries. I come at this knowing from previous experience with trying to fix the static elements issue, that delving into a file handling/media source-centered issue _could_ become a longer than expected headache. So I was going with more immediacy toward just getting the download working when it _should_, leaving the intended error handling (which was not clearly self-evident from the original code) to a date soon after an initial fix.

At any rate, thankfully you identified a simple solution in #16152 (which I just reviewed). I had just a couple small suggestions there.

I think the push notification-like feedback indicating a successful download is interesting (icing on the cake). You could look to replicate the same feedback mechanism/style as is shown when saving Resources and Elements.

smg6511:
Closing in favor or #16152.

PR #16145, opened 2022-04-14
Fix Element/Template/GetList processor for >20 templates in manager

### LINKED ISSUE #16135: MODX 3.0.0 Template Menu in Quick and regular Edit mode
### Summary
The Template menu lists the first 20 templates in Quick Edit and Edit modes, and sporadically shows the Pagination navigation arrows.

### Step to reproduce
Create Document, Quick Create Document, Quick Edit, and Edit

### Observed behavior
Regardless of whether it displays the Pagination or not, it displays only 20 templates. I have 65. It's a mess because it's the original Cloud from 10 years ago.

QuickEdit1:
QuickEdit1

QuickEdit2:
QuickEdit2

RegEdit1: I don't know how this displayed the newer Templates and the Pagination nav.
RegEdit1

### Expected behavior
It should display all of the Templates.

### Environment
MODXcloud 3.0.0, upgraded from 2.8.3

halftrainedharry:
### What does it do?
Currently the Element/Template/GetList processor always returns a 'total' value of `0`. This causes the pagination of the template-dropdown to not work properly.

Also in the "Create Document" window, there is no pagination and the filtering happens without further AJAX-requests. Therefore the processor should return all templates if no limit is specified.
(Maybe a better fix would be to send a limit = `0` request parameter from the form though.)

### Why is it needed?
If you have more than 20 templates, not all of them can be selected when creating/updating a resource in the manager.

### How to test
Create more than 20 templates.
Create/update a resource and test if all templates can be selected.

### Related issue(s)/PR(s)
Resolves #16135

### DIFF
--- a/core/src/Revolution/Processors/Element/Template/GetList.php
+++ b/core/src/Revolution/Processors/Element/Template/GetList.php
@@ -33,7 +33,7 @@ class GetList extends \MODX\Revolution\Processors\Element\GetList
     public $languageTopics = ['template', 'category'];
     public $defaultSortField = 'templatename';
     public $permission = 'view_template';
-    
+
     public function initialize()
     {
         $this->setDefaultProperties([
--- a/core/src/Revolution/Processors/Element/Template/GetList.php
+++ b/core/src/Revolution/Processors/Element/Template/GetList.php
@@ -33,6 +33,15 @@ class GetList extends \MODX\Revolution\Processors\Element\GetList
     public $languageTopics = ['template', 'category'];
     public $defaultSortField = 'templatename';
     public $permission = 'view_template';
+    
+    public function initialize()
+    {
+        $this->setDefaultProperties([
+            'limit' => 0
+        ]);
+
+        return parent::initialize();
+    }
 
     public function prepareQueryBeforeCount(xPDOQuery $c)
     {
@@ -44,7 +53,7 @@ class GetList extends \MODX\Revolution\Processors\Element\GetList
             ]);
         }
 
-        $c->sortby('category_name');
+        $c->sortby('Category.category');
         $c->sortby('templatename');
 
         return $c;


smg6511 (approved):
Works as advertised!

Jako:
The limit line should not make any difference. limit = 0 is sent with the initial `Element/Template/GetList` request.

The sorting should be moved to an additional prepareQueryAfterCount method 

```
    public function prepareQueryAfterCount(xPDOQuery $c)
    {
        $c = parent::prepareQueryAfterCount($c);
        $c->sortby('category_name');
        $c->sortby('templatename');

        return $c;
    }
```

for two reasons:

- category_name is selected (created) in parent::prepareQueryAfterCount
- sorting (normally) does not change the count and without sorting the count is less resource consuming.

smg6511:
@Jako - Unless the sorting (in it's current placement) is affecting the number of returned results, altering the limit _might_ be necessary, as it's set to 20 in one of the ancestor classes (`core/src/Revolution/Processors/Model/GetListProcessor.php`). I see what you're saying about it being set to 0 in the `baseParams` in the `MODx.combo.Template` definition. Did you test what you posted above while removing the limit proposed in this PR?

halftrainedharry:
@Jako `Element/Template/GetList` is used in various places. At least in the "Create Document" window, there is no limit sent in the request. There may be other places.

Screenshot 2022-04-23 at 07-55-51 Dashboard MODX Revolution

Jako:
I have tested my comment and there was no issue with about 25 templates. It is somehow a BC if the default limit is changed. Setting the limit has to go in the JS code.

halftrainedharry:
Sure. Feel free to change this PR or create a new one.

halftrainedharry:
Closing in favor or #16160.

PR #16147, opened 2022-04-14
Ignore E_DEPRECATED and E_STRICT errors in setup

opengeek:
### What does it do?
Ignore E_DEPRECATED and E_STRICT errors in setup

### Why is it needed?
In PHP 8.1, functions are deprecated that will not be resolved in essential dependencies such as Smarty. If we do not ignore these errors during setup, whether a production server is properly set up to ignore these or not, MODX setup will be problematic, displaying concerning error messages on screen. This should have never been explicitly set, as these are meant to be informational errors.

### How to test
Test setup in PHP 8.1 and make sure deprecated errors are not displayed.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/setup/includes/request/modinstallrequest.class.php
+++ b/setup/includes/request/modinstallrequest.class.php
@@ -20,7 +20,7 @@
  * @package setup
  */
 
-error_reporting(E_ALL & ~E_NOTICE);
+error_reporting(E_ALL & ~E_NOTICE & ~E_DEPRECATED & ~E_STRICT);
 @ ini_set('display_errors', 1);
 
 class modInstallRequest {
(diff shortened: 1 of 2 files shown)


opengeek:
I'm torn between solving it this way or by removing the `error_reporting` and `display_errors` code completely.

JoshuaLuckers:
> I'm torn between solving it this way or by removing the `error_reporting` and `display_errors` code completely.

It can be useful to have errors shown during setup, especially when it's turned off on server level. It's already a lot of guessing when people report something not working during setup.

opengeek:
> > I'm torn between solving it this way or by removing the `error_reporting` and `display_errors` code completely.
> 
> It can be useful to have errors shown during setup, especially when it's turned off on server level. It's already a lot of guessing when people report something not working during setup.

I get that, but there are logs that are supposed to capture those during setup specifically.

JoshuaLuckers:
I assume you just gave a good reason to delete it haha.

PR #16149, opened 2022-04-15
Adjust placeholder filtering and accessible options for @SELECT

### LINKED ISSUE #16148: MODX 3.0.0 – getResources & getPage fails to load TPLs using @FILE Binding
### Summary
```@FILE``` Binding in  "getResources" & "getPage" no longer works in MODX 3.0.0.

### Step to reproduce
In one of my longstanding projects I have always used the following way to call child resources:

```
[[!getResources?
	&parents=`2`
	&where=`{"template:=":8}`
	&tplPath=`[[++assets_path]]templates/page/chunks/getResources/`
	&tplWrapper=`@FILE product_outer.tpl`
	&wrapIfEmpty=`0`
	&tpl=`@FILE product_item.tpl`
	&hideContainer=`1`
	&includeTVs=`1`
	&processTVs=`1`
	&.....
]]
```

### Observed behavior
and now the same call doesn't work anymore in Modx 3.0.0. It is the same with "getPage".
The call only works if I load the chunks directly from the element tree in the manager system, but not from the filesysrem.

This Example does still works:
```
[[!getResources?
	&parents=`2`
	&where=`{"template:=":8}`
	&tplWrapper=`product_outer`
	&wrapIfEmpty=`0`
	&tpl=`product_item`
	&hideContainer=`1`
	&includeTVs=`1`
	&processTVs=`1`
	&.....
]]
```
Does this issue have anything to do with the latest changes with ```@EVAL```?
I know that ```@EVAL``` is no longer supported, but the ```@FILE``` Binding should still work, right?

### Expected behavior
It should be possible to output external chunks with the use of ```@FILE``` binding in "getResources" & "getPage".

### Environment
MODX Revolution: 3.0.0-pl (traditional)
PHP Version: 7.4.28
Linux: 4.14.132
Apache: 2.0 Handler (20120211)
mysql: 5.7.37-40-log
mysqlnd: 7.4.28...
Firefox, Firefox Developer Edition, Safari, Chrome

opengeek:
### What does it do?
Port adjustments to placeholder filtering and @SELECT options from 2.x

### Why is it needed?
This restores functionality that was originally lost and then restored in 2.x

### How to test
Test the usage of placeholders for assets_path, base_path, and core_path in parsed content, as well as legacy placeholders for @SELECT (table_prefix and dbname).

### Related issue(s)/PR(s)
Port of #15714 
Resolves #16148

### DIFF
--- a/core/src/Revolution/modTemplateVar.php
+++ b/core/src/Revolution/modTemplateVar.php
@@ -916,8 +916,8 @@ class modTemplateVar extends modElement
                     if ($modx->resource && $modx->resource instanceof modResource) {
                         $dbtags = $modx->resource->toArray();
                     }
-                    $dbtags['DBASE'] = $this->xpdo->getOption('dbname');
-                    $dbtags['PREFIX'] = $this->xpdo->getOption('table_prefix');
+                    $dbtags['DBASE'] = $dbtags['+dbname'] = $this->xpdo->getOption('dbname');
+                    $dbtags['PREFIX'] = $dbtags['+table_prefix'] = $this->xpdo->getOption('table_prefix');
                     foreach ($dbtags as $key => $pValue) {
                         if (!is_scalar($pValue)) continue;
                         $param = str_replace('[[+'.$key.']]', (string)$pValue, $param);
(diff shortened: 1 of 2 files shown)


opengeek:
FWIW, I am NOT fixing the code quality errors for this patch. This can be addressed at another time.

JoshuaLuckers (approved):

Jako (approved):

PR #16150, opened 2022-04-15
New MODx.util.FileDownload helper method

Jako:
### What does it do?
Add a MODx.util.FileDownload helper method.

### Why is it needed?
An improved fix for #16134 than #16140 

### How to test
The new method can be called the following :
```
MODx.util.FileDownload({
    url: MODx.config.connector_url,
    params: {
        action: 'Browser/File/Download',
        file: data.pathRelative,
        wctx: MODx.ctx || '',
        source: this.config.source,
        download: true
    },
    success: function (response) {
        MODx.msg.status({
            title: _('success'),
            message: response.message,
            delay: 2
        });
    },
    failure: function (response) {
        MODx.msg.alert(_('error'), response.message);
    }
});
```

The allowed settings for the method are documented in the code.

To use the success callback, the download processor has to set a cookie i.e. by:

```
if ($this->getProperty('cookieName')) {
     setcookie($this->getProperty('cookieName'), 'true', time() + 10, '/');
}
```

### Related issue(s)/PR(s)
#16140, #16134


Jako:
The cookie key has to be changed to something unique.

core/lexicon/en/file.inc.php
```diff
@@ -75,6 +75,9 @@
 $_lang['file_folder_rename'] = 'Rename Directory';
 $_lang['file_last_accessed'] = 'Last Accessed';
 $_lang['file_last_modified'] = 'Last Modified';
+$_lang['file_msg_download_error'] = 'Download error!';
+$_lang['file_msg_download_succees'] = 'Download success!';
```
JoshuaLuckers:
```suggestion
$_lang['file_msg_download_success'] = 'Download success!';
```

PR #16151, opened 2022-04-16
Don't show the Configuration Check dashboard widget if all checks pass

### LINKED ISSUE #16109: MODX 3.0 — Configuration Check widget no longer goes away
### Summary
In 2.8 and before, the Configuration Check widget was automatically hidden if there were no configuration errors.

### Observed behavior
In the 3.0 dashboard, Configuration Check appears even if there are no errors.

### Expected behavior
Why did this change? It's much better for this widget to be hidden if there are no problems so that it doesn't clutter up the dashboard.

### Environment
MODX 3.0.0.

JoshuaLuckers:
### What does it do?
Don't show the success message for the configuration check widget if all checks pass.

### Why is it needed?
This will hide the widget, otherwise it would stay visible indefinitely unless you manually remove the widget.

### How to test
Resolve all warnings shown in the configuration check widget.

### Related issue(s)/PR(s)
Resolves #16109

### DIFF
--- a/manager/templates/default/dashboard/configcheck.tpl
+++ b/manager/templates/default/dashboard/configcheck.tpl
@@ -8,6 +8,4 @@
             </li>
         {/foreach}
     </ul>
-{else}
-    <h4 class="green">{$_lang.configcheck_ok}</h4>
 {/if}


Ibochkarev (approved):

rthrash:
This makes total sense to me … and in theory could lose another line in the Lexicon.

opengeek (approved):

PR #16152, opened 2022-04-17
Fix download files from the Media Browser

Jako:
### What does it do?
- Fix download files from Media Browser when the media source base_path/base_url are set
- Fix download url is returned even if the file does not exist

### Why is it needed?
See #16134. During discussing the fix in #16140, the second issue was detected.

The fix in #15385 has to be reverted and the failure action of this patch can be added. 

This PR replaces #16140.

### How to test
Create a file in a media source that has base_path/base_url set, and download that file. You get a blank page with the error message "Cannot open file: /path/to/the/file". The same thing happens when a file is deleted elsewhere in the filesystem without reloading the Media Browser.

### Related issue(s)/PR(s)
#16134, #16140


core/src/Revolution/Processors/Browser/File/Download.php
```diff
@@ -40,7 +40,11 @@ public function process()
 
         // Manager asks for file url
         if (!$this->getProperty('download')) {
-            return $this->success('', ['url' => rawurlencode($this->source->getObjectUrl($file))]);
+            if ($this->source->isFile($file)) {
```
smg6511: You don't need to create a new method in modMediaSource. I'd change line 43 to:
`if ($this->source->getFilesystem()->fileExists($file)) {`

core/src/Revolution/Processors/Browser/File/Download.php
```diff
@@ -40,7 +40,11 @@ public function process()
...
+                return $this->success('', ['url' => rawurlencode($file)]);
+            } else {
+                return $this->failure($this->modx->lexicon('file_err_nf') . ': ' . $file);
```
smg6511, line 46: The lex entry `file_err_nf` has punctuation at the end of if, so it looks strange adding the colon. A couple of suggested options:
1. Change line 46 to `return $this->failure($this->modx->lexicon('file_err_nf') . ' (' . $file . ')');` OR
2. Consider adding a new file lex entry to provide a smoother, more descriptive message:
`$_lang['file_err_notfound'] = 'The requested file “[[+file]]” could not be found. Please refresh this page to ensure its file listings are up to date.';`
and adjust line 46 to:
`return $this->failure($this->modx->lexicon('file_err_notfound', ['file' => $file]));`

core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2332,6 +2332,24 @@ protected function getImageDimensions($path, $ext)
...
+    {
+        try {
+            return $this->filesystem->fileExists($file);
+        } catch (Exception $e) {
+            // pass
+        }
+
+        return false;
+    }
+
```
smg6511: You can remove this method, if you take my first suggestion above.

smg6511 (changes requested):
Ok, this does the job. Whatever the final submission is, the file tree js should be updated as well to carry through this error handling behavior.

core/src/Revolution/Processors/Browser/File/Download.php
Jako: The other messages (i.e. deleting a not existing file) are created the same. That can be changed later on.

Jako:
I pushed the requested changes (not the lexicon ones) and I am away for a week. If there is something else to do fast, this has to be created in a fork or by editing this PR (Jason, Mark and Joshua should be able to this).

smg6511 (approved):
Ok, this will work for now. I'll propose the lexicon change in a different PR.

PR #16160, opened 2022-04-23
Fix the count in Element/TemplateGetList processor

Jako:
### What does it do?
- Move the sortby queries into the prepareQueryAfterCount.
- Add a limit: 0 to the baseParams of the TemplatePicker combo in the resource creating window.
- Some normalizing of the boolean value of the `combo` property

### Why is it needed?
- The category_name column is not available in prepareQueryBeforeCount. This causes issues during retrieving the count  in the Template combo. Because of this, the pagination was not shown with more than 20 templates available.
- The default limit is 20 in the Model/GetListProcessor and only the first 20 templates are selectable without a search
- The `combo` property contains a boolean value and not a string or an integer.

### How to test
- Create more than 20 templates and edit an existing resource. Only the first 20 templates are selectable without a search in the combo.
- Create more than 20 templates and create a resource. Only the first 20 templates are selectable without a search in the template list.

### Related issue(s)/PR(s)
#16135, this PR improves and replaces #16145


JoshuaLuckers (approved):
It works fine, however when you select a template from the second "page" it doesn't select this template after you refresh and open the template picker again. This works fine for templates from the first page.

Not a showstopper, just something I've noticed.

Jako:
This must be an Ext JS issue and has to be investigated later.

Ibochkarev (approved):

Jako:
Removing `limit: 0` caused this, since then the loaded store does not contain all records. However, it is better not to load all records in every combo by default for efficiency. Maybe there is a way to add the x-combo-selected class later on.

Jako:
During my tests it worked: If the selected Template is on page 2, the right template will be selected, when you go to page 2 in the combo.

It is not possible to have page 2 automatically opened, since the current value is not sent in the request that retrieves the paginated list. This can be somehow done in the combo code, but then the page has to calculated with a minimum of database queries. Is this possible?

JoshuaLuckers:
> During my tests it worked: If the selected Template is on page 2, the right template will be selected, when you go to page 2 in the combo.

Indeed it does! Thanks.


> It is not possible to have page 2 automatically opened, since the current value is not sent in the request that retrieves the paginated list. This can be somehow done in the combo code, but then the page has to calculated with a minimum of database queries. Is this possible?

I'm fine if we leave it like it is for now. For me it's not a real problem that needs to be solved.

smg6511 (approved):
This does the job. I am curious though why it is not necessary to set the limit to 0 on the template combo config (not the picker variation) — somehow the processor's default of 20 is being overridden, right?

Jako:
The default pagination of a combo is set with the pageSize property of 20 in Ext JS.

smg6511:
I understand the pagination config. It's the limit of _total_ results (which wasn't working before, but now is). Why would the query only fetch a total of 20 before this fix?

Jako:
Before, the count query was sorted by a missing column, which returned a total of 0 then. This disabled the pagination. The final query for the results contains the missing column (it is added in prepareQueryAfterCount). The processor uses also an internal default limit of 20.

PR #16163, opened 2022-04-24
[3.x] Make hidden TV's work for weblinks, symlinks and static resources

JoshuaLuckers:
### What does it do?
Add the hidden TV inputs to the create/update manager templates for weblinks, symlinks and static resources

### Why is it needed?
To make hidden TV's work for weblinks, symlinks and static resources. Otherwise it would only work for resources with Resource Type "Document" .

### How to test
- Create a hidden TV
- Create/update a weblink, symlink and static resource.
- Set the value of the hidden TV (e.g install the DaterangeTV and set the `End Value Template Variable`)
- Save the resource.

### Related issue(s)/PR(s)
PR #16162

### DIFF
--- a/manager/templates/default/resource/staticresource/create.tpl
+++ b/manager/templates/default/resource/staticresource/create.tpl
@@ -1,5 +1,9 @@
 <div id="modx-panel-static-div"></div>
 <div id="modx-resource-tvs-div" class="modx-resource-tab x-form-label-left x-panel">{$tvOutput|default}</div>
+{foreach from=$hidden item=tv name='tv'}
+    <input type="hidden" id="tvdef{$tv->id}" value="{$tv->default_text|escape}" />
+    {$tv->get('formElement')}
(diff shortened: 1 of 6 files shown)


Jako (approved):

Jako:
Does anyone know why the resource/update.tpl does not contain the code `class="modx-resource-tab x-form-label-left x-panel"`, but the weblink/symlink/staticresource one does? It makes no difference during rendering the page, but some scripts can use these classes.

JoshuaLuckers:
> Does anyone know why the resource/update.tpl does not contain the code `class="modx-resource-tab x-form-label-left x-panel"`, but the weblink/symlink/staticresource one does? It makes no difference during rendering the page, but some scripts can use these classes.

It's been this way since the commit that introduced it: https://github.com/modxcms/revolution/commit/8ffb9b949a8a96c77526b931a97130be9aa47ac4

Issue #16044, opened 2022-02-07
wrong parse when having a semi-column with a setting variables in output filter 'then' or 'else'


intersel:
### Summary
When we add a semi-column (';') in the output filter "then", modx can't parse properly and the output is strange....

This behaviour needs to use twice a same setting value... see example.

### Step to reproduce
create snippet "echoinput"
```
<?php
return $input;
```
create a resource "test bug" with a void template and the  following content:
```
[[echoinput?&input=`[[++mail_smtp_hosts]] [[echoinput:eq=`0`:then=`test1:[[++mail_smtp_hosts]];`:else=`test2:[[++mail_smtp_hosts]]`?input=`0`]]`]]
```

### Observed behavior
result:
```
`localhost [[echoinput:eq=`0`:then=`test1:localhost
```

### Expected behavior

```
localhost test1:localhost;
```
If you remove the ';' in the then filter like `test1:[[++mail_smtp_hosts]]` instead of `test1:[[++mail_smtp_hosts]];`

then everything is parsing ok....

If you add a blank after the name of the setting variable "[[++mail_smtp_hosts ]]" , like 
```
[[echoinput?&input=`[[++mail_smtp_hosts ]] [[echoinput:eq=`0`:then=`test1:[[++mail_smtp_hosts]];`:else=`test2:[[++mail_smtp_hosts]]`?input=`0`]]`]]
```

The parsing is ok too !

### Related issue(s)/PR(s)
can not say...

### Environment
MODX Revolution 2.8.3-pl, Apache/2.4.38 ,  10.3.31-MariaDB, Firefox/chrome/...

JoshuaLuckers:
@intersel are you sure the syntax is right? Does it also occur if you only use ```[[echoinput:eq=`0`:then=`test1:[[++mail_smtp_hosts]];`:else=`test2:[[++mail_smtp_hosts]]`]] ``` ?

intersel:
@JoshuaLuckers 
Syntax seems right... can't see what's wrong actually :s

if I only use
``
[[echoinput:eq=`0`:then=`test1:[[++mail_smtp_hosts]];`:else=`test2:[[++mail_smtp_hosts]]`]]
``
it gives me 
``
test2:localhost
``
That's seem correct to me...

intersel:
It is really a weird bug! as it is really in this kind of steps to reproduce that you have it... as soon as I remove or change something... it's working as expected!

JoshuaLuckers:
[...]

PR #16165, opened 2022-04-26
[3.x] Unit Test for modInputFilter

JoshuaLuckers:
### What does it do?
Adds a Unit Test for modInputFilter.

### Why is it needed?
To make it easier to test if modInputFilter works as it should when we adjust things.

In another PR I will add data to the provider that will make the test fail, this is an issue we should solve. 
This will help with fixing issue #16044

### How to test
Run the Unit Tests suite (or the single test) and see it pass.

### Related issue(s)/PR(s)
Issue #16044

### DIFF
--- a/_build/test/Tests/Model/Filters/modInputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modInputFilterTest.php
@@ -12,6 +12,8 @@
 namespace MODX\Revolution\Tests\Model\Filters;
 
 
+use MODX\Revolution\Filters\modInputFilter;
+use MODX\Revolution\modFieldTag;
 use MODX\Revolution\MODxTestCase;
 
 /**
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Adding this to the data provider will make the Unit Test fail:

```
 [
    'pagetitle:eq=`0`:then=`foo:bar`:else=`bar:foo`',
    'pagetitle',
    [
        'eq',
        'then',
        'else',
    ],
    [
        '0',
        'foo:bar',
        'bar:foo',
    ]
]
```

Failed asserting that two arrays are equal.

Expected:
```
Array (
    0 => 'eq'
    1 => 'then'
    2 => 'else'
)
```

Actual:
```
Array (
    0 => 'eq'
    1 => 'then'
    2 => 'foo`'
)
```

opengeek (approved):

PR #16168, opened 2022-04-28
Use MODx.util.FileDownload for downloads

Jako:
### What does it do?
Use MODx.util.FileDownload for downloads

### Why is it needed?
Allow success and failure messages during downloading files

### How to test
The new download method is used in the MODX file browser download, in the file tree download, in the Form Customisation export and in the Properties Set export.

### Related issue(s)/PR(s)
#16150, Discussion in #16140

### DIFF
--- a/core/src/Revolution/Processors/System/DownloadOutput.php
+++ b/core/src/Revolution/Processors/System/DownloadOutput.php
@@ -10,6 +10,7 @@
 
 namespace MODX\Revolution\Processors\System;
 
+use MODX\Revolution\File\modFileHandler;
 use MODX\Revolution\Processors\Processor;
 
 /**
@@ -23,46 +24,23 @@ class DownloadOutput extends Processor
      */
     public function process()
     {
-        if ($this->getProperty('download')) {
-            $output = $this->download();
-        } else {
-            $output = $this->cache();
+        $cookieName = $this->getProperty('cookieName');
+        if ($cookieName) {
+            setcookie($cookieName, 'true', time() + 10, '/');
         }
-        return $output;
+
+        return $this->download();
     }
 
     /**
-     * Download the output to the browser
-     * @return string
+     * Create a temporary file object and serve as a download to the browser
+     *
+     * @return bool|string
      */
     public function download()
     {
-        $dl = $this->getProperty('download');
-        $dl = str_replace(['../', '..', 'config'], '', $dl);
-        $dl = ltrim($dl, '/');
-
-        $f = $this->modx->getOption('core_path') . $dl;
-        $o = $this->modx->cacheManager->get($dl);
-        if (!$o) {
-            return '';
-        }
-
-        $this->modx->cacheManager->delete($dl);
-
-        $bn = basename($f);
-        @session_write_close();
-        header('Content-Type: application/force-download');
-        header('Content-Disposition: attachment; filename="' . $bn . '-' . date('Y-m-d Hi') . '.txt"');
-        return $o;
-    }
-
-    /**
-     * Cache the data stored
-     * @return array|string
-     */
-    public function cache()
-    {
         $data = $this->getProperty('data');
+
         if (empty($data)) {
             return $this->failure($this->modx->lexicon('invalid_data'));
         }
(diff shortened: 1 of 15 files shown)


core/src/Revolution/Processors/System/DownloadOutput.php
```diff
@@ -23,69 +24,44 @@ class DownloadOutput extends Processor
...
         $data = strip_tags($data);
-        $o = '/*
-* MODX Console Output
-*
-* @date ' . date('Y-m-d H:i:s') . '
-*/
-' . $data . '
-/* EOF */
-';
+        $output = "/*\r\n* MODX Console Output\r\n*\r\n* @date " . date('Y-m-d H:i:s') . "\r\n*/\r\n" . $data . "\r\n/* EOF */\r\n";
```
Ruslan-Aleev, line 51: Maybe we should use the system setting to display the time format? Or is it not needed here?

core/src/Revolution/Processors/System/DownloadOutput.php
```diff
@@ -23,69 +24,44 @@ class DownloadOutput extends Processor
...
-        $fileName = $this->modx->getOption('core_path') . $file;
-        if (file_exists($fileName)) {
-            $this->modx->cacheManager->delete($fileName);
-        }
-        $success = $this->modx->cacheManager->set($file, $o);
-        return $success ? $this->success($file) : $this->failure($this->modx->lexicon('cache_err_write'));
+        /** @var modFileHandler $fileHandler */
+        $fileHandler = $this->modx->getService('fileHandler', modFileHandler::class);
+
+        $fileName = 'output-' . date('Y-m-d Hi') . '.txt';
```
Ruslan-Aleev: There is also a question about the output of the time format, and it differs from the format above. Or is it not important?

opengeek:
Why was this not part of the PR that introduced the new method? Now we have included an unused method in a release.

Jako:
Because #16150 was targeted to 3.1 🤷

opengeek:
> Because #16150 was targeted to 3.1 🤷

I must have missed that, but in the future, I would like to avoid this kind of separation. There is no point in having a PR introduce code that is not used in that PR.

core/src/Revolution/Processors/System/DownloadOutput.php
Jako: The system settings are based on javascript date format. This is code is PHP.

Jako:
It was posted as a possible solution for #16134 but it has been solved different. Next time I will add those PRs combined.

core/src/Revolution/Processors/System/DownloadOutput.php
Jako: Sorry, this was already wrong in the original code and I did not noticed that. Fixed now.

smg6511 (changes requested):
There is one major issue I came across while testing: If you attempt to download a file that no longer exists, the process ends up in an infinite loop (which eventually crashes and pops up an error modal with no text). This happens in the try/catch block.

Otherwise, I'd like to see everyone get out of their old ES5 ways — use ES6+ syntaxes and methods and more contemporary best practices. We were limited in our ability to do this until Sept 2021, when updates I made to the build process were merged — allowing the use of ES6+ syntax. 

A couple suggestions in that regard could be applied in the `MODx.util.FileDownload` utility:
- Use `const` and `let` to declare your variables. The only reason to use `var` these days is when you need the variable in the global scope.
- Comma separate your variable declarations (that use a common keyword) instead of using the keyword for every variable, maintaining separation by having only one variable per line.

smg6511:
The only other thing I'm wondering is what true advantage of this utility is. I'm not saying I know there are _not_ advantages, I'm just not readily seeing them and could use some enlightening in that regard. One of the things I thought would be nice (which this _almost_ does) is provide the same post-success notification that you get in other areas of the UI. 

The problem here is the timing and what that success is actually meaning: The success shown occurs when the background process of the download occurs (often at the same time the system's file dialog window appears and before the user clicks save or cancel), not when the user additionally _accepts_ the download and it has been saved to its target location. This is because, at least from the initial research I've done, it's impossible to hook into the browser/system in that way; there's no event that's exposed that you can take advantage of to know what ultimately happens at the end of the download process.

Anyway, this seems to point to it being really difficult to reliably provide success feedback with the right timing (and meaning for that matter—I think the modern browsers make it pretty self evident when a download is occurring and whether the file completely downloaded or not).

JoshuaLuckers:
> - Comma separate your variable declarations (that use a common keyword) instead of using the keyword for every variable.

Personally I find this pretty annoying and makes the code less readable and requires it more cognitive function

smg6511:
> Personally I find this pretty annoying and makes the code less readable

I agree with you on that when each variable is _not_ on its own line. (I wasn't clear about that above; I _do_ prefer to have each variable on its own line.) The way I format these is easily readable, it just doesn't have all those declaration keywords.

Jako:
> The only other thing I'm wondering is what true advantage of this utility is.

There was no download utility available. Some download code was wrapped with MODx.Ajax.request to have some error handling available. Instead of redirecting the browser to an error page. The reason for the AJAX requests was not documented in the code, so it was removed in #16140 and #15385. Some download code did not have this check.

The first idea was just to have a download error handler available in https://github.com/modxcms/revolution/pull/16140#issuecomment-1098814801

But then I found some code in an [stackoverflow answer](https://stackoverflow.com/questions/20499959/extjs-4-downloading-a-file-through-ajax-call/40688974#40688974) that allowed a success handler.

The success handler is triggered at the moment, when the file is fully sent to the browser (then the polling cookie is set). When generating a download needs some time, this can be helpful.

> use ES6+ syntaxes and methods and more contemporary best practices.

@smg6511 I can rewite the code do use const/let and maybe it is possible to use `promise` instead of polling.

Jako:
The infinite loop is ended with a timeout message in MODx.util.FileDownload. The issue with the download of a not existing file is fixed now.

Mark-H (changes requested):
I'm so sorry for being late to the party (been a bit preoccupied with the MODXpo prep these last two weeks and all of this went straight past me) but wow this whole utility seems waaaaaaaaayyy more complex than needed for the simple task of downloading a file. 

All we need to do is stream the file through to the media source API, which #16134 aimed to fix.  

If there's insufficient user-friendly error handling in a processor, a much simpler solution would be to introduce a download _controller_ instead that will either stream the file or return the nice error page we added in 3.0. 
 
I would also propose we revert #16150.

Jako:
Do you really want the user to be redirected to an error page, when a download can't be executed? Or is it better to show an error popup on the same page in that case?

Mark-H (changes requested):
After a chat with Thomas in Slack we can't seem to find agreement. I remain with my position this is way too overengineered for a very narrow use case. Showing an error in the edge case of a file being removed in between listing files and hitting download does not IMO constitute any of this complexity. I still believe it is best off reverted and removed before third parties start relying on it. 

If a majority of integrators feel otherwise I will not block it and will remove my review. Just don't expect me to get involved when it breaks or is found to contain security issues.

Jako:
It is useful to have a download component in the core javascripts, that allows an error and a success handler. If the success handler is not needed, the polling + cookie part can be removed (maybe the polling part can be rewritten with some 'promise' code). But a good error handler is essential.

The code is not very complex and hopefully easy to follow. There should be no security issues in the code, since it just forwards the download processor properties to an iframe calling the the download processor. The displayed message content is handled by MODx.util.safeHtml.

The component can be used not only in the core, but also in any extra that offers a download.

Issue #16172, opened 2022-05-01
MODX 3.0.1 User Settings - Typeahead doesn't work


pepebe:
### Summary

Typeahead inside user settings doesn't work.

### Step to reproduce
Add a new system setting to a user:

manage->users->USER1->Settings->Add

Click into the key input field and start typing.

### Observed behavior

If I add a new system setting to a user in MODX 3.x this happens if I start typing:
<img width="602" alt="image" src="https://user-images.githubusercontent.com/667315/166148941-bd82d3d3-286d-4087-a3e8-8eb011e3d299.png">

A list of system settings pops up, but not a single one fits,

### Expected behavior
If I add a new system setting to a user in MODX 2.8.4 this happens if I start typing:

<img width="610" alt="image" src="https://user-images.githubusercontent.com/667315/166163482-097dcbf6-6abf-4c35-97f6-030ed0d49632.png">

Typeahead searches for fitting settings and presents one or more results.

### Environment
Latest version of MODX 3.1

halftrainedharry:
It seems that in 2.x the parameter `key` was used to filter the settings

https://github.com/modxcms/revolution/blob/7d5e18753c0629aa4d60de134ac2945600e5cb0c/core/model/modx/processors/system/settings/getlist.class.php#L54

but in 3.x the parameter is called `query`.

https://github.com/modxcms/revolution/blob/b5d9b3459bc56590a4bacda89302100a35ca8212/core/src/Revolution/Processors/System/Settings/GetList.php#L63

---

Maybe the query parameter in the combobox can just be changed to the  value `'query'`.

https://github.com/modxcms/revolution/blob/b5d9b3459bc56590a4bacda89302100a35ca8212/manager/assets/modext/widgets/core/modx.combo.js#L1074

Ibochkarev:
Fixed https://github.com/modxcms/revolution/pull/16176

PR #16176, opened 2022-05-05
Change queryParam for typeahead in settings combo for Context/User section

Ruslan-Aleev:
### What does it do?
Change queryParam for typeahead in settings combo for Context/User section.

As suggested @halftrainedharry https://github.com/modxcms/revolution/issues/16172#issuecomment-1114640804

Now the typeahead is working.

query_combo

### Why is it needed?
Typeahead inside Context/User settings doesn't work.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/16172


Ibochkarev (approved):

smg6511 (changes requested):
This works, but there are a couple things you can do to fine-tune this config:

1. Remove `typeAhead: false`, as it's already set in the parent `MODx.combo.ComboBox` class
2. Remove the `queryParam` property; it's not needed here because the base `Ext.form.ComboBox` class sets it to 'query' by default

Ruslan-Aleev:
> This works, but there are a couple things you can do to fine-tune this config:
> 
> 1. Remove `typeAhead: false`, as it's already set in the parent `MODx.combo.ComboBox` class
> 2. Remove the `queryParam` property; it's not needed here because the base `Ext.form.ComboBox` class sets it to 'query' by default

@smg6511 You are right, corrected the code, check.

smg6511 (approved):

PR #16196, opened 2022-05-31
Update uploadObjectsToContainer method

sdrenth:
Update uploadObjectsToContainer so the files array passed to the OnFileManagerUpload event includes an up-to-date version of $objects. So if a file name has been renamed, $objects passed to the OnFileManagerUpload also reflects these changes.

### What does it do?
This ensures that $objects is kept up-to-date after looping through the array.

### Why is it needed?
Having an old file name triggered an unforced error in FileSluggy because the old file name was passed instead of the renamed filename.

### How to test
Install FileSluggy and try to upload a file with a name such as "Modx - Test.jpg". Now you'll see the unforced error:
```
 Error
- Modx - Test.jpg: The file is not a regular file and cannot be removed.
```
Even though the file has been uploaded successfully and can also be removed.

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -1108,7 +1108,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
         $this->xpdo->context->prepare();
         $maxFileSize = $this->xpdo->getOption('upload_maxsize', null, 1048576);
 
-        foreach ($objects as $file) {
+        foreach ($objects as $key => $file) {
             $this->xpdo->invokeEvent('OnFileManagerBeforeUpload', [
                 'files' => &$objects,
                 'file' => &$file,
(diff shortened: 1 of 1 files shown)


smg6511 (commented):
So, I verified that this change _does_ fix the error. However, that error is triggered due to how FileSluggy works, not a bug in MODX. I wonder if it's better to address this issue in the Extra's code rather than MODX's. To my eye, it's a harmless looking change, even if it is to accommodate an issue introduced outside of the core code. 

@Mark-H @opengeek - You guys will have to make the judgement call on this one...

JoshuaLuckers:
I would expect to receive up-to-date objects in `OnFileManagerUpload`.

@sdrenth  Can you think of any unwanted behaviors this change might introduce?

sdrenth:
@JoshuaLuckers cant think of any. But I also feel like this is a bug in the core that should be fixed, i would expect this behavior by default

Mark-H (approved):

JoshuaLuckers (approved):
I tested this with FileSluggy and the error message `The file is not a regular file and cannot be deleted.` no longer appears after applying the changes in this PR.

opengeek:
So is this a bug or a feature/enhancement? Should this be ported back for inclusion in a 3.0.x patch release?

sdrenth:
@opengeek it's a bug, method should return the new renamed filename

rthrash:

https://community.modx.com/t/resizeimageonupload-plugin-and-modx3/6703/2

rthrash:

https://community.modx.com/t/resizeimageonupload-plugin-and-modx3/6703/11

Issue #16015, opened 2022-01-30
There is no port in the URL during installation


jbl3010:
I use port HTTPS 88 as the port. After the installation process and installation, the URL does not carry the port number. As a result, I need to manually add':88'to the address bar to install and use it.

Resolvent

----

After installation, at core/config/config. Inc.php, change line 75 to:

$site_ Url=$url_ Scheme. $ Http_ Host.':'. $_ SERVER['SERVER_PORT']. MODX_ BASE_ URL;

Now I can use it normally.

----
From Machine Translation

Ruslan-Aleev:
Related with https://github.com/modxcms/revolution/pull/15622

opengeek:
> Related with #15622

This has nothing to do with the database connection string.

Ruslan-Aleev:
Yes, my bad, I confused the port for the database (https://github.com/modxcms/revolution/pull/15622#discussion_r733910682) with the port of the server.
Although we can fix it in one PR (if necessary), because refers to one file `core/config.inc.php`. Or delete the comments if I'm wrong.

JoshuaLuckers:
@jbl3010 if you run setup, does it keep the port or do you have to manually add it there as well?

jbl3010:
> @jbl3010 if you run setup, does it keep the port or do you have to manually add it there as well?

My address is https://xxxxxx:88

In the next step of setting, the ": 88" of the URL will disappear. Every time next, I have to add ": 88" manually

If you use non-standard ports (other than 80 and 443)

This phenomenon can be reproduced

smg6511:
Yes, this is very annoying behavior, especially for those operating dev environments on non-standard ports. The issue on the front end is how the config makes its assumptions, and then strips out the port in most cases on L66-L68:
```
if ($_SERVER['SERVER_PORT'] != 80) {
    $http_host= str_replace(':' . $_SERVER['SERVER_PORT'], '', $http_host); // remove port from HTTP_HOST
}
$http_host .= ($_SERVER['SERVER_PORT'] == 80 || $isSecureRequest) ? '' : ':' . $_SERVER['SERVER_PORT'];
```

A similar issue is present in the setup's bootstrap file (setup/provisioner/bootstrap.php), L22-L25:
```
if (isset($_SERVER['SERVER_PORT']) && (string)$_SERVER['SERVER_PORT'] !== '' && $_SERVER['SERVER_PORT'] !== 80) {
[...]

PR #16201, opened 2022-06-03
[2.x] Persist non-standard ports in site URL

smg6511:
### What does it do?
Ensures that non-standard ports persist in the site_url.

### Why is it needed?
Currently, when using a non-standard port for https (any one other than 443), the port gets stripped from the site's URL, making it impossible to properly navigate the site. Only the standard ports of 80 and 443 should be stripped.

### How to test
Configure the testing site on standard and non-standard ports, both with https on and off. Verify that setup, subsequent usage of the manager, and front end browsing all work as expected (with non-standard ports being maintained in the URL and 80/443 being stripped out).

### Related issue(s)/PR(s)
This is a backport of #16087 (which solved #16015 for 3.x).

### DIFF
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -57,14 +57,14 @@ if (!defined('MODX_URL_SCHEME')) {
 }
 if (!defined('MODX_HTTP_HOST')) {
     if(defined('PHP_SAPI') && (PHP_SAPI == "cli" || PHP_SAPI == "embed")) {
-        $http_host='{http_host}';
+        $http_host = '{http_host}';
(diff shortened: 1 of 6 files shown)


JoshuaLuckers (approved):

Ibochkarev (approved):

dimasites:
Hi! Thank for updates! But this still not fully correct working with non-standart ports. In sreenshot below, port :44333 MODX cut ":443" and "33" part breaks the links:

Still not correct working with non-standart ports

I made [new issue](https://github.com/modxcms/revolution/issues/16428) about this

PR #16202, opened 2022-06-03
[2.x] Fix inability to drag/drop files and folders in tree

### LINKED ISSUE #16194: Dragging of files and folders in Files tab in left bar does not work in MODX 2.8.4-pl
### Summary
When attempting to drag files or folders in the Files tab, the UI suggests that the move will be successful, but it fails.

### Step to reproduce
In MODX 2.8.4-pl, in the left column, use the Files tab to drag files or folders into new locations.

### Observed behavior
A green + will appear when you drag and hover over an acceptable folder. However, on release, the tree will refresh [collapsed] and the move has not taken place.

### Expected behavior
The file or folder should be moved to the new location.

### Environment
MODX 2.8.4-pl PHP 8.1.6

smg6511:
### What does it do?
Removes improper application of url and entity encoding on File tree config's id parameters.

### Why is it needed?
To re-enable the drag/drop of files and folders in the File tree, while also preserving the ability to use multibyte characters in folder and file names.

### How to test
Create a handful of folders and files in your tree, with some having mb characters such as ö, and verify that all can be interacted with as expected (moved, renamed, etc.).

### Related issue(s)/PR(s)
Resolves #16194, reverts PR #16071

### DIFF
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -57,14 +57,14 @@ if (!defined('MODX_URL_SCHEME')) {
 }
 if (!defined('MODX_HTTP_HOST')) {
     if(defined('PHP_SAPI') && (PHP_SAPI == "cli" || PHP_SAPI == "embed")) {
-        $http_host='{http_host}';
+        $http_host = '{http_host}';
         define('MODX_HTTP_HOST', $http_host);
     } else {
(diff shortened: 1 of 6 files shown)


Ruslan-Aleev (approved):
Thank you! Visually everything works as it should!

dd_tree

Ruslan-Aleev:
Also the file is now highlighted in the tree, although this behavior was gone with the latest PRs.

Ibochkarev (approved):

PR #16213, opened 2022-06-22
Bug 16212

### LINKED ISSUE #16212: findPolicy() in modMediaSource always looks at the mgr context
### Summary

### Step to reproduce

```
    {
        $policy = [];
        $enabled = true;
        $context = 'mgr';
```

### Observed behavior

### Expected behavior

### Environment

CarlBohman:
### What does it do?
Removes the forced mgr context for the findPolicy() function.

### Why is it needed?
See bug #16212

### How to test
1. Turn on info level logging.
2. Turn off the access_resource_group_enabled setting.
3. Access a page in the web context as an anonymous user.
4. No log messages about the mgr context should show up.

### Related issue(s)/PR(s)
Resolves #16212


CarlBohman:
This was against 3.x instead of 2.x.

PR #16219, opened 2022-07-10
Add filtering for ACL policy and policy template grids

### LINKED ISSUE #5822: Add Filter to Access Policy Editor View
smashingred created Redmine issue ID 5822

It would be nice if you could filter the items in the access policy editor view so you could filter on terms such as "resource" or "save" and manage the access for those matching.

This is the view being referenced:
!https://img.skitch.com/20110929-njwc9sikknx8qb9db755pjhd12.png!

smg6511:
### What does it do?
This update adds a filtering toolbar and associated methods for the referenced grids. It also provides some housekeeping and enhancements:
- Removes some unneeded code, such as the `url` and `baseParams` config entries where they are not applicable (since these are local not remote grids)
- For the Create form, changed the `name` field's xtype to textfield and removed code associated with the former combobox (which wasn't working and was irrelevant in the context of creating a new entry).
- Added dirty marker to new entries, as new entries are only saved when the user clicks save, and added new row highlight give additional cue to user that their new data was added but needs further action (save).
- New entries now appear at the top of the grid instead of the bottom
- Grids re-sort on save so new entries are put in order
- Added ability to search both the `name` _and_ `description` (or just name via checkbox control in toolbar)

### Why is it needed?
Addresses a very old request for this functionality, which makes the longer policy/policy template grids (the administrator ones in particular) much easier to navigate and work with.

### How to test
1. Run `grunt build` from `_build/templates/default` and ensure your caches are fully cleared.
2. Go to both the Policies and Policy Templates areas, select a specific Policy / Policy Template (especially the admin ones) and experiment with the new search filter.
3. Make changes/additions/updates to each to verify expected functionality.

### Related issue(s)/PR(s)
Resolves #5822.


smg6511:
Need a couple reviewers to check this PR out. That means you!!! ;-)

JoshuaLuckers (approved):

smg6511:
@Mark-H - While you're on your reviewing rampage this week, wanted to see if you could get a look at this and possibly include for 3.0.3 ;-)

opengeek:
> @Mark-H - While you're on your reviewing rampage this week, wanted to see if you could get a look at this and possibly include for 3.0.3 ;-)

I think this is a feature. It should be reviewed for inclusion in 3.1.

smg6511:
More accurately a _missing_ feature (for especially the admin template), as all other grids of this length have this same filtering. But I do get that some issues might be of greater importance (although this is a pretty easy one to review) ;-)

opengeek:
Fair. But features still go in minor versions, regardless of any qualifiers. 😉

Mark-H (approved):

PR #16220, opened 2022-07-11
Change behavior of ENTER key in settings grid cell editor

### LINKED ISSUE #5943: Don't Go Into Edit Mode on Next Setting On System Edit
johnywhy created Redmine issue ID 5943

When you edit a system setting, "Enter" takes you into edit mode on the next setting in the list.

This is very high-risk for me (i'm clumsy) and my non-expert users (they are non-expert). I have to gingerly escape out of edit mode on the next item, hoping i did not screw something up.

It's an invalid assumption to assume that the admin is going to edit each system setting one after the other, down the list like that. I have never done that. It's always a one-shot deal-- I have to edit just one setting. I edit it, and then i'm done. 

I think it's also invalid to assume users EXPECT the next item to go into edit mode, particularly when it takes two clicks to get into edit mode on the first item, and NO clicks to get into edit mode on the next item. That is not expected. This is not an Excel spreadsheet, and if it was, hitting Enter after editing a cell does NOT take you into edit mode on the next cell in the spreadsheet. 

thanks

smg6511:
### What does it do?
Overrides the grid selection model's `onEditorKey` method with an Interceptor so pressing the ENTER key (alone) does _not_ navigate to and open the next setting's value editor. Pressing the key combination of _control-ENTER_ will, however, behave as before (navigating down and opening that cell's editor).

### Why is it needed?
The current behavior is error-prone, and most users likely only edit one or a few setting(s) at a time (and not sequentially up and down the grid of settings).

### How to test
Simply change the value of a few settings using the grid editor using the ENTER key to commit your changes. Verify that changes are saved and that navigating the grid works as expected. Note that you can still TAB up and down once a cell editor is active.

### Related issue(s)/PR(s)
Resolves #5943.

### Special Note
There are a handful of grids where this solution could be applied as well (namespaces, content types, roles, etc.); or perhaps, if this is in fact the preferred behavior, we could consider applying it globally (via the base grid class).

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.settings.js
+++ b/manager/assets/modext/widgets/core/modx.grid.settings.js
@@ -237,7 +237,7 @@ MODx.grid.SettingsGrid = function(config) {
 
     // prevents navigation to next cell editor field when pressing the ENTER key
     this.selModel.onEditorKey = this.selModel.onEditorKey.createInterceptor(function(field, e) {
-        if (e.getKey() == Ext.EventObject.ENTER) {
+        if (e.getKey() == Ext.EventObject.ENTER && !e.ctrlKey) {
             e.stopEvent();
             return false;
         }
--- a/manager/assets/modext/widgets/core/modx.grid.settings.js
+++ b/manager/assets/modext/widgets/core/modx.grid.settings.js
@@ -234,6 +234,14 @@ MODx.grid.SettingsGrid = function(config) {
         ,scrollOffset: 0
     });
     MODx.grid.SettingsGrid.superclass.constructor.call(this,config);
+
+    // prevents navigation to next cell editor field when pressing the ENTER key
+    this.selModel.onEditorKey = this.selModel.onEditorKey.createInterceptor(function(field, e) {
+        if (e.getKey() == Ext.EventObject.ENTER) {
+            e.stopEvent();
+            return false;
+        }
+    });
 };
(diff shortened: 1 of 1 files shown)


Jako:
I am not sure, wether the cell field can be a textarea. Then the enter has to work inside.

smg6511:
I double-checked that — it works fine with the textarea. With TAs, you have to TAB or shift-ENTER to commit the change in the grid (but that's standard behavior).

JoshuaLuckers:
I'm sure someone will dislike this change. I find it useful from time to time, but most of the times it's annoying so I like this change.

smg6511:
How about this to retain the old behavior (albeit in a different way): Only navigate with ENTER when also pressing the control key?

smg6511:
Ok, ctrl-ENTER now navigates. I'm of the mind to make this a system-wide behavior, in which case I'd move this into the main grid class instead of the system settings one (where it currently resides). What do you all think?

JoshuaLuckers:
> I'm of the mind to make this a system-wide behavior, in which case I'd move this into the main grid class instead of the system settings one (where it currently resides). What do you all think?

I agree to have a consistent system-wide behavior for for now I suggest to test it with the settings grid first and see how users like it.

JoshuaLuckers (approved):

Ibochkarev (approved):

Jako (approved):

PR #16225, opened 2022-07-14
Provide global control to set autocomplete="off"

smg6511:
### What does it do?
Created a new constant in the main MODx class to enable system-wide setting of autocomplete attribute, and made changes and overrides in the appropriate places to apply this constant. 

Note that this will not work in Chrome, as Google has opted to force the issue and not comply with the setting of this attribute. The only real way to turn it off in that browser is to do it in the browser's settings.

### Why is it needed?
While autofill may be handy for user's typical front end browsing, it's more a burden working in the back end. Therefore this PR opts to turn it off, but also provides a very easy way of turning it back on should users desire it in the future for some reason.

### How to test
Test in any browser except Chrome to verify that no field triggers the browser's autofill feature.

### Related issue(s)/PR(s)
Provides feature requested in #6703 (although it widens the scope to all form fields).


JoshuaLuckers (approved):

Ibochkarev (approved):

JoshuaLuckers:
@smg6511 what changes did you do in your force-push?

smg6511:
That was just a rebase. I've been doing that every couple months on most of my PRs that have been open for a long time.

Mark-H (approved):

PR #16227, opened 2022-07-15
Fix only_full_group_by issue in permission processor query

### LINKED ISSUE #16226: MODx.combo.Permission combobox not working in MODX3
### Summary
The combobox `MODx.combo.Permission` is not working in MODX3.
I'll look into this shortly.

Edit: 
It's an `only_full_group_by` issue.

### Step to reproduce
It's used in two places, on the Edit Widget panel of the Dashboards page, and in the Create New Permission window when editing a policy template.

### Observed behavior
The connector response shows success and gives the total count, but no results are returned.
Thus, the dropdown doesn't get populated.

### Expected behavior
It should load a list of permissions that can be selected.

### Environment
MODX 3 (both 3.0.1 and latest git), PHP 7.4, MySQL 8.0.27

muzzwood:
### What does it do?
Fixes an `only_full_group_by` issue in `MODX\Revolution\Processors\Security\Access\Permission\GetList`

### Why is it needed?
So the permission combobox works when `only_full_group_by` is enabled.

### How to test
Have `only_full_group_by` enabled in MySQL and try to select a permission on either the widget edit panel on the Dashboards page, or create a new permission in a policy template.

### Related issue(s)/PR(s)
Resolves #16226

### DIFF
--- a/core/src/Revolution/Processors/Security/Access/Permission/GetList.php
+++ b/core/src/Revolution/Processors/Security/Access/Permission/GetList.php
@@ -68,7 +68,7 @@ class GetList extends GetListProcessor
             'modAccessPermission.description',
             'Template.lexicon',
         ]);
-        $c->groupby('modAccessPermission.name');
+        $c->groupby('modAccessPermission.id');
         $name = $this->getProperty('name', '');
         if (!empty($name)) {
             $c->where([
--- a/core/src/Revolution/Processors/Security/Access/Permission/GetList.php
+++ b/core/src/Revolution/Processors/Security/Access/Permission/GetList.php
@@ -63,12 +63,12 @@ class GetList extends GetListProcessor
     public function prepareQueryAfterCount(xPDOQuery $c)
     {
         $c->select([
-            'modAccessPermission.id',
+            'id' => 'MIN(modAccessPermission.id)',
             'modAccessPermission.name',
             'modAccessPermission.description',
             'Template.lexicon',
         ]);
-        $c->groupby('modAccessPermission.id');
+        $c->groupby('modAccessPermission.name, modAccessPermission.description, Template.lexicon');
         $name = $this->getProperty('name', '');
         if (!empty($name)) {
             $c->where([


smg6511 (changes requested):
This actually causes another problem, in that duplicates show up in the list. What you need to do is change lines 65-71 to:
```
$c->select([
    'id' => 'MIN(modAccessPermission.id)',
    'modAccessPermission.name',
    'modAccessPermission.description',
    'Template.lexicon',
]);
$c->groupby('modAccessPermission.name, modAccessPermission.description, Template.lexicon');
```

muzzwood:
Thanks @smg6511, you're not wrong! I've updated the PR with your suggestion. 👍🏻

smg6511:
@muzzwood - Hey, I put up an alternate PR on this subject after finding an issue while testing this (not an issue caused by your code, but an underlying one that wasn't initially apparent). I tried to think of a way to just suggest the solution I arrived at to add to your PR, but it involved changing other files not present here. Take a look and see what you think (#16273).

Mark-H:
Question... why is groupby() being used here anyway? I'm not seeing why this is supposedly aggregating rows beyond the (new) `MIN()` select? Wouldn't it be a simpler solution to remove the groupby line?

smg6511:
@Mark-H - We only want to group by the `name` so we can generate a unique list for the combo; remember that we don't have a separate table where each permission is defined (_once_, that is), so the table from which this list is derived will contain multiple rows for the same permission once a user starts adding new ACLs. That is why MIN is used on the id field (or MAX could've been used). If there's a better way, I'm all ears ;-) 

I'd personally rather see the permissions defined in a separate table, but that'd be a somewhat major structural change.

Mark-H:
Ah, so because it's fetching permissions from different access policies/templates it may have duplicates. Gotcha.

muzzwood:
Hi @smg6511 ,
Sorry for being unresponsive, it slipped off my radar. 🙈 
I'm more than happy for you to handle the fix in your larger PR.
Nice work, and thanks for looking into it! 
Closing this one for now.

PR #16230, opened 2022-07-24
[Port to 3.x] Check if modx-content is defined before calling doLayout

jscaltreto:
### What does it do?
Ports the fix from my previous PR https://github.com/modxcms/revolution/pull/14593 to 3.x

After upgrading to 3.0.1 the following issue resurfaced: occasionally widgets fail to load properly in the manager. I tracked it down to the call to doLayout() and found that sometimes when the listener gets called getCmp('modx-content') returned undefined. If I had to guess I'd say it's a race condition wherein sometimes the listener fires too early. Adding a check has resolved the issue entirely and I haven't encountered any ill effects.

### Why is it needed?
Adds a check to verify modx-content is defined before calling doLayout

### How to test
Intermittent issue of certain widgets not loading in the manager

### Related issue(s)/PR(s)

Previously fixed in 2.x: https://github.com/modxcms/revolution/pull/14593


jscaltreto:
@cla-bot check

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -364,7 +364,10 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                 ,autoDestroy: true
                 ,listeners:{
                     load: function(){
-                        Ext.getCmp('modx-content').doLayout(); /* Fix layout bug with absolute positioning */
+                        var cmp = Ext.getCmp('modx-content');
+                        if(typeof cmp !== "undefined") {
+                            cmp.doLayout(); /* Fix layout bug with absolute positioning */
+                        }
```
smg6511: What you've done certainly works, but please consider this:

I'd like to encourage everyone to employ ES6+ syntaxes and methods when submitting new or updated js code. Generally the `var` keyword should be avoided whenever possible, opting for the newer `let` or `const` keywords. Also, since you're just testing for truthiness/falseyness of your `cmp` variable, you can simply do this:
```
const cmp = Ext.getCmp('modx-content');
if (cmp) {
    cmp.doLayout();
}
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -380,7 +383,10 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
```
smg6511: Same comment as above.

smg6511 (changes requested):
Just FYI, in 3.x you can now use ES6+ code since we updated the build workflow a while back. In 2.x you're stuck with <=ES5, so the comment about the variable declaration will only apply to your port here. Thanks for your submission!

manager/assets/modext/widgets/core/modx.grid.js
jscaltreto: Thanks for the suggestion, @smg6511! I've updated the PR accordingly (and verified manually on my local installation that the behavior is unchanged).

manager/assets/modext/widgets/core/modx.grid.js
jscaltreto: Done!

smg6511 (approved):
I've approved the change but, if you would, can you let us know a way to replicate the issue? I'd like to see if there's something else in the process that should be looked into.

jscaltreto:
I don't have a way to reliably reproduce the issue. It seems to only happen intermittently with the `migxdb` input type, so it may be something specific to MIGX (CC @Bruno17) that isn't typically encountered in vanilla. When it happens, the result is that a `Loading` modal appears over the TV in the manager and it cannot be interacted with. 

From the debugging I was able to do, the stack trace when this is encountered points back to this: https://github.com/Bruno17/MIGX/blob/7d6d8b8bd6e1d36f12c48c4d09f1572aac343c26/core/components/migx/templates/mgr/grids/default.grid.tpl#L72 So it would seem that it's possible for that to run before the `modx-content` component has been loaded. It throws an exception which prevents the TV from loading fully.

smg6511:
Ok, yes, I've seen this issue myself on a couple of sites where migxdb is used. It's not something to address here in this PR. It may well be a timing issue. I might look into it a little further at some point... Thanks for the details!

Mark-H (approved):

Issue #16229, opened 2022-07-22
Blank page at sub.domain.com/manager/


tvquizphd:
### Summary
After "Installation finished in 1.9130 s" following [instructions](https://docs.modx.com/current/en/getting-started/installation/cli), I see only a blank page at `/manager/`. I'm hosting the page with nginx.

### Step to reproduce

```
wget https://modx.com/download/direct/modx-3.0.1-pl-sdk.zip
unzip modx-3.0.1-pl.zip
mv modx-3.0.1-pl sub.domain.com
cd sub.domain.com/setup/
```

Vim replacements after I copy `config.dist.new.xml` to `config.xml`:
```
:%s@/www/modx/@/var/www/sub.domain.com@gc
:%s/db_username/root/gc
:%s/db_password/pass/gc
:%s@/modx/@/@gc
```

I change `<http_host>localhost</http_host>` to `<http_host>sub.domain.com</http_host>`

I run:
```
php ./index.php --installmode=new --core_path=/var/www/sub.domain.com/core/
```

My /etc/php/8.0/fpm/php.ini has this:
```
pdo_mysql.default_socket=/var/run/mysqld/mysqlx.sock
mysqli.default_socket = /var/run/mysqld/mysqlx.sock
```

I ensure the fastcgi_pass in my nginx config is 

```
fastcgi_pass unix:/run/php/php-fpm.sock;
```

which matches my `/etc/php/8.0/fpm/pool.d/www.conf `:

```
listen = /run/php/php-fpm.sock
```

This code runs correctly:

```
<?php
/* Connect to an ODBC database using driver invocation */
$dsn = 'mysql:dbname=testdb;host=localhost';
$user = 'dbuser';
$password = 'dbpass';

try {
  $dbh = new PDO($dsn, $user, $password);
} catch (PDOException $e) {
  echo 'Connection failed: ' . $e->getMessage();
}
?>
```

### Observed behavior
The manager page is blank. It is a 200 response with a completely blank file. Here are the response headers:

```
Cache-Control: no-store, no-cache, must-revalidate
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/html; charset=UTF-8
Date: Fri, 22 Jul 2022 20:17:52 GMT
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Pragma: no-cache
Server: nginx/1.18.0 (Ubuntu)
Transfer-Encoding: chunked
```

In my nginx logs, i see the following:

```
[...]

PR #16231, opened 2022-07-25
Improve default config xml files

tvquizphd:
### What does it do?
I've changed the default database server from "localhost" to "127.0.0.1".

### Why is it needed?
This circumvents the `Connection failed: SQLSTATE[HY000] [2002] No such file or directory` that I encountered on a fresh install of Modx Revolution. See [Connection failed: SQLSTATE[HY000] [2002] No such file or directory](https://forum.infinityfree.net/t/connection-failed-sqlstate-hy000-2002-no-such-file-or-directory/43590).

### How to test

**Prepare php and mysql for installation**:

- In mysql, run `create database modx_modx`
- Modify /etc/php/8.0/*/php.ini by setting `mysqli.default_socket` and `pdo_mysql.default_socket` correctly
- Modify /etc/php/8.0/*/php.ini by setting `date.timezone = America/New_York`

**Run CLI installation script with config**

```
wget https://gist.githubusercontent.com/tvquizphd/b4eaab96ac944328b9939deafc9940f5/raw/aa095175827519c8456d012e4543e54caf297c82/config.xml
wget https://raw.githubusercontent.com/craftsmancoding/modx_utils/master/installmodx.php
wget https://modx.com/download/direct/modx-3.0.1-pl-sdk.zip
```

Replace "user", "pass", and "domain.example.com" in config.xml, then run installation:

```
php installmodx.php --config=config.xml --zip=modx-3.0.1-pl-sdk.zip --target=jeannie.hoff.in --version=3.0.1 --installmode=new
```
### Related issue

This resolves [https://github.com/modxcms/revolution/issues/16229](issue 16229).

### DIFF
--- a/setup/config.dist.new.xml
+++ b/setup/config.dist.new.xml
@@ -8,7 +8,7 @@
 -->
 <modx>
 	<database_type>mysql</database_type>
-    <database_server>localhost</database_server>
+    <database_server>127.0.0.1</database_server>
     <database>modx_modx</database>
     <database_user>db_username</database_user>
     <database_password>db_password</database_password>
--- a/setup/config.dist.upgrade-advanced.xml
+++ b/setup/config.dist.upgrade-advanced.xml
@@ -8,7 +8,7 @@
 -->
 <modx>
 	<database_type>mysql</database_type>
-    <database_server>localhost</database_server>
+    <database_server>127.0.0.1</database_server>
     <database>modx_modx</database>
     <database_user>db_username</database_user>
     <database_password>db_password</database_password>


tvquizphd:
@cla-bot check

smg6511:
There's just one problem with this: although it typically is, `localhost` does not have to be assigned to 127.0.0.1; it can be assigned to other addresses within that block (1-255).

tvquizphd:
> There's just one problem with this: although it typically is, `localhost` does not have to be assigned to 127.0.0.1; it can be assigned to other addresses within that block (1-255).

In that case, I agree this isn't the solution. Would you instead support a new section in the "troubleshooting installation" section of the documentation? I've observed that this localhost mysql error causes any valid modx page (such as the manager or root) to display a blank white screen with a 0-byte file sent with a 200 response code.

If I were more comfortable with PHP, I'd consider catching this error and displaying an error screen instead of just an empty file, but I would be competent enough to at least update the documentation.

Ibochkarev (approved):

JoshuaLuckers:
> Would you instead support a new section in the "troubleshooting installation" section of the documentation?

@tvquizphd if you have to opportunity to do so, yes please update the documentation.

JoshuaLuckers:
Since there is an agreement about this not being the best solution I will close this PR.

Issue #15136, opened 2020-06-26
User-entered data in a TV with a Listbox (Multi-Select) type is not displayed in the resource form


Ruslan-Aleev:
### Summary
There is another problem with setting "**Force Selection to List:**" (forceSelection). It allows you to enter values not from the list. And for a multiple lis the values entered by the user are not displayed in the TV, in the resource form.

Listbox (Multi-Select):

![listbox_m_forse](https://user-images.githubusercontent.com/12523676/85009571-50965a00-b167-11ea-908a-6ed262004f51.gif)

Although there is no such problem for Listbox (Single-Select):

![listbox_s_forse](https://user-images.githubusercontent.com/12523676/85009578-512ef080-b167-11ea-931d-bcfede1e0127.gif)

p.s. Personally, I am against the "forceSelection" setting as a whole, it creates potential bugs. Why let the user enter values other than specified in the "Input Option Values"? You can simply use TV text if you need data not from the "Input Option Values".

### Expected behavior
Display user input

### Environment
MODX 2.7>
Related with https://github.com/modxcms/revolution/pull/14957

JoshuaLuckers:
To clarify: the TV doesn't show the manually entered value that's saved in the database for that particular resource, which it should just like the single select.

PR #16241, opened 2022-08-07
View values in TV listbox-multiple then it manually entered

Alexij2:
### What does it do?
- Displays values manually entered by the manager in the admin panel. For MODX 2.x

### Why is it needed?
- It is necessary to fix the display problem. #15136 (it for MODX 3.x, but for 2.x acutally too)

### How to test
- Created TV listbox-multiple add to template
- Open resource and type in this TV manually some words. -> press Enter -> Save Document.
- Reload page (F5), Open TV and see changes.

### Related issue(s)/PR(s)
#15136

### DIFF
--- a/core/model/modx/processors/element/tv/renders/mgr/input/listbox-multiple.class.php
+++ b/core/model/modx/processors/element/tv/renders/mgr/input/listbox-multiple.class.php
@@ -35,6 +35,14 @@ class modTemplateVarInputRenderListboxMultiple extends modTemplateVarInputRender
         $orderedItems = array();
         // loop trough the selected values
         foreach ($value as $val) {
+            // add custom value (from TV) if it isset
+            if (isset($val)) {
+                $orderedItems[] = array(
+                    'text' => $val,
+                    'value' => $val,
+                    'selected' => 1,
+                );
+            }
             // find the corresponding option in the items array
             foreach ($items as $item => $values) {
                 // if found, add it in the right order to the $orderItems array


Jako:
Thanks for helping to make MODX better. This fix should check, wether the TV option forceSelection is deactivated and not add the unsupported values then. Otherwise the patch looks ok.

What happens if an unsupported value is in the TV and the TV is saved?

smg6511:
I just put up a PR to fix this in 3.x. Take a look (_see_ #16242), as it may inform how you approach this. I believe the un-patched file is virtually the same for 2.x and 3.x, so you may be able to use what I've done if you like. The one thing I see (that would need fixing) is that you probably end up duplicating selected values in the list when they are among the pre-defined input options values.

Alexij2:
> What happens if an unsupported value is in the TV and the TV is saved?

It is stored normally. It is not duplicated in the Database. Also, there is no duplication of the value, JS does not allow this to be done in the admin panel. And even if it is possible to write the value, then after saving and when updating the page (F5), the data is displayed correctly.

> The one thing I see (that would need fixing) is that you probably end up duplicating selected values in the list when they are among the pre-defined input options values.

Duplication of the value does not occur. Js does not allow this to be done in the admin panel. I also thought that it would be duplicated, but as practice has shown, this does not happen, because unique values come to be recorded in the field DB.

core/model/modx/processors/element/tv/renders/mgr/input/listbox-multiple.class.php
```diff
@@ -35,6 +35,14 @@ public function process($value,array $params = array()) {
         $orderedItems = array();
         // loop trough the selected values
         foreach ($value as $val) {
+            // add custom value (from TV) if it isset
+            if (isset($val)) {
+                $orderedItems[] = array(
+                    'text' => $val,
+                    'value' => $val,
+                    'selected' => 1,
+                );
```
smg6511, line 44: Small change to treat the custom input values the same as the pre-defined ones (htmlspecialchars). Also, I encourage the modernization of any passage of code you add or edit (here, use the short array declaration using only brackets).
```suggestion
            $val = htmlspecialchars($val, ENT_COMPAT, 'UTF-8');
            $orderedItems[] = [
                'text' => $val,
                'value' => $val,
                'selected' => 1
            ];
```

smg6511 (changes requested):
I tested this solution and it does fix the issue. The solution I created for 3.x could also just be dropped in, as the class definitions are exactly the same (comparing the original 3.x one before my merged changes and the 2.x one being changed here). Honestly, it's been too long for me to _fully_ remember the motive for my 3.x approach although I do remember finding the original code somewhat hard to follow and wanted to make the process more immediately readable. 

My first choice would be to drop in the 3.x solution, but I'd also approve this one with the more minor change I specified above.

theboxer:
@Alexij2 can you incorporate the suggested changes? or @smg6511 would you be able to port your 3.x fix to 2.x?

smg6511:
@theboxer - Sure thing, just posted the back port.

theboxer:
Closing this in favour of #16640 thank you @Alexij2 and @smg6511

PR #16242, opened 2022-08-08
Fix rendering of user-entered values in editable listbox-multiple TVs

### LINKED ISSUE #15136: User-entered data in a TV with a Listbox (Multi-Select) type is not displayed in the resource form

smg6511:
### What does it do?
Refactored the listbox-multiple render class to ensure values that are not in a TV’s pre-defined options are rendered properly. The code has been restructured to make it more intuitive, as the previous code had a rather convoluted way of collecting and separating the selected (saved) and unselected values.

### Why is it needed?
User-entered values are saved to the database but are not subsequently rendered when reloading the Resource form.

### How to test
1. Create a listbox-multiple TV with _Enable Type-Ahead_ on and _Require Match_ set to “No.”
2. Enter a set of Input Options for the TV.
3. Make the TV available to a Resource and experiment with making selections from the dropdown, as well as entering ones directly in the field. Save the Resource and re-load it to verify all values show in the TV field as expected.

### Related issue(s)/PR(s)
Resolves #15136.

### DIFF
--- a/core/src/Revolution/Processors/Element/TemplateVar/Renders/mgr/input/listbox-multiple.class.php
+++ b/core/src/Revolution/Processors/Element/TemplateVar/Renders/mgr/input/listbox-multiple.class.php
@@ -61,10 +61,10 @@ class modTemplateVarInputRenderListboxMultiple extends modTemplateVarInputRender
 
         // Ensure custom values are displayed when the listbox is editable
         if (isset($params['forceSelection']) && empty($params['forceSelection'])) {
-            $customSelections = array_diff($savedValues, $optsValues);
-            if (!empty($customSelections)) {
+            $customValues = array_diff($savedValues, $optsValues);
+            if (!empty($customValues)) {
                 $customData = [];
-                foreach ($customSelections as $customValue) {
+                foreach ($customValues as $customValue) {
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

smg6511:
@JoshuaLuckers - Quick note: Because this is a pretty major bug fix and not a new feature or just an improvement, I'd suggest changing the milestone to 3.0.2. (I'm actually surprised we haven't seen more complaints about the bug.) While I know I could change this myself, I thought maybe you had a particular reason for making it a minor release instead of a patch one.

PR #16244, opened 2022-08-08
Prioritize exact matches in manager search result

### LINKED ISSUE #16236: Manager Search should always find named elements
### Summary
If a site has a lot of elements with similar names and/or elements containing a search term, any element with a name that is an exact match should be shown. 

Examples are my own implementations.

Example 1, MODX 2.8.4:
I search for a snippet named 'GetImage.' I have many snippets with that prefix but the actual 'GetImage' is not shown.

Example 2 MODX 3.0.1:
I search for a snippet named 'functions.' I run that snippet in other snippets, many of which are found, but the actual 'functions' is not shown.

It many take an extra moment to show the exact match, but this is back-end time, so will not impact site visitors.

halftrainedharry:
### What does it do?
Adds a sort to the queries for the manager search, that puts an exact match of the "name" field at the top of the result.

### Why is it needed?
The search in the manager only shows 5 results for each type of element. If there are a lot of elements with similar names, even one that matches the search term exactly may not be shown in the result.

### How to test
Create some snippets/chunks/etc. with similar names. Verify that the snippet/chunk that exactly matches the search term is shown first.

### Related issue(s)/PR(s)
Resolves #16236 (for MODX 2.x)

### DIFF
--- a/core/model/modx/processors/search/search.class.php
+++ b/core/model/modx/processors/search/search.class.php
@@ -184,6 +184,7 @@ class modSearchProcessor extends modProcessor
                 'modResource.context_key:IN' => $contextKeys,
             )
         ));
+        $c->sortby('IF(`modResource`.`pagetitle` = ' . $this->modx->quote($this->query) . ', 0, 1)');
         $c->sortby('modResource.createdon', 'DESC');
 
         $c->limit($this->maxResults);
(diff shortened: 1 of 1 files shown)


JoshuaLuckers (approved):

Mark-H (approved):

opengeek:
@halftrainedharry — any chance you could port this to 3.x? I've tried cherry-picking in various ways to apply it, but I'm not having any luck.

PR #16249, opened 2022-08-18
Made it possible to edit htaccess from the admin panel

### LINKED ISSUE #16221: Can't make .htaccess editable from manager in 3.0
### Summary

In 2.x, adding `htaccess` to the `upload_files` system setting would allow you to edit .htaccess files from the manager. That doesn't seem to work in 3.0.1 anymore.

### Step to reproduce
Add `htaccess` to the `upload_files` system setting.

### Observed behavior
Error `File extension `` is not permitted.`

### Expected behavior
Allow editing .htaccess files when opted in.

### Environment
Probably all.

Ibochkarev:
### What does it do?
Made it possible to edit htaccess from the admin panel

### Why is it needed?
For ease of editing from the admin panel

### How to test
Add `htaccess` to the `upload_files` system setting. And rename `ht.access` to `.htaccess`

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/16221

### DIFF
--- a/core/src/Revolution/Processors/Browser/Browser.php
+++ b/core/src/Revolution/Processors/Browser/Browser.php
@@ -143,7 +143,7 @@ abstract class Browser extends Processor
     {
         $file = rawurldecode($file);
         $file = strip_tags($file);
-        $file = preg_replace('#^[.\/]+#u', '', $file);
+        $file = preg_replace('#^(\.{2}|/)+#u', '', $file);
         $file = strip_tags($file);
 
         return $file;
--- a/core/src/Revolution/Processors/Browser/Browser.php
+++ b/core/src/Revolution/Processors/Browser/Browser.php
@@ -143,7 +143,8 @@ abstract class Browser extends Processor
     {
         $file = rawurldecode($file);
         $file = strip_tags($file);
-        $file = preg_replace('#^(\.{2}|/)+#u', '', $file);
+        $file = preg_replace('#\.(?![\w\-\~])#u', '', $file);
+        $file = preg_replace('#\/{2,}#', '/', $file);
         $file = strip_tags($file);
 
         return $file;
--- a/manager/controllers/default/system/file/create.class.php
+++ b/manager/controllers/default/system/file/create.class.php
@@ -84,7 +84,7 @@ class SystemFileCreateManagerController extends modManagerController
         $directory = !empty($scriptProperties['directory']) ? $scriptProperties['directory'] : '';
 
         $this->fileRecord = [
-            'directory' => strip_tags(preg_replace('#^[.\/]+#u', '', $directory)),
+            'directory' => strip_tags(preg_replace('#^(\.{2}|/)+#u', '', $directory)),
             'source' => $source->get('id'),
         ];
 


JoshuaLuckers (approved):

benniledl (changes requested):
Use this Regex `^(\.{2,}|/)+`
to match 2 or more dots, otherwise any uneven amount of dots would not be matched

smg6511 (approved):

JoshuaLuckers:
> Use this Regex `^(\.{2,}|/)+`
> to match 2 or more dots, otherwise any uneven amount of dots would not be matched

@benni1516 With `preg_replace` the regex `'#^(\.{2}|/)+#u` seems to work fine for me. 
E.g: `.....htaccess` ends up in `.htaccess` which is the expected result.

benniledl:
@JoshuaLuckers 
hello, sorry if I'm wrong, I thought about something like this
https://regex101.com/r/JPrLdx/1

smg6511:
@benni1516 - Just curious (and not making any assumptions); have you pulled this down and tested this or are you commenting based solely upon looking at the code changes? I ask because if you've not tested this I'd really like to encourage you to do so. While comments and observations are helpful, we're _really_ in need of folks willing to go the extra mile and test PRs [shout out to anyone reading this who's not testing ... yet ;-) ]. 

@Ibochkarev, @JoshuaLuckers, @benni1516 - Anyway, after playing with this a bit more, if you really push it with a file name like `./../.../..../..htaccess`, then you end up with likely an unwanted file `..htaccess`. Interestingly, it's the presence of the various dots and slashes ahead of the last part of the path that causes `..htaccess` to be created. Bottom line is you can end up with a file where part of the name consists of dots. Here's the rub: That does not happen _without_ the change this PR applies. That said, the regex could be thought through more to avoid this issue OR this could be considered a non-issue under the assumption that virtually no one is going to enter file names with a bunch of dots and slashes.

opengeek:
The proper way to avoid directory traversal is to compare the realpath() of the path provided by the user with a pre-defined base path for the operation. I don't see any defined base path here, but it absolutely should be using one.

benniledl:
@smg6511 sorry I didn't test it out, I will test things out before commenting in the future :)

smg6511:
@benni1516 - No apologies necessary ... just wanted to encourage you test ;-)

All - FYI, changed status as a little more needs to be done with the regex. I'm playing with something to propose and will put that up as a requested change (or in a comment) momentarily...

core/src/Revolution/Processors/Browser/Browser.php
```diff
@@ -143,7 +143,7 @@ public function sanitize($file)
     {
         $file = rawurldecode($file);
         $file = strip_tags($file);
-        $file = preg_replace('#^[.\/]+#u', '', $file);
+        $file = preg_replace('#^(\.{2}|/)+#u', '', $file);
```
smg6511:
```suggestion
        $file = preg_replace('#\.(?![\w\-\~])#u', '', $file);
        $file = preg_replace('#\/{2,}#', '/', $file);
```
This strategy here is to take care of the dots first, then simply reduce multiple slashes to single ones. The one thing to consider here is whether the class of allowable chars directly after a valid dot is flexible enough; I think so, but others should chime in if not.

manager/controllers/default/system/file/create.class.php
```diff
@@ -84,7 +84,7 @@ public function process(array $scriptProperties = [])
         $directory = !empty($scriptProperties['directory']) ? $scriptProperties['directory'] : '';
 
         $this->fileRecord = [
-            'directory' => strip_tags(preg_replace('#^[.\/]+#u', '', $directory)),
```
smg6511, line 87: If the above change to the Browser class is used, this controller change can be reverted.

smg6511 (changes requested):
Made some suggestions after thinking about the regex further.

opengeek:
If you can specify any directory, why are we preventing directory traversal with regex? This makes no sense. We need to be using basename() to get the filename and using realpath() to compare the provided path with a properly limited "base" path for the operation.

smg6511:
What I'm trying to do is, taking into account that you _can_ traverse (and dynamically create) directories moving downward, keep the resulting sanitized path as clean as possible (and non-helpful mysterious error messages) while allowing .files and .directories to be written and edited. Upward traversal wasn't possible anyway. 

Take for instance creating a new file with this path: `newdir5/../test.txt`. With the current regex, that path will make it past this initial sanitize (in Browser) and ultimately create a file at `newdir5/text.txt` because a second pass is made by sanitizePath (in modMediaSource) when createObject is called. However, an error is triggered along the way — a possible explanation is when the file is read back in, the initial invalid path (due to `/../`) is perhaps being used in that process (unconfirmed). At any rate, I adjusted the regex to avoid this sort of thing, as the on-screen error is confusing and useless IMO (there's no error given in the modx log).

Issue #16186, opened 2022-05-16
Notempty Output Filter cuts off value if it includes snippet properties


newswanger:
### Summary
Not empty output filter will truncate the value if it includes a snippet with properties.

### Step to reproduce
Add the following to a template:
[[*content:notempty=\`[[!Wayfinder? &startId=\`0\` &level=\`1\`]]\`]]

### Observed behavior
Rendering this in a browser will display:
[[!Wayfinder? &startId=

### Expected behavior
The complete snippet tag should be returned from the filter and processed. This filter does work  MODX 2.8.4

### Environment
MODX 3.0.1, apache2 , mariaDB

Bruno17:
you know, conditionals like that, will allways parse the snippet inside, also when content is empty.
Better try using the mosquito method, like that:
```[[[[*content:notempty=`Wayfinder? &startId=`0` &level=`1``]]]]```

Maybe, this will also help with your issue.
see: https://modx.com/blog/2012/09/14/tags-as-the-result-or-how-conditionals-are-like-mosquitoes/

JoshuaLuckers:
Related to issue #16044

I've added test data to the Input Filter Unit Test in my repo, this will fail the Unit Test: https://github.com/JoshuaLuckers/revolution/commit/004b877ad0e597e77b19d4a6a09477495bdde4ad

newswanger:
I tracked the problem to the `preg_match_all` in modInputFilter.php.

[revolution/core/src/Revolution/Filters/modInputFilter.php](https://github.com/modxcms/revolution/blob/697a0286117a7e30b19c7f78ef30803806daf6c3/core/src/Revolution/Filters/modInputFilter.php#L56)

I think I have a fix that works:
            if (preg_match_all('~:([^:=]+)(?:=\`((?:(?:[^:=][\s\S]\*?|(?R))\*?))\`$)?~s', $modifiers, $matches)) {

The addition of the '$' makes sure it matches the last tick mark in the string.

newswanger:
Thanks @Bruno17, the example is just for testing. In the real world scenario the output filter looks like:

[[*content:notempty=\`<div class="content'>[[*content]]<\/div>\`]]

Then the end user includes a snippet with properties in the content field.

JoshuaLuckers:
> I think I have a fix that works: if (preg_match_all('~:([^:=]+)(?:=`((?:(?:[^:=][\s\S]*?|(?R))*?))`$)?~s', $modifiers, $matches)) {
[...]

PR #16254, opened 2022-08-28
Revert and break nested output filters

JoshuaLuckers:
### What does it do?
Revert and break nested output filters to fix a lot of other issues with output filters introduced in PR #14458

### Why is it needed?
To fix a lot of issues with output filters being broken.

### Related issue(s)/PR(s)
Related issues: 
- #16044 
- #16186

Related PR: #14458

### DIFF
--- a/core/src/Revolution/Filters/modInputFilter.php
+++ b/core/src/Revolution/Filters/modInputFilter.php
@@ -53,7 +53,7 @@ class modInputFilter
             $matches = [];
             $name = trim(substr($output, 0, $splitPos));
             $modifiers = substr($output, $splitPos);
-            if (preg_match_all('~:([^:=]+)(?:=`((?:(?:[^:=][\s\S]*?|(?R))*?))`)?~s', $modifiers, $matches)) {
+            if (preg_match_all('~:([^:=]+)(?:=`(.*?)`[\r\n\s]*(?=:[^:=]+|$))?~s', $modifiers, $matches)) {
                 $this->_commands = $matches[1]; /* modifier commands */
                 $this->_modifiers = $matches[2]; /* modifier values */
             }


Jako (approved):

Jako:
The tests have to be reverted too, since nested output filters don't work anymore (at least in the test). The code is reverted to the state before #14458.

JoshuaLuckers:
> The tests have to be reverted too, since nested output filters don't work anymore (at least in the test). The code is reverted to the state before #14458.

Yes you’re right, I hope to find some time to actually fix the issue but it’s complicated to do. Once everyone agrees with reverting the change and breaking nested output filters I’ll remove the breaking test cases.

JoshuaLuckers:
@opengeek @Mark-H what are your thoughts on this?

Mark-H:
I've not followed this issue closely enough to have a strong opinion, though it sounds like this may be related to some upgrade issues I've seen myself. 

I'm seeing two integrators who have been working on this issue saying we need this, and am happy to lend those my support. 

Maybe as the next step after this we can build a ton more extra test cases (some of which may fail for now) to make sure we avoid going back and forth on this repeatedly. Clearly there needs to be work on the parser but very few people are comfortable with this particular level of spaghetti and inferred behavior.

rthrash:

https://community.modx.com/t/calling-uncached-inside-if-or-similar-output-modifier/5740/2

rthrash:

https://community.modx.com/t/migx-php-8-0-nested-getimage-calls-is-it-a-bug/5750/12

rthrash:

https://community.modx.com/t/template-tags-are-not-parsed-under-certain-conditions/5779/2

rthrash:

https://community.modx.com/t/modx3-broken-output-filters/5906/1

Mark-H:
I've just created an alternative proposal to fix this issue, and would appreciate lots and lots of testing and feedback on that: #16288

modxcommunity:

https://community.modx.com/t/modx3-output-filters-break-with-nested-snippets-chunks/6878/1

PR #16271, opened 2022-09-24
Enable SVG thumbnail preview for image TVs

### LINKED ISSUE #16270: SVG support for image TV preview #13556...
Hi @JoshuaLuckers,
this fix was for Modx Revolution 2.5.8-dev. Is it possible to find a corresponding fix for Modx 3.0 - 3.0.2?
Because I have the same problem with Modx 3. - BTW. the files which are listed above do not exist in Modx 3.
(btw 2: the SVG files are shown in my collection view but not on my TV.)

_Originally posted by @eire69k in https://github.com/modxcms/revolution/issues/13556#issuecomment-1255229485_

smg6511:
### What does it do?
Skips thumbnailing of SVGs via phpThumb.

### Why is it needed?
Fixes inability to preview image TVs containing SVG images.

### How to test
1. Run `grunt build` at `_build/templates/default` and ensure your browser and modx caches are cleared.
2. Save two SVG images, associating one with the default Filesystem source and the other with an alternate media source.
3. Create two image TVs, one for each source you saved your SVGs to.
4. Assign the TVs to a Template and insert your images into a Resource assigned that Template to verify the preview functionality.

### Related issue(s)/PR(s)
Resolves #16270 and is a port of #15865.


Ibochkarev (approved):

core/src/Revolution/Processors/System/PhpThumb.php
```diff
@@ -71,6 +72,13 @@ public function process()
             return '';
         }
 
+        if (pathinfo($src)['extension'] == 'svg') {
```
JoshuaLuckers:
```suggestion
        if (strtolower(pathinfo($src, PATHINFO_EXTENSION)) === 'svg') {
```

smg6511: Don't know if using the constant is less expensive, but the lowercasing before comparison is definitely a good move. This should be done in the fix for 2.x as well.

JoshuaLuckers: @smg6511 i took this part of the code from the port you mentioned.

smg6511: Too funny. I must not have looked at it that closely! ;-)

Mark-H (approved):

PR #16272, opened 2022-09-25
13970 ignore synlink

### LINKED ISSUE #13970: MODX 3 - No files displayed if a symlink exists in the directory
### Step to reproduce
Create a symlink in a directory that you have a media source set up for. Try to browse that directory (not the symlink, but the directory it is contained in). A popup error is shown saying links aren't supported, but none of the other files/directories are shown.

### Observed behavior
Error popup, but no files/directories.

### Expected behavior
If it cannot support the symlink, it should still show other files and directories.

### Environment
MODX 3 @ 0ba86ec05d9d99327d55d45083312b98ac80229a
PHP 7.1.12

sepiariver:
### What does it do?
Changed the `LocalFilesystemAdapter` config in `core/src/Revolution/Sources/modFileMediaSource.php` to skip (ignore) symlinks. Also updated unit tests and test readme to correctly instruct how to run phpunit.

### Why is it needed?
Prior to this, in the base branch, `modFileMediaSource`  would error when listing directory contents if that directory contained a symlink. Users could not inspect the contents of the directory at all in this case. With this PR it ignores symlinks and lists the other content.

### How to test
`composer run-script phpunit`
You can also test manually by logging into the manager, listing the contents of a directly like `assets` in the File Tree, then create a symlink there, and in the manager you will still be able to list the contents of the directory. It is expected that the symlink will simply be ignored and not displayed.

### Related issue(s)/PR(s)
Resolves #13970  and related issues that have the same cause

### DIFF
--- a/_build/test/README.md
+++ b/_build/test/README.md
@@ -7,4 +7,4 @@ Copy `_build/test/properties.sample.inc.php` to `_build/test/properties.inc.php`
 * `$properties['mysql_string_username']= '';` database username
 * `$properties['mysql_string_password']= '';` database password
 
-From the commandline at the root of the projeect, run `composer run-script phpunit`
+From the commandline at the root of the project, run `composer run-script phpunit`
--- a/_build/test/README.md
+++ b/_build/test/README.md
@@ -7,4 +7,4 @@ Copy `_build/test/properties.sample.inc.php` to `_build/test/properties.inc.php`
 * `$properties['mysql_string_username']= '';` database username
 * `$properties['mysql_string_password']= '';` database password
 
-from the commandline change to the test directory and run `phpunit`
+From the commandline at the root of the projeect, run `composer run-script phpunit`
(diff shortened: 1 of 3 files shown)


sepiariver:
@cla-bot check

_build/test/README.md
```diff
@@ -7,4 +7,4 @@ Copy `_build/test/properties.sample.inc.php` to `_build/test/properties.inc.php`
 * `$properties['mysql_string_username']= '';` database username
 * `$properties['mysql_string_password']= '';` database password
 
-from the commandline change to the test directory and run `phpunit`
\ No newline at end of file
+From the commandline at the root of the projeect, run `composer run-script phpunit`
```
smg6511: Just one little typo to fix:
```suggestion
From the commandline at the root of the project, run `composer run-script phpunit`
```

smg6511 (approved):
This works and is the best way to handle the incompatibility. Flysystem purposely does not support symlinks (it's not a bug or oversight); as such, it's better to have the reading of symlinks silently fail than to have all files in a directory be inaccessible.

This behavior should be noted in the MODX documentation if it's not already.

Mark-H (approved):

PR #16273, opened 2022-09-25
Alternate Permissions Combo Fix

smg6511:
### What does it do?
This follows on work by @muzzwood (#16227) with additional fixes.

### Why is it needed?
The referenced PR works to fix the immediate issue, but I found an additional one when doing some final testing (playing with the paging of the combo): The total count was wrong, due to issues with how getCount works and the fact that aggregate functions need to be used to produce the combo's list correctly (without duplicates). Note that the original code (pre-@muzzwood's changes) attempted to use DISTINCT to do this, but that's not possible in this case, as there's only one column we desire to be distinct but we're selecting four.

The proposed solution queries for the total count in a different way, setting its value in a class property. In this way, the default getCount can be side-stepped for instances such as this without changing any method signatures or overriding the getData method in the child class (which would just produce 99% duplication of its parent class method).

Otherwise, a separate commit makes a minor change to the js, adding listeners to load the store when one happens to delete the current value in the combo's field. Previously to this addition, the paging would not get loaded correctly (or at all) after erasing the field's value.

### How to test
Rebuild, clear cache, and test the combo at:
`manager/?a=system/dashboards/widget/update&id=1` and
`manager/?a=security/access/policy/template/update&id=2` (Shown when you Create a new permission on the grid)

### Related issue(s)/PR(s)
Based on #16227

### DIFF
--- a/core/src/Revolution/Processors/Security/Access/Permission/GetList.php
+++ b/core/src/Revolution/Processors/Security/Access/Permission/GetList.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
@@ -44,14 +45,21 @@ class GetList extends GetListProcessor
      */
     public function prepareQueryBeforeCount(xPDOQuery $c)
     {
+        $n = $this->modx->newQuery($this->classKey);
+        $n->select([
+            'total' => 'COUNT(DISTINCT `modAccessPermission`.`name`)'
+        ]);
+
         $c->leftJoin(modAccessPolicyTemplate::class, 'Template');
-        $c->query['DISTINCT'] = 'DISTINCT';
+
         $query = $this->getProperty('query', '');
         if (!empty($query)) {
-            $c->where([
-                'modAccessPermission.name:LIKE' => '%' . $query . '%',
-            ]);
+            $query = ['modAccessPermission.name:LIKE' => '%' . $query . '%'];
+            $c->where($query);
+            $n->where($query);
         }
+
+        $this->listTotal = $this->modx->getObject($this->classKey, $n)->get('total');
 
         return $c;
     }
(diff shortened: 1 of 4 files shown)


core/src/Revolution/Processors/Security/Access/Permission/GetList.php
```diff
@@ -44,15 +45,22 @@ public function initialize()
...
-        $c->query['DISTINCT'] = 'DISTINCT';
+
         $query = $this->getProperty('query', '');
         if (!empty($query)) {
-            $c->where([
-                'modAccessPermission.name:LIKE' => '%' . $query . '%',
-            ]);
+            $query = ['modAccessPermission.name:LIKE' => '%' . $query . '%'];
+            $c->where($query);
+            $n->where($query);
```
Mark-H: ~~duplicate line~~ very similar looking line 😂

smg6511: Yes, it's the main query vs. the total count ($n) one. If that bothers you, I can change the variable name. Let me know...

Mark-H: When it comes to readability of the code, there's a perfect balance somewhere between simple variable names like `$c` and `$accessPermissionNameQueryInstance`. :D 

Because of the similarity I probably would've named it slightly differently (perhaps `$totalQuery` to indicate its purpose) but I don't feel very strongly about it so no need to change imo.

smg6511: I debated about this detail before you mentioned it. I went ahead and applied your suggestion; honestly I normally make my vars descriptive so I'm not sure why I landed on $n ... maybe matching the style of $c.

JoshuaLuckers (approved):

Mark-H (approved):

PR #16276, opened 2022-10-01
REST service case insensitive sub level search

Bournwog:
Problem: 
request `/rest/test/test2` search only in `Controllers/test/test2.php`
if path is like that `Controllers/Test/Test2.php` need to send this request `/rest/Test/Test2`

Fixed for case insensitive search, for all cases
`Controllers/test/test2.php`
`Controllers/Test/Test2.php`
`Controllers/test/Test2.php`
etc

### DIFF
--- a/core/src/Revolution/Rest/modRestService.php
+++ b/core/src/Revolution/Rest/modRestService.php
@@ -219,7 +219,7 @@ class modRestService
         
         $iterateDirectories = $this->iterateDirectories($basePath . '/*' . $controllerClassFilePostfix, GLOB_NOSORT);
     	if ($iterateDirectories) {           
-        	$contollersList = array();     
+        	$contollersList = [];     
         	
 	        foreach ($iterateDirectories as $controller) {
 	            $controller = $basePath != '/' ? str_replace($basePath, '', $controller) : $controller;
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
@cla-bot check

core/src/Revolution/Rest/modRestService.php
```diff
@@ -216,28 +216,42 @@ protected function getController()
...
-            if (strnatcasecmp($expectedFile, $controllerFile) == 0) {
-                require_once $basePath . $controller;
-
-                return $controllerClassPrefix . $controllerClassSeparator . $controllerClass;
-            }
-        }
+        
+        $iterateDirectories = $this->iterateDirectories($basePath . '/*' . $controllerClassFilePostfix, GLOB_NOSORT);
+    	if ($iterateDirectories) {           
+        	$contollersList = array();
```
JoshuaLuckers:
```suggestion
        	$contollersList = [];     
```

Ibochkarev: @Bournwog Up

Mark-H (approved):
I don't personally have an implementation of the rest service to properly test this with, but I don't see any issues with it from a code perspective.

opengeek:
Like Mark, I do not have any implementations of this code to test. If we get another review approval, I'll assume it was tested appropriately and can be merged.

Bruno17:
I could test it on 2.8.4 and it seems to solve case insensitive sub level search not working and works everything as expected.
Codewise, there isn't any difference on 3.x in this part, so I think it should work on both 2.8.x and 3.x

JoshuaLuckers (approved):
Based on @Bruno17's comment.

PR #16280, opened 2022-10-10
Remove the scrolling overflow from the modx-browser panel

Jako:
### What does it do?
Remove margins from the container and add them to the border layout

### Why is it needed?
Remove the scrolling overflow from the modx-browser panel

Before:

Bildschirmfoto 2022-10-10 um 14 09 38

After:

Bildschirmfoto 2022-10-10 um 14 11 20

### How to test
Add the changes and check the browser panel in the MODX media browser CMP. It currently has to be tested with responsive sizes. The CSS can perhaps be simplified there. 

Maybe it is better to compose the page with ExtJS directly on small screens instead of overriding the ExtJS functionality by CSS.

### Related issue(s)/PR(s)
None known.


smg6511 (approved):
This does make a subtle improvement for the desktop view. FYI, `border: false` has no effect (both the one added and the one in the center region item), so you could delete those config lines.

Jako:
`border: false` removes a not needed class in the wrapper element.

PR #16286, opened 2022-10-18
Exclude @SELECT binding from properties check in MODX 3

### LINKED ISSUE #16200: modTemplateVar::parseBinding - third parameter is invalid JSON
### Summary
I have a template variable which contains the input option values as:
```php
@SELECT CONCAT(`pagetitle`, ' (', `id`, ')') AS `name`,`id` FROM `[[+PREFIX]]site_content` WHERE `published` = 1 AND `deleted` = 0 AND `context_key` = '[[+context_key]]'
```

When I open a resource that contains this template variable, the following error is logged in the MODX error log:

```php
[2022-06-02 16:15:22] (ERROR @/core/src/Revolution/modTemplateVar.php : 1016) modTemplateVar::parseBinding - third parameter is invalid JSON :@SELECT CONCAT(`pagetitle`, ' (', `id`, ')') AS `name`,`id` FROM `[[+PREFIX]]site_content` WHERE `published` = 1 AND `deleted` = 0 AND `context_key` = '[[+context_key]]'
```

Looking at the code it expects to contain a JSON string but it's not. I think the parseBinding method requires some additional logic maybe?

```php
public function parseBinding($binding_string)
    {
        $match = [];
        $match2 = [];
        $binding_string = trim($binding_string);
        $regexp = '/@(' . implode('|', $this->bindings) . ')\s*(.*)/is'; /* Split binding on whitespace */

        if (preg_match($regexp, $binding_string, $match)) {
            /* We can't return the match array directly because the first element is the whole string */

            $regexp2 = '/(\S+)\s+(.+)/is'; /* Split binding on second whitespace to get properties */

            $properties = [];
            if (preg_match($regexp2, $match[2] , $match2)) {
                if (isset($match2[2])) {
                    $props = json_decode($match2[2],true);
                    $valid = json_last_error() === JSON_ERROR_NONE;
                    if ($valid && is_array($props)){
                        $properties = $props;
                        $match[2] = $match2[1];
                    } else {
                        $this->xpdo->log(modX::LOG_LEVEL_ERROR, 'modTemplateVar::parseBinding - third parameter is invalid JSON :' . $binding_string);
                    }
                }
            }

            $binding_array = [
                strtoupper($match[1]),
                trim($match[2]),
                $properties
            ]; /* Make command uppercase */

            return $binding_array;
        }
    }
```

### Step to reproduce

1. Create a TV with input option values:

```php
@SELECT CONCAT(`pagetitle`, ' (', `id`, ')') AS `name`,`id` FROM `[[+PREFIX]]site_content` WHERE `published` = 1 AND `deleted` = 0 AND `context_key` = '[[+context_key]]'
```

2. Open resource
3. Open MODX error log

### Observed behavior
An error is written in the error log which I don't think is okay.

### Expected behavior
I'd expect this to not trigger any errors, because it is a valid/working value.

### Environment
MODX 3.0.1-pl
PHP 8.1

halftrainedharry:
### What does it do?
PR #15488 added a way to add properties to bindings in the form `@SNIPPET someSnippet {"someKey":"someValue"}` by looking for a value after **the second whitespace** and trying to parse this value as a JSON string.

Because every `@SELECT` binding contains a lot of whitespace an no JSON, it should be excluded from the check.

### Why is it needed?
To avoid constant (misleading) error log messages for every `@SELECT` binding.

### How to test
Create a TV with a `@SELECT` binding in the input option values and verify, that there are no more "third parameter is invalid JSON ..." error log messages.

### Related issue(s)/PR(s)
Resolves #16200

### DIFF
--- a/core/src/Revolution/modTemplateVar.php
+++ b/core/src/Revolution/modTemplateVar.php
@@ -1003,7 +1003,7 @@ class modTemplateVar extends modElement
             $regexp2 = '/(\S+)\s+(.+)/is'; /* Split binding on second whitespace to get properties */
 
             $properties = [];
-            if (strtoupper($match[1]) != 'SELECT' && preg_match($regexp2, $match[2] , $match2)) {
+            if (strtoupper($match[1]) != 'SELECT' && preg_match($regexp2, $match[2], $match2)) {
                 if (isset($match2[2])) {
                     $props = json_decode($match2[2],true);
                     $valid = json_last_error() === JSON_ERROR_NONE;
--- a/core/src/Revolution/modTemplateVar.php
+++ b/core/src/Revolution/modTemplateVar.php
@@ -1003,7 +1003,7 @@ class modTemplateVar extends modElement
             $regexp2 = '/(\S+)\s+(.+)/is'; /* Split binding on second whitespace to get properties */
 
             $properties = [];
-            if (preg_match($regexp2, $match[2] , $match2)) {
+            if (strtoupper($match[1]) != 'SELECT' && preg_match($regexp2, $match[2] , $match2)) {
                 if (isset($match2[2])) {
                     $props = json_decode($match2[2],true);
                     $valid = json_last_error() === JSON_ERROR_NONE;


JoshuaLuckers:
@sdrenth can you help us test this PR?

Mark-H (approved):

rthrash:

https://community.modx.com/t/modx-3-0-1-error-if-list-createdby-via-select/5981/2

Issue #16198, opened 2022-06-01
Conditional output filter fails if a colon is used inside `:then=` or `:else=`


superspace:
### Summary
Conditional output filter fails if a colon is used inside `:then=` or `:else=` on Linux Systems

### Step to reproduce
A conditional output modifier:
```
[[*id:eq=`1`:then=`Resource ID: 1`:else=`Resource ID: not 1`]]
```

### Observed behavior
The value of [[*id]] is displayed

If id is 1:
```
1
```
If id is not 1, the actual ID is displayed

There are some more observations: if there is no colon used in `then` and id is not 1, `else` is displayed correctly even with colon used inside. Unless you chain another modifier, then it will also fail.

### Expected behavior

If id is 1:
```
Resource ID: 1 
```
If id is not 1
```
Resource ID: not 1
```

### Environment
MODX 3.0.1
PHP 7.4.29 / 8.1 (FPM)
Apache 2.4
MariaDB 10.3.32
Linux 3.10.0-1160.49.1.el7.x86_64

The Bug doesn't show on MODX 2.8.* and so far only on Linux Systems. On local MAMP installation on MacOS everything works fine.

halftrainedharry:
This is a known issue related to this line in the code:

https://github.com/modxcms/revolution/blob/cb6f56a19af818ad4d1d62dc0d31b9819aaa876f/core/src/Revolution/Filters/modInputFilter.php#L56

The regular expression was changed in #14458 to allow nested output filters, but causes a lot of problems.

Same issue in the MODX forum: https://community.modx.com/t/output-modifier-issue-in-modx-3-0-1-pl/5376
Related issues: #16044, #16186

---

This particular problem can probably be fixed by adding a backtick to the first character set (``[^:=`]`` instead of `[^:=]`), but that doesn't solve all the other issues with this new regex pattern.

jmacdonald-spectacle:
My workaround is to use the html entity ``&colon;`` instead.

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/hiding-fields-does-not-work-correctly/5976/6

Mark-H:
Should be fixed with #16288.

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/a-colon-in-a-resources-content-is-breaking-a-template/6127/4

PR #16288, opened 2022-10-19
Refactor input filter parsing to properly support nested tags

Mark-H:
### What does it do?

Refactors the parsing in the modInputFilter, which gets passed the inside of a tag to break up into a filter command and it's modifiers (aka options). Adds more tests. 

Rather than relying on a very daunting regex that is clearly not handling nesting well, or reverting back to behavior that still has known issues with nested parsing, this refactor just changes the way that thing works entirely. 

It now parses the inside of a tag one character at a time to really understand what the implications of a certain nested tag are. It implements the parsing rules so that a `` ` `` inside a nested tag doesn't "close" the outer tag. It understands to not do anything with a nested tag within its own set of brackets. 

If there are other situations where this still doesn't address a nested tag yet, let me know.

### Why is it needed?

To fix the shitshow that has been recent parsing issues in 3.0.1 without reintroducing other limitations.  

### How to test

The expanded test suite should be looking green, at least it is for me locally 🤞 

I've also tried throwing some more complex ContentBlocks-powered 3-level-nested chunks against it, and it seems to be holding up where it would previously (even in 2.x) fail.

Getting even more tests in here would be preferred, as well as some real-world tests, would be ideal. 

### Related issue(s)/PR(s)

Related to / fixes:

- #13250: one of the earliest reports of issues with nested tags, with lots of discussions.
- #14095: report of nested conditionals causing broken tags
- #16254: proposed revert of #14458 which introduced issues with nested output filters (` characters)
- #16186: cut-off with nested snippet
- #16198: `:` inside an output filter breaking
- Passes Joshua's [test addition](https://github.com/JoshuaLuckers/revolution/commit/004b877ad0e597e77b19d4a6a09477495bdde4ad) and [tests added in #14458](https://github.com/modxcms/revolution/pull/14458/files) too

Does not yet seem to fix:

- #16044: the `;` seems to be tripping something up somewhere else. (`:` works fine.)

### DIFF
--- a/_build/test/Tests/Model/Filters/modInputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modInputFilterTest.php
@@ -91,6 +91,38 @@ class modInputFilterTest extends MODxTestCase
                     'title',
                 ]
             ],
+            [
+                'echoinput:input=`[[++mail_smtp_hosts]] [[echoinput:eq=`0`:then=`test1:[[++mail_smtp_hosts]];`:else=`test2:[[++mail_smtp_hosts]]`?input=`0`]]`',
+                'echoinput',
+                [
+                    'input',
+                ],
+                [
+                    '[[++mail_smtp_hosts]] [[echoinput:eq=`0`:then=`test1:[[++mail_smtp_hosts]];`:else=`test2:[[++mail_smtp_hosts]]`?input=`0`]]',
+                ]
+            ],
+            [
+                'content:notempty=`[[!Wayfinder? &startId=`0` &level=`1`]]`',
+                'content',
+                [
+                    'notempty',
+                ],
+                [
+                    '[[!Wayfinder? &startId=`0` &level=`1`]]',
+                ]
+            ],
+            [
+                'content:up:trim',
+                'content',
+                [
+                    'up',
+                    'trim',
+                ],
+                [
+                    '',
+                    '',
+                ]
+            ]
         ];
     }
 }
--- a/_build/test/Tests/Model/Filters/modInputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modInputFilterTest.php
@@ -68,7 +68,29 @@ class modInputFilterTest extends MODxTestCase
                     'foo',
                     'bar',
                 ]
-            ]
+            ],
+            [
+                'content:notempty=`[[!Wayfinder? &startId=`0` &level=`1`]]`',
+                'content',
+                [
+                    'notempty',
+                ],
+                [
+                    '[[!Wayfinder? &startId=`0` &level=`1`]]',
+                ]
+            ],
+            [
+                'pagetitle:default=`[[*longtitle]]`:toPlaceholder=`title`',
+                'pagetitle',
+                [
+                    'default',
+                    'toPlaceholder',
+                ],
+                [
+                    '[[*longtitle]]',
+                    'title',
+                ]
+            ],
         ];
     }
 }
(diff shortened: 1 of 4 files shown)


smg6511 (commented):
I've taken an initial run-through testing against all of the referenced issues and running the unit tests and, with the exception of the issue I mentioned on Slack re formatting (returns and other space) throwing everything off, all works well.

The last-referenced issue that remains unsolved (#16044) is not caused within the `modInputFilter` class—I confirmed that `process()` returns what it should from there. Perhaps that's a separate issue.

I have a suggestion for you to solve the formatting issue. Because its code is beyond what can be marked here, I'm just posting a chunk here for you to consider:
```php
public function filter(&$element)
    {
        /* split commands and modifiers and store them as properties for the output filtering */
        $output = $element->get('name');
        $initialSplitPos = strpos($output, ':');
        $hasFilters = $initialSplitPos !== false && $initialSplitPos > 0;

        // Normalize filter format
        if ($hasFilters) {
            // Collapse filter to one line
            $output = preg_replace('/[\n\r]\s*/m', '', $output);
            // Remove space around = and :
            $search = ['/\s*=\s*(?=`)/', '/\s*:\s*(?=[^=`\s]+\s*=\s*`)/'];
            $replace = ['=', ':'];
            $output = preg_replace($search, $replace, $output);
        }

        $name = $output;

        if ($hasFilters) {
            $splitPos = strpos($output, ':');
            $name = trim(substr($output, 0, $splitPos));
            $modifiers = substr($output, $splitPos);

            $chars = mb_str_split($modifiers);
```
UPDATE: @Mark-H's latest commit solves the formatting (spacing) issues in a simpler way, thus the above suggestion is obsolete.

rthrash:

https://community.modx.com/t/migx-php-8-0-nested-getimage-calls-is-it-a-bug/5750/13

rthrash:

https://community.modx.com/t/modx3-broken-output-filter-modifiers/5906/3

smg6511 (approved):
I tested a number of scenarios up to three levels deep—including calling a snippet within a nested conditional, calling a snippet with an input modifier within, using the makeUrl tag syntax within, adding a TV within, throwing various spacing around `:` and `=`—and didn't detect any issues.

I'll keep poking at it while others review, but it's looking good to me!

rthrash:

https://community.modx.com/t/modx3-broken-output-filter-modifiers/5906/4

Mark-H:
Based on the feedback and my own testing plus the unit tests, I'm starting to get pretty confident about this and would propose it get merged and released as nightly. Obviously if there is any review feedback on the code that's more than welcome.  



There are still some cases where the parser gets tangled up inside itself, however those trace back to `modParser::collectElementTags` which is not touched here, so I don't personally consider those blockers. 

For example, this minimal example extracted from a pretty nightmarish generated content caused by using `@CHUNK` on every level in a complex ContentBlocks repeater, breaks when `[[~3]]` appears twice. Changing one of the instances to a different ID avoids it, so this seems to be a replacement issue of sorts. 

```
[[$testTemplateWrapper?
  &template=`[[$testTemplate?
    &rows.0.richtext.value=`<p>Test <a href="[[~3]]">testhyperlink</a></p>`
    &richtext=`[[$testTemplateRichtext?
      &value=`<p>Test <a href="[[~3]]">testhyperlink</a></p>`
    ]]`
  ]]`
]]
```

(Each of the 3 chunks basically only have the relevant placeholder `[[+template]]`/`[[+richtext]]`/`[[+value]]` inside them, though the real world usage may do more useful things. Personally I'd recommend `@FILE` for this type of content when the goal is just to have static file templates, as those are parsed on save/rebuild so has less of a performance impact, and parsing issues don't happen.)

halftrainedharry:
> so this seems to be a replacement issue of sorts.

This is probably the same issue as with #16044.
Here is a possible solution: https://github.com/modxcms/revolution/issues/16044#issuecomment-1287056970

Mark-H:
Ah, I had missed that. Thanks for the bump! That does sound like it's the same problem and a potential fix!

rthrash:

https://community.modx.com/t/hiding-fields-does-not-work-correctly/5976/2

PR #16289, opened 2022-10-31
Catch dashboard widget errors on a per-widget basis

Mark-H:
### What does it do?
Currently, if an error happens inside any dashboard widget, the entire page shows an error:
 
Schermafbeelding 2022-10-31 om 21 31 42

That's a step up from 2.x where an error on the dashboard makes the entire manager unusable, but still, not great. 

With this patch, errors are caught when rendering each of the widgets and the error is placed inside the widgets own frame and the rest of the manager remains unaffected:

Schermafbeelding 2022-10-31 om 21 33 14

### Why is it needed?

This change makes it clearer in what widget an error occurred, while still providing the same information to help with debugging. And any other widgets on the dashboard remain functional. 

### How to test

Add a broken dashboard widget. What I've done for testing is call an undefined function `bug()` inside `manager/controllers/default/dashboard/widget.grid-online.php` but there are many ways to cause an error, and `throw new Exception('Invalid condition');` for throwing an exception.

### Related issue(s)/PR(s)
N/a

### DIFF
--- a/core/src/Revolution/modDashboardWidgetInterface.php
+++ b/core/src/Revolution/modDashboardWidgetInterface.php
@@ -11,6 +11,7 @@
 namespace MODX\Revolution;
 
 
+use Throwable;
 use xPDO\xPDO;
 
 /**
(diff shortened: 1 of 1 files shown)


halftrainedharry:
Your commit still contains your testing code (`bug();throw new Exception('Invalid condition');`).

opengeek:
> Your commit still contains your testing code (`bug();throw new Exception('Invalid condition');`).

@Mark-H — does this need to be resolved before merging?

Mark-H:
Whoops, definitely didn't mean to have that there. And missed your reply until now. 

Fixed!

opengeek (approved):

Issue #15918, opened 2021-11-24
PHP8 Fatal error when using Snippet Widget in Dashboard


Peeet93:
### Summary
If you use "Inline-PHP-Widget" then create_function is called in core/model/modx/moddashboardwidget.class.php on line 321.

### Step to reproduce
Create Widget. Use "inline-php-widget". Use widget in dashboard. Go to dashboard. See the error.

### Environment
Modx 2.8.3

JoshuaLuckers:
[create_function](https://www.php.net/manual/en/function.create-function.php) has been DEPRECATED as of PHP 7.2.0, and REMOVED as of PHP 8.0.0. Relying on this function is highly discouraged.

JoshuaLuckers:
Suggestion from @Mark-H on how to fix this:

> Write to an include file like snippets, or deprecate and remove

matdave:
Just a bump on this. Still an issue a year later.

matdave:
Created a PR that runs script treating it as a psuedo snippet.

smg6511:
This appears to have been addressed with the above referenced PR #16290.

PR #16290, opened 2022-10-31
Remove create_function from moddashboardwidget

matdave:
### What does it do?
Remove create_function from the PHP dashboard widget processor, and run the content as a snippet. 

### Why is it needed?
create_function is removed in PHP 8.0

### How to test
Create a PHP Dashboard widget in MODX running on PHP 8+
e.g. 
```
<?php 
echo "Hello world!"
```

### Related issue(s)/PR(s)
#15918

### DIFF
--- a/core/model/modx/moddashboardwidget.class.php
+++ b/core/model/modx/moddashboardwidget.class.php
@@ -317,10 +317,17 @@ abstract class modDashboardWidgetInterface {
     public function renderAsSnippet($content = '') {
         if (empty($content)) $content = $this->widget->get('content');
         $content = str_replace(array('<?php','?>'),'',$content);
-        $snippet = $this->modx->newObject('modSnippet');
-        $snippet->set('content', $content);
-        return $snippet->process(array(
+        $closure = function ($scriptProperties) use ($content) {
+            global $modx;
+            if (is_array($scriptProperties)) {
+                extract($scriptProperties, EXTR_SKIP);
+            }
+
+            return eval($content);
+        };
+
+        return $closure([
             'controller' => $this->controller,
-        ));
+        ]);
     }
 }
--- a/core/model/modx/moddashboardwidget.class.php
+++ b/core/model/modx/moddashboardwidget.class.php
@@ -326,8 +326,8 @@ abstract class modDashboardWidgetInterface {
             return eval($content);
         };
 
-        return $closure([
+        return $closure(array(
             'controller' => $this->controller,
-        ]);
+        ));
     }
 }
--- a/core/model/modx/moddashboardwidget.class.php
+++ b/core/model/modx/moddashboardwidget.class.php
@@ -317,8 +317,9 @@ abstract class modDashboardWidgetInterface {
     public function renderAsSnippet($content = '') {
         if (empty($content)) $content = $this->widget->get('content');
         $content = str_replace(array('<?php','?>'),'',$content);
-        $closure = create_function('$scriptProperties','global $modx;if (is_array($scriptProperties)) {extract($scriptProperties, EXTR_SKIP);}'.$content);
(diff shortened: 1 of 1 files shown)


Mark-H:
It's worth noting this is the fix for this in 3.0:

```
        $closure = function ($scriptProperties) use ($content) {
            global $modx;
            if (is_array($scriptProperties)) {
                extract($scriptProperties, EXTR_SKIP);
            }

            return eval($content);
        };

        return $closure([
            'controller' => $this->controller,
        ]);
```

Yours is better though as it avoids using `eval()` as well. 

Only concern I have is that the generated script might not have a valid filename because the snippet has no ID? Maybe it works fine as the default ID 0, though.

matdave:
@Mark-H I tested it on a site with quite a few custom dashboard widgets all loading at once and it seems to work fine. :fist_right: :wood:

JoshuaLuckers:
> Only concern I have is that the generated script might not have a valid filename because the snippet has no ID? Maybe it works fine as the default ID 0, though.

I share a similar concern, when `process` is called it will get the include filename for the script, generating it if it does not exist: https://github.com/modxcms/revolution/blob/8e400704988184f6e396cf19d9047b6a34ad217e/core/src/Revolution/modScript.php#L77

The name of the script/file will use the ID (in this case 0): https://github.com/modxcms/revolution/blob/8e400704988184f6e396cf19d9047b6a34ad217e/core/src/Revolution/modScript.php#L147

I'm not totally sure but later on the cacheManager is used to create a file: https://github.com/modxcms/revolution/blob/8e400704988184f6e396cf19d9047b6a34ad217e/core/src/Revolution/modScript.php#L193

If you have multiple dashboards and they all have a `modSnippet` with id `0` this might cause some conflicts if they try to read/write to the same file at the same time.

matdave:
> If you have multiple dashboards and they all have a `modSnippet` with id `0` this might cause some conflicts if they try to read/write to the same file at the same time.


It overwrites with each call, so it never double responds with the same widget if that's what your concern is.

core/model/modx/moddashboardwidget.class.php
```diff
@@ -317,8 +317,9 @@ public function getFileChunk($tpl,array $placeholders = array()) {
     public function renderAsSnippet($content = '') {
         if (empty($content)) $content = $this->widget->get('content');
         $content = str_replace(array('<?php','?>'),'',$content);
-        $closure = create_function('$scriptProperties','global $modx;if (is_array($scriptProperties)) {extract($scriptProperties, EXTR_SKIP);}'.$content);
-        return $closure(array(
+        $snippet = $this->modx->newObject('modSnippet');
```
opengeek: I think it would be good to set the name to something like `md5($this->widget->get('content'))` or similar.

opengeek (changes requested):

matdave:
I'm not sure why the GH Compare got so screwed up, but I adjusted Line 317 onward to match the solution for [rendering files as snippets in MODX 3](https://github.com/modxcms/revolution/blob/f2bdbec64b10933146712d21fc8819f43ef334a2/core/src/Revolution/modDashboardWidgetInterface.php#L157).

opengeek:
> I'm not sure why the GH Compare got so screwed up, but I adjusted Line 317 onward to match the solution for [rendering files as snippets in MODX 3](https://github.com/modxcms/revolution/blob/f2bdbec64b10933146712d21fc8819f43ef334a2/core/src/Revolution/modDashboardWidgetInterface.php#L157).

You changed the line-endings.

opengeek (approved):

rthrash:

https://community.modx.com/t/custom-dashboard-widget-issue-after-upgrade-php8/6254/2

PR #16291, opened 2022-10-31
Prevent errors or incompatibilities in extras from breaking the setup

Mark-H:
### What does it do?

Avoid instantiating namespaces, extension packages, and the OnMODXInit event while in the setup.

### Why is it needed?

The upgrade from 2.x to 3.0 isn't always super smooth due to breaking changes. Some extras have different versions for different major releases, or other errors pop up. 

Especially extras with extension packages or those with the new `bootstrap.php` that runs from the namespace root can inadvertently be run in the setup, and cause the setup itself to throw an error. 

### How to test

I've not yet tested this extensively personally, though a quick test on my development install completed as expected. If a nightly would get built with this patch included, that would help test it for a client that's currently having this issue (S34224). 

The issue mostly seems to happen on a 2.x > 3.0 upgrade but could theoretically also occur on regular patch updates.

Basic testing requires [building a core package](https://docs.modx.com/current/en/contribute/code/tooling/core) and then running the setup. 

### Related issue(s)/PR(s)

It comes up every now and then, but I don't think this has a specific issue.

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -564,9 +564,11 @@ class modX extends xPDO {
 
             $this->getCacheManager();
             $this->getConfig();
-            $this->_initNamespaces();
-            $this->_initContext($contextKey, false, $options);
-            $this->_loadExtensionPackages($options);
+            if (!$this->getOption(xPDO::OPT_SETUP)) {
+                $this->_initNamespaces();
+                $this->_initContext($contextKey, false, $options);
+                $this->_loadExtensionPackages($options);
+            }
             $this->_initSession($options);
             $this->_initErrorHandler($options);
             $this->_initHttpClient();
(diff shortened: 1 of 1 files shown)


opengeek:
Aren't those namespaces going to need to be loaded in some cases to complete the setup successfully? 🤔 What if you have custom resource classes, for instance, and an upgrade affects those somehow?

Mark-H:
I guess that _could_ become an issue when a core upgrade to the resources table needs to iterate over all resources as xPDO objects, but is that something we'd do or allow?  

A quick search in the current migrations shows only things like this:

```
$modx->updateCollection($class, ['class_key' => modDocument::class], ['class_key' => 'modDocument']);
```

That executes a standard UPDATE query to update the table en-masse. Schema changes and such would not be loading individual resources either, those just use the xPDOManager add/alterField/Index and such. 

I think this change is safe unless proven otherwise :D Perhaps there is a exotic plugin somewhere that does intend to operate during the setup, but I haven't seen one.

opengeek:
Plugins should definitely not execute during setup. I guess we'll give it all a try and see what happens. 🤷🏻

rthrash:

https://community.modx.com/t/database-error-using-the-upgrademodx-package/6125/11

PR #16292, opened 2022-11-02
Determine upload path based on the tree node's path attribute

Mark-H:
### What does it do?

Change the path determination for uploads. Before, it was recursively going up the tree from the selected node and adding the **display text** for each level. 

Now, it's relying on using the **path** that is provided by the media source for each directory (i.e. `modMediaSource::getContainerList()`) which allows more control. 

### Why is it needed?

I'm currently working on a new media source where internal **paths** are actually IDs that do not match the rendered file **names/text**. Support for such sources isn't exactly straightforward, but it's almost feature complete and ready to go.  

One of the challenges was accepting uploads into a subdirectory, because for some reason this code (originating from the very first MODX commit from 2009) was relying on the **text**. I suppose there may have once upon a blue moon been an edge case that led to this complication, but I cannot see anything not handled by the media sources' path.

### How to test

Assets rebuild is required for this patch. If you have `compress_js` disabled you can test right away.

Make sure that uploading to different paths continues works. 

I've personally tested uploads to the root and 2 or 3 levels deep in the following media sources: standard local filesystem from the root, local from /assets/,  AWS S3 (core), and my new one that is not yet released. They all seem to work as expected.

### Related issue(s)/PR(s)
N/a.


manager/assets/modext/widgets/system/modx.tree.directory.js
```diff
@@ -365,27 +365,10 @@ Ext.extend(MODx.tree.Directory,MODx.tree.Tree,{
...
-
-            // path for root node is it's path attribute
-            else {
-                path = node.attributes.path || '';
-            }
+    ,getPath: function(node) {
+        let path = '';
+        if (node !== undefined && node.attributes) {
+            path = node.attributes.path || '';
         }
```
GulomovCreative:
```suggestion
        let path = node?.attributes?.path || '';
```

PR #16293, opened 2022-11-04
Improve grid error handling when processor didn't return json

Mark-H:
### What does it do?

Title.

### Why is it needed?

If a processor does not return valid JSON, but for example an uncaught PHP error, right now that causes a client-side error when attempting to parse the response. This is the infamous `Uncaught SyntaxError: expected expression, got '<'` error. 

This PR makes sure to not break the client-side if the response is not valid JSON, and actually shows the error in the grid contents. This makes getting to the actual root error a lot easier!

For example, I had some misconfigurations that caused a VersionX grid to fail to load. Rather than an empty grid and JS error, with this PR applied that now looks like this:

Schermafbeelding 2022-11-04 om 20 28 06

Easy to debug, easy to report. 

### How to test

Break a getlist processor, for example call a non-existent function or break a class name, and test the grid.

Note that an assets rebuild is required if compress_js is enabled.   

### Related issue(s)/PR(s)

N/a

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -173,11 +173,16 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
     windows: {}
 
     ,onStoreException: function(dp,type,act,opt,resp){
-        var r = Ext.decode(resp.responseText);
-        if (r.message) {
-            this.getView().emptyText = r.message;
-            this.getView().refresh(false);
+        let r;
+        try {
+            r = Ext.decode(resp.responseText);
+        } catch (e) {
+            this.getView().emptyText = 'Invalid response from server: <pre><code>' + MODx.util.safeHtml(resp.responseText) + '</code></pre>';
         }
+        if (r && r.message) {
+            this.getView().emptyText = r.message;
+        }
+        this.getView().refresh(false);
     }
     ,saveRecord: function(e) {
         e.record.data.menu = null;


smg6511:
@Mark-H - Something to consider:
- ~Are you meaning to only handle this for grids with in-grid editing or all grids? If the latter, you might want to handle this in the base grid class.~ (Nevermind on this - I misread the code at first.)
- To me, assuming many of our users are of the non tech-oriented variety, a more user-friendly flow might be to issue a simple message stating there was an internal error and output the stack trace via `console.error`. That way you're not throwing that big ugly trace at typical users. I'd argue it's even more friendly for us techie types too to not see that degree of info output directly to the UI.

JoshuaLuckers:
I love this idea however I agree with @smg6511  about using `console.error`.

Mark-H:
I kinda think we shouldn't hide error details in the console. Processors can of course still return a clean error and put details in the MODX, this is just for fatal issues.

JoshuaLuckers:
> I kinda think we shouldn't hide error details in the console. Processors can of course still return a clean error and put details in the MODX, this is just for fatal issues.

Wouldn’t it break the UI if the error would be thrown at another moment in time? E.g when rendering the resource tree. 
It is it for grids only?

Mark-H:
Yes this is specific to grids. Title or changed file didn't give that away? :P 

I'm sure trees could use improved error handling, too, though. This is not my first improved error handling PR and it won't be the last.

Jako:
I prefer a direct Ext.msg and and/or a smaller clickable error message in the grid that shows an Ext.msg.

Mark-H:
If we're gonna bike shed this much over rendering a previously hidden error message, I look forward to seeing other people's preferred solutions to review instead.

JoshuaLuckers:
Personally I'm ok with this solution, it's better than we have now. I apologize for nitpicking, I would be frustrated as well if something similar happend to my PR.

smg6511:
@Mark-H - I'm gonna take you up on providing an alternate solution and was curious how you triggered the error for your example. For some reason I'm finding it tricky to force an uncaught error that finds its way to the grid component in the way that you did. I can trip things up and get a general error, but not one with a stack trace returned in the response.

Mark-H:
This specific case was some misconfiguration but just cause a type error or call an undefined method or something in the processor. And to have it show in the grid is what the PR does, before it was hidden in the ajax response only.

smg6511:
So, here's the thing: Using the Contexts grid as the test, I inserted this line in the `prepareRow` method to break the `GetList` processor:
`$contextArray = $this->noExist($contextArray);`

This indeed does the job, but the stack trace only appears in the php logs and does not get output into the ajax response, which is: 
```javascript
{
    "tId": 3,
    "status": 500,
    "statusText": "Internal Server Error",
    "responseText": "",
    "responseXML": null
}
```
The trace in the log shows:
```
[13-Dec-2022 12:41:29 America/New_York] PHP Fatal error:  Uncaught Error: Call to undefined method MODX\Revolution\Processors\Context\GetList::noExist() in /Users/jimg/GitHub/modx/revolution/core/src/Revolution/Processors/Context/GetList.php:124
Stack trace:
#0 /Users/jimg/GitHub/modx/revolution/core/src/Revolution/Processors/Model/GetListProcessor.php(101): MODX\Revolution\Processors\Context\GetList->prepareRow(Object(MODX\Revolution\mysql\modContext))
#1 /Users/jimg/GitHub/modx/revolution/core/src/Revolution/Processors/Model/GetListProcessor.php(69): MODX\Revolution\Processors\Model\GetListProcessor->iterate(Array)
#2 /Users/jimg/GitHub/modx/revolution/core/src/Revolution/Processors/Processor.php(189): MODX\Revolution\Processors\Model\GetListProcessor->process()
#3 /Users/jimg/GitHub/modx/revolution/core/src/Revolution/modX.php(1775): MODX\Revolution\Processors\Processor->run()
#4 /Users/jimg/GitHub/modx/revolution/core/src/Revolution/modConnectorResponse.php(151): MODX\Revolution\modX->runProcessor('Context/GetList', Array, Array)
#5 /Users/jimg/GitHub/modx/revol in /Users/jimg/GitHub/modx/revolution/core/src/Revolution/Processors/Context/GetList.php on line 124
```
This was the case for a couple other places I threw that fake method into.

Mark-H:
I assume you're aware of the display_errors php.ini setting.

Mark-H:
Was thinking last night that while rendering an invalid response is still useful, as extjs just shits the bed when that happens otherwise, we could also wrap a try/catch around running processors in runProcessoe or the request handler to catch any throwable and make that return a proper JSON response instead. That would also work when display_errors is off.

smg6511:
> I assume you're aware of the display_errors php.ini setting.

Ahhh, thank u - that explains it. I was assuming you were getting the trace returned with the typical production setting of display_errors = off.

PR #16294, opened 2022-11-06
modElement: Fix getTag() side effects in MODX 3

### LINKED ISSUE #16283: MODX 3: Object in snippet properties gets converted to array
### Summary
In some cases an object that is passed by reference as a property to a snippet, gets converted to an array.

### Step to reproduce
Here is some example snippet code to showcase the issue:

```
<?php
$output = '';

$snippet = $modx->getObject('MODX\\Revolution\\modSnippet', ['name' => 'someSnippet']);
$user = $modx->getObject('MODX\\Revolution\\modUser', 1);
$properties = ['fields' => ['user' => &$user] ];
//$properties = ['user' => &$user]; //this works correctly

$output = "Type before: " . get_class($user);

$success = $snippet->process($properties);

$output .= " | Type after: " . (is_array($user) ? "ARRAY!" : get_class($user));
return $output;
```

Code of the snippet "someSnippet" (just to show that the variable `$user` isn't even used):
```
<?php
return "";
```

### Observed behavior

The output of the snippet is `Type before: MODX\Revolution\mysql\modUser | Type after: ARRAY! `. 
The variable `$user` gets converted to an array.

In MODX 2.x this is not the case.

This behavior creates a problem with custom hooks in the _Login_ extra.
https://github.com/modxcms/Login/issues/180

### Environment
MODX 3.0.1

halftrainedharry:
### What does it do?
Creates the tag without accidentally changing the type of xPDOObjects in the properties.

### Why is it needed?
Currently in some cases an xPDOObject that is passed by reference as a property to a snippet, gets converted to an array.

### How to test

Run a snippet that has a property that is a _reference to an xPDOObject_ inside an array. Make sure the reference is of the same type after executing the snippet.

```php
<?php
$output = '';

$snippet = $modx->getObject('MODX\\Revolution\\modSnippet', ['name' => 'someSnippet']);
$user = $modx->getObject('MODX\\Revolution\\modUser', 1);
$properties = ['fields' => ['user' => &$user] ];

$output = "Type before: " . get_class($user);

$success = $snippet->process($properties);

$output .= " | Type after: " . (is_array($user) ? "ARRAY!" : get_class($user));
return $output;
```

### Related issue(s)/PR(s)
Resolves #16283

### DIFF
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -278,9 +278,9 @@ class modElement extends modAccessibleSimpleObject
                     } else {
                         if (is_array($value)) {
                             $func = function ($item) use (&$func) {
-                                if (is_array($item)){
+                                if (is_array($item)) {
                                     return array_map($func, $item);
-                                } else if ($item instanceof \xPDOObject) {
+                                } elseif ($item instanceof \xPDOObject) {
                                     return $item->toArray('', false, true);
                                 } else {
                                     return $item;
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -277,11 +277,16 @@ class modElement extends modAccessibleSimpleObject
                         $propTemp[$key] = $key . '=`' . $value . '`';
                     } else {
                         if (is_array($value)) {
-                            array_walk_recursive($value, function (&$item, $key) {
-                                if ($item instanceof \xPDOObject) {
-                                    $item = $item->toArray('', false, true);
+                            $func = function ($item) use (&$func) {
+                                if (is_array($item)){
+                                    return array_map($func, $item);
+                                } else if ($item instanceof \xPDOObject) {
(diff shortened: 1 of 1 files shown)


opengeek:
I thought that xPDOObjects were intentionally being changed to arrays in element properties. 🤔 I'll have to go back through the commit history to recall when/why…

JoshuaLuckers:
@opengeek yes, see https://github.com/modxcms/revolution/pull/15809

halftrainedharry:
This PR should still do what #15809 does, just without creating new problems (like https://github.com/modxcms/Login/issues/180).
Or am I missing something here?

opengeek:
No, I've got it now. I missed the part where just the "reference" being passed was getting converted to an array improperly while the values in the properties were still getting converted. I was simply confused because it looked like it was still doing the same thing. 😂

PR #16303, opened 2022-11-22
Fix TypeError exceptions when restoring deleted resources

### LINKED ISSUE #16297: Trash restore processor causes uninitialised/null exception
### Summary
When you try to "Restore all" in the trash manager (tested on a PHP8.1 instance), this will fail.

### Step to reproduce
See above.

### Observed behavior
Nothing is restored, strange error dialog pops up in the trash manager: 
image

Logfile shows: 
```bash
ERROR:  Uncaught Exception TypeError: "count(): Argument #1 ($value) must be of type Countable|array, null given" at /home/www/core/src/Revolution/Processors/Resource/Trash/Restore.php line 132
```

### Expected behavior
The resource(s) should be restored. If that failed due to whatever, at least a better error message should be shown. 

### Root cause
One line of code is missing in the restore processor, which must initialize the `$failures` array (around line 79):
```php
   public function process()
    {
        $this->success = [];
        $this->failures = [];    // <------- that's missing
        $contexts = [];

        foreach ($this->resources as $resource) {
```

### Environment
MODX version: 3.0.1-pl advanced,
PHP 8.1.11

JoshuaLuckers:
### What does it do?
Initialize array class properties as an empty array and improved some other pieces of code.

### Why is it needed?
To prevent TypeErrors exceptions. 

### How to test
Use "Restore all" in the trash manager.

### Related issue(s)/PR(s)
Resolves #16297

### DIFF
--- a/core/src/Revolution/Processors/Resource/Trash/Restore.php
+++ b/core/src/Revolution/Processors/Resource/Trash/Restore.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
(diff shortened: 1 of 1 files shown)


Mark-H:
To fix that phpcs failure, I think you need to move the `]` to the next line (with the `));`) It's not something that's wrong in your change, but would still prevents merging the PR.

Mark-H (approved):

Ibochkarev (approved):

PR #16311, opened 2022-11-28
Include the media source id in MODX media browser data [ReUp]

muzzwood:
### What does it do?
Includes the media source id in the "onSelect" callback data. 

### Why is it needed?
Currently the media source id is missing from that data. It would be useful for custom handling.

### How to test
Check that the correct "source" is now in the data array when listening for the select event.

e.g.
```
let browser = MODx.load({
    xtype: 'modx-browser',
    multiple: true,
    listeners: {
        select: {fn: function(data) {
            console.log(data.source);
    }, scope:this}
...
```

### DIFF
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -864,13 +864,15 @@ Ext.extend(MODx.browser.Window,Ext.Window,{
     }
 
     ,onSelect: function(data) {
-        var selNode = this.view.getSelectedNodes()[0];
-        var callback = this.config.onSelect || this.onSelectHandler;
-        var lookup = this.view.lookup;
-        var scope = this.config.scope;
+        var selNode = this.view.getSelectedNodes()[0],
+            callback = this.config.onSelect || this.onSelectHandler,
+            lookup = this.view.lookup,
+            scope = this.config.scope,
+            source = data.source;
         this.hide(this.config.animEl || null,function(){
             if(selNode && callback){
                 var data = lookup[selNode.id];
+                data.source = source;
                 Ext.callback(callback,scope || this,[data]);
                 this.fireEvent('select',data);
             }
@@ -1596,12 +1598,14 @@ Ext.extend(MODx.browser.RTE,Ext.Viewport,{
     }
 
     ,onSelect: function(data) {
-        var selNode = this.view.getSelectedNodes()[0];
-        var callback = this.config.onSelect || this.onSelectHandler;
-        var lookup = this.view.lookup;
-        var scope = this.config.scope;
+        var selNode = this.view.getSelectedNodes()[0],
+            callback = this.config.onSelect || this.onSelectHandler,
+            lookup = this.view.lookup,
+            scope = this.config.scope,
+            source = data.source;
         if (callback) {
             data = (selNode) ? lookup[selNode.id] : null;
+            data.source = source;
             Ext.callback(callback, scope || this, [data]);
             this.fireEvent('select', data);
             if (window.top.opener) {
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -868,7 +868,7 @@ Ext.extend(MODx.browser.Window,Ext.Window,{
             callback = this.config.onSelect || this.onSelectHandler,
             lookup = this.view.lookup,
             scope = this.config.scope,
-            source = data.source;
+            source = this.config.source;
         this.hide(this.config.animEl || null,function(){
             if(selNode && callback){
                 var data = lookup[selNode.id];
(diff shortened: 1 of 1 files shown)


muzzwood:
Actually that's not reliable. I'll come back with a different approach.

Jako:
I don't get the reason for this patch. The scope of the callback is `this` and so you can access the source inside with `this.config.scope`. Or am I wrong with that?

Jako (changes requested):
I don't get the reason for this patch. The scope of the callback is `this` in all cases in MODx.browser and so you can access the source inside with `this.config.source`. Or am I wrong with that?

smg6511:
@muzzwood - It would be helpful to know from where and how you are accessing the select callback as, like @Jako points out, the source may also be accessible via `this.config`. I don't think there's anything wrong with explicitly adding the source to the callback's data, but it might not be needed.

muzzwood:
Thanks for the reviews! IIRC it wasn't available in the callback but I could have missed it. 
I'll double check and report back after the holidays.

muzzwood:
Happy New Year! :)
I went and had another look. 
You're right @Jako, if the scope isn't specified/overridden on the listener then the scope is the browser and it's available via `this.config.source`.

However, if a different scope is being set on the listener (like it the code I was working with), then it's not available. e.g.

```
            listeners: {
                select: {fn: function(file) {
                    // Not available here as scope is set to outer component
                }, scope: this}
            },
```

```
            listeners: {
                select: {fn: function(file) {
                    // Available here because the scope is the browser
                }}
            },
```

While a developer could just refactor code to make sure the callback scope is always the browser object, I reckon it would still be handy to have the source available via the passed callback object too. e.g. `file.source` in the above example.

...

Perhaps a better way could be to pass the entire browser scope back as part of the callback data. So the source id could then be accessed as `file.browser.source`.

Mark-H:
I'd say in most cases the scope will in fact be overwritten, because you're asking the browser to select an image - you'll be wanting to handle that image in your own code. This gets my approval, but then again, I'm biased having asked Murray to work on this in the first place. 😛 

Just realized this will conflict with #16348 however I'm happy to adjust that PR if needed.

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -864,13 +864,15 @@ Ext.extend(MODx.browser.Window,Ext.Window,{
...
-        var scope = this.config.scope;
+        var selNode = this.view.getSelectedNodes()[0],
+            callback = this.config.onSelect || this.onSelectHandler,
+            lookup = this.view.lookup,
+            scope = this.config.scope,
+            source = this.config.source;
         this.hide(this.config.animEl || null,function(){
             if(selNode && callback){
                 var data = lookup[selNode.id];
+                data.source = source;
```
smg6511, line 875: Aside from some minor formatting fixes, just a few substantive changes here:

1. With ES6+, avoid `var` declaration unless there's a specific reason to use it. Use `const`, or when a variable will later be assigned a new value `let`.
2. Added the default source, as the source will not show up in `this.config.source` until it is _changed_ in the tree.
3. Cast the source to Integer when a value is present.

```suggestion
        const selNode = this.view.getSelectedNodes()[0],
              callback = this.config.onSelect || this.onSelectHandler,
              lookup = this.view.lookup,
              scope = this.config.scope,
              source = this.config.source || MODx.config.default_media_source
        ;
        this.hide(this.config.animEl || null, function() {
            if (callback){
                const data = lookup[selNode.id];
                data.source = source ? parseInt(source) : undefined ;
```

manager/assets/modext/widgets/media/modx.browser.js
```diff
@@ -1596,12 +1598,14 @@ Ext.extend(MODx.browser.RTE,Ext.Viewport,{
...
-        var lookup = this.view.lookup;
-        var scope = this.config.scope;
+        var selNode = this.view.getSelectedNodes()[0],
+            callback = this.config.onSelect || this.onSelectHandler,
+            lookup = this.view.lookup,
+            scope = this.config.scope,
+            source = this.config.source;
         if (callback) {
             data = (selNode) ? lookup[selNode.id] : null;
+            data.source = source;
```
smg6511, line 1608: Same moves as above here plus simplified the reassignment of `data` to match above (there's no need to test for truthiness of `selNode`, as a selection had to have been made for `onSelect` to be called).
```suggestion
        const selNode = this.view.getSelectedNodes()[0],
              callback = this.config.onSelect || this.onSelectHandler,
              lookup = this.view.lookup,
              scope = this.config.scope,
              source = this.config.source || MODx.config.default_media_source
        ;
        if (callback) {
            data = lookup[selNode.id];
            data.source = source ? parseInt(source) : undefined ;
```

smg6511 (changes requested):
Mostly fine-tuning here. The one really important change is the addition of default source in both `onSelect` methods:
```javascript
source = this.config.source || MODx.config.default_media_source
```

Mark-H:
Closing as this change was also included in #16348 which is now merged.

PR #16319, opened 2022-12-02
Fix Validation Error When Switching TV Types

### LINKED ISSUE #16314: Unable to save TV when it is created if you accidentally selected a type with a required field
### Summary
For example, I need to create a TV with **textarea** input type.
textarea input type at the very bottom of the list with input types, and if I clicked/loaded another input type (accidentally or not) where there is a required field (for example, for "Listbox (Multi-Select)" is "Dropdown List Options"), then my TV with textarea input type cannot be saved - an error appears.

tv-save-bug

Also, the tab with the error does not activate, although I think @smg6511 fixed it.

### Environment
MODX 3.0.3-dev

smg6511:
### What does it do?
Explicitly sets allowBlank for `modx-tv-elements` (Input Options) when switching between TV types that hide that field and those that don’t (in which case it’s a required field).

### Why is it needed?
Even when hiding the referenced field, its other parameters (namely the allowBlank one) still take effect; this renders the TV unsavable if a user happens to switch from a type that requires Options to one that doesn't.

### How to test
1. Begin the process of creating a TV and choose a type that requires Options (_e.g._, listbox, radio, etc.) but do not fill in those options. ( Remember to fill in the Name under General Information ;-) )
2. Try to save; you should receive a validation error.
3. Switch the type to one without Options (_e.g._, text, email, hidden, etc.) and attempt to save; the TV should save without error.

### Related issue(s)/PR(s)
Resolves #16314

### DIFF
--- a/manager/assets/modext/widgets/element/modx.panel.tv.js
+++ b/manager/assets/modext/widgets/element/modx.panel.tv.js
@@ -1388,6 +1388,7 @@ Ext.extend(MODx.panel.TVInputProperties,MODx.Panel,{
 
         if(inputOptValsItem){
             if(hideInputOptValsItemFor.indexOf(type) !== -1){
+                inputOptValsItem.allowBlank = true;
                 inputOptValsItem.hide().nextSibling().hide();
             } else {
                 this.updateSharedComponent(type, inputOptValsItem, inputOptValsItemVal);
@@ -1399,6 +1400,9 @@ Ext.extend(MODx.panel.TVInputProperties,MODx.Panel,{
                 */
                 if (typeChanged) {
                     Ext.getCmp('modx-tv-elements').clearInvalid();
+                    if(this.isNativeType){
+                        inputOptValsItem.allowBlank = false;
+                    }
                 }
             }
         }


Mark-H (approved):

manager/assets/modext/widgets/element/modx.panel.tv.js
```diff
@@ -1399,6 +1400,9 @@ Ext.extend(MODx.panel.TVInputProperties,MODx.Panel,{
                 */
                 if (typeChanged) {
                     Ext.getCmp('modx-tv-elements').clearInvalid();
+                    if(this.isNativeType){
+                        inputOptValsItem.allowBlank = false;
+                    }
```
JoshuaLuckers, line 1405: Out of curiosity:

What happens if you change the type, this statement is triggered because it's a nativeType and you switch the type again to a type that's not a native type? Will it switch back to `inputOptValsItem.allowBlank = true` somewhere?

smg6511: The `allowBlank` property is only automatically set internally for native types, so there should be no problem with non-native ones. I believe I checked that when testing, but will double-check to be sure.

smg6511: Yup - confirmed - everything works as expected when switching from native to non-native types.

Ruslan-Aleev (approved):
Visually everything works as it should :)

Mark-H (approved):

PR #16321, opened 2022-12-03
JavaScript linter: step one

mishanthrop:
### What does it do?
JavaScript linter installed.

### Why is it needed?

### How to test
1. `npm i`
1. `npm run js:lint`

image

At the moment there are many issues from the linter.
In this PR, I just installed the linter and made its initial configuration. Next, I plan to correct the linter's remarks and tweak the rules.


JoshuaLuckers:
Very nice! I’ll test it tomorrow!

smg6511:
Thanks @mishantrop, this is no doubt a step we need to take! I've actually been working on this as well, and have eslint running in my local modx dev workspace to develop the rule set I think we should have. 

On this PR, I have a couple questions/suggestions:
- Is there an advantage to setting up the rc file as a js module instead of a straight JSON config?
- Can we set this up such that the configs (`eslintrc.*` and `package.json`) are placed in the root of the workspace instead of the default template (i.e., matching how `phpcs.xml` and `composer.json` for php are placed in the root)? This may make it easier to work with from the standpoint of your local IDE picking up the config more easily and integrating with the git workflow.
- I'd like to see us base our rules off of the airbnb-base ruleset.
- Before ultimately implementing the linter, we should allow time for interested parties to suggest their rule preferences and come to agreement on those. Apart from new rules, we're going to want to avoid changing rules that get finalized and rolled out unless there's really good reason to do so. (I'm just suggesting we be thoughtful about this and not rush it into action.)

Lastly, if possible, you should remove the lock file from your commit.

mishanthrop:
@smg6511

1. In theory, there shouldn't be any difference.
2. I did not dare to take out all the files to the root, but I like this idea. We need to know the opinion of the leaders. I didn't plan on making any drastic changes, but it really does make things easier in the IDE.
3. As I know, airbnb-config is React-specific.
4. Simple rules usually are autofixable (mostly a matter of taste: trailing comma, indents, semicolon, etc). Therefore, there is no need to sweat too much, changing the new rules should not greatly interfere with the development process.

`> Lastly, if possible, you should remove the lock file from your commit.`
Why? Is it generated in the CI?

smg6511:
> 3\. As I know, airbnb-config is React-specific

The one I'm suggesting (`eslint-config-airbnb-base`) is for non-React projects. It's `eslint-config-airbnb` that includes the React rules and plugins.

Mark-H:
I don't have any strong opinions on the javascript code style standard to use (I kinda thought prettier was the defacto standard on many big projects? I could be very wrong.), but happy to lend this my support once there's agreement.

smg6511:
> I kinda thought prettier was the defacto standard on many big projects

Prettier is only a code formatter, not a linter. You integrate the formatter with the linter, but I think the choice of formatter is more of and end-user choice and not something you'd enforce in our dev dependencies.

@Mark-H @JoshuaLuckers - Do you guys see any problem with dropping this in the project root? I think it's most appropriate there as, just like the php linting config, it should apply project-wide (minus select areas).

Mark-H:
Fair enough. 

Is there a reason we need a non-standard eslint rule set? Just checked a recent project and that's using `eslint:recommended` and some vue-specific stuff. (`extends: ["plugin:vue/vue3-essential", "eslint:recommended", "@vue/eslint-config-prettier"]`. The prettier config seems twice as popular than airbnb's base config based on weekly npm downloads, but both are pretty big. 

I don't know enough about either to fully appreciate what the impact of either will be on our project, but I do know this is a decision we'll have to live with for the next decade as we wont be changing code styles constantly so we do need to discuss options.

--

IMO moving the templates directory seems out of scope here. Where files are located isn't relevant to linting/formatting. 

The assets are admittedly somewhat hidden, but there's an old reason it's a "default" directory in a "templates" directory. Let's set up the linting correctly for manual running first (code standard included), work on getting all existing code compliant (something we need to do for php too, been chatting with Jason about that), and if there's a strong reason to move it deal with that separately.

smg6511:
> IMO moving the templates directory seems out of scope here. Where files are located isn't relevant to linting/formatting.

I'm not wanting to move the templates ... just suggesting the linting config and related package.json is placed in the root.

Mark-H:
`/_build/templates/default/package.json` is specific to the theme so I'd leave that in place as-is. I would be fine however introducing a separate `/package.json` in the root that provides the js linting across the code base.

smg6511:
Yep, that's exactly what I had in mind ;-)

mishanthrop:
Whats new:
1. Moved package.json and .eslintrc.js to the project root
1. Installed rules by airbnb (`eslint-config-airbnb-base`)

Reminder:
This PR is just the first step in setting up linting. It makes no sense to set up the linter perfectly, it can be done later without degrading the development process.

Current linting state:

image

mishanthrop:
@smg6511 

`> Is there an advantage to setting up the rc file as a js module instead of a straight JSON config?`

It is possible to use comments in js file, while rc is a pure JSON.

JoshuaLuckers (approved):

Mark-H (approved):

.eslintrc.js
```diff
@@ -0,0 +1,41 @@
...
+        indent: 0,
+        quotes: ['error', 'single'],
+        semi: 0,
+        'space-before-function-paren': 0,
+        'comma-dangle': 0,
+        'prefer-arrow-callback': 0,
+        'space-before-blocks': 0,
+        'object-shorthand': 0,
+    },
+}
```
smg6511, line 41: To keep it simple, I've selected the whole rc, as one of the changes I'd like to see is the non-use of the comma-dangle (which would have entailed a bunch of one-line suggestions). The real primary request here is the rules set I'm proposing, as well as _not_ ignoring the workspace panel file. Otherwise, this is great!
```suggestion
module.exports = {
    env: {
        browser: true,
        es2021: true
    },
    extends: [
        'eslint:recommended',
        'airbnb-base'
    ],
    globals: {
        MODx: 'readonly',
        Ext: 'readonly',
        _: 'readonly'
    },
    ignorePatterns: [
        'manager/assets/ext3/**/*.js',
        'manager/assets/fileapi/**/*.js',
        'manager/assets/lib/**/*.js',
        'manager/assets/modext/modx.jsgrps-min.js',
        'setup/assets/js/ext-core.js',
        'setup/assets/js/ext-core-debug.js'
    ],
    overrides: [],
    parserOptions: {
        ecmaVersion: 'latest'
    },
    rules: {
        'arrow-parens': ['error', 'as-needed'],
        'comma-dangle': ['error', 'never'],
        'consistent-return': 0,
        curly: ['error', 'all'],
        eqeqeq: ['error', 'smart'],
        'func-names': ['warn', 'as-needed'],
        indent: ['error', 4, {
            VariableDeclarator: 'first',
            SwitchCase: 1
        }],
        'max-len': ['warn', {
            code: 140,
            ignoreComments: true
        }],
        'no-continue': 'warn',
        'no-new': 'warn',
        'no-param-reassign': 'warn',
        'no-plusplus': ['error', {
            allowForLoopAfterthoughts: true
        }],
        'no-underscore-dangle': 'warn',
        'no-unused-vars': ['error', { args: 'none' }],
        'no-use-before-define': ['error', 'nofunc'],
        'object-shorthand': ['error', 'consistent'],
        'one-var': ['error', 'consecutive'],
        'prefer-arrow-callback': 'warn',
        'prefer-rest-params': 'warn',
        'semi-style': ['warn', 'last'],
        'space-before-function-paren': ['error', 'never']
    }
};
```

smg6511 (changes requested):
Sorry to chime in late on this with an actual review. I've been working with this quite a bit with one of the larger Ext files to hone the ruleset I'd like to see. As @mishantrop suggests it's a starting point, but I wanted to develop it a bit further before implementation.

.eslintrc.js
mishanthrop: @smg6511 you can commit this changes/rules with fixed that big js file in next PR.
As has been said several times, this request is just starting point.

PR #16322, opened 2022-12-03
Fix non-existent snippet tags being broken up by an @ tag 

### LINKED ISSUE #16318: At sign (@) inside chunk parameter value nested into tag output filter breaks tag parsing
Consider this content:
```
[[+xyz:empty=`
aaa
[[xxx? &x=`bbb@ccc`]]
ddd
`]]
eee
```

Supposed output should be:
```
aaa xxxSnippetOutput ddd eee
```

Instead it is:
```
aaa [[xxx? &x=`bbb eee
```

This works fine as expected:
```
[[+xyz:empty=`
aaa
[[xxx? &x=`bbbccc`]]
ddd
`]]
eee
```

At sign (@) inside parameter value breaks parsing, but only when nested inside output filter parameter value. Workaround here is to put inner tag inside separate chunk.
With some more complicated scenarios this even causes Server error with `End of script output before headers: index.php` in error log.
Clean MODx 3.0.2 install, the same with older 2.7.3

Mark-H:
### What does it do?
Described in #16318, if a non-existent snippet tag is found within an output filter, an @ tag inside the filter can cause the output to break.

The provided example where `xxx` doesn't exist:

```
[[+xyz:empty=`
aaa
[[xxx? &x=`bbb@ccc`]]
ddd
`]]
eee
```

is getting broken up by modTag/modElement::getPropertySet on the `@`, treating the part in front of it as the name of the tag/element, including half the snippet tag.

This fix makes sure that it only considers the bit up to the first `:`. 

To ensure property sets continue to work, I wanted to add some property set tests but I can't get this structure:

```
[[!Snippet@PropSet:default=`foo`]]
``` 

to return anything when the snippet returns an empty value. If someone can figure out why that's not working (in either the test or real world cases!) that would be great. 

### Why is it needed?
Make parser more resilient.

### How to test
Unit tests or create your own similar test cases with snippets on a resource.

### Related issue(s)/PR(s)
Fixes #16318

### DIFF
--- a/_build/test/Tests/Model/modParserTest.php
+++ b/_build/test/Tests/Model/modParserTest.php
@@ -12,6 +12,9 @@
 namespace MODX\Revolution\Tests\Model;
 
 
+use MODX\Revolution\modElementPropertySet;
+use MODX\Revolution\modPropertySet;
+use MODX\Revolution\modSnippet;
 use MODX\Revolution\modX;
 use MODX\Revolution\MODxTestCase;
 use MODX\Revolution\MODxTestHarness;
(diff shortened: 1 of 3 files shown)


_build/test/Tests/Model/modParserTest.php
```diff
@@ -646,7 +649,172 @@ public function providerProcessElementTags() {
...
+                    'removeUnprocessed' => false,
+                    'prefix' => '[[',
+                    'suffix' => ']]',
+                    'tokens' => [],
+                    'depth' => 10
+                ]
+            ],
+            [
+                'This is a [[snippetName@propSetName:default=`default value`]]',
+                'This is a default value',
```
Mark-H, line 802: This is the test that's failing. The snippet returns the value of the `prop` property, which doesn't exist in this scenario, so the snippet returns an empty value. That should make it return the `default` filter, but instead it returns nothing at all (`This is a `). 

Can reproduce the same behavior when creating things manually on a template/resource, so that suggests _something_ is broken?

Mark-H: There is a chance this may've been related to #16337, so will need to test again.

theboxer:
@Mark-H is this still in a draft state?

Mark-H:
Yes, I was hoping for some help trying to figure out the failing test but didn't get any.

core/src/Revolution/modElement.php
```diff
@@ -726,19 +726,14 @@ public function getPropertySet($setName = null)
...
-                $this->set('name', $name);
-            }
+
+        $nameSplit = explode(':', $name);
+        $tagName = array_shift($nameSplit);
+        $remainingTag = implode(':', $nameSplit);
+        if (strpos($tagName, '@') !== false) {
+            $split = xPDO:: escSplit('@', $tagName);
+            $psName = $split[1];
+            $this->set('name', $split[0] . $remainingTag);
```
smg6511, line 736: Note that this is where the issue lies (same goes in the modTag file L520). Here, when modifiers are present, you end up with ``snippetNamedefault=`default val` `` instead of  ``snippetName:default=`default val` `` (note the colon missing in the former) when setting the name on the Element.

smg6511: A couple other things:
- I made my own submission for this (#16682) only because it was easier to do, particularly because I wanted to make sure whatever I came up with passed on the latest 3.1 dev branch _and_ I've been having issues getting the unit tests to run on my local dev install.
- I refactored how the two parts of the tag are broken up (using substr approach instead of explode/implode), as it seemed easier to see what was happening.

opengeek:
Issue resolved in #16682

Issue #16208, opened 2022-06-13
MagpieRSS::feed_start_element() error with PHP 8.0 and 8.1


andrewcsp:
### Summary
I've noticed the following error appears hundreds of times related to MagpieRSS when running PHP 8.0 or 8.1

(ERROR @ /home/user/core/model/modx/xmlrss/magpierss.class.php : 133) PHP warning: MagpieRSS::feed_start_element(): Argument #3 ($attrs) must be passed by reference, value given

### Step to reproduce
To reliably reproduce the Magpie errors, delete the contents of your cache folder (within core), then load the MODX manager homepage/dashboard (i.e. domain.com/manager/)

### Observed behavior
Review the errors in Manage > Reports > Error Log. You should see hundreds of the above MagpieRSS error (186 of them for me).

### Expected behavior
I would expect to see no MagpieRSS errors.

### Environment
MODX Revolution 2.8.4-pl
Apache/2.4.53 (cPanel)
10.3.35-MariaDB
Confirmed with PHP Version 8.0.19 and 8.1.6. The problem does not occur with PHP 7.4.29.

claytonk:
The feed_start_element() function does indeed show the third argument as being passed by reference. That function is on line 152:

`function feed_start_element($p, $element, &$attrs)`

It appears that removing the reference indicator ('&') will fix this issue, or at least it did in my case.

matdave:
I've got a PR in for this. Thanks for reporting!

PR #16325, opened 2022-12-06
MagpieRSS::feed_start_element() error with PHP 8.0 and 8.1

matdave:
### What does it do?
Adjust the type declaration for argument $attrs from a reference to a value on function feed_start_element

### Why is it needed?
PHP 8+ is much more strict about variable types. This resulted in an excess of warnings in the error log. 

### How to test
Open the dashboard in MODX 2.x and PHP 8.1, and then check for errors generated in the error.log. 

### Related issue(s)/PR(s)
Issue #16208

### DIFF
--- a/core/model/modx/xmlrss/magpierss.class.php
+++ b/core/model/modx/xmlrss/magpierss.class.php
@@ -149,7 +149,7 @@ class MagpieRSS {
         $this->normalize();
     }
 
-    function feed_start_element($p, $element, &$attrs) {
+    function feed_start_element($p, $element, $attrs) {
         $el = $element = strtolower($element);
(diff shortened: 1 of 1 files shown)


Mark-H (approved):

Mark-H:
You're killing it with the PRs lately, thanks @matdave 🔥

Ibochkarev (approved):

opengeek:
If the function no longer operates on the variable as a reference, how does it return the changed values where it is used? Are we sure this change does not affect behavior?

opengeek:
Looking closer, I don't see why this ever was set as a reference. Ignore my question.

PR #16331, opened 2022-12-15
Add new error reporting for grids

smg6511:
### What does it do?
Changed to base grids class `onStoreException` method to provide in-grid reporting of errors.

### Why is it needed?
Picking up where @Mark-H started, this displays an error message where previously a grid was mysteriously empty when a server error occurred. Note, however, this solution shows stack traces in the browser console by default and only outputs them to screen when MODx is in debug mode. Otherwise, an elegant message is shown in the grid, notifying users of an error and suggesting ways to view more detailed information.

Here are some screenshots of various scenarios when this solution is applied:

modx-pr-example-outputs--xl

### How to test
1. Clear all caches and run a rebuild of the default template.
2. Break a GetList processor. I used the Contexts grid for testing and added this line before the return in the `prepareRow` method: `$contextArray = $this->noExist($contextArray);`
3. Test the output in your test grid by changing your php `display_errors` and MODx `debug` system setting to various values. Note that stack traces will only appear when `display_errors` is set to "on" and `debug` is not "0". Verify receiving messages as shown in the example image above.
4. Note that I've found that the various caches (modx, browser, etc) can be finicky and it may take multiple clears when changing your config scenarios to get the expected output.

### Related issue(s)/PR(s)
Alternate solution to the withdrawn PR #16293.


Mark-H:
Why did you go through the extra effort to hide it unless people opt-in to see it through a system setting that enables so much more debugging? That'll make all PHP notices and warnings go into the error log, no? This is meant to handle errors not debug statements. 

The whole problem is that errors are hidden and people have to jump through hoops to see what actually went wrong. This still seems to hide them until jumping through the hoop to enable the debug setting, and then spamming the log with notices too. I feel like this is two steps forward and one step back. Still a step forward. But I'm kinda sad you spent _more time to take an unnecessary step back_. 

I'll approve it just because you've already spent more time on this than the problem is worth but I really want to encourage us to collectively focus our energy and time on getting the bigger picture things done right over the tiny details of something most users shouldn't see often (if ever).

Mark-H (approved):

smg6511:
OK, thanks for putting it through! I do disagree on the one step back opinion. Think of it this way: The only people who would be motivated to jump through that hoop of turning debugging on are the ones who actually have the expertise to solve it. IMO, it's your typical MODx end user you want to cater to first and that means giving them feedback that's useful but not needlessly intimidating. 

In a whole other context, think of other consumer software — say Photoshop. I know how to use it but I were thrown a stack trace for something that went wrong (and this is given that I'm at least comfortable with seeing code), I'd think less of the software. I'd get the impression that the developers dropped the ball on something or that the problem was more serious than it actually was. Of course, Adobe's stack traces are huge so they're a little intimidating no matter what ;-)

Anyway, the point of my blathering on is that what I saw you touching on in your original PR is actually a very important aspect that's not paid much attention in MODx and has always bothered me. So I look at what I was doing here as not only solving the small problem at hand, but perhaps beginning to model how we provide better error feedback.

core/lexicon/en/default.inc.php
```diff
@@ -143,6 +143,9 @@
 $_lang['error'] = 'Error';
 $_lang['error_sending_email'] = 'Error sending email';
 $_lang['error_sending_email_to'] = 'Error while sending mail to ';
+$_lang['error_grid_get_content_no_msg'] = 'A server error prevented this grid’s content from loading. To get additional information, change MODx’s debug system setting to either 1 or empty (not 0, as this will turn reporting off).';
```
JoshuaLuckers:
```suggestion
$_lang['error_grid_get_content_no_msg'] = 'A server error prevented this grid’s content from loading. To get additional information, change MODX’s debug system setting to either 1 or empty (not 0, as this will turn reporting off).';
```

Mark-H: Critical tweak.

Jako (approved):

Issue #14442, opened 2019-03-01
Revert lexicon entries in all namespaces


JonikUl:
### Summary
At the moment only the core lexicon entries could be reset with the button 'Revert All Core Entries' in the lexicon management. It would be nice to reset other lexicon namespaces too. 

### Current behaviour
- Change a value in a lexicon not in the core namespace.
- Use **Revert All Core Entries**  button.
- The same value remains.

### Improved behavior
- The button should be renamed to 'Revert the entries in the selected namespace' and the namespace should be send to https://github.com/modxcms/revolution/blob/2.x/core/model/modx/processors/workspace/lexicon/reloadfrombase.class.php#L26

### Environment
MODX 2.7.1

![lex](https://user-images.githubusercontent.com/18344088/53652132-8f965880-3c61-11e9-8f91-fce7d7fcbace.gif)

Jako:
The button works well for the core lexicon workspace on my local 2.7.1 installation. But not in other (selected) namespaces.

This is not a bug because of the text of the button. But the behaviour could be changed by sending the selected namespace to the processor and change the button text.

Jako:
I have changed the issue into a feature request. Is this ok?

Ruslan-Aleev:
Very strange button :) The button work only for "core", although the first one in the list of namespaces may be not "core", but, for example, "ace".
Today we noticed this bug, the problem is also in translation, for example, in Russian, there is no mention that the button only works for "core". We tested in Russian, and when we recorded gif we didn’t pay attention to the original name in English.

PR #16333, opened 2022-12-21
Add reloading of lexicon entries for any namespace

Ruslan-Aleev:
### What does it do?
Add reloading of lexicon entries for any namespace:

- Move the button up, it's was almost invisible.
- Added a pop-up window with confirmation, where the namespace, topic and language are indicated.
- The name of the strings is now returned to the console, and not just their amount.

lexicons_reload

p.s. In a good way, this PR should be improved and merged after this (https://github.com/modxcms/revolution/pull/15942), although the current version is also working.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14442


core/lexicon/en/lexicon.inc.php
```diff
@@ -25,7 +25,8 @@
...
 $_lang['lexicon_management'] = 'Lexicon Management';
 $_lang['lexicon_management_desc'] = 'Here you can override any Lexicon Entries across Namespaces and their respective Topics. Just double-click on any Entry\'s value to override it. To add new Lexicon Entries or Topics, just create their Topic files in their Namespace\'s respective directory.<br /><b>Each Entry will be available via the [[%name]] placeholder.</b>';
-$_lang['lexicon_rlfb_msg'] = 'Successfully regenerated [[+num]] strings.';
-$_lang['reload_from_base'] = 'Revert All Core Entries';
-$_lang['reload_success'] = 'Successfully reloaded [[+total]] strings.';
+$_lang['lexicon_revert'] = 'Revert Entries';
+$_lang['lexicon_revert_confirm'] = 'Are you sure you want to reloade this strings for "[[+namespace]]" namespase, "[[+topic]]" topic and "[[+language]]" language?';
+$_lang['lexicon_revert_success'] = 'Successfully reloaded [[+total]] strings with "[[+names]]" names.';
+$_lang['lexicon_revert_error'] = 'Strings for reloading not found.';
```
JoshuaLuckers: @rthrash It's a good idea if a native English speaker would check the lexicon's for grammar and typo's.

rthrash: Will do that ASAP

rthrash: Three minor changes (the first one is fine). 

```
$_lang['lexicon_revert_confirm'] = 'Are you sure you want to revert the "[[+namespace]]" namespace > "[[+topic]]" topic > "[[+language]]" lexicon to its default strings?';
 $_lang['lexicon_revert_success'] = 'Reverted [[+total]] total string(s): [[+names]]';
 $_lang['lexicon_revert_error'] = 'No lexicon entries found to revert.'
```

rthrash: @JoshuaLuckers @Ruslan-Aleev that was done earlier (see other comments)

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +41,55 @@ public function getLanguageTopics()
...
-        $language = 'en';
-        $namespace = 'core';
-        $corePath = $this->modx->getOption('core_path', null, MODX_CORE_PATH);
+        $namespace = htmlentities($_POST['namespace']);
+        $language = htmlentities($_POST['language']);
+        $topic = htmlentities($_POST['topicLexicon']);
+        if ($namespace == 'core') {
+            $namespacePath = $this->modx->getOption('core_path', null, MODX_CORE_PATH);
+        } else {
+            $namespacePath = $this->modx->getOption('core_path', null, MODX_CORE_PATH) . 'components/' . $namespace;
```
JoshuaLuckers: The path to a namespace should not be hardcode like this since it can be in a different path as well.

You have to ask this info from the right namespace object. See: https://github.com/modxcms/revolution/blob/1e7c048833a8419c113a0f10c0daa3bcad5b4d8e/core/src/Revolution/modLexicon.php#L363-L382

JoshuaLuckers (changes requested):

core/lexicon/en/lexicon.inc.php
```diff
@@ -25,7 +25,8 @@
...
 $_lang['lexicon_management'] = 'Lexicon Management';
 $_lang['lexicon_management_desc'] = 'Here you can override any Lexicon Entries across Namespaces and their respective Topics. Just double-click on any Entry\'s value to override it. To add new Lexicon Entries or Topics, just create their Topic files in their Namespace\'s respective directory.<br /><b>Each Entry will be available via the [[%name]] placeholder.</b>';
-$_lang['lexicon_rlfb_msg'] = 'Successfully regenerated [[+num]] strings.';
-$_lang['reload_from_base'] = 'Revert All Core Entries';
-$_lang['reload_success'] = 'Successfully reloaded [[+total]] strings.';
+$_lang['lexicon_revert'] = 'Revert Entries';
+$_lang['lexicon_revert_confirm'] = 'Are you sure you want to revert the "[[+namespace]]" namespace > "[[+topic]]" topic > "[[+language]]" lexicon to its default strings?';
+$_lang['lexicon_revert_success'] = 'Reverted [[+total]] total string(s): [[+names]]';
+$_lang['lexicon_revert_error'] = 'No lexicon entries found to revert.';
```
smg6511, line 31: The language used still is a bit clunky IMO and the term _strings_ is not going to be clear to most users unfamiliar with coding. Consider the following: 
```suggestion
$_lang['lexicon_revert_confirm'] = 'This action will permanently replace all overridden Lexicon entries in the selected namespace/topic/language with their original text. User-created entries will not be affected. Continue?';
$_lang['lexicon_revert_multiple_success'] = 'Successfully reverted [[+total]] overridden entries: [[+names]]';
$_lang['lexicon_revert_success'] = 'Successfully reverted one overridden entry: [[+names]]';
$_lang['lexicon_revert_error'] = 'No Lexicon entries found to revert.';
```
Note that the only change to the final entry is capitalization of “Lexicon.”

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +42,53 @@ public function getLanguageTopics()
...
                 $entry->remove();
             }
         }
 
-        $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('reload_success', ['total' => $i]));
+        if ($i) {
+            $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_success', ['total' => $i, 'names' => implode(', ', $names)]));
+        } else {
+            $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_error'));
+        }
```
smg6511, line 92: Consider this change to use more natural language for singular vs. multiple reversion (makes use of proposed new `lexicon_revert_multiple_success` entry):
```suggestion
        switch (true) {
            case $i === 1:
                $this->modx->log(
                    modX::LOG_LEVEL_WARN,
                    $this->modx->lexicon('lexicon_revert_success', ['names' => $names[0]])
                );
                break;
            case $i > 1:
                $this->modx->log(
                    modX::LOG_LEVEL_WARN,
                    $this->modx->lexicon('lexicon_revert_multiple_success', ['total' => $i, 'names' => implode(', ', $names)])
                );
                break;
            default:
                $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_error'));
        }
```

manager/assets/modext/workspace/lexicon/lexicon.grid.js
```diff
@@ -248,30 +247,46 @@ Ext.extend(MODx.grid.Lexicon,MODx.grid.Grid,{
...
-        this.console = MODx.load({
-           xtype: 'modx-console'
-           ,register: 'mgr'
-           ,topic: topic
-        });
+        var tb = this.getTopToolbar(),
+        namespace = tb.getComponent('namespace').getValue(),
+        topicLexicon = tb.getComponent('topic').getValue(),
+        language = tb.getComponent('language').getValue(),
+        topic = '/workspace/lexicon/reload/';
```
smg6511: I'm just encouraging that any time javascript is being updated, we look to bring it to current coding conventions (namely ES6+). Also, I'd like to see us replacing abbreviated variable names to descriptive ones to improve readability. Here, and in any other place where the variable will not be re-defined, use `const` in the declaration.
```suggestion
        const toolbar = this.getTopToolbar(),
              namespace = toolbar.getComponent('namespace').getValue(),
              topic= toolbar.getComponent('topic').getValue(),
              language = toolbar.getComponent('language').getValue(),
              registryTopic = '/workspace/lexicon/reload/'
        ;
```

manager/assets/modext/workspace/lexicon/lexicon.grid.js
```diff
@@ -248,30 +247,46 @@ Ext.extend(MODx.grid.Lexicon,MODx.grid.Grid,{
...
+                ,topicLexicon: topicLexicon
+                ,language: language
+            }
+            ,listeners: {
+                'success': {
+                    fn:function() {
+                        this.console = MODx.load({
+                            xtype: 'modx-console'
+                            ,register: 'mgr'
+                            ,topic: topic
```
smg6511: The following suggestion is much less than it looks like, mostly I'd like to encourage that as contributors make changes we apply better code style and updated syntax/grammar along the way; in this case:
- Move commas from beginning to end of statements
- Remove unneeded quotes from object keys (this will be most of the time, unless the key contains a character such as a dash and others that then require the key to be quoted)

I also suggest minor naming tweaks to the topic variables in an attempt to clarify them a little. The clash between the completely unrelated `topic` variables for Lexicons and the Registry threw me off at first.

```suggestion
                namespace: namespace,
                topic: topic,
                language: language
            }),
            url: this.config.url,
            params: {
                action: 'Workspace/Lexicon/ReloadFromBase',
                register: 'mgr',
                topic: registryTopic,
                namespace: namespace,
                lexiconTopic: topic,
                language: language
            },
            listeners: {
                success: {
                    fn: function() {
                        this.console = MODx.load({
                            xtype: 'modx-console',
                            register: 'mgr',
                            topic: registryTopic
```

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +42,53 @@ public function getLanguageTopics()
      */
     public function process()
     {
-        $language = 'en';
-        $namespace = 'core';
-        $corePath = $this->modx->getOption('core_path', null, MODX_CORE_PATH);
+        $namespace = htmlentities($_POST['namespace']);
+        $language = htmlentities($_POST['language']);
+        $topic = htmlentities($_POST['topicLexicon']);
```
smg6511: If taking my suggestion re variable renaming in the JS file:
```suggestion
        $topic = htmlentities($_POST['lexiconTopic']);
```

smg6511 (commented):
First off, nice job extending this to reversion of Lexicons beyond the core ones! Re my suggestions — they are mostly of the fine-tuning variety and this PR works as intended without them, so I intend to approve once you get a look. I am hoping you'll agree with most of what I proposed though ;-)

For the future: I do think it's a bit of a weird UI move to use the MODX console here. This has nothing to do with @Ruslan-Aleev's work (it was already that way). I think that should be reserved only for installation procedures and cache-clearing. At worst a regular Window showing success/fail feedback makes more sense; at best IMO we move toward a UI model where the user doesn't have to click or otherwise be interrupted by this feedback (such as a "Toast"-like message that is less obtrusive and disappears on its own). That's a job for another day.

Better-yet, for this area and others like it, a more elegant flow would report how many and which items are _about_ to be changed (and from "x" to "y") in the confirmation window and have the final notification window (or Toast) simply say that the action succeeded (or failed).

manager/assets/modext/workspace/lexicon/lexicon.grid.js
Mark-H: Changes like this make diffs on PRs very hard to read (tho github helps). It's a tricky problem and each approach has its own drawbacks, but I'm leaning towards doing one massive automated code style swoop at some point to give it all consistency.

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +42,53 @@ public function getLanguageTopics()
      */
     public function process()
     {
-        $language = 'en';
-        $namespace = 'core';
-        $corePath = $this->modx->getOption('core_path', null, MODX_CORE_PATH);
+        $namespace = htmlentities($_POST['namespace']);
```
Mark-H: Processors should NOT use $_POST (or other superglobals). Instead, use `$this->getProperty('namespace')`.

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
Mark-H: I find the original easier to read and understand than the switch in this case 😅

Mark-H (changes requested):

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
smg6511: Note that I changed to `switch` because I added another condition. It looks kinda ugly here because of the space constraints, but I think it reads well in the file. Still prefer if-elseif-else ?

manager/assets/modext/workspace/lexicon/lexicon.grid.js
JoshuaLuckers: Wasn't there a discussion in the core-contributors channel where we sort of decided/leaned towards against this?

core/lexicon/en/lexicon.inc.php
Ruslan-Aleev: 1) I'm still in favor of displaying both "[[+namespace]]", and "[[+topic]]", and "[[+language]]" in the message, because pop-up window may cover the selection in the grid and it will not be clear what is selected.
2) @rthrash @smg6511  Can you decide on the final spelling? So that the final version is as clear **or we can leave the current one**?

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
Ruslan-Aleev: I do not see much point in the extra `lexicon_revert_multiple_success` lexicon, and without it it is clear that the line / lines have been reloaded. And there will be no extra work for translations, and the code is simpler.

manager/assets/modext/workspace/lexicon/lexicon.grid.js
Ruslan-Aleev: Yes, such global changes should be made in another PR, let's leave it as it is for now.

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +42,53 @@ public function getLanguageTopics()
...
+
+        if (empty($namespace) || empty($topic) || empty($language)) {
+            $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_error'));
+            $this->modx->log(modX::LOG_LEVEL_INFO, 'COMPLETED');
+
+            return $this->success();
+        }
+
+        $namespaceObj = $this->modx->getObject(modNamespace::class, $namespace);
+        $namespacePath = $namespaceObj->getCorePath();
```
Mark-H, line 58: A null exception can occur here if the namespace doesn't exist or isn't specified. That's definitely an edge case but I'd prefer that to be caught by checking if $namespaceObj is in fact a modNamespace.

Ruslan-Aleev: And how can there be null or emptiness here, if I check this above, or have I misunderstood something?

Mark-H: You're checking if the namespace is empty but not if it is a valid namespace that exists and can be loaded. So if an invalid namespace is passed, `$this->modx->getObject()` will return null, and a fatal error will occur trying to call `getCorePath()` on null.

PR #16334, opened 2022-12-22
fixes on some bad translations in FR

intersel:
### What does it do?
* it fixes a placeholder name in 'upload.upload_progress' that was translated in french ('loaded' -> 'chargé') that breaks feedback during uploading files

Fixes on some translations:
* Fix translation on file_err_ext_not_allowed (remove an 's' in 'extension') 
* Fix on file_err_ext_not_allowed ('autorisée' should have an 'e' at the end) 
* Translate file_folder_remove_confirm
* Translate file_remove_confirm

### Why is it needed?
The dynamic progression bar when uploading a file shows "[+chargé+] / 1.29 Mib" instead of the dynamic value of loaded bytes

### How to test
* Set Modx in French
* upload a file
* the dynamic uploading progression bar would now correctly show how many bytes are uploaded...


Ruslan-Aleev:
Thank you for participating in MODX!
But translations should be edited here - https://crowdin.com/project/modx-revolution/fr , only the English version is edited on github, on the basis of which translations are created in crowdin.

intersel:
Ok thanks! Will do it there so...

JoshuaLuckers:
@insytes I accepted your request so you should have access!

intersel:
Thanks Joshua!

PR #16348, opened 2023-01-10
Support multi select in MODx.Browser when multiSelect: true is provided

Mark-H:
### What does it do?

Allows the media browser to be instantiated with `multiSelect: true` to indicate the caller supports multiple file selections. Selecting multiple files is done with ctrl/shift  

By default, this is off, which preserves the current behavior of only allowing a single image to be inserted. 

(*Note: _technically_ the current 3.0.2 behavior is to allow multiple _selections_, but only the _first_ selected image is ever returned to the caller, so that's kinda broken)

For ease of implementation, the provided `onSelect` callback is simply called repeatedly for each of the selected files rather than introducing a new mechanism. 

### Why is it needed?

Enable implementations to allow multiple file selections in a single action. In this specific case, I'm looking at MoreGallery and Redactor, but anything else that uses MODx.Browser and can support multiple selections will benefit from this. 

### How to test

Importing images in MoreGallery 1.16.0-pl supports multiple selections and specifies `multiSelect: true` to enable it, so that can be used to test multiple files are being passed to the callback. (* this release will be available from Jan 11th.)

Alternatively, running something like this in the console can be used to confirm multiple callbacks are run:

```
MODx.load({
    xtype: 'modx-browser',
    closeAction: 'close',
    id: Ext.id(),
    multiple: true, /* indicates multiple browsers may be used on the page, without this it wont show up a second time */
    multiSelect: true, /* allow multiple selection inside the browser */
    onSelect: function(data) {
         console.log('received onSelect callback', data);
    },
}).show();
```

Also change `multiSelect: true` to `multiSelect: false` to see it support just the one selection. 

### Related issue(s)/PR(s)

Surprisingly enough, I haven't been able to find an issue for this even though it's been a common request for users of my extras over the years.


smg6511:
@Mark-H - Just a couple things as I get ready to test:

- Maybe consider a new var name for the passed data in the try-catch block, e.g.:
   ```javascript
   const callbackData = lookup[selNode.id];
   Ext.callback(callback, scope || this, [callbackData]);
   this.fireEvent('select', callbackData);
   ```
- You probably will want to replicate what you did down in the `MODx.browser.RTE` definition (~L1600)

Mark-H:
I think this might be the first I've heard of MODX.browser.RTE! There don't seem to be any usage of that in core nor the obvious extras (tinymce, tinymce-rte). Do you have any idea where it's used (or why/how it's different)? 

It does seem to be duplicated in `manager/assets/modext/widgets/core/modx.rte.browser.js` as well.

Mark-H:
Updated to adjust all duplicates of this, and also to include the source and review comments from #16311.

smg6511:
> It does seem to be duplicated in `manager/assets/modext/widgets/core/modx.rte.browser.js` as well.

Yes, that file needs to be deleted (it is a duplicate and is not included by the template build or any of the php connectors). 

> There don't seem to be any usage of that in core nor the obvious extras (tinymce, tinymce-rte). Do you have any idea where it's used (or why/how it's different)?

The component is in fact used by TinyMCE Rich Text Editor (@Jako's newer version of the TInyMCE editor). I didn't test whether the other RTE extras used it. I saw some nuances to the RTE implementation (such as how the browser gets closed after selection) but didn't pick through it enough to note other differences.

One other thing to mention, and you can tell me if I was overthinking it (which I've been known to do), but I see the place and way we applied parseInt on the source value differs and I wanted to just explain the reasoning behind my version. I actually was going to do the same as you did but found there was one scenario where the result of the cast would be `NaN`; I felt it was cleaner to only pass an int or undefined in the callback data. Anyway, it may be all the same - up to you ;-)

Last, since you incorporated the feature of @muzzwood's PR, shall we have him close that one out to avoid confusion?

Mark-H:
> One other thing to mention, and you can tell me if I was overthinking it (which I've been known to do), but I see the place and way we applied parseInt on the source value differs and I wanted to just explain the reasoning behind my version. I actually was going to do the same as you did but found there was one scenario where the result of the cast would be NaN; I felt it was cleaner to only pass an int or undefined in the callback data. Anyway, it may be all the same - up to you ;-)

Sorry, I'm confused what you're referring to here. What different way of applying parseInt?

smg6511:
You've got:
```javascript
const [...],
source = parseInt(this.config.source || MODx.config.default_media_source);
```
then later:
```javascript
selData.source = source;
```
whereas, I'd suggested:
```javascript
const [...],
source = this.config.source || MODx.config.default_media_source;
```
then:
```javascript
selData.source = source ? parseInt(source) : undefined ;
```
Basically, only casting to int if source has a value.

Mark-H:
I'd argue it's a safe assumption `MODx.config.default_media_source` always has a value. If `this.config.source` does not have a truthy value (a `0`, `null`, `''`, etc), it will fallback to that. So that conditional is moot.

smg6511:
I do want to point out that you _can_ end up with a non-value for `MODx.config.default_media_source`, which may or may not be a bug. (It happens if you enter edit mode in the settings grid but don't choose a value.) That's the scenario where you end up with NaN. Just pointing it out ... not making a big deal over it.

smg6511 (approved):
Tested and works as expected.

JoshuaLuckers (approved):

PR #16349, opened 2023-01-13
Remove unnecessary settings for uploaded files

Ruslan-Aleev:
### What does it do?
Remove unnecessary settings for uploaded files. Basically, the file settings are set in the media source, so so many settings are not needed.

Previously, you had to add extensions in 3 places, in the media source settings, and in the `upload_files` and `upload_images/upload_media` system settings, while the extensions were duplicated. Now there are fewer settings, at least 2.

Perhaps in the future it is worth transferring them completely to the media source.

### Related issue(s)/PR(s)
N\A


Mark-H (changes requested):
I'd like this to make sure on upgrade that anything in the removed settings is copied into `upload_files`, other than that I'm very happy to see these go.

Ruslan-Aleev:
> I'd like this to make sure on upgrade that anything in the removed settings is copied into upload_files

Is it possible to upload files if they are not set in the `upload_files` setting? I don't think so.

setup/includes/upgrades/common/3.0.0-update-sys-setting_upload_files.php
```diff
@@ -0,0 +1,78 @@
...
+        $upload_images_arr = explode(',', $upload_images_value);
+    }
+
+    if ($upload_media_setting) {
+        $upload_media_value = $upload_media_setting->get('value');
+        $upload_media_arr = explode(',', $upload_media_value);
+    }
+
+    $upload_files_arr_new = array_unique(array_merge($upload_files_arr, $upload_images_arr, $upload_media_arr));
+    $upload_files_arr_new = array_filter($upload_files_arr_new, fn($value) => !is_null($value) && $value !== '');
```
Mark-H: [Arrow shorthand functions](https://www.php.net/manual/en/functions.arrow.php) were introduced in php 7.4, officially MODX3 still supports 7.2.

(Bumping that is a good topic to discuss elsewhere.)

Mark-H: It also looks like this particular callback could simply be removed as the default filter array_filter does will already filter out empty values.

Mark-H (commented):
Looking good minus the phpcs errors (red).

Ibochkarev:
> Looking good minus the phpcs errors (red).

Fixed

Mark-H (approved):

Ibochkarev (approved):

PR #16352, opened 2023-01-16
JavaScript linter: step two

mishanthrop:
### What does it do?
Continued implementation of the eslint: https://github.com/modxcms/revolution/pull/16321

I think its time to adjust the critical rules (unused variables, undefined variables, dangerous `var` -- the most popular problem types) and correct the relevant comments.

Rules from the category of "a matter of taste" are disabled in order to focus on more important issues.

I also suggest agreeing on the maximum number of changes (count of files/lines), since large Pull Requests are difficult to test.

### Why is it needed?

### How to test
```sh
npm i
```

```sh
npm run js:lint
```

Current state:
```sh
✖ 5221 problems (5114 errors, 107 warnings)
  3717 errors and 0 warnings potentially fixable with the `--fix` option.
```

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/16321


Ibochkarev (approved):

Ibochkarev:
🥳

manager/assets/modext/workspace/package.browser.panels.js
```diff
@@ -413,13 +413,15 @@ MODx.panel.PackageBrowserDetails = function(config) {
 				+'<tpl for=".">'
 					+'<div class="selected">'
 						+'<tpl if="screenshot">'
-							+'<a href="{screenshot}" title="'+_('package_preview_view')+'" alt="'+_('package_preview_view')+'" class="lightbox" />'
+							+'<a href="{screenshot}" title="package_preview_view" alt="package_preview_view" class="lightbox">'
```
Mark-H: This change is wrong; it removes the title/alt values which need to come from the lexicons.

setup/assets/js/sections/database.js
```diff
@@ -20,23 +20,23 @@ MODx.DB = function () {
           url: 'processors/connector.php',
           success: function (r) {
             r = Ext.decode(r.responseText);
-            var msg = Ext.select('#modx-db-step1-msg'),
-              dbInfo = Ext.select('#modx-db-info'),
-              step2 = Ext.select('#modx-db-step2');
+            const msg = Ext.select('#modx-db-step1-msg');
+            const dbInfo = Ext.select('#modx-db-info');
+            const step2 = Ext.select('#modx-db-step2');
```
Mark-H, line 25: Not trying to nitpick, but for this block, wouldn't the preferred code style be this?

```
const msg = Ext.select('#modx-db-step1-msg'),
    dbInfo = Ext.select('#modx-db-info'),
    step2 = Ext.select('#modx-db-step2');
```

smg6511:
Guys - I would prefer this be paused, as I have some strong opinions on the code style that I suggested in the "step one" iteration of this that went unheeded.

Also, the changes to files should wait until an agreed-upon ruleset is finalized. And those who are contributing the most to the front end should have the most influence on that. Also, It's hard enough getting reviews; the notion of applying a draft set of rules to files already doesn't seem like a good idea to me.

Mark-H:
> I also suggest agreeing on the maximum number of changes (count of files/lines), since large Pull Requests are difficult to test.

There's not a hard number, but the size of this PR currently is nice. It lets me scroll through to spot potential issues without wearing me down because it's too much. But other PRs with 300 changed lines could be much more work to review and test if they change logic - this PR doesn't actually change much. 

If we're 100% sure nothing but code style is tweaked I'd personally also be happy to run the autofixer on all code in one big swoop. 

> Guys - I would prefer this be paused, as I have some strong opinions on the code style that I suggested in the "step one" iteration of this that went unheeded.

It's still a draft, so go for it. ;)

smg6511:
@Mark-H - I'd hesitate on auto fixing, as not only do we want to fix simple things like spacing, comma placement, and the like, we also want to modernize the code (using contemporary syntax and methods, for example, preferring arrow functions where possible, replacing for-in loops with forEach where appropriate, etc.). Some of those modernizations are risky to do automatically. I'd also suggest strategizing the sequence of files to be updated (maybe individual panels and grids first, then the larger base and utility classes last) and perhaps just update one per PR, or at least make separate commits for each file changed.

setup/assets/js/sections/database.js
mishanthrop: I noticed that this approach to variable initialization is often used in the project.
This was OK when only `var` were used. But now the variables are initialized with `let` and `const`. Because of this, visually strange grouping can occur (fictional code example):

```javascript
const user = getUser(),
    ui = getUI()

let userId = getUserId(),
    output = ''
```

Therefore, it makes sense to group the initialization of variables by meaning, and not by type of variables:
```javascript
const user = getUser()
let userId = getUserId()

const ui = getUI()
let output = ''
```

manager/assets/modext/workspace/package.browser.panels.js
mishanthrop: Thanks, I didn't notice that the eslint underscore overlapped the function name. More precisely, I noticed this, but did not fix it in all my edits.

image

mishanthrop:
@smg6511 
`> I have some strong opinions on the code style that I suggested in the "step one" iteration of this that went unheeded.`

[That PR](https://github.com/modxcms/revolution/pull/16321) was about installing a linter, nothing more.
Also, your last [rule suggestions](https://github.com/modxcms/revolution/pull/16321#discussion_r1067170834) were mostly about the code style, not its quality in terms of reliability (indentation, semicolons, spaces, etc.)

I am grateful to you for the tips regarding the choice of the basic configuration. 

In this PR I want to deal with 3-4 rules, which I wrote about in the PR description.
We will have the opportunity to discuss further development: I wrote it in the first request and I write it here.

mishanthrop:
@smg6511 
`> It's hard enough getting reviews; the notion of applying a draft set of rules to files already doesn't seem like a good idea to me.`
Any arguments?

JoshuaLuckers:
If we target a major release (e.g 4.x) I wouldn't mind if we have risky rules that might break something. But for minor and patch releases (3.x) I would prefer a set of "code style" only rules. 

@smg6511 do you see anything risky with the current set of rules in this draft PR? 

Edit: we also need to make sure that changing method signatures, etc. etc. need to go into a new major/minor release since they are breaking changes.

mishanthrop:
@JoshuaLuckers 

`> if we have risky rules that might break something`
Any feature/bugfix might break something, this PR is not exception.

`> we also need to make sure that changing method signatures`
I removed unused parameters in internal calbacks only. These changes are not breaking any compatibility. 

Currently the only way to check for breaking changes is to open these files in IDE and installed root dependencies and see if the functions arguments are marked as unused.

`> But for minor and patch releases (3.x) I would prefer a set of "code style" only rules.`
Why?

Mark-H:
Looks like you did a merge of upstream, it's better to do a rebase (`git rebase -i upstream/3.x`) when trying to bring things up-to-date to avoid adding unrelated commits. If you need help sorting that out feel free to drop me a DM in the MODX community slack.

mishanthrop:
@Mark-H yep, git is my weak point. 

`> feel free to drop me a DM`
Thank you. Perhaps when we come to a compromise with smg6511, this time I'll just create a new PR. And next time I'll use your `git rebase` advice.

PR #16251, opened 2022-08-22
Fix and Enhance Usergroup ACL Permissions Grids and Filtering

smg6511:
### What does it do?
Made a number of changes and additions in the processors and grid JS files to correct issues with the grids’ grouping functionality and more accurately build the grid filters depending on the data shown. More specifically:

1. Replaced the column model renderers (used to render links on column data) with a `templatecolumn` specification (generated by a new method in the grids base class), as using renderers on grouping grids interferes with the grouping functionality (because a renderer is an interceptor method in Ext).
2. Moved menu specifications (grid context menus) out of processors and into the grid js for all sections (context was already this way). The primary reason being that some events (such as `afterRemoveRow`) were not firing when set up this way; monitoring these events is necessary for maintaining accurate filter options when making changes to a grid's data. Also, there's no reason to specify js in php unless the item in question will change on the fly and needs setup data from the processor.
3. Grid view, including the bottom toolbar, now correctly updates when a row is deleted.
4. Created new methods to update dependent filters (when two or more filters are present) based upon the selected value of any given filter.
5. Fixed some filters that were missing options due to an incorrect specification of their `baseParams.group` property.

### Why is it needed?
Grouping in these grids was not working correctly. Also, providing filter options where no relevant data is available in a given grid is a UX problem that needs to be fixed not only here, but across all grids with filters. 

### How to test
1. Ensure modx and browser caches are fully cleared.
2. Create a number of policies in each area (if you don't already have some set up) and play with the grouping and filtering of items in each grid. (These are the five Access Permissions grids — Context, Resource Groups, Element Categories, Media Sources, and Namespaces.) Verify that filters only show relevant options depending on the grid data and other filters you've made selections on.
3. Verify that other operations (editing and deleting) affect filter options as expected.

### Related issue(s)/PR(s)
Resolves #16239.

### Special Note
Because these updates involve a lot of changes, I didn't want to execute them for _all_ relevant grids at once. This PR completes a logical section and, depending on any requested changes, other PRs will follow what I've done here for other grids.

### Before and After Examples
Currently, in MODx 3.0.3-dev, note each of the following problems (in order of appearance in the video):

1. Each ACL area grid does not group properly; where an area's rows should be grouped under a common heading, they are not. To see this clearly, pause each time I click on each area and note where grouping should be occurring (for example, in Contexts, the three assigned the _Editors - 1000_ Role should be grouped under one row heading for that Role).
[...]

PR #16355, opened 2023-01-18
Fix and Enhance Usergroup ACL Permissions Grids and Filtering

### LINKED ISSUE #16239: Access Policy Context not selectable
### Summary
On an almost fresh MODX installation, the Context policy is not selectable on usergroup/update -> contexts. 

The connector of that combo uses the following properties:

start: 0
limit: 20
action: Security/Access/Policy/GetList
group: Administrator,Object
combo: true
query:

### Environment
MODX 3.0.1, PHP 7.4

smg6511:
After difficulties rebasing and resolving conflicts of the original PR (#16251) in preparation for final merge, elected to create this clean re-submission that resolves all issues.

### What does it do?
Made a number of changes and additions in the processors and grid JS files to correct issues with the grids’ grouping functionality and more accurately build the grid filters depending on the data shown. More specifically:

1. Replaced the column model renderers (used to render links on column data) with a `templatecolumn` specification (generated by a new method in the grids base class), as using renderers on grouping grids interferes with the grouping functionality (because a renderer is an interceptor method in Ext).
2. Moved menu specifications (grid context menus) out of processors and into the grid js for all sections (context was already this way). The primary reason being that some events (such as `afterRemoveRow`) were not firing when set up this way; monitoring these events is necessary for maintaining accurate filter options when making changes to a grid's data. Also, there's no reason to specify js in php unless the item in question will change on the fly and needs setup data from the processor.
3. Grid view, including the bottom toolbar, now correctly updates when a row is deleted.
4. Created new methods to update dependent filters (when two or more filters are present) based upon the selected value of any given filter.
5. Fixed some filters that were missing options due to an incorrect specification of their `baseParams.group` property.

### Why is it needed?
Grouping in these grids was not working correctly. Also, providing filter options where no relevant data is available in a given grid is a UX problem that needs to be fixed not only here, but across all grids with filters. 

### How to test
1. Ensure modx and browser caches are fully cleared.
2. Create a number of policies in each area (if you don't already have some set up) and play with the grouping and filtering of items in each grid. (These are the five Access Permissions grids — Context, Resource Groups, Element Categories, Media Sources, and Namespaces.) Verify that filters only show relevant options depending on the grid data and other filters you've made selections on.
3. Verify that other operations (editing and deleting) affect filter options as expected.

### Related issue(s)/PR(s)
Resolves #16239.

### Special Note
Because these updates involve a lot of changes, I didn't want to execute them for _all_ relevant grids at once. This PR completes a logical section and, depending on any requested changes, other PRs will follow what I've done here for other grids.

### Before and After Examples
Currently, in MODx 3.0.3-dev, note each of the following problems (in order of appearance in the video):

1. Each ACL area grid does not group properly; where an area's rows should be grouped under a common heading, they are not. To see this clearly, pause each time I click on each area and note where grouping should be occurring (for example, in Contexts, the three assigned the _Editors - 1000_ Role should be grouped under one row heading for that Role).
2. [0:19] Next, notice how the grid filters show irrelevant options where they should only show those present in the grid data. The _Filter by Context_ combo shows all Contexts instead of those present (dupctx1, dupctx2, dev2, and www2). The same goes for the _Filter by Policy_ combo. [0:42] Additionally, the Filter by Policy combo is mis-configured in places; here the problem in the Contexts area is illustrated. When adding a new Context, you are given the full Policies list in the Create window, but the filter combo shows an incomplete subset. [1:02] Lastly, when attempting to create additional policies for the same Context, an error occurs due to an incorrect validation for a unique context key. 
3. [1:32] When deleting a row, see how the count is not updated in the lower right grid status area.
4. [1:50] Next, when filtering, a choice in one combo should update the others to show only entries relevant to the current state of the query. They do not currently react to changes as one would expect.

https://user-images.githubusercontent.com/689075/209720912-944ca7bb-2d91-4aae-9bdd-f0b5b7d71b18.mov

Now, note how each problem is fixed (action taken in same order as above):

https://user-images.githubusercontent.com/689075/209720965-f3b53940-7554-4db1-a695-86bc2b25c8b7.mov

### DIFF
--- a/core/src/Revolution/Processors/Security/Access/UserGroup/Context/GetList.php
+++ b/core/src/Revolution/Processors/Security/Access/UserGroup/Context/GetList.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
@@ -67,7 +68,7 @@ class GetList extends GetListProcessor
         $userGroup = $this->getProperty('usergroup');
         $c->where([
             'principal_class' => modUserGroup::class,
-            'principal' => $userGroup,
+            'principal' => $userGroup
         ]);
         $context = $this->getProperty('context', false);
         if (!empty($context)) {
@@ -94,7 +95,6 @@ class GetList extends GetListProcessor
             'policy_name' => 'Policy.name',
             'policy_data' => 'Policy.data',
         ]);
-
         return $c;
     }
 
@@ -109,7 +109,10 @@ class GetList extends GetListProcessor
         if (empty($objectArray['name'])) {
             $objectArray['name'] = '(' . $this->modx->lexicon('none') . ')';
         }
-        $objectArray['authority_name'] = !empty($objectArray['role_name']) ? $objectArray['role_name'] . ' - ' . $objectArray['authority'] : $objectArray['authority'];
+        $objectArray['authority_name'] = !empty($objectArray['role_name'])
+            ? $objectArray['role_name'] . ' - ' . $objectArray['authority']
+            : $objectArray['authority']
+            ;
 
         /* get permissions list */
         $data = $objectArray['policy_data'];
@@ -127,7 +130,11 @@ class GetList extends GetListProcessor
         }
 
         $cls = '';
-        if (($objectArray['target'] === 'web' || $objectArray['target'] === 'mgr') && $objectArray['policy_name'] === 'Administrator' && ($this->userGroup && $this->userGroup->get('name') === 'Administrator')) {
+        if (
+            ($objectArray['target'] === 'web' || $objectArray['target'] === 'mgr')
+            && $objectArray['policy_name'] === 'Administrator'
+            && ($this->userGroup && $this->userGroup->get('name') === 'Administrator')
+        ) {
         } else {
             $cls .= 'pedit premove';
         }
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -79,7 +79,6 @@ MODx.grid.Grid = function(config) {
         };
 
         Ext.applyIf(config.groupingConfig, groupingConfig);
-
         Ext.applyIf(config,{
             view: new Ext.grid.GroupingView(config.groupingConfig)
         });
@@ -392,6 +391,7 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
                     ,direction: this.config.sortDir || 'ASC'
                 }
                 ,remoteSort: this.config.remoteSort || false
+                ,remoteGroup: this.config.remoteGroup || false
                 ,groupField: this.config.groupBy || 'name'
                 ,storeId: this.config.storeId || Ext.id()
                 ,autoDestroy: true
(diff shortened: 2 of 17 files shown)


smg6511:
@Mark-H @JoshuaLuckers - I see you're active on GH right now ... any chance you guys can give this a quick peek and (re)give it your blessing?

Mark-H:
Is there an urgency I'm missing that warrant 2 pings for a review on a PR created 2 days ago? ;) 

It's a big one so it'll take up a bit more time to check it again compared to giving a quick thumbs up on something small. Just trying to keep up with notifications in between actual work. I haven't done a full review round since before the 3.0.3 release so I'm not ignoring you, simply just haven't sat down for something of this magnitude.

smg6511:
I'm only hounding you guys (sorry) because this work has already been reviewed and approved — it's just been repackaged because of the rebasing snags I hit in the original. So my thought is that it should be quick and easy.

JoshuaLuckers (approved):

Mark-H:
> I'm only hounding you guys (sorry) because this work has already been reviewed and approved — it's just been repackaged because of the rebasing snags I hit in the original. So my thought is that it should be quick and easy.

That'd be my excuse if I want to sneak in something nasty in a PR with 2k changed lines, too! :D 

(Just kidding, but I'm still gonna go through it before giving you my thumbs.)

core/src/Revolution/Processors/Security/Access/UserGroup/Context/GetList.php
```diff
@@ -127,7 +130,11 @@ public function prepareRow(xPDOObject $object)
         }
 
         $cls = '';
-        if (($objectArray['target'] === 'web' || $objectArray['target'] === 'mgr') && $objectArray['policy_name'] === 'Administrator' && ($this->userGroup && $this->userGroup->get('name') === 'Administrator')) {
+        if (
+            ($objectArray['target'] === 'web' || $objectArray['target'] === 'mgr')
+            && $objectArray['policy_name'] === 'Administrator'
+            && ($this->userGroup && $this->userGroup->get('name') === 'Administrator')
+        ) {
         } else {
```
Mark-H, line 138: Is this if/else right? It looks like it was the same before, so I assume yes, but the empty body for the `if {}` looks weird.

smg6511:
@Mark-H - Yes, and I'd assume that — gotta make sure I didn't drop something along the way. Just saying it should be easier to look at your second time around this block ;-)

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -892,6 +930,82 @@ Ext.extend(MODx.grid.Grid,Ext.grid.EditorGridPanel,{
...
+     */
+    ,refreshFilterOptions: function(filterData = [], clearDependentParams = true) {
+        if (filterData.length > 0) {
+            filterData.forEach(data => {
+                const filter = this.getFilterComponent(data.filterId);
+                if (filter) {
+                    const store = filter.getStore();
+                    filter.setValue('');
+                    if (store) {
+                        if (data.hasOwnProperty('dependentParams')) {
```
Mark-H, line 980: Code style wise, early returns help keep the code simple to understand and debug. This is 4 levels deep. No need to change that immediately as it's a nitpick, but for future consideration when implementing new features.

Mark-H (approved):

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Yes, I'm more aware these days of the utility in using early returns, though I might miss some here and there ;-) In this case, you can't as you cannot break out of `forEach`. That's fine, because I don't expect or want to break out of this loop, I'm really just trying to account for something going wrong (maybe too much so?).

core/src/Revolution/Processors/Security/Access/UserGroup/Context/GetList.php
smg6511: It is right, but perhaps this would read better (replacing the entire if else block):
```php
$cls .= ($objectArray['target'] === 'web' || $objectArray['target'] === 'mgr')
            && $objectArray['policy_name'] === 'Administrator'
            && ($this->userGroup && $this->userGroup->get('name') === 'Administrator')
            ? ''
            : 'pedit premove'
        ;
```
That said, I'd leave it for now as I have a different [proposed way](https://github.com/modxcms/revolution/pull/15919#issuecomment-980792211) of expressing the permissions that makes way more sense than sticking them in css class names, which would make this moot if accepted.

opengeek:
@smg6511 — just to confirm, this doesn't need to be ported back into 3.0.x, correct?

smg6511:
> @smg6511 — just to confirm, this doesn't need to be ported back into 3.0.x, correct?

Correct, this can wait for the feature release.

PR #16327, opened 2022-12-09
Add PHP 8.2 as CI target + minor associated fixes

Mark-H:
### What does it do?

- Add PHP 8.2 as CI target
- Include php compat checker in phpcs, and fix some minor compatibility issues
- update actions/checkout@v2 to @v3 as @v2 is deprecated
- fix potential mb_str_split error on PHP 7.2
- fix potential typeerror on PHP 8 in modLexicon with invalid lexicon configurations
- remove a few legacy (php <5.4) checks that are obsolete

### Why is it needed?
PHP 8.2 is out! 🎉 

Tried running php compat to see if there's any issues to be expected. Found a few things, tho mostly legacy stuff. Figured these things are so small it's okay to review as one. 

### How to test
N/a

### Related issue(s)/PR(s)
N/a

PR #16359, opened 2023-01-21
Cleanup legacy PHP version references

JoshuaLuckers:
### What does it do?
Did some clean up of PHP version specific fixes that are no longer needed.
I also made sure we check for the correct current minimum PHP version we support.

### Why is it needed?
Clean up old code and make things more consistent. 

### How to test
n/a

### Related issue(s)/PR(s)
n/a


Ibochkarev (approved):
LGTM 👍

Mark-H (approved):

Mark-H:
Related to #16327, I don't think it should conflict from the looks of it.

PR #16361, opened 2023-01-23
Add resource alias in template picker as needed

### LINKED ISSUE #16178: It is impossible to create a New Document form Resources Tree if "Automatically generate alias" is set to "No" in System Settings
### Summary
If "Automatically generate alias" is set to "No" in System Settings it is impossible to create a New Document from Resources Tree since **there is no field for "Alias" in the Popup window**.

### Step to reproduce
Set "Automatically generate alias" to "No" in System Settings and try to create new Document from Resources Tree.

### Expected behavior
There shloud be field for "Alias" in the Popup window.

### Environment
MODX Revolution 3.0.1

smg6511:
### What does it do?
Conditionally includes the `alias` field when the `friendly_urls` is enabled and `automatic_alias` is disabled.

### Why is it needed?
Given the scenario above, a resource can not be saved via the template picker because the alias is not present.

### How to test

1. Clear all caches
2. Enable the template picker system setting (`enable_template_picker_in_tree`)
3. Toggle the `friendly_urls` and `automatic_alias` system settings in various combinations, each time performing a hard reload of the page
4. Create a new Resource via the add ("+") button in the Resources tree and verify the ability to save no matter what settings values you have chosen.

### Related issue(s)/PR(s)
Resolves #16178


manager/assets/modext/widgets/resource/modx.window.resource.js
```diff
@@ -1,7 +1,69 @@
...
-    this.ident = config.ident || 'tplpick'+Ext.id();
-    var id = this.ident;
+MODx.window.CreateResource = function(config = {}) {
+    this.ident = config.ident || `tplpick${Ext.id()}`;
+    const id = this.ident,
+          requireAlias = parseInt(MODx.config.friendly_urls, 10) && !parseInt(MODx.config.automatic_alias, 10),
+          aliasLength = parseInt(MODx.config.friendly_alias_max_length, 10) || 0,
+          resourceDetail = [
+              {
+                  columnWidth: requireAlias ? 0.33 : 0.5,
```
Mark-H, line 8: Widths don't actually have to add up to 1 specifically. ExtJS will total all widths, divide by number of columns, and size accordingly. So this conditional isn't technically needed. The conditional also doesn't hurt, so no need to go back and remove it, but just FYI.

smg6511: Will take a look at that for future consideration.

Mark-H (approved):

rthrash:

https://community.modx.com/t/modx-3-create-document-inside-fails-to-create-anything/6140/21

PR #16378, opened 2023-02-02
JavaScript linter: step two

mishanthrop:
### What does it do?
Continued implementation of the eslint: https://github.com/modxcms/revolution/pull/16321

In this PR I went through rule `no-var`, which recommends using `let` or `const` instead of `var`.

Сode example from https://eslint.org/docs/latest/rules/no-var:
```javascript
var count = people.length;
var enoughFood = count > sandwiches.length;

if (enoughFood) {
    var count = sandwiches.length; // accidentally overriding the count variable
    console.log("We have " + count + " sandwiches for everyone. Plenty for all!");
}

// our count variable is no longer accurate
console.log("We have " + count + " people and " + sandwiches.length + " sandwiches!");
```

During fixing issues, I found many problematic places where the variable are not declared or it is difficult to understand what the scope of the variable is. I'll save it for other PRs.

P.S. I limited the number of edits and files so that this PR was easy to review.

### Why is it needed?

### How to test
```sh
npm i
```

```sh
npm run js:lint
```

How to test specific rule:
```
./node_modules/eslint/bin/eslint.js "manager/assets/**/*.js" --no-eslintrc --env "es6" --env "node" --parser-options "{ecmaVersion: 2018}" --rule "{no-var: error}"
```

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/16321


Mark-H:
Just to make sure I'm following correctly, this just applies the one rule (which admittedly can be complex to apply if it wasn't being applied correctly before like your example), so this does not cover the _full set_ of rules you would propose we adopt?

We've been chatting in Slack that it might be useful to collect the desired code style proposals from both you and @smg6511 (who indicated you could not reach consensus amongst just the two of you), and to put it to a vote so we have a clear idea of where we're headed with this.

mishanthrop:
@Mark-H 

`> this just applies the one rule, so this does not cover the full set of rules you would propose we adopt`
In this PR -- yes. 

Initially I stuck to adding rules gradually. While smg6511 wanted to create the most ideal ruleset in first PR. The only disagreement as far as I know.

In this PR I used the ruleset (99%) that smg6511 suggested in previous PR.

P.S. I'd love to discuss this on Slack, but I couldn't quickly find information on how to join the MODX Slack channel.

JoshuaLuckers:
@mishantrop to join Slack you need to have an invite, if you go to this link it should work: https://modx.org/

smg6511:
Guys - Love Slack, but there's one problem: Since MODX as does not have a pro-level Slack account, all discussion older than 90 days becomes inaccessible. I get the feeling this subject of code style might be ongoing for a while—not that we'll not be making progress in the next 90 days, just that it _could_ be a multi-pass process that'll take longer than that time period to complete. It might be best to have the conversation in Community/Development under a new thread that we bookmark or perhaps gets pinned (which I think only the code owners can make happen?) so it'll be easily visible for a while.

@mishantrop - A couple things:
- Thanks for mostly aligning the rules to what I initially suggested. There are a few changes I'm not crazy about:
   - **comma-dangle:** I won't put up a big fight about this one but, even after researching why many devs prefer these, I don't like the use of them for a couple reasons—I don't like the look (less important) and I think it can give the impression that something is missing (the dev forgot to or intends to add another entry to an array or object). I understand it makes diffs a _little_ cleaner, but IMO it's a small gain in the context of our codebase and how much activity there is in changing/adding to it.
   - **quotes:** I see you probably prefer single quotes as I do since you set the rule to "warn," but I'd be more aggressive with enforcing the preference for them. Again, not going to make a big deal about this one, but think about it.
   - **one-var:** This one I do object pretty strongly to—I think using declaration keywords for every single variable just produces a lot of visual noise.
- Know that I'm not looking for perfection in the initial rollout of the rules, but I know from experience how few people we have reviewing PRs and sometimes even small changes take a long time to get reviewed. The idea of taking multiple passes to update the codebase is nice, I just think it'll take forever—even with the smallest of changes (for example, fixing comma placement), the results will still need to be tested. That's why I encourage taking a little time to establish the ruleset itself, then applying that in a strategic way so we have a chance of the style improvements making it into releases sooner than later.

PR #16383, opened 2023-02-08
Fix vtabs + grid horizontal overflow

### LINKED ISSUE #16380: Access Permissions grid horizontal overflow
### Summary

The context menu action column is pushed off screen. See screenshot:

Screenshot 2023-02-07 at 10-24-39 User Group Editor MODX Revolution

### Step to reproduce
Go to ACLs -> Access Permissions Tab 
You may need a longer access policy name to reproduce (e.g. see mine in screenshot above)

### Observed behavior
Overflow hides column which should be visible.

### Expected behavior
Column should be visible

### Environment
MODX version 3.x git

muzzwood:
### What does it do?
Prevents `modx-vtabs` from causing part of the user group access permissions grid from being hidden.

### Why is it needed?
So users can see the right-hand column.

### How to test
Navigate to the grid and see the column is now visible.

### Related issue(s)/PR(s)
Resolves #16380

### DIFF
--- a/manager/assets/modext/widgets/security/modx.panel.user.group.js
+++ b/manager/assets/modext/widgets/security/modx.panel.user.group.js
@@ -135,6 +135,7 @@ MODx.panel.UserGroup = function(config) {
                         ,itemId: 'user-group-context-access'
                         ,hideMode: 'offsets'
                         ,layout: 'form'
+                        ,autoWidth: false
                         ,items: [{
                             html: '<p>'+_('user_group_context_access_msg')+'</p>'
                             ,xtype: 'modx-description'


smg6511 (commented):
I do see the issue, but this change has no effect for me (in Chrome v109.0.5414.119). Another angle, temporarily at least, would be to lower the width of the policy column from 200 to say 100 in `modx.grid.user.group.context.js` (line 64). I'm thinking the issue might stem from the row expander when there is a large set of permissions (but that's just a hunch).

muzzwood:
@smg6511 Thanks for testing it.
Hmm I've just tested it again, and the fix works for me in Chromium, Brave and Firefox. 
I am on Linux though, so perhaps it differs by OS.

I tried switching to a _load only_ access policy with only one permission showing in the row expander, but the issue persists when `autoWidth` isn't set to `false`.

I've discovered it's not just that grid, but all grids nested within the vtabs component that are affected. I didn't see it before because they had no rows, and the action column doesn't have a header label. The autoWidth change fixes them all (for me at least). 

That's certainly a pain it's not working for you. 🤔

smg6511:
@muzzwood - Ok, I'll keep looking at it too, and on a couple other browsers. I do think it centers around the rowExpander, as when I disable that in the column model, the issue disappears. I do not see the issue in the other grids on Chrome. It probably has something to do with how different browsers handle the box model and calculate the widths; perhaps there's something in the css that's contributing to the issue.

In your testing environment, do all the permissions grids contain enough rows to trigger the scroll bar?

Mark-H:
I did manage to reproduce and confirm this, thanks @muzzwood.

PR #16392, opened 2023-02-20
Remove Unused Window Config for Lexicons Grid

smg6511:
Came across this when working on another PR. There is no export Lexicons feature, nor was there one in 2.x. 

### How to test
No testing to do per-se. Just verify that there is, in fact, no implementation that removing this code would break.

### Related issue(s)/PR(s)
None


Mark-H:
Huh, that's funny. There is lexicon entry for `lexicon_export_desc` and a window definition, but no processor or any usages I can find either. 

Looks like it was removed [in 2.1.0](https://github.com/modxcms/revolution/commit/05551131788a8cd6ef343bdc951aea8223b98a4e#diff-67b383681ba332c789685cab38f07a646ad972d1c5cc725e4c8ab18d6ef8bcbc), just 12 years ago 😂

Mark-H (approved):

JoshuaLuckers:
It's a feature that's been requested a few times, wouldn't it be a better idea to bring back this feature?

smg6511:
@JoshuaLuckers - While it might be a nice feature for someone to redevelop, it's been an awful long time since it's been around. The code being deleted is just stray at this point and shouldn't been in the production codebase, right?

JoshuaLuckers:
> @JoshuaLuckers - The code being deleted is just stray at this point and shouldn't been in the production codebase, right?

You're right!

PR #16394, opened 2023-02-22
Use History API to save path of Media Browser

bezumkin:
### What does it do?
Enables using History API for MODX Media Browser.

### Why is it needed?
The current MODX media browser doesn’t save the last file location path properly. Content editors are spending a lot of time to navigate to the right folder within the MODX Media Browser. 

After every page refresh the Media Browser doesn’t remember the last file path. This can be a pain in case of deep file structures. 

### How to test
Just use browser navigation keys in Media Browser. 

### Related issue(s)/PR(s)
N/a

### Demo

Media Browser:

https://user-images.githubusercontent.com/1257284/220913381-407f0d24-df9c-47a7-a62a-0b4df03bc8dd.mov

<br/><br/>
Browser Window:

https://user-images.githubusercontent.com/1257284/220913430-89829998-dd31-491d-89cb-a3b8af560f67.mov

---

1. `MODx.tree.Tree` now has new method to expand and select node by specified path: `expandTreePath(path)`

2. Default components `source` and `openTo` has priority over URL params. This should prevent the ability to load another source for TV.

3. Browser Window remove saved params from URL on regular close. But if you reload page or open by direct link, window should be opened on saved path.

I didn't added work with History API to `MODx.browser.RTE`, because of 2 reasons:
- as I see, it is not used in MODX manager by default
- I'm not familiar with RTE solutions

Maybe RTE authors will update this component too.

### DIFF
--- a/manager/assets/modext/util/utilities.js
+++ b/manager/assets/modext/util/utilities.js
@@ -36,6 +36,37 @@ MODx.util.Progress = {
     }
 };
 
+MODx.util.UrlParams = {
+    get() {
+        return this.parse(window.location.search)
+    },
+    set(data) {
+        const params = decodeURIComponent(new URLSearchParams(data).toString())
+        if (params.length) {
+            window.history.pushState(params, '', document.location.pathname + '?' + params);
+        } else {
+            window.history.pushState('', '', document.location.pathname);
+        }
+    },
+    add(key, val) {
+        const params = this.get()
+        params[key] = val
+        this.set(params)
+    },
+    remove(key) {
+        const params = this.get()
+        delete params[key]
+        this.set(params)
+    },
+    clear() {
+        this.set({})
+    },
+    parse(str) {
+        const params = new URLSearchParams(str)
+        return Object.fromEntries(params.entries())
+    }
+}
+
 /** Adds a lock mask to an element */
 MODx.LockMask = function(config) {
     config = config || {};
--- a/manager/assets/modext/widgets/core/tree/modx.tree.js
+++ b/manager/assets/modext/widgets/core/tree/modx.tree.js
@@ -901,6 +901,19 @@ Ext.extend(MODx.tree.Tree,Ext.tree.TreePanel,{
         return langs;
     }
 
+    ,expandTreePath(dir = '/') {
+        const root = this.getRootNode().getPath('text')
+        const path = root.replace(/\/$/, '') + '/' + dir.replace(/^\//, '')
+        this.expandPath(path, 'text', () => {
+            let node = this.getNodeById(encodeURIComponent(dir))
+            if (!node) {
+                node = this.getRootNode()
+            }
+            node.select()
+            this.cm.activeNode = node
+        })
+    }
+
 });
 Ext.reg('modx-tree',MODx.tree.Tree);
 
--- a/manager/assets/modext/widgets/media/modx.browser.js
+++ b/manager/assets/modext/widgets/media/modx.browser.js
@@ -522,10 +522,11 @@ MODx.browser.Window = function(config) {
             this.view.run();
         }
         ,scope: this
-        ,source: config.source || MODx.config.default_media_source
+        ,source: config.source || MODx.util.UrlParams.get().source || MODx.config.default_media_source
+        ,originalSource: config.source || MODx.config.default_media_source
         ,hideFiles: config.hideFiles || MODx.config.modx_browser_tree_hide_files
         ,hideTooltips: config.hideTooltips || MODx.config.modx_browser_tree_hide_tooltips || true // by default do not request image preview tooltips in the media browser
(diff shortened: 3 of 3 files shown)


Mark-H:
Long time no see! ;) Hope you're well, Vasily.

Can confirm it works as advertised, but I do have a few questions/thoughts: 

1. Is it intentional that this only affects the standalone media browser (`/manager/?a=media/browser`), and not, for example, a media browser that is opened by things like the image TV or RTEs? Users would likely spend more time with TVs and RTEs so I'm wondering if that was overlooked, or if I'm the one overlooking something. :)

2. Building on 1), I do think that when used in the image TV or from RTEs there should be a way for those to force it to open in a certain place. Or at least, force it to be be re-opened within a certain source or path if that is what the opener expects. Right now that can be done by providing the `openTo` and `source` properties, but those would be overridden. In the current iteration it only affects the standalone browser, so this is irrelevant, but wondering if you have thoughts on that in case this does get extended to also affect image TVs/RTEs.

3. Is `MODx.util.History` the right name for a utility that stores information in the URL? I get it is based on the History API but it might give the impression the "history" will be stored, and that it will remember, for example, what URL was opened the last time if you return to the manager without the query parameters there. 

4. Rather than a full page reload on `popstate`, it would be better if it would navigate to the source/path without a refresh. While it's hard to tell the refresh in your video, when using remote media source types (S3, Google Drive) or on cheap shared servers that should be much faster to navigate, especially if the path already is loaded in the view.


@Ibochkarev We're now targeting 3.1 for features and improvements, with only bugfixes being backported to 3.0.4. So that milestone assignment should've been 3.1. ;)

Ibochkarev:
@Mark-H ok

JoshuaLuckers:
> I didn't found a way to smoothly manipulate files tree without reloading the page, but even with reload it is much better than it was before.

Would this be a possible way to do it?
https://github.com/modxcms/revolution/blob/e7607a239218ba855dbd60805948e34119f913c5/manager/assets/modext/widgets/media/modx.browser.js#L562-L571

bezumkin:
@Mark-H Hello!

1. Yes, it was intentional.

2. I believe, If user have direct link in URL to some directory - it is okay to override default values.

3. I can rename it to `MODx.util.UrlParams`, no problem

4. I had not enough time for deep digging ExtJS components, to  use internal methods for tree manipulation. I'm not sure it's even possible without major changes.

Right now Sterc provided some more hours for this task, so I will try to improve this pull request.

bezumkin:
@Mark-H updated PR and description

smg6511:
I'll try to dig into this in about a week or so...

Ibochkarev:
@smg6511 ping...

smg6511:
Sorry guys, I was hoping to review this before traveling on vacation (which I'm currently in the middle of). I will focus on this after I get back (the week of the 25th).

JoshuaLuckers:
ping @smg6511

smg6511:
I finally have had a chance to start digging into this. There is a built-in singleton class (Ext.History) meant for handling what this PR seeks to do. Ideally, we don't reinvent the wheel and we leverage that class to do the job. In a test branch I created this isn't hard to achieve when navigating within a single source; it's when you change sources or navigate to another page/view and return to the file browser that a challenge is presented. It's a bit of a brain-teaser but I think it's totally doable. I'll continue playing with that.

As to the functionality of this PR as-is, it works well for the most part. I did find that certain characters (like a "+") in folders/files caused it to break, so if this PR's strategy were carried through a harder look at the encoding/decoding during the process would need to be done. Also there are a couple code style issues, most prominently the omission of terminating semicolons, that would need to be fixed for me to approve.

At any rate, it's great that this feature has been put forward, as the file browser can be _really_ cumbersome to work in!

gpsietzema:
So, just wondering here: @smg6511 , Jim, would you give this your blessing to merge this? If not, could you specify what you would like to have seen different, then Vasily could modify that to get it merged. If not, it feels like this PR will die a slow death and we'll look at in 2 years and ask ourselves why its still here.

smg6511:
@gpsietzema - Yes, I'll finally have some time to refocus on this over the weekend and provide specific feedback or an alternative next week.

OlegAnTo2000:
@smg6511 Hi) this is a very important feature for working with content in modx. i have several sites and miss it very much. only recently i found out that Vasily has already made it. Please check it 😊.

joeke (approved):

Mark-H (approved):

Mark-H:
Thanks for updating it based on my earlier feedback, looking good 👍 

Merged in https://github.com/modxcms/revolution/commit/dcf11a9b1b25b79f3e328430e5aa8f50c66e5974 even though github isn't automatically setting it as merged for some reason.

Mark-H:
Figured out why it didn't autoclose, I messed up on my end, oops 😂

PR #16369, opened 2023-01-30
Grid filtering refactor for improved accuracy and persistence via URL parameters

smg6511:
### What does it do?
Built upon work merged in #16089 and #16355 to provide better filtering functionality and persistence for the majority of grids. Additional methods were added to the Tabs and Grids base classes to further optimize the code (_e.g._, centralizing the repetitive task of creating the query and clear buttons, which are for the most part exactly the same everywhere they are applied). 

### Why is it needed?
To apply the new global persistence feature to the referenced grids and improve their UIs.

### Grids covered

1. **Media Sources** - Media > Media Sources (/manager/?a=source)
2. **Packages** - Extras > Installer (/manager/?a=workspaces)
3. **Users** - Manage > Users (/manager/?a=security/user)
4. **User Settings** - Manage > Users > [User] Settings (manager/?a=security/user/update&id=1)
5. **Messages** - User (user icon) > Messages (/manager/?a=security/message)
6. **System Settings** - Settings (gear) > System Settings (/manager/?a=system/settings)
7. **System Events** - Settings (gear) > System Settings, 2nd tab (/manager/?a=system/settings)
8. **Form Customization Profiles** - Settings (gear) > Form Customization (/manager/?a=security/forms)
9. **Profile** - Settings (gear) > Form Customization > [Profile] (manager/?a=security/forms/profile/update&id=1)
10. **Dashboards** - Settings (gear) > Dashboards (/manager/?a=system/dashboards)
11. **Widgets** - Settings (gear) > Dashboards, 2nd tab (/manager/?a=system/dashboards)
12. **Contexts** - Settings (gear) > Contexts (/manager/?a=context)
13. **Context Settings** - Settings (gear) > Contexts > [Context], 2nd tab (/manager/?a=context/update&key=[yourcontext])
14. **User Groups & Users** - Settings (gear) > Access Control Lists [click UG in left tree] (/manager/?a=security/permission&group=1)
15. **Access Policies** - Settings (gear) > Access Control Lists, 3rd tab (/manager/?a=security/permission)
16. **Policy Templates** - Settings (gear) > Access Control Lists, 4th tab (/manager/?a=security/permission)
17. **Access Permissions/Contexts** - Settings (gear) > Access Control Lists > [User Group], 2nd tab (manager/?a=security/usergroup/update&id=1)
18. **Access Permissions/Resource Groups** - (same path as # 17, 2nd vertical tab)
19. **Access Permissions/Element Categories** - (same path as # 17, 3rd vertical tab)
[...]

PR #16397, opened 2023-02-23
Use Codecov in informational mode.

JoshuaLuckers:
### What does it do?
Use Codecov in informational mode

### Why is it needed?
See status checks but prevent them from blocking / failing the build status.

> Use Codecov in informational mode. Default is false. If true is specified the resulting status will pass no matter what the coverage is or what other settings are specified. Informational mode is great to use if you want to expose codecov information to other developers in your pull request without necessarily gating PRs on that information.

### How to test
See builds pass the code coverage checks.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,11 @@
 comment:
     require_changes: true
+
+coverage:
+    status:
+        project:
+            default:
+                informational: true
+        patch:
+            default:
+                informational: true


Ibochkarev (approved):

smg6511 (approved):

Mark-H:
I get _why_.. but wouldn't it be better if we consistently encourage adding tests for code they change to avoid a failing status being given? That way at least the coverage can't get worse.

JoshuaLuckers:
> I get _why_.. but wouldn't it be better if we consistently encourage adding tests for code they change to avoid a failing status being given? That way at least the coverage can't get worse.

Yes I agree, it would be better but I don’t think it’s a good idea right now. It’s already hard enough to encourage people to contribute.. The status checks would fail most of the time with the current coverage we have right now.

Mark-H:
The status check only fails if the coverage gets _lower_. I.e. lots of code being added that is not covered. 

I think I did see a PR (one of yours iirc?) where some one-line changes caused it to see a large drop relative to a small file, but perhaps cases like that can be averted just by tweaking the rules a bit..

smg6511:
I agree that throwing the failures would tend to _discourage_ and, similar to when the PSR rules were first introduced in the workflow, can give reviewers a false impression that the PR is _not_ ready. I think the notifications are enough for now. You guys will see where they are popping up and can request the addition of unit tests where you deem them necessary.

As an example, take a look through #16369; I did take note of the coverage notices but have a hard time even conceiving of how many of the marked areas could _practically_ be unit-tested. (Note that I'm inexperienced in building unit tests and may well just not be seeing it.) Also, areas that contain simple formatting changes (e.g., to satisfy a PSR rule) are triggering coverage notices. I'm guessing that's one where a rule tweak comes into play.

JoshuaLuckers:
> I think I did see a PR (one of yours iirc?) where some one-line changes caused it to see a large drop relative to a small file, but perhaps cases like that can be averted just by tweaking the rules a bit..

Yes, that’s the patch status that caused it to fail. And I still agree with you that it would be better to motivate people to increase code coverage. So if someone has time to tweak the rules I don’t have any objection to it.

opengeek:
@Mark-H — do you approve of this? Or do we need to have further discussion on it?

JoshuaLuckers:
ping @Mark-H

Mark-H:
I have nothing better to contribute at the moment, so no objections. I may revisit the configuration in the future and propose a different way to go about it.

PR #16398, opened 2023-02-24
Fix date TV parsing bug when time is hidden

### LINKED ISSUE #16393: After enabling the hidden time option a Date TV looses its value in MODX 3.x
### Summary
After enabling the hidden time option a Date TV looses its value in MODX 3.x.

&#x1F534; **IMPORTANT**: The issue occurs also after updating the site from MODX 2.x to MODX 3.x In MODX 2.x. the value was displayed and saved right. After updating to MODX 3.x the value is displayed AND saved wrong.

### Step to reproduce
Create a date TV and edit a resource with that TV. Set the time to some value different than 0:00. Enable the 'Hide Time Option' in the TV and edit the resource again.

### Observed behavior
The displayed date is set to 1.1.1970. The wrong value is saved, when the resource is saved.

### Expected behavior
Only the time part of the date should be removed and the date part should remain as it was previously entered.

### Environment
MODX 3.0.3

smg6511:
### Why is it needed?
Dates without time were not rendering correctly for TVs

[Some more context here](https://github.com/modxcms/revolution/issues/16393#issuecomment-1442461134).

### How to test
1. Create a new date TV with "Hide Time Option" enabled, assign it to a Template, and enter/save values in a Resource to verify correct behavior.
2. Edit an existing date TV (with time enabled, and having already assigned/saved a value in your test Resource), enabling "Hide Time Option." Verify that the TV's saved date still shows as expected.

### Related issue(s)/PR(s)
Resolves #16393


JoshuaLuckers:
It sometimes still picks the current time instead of `00:00:00`.
I have the default date and time set to: `Today (midnight)`.

1. Select a date and save (time is saved as 00:00:00).
2. Empty the input and save (empty value is saved).
3. Pick a new date (time is saved to the current time).

smg6511:
@JoshuaLuckers - Interesting, I'll take a look at that. It gets into how default values are evaluated and the question is: How to make the processor aware that hide time is enabled when saving that default value (might be easy, hopefully)? Will investigate in the next couple days...

smg6511:
@JoshuaLuckers - Alright, it was easy! Take a second look...

rthrash:

https://community.modx.com/t/every-upgrade-from-2-to-3-ends-in-a-disaster/6587/32

JoshuaLuckers (approved):

Jako (approved):

PR #16403, opened 2023-02-27
Fix visibility of Symlink, Weblink, and Static Resource in contextual menus (create) and combos

smg6511:
### What does it do?
Properly sets the `showInContextMenu` and `allowListingInClassKeyDropdown` class properties to ensure correct visibility of context menu items and combo options based on user permissions to create Symlinks, Weblinks, and/or Static Resources.

### Why is it needed?
This picks up on what #16067 is trying to solve, but uses functionality that is already present (just not taken advantage of) to achieve the end result. In this way, the visibility applies not only to the template picker but to other areas that utilize/extend the Derivatives classes, as well as the tree's context menu (applies to create visibility only).

### How to test

1. Create a secondary user and a new ACL policy for that user's usergroup.
2. Toggle the `new_weblink`, `new_symlink`, and `new_static_resource` permissions and verify that, when off, the related Resource Type combo option items do not show (in the template picker window, in the main Resource editing panel [under Settings], and if you have Collections installed in the Collections Settings are [_Default children's resource type_ combo]). Also note that when using the context menu in the tree, these items should only show under Create and Quick Create when the user has the appropriate permissions.

### Note ###
This only addresses visibility in the tree for create/new permission. More should be done on this subject to control the visibility of edit and delete on tree nodes where permission is not granted for the give Resource type. But that should be handled separately, as it may take a bit more "doing" to get that sorted.

### Related issue(s)/PR(s)
This is an alternate to PR #16067 to solve #15893 and more.

### DIFF
--- a/core/src/Revolution/modStaticResource.php
+++ b/core/src/Revolution/modStaticResource.php
@@ -27,7 +27,9 @@ class modStaticResource extends modResource
     {
         parent:: __construct($xpdo);
         $this->set('class_key', __CLASS__);
-        $this->showInContextMenu = true;
+        $canCreate = (bool)$this->xpdo->hasPermission('new_static_resource');
+        $this->allowListingInClassKeyDropdown = $canCreate;
+        $this->showInContextMenu = $canCreate;
     }
 
     /**
--- a/core/src/Revolution/modSymLink.php
+++ b/core/src/Revolution/modSymLink.php
@@ -20,7 +20,9 @@ class modSymLink extends modResource
         parent:: __construct($xpdo);
         $this->set('type', 'reference');
         $this->set('class_key', __CLASS__);
-        $this->showInContextMenu = true;
+        $canCreate = (bool)$this->xpdo->hasPermission('new_symlink');
+        $this->allowListingInClassKeyDropdown = $canCreate;
+        $this->showInContextMenu = $canCreate;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


JoshuaLuckers:
We also need to check if you are allowed to `save` the objects right? 

Can someone confirm (@Mark-H @opengeek) if instead of doing this: `$canCreate = (bool)$this->xpdo->hasPermission('new_static_resource');`  it would be better to do it like this?
`$canCreate = $this->checkPolicy(['save','new_static_resource']);`

smg6511:
@JoshuaLuckers - According to my tests, neither the basic `save` nor the basic `create` permissions are required to create new derivative Resources.

smg6511:
@JoshuaLuckers - Hey there! Can you revisit this? Thx ;-)

opengeek:
> @JoshuaLuckers - According to my tests, neither the basic `save` nor the basic `create` permissions are required to create new derivative Resources.

`save` and `create` would only be applicable when checking the modAccessResourceGroup target, so those would not be applicable in a check that was searching for more specific context-assigned permissions like `new_static_resource`. Otherwise, you would be checking against any number of policy targets for the generic `save` and `create` permissions, which could simply be saying whether a user has access to save or create a context or a media source.

smg6511:
@Mark-H @muzzwood - Hey guys, since you both were involved in this (Mark, initial issue report; Murray, preliminary PR) I'm hoping you can carve out some time to give this a quick review.

PR #16404, opened 2023-02-27
Add guard-clause to clearDirty method

### LINKED ISSUE #143: Please add the field multiselect
Thank you for completing the package and solving issue # 61. We also ask you to expand the possibilities and add a new type of the multiselect field

### LINKED ISSUE #202: Feature-Request: Radio-Button Settings
I would like to be able to use radio-inputs for some settings. 
In my opinion that would be useful for little "switch-settings" like for example "enable/disable custom scrollbar".
Also, as currently I am using dropdowns for these two options, this would save the user 1 click :-) 

As Mark attempted in #176 with the multi-select checkboxes, I also ran into the same JS-issue. 
Jako also stumbled upon the exact same issue https://github.com/modxcms/revolution/issues/12177 some time ago.

I will prepare a PR for my own feature-request here, and I am attempting a PR on the actual issue in the MODX code.

### LINKED ISSUE #12177: checkboxgroup/radiogroup issue in MODx.FormPanel
If a MODx.FormPanel contains a checkboxgroup/radiogroup this panel is not rendered because of nodeToRecurse.items is an array in utilities.js (line 65) for the checkboxgroup/radiogroup item and that has not the method each.

I don't know how to fix that without breaking recursive work of that function.

sebastian-marinescu:
### What does it do?
This makes sure the JS doesn't throw an error in the case that the items-object is an array.

### Why is it needed?
When using ExtJS to create components of the `xtype: 'radiogroup'` to be used in CMPs inside a form, 
this function gets called and tries to iterate over it's items array with a method that doesn't exist. 

This PR isn't changing the way forms or the clearDirty-method are handled, it just assures no error is thrown when using radiogroup-components.

### Related issue(s)/PR(s)
Resolves #12177
Resolves https://github.com/modmore/ClientConfig/issues/202
Resolves https://github.com/modmore/ClientConfig/pull/176
Resolves https://github.com/modmore/ClientConfig/issues/143


manager/assets/modext/util/utilities.js
```diff
@@ -82,6 +82,9 @@ Ext.override(Ext.form.NumberField, {
 Ext.override(Ext.form.BasicForm,{
     clearDirty : function(nodeToRecurse){
         nodeToRecurse = nodeToRecurse || this;
+        if (!nodeToRecurse || !nodeToRecurse.items || typeof nodeToRecurse.items.each !== 'function') {
+            return;
+        }
         nodeToRecurse.items.each(function(f){
```
smg6511: I believe you could use optional chaining to simplify this to:
```suggestion
        nodeToRecurse?.items.each?.(function(f){
```
I don't have a quick way to verify this with the case you mention, but you probably can.

sebastian-marinescu: Also thought of that, but as in the docs it states that the [manager also supports Internet Explorer 11](https://docs.modx.com/3.x/en/getting-started/server-requirements), I don't think we can use ultra-modern JavaScript. The [Internet Explorer does not support](https://caniuse.com/?search=optional%20chaining) optional chaining.

smg6511: We're already using optional chaining as well as other ES6+ features elsewhere in the code. The support page probably needs updating, especially since IE was officially retired by Microsoft as of June last year.

sebastian-marinescu: Well, in that case 😃

smg6511 (approved):
For the simple purpose of avoiding error, this does the job. Supporting checkboxgroup/radiogroup types may need to be revisited, as it's hard to know whether the clearDirty method is taking effect on those component types without setting up a special CMP for the purpose of testing.

Ibochkarev (approved):

sebastian-marinescu:
Very nice! :mechanical_arm:

PR #16405, opened 2023-02-28
Update minimum PHP version to 7.4 and recommended version to 8.0.0

JoshuaLuckers:
### What does it do?
Drop support for PHP 7.2 and 7.3 and update minimum PHP version to 7.4 and recommended version to 8.0.0

### Why is it needed?
Reduce the amount of PHP versions we have to support, especially if they are End of Life. 
We can't drop support for 7.4 just yet because it's still widely in use.

### How to test
-

### Related issue(s)/PR(s)

### DIFF
--- a/setup/includes/test/modinstalltest.class.php
+++ b/setup/includes/test/modinstalltest.class.php
@@ -66,8 +66,8 @@ abstract class modInstallTest
         $this->title('php_version', $this->install->lexicon('test_php_version_start') . ' ');
         $phpVersion = phpversion();
 
-        $recommended_version = "7.3";
-        $required_version = "7.2.5";
+        $recommended_version = "8.0.0";
+        $required_version = "7.4.0";
 
         $php_ver_comp = version_compare($phpVersion, $required_version, '>=');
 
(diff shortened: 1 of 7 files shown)


Ibochkarev (approved):

setup/includes/test/modinstalltest.class.php
```diff
@@ -66,8 +66,8 @@ protected function _checkPHPVersion()
         $this->title('php_version', $this->install->lexicon('test_php_version_start') . ' ');
         $phpVersion = phpversion();
 
-        $recommended_version = "7.3";
-        $required_version = "7.2.5";
+        $recommended_version = "8.0.0";
```
rthrash, line 69: At this stage, PHP 8.0.x is no longer supported. Should this be bumped to 8.1.0 (supported through Nov 25, 2024)?

JoshuaLuckers: I’m no longer involved in this project so I don’t know. Maybe @opengeek or @Mark-H can help decide this.

PR #16406, opened 2023-03-05
Update modx.layout.js to fix margin issue

### LINKED ISSUE #16182: When sidebar is retracted the gap on the left side is too narrow
### Summary
Retract the sidebar and you'll see the gap is about half the width of when it's expanded.
The main content div overlaps the side menu bar. 

Here I'm using dev tools to hightlight the div:
Screenshot from 2022-05-13 17-41-39

### Step to reproduce
Retract the sidebar, and notice the gap is half the size. Go to the error log page as an example, and press the refresh button and you'll see the loading mask overlaps.

### Observed behavior
As above

### Expected behavior
The gap should be the same width as when the side bar is extended.

### Environment
MODX 3.0.1

smg6511:
### What does it do?
Sets the content margin not only on render, but on collapse and expand of the tree panel. A handful of code optimization and formatting fixes were made as well.

### Why is it needed?
Makes content panel's left margin consistent when tree panel state changes.

### How to test
Simply expand and contract the tree panel, and reload a page while in each state to verify the content area margin remains consistent.

### Related issue(s)/PR(s)
Resolves #16182

### DIFF
--- a/manager/assets/modext/core/modx.layout.js
+++ b/manager/assets/modext/core/modx.layout.js
@@ -10,11 +10,10 @@ Ext.apply(Ext, {
     isFirebug: (window.console && window.console.firebug)
 });
 
-MODx.Layout = function(config){
-    config = config || {};
+MODx.Layout = function(config = {}) {
     Ext.BLANK_IMAGE_URL = MODx.config.manager_url+'assets/ext3/resources/images/default/s.gif';
     Ext.Ajax.defaultHeaders = {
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
Since this is PR adds new public properties/functions we need to target 3.1

theboxer (approved):

theboxer:
I'd be fine with this in 3.0.x

PR #16407, opened 2023-03-07
Fix TV sorting

christianseel:
### What does it do?
Fix sorting of TVs in manager

### Why is it needed?
There is currently no user interface to manage ranks of `TemplateVarTemplate` (template!). Users can only manage the rank of the `modTemplateVar` itself. MODX should not sort by a value that a manager user can not manage.

### How to test
Specify the rank of TV in the TV edit view. Add more TVs to test. Verify sorting on the resource edit page in the TV tab.

### Related issue(s)/PR(s)
-

### DIFF
--- a/manager/controllers/default/resource/resource.class.php
+++ b/manager/controllers/default/resource/resource.class.php
@@ -354,7 +354,7 @@ abstract class ResourceManagerController extends modManagerController
                     $c->select($this->modx->getSelectColumns(modTemplateVarResource::class, 'TemplateVarResource', '', ['value']));
                 }
                 $c->select($this->modx->getSelectColumns(modTemplateVarTemplate::class, 'TemplateVarTemplate', '', ['rank']));
-                $c->sortby('cat_category,TemplateVarTemplate.rank,modTemplateVar.rank', 'ASC');
+                $c->sortby('cat_category,modTemplateVar.rank', 'ASC');
                 $tvs = $this->modx->getCollection(modTemplateVar::class, $c);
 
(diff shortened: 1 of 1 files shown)


christianseel:
@cla-bot check

christianseel:
Not sure whtats the issue with cla-bot, but CLA was signed in 2012.

opengeek:
> Not sure whtats the issue with cla-bot, but CLA was signed in 2012.

It seems you have two user accounts that both have @christianseel set as the github_username. The `seel_media` one appears first in the database and does not have the CLA signed.

Mark-H:
Isn't this managed when editing the template, on the TVs tab? 

Scherm­afbeelding 2023-03-07 om 17 23 35

Perhaps the better fix is adjust the grid when editing a TV to have the rank there, but I can imagine that's not there because you don't have the information of the rank relative to other TVs.  

Scherm­afbeelding 2023-03-07 om 17 25 04

smg6511:
> I can imagine that's not there because you don't have the information of the rank relative to other TVs

That, and there's no real meaning for a TV’s rank outside of a template. Also, the TV may be ranked differently depending on template and possible Form Customization rules.

christianseel:
> Isn't this managed when editing the template, on the TVs tab?
> 
> Scherm­afbeelding 2023-03-07 om 17 23 35

Gosh! I was also asking Ben and we both didn't find that screen! Ok, so thats the place where to manage the rank…

Maybe the rank on the "edit tv" screen should be removed because it has no effect? But anyway, this MR can be closed.

PR #16408, opened 2023-03-11
Add IDs to toolbar elements

Jako:
### What does it do?
Add IDs to toolbar elements in MODx.grid.SettingsGrid

### Why is it needed?
If toolbar elements i.e. have to be hidden in a class that extends MODx.grid.SettingsGrid, Ext.getCmp needs an ID to access the toolbar element.

In #16089 the ID was removed. This PR fixes that, but with an ID that is variably set with the Ext.id of the current MODx.grid.SettingsGrid. This way multiple MODx.grid.SettingsGrid will not interfere with each other by using the same ID multiple times.

### How to test
At the moment only in Agenda. There I extended the code to get the right toolbar elements.

before:
```
            afterrender: function () {
                var filterNamespace = Ext.getCmp('modx-filter-namespace');
                if (filterNamespace) {
                    filterNamespace.hide();
                }
            }
```

after:
```
            afterrender: function (cmp) {
                var filterNamespace = Ext.getCmp('modx-filter-namespace');
                filterNamespace = filterNamespace || Ext.getCmp(cmp.ident + '-filter-ns');
                if (filterNamespace) {
                    filterNamespace.hide();
                }
            }
```

### Related issue(s)/PR(s)
#16089

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.settings.js
+++ b/manager/assets/modext/widgets/core/modx.grid.settings.js
@@ -8,6 +8,7 @@
  */
 MODx.grid.SettingsGrid = function(config) {
     config = config || {};
+    this.ident = config.ident || 'modx-ssg-' + Ext.id();
     this.exp = new Ext.grid.RowExpander({
         tpl : new Ext.XTemplate(
             '<p class="desc">{[MODx.util.safeHtml(values.description_trans)]}</p>'
@@ -30,6 +31,7 @@ MODx.grid.SettingsGrid = function(config) {
     '->'
     ,{
         xtype: 'modx-combo-namespace'
+        ,id: this.ident + '-filter-ns'
         ,itemId: 'filter-ns'
         ,emptyText: _('namespace_filter')
         ,allowBlank: false
@@ -66,6 +68,7 @@ MODx.grid.SettingsGrid = function(config) {
         }
     },{
         xtype: 'modx-combo-area'
+        ,id: this.ident + '-filter-area'
         ,itemId: 'filter-area'
         ,emptyText: _('area_filter')
         ,value: MODx.request.area || null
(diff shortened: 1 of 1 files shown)


Mark-H (approved):

smg6511:
Guys - FYI, a number of filtering deficiencies are handled in #16369, albeit in a different way. (As I re-read your description, I see the issue has to do with extending the grid so I'll need to take a look at that.) Please note that the best practice is to avoid specifying ids for components unless absolutely necessary. There are other ways to access specific components that have an `itemId`, which is the safer way to go. 

@Jako  - I have Agenda installed in my dev install and will test to see what's happening there. What's the bare minimum I'd need to set up in Agenda to see the issue in action?

Jako:
@smg6511 You can also use CrossContextsSettings 1.2.4 for the check.

The CMP of both extras has a dedicated system settings grid for the component. In this grid the namespace filter combo is hidden with the code examples above.

This ID drop should have been documented somewhere. And it should not be changed in a patch release.

If I read the documentation right, I just have to use the following code to avoid my patch:

```
            afterrender: function (cmp) {
                var filterNamespace = Ext.getCmp('modx-filter-namespace');
                filterNamespace = filterNamespace || cmp.topToolbar.getComponent('filter-ns');
                if (filterNamespace) {
                    filterNamespace.hide();
                }
            }
```

Jako:
Using `cmp.topToolbar.getComponent('filter-ns');` works fine, so I will close this PR later on.

The '&tab' url property set by the new settings grid properties interferes with other extras, that don't have a tab handling or use the statCache.

smg6511:
> The '&tab' url property set by the new settings grid properties interferes with other extras

Could you be more specific about this? Although the current change was not done by me, my open PR that I referenced also uses the `tab` key in the URL. I'd like to assess whether that key should be changed.

Jako:
Change the area in CrossContextsSettings 1.2.4 'system settings' tab (cog icon) to 'system'. Open the 'context settings' tab. Reload the manager. The 'system settings' tab is opened. This should be the case with every extra using an extended core grid that remembers open tabs etc. in the url.

Jako:
Is it possible to reactivate the id in 3.0.4, deprecate it and remove it in 3.1?

PR #16414, opened 2023-04-04
Settings grid adjustments for Extras support

smg6511:
### What does it do?
Re-introduced original id property on settings grid filter combos and added a couple small changes already developed for the much larger #16369.

### Why is it needed?
Provides a quick solution to a couple problems caused by the unintentional early implementation of #16089 into the 3.0.3 release, namely those related to Extras that  extend the settings grid (specifically @Jako's Agenda and CrossContextsSettings).

### How to test

- Ensure that the ACL grids in the Usergroup Permissions area (5 grids) and the Settings grid behave as they currently do in 3.0.3.
- Test the two Extras mentioned above to verify they behave as intended.

### Related issue(s)/PR(s)
Fulfills request made in the closed #16408


JoshuaLuckers (approved):

Jako (approved):
Works well. Is it possible to deactivate the tab handling by option in an extended grid?

smg6511:
> Is it possible to deactivate the tab handling by option in an extended grid?

I suppose that could be done, I'll have to give it a look. Note that the only things happening with the tab handling are:
- when a filter is applied, the tab index is added to the URL for later retrieval,
- when a URL with a tab parameter is directly accessed, the tab specified will be activated (currently only for panels having a grid), and
- the filters for the grid within a panel being moved away from are always cleared. 

Is this undesirable in your case?

Jako:
Currently I am stumbling over the following problem: I save and retrieve the tabs in my extras with a stateful property and not with a url property. So when using the system settings grid and changing the area there for example, the url gets the tab property. If I switch to another tab and reload the manager, the system settings tab is opened. This is because the url property takes precedence over the stateful property and my extras do not reset the tab url property when I switch tabs.

smg6511:
@Jako - Let's do this: I'll create a way for custom (i.e., Extras) grid classes to opt out of the system's tab/grid filter tracking; but I'll do it in a separate PR since it solves a different issue and might be considered more of a feature than a fix.

PR #16417, opened 2023-04-08
Bug fix "Form Customization" (Constraint)

tanzirev:
When using comma delimited list of constraints, customizations for TVs doesn't work.

Example setup:
- create any TV and assign it to template
- create FC with constraint field = id and constraint = 1,2
- Change Label for pagetitle (or any other field)
- Change Label for the previously created TV
- Open resource 1 or 2, pagetitle will have the new label set from FC, but TV won't

original content
---
In the "Form Customization" tool, in the rule settings, in the "Constraint" field, you can specify multiple ids separated by commas. The rule worked for all fields except TV.

I copied the code from here:
https://github.com/modxcms/revolution/blob/v2.8.4-pl/core/model/modx/modmanagercontroller.class.php#L840

### DIFF
--- a/core/model/modx/modtemplatevar.class.php
+++ b/core/model/modx/modtemplatevar.class.php
@@ -621,10 +621,12 @@ class modTemplateVar extends modElement {
 
                     $constraintClass = $rule->get('constraint_class');
                     if (!empty($constraintClass)) {
-                        if (!($resource instanceof $constraintClass)) continue;
+                        if (empty($resource) || !($resource instanceof $constraintClass)) continue;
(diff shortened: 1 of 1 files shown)


smg6511:
First, thank you for submitting your first PR!

Please be more specific about 
- what you're entering into the Form Customization Constraints (field and value list)
- at least some of the changes your making to Regions or Template Variables, and 
- what you're expecting (but not seeing) when editing a Resource that the customization applies to. 

It's best to share as much information as you can to make your PRs easier to review ;-)

theboxer (approved):
works as described

PR #16418, opened 2023-04-10
Fixed steal processor 

Semdevmaster:
### What does it do?
Fix missing scriptProperties array

### DIFF
--- a/core/model/modx/processors/resource/locks/steal.class.php
+++ b/core/model/modx/processors/resource/locks/steal.class.php
@@ -20,6 +20,7 @@ class modResourceLocksStealProcessor extends modProcessor {
     }
 
     public function process() {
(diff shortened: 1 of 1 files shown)


Ibochkarev:
@cla-bot check

Semdevmaster:
@cla-bot check

JoshuaLuckers:
@Semdevmaster thanks for taking time to help make MODX better! For future reference, did you encounter a specific issue this might solve (except the logic issue in the code)?

Semdevmaster:
> 

During the data import process, I needed to intercept possible resource locks so as not to receive resource update errors. But this processor always returned an error. The problem turned out to be that the required array containing the input parameter was simply not declared in the processor. After adding the fix - everything worked as it should.

opengeek (approved):

PR #16421, opened 2023-04-12
New: Native DKIM capability for emails

### LINKED ISSUE #16396: Enable DKIM capabilities by default via settings
### Summary
Native DKIM capabilities in modMailer would be very useful. The functionality is already there but can't be used by default or with an extension like FormIt.

### Why is it needed?
More and more companies moving to DKIM and also using 2-factor authentication for email accounts, so even secure SMTP is not an option for those. SPF record is often not enough to be able to deliver email reliably. 

### Suggested solution(s)
MODx uses PHPMailer which is capable of sending emails using DKIM, even modMail and modPHPMailer implementation are capable of using them. It can be used from a snippet but it can't be used with in-built functionality or with an extra like FormIt. So a simple solution would be adding those DKIM options to the settings and also modifying modMailer::getDefaultAttributes() function so it would utilise those settings.

krisznet:
### What does it do?
Adding native DKIM capabilities to MODX without creating a custom email script. After setting up the correct values in system settings, all mail sent by MODx will be signed.

### Why is it needed?
More and more companies moving to DKIM and also using 2-factor authentication for email accounts, so even secure SMTP is not an option for those. SPF record is often not enough to be able to deliver email reliably.

### How to test
Regular mail and SMTP should work like before. If you add DKIM details in settings the Emailer should use it and the email should be signed with the correct key.

### Related issue(s)/PR(s)
Resolves #16396

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -952,7 +952,7 @@ $settings['mail_dkim_passphrase'] = $xpdo->newObject(modSystemSetting::class);
 $settings['mail_dkim_passphrase']->fromArray([
   'key' => 'mail_dkim_passphrase',
   'value' => '',
-  'xtype' => 'textfield',
+  'xtype' => 'text-password',
   'namespace' => 'core',
   'area' => 'mail',
   'editedon' => null,
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -902,6 +902,61 @@ $settings['mail_smtp_user']->fromArray([
   'area' => 'mail',
   'editedon' => null,
 ], '', true, true);
+
+$settings['mail_dkim_selector'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_selector']->fromArray([
+  'key' => 'mail_dkim_selector',
+  'value' => '',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_identity'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_identity']->fromArray([
+  'key' => 'mail_dkim_identity',
+  'value' => '',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_domain'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_domain']->fromArray([
+  'key' => 'mail_dkim_domain',
+  'value' => '',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_privatekeyfile'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_privatekeyfile']->fromArray([
+  'key' => 'mail_dkim_privatekeyfile',
+  'value' => '',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_privatekeystring'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_privatekeystring']->fromArray([
+  'key' => 'mail_dkim_privatekeystring',
+  'value' => '',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_passphrase'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_passphrase']->fromArray([
+  'key' => 'mail_dkim_passphrase',
+  'value' => '',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
 $settings['manager_date_format'] = $xpdo->newObject(modSystemSetting::class);
 $settings['manager_date_format']->fromArray([
   'key' => 'manager_date_format',
--- a/core/lexicon/ar/setting.inc.php
+++ b/core/lexicon/ar/setting.inc.php
@@ -401,24 +401,6 @@ $_lang['setting_mail_smtp_timeout_desc'] = 'ضبط مهلة المخدم SMTP م
 $_lang['setting_mail_smtp_user'] = 'مستخدم SMTP';
 $_lang['setting_mail_smtp_user_desc'] = 'المستخدم للمصادقة مقابله إلى SMTP.';
 
-$_lang['setting_mail_dkim_selector'] = 'DKIM Selector';
-$_lang['setting_mail_dkim_selector_desc'] = 'The DKIM domain selector where the public key stored.';
-
-$_lang['setting_mail_dkim_identity'] = 'DKIM Identity';
-$_lang['setting_mail_dkim_identity_desc'] = 'DKIM identity you\'re signing as - usually your From address';
-
-$_lang['setting_mail_dkim_domain'] = 'DKIM Domain';
-$_lang['setting_mail_dkim_domain_desc'] = 'DKIM signing domain name.';
-
-$_lang['setting_mail_dkim_privatekeyfile'] = 'DKIM Private key file';
-$_lang['setting_mail_dkim_privatekeyfile_desc'] = 'DKIM private key file path. You can use DKIM Private key string instead of this.';
-
-$_lang['setting_mail_dkim_privatekeystring'] = 'DKIM Private key string';
-$_lang['setting_mail_dkim_privatekeystring_desc'] = 'Takes precedence over DKIM Private key file.';
-
-$_lang['setting_mail_dkim_passphrase'] = 'DKIM Passphrase';
-$_lang['setting_mail_dkim_passphrase_desc'] = 'Used only if your key is encrypted.';
-
 $_lang['setting_main_nav_parent'] = 'أصل القائمة الرئيسية';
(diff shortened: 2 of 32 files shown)


zaigham:
@krisznet You just need to commit translations in `core/lexicon/en/*` folder, the translations will be handled via Crowdin.

Bruno17:
Had that same issue and got this answers on slack:
https://modxcommunity.slack.com/archives/C042Q4YS7/p1681303237481139?thread_ts=1678262173.500949&cid=C042Q4YS7
I could solve it then by setting up SMTP and DNS correctly

krisznet:
> Had that same issue and got this answers on slack: https://modxcommunity.slack.com/archives/C042Q4YS7/p1681303237481139?thread_ts=1678262173.500949&cid=C042Q4YS7 I could solve it then by setting up SMTP and DNS correctly

@Bruno17 The problem is that some company has strict email policy and 2FA for emails, so the SMTP not an option. And it is 2023, DKIM and DMARC is essential.

Omeryl:
Huh? In what case is the company completely fine with you setting a random DKIM key in the DNS, but won't provide you with SMTP credentials for an account authorized to send for the domain? Storing these in a web application is a terrifying prospect, the mail server is a much better place to be signing emails. Furthermore, if an email fails to send one time with mail(), it won't be retried. Mail servers retry every so often on an exponential back off. This is better for the internet, sometimes receiving mail servers go down (even Google!).

krisznet:
> Huh? In what case is the company completely fine with you setting a random DKIM key in the DNS, but won't provide you with SMTP credentials for an account authorized to send for the domain? Storing these in a web application is a terrifying prospect, the mail server is a much better place to be signing emails. Furthermore, if an email fails to send one time with mail(), it won't be retried. Mail servers retry every so often on an exponential back off. This is better for the internet, sometimes receiving mail servers go down (even Google!).

@Omeryl I agree that SMTP is more reliable but sometimes you can't use it because of the technical limitations. If 2FA is required for emails SMTP is not an option. You can set up DKIM selectors, so you can just create a key solely for this purpose. If it is compromised, you can delete it and set up a new one. The private key can be stored in a file or DB and can be encrypted. If your database and/or filesystem are compromised, you are already in big trouble.

JoshuaLuckers:
@opengeek @Mark-H i would love to get your opinion on this, to me it seems like a great addition even if it’s not the preferred solution (to have it as option in the CMS).

opengeek (changes requested):
All lexicon entries other than the en/ sources must be removed from this PR.

I'm also not clear on how this works by just setting some options.

Bruno17:
my idea was to add a plugin event to modPHPMailer::send 
'onMailerSend' or something like that, where we could hook in and add for example this DKIM stuff, if needed.

opengeek:
> my idea was to add a plugin event to modPHPMailer::send 'onMailerSend' or something like that, where we could hook in and add for example this DKIM stuff, if needed.

I guess my question is why wouldn't this be settings of an extra that installs this plugin?

Bruno17:
> > my idea was to add a plugin event to modPHPMailer::send 'onMailerSend' or something like that, where we could hook in and add for example this DKIM stuff, if needed.
> 
> I guess my question is why wouldn't this be settings of an extra that installs this plugin?

Yeah. But thats currently not possible, because there isn't an event like that.

opengeek:
> > > my idea was to add a plugin event to modPHPMailer::send 'onMailerSend' or something like that, where we could hook in and add for example this DKIM stuff, if needed.
> > 
> > 
> > I guess my question is why wouldn't this be settings of an extra that installs this plugin?
> 
> Yeah. But thats currently not possible, because there isn't an event like that.

But this PR does not add an event. I guess I'm still confused how this is supposed to work as is in the PR.

Bruno17:
> > > > my idea was to add a plugin event to modPHPMailer::send 'onMailerSend' or something like that, where we could hook in and add for example this DKIM stuff, if needed.
> > > 
> > > 
> > > I guess my question is why wouldn't this be settings of an extra that installs this plugin?
> > 
> > 
> > Yeah. But thats currently not possible, because there isn't an event like that.
> 
> But this PR does not add an event. I guess I'm still confused how this is supposed to work as is in the PR.

oh.. its not my PR. The Opener has added his new systemsettings as default attributes into modMail
https://github.com/modxcms/revolution/pull/16421/files#diff-919cb7414fac156184fcd82007638be9e5946883c0348c3c6957ea355060d390

opengeek:
From what I can see, all of these "settings" are part of modPHPMailer. modMail is an interface that is intended to define common settings for any mailer implementation. In order to set the modPHPMailer settings, those entries can be manually added to system settings when needed, and they will just work. So I'm not sure of the actual necessity of this core change.

krisznet:
> > > > > my idea was to add a plugin event to modPHPMailer::send 'onMailerSend' or something like that, where we could hook in and add for example this DKIM stuff, if needed.
> > > > 
> > > > 
> > > > I guess my question is why wouldn't this be settings of an extra that installs this plugin?
> > > 
> > > 
> > > Yeah. But thats currently not possible, because there isn't an event like that.
> > 
> > 
> > But this PR does not add an event. I guess I'm still confused how this is supposed to work as is in the PR.
> 
> oh.. its not my PR. The Opener has added his new systemsettings as default attributes into modMail https://github.com/modxcms/revolution/pull/16421/files#diff-919cb7414fac156184fcd82007638be9e5946883c0348c3c6957ea355060d390

Hey guys, if you read the original PR and the Issue it is described clearly. DKIM capability is already in modPHPMailer but you can't use it through modMail with any existing extension like FormIt. I added the settings and also modified modMail, so it uses those settings. That is it. It is much easier than creating a plugin and it is not much code and also not a difficult change.

krisznet:
> From what I can see, all of these "settings" are part of modPHPMailer. modMail is an interface that is intended to define common settings for any mailer implementation. In order to set the modPHPMailer settings, those entries can be manually added to system settings when needed, and they will just work. So I'm not sure of the actual necessity of this core change.

@opengeek You are right but not 100% right. You can use DKIM if you write a custom snippet but can't use it with emails sent by modX itself or an extra like FormIt.

opengeek:
> > From what I can see, all of these "settings" are part of modPHPMailer. modMail is an interface that is intended to define common settings for any mailer implementation. In order to set the modPHPMailer settings, those entries can be manually added to system settings when needed, and they will just work. So I'm not sure of the actual necessity of this core change.
> 
> @opengeek You are right but not 100% right. You can use DKIM if you write a custom snippet but can't use it with emails sent by modX itself or an extra like FormIt.

I see now. If the default attributes are not set, the initialization does not look for those attributes in the settings. Let me think on this. It would, in my opinion, be better if we modified the initialization of modPHPMailer to look for attributes from MODX settings in all cases. Otherwise, we will run into other options in this mailer—or in any mailer implementation—that would have to manually be added to the modMail interface. And this is not really the job of an interface.

krisznet:
> > > From what I can see, all of these "settings" are part of modPHPMailer. modMail is an interface that is intended to define common settings for any mailer implementation. In order to set the modPHPMailer settings, those entries can be manually added to system settings when needed, and they will just work. So I'm not sure of the actual necessity of this core change.
> > 
> > 
> > @opengeek You are right but not 100% right. You can use DKIM if you write a custom snippet but can't use it with emails sent by modX itself or an extra like FormIt.
> 
> I see now. If the default attributes are not set, the initialization does not look for those attributes in the settings. Let me think on this. It would, in my opinion, be better if we modified the initialization of modPHPMailer to look for attributes from MODX settings in all cases. Otherwise, we will run into other options in this mailer—or in any mailer implementation—that would have to manually be added to the modMail interface. And this is not really the job of an interface.

All other settings are in the modMail interface as well like SMTP settings and charset, etc. That is why I put it there.

krisznet:
> All lexicon entries other than the en/ sources must be removed from this PR.
> 
> I'm also not clear on how this works by just setting some options.

@opengeek I removed the unwanted lexicon changes.

opengeek:
@krisznet — can you sign the CLA and make sure you add your GitHub username when doing so?

krisznet:
> @krisznet — can you sign the CLA and make sure you add your GitHub username when doing so?

@opengeek I did it yesterday.

opengeek:
@cla-bot check

opengeek (dismissed):

_build/data/transport.core.system_settings.php
```diff
@@ -902,6 +902,61 @@
...
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_privatekeystring'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_privatekeystring']->fromArray([
+  'key' => 'mail_dkim_privatekeystring',
+  'value' => '',
+  'xtype' => 'textfield',
```
JoshuaLuckers, line 946:
```suggestion
  'xtype' => 'text-password',
```

I suggest converting the input field into a password field.

_build/data/transport.core.system_settings.php
```diff
@@ -902,6 +902,61 @@
...
+  'namespace' => 'core',
+  'area' => 'mail',
+  'editedon' => null,
+], '', true, true);
+$settings['mail_dkim_passphrase'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_dkim_passphrase']->fromArray([
+  'key' => 'mail_dkim_passphrase',
+  'value' => '',
+  'xtype' => 'textfield',
```
JoshuaLuckers:
```suggestion
  'xtype' => 'text-password',
```

I suggest converting the input field into a password field.

JoshuaLuckers (changes requested):
@Mark-H @opengeek `mail_dkim_privatekeyfile`, `mail_dkim_privatekeystring` and `mail_dkim_passphrase` should also be added to the list of things we need to unset in the following two places right?

https://github.com/modxcms/revolution/blob/39a5672363fea11e982b7887372b43ebbbf47f3f/core/src/Revolution/Processors/System/ConfigJs.php#L115

https://github.com/modxcms/revolution/blob/39a5672363fea11e982b7887372b43ebbbf47f3f/core/src/Revolution/modX.php#L583

Mark-H (approved):

PR #16422, opened 2023-04-12
Fix undefined global variable $_SESSION in a CLI connector

Jako:
### What does it do?
Use an empty array, if the global $_SESSION variable does not exist

### Why is it needed?
Fix the warning `(ERROR @ /www/core/src/Revolution/modConnectorRequest.php : 41) PHP warning: Undefined global variable $_SESSION`, when a connector is called via CLI. The issue occurred first with the more strict checks of PHP 8.

### How to test
It currently occurs with CronManager, when `assets/components/cronmanager/cron.php` is executed by CLI. Agenda has the same issue.

### Related issue(s)/PR(s)
None known.

### DIFF
--- a/core/src/Revolution/modConnectorRequest.php
+++ b/core/src/Revolution/modConnectorRequest.php
@@ -38,7 +38,7 @@ class modConnectorRequest extends modManagerRequest
         if ($this->modx && is_object($this->modx->context) && $this->modx->context instanceof modContext) {
             $ctx = $this->modx->context->get('key');
             if (!empty($ctx) && $ctx == 'mgr') {
-                $ml = $this->modx->getOption('manager_language', $_SESSION,
+                $ml = $this->modx->getOption('manager_language', $_SESSION ?? [],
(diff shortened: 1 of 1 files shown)


MaticSulc:
@Jako thank you for all the help.
I've also reproduced this with [CronManager](https://github.com/Jako/CronManager/)

JoshuaLuckers (approved):

Ibochkarev (approved):

opengeek:
Thanks for this, but I wonder why a connector is being called via CLI. Connectors are meant to provide web access to processors, so the fact CLI scripts are calling them is a little confusing.

opengeek:
Also, is this not an issue in 2.x?

Jako:
I have seen this CLI usage of a connector in a lot extras. The code for CLI and connector usage can be the same.

It should also be relevant in 2.x and it occurs only with the more strict checks of PHP 8 in both versions (added this above).

opengeek (approved):

PR #16423, opened 2023-04-19
Add triggers for OnUserBeforeRemoveFromGroup and OnUserRemoveFromGroup

theboxer:
### What does it do?
Adds missing triggers for `OnUserBeforeRemoveFromGroup` and `OnUserRemoveFromGroup` events.

### Why is it needed?
Plugins may want them :)


opengeek:
Can you clarify? Are these new events or events that were added to 2.x after 3.x was branched and never ported?

theboxer:
I believe they were just not properly implemented during the processors rewrite for 3.x.

https://github.com/modxcms/revolution/blame/2.x/core/model/modx/processors/security/group/removeuser.php#L69

PR #16430, opened 2023-05-09
Send all args to provider download endpoints

opengeek:
### What does it do?
Sends all of the arguments sent to other provider endpoints to the download endpoint.

### Why is it needed?
The UUID and other important information were not sent to provider download endpoints (seemingly the only endpoint not receiving these arguments). Having that information can be useful to a package provider implementation.

### How to test
Provider implementation owners can validate that `uuid` and the other arguments sent to most provider endpoints are being received on download endpoint requests.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/core/src/Revolution/Transport/modTransportProvider.php
+++ b/core/src/Revolution/Transport/modTransportProvider.php
@@ -414,10 +414,9 @@ class modTransportProvider extends xPDOSimpleObject
 
         $uri = $location;
         $uri .= (strpos($uri, '?') > 0) ? '&' : '?';
-        $uri .= http_build_query([
-            'revolution_version' => $this->arg('revolution_version', $this->args($args)),
+        $uri .= http_build_query(array_merge($this->args($args), [
             'getUrl' => true,
-        ]);
+        ]));
         $request = $requestFactory->createRequest('GET', $uri)
             ->withHeader('Accept', 'text/plain');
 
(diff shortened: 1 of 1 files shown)


opengeek:
This will need to be ported to the 2.x and 3.x branches, as well.

Mark-H:
IIRC the package listing that provides the download url already provides the parameters necessary for the actual download, or at least it used to back in the day. Adding this in could potentially break other package providers because parameters are then provided twice.

opengeek:
> IIRC the package listing that provides the download url already provides the parameters necessary for the actual download, or at least it used to back in the day. Adding this in could potentially break other package providers because parameters are then provided twice.

The actual download URL is not provided in the listing. We need the UUID in the download endpoint in order to properly track download counts. Using IP address for this is problematic for various reasons. This is the only endpoint that does not receive all of the arguments, and I don't see why it would not. I tested this with the modx.com provider and there is no issue at all with providing the additional arguments before making any changes to make use of them. How would this break a provider?

opengeek:
> IIRC the package listing that provides the download url already provides the parameters necessary for the actual download, or at least it used to back in the day. Adding this in could potentially break other package providers because parameters are then provided twice.

I see what you mean now. The request that builds the listing receives the parameters and I can provide them to the download URL when building the list. That will work retroactively and not require a patch.

PR #16431, opened 2023-05-15
Add namespaces partition to cache refresh process

opengeek:
### What does it do?
Ensures that the namespaces cache partition is refreshed when modCacheManager::refresh() is called without specific partitions and that the ClearCache processor includes the namespaces partition.

### Why is it needed?
The namespaces cache partition was not being refreshed which caused problems when injecting teleport snapshots or other processes that affect namespaces when xPDO::OPT_SETUP is enabled.

### How to test
Inject a snapshot with teleport that includes various namespaces in 3.x that add their component src directories to the autoloader. Ensure that those components work after injection without manually deleting the cache directory or reinstalling MODX.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/core/src/Revolution/Processors/System/ClearCache.php
+++ b/core/src/Revolution/Processors/System/ClearCache.php
@@ -133,6 +133,7 @@ class ClearCache extends Processor
 
         $partitions['system_settings'] = [];
         $partitions['context_settings'] = ['contexts' => $contextKeys];
+        $partitions['namespaces'] = [];
 
         if ($this->modx->getOption('cache_db', null, false)) {
             $partitions['db'] = [];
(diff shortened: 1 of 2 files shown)


JoshuaLuckers:
> Inject a snapshot with teleport that includes various namespaces in 3.x that add their component src directories to the autoloader.

@opengeek would it be possible for you to share a test snapshot I can use to inject and test this?

opengeek:
> @opengeek would it be possible for you to share a test snapshot I can use to inject and test this?

I don't have one that can reproduce the issue, but it seems to have solved the problem @jaygilmore and @rthrash were experiencing with one in MODX Cloud.

Mark-H (approved):

Ibochkarev (approved):

PR #16435, opened 2023-06-02
Update subMenuBuilder so "installer-entry" is always at the top [v3.x]

### LINKED ISSUE #11694: Make Installer always top menu item

sebastian-marinescu:
### What does it do?
Due to the default-value of menuindex in menu-entries of newly installed extras being most of the time "0", the installer entry will jump up and/or down after an extra gets installed/updated.

Therefore it should at least be fixed at the top of the dropdown-menu, if the User wishes.

### Related issue(s)/PR(s)
Fixes https://github.com/modxcms/revolution/issues/11694 pragmatically
Related PR for v2.x #14482

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1073,6 +1073,15 @@ $settings['modx_charset']->fromArray([
   'area' => 'language',
   'editedon' => null,
 ], '', true, true);
+$settings['package_installer_at_top'] = $xpdo->newObject('modSystemSetting');
+$settings['package_installer_at_top']->fromArray(array (
+  'key' => 'package_installer_at_top',
+  'value' => true,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'manager',
+  'editedon' => null,
+), '', true, true);
 $settings['principal_targets'] = $xpdo->newObject(modSystemSetting::class);
 $settings['principal_targets']->fromArray([
   'key' => 'principal_targets',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1073,15 +1073,15 @@ $settings['modx_charset']->fromArray([
   'area' => 'language',
   'editedon' => null,
 ], '', true, true);
-$settings['package_installer_at_top'] = $xpdo->newObject('modSystemSetting');
-$settings['package_installer_at_top']->fromArray(array (
+$settings['package_installer_at_top'] = $xpdo->newObject(modSystemSetting::class);
+$settings['package_installer_at_top']->fromArray([
   'key' => 'package_installer_at_top',
   'value' => true,
   'xtype' => 'combo-boolean',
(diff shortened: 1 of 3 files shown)


core/lexicon/en/setting.inc.php
```diff
@@ -471,6 +471,9 @@
 $_lang['setting_new_folder_permissions'] = 'New Folder Permissions';
 $_lang['setting_new_folder_permissions_desc'] = 'When creating a new folder in the File Manager, the File Manager will attempt to change the folder permissions to those entered in this setting. This may not work on some setups, such as IIS, in which case you will need to manually change the permissions.';
 
+$_lang['setting_package_installer_at_top'] = 'Pin Package-Installer at top';
+$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the package-installer menu-entry inside the Extras-menu will be pinned at the top - regardless of its "menuindex"-attribute. If disabled, the package-installer will be positioned according to its menuindex.';
```
smg6511: There are a lot of unnecessary dashes in the description. Also suggesting other tweaks to keep it concise.
```suggestion
$_lang['setting_package_installer_at_top_desc'] = 'If enabled, the Installer entry will be pinned to the top of the Extras menu. Otherwise it will be positioned according to its menuindex.';
```

smg6511 (changes requested):
This needs a couple changes to work (the second one to the lexicon is arguably optional, but makes a necessary improvement IMO). Once these changes are made, it's ready to approve.

_build/data/transport.core.system_settings.php
```diff
@@ -1073,6 +1073,15 @@
...
 ], '', true, true);
+$settings['package_installer_at_top'] = $xpdo->newObject('modSystemSetting');
+$settings['package_installer_at_top']->fromArray(array (
+  'key' => 'package_installer_at_top',
+  'value' => true,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'manager',
+  'editedon' => null,
+), '', true, true);
```
smg6511: For 3.x, the system setting class needs to be specified in this way, as otherwise a fatal error is thrown when running `_build/core.transport.php`. Also, we no longer use the `array` keyword and prefer the short (bracket) syntax when creating arrays.
```suggestion
$settings['package_installer_at_top'] = $xpdo->newObject(modSystemSetting::class);
$settings['package_installer_at_top']->fromArray([
  'key' => 'package_installer_at_top',
  'value' => true,
  'xtype' => 'combo-boolean',
  'namespace' => 'core',
  'area' => 'manager',
  'editedon' => null,
], '', true, true);
```

sebastian-marinescu:
Nice - thank you very much @smg6511 !

smg6511 (approved):
Ok, looks great and works as expected!

Mark-H (approved):

PR #16436, opened 2023-06-06
Fix PHP warning: Constant IMG_AVIF already defined

Maef:
### What does it do?
Fixes unnecessary PHP warning entries in the Modx  Error log about an already defined IMG_AVIF constant. 

### Why is it needed?
Some phpThumb operations on images produce this line in the error log. Let's keep the error log clean(er).

### DIFF
--- a/core/model/phpthumb/modphpthumb.class.php
+++ b/core/model/phpthumb/modphpthumb.class.php
@@ -41,8 +41,10 @@ class modPhpThumb extends phpThumb
         }
 
         if (version_compare(PHP_VERSION, '8.1.0', '<=')) {
-            // The constant IMG_AVIF is available as of PHP 8.1.
-            define('IMG_AVIF', 256);
+            // The constant IMG_AVIF is available as of PHP 8.1.            
+            if (!defined('IMG_AVIF')) {
+                define('IMG_AVIF', 256);
+            }
         }
 
         parent::__construct();


JoshuaLuckers (approved):

smg6511:
Just a couple questions on this:
1. Can you be more specific about the condition that triggers the errors for you? Although it's reasonable to approve this by just reading the code and concluding it's a logical move, it'd be nice to make sure this isn't just patching a symptom of a root cause that might need to be fixed.
2. Assuming this fix is indeed needed, should the same condition be added to the IMG_WEBP definition?

JoshuaLuckers:
> 2. Assuming this fix is indeed needed, should the same condition be added to the IMG_WEBP definition?

You mean this? https://github.com/modxcms/revolution/commit/5e45c556b74551b3b4a03f3db4c575c187fb1f5a

smg6511:
Yes, but wrapping that with the conditional `if (!defined('IMG_WEBP')) ...`, which is the only thing being done in this PR for IMG_AVIF.

Maef:
> 1. Can you be more specific about the condition that triggers the errors for you? Although it's reasonable to approve this by just reading the code and concluding it's a logical move, it'd be nice to make sure this isn't just patching a symptom of a root cause that might need to be fixed.

Did some digging ... hopefully the correct conclusion.

Traced it back to a commit by myself on [phpThumb 1.7.18-202208061319](https://github.com/modxcms/revolution/commit/f08c189d05053d5a6cd10b6b0f4028434147071b) in November 2022. The constant was added in core/model/phpthumb/modphpthumb.class.php

The next [phpThumb v1.7.19-202210110924 update](https://github.com/modxcms/revolution/commit/d750869c9769a1573f6a9c91207182ac391566c5) in December 2022 includes logic  [by the author of phpThumb](https://github.com/JamesHeinrich/phpThumb/commit/9914236397b1796e74e0fd681e13dfd7fd78467d) to define the constant if it was undefined. 

You can simulate the warning by uploading an image and hovering over the image in the tree / filebrowser to get the error log. (with a PHP < 8.1)

smg6511:
In light of what you found, would it be reasonable to simply remove the const definitions for both `IMG_WEBP` and `IMG_AVIF` given that the currently-integrated version of the phpthumb lib defines them already?

smg6511:
@Maef @JoshuaLuckers - Your thoughts on removing the const definitions (as mentioned above)?

Maef:
Only tested `IMG_AVIF`

Keep the test or remove the constant. Either way is fine with me. It's up to you guys.

Maef:
@opengeek @Mark-H Your view? Would be nice if the next Modx 2.8.x release would not have the unnecessary PHP warnings.

Mark-H:
If phpthumb already does it, we don't need the constants anymore and can remove them. Do we still need to update phpthumb though?

smg6511:
@Maef @Mark-H - Actually, the merged #16456 makes this PR irrelevant, as AVIF is defined in that updated version of phpthumb. Shall we close this?

Note also, for 3.x, composer already pulls down a recent enough version of phpthumb to _not_ need to port a fix there.

opengeek:
@smg6511 @Maef — should this be closed?

smg6511:
Yes, I think this can be closed.

Maef:
Yes, it's better close this.

PR #16438, opened 2023-06-09
S3 Speed Up

matdave:
### What does it do?
Removed the functions that process each file individually when listing files, e.g. lastModified, fileSize, mimeType, e.g. 

### Why is it needed?
Currently, the S3 bucket will time out as you grow the contents of your folders. This dramatically increases the responses and has been tested on folders containing hundreds of files. 

### How to test
The best way to test is with an S3 bucket with numerous sample files. The file size doesn't matter, just a large number. As the number grows, the response time increases. 

### Related issue(s)/PR(s)
No related public issues, just personal experience

### DIFF
--- a/core/src/Revolution/Sources/modS3MediaSource.php
+++ b/core/src/Revolution/Sources/modS3MediaSource.php
@@ -677,7 +677,9 @@ class modS3MediaSource extends modMediaSource
             'xml',
         ];
         foreach ($binary_extensions as $a) {
-            if (stripos($file,$a) !== false) return true;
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
I don't have access to an S3 bucket, is there another way I can test this?

matdave:
@theboxer can you review this

matdave:
@JoshuaLuckers I believe Amazon offers a free tier https://aws.amazon.com/free/storage/s3/

theboxer (approved):

PR #16440, opened 2023-06-13
Fix deprecation of dynamic properties

Jako:
### What does it do?
Define magic method __set and __get in the class to avoid the deprecation message. See: https://php.watch/versions/8.2/dynamic-properties-deprecated#__get-__set

There is also a fix included for an uncaught error: `Cannot assign by reference to overloaded object in /core/src/Revolution/Hashing/modHashing.php:107`.

### Why is it needed?
Avoid the message `Creation of dynamic property MODX\Revolution\Hashing\modHashing::$native is deprecated in core/src/Revolution/Hashing/modHashing.php on line 107`

### How to test
Those messages seem to occur during installing MODX 3.0.3 on PHP 8.2. See the screenshot from Slack:

image

There are other similar issues xPDO and modAccessibleObject that screenshot.

### Related issue(s)/PR(s)
None known

### DIFF
--- a/core/src/Revolution/Hashing/modHashing.php
+++ b/core/src/Revolution/Hashing/modHashing.php
@@ -19,6 +19,7 @@ use xPDO\xPDO;
  *
  * @package MODX\Revolution\Hashing
  */
+#[\AllowDynamicProperties]
 class modHashing
 {
     /**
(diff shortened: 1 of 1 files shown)


JoshuaLuckers:
@Jako @opengeek 
> In PHP 9.0, dynamic properties will result in a fatal error.

opengeek:
> @Jako @opengeek
> 
> > In PHP 9.0, dynamic properties will result in a fatal error.

Not on classes that have that attribute they won't.

Mark-H (approved):

PR #16447, opened 2023-06-18
Add reloading of lexicon entries for any namespace

Ruslan-Aleev:
### What does it do?
Add reloading of lexicon entries for any namespace:

- Move the button up, it's was almost invisible.
- Added a pop-up window with confirmation, where the namespace, topic and language are indicated.
- The name of the strings is now returned to the console, and not just their amount.

lexicons_reload

p.s. In a good way, this PR should be improved and merged after this (https://github.com/modxcms/revolution/pull/15942), although the current version is also working.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/16333
https://github.com/modxcms/revolution/issues/14442


Ruslan-Aleev:
Guys, some time ago I deleted several PRs out of emotion, thereby devaluing not only my work, but also the work of the community.
@JoshuaLuckers @Ibochkarev @Jako @Mark-H @smg6511  I'm sorry, I'll try to restore these PRs soon so that the community's work is not lost :)

smg6511:
@Ruslan-Aleev - No worries, just glad to see you're back!

Jako:
@Ruslan-Aleev There is no need to excuse. I understood your reaction at the end of January 2023.

Mark-H (dismissed):

core/lexicon/en/lexicon.inc.php
```diff
@@ -25,7 +25,8 @@
...
 $_lang['lexicon_management'] = 'Lexicon Management';
 $_lang['lexicon_management_desc'] = 'Here you can override any Lexicon Entries across Namespaces and their respective Topics. Just double-click on any Entry\'s value to override it. To add new Lexicon Entries or Topics, just create their Topic files in their Namespace\'s respective directory.<br /><b>Each Entry will be available via the [[%name]] placeholder.</b>';
-$_lang['lexicon_rlfb_msg'] = 'Successfully regenerated [[+num]] strings.';
-$_lang['reload_from_base'] = 'Revert All Core Entries';
-$_lang['reload_success'] = 'Successfully reloaded [[+total]] strings.';
+$_lang['lexicon_revert'] = 'Revert Entries';
+$_lang['lexicon_revert_confirm'] = 'Are you sure you want to revert the "[[+namespace]]" namespace > "[[+topic]]" topic > "[[+language]]" lexicon to its default strings?';
+$_lang['lexicon_revert_success'] = 'Reverted [[+total]] total string(s): [[+names]]';
+$_lang['lexicon_revert_error'] = 'No lexicon entries found to revert.';
```
smg6511, line 31: It's not really an error, so consider this key name change:
```suggestion
$_lang['lexicon_revert_none_found'] = 'No lexicon entries found to revert.';
```

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +42,54 @@ public function getLanguageTopics()
...
-        $corePath = $this->modx->getOption('core_path', null, MODX_CORE_PATH);
+        $namespace = $this->getProperty('namespace');
+        $topic = $this->getProperty('lexiconTopic');
+        $language = $this->getProperty('language');
+
+        $namespaceObj = $this->modx->getObject(modNamespace::class, $namespace);
+
+        if (empty($namespace) || empty($topic) || empty($language) || empty($namespaceObj)) {
+            $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_error'));
+            $this->modx->log(modX::LOG_LEVEL_INFO, 'COMPLETED');
```
smg6511, line 53: I know you're following what was done later in this class, but I get the feeling this was not meant to make it into production code. If this logging does have some use, the content should point to a lexicon and not be hard-coded as “COMPLETED.” My guess is this log and the one later on should be removed.
```suggestion

```

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
```diff
@@ -41,42 +42,54 @@ public function getLanguageTopics()
...
             }
         }
 
-        $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('reload_success', ['total' => $i]));
+        if ($i) {
+            $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_success', ['total' => $i, 'names' => implode(', ', $names)]));
+        } else {
+            $this->modx->log(modX::LOG_LEVEL_WARN, $this->modx->lexicon('lexicon_revert_error'));
+        }
         $this->modx->log(modX::LOG_LEVEL_INFO, 'COMPLETED');
```
smg6511, line 95: Same comment as above re this log line.
```suggestion

```

manager/assets/modext/workspace/lexicon/lexicon.grid.js
```diff
@@ -248,30 +247,46 @@ Ext.extend(MODx.grid.Lexicon,MODx.grid.Grid,{
...
+            topic = tb.getComponent('topic').getValue(),
+            language = tb.getComponent('language').getValue(),
+            registryTopic = '/workspace/lexicon/reload/';
 
-        this.console.on('complete',function(){
-            this.refresh();
-        },this);
-        this.console.show(Ext.getBody());
+        MODx.msg.confirm({
+            text: _('lexicon_revert_confirm', {
```
smg6511, line 265: Consider changing the title for this window, as it's not really a warning. If you elect to take this suggestion, you may even want to create a new `confirm_action` entry that reads “Confirm Action” to make the title even more clear.
```suggestion
            title: _('confirm')
            ,text: _('lexicon_revert_confirm', {
```

smg6511 (commented):
This works well and is a good change to the UI. I have a few suggestions (above) I'd like you to consider before approving the PR, but none are deal-breakers.

core/src/Revolution/Processors/Workspace/Lexicon/ReloadFromBase.php
theboxer: The `COMPLETED` log message tells the MODx.console that we're done

smg6511: While that may be the case, shouldn't it _not_ be hard-coded in English?

theboxer: well it only works in english, it could be a constant somewhere, but not a scope of this PR to change how it works

smg6511: A couple things, and I could be wrong on this (and in any case it's not a big deal at all), but...
- I believe this functionality works for all languages
- Looking at the entire file, I just don't see how this info-level log provides _any_ useful information

Again, it's nothing to haggle about, I just figure if there's an opportunity to clean up something in the process why not? ;-)

BTW @theboxer - Really glad to see you digging into some of these PRs lately ... we need all the help we can get!

opengeek: @smg6511 — it is just the way the console works. It has to get a "COMPLETED" log message—exactly that string—to know the logging should be returned back away from the console (i.e. to whatever log settings are configured in the MODX instance).

theboxer: https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/Processors/System/Console.php#L86

smg6511: Thanks for that clarification! Now I get why John said it's out of scope to change how it works. (Ya gotta admit though, that's a pretty odd and un-intuitive way to communicate with the console component.)

opengeek: You will get no disagreement from us. 😉

smg6511: Quick question: Has marking a conversation as "resolved" changed recently? I'd like to mark this and my other comment on this subject, but see no way to do it now.

opengeek: I still see a "Resolve conversation" button at the end of each review comment thread. Are you not seeing this?

smg6511: Nope, not seeing the button.

manager/assets/modext/workspace/lexicon/lexicon.grid.js
Ruslan-Aleev: There is a question in `lexicon_revert_confirm`, so no additional title is required.
Moreover, I did not find a single window in MODX with `title: _('confirm')`, and `title: _('warning')` is used by default, this is enough.

core/lexicon/en/lexicon.inc.php
Ruslan-Aleev: I don't think it makes much sense. There is a pair `error \ success`, everything is clear.
Otherwise, we will have to change `lexicon_revert_error` in `lexicon_revert_none_found`, and `lexicon_revert_success` then in `lexicon_revert_all_found` =) ? Sophistry...

Mark-H:
After merging the grid update by @smg6511, this is now conflicting. I tried to resolve them, unsuccessfully. Would you be able to rebase?

Ruslan-Aleev:
@Mark-H Done, check.

Mark-H (approved):

PR #16449, opened 2023-06-18
Added names of elements in context menu

Ruslan-Aleev:
### What does it do?
Added names of elements in tree. Now it takes less clicks to find out the key / id of an element in the tree + uniformity in the interface.

1) Added a source name to context menu.

Before:
src_tree_1

After:
src_tree_name

2) Added names of elements (directory and file) to their context menu. Now it is clear for which directory / file the context menu is called.

Before:
dir_tree_1
file_tree_1

After:
dir_tree_2
file_tree_2

### Why is it needed?
UX improvements

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/15922/


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -2087,13 +2087,19 @@ protected function getExtJSDirClasses()
      *
      * @return array
      */
-    protected function getListDirContextMenu()
+    protected function getListDirContextMenu($dirName)
```
Mark-H: Any chance we can accomplish this without changing the function signature? As it's not a private method, this can cause issues in extended classes.

Ruslan-Aleev: In my opinion, it is impossible to do without changing the signature.
Alternatively, we can set a default value:

```
protected function getListDirContextMenu($dirName = null) {
    ...

    $menu = [];
    if (isset($dirName)) {
        $menu[] = [
            'text' => '<b>' . htmlentities($dirName) . '</b>',
            'handler' => '',
            'header' => true,
        ];
        $menu[] = '-';
    }
    ...

    return $menu;
}
```

Mark-H: What about adding it client-side, like you did in modx.tree.directory.js?

Ruslan-Aleev: You are right, I found a simpler solution, and without editing the signature.

Mark-H (approved):

Ibochkarev (approved):

PR #16451, opened 2023-06-20
Changed the order of variables in config.inc.php

Ruslan-Aleev:
### What does it do?
Changed the order of variables in config.inc.php:

1) Changed the order of db connection variables.
For example, `$dbase` and `$database_user`, since the fields often have the same values and it will be easier to edit them visually.

3) Changed the order of path constant.
Sorted them alphabetically, just like the MODX folders located on the server.

p.s. PR does not change anything, but when transferring a site from server to server or changing the config file, such inconsistency is striking.

### Why is it needed?
Fewer edits and fewer possible bugs in config.inc.php.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/10479#issuecomment-568462571

### DIFF
--- a/core/docs/config.inc.tpl
+++ b/core/docs/config.inc.tpl
@@ -4,11 +4,11 @@
  */
 $database_type = '{database_type}';
 $database_server = '{database_server}';
+$database_connection_charset = '{database_connection_charset}';
+$table_prefix = '{table_prefix}';
+$dbase = '{dbase}';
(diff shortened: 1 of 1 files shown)


smg6511 (approved):
Although I don't feel strongly one way or the other on this, it _does_ make the order more logical. And there's no functional difference, so no worries there.

Mark-H (approved):

PR #16460, opened 2023-07-27
Hopefully fixed recently edited resources (sub)query

rtripault:
### What does it do?
Adjusted the (sub) query to retrieve last edited resources

### Why is it needed?
When having many manager log records, the existing query already returned "outdated" recently edited resources

### How to test
* Generate a bunch of manager log entries (we tested with 500) for resource edition
* manually edit a resource
* check the Recently edited dashboard widget (or `/?a=security/profile`) and confirm last edited resource is not accurate
* apply patch and confirm last edited resource is correct ;)

### Related issue(s)/PR(s)
none found

### DIFF
--- a/core/src/Revolution/Processors/Security/User/GetRecentlyEditedResources.php
+++ b/core/src/Revolution/Processors/Security/User/GetRecentlyEditedResources.php
@@ -60,22 +60,21 @@ class GetRecentlyEditedResources extends GetListProcessor
     {
         $user = $this->getProperty('user');
         $q = $this->modx->newQuery($this->classKey, ['classKey:IN' => $this->classKeys]);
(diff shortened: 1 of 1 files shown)


smg6511 (approved):
Tested this out and it works. Nice fix!

A couple of side notes to all ... which may be addressed in another PR:
1. Consider excluding empty_trash and _perhaps_ resource_delete
2. Consider whether this widget should be connected to the actions log, as once the log is purged this widget will incorrectly show no recent activity (correct?)
3. The word "edited" in the widget heading and in the date column is a little misleading; we may want to remove or reword somehow.
4. Maybe re-label the actions to be more natural sounding ("Create" instead of "resource_create", etc.)

PR #16461, opened 2023-07-28
Change default gravatar type for user

Ruslan-Aleev:
### What does it do?
In the 2.x branch, the gravatar image is enabled by default, BUT, because in code set `mm` gravatar type - the picture is always the same, which is not very useful for UI\UX.

Changed the type to `retro`, which generates a unique gravatar by email hash, and now you can better see which user is logged into the manager panel.

Before:
gravatar_1

After:
gravatar_2

gravatar_3

### Why is it needed?
UI\UX

### Related issue(s)/PR(s)
N\A

### DIFF
--- a/core/model/modx/moduser.class.php
+++ b/core/model/modx/moduser.class.php
@@ -918,7 +918,7 @@ class modUser extends modPrincipal {
      *
      * @return string The Gravatar photo URL
      */
-    public function getGravatar($size = 128, $default = 'mm') {
+    public function getGravatar($size = 128, $default = 'retro') {
         $gravemail = md5(
             strtolower(
                 trim($this->Profile->email)


JoshuaLuckers:
@cla-bot check

SnowCreative:
Hmm. Interesting feature. But maybe toggle this on using a system setting? I'd personally prefer the generic image, while others may enjoy the auto-created ones.

Ruslan-Aleev:
@SnowCreative The 2.x branch is frozen, so there is no point in making additional system settings. Even in such a simple version, PR is unlikely to be merged in...

opengeek:
> The 2.x branch is frozen, so there is no point in making additional system settings.

If it is frozen, then why should this change be accepted?

> Even in such a simple version, PR is unlikely to be merged in...

And why do you say that?

opengeek:
> > The 2.x branch is frozen, so there is no point in making additional system settings.
> 
> If it is frozen, then why should this change be accepted?
> 
> > Even in such a simple version, PR is unlikely to be merged in...
> 
> And why do you say that?

And for clarity, I thought this was a good change…

opengeek (approved):

Ruslan-Aleev:
> If it is frozen, then why should this change be accepted?

Here the changes are not global, so I sent a PR, if the PR is merged - it's good, if not - it's not critical =)

> And why do you say that?

More valuable PRs are waiting for testing, there are almost no testers, so I have little optimism, but I will be glad to be wrong.

JoshuaLuckers (approved):

matdave:
I'd be +1 for making this a system setting

PR #16462, opened 2023-08-04
fix return type GetProcessor.php

catsmeatman:
### What does it do?
add string type to return types initialize method

### Why is it needed?
method initialize GetProcessor.php can return bool or string value.

### How to test
Create a inherited class of GetProcessor.php. Add return types in initialize method `function initialize() :bool`. Now in all cases will returned `true`. If add return types `function initialize() :bool|true` (as php DocBlock) everything will be alright.

### DIFF
--- a/core/src/Revolution/Processors/Model/GetProcessor.php
+++ b/core/src/Revolution/Processors/Model/GetProcessor.php
@@ -28,7 +28,7 @@ abstract class GetProcessor extends ModelProcessor
 
     /**
      * {@inheritDoc}
-     * @return boolean
+     * @return boolean|string
      */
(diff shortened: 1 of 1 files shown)


opengeek:
Why would initialize() return a string value?

Mark-H:
Errors or validation that prevents a processor from continuing (i.e. it's process() method should not be run) are returned as a lexicon string. If you expand the changed code block there are a number of examples right there.

catsmeatman:
> Why would initialize() return a string value?

core/src/Revolution/Processors/Model/GetProcessor.php in line 41 or 46 return string from lexicon

PR #16473, opened 2023-09-19
Fix ability to view Manager Actions

### LINKED ISSUE #16445: Can't see Manager Actions Log unless I'm a sudo user
### Summary
In MODX 3.0.3 if I attempt to access the Manager Actions Log, I get an "Access Denied: Not allowed to access this controller" error. I'm an admin super user with all permissions, including the `mgr_log_view` permission which is the only one required for the log.

If I make myself a sudo user, I can see the log as usual.

smg6511:
### What does it do?
Update view permissions key in logs controller.

### Why is it needed?
The old 2.x key was still being specified in this controller, making it impossible for those without sudo permissions to view the Manager Actions log.

### How to test
Test against a user having the `mgr_log_view` (and without sudo permissions) to verify Manager Actions can be viewed and navigated.

### Related issue(s)/PR(s)
Resolves #16445

### DIFF
--- a/manager/controllers/default/system/logs/index.class.php
+++ b/manager/controllers/default/system/logs/index.class.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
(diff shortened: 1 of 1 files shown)


BobRay:
It looks like there may be [some problems](https://github.com/modxcms/revolution/actions/runs/6230479403/job/16910549741?pr=16473) with the PR that prevent it from being merged. Wrapping line 37 should solve one of them. The namespace one makes no sense to me. 

Just changing the permission should solve the problem. Are the other changes really necessary?

smg6511:
@BobRay - Yes, don't worry about those; the namespace one is something that'll be taken care of when another of my recent PRs gets merged that relaxes the rule that causes the error you see. The other changes are just code formatting fixes to conform to our PSR12 rules (otherwise you'd see a bunch more errors about those).

PR #16475, opened 2023-09-20
Make “root” a reserved (restricted) key for Contexts

### LINKED ISSUE #16457: Naming a context root crashes the manager resource tree
### Summary
When adding a context with context key `root` it crashed the resource tree.

### Step to reproduce
Create a context with key `root`

### Observed behavior
It duplicates all existing contexts and formats them twice in the resource tree.

### Expected behavior
It should just add a context named root.

### Environment
MODX 2.8.5
Screenshot_2023-07-19_at_15_00_59

smg6511:
### What does it do?

1. Prevents "root" from being specified as a Context key.
2. Slightly refactors the `Create->beforeSave() `error testing conditional to use a `switch` statement so only relevant code will execute in that block.
3. Additionally, in the second commit, leverages the newly-added `RESERVED_KEYS` constant (in the `modContext` base class) in the Context processor `GetList` class.

### Why is it needed?
The keyword "root" is used as the default id for all trees in the core. As such, attempting to create a Context with that key results in errors in the rendering of the Resources tree.

### How to test
Attempt to create a new Context with the key "root." You should receive an error message indicating the key is reserved.

### Related issue(s)/PR(s)
Resolves #16457

### DIFF
--- a/core/lexicon/en/context.inc.php
+++ b/core/lexicon/en/context.inc.php
@@ -18,6 +18,7 @@ $_lang['context_err_nfs'] = 'Context not found with key: [[+key]]';
 $_lang['context_err_ns'] = 'Context not specified.';
 $_lang['context_err_ns_key'] = 'Please specify a valid key for the Context.';
 $_lang['context_err_remove'] = 'An error occurred while trying to delete the Context.';
+$_lang['context_err_reserved'] = 'The Context key you chose is reserved for system use only. Please specify a different key.';
 $_lang['context_err_save'] = 'An error occurred while saving the Context.';
 $_lang['context_id'] = 'Ctx ID';
 $_lang['context_key'] = 'Context Key';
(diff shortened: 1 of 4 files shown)


theboxer (dismissed):

Jako (approved):

Mark-H:
While `root` gets blocked correctly, `ROOT` or `roOt` bypasses the check. Not 100% sure, but we probably want to guard against this regardless of casing?

smg6511:
> While `root` gets blocked correctly, `ROOT` or `roOt` bypasses the check. Not 100% sure, but we probably want to guard against this regardless of casing?

Yeah, while it may or may not be needed technically, it'd be a good idea. I'll make a small update to account for casing later today...

Mark-H (approved):

Mark-H:
Thank you @smg6511.

Ibochkarev (approved):

Issue #16376, opened 2023-02-01
Missing sanity check in resource/create controller


BobRay:
### Summary
Modx 2
Line 254 of manager/controllers/default/resource/create.class.php 
throws an "empty IN condition error if `$usergroups` is empty.

```
array(
                'ProfileUserGroup.usergroup:IN' => $userGroups,
                array(
                    'OR:ProfileUserGroup.usergroup:IS' => null,
                    'AND:UGProfile.active:=' => true,
                ),
            ),
```

### Step to reproduce
No sure how to reproduce this, but a user reported it [here](https://community.modx.com/t/error-log-question/6359/2).

It's Form Customization code, so possibly a user who doesn't belong to a group or a reference to a group that no longer exists.

Other similar examples only make the query if the variable is not empty.

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/error-log-question/6359/3

PR #16477, opened 2023-09-27
Fix sanity check in resource/create controller

arjen-t:
### What does it do?
Add sanity check for $userGroups in XPDO query during creating a new resource.

### Why is it needed?
In the case the user is a sudo user and isn't assigned to any groups, the $userGroups which is a reflection of the user assigned groups can be empty. This cause a SQL error during creating a new resource.

```
Array
(
    [0] => 42000
    [1] => 1064
    [2] => You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') OR  ( `ProfileUserGroup`.`usergroup` IS NULL AND `UGProfile`.`active` = 1 )  )' at line 1
)
```

### How to test

1. Login in the manager with only sudo rights.
2. Create a new resource
3. See error log for the SQL error.

### Related issue(s)/PR(s)
#16376

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -249,14 +249,20 @@ class ResourceCreateManagerController extends ResourceManagerController {
             'FCSet.active' => true,
             'Profile.active' => true,
         ));
+
+        $criteriaUserGroups = [];
+
+        if(!empty($userGroups)){
+            $criteriaUserGroups['ProfileUserGroup.usergroup:IN'] = $userGroups;
+        }
+
+        $criteriaUserGroups[] = array(
+            'OR:ProfileUserGroup.usergroup:IS' => null,
+            'AND:UGProfile.active:=' => true,
+        );
+
         $c->where(array(
-            array(
-                'ProfileUserGroup.usergroup:IN' => $userGroups,
-                array(
-                    'OR:ProfileUserGroup.usergroup:IS' => null,
-                    'AND:UGProfile.active:=' => true,
-                ),
-            ),
+            $criteriaUserGroups,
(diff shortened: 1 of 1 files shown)


manager/controllers/default/resource/create.class.php
```diff
@@ -249,14 +249,20 @@ public function getDefaultTemplate() {
...
         ));
+
+        $criteriaUserGroups = [];
+
+        if(!empty($userGroups)){
+            $criteriaUserGroups['ProfileUserGroup.usergroup:IN'] = $userGroups;
+        }
+
+        $criteriaUserGroups[] = array(
+            'OR:ProfileUserGroup.usergroup:IS' => null,
```
Mark-H, line 260: Should this OR be there if the previous condition isn't added?

arjen-t: As far I know this OR will be ignored by xpdo when the previous condition isn't added.

When logging the statement in MODX the following occurs:

Without the $userGroups

> AND  (  ( `ProfileUserGroup`.`usergroup` IS NULL AND `UGProfile`.`active` = 1 )  OR `ProfileUserGroup`.`usergroup` IS NULL )  )

With $userGroups:

> AND  (  ( `ProfileUserGroup`.`usergroup` IN ('1') OR  ( `ProfileUserGroup`.`usergroup` IS NULL AND `UGProfile`.`active` = 1 )  )  OR `ProfileUserGroup`.`usergroup` IS NULL )  )

side note which has nothing to do with this  bug.

> OR ( ProfileUserGroup.usergroup IS NULL AND UGProfile.active = 1 ) 

This nested OR seems to be redundant.

theboxer (approved):
👍

PR #16484, opened 2023-10-17
Fix errors when hovering over the action column header

theboxer:
### What does it do?
ExtJS is grabbing the header by regex on class `'x-grid3-td-([^\\s]+)'`, for some reason the class for action column header doesn't populate correctly (when using column model). Manually specifying an id fixes the issue.

### Why is it needed?
Get rid of the errors

### How to test
Open system settings & console. hover over the action column header.
CleanShot 2023-10-17 at 15 15 59


manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -115,7 +115,8 @@ MODx.grid.Grid = function(config) {
             }
 
             config.columns.push({
-                width: config.actionsColumnWidth || defaultActionsColumnWidth
+                id: config.columns.length
```
smg6511:
```suggestion
                id: 'modx-actions'
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -129,7 +130,8 @@ MODx.grid.Grid = function(config) {
             }
 
             config.cm.columns.push({
-                width: config.actionsColumnWidth || defaultActionsColumnWidth
+                id: config.cm.columns.length
```
smg6511:
```suggestion
                id: 'modx-actions'
```

smg6511 (changes requested):
First, this does indeed fix the error. I suggest an id that's more targetable (via css) and specific to the column's purpose. That it's an _id_ is of no consequence in this case, because Ext just uses this internally and it doesn't end up creating that as an actual _id_ in the DOM.

FWIW, the root cause of this is actually the different way the settings js class is constructed. I verified this via something else I'm working on where I've refactored that base class a bit where, among other things, the column model is just specified in the config (_not_ explicitly created via `new Ext.grid.ColumnModel` as it currently is). In this scenario, the id is not needed on the actions column. That's why the error doesn't appear in any of the non-settings grids.

manager/assets/modext/widgets/core/modx.grid.js
theboxer: `modx-actions` fails the regex test from ExtJS, the `column.length` matches ids of the other header ids, don't see a reason why this should be different

manager/assets/modext/widgets/core/modx.grid.js
smg6511: Couple things (note, I'm not looking to squabble about this, I'm just a stickler on details and am curious)—
- Where is the regex you refer to and how do you see it fail (i.e., where do you see the effects)? It's actually kind of an odd rule; why would one be looking specifically for one or more of a backslash or the letter "s"? I do wonder if that's supposed to be `[^\s]+`.
- Regardless of the answer to that, the id 'modx-actions' does work just fine in the end—causes no errors and the classes get added to the header and column cells as intended.
- True, using the length matches the default way Ext generates the unique col classes, but that is just a default.

In any case, if you feel strongly about doing it the way you did that's not a big deal - I just wanted to throw this out there, especially if it might be desirable to _easily_ style this column in a particular way in the future.

smg6511: BTW, whatever you decide, if you see the "Resolve Conversation" button for these comments I've made go ahead and close 'em out. I don't see the button for some reason.

manager/assets/modext/widgets/core/modx.grid.js
smg6511: See below (meant to reply here but managed to do so below ... haven't had enough coffee yet this morning!)

smg6511 (approved):

manager/assets/modext/widgets/core/modx.grid.js
theboxer: bah, I had an issue with double escaping while testing the regex, you were right it works. adjusted it.

for your reference, the regex is in https://docs.sencha.com/extjs/3.4.0/source/GridView.html#Ext-grid-GridView
search for: `this.colRe = new RegExp('x-grid3-td-([^\\s]+)', '');`

opengeek:
@cla-bot check

opengeek:
@cla-bot[bot] check

Mark-H:
Probably worth double checking on pages with more than one grid.

smg6511:
> Probably worth double checking on pages with more than one grid.

I double-checked this in the one area that has multiple grids that include an actions column (the ACLs) and all works as expected.

Mark-H (approved):

opengeek:
I was unable to apply this to 3.0.x. If we want this fix on 3.0.x, I'll need the relevant changes submitted against that branch directly.

PR #16488, opened 2023-10-23
[2.x] Fix functionality in the section "Categories" of the elements tree

halftrainedharry:
### What does it do?
In some handler-functions, the information about the clicked item (type and id) gets extracted from the **node-ID** in the tree.
In most sections, the IDs have the prefix `n_`.
In the "Categories" section, nodes for other types have the prefix `n_c_` instead:

https://github.com/modxcms/revolution/blob/3f887dc6494fe32d3b2a52da20ec76a63bc8fb7d/core/model/modx/processors/element/getnodes.class.php#L480

This is now taken into account.

### Why is it needed?
Some context menu items don't work correctly, when executed for element types (other than "Category") in the "Categories" section of the elements tree.

modx_manager_elements_categories_tv_remove

### How to test
Go to the section "Categories" in the elements tree and check that the following options (in the context menu) now work:

- Remove `<type>` (for the types "Template", "TV", "Chunk", "Snippet" and "Plugin")
- Activate and Deactivate Plugin (for type "Plugin")
- Create a New `<type>` here (for the types "Template", "TV", "Chunk", "Snippet" and "Plugin")

### Related issue(s)/PR(s)
Resolves #16487 for MODX 2.x

PR #16489, opened 2023-10-23
Fix Element handling from within Categories tree

### LINKED ISSUE #16487: tv will not be deleted if delete from categories in left menu
### Summary
Try to delete a tv, by clicking right mouse - delete, in left menu  in categories.
Will throw - cant find processor 'element/c/remove'

### Step to reproduce
Try to delete a tv, by clicking right mouse - delete, in left menu  in categories.

### Observed behavior
Will throw - cant find processor 'element/c/remove'

### Expected behavior
To delete a tv.

### Environment
MODX version 2.8.5

smg6511:
### What does it do?
Adds a new tree method to properly extract Element data from the working tree Node. There are three patterns for a Node's id, but only one was being accounted for in the code. Additionally, cleans up code formatting in the affected methods.

### Why is it needed?
Within the Categories tree, these actions are currently inoperable:
- Element removal
- Plugin activation/deactivation

### How to test
1. Create a handful of Elements (at least one of each type), both within a Category and within the root (no Category)
2. Verify the ability to remove each Element (from both the Categories tree and from within the individual Element trees)
3. Verify the ability to activate/deactivate Plugins (from both the Categories tree and from within the individual Element trees)

### Related issue(s)/PR(s)
Resolves #16487 for MODX 3.x
Related to 2.x solution #16488


theboxer (approved):
Works nicely

theboxer:
@smg6511 would it make sense for you to implement same approach for 2.x?

smg6511:
> @smg6511 would it make sense for you to implement same approach for 2.x?

It would, I just didn't want to step on any toes, since @halftrainedharry already put up a different solution for 2.x.

PR #16496, opened 2023-11-15
S3 Update: Add No Bucket Check

matdave:
### What does it do?
Add a "No Bucket Check" option for users with limited access.

### Why is it needed?
Some access keys aren't authorized to list buckets, so will fail to initialize the S3 source when it checks if the bucket exists.

### How to test
This would take some set up. Primarily, this is addressing an issue we are facing with updating our internal S3 provider. However, it is a common feature and the logic was ported from rclone https://rclone.org/s3/#s3-no-check-bucket

### DIFF
--- a/core/src/Revolution/Sources/modS3MediaSource.php
+++ b/core/src/Revolution/Sources/modS3MediaSource.php
@@ -161,9 +161,9 @@ class modS3MediaSource extends modMediaSource
                 'value' => '',
                 'lexicon' => 'core:source',
             ],
-            'no_bucket_check' => [
-                'name' => 'no_bucket_check',
-                'desc' => 'prop_s3.no_bucket_check_desc',
+            'no_check_bucket' => [
+                'name' => 'no_check_bucket',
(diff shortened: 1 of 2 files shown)


core/src/Revolution/Sources/modS3MediaSource.php
```diff
@@ -54,13 +55,15 @@ public function initialize()
...
             $client = new S3Client($config);
-            if (!$client->doesBucketExist($bucket)) {
-                $this->xpdo->log(
-                    xPDO::LOG_LEVEL_ERROR,
-                    $this->xpdo->lexicon('source_err_init', ['source' => $this->get('name')])
-                );
-
-                return false;
+            if (!$noBucketCheck) {
+                if (!$client->doesBucketExist($bucket)) {
```
Mark-H: Code-style nitpick, merge these conditions into one `if`. Keeps the nesting level lower and generally lowers complexity of the code.

theboxer (approved):
👍

Mark-H (approved):

PR #16502, opened 2023-12-13
Clarify message when changing user passwords

Ruslan-Aleev:
### Why is it needed?
When you change passwords for several users, it is not always clear which user's password was changed. The user's login and password are not visible.

user_before

user_after

### Related issue(s)/PR(s)
N\A


Ibochkarev (approved):

core/src/Revolution/Processors/Security/User/Create.php
```diff
@@ -239,9 +239,16 @@ public function sendNotificationEmail() {
      * @return array|string
      */
     public function cleanup() {
+        $userArray = $this->object->toArray();
+        $profile = $this->object->getOne('Profile');
```
Mark-H: As we're only trying to access the `username`, we don't need to load the profile, or even turn it into an array - we can just do `$this->object->get('username')` in the lexicon call a few lines down.

Ruslan-Aleev:
@Mark-H Corrected, check.

Mark-H (approved):

Mark-H:
Thanks @Ruslan-Aleev. 

(CI failing is not related to your PR.)

PR #16505, opened 2023-12-14
Fix saving time for date TVs created in MODX 2.x

halftrainedharry:
### What does it do?
Fixes the code in the "Resource/Create" and "Resource/Update" processors (that was added in #16398) to also work for date TVs that were created in MODX 2.x.

### Why is it needed?
In MODX 2.x, the input property **"hideTime"** of a Date-TV is stored as the value "false" (or "true") -> `s:8:"hideTime";s:5:"false";`
In MODX 3, the same property is stored as the value "0" (or "1") -> `s:8:"hideTime";s:1:"0";`

In MODX installations, that were updated from MODX 2.x to MODX 3, the time part of the date-TV-value always gets deleted.

### How to test

- Create a TV of type "date".
- Change the value of "hideTime" from `s:1:"0";` to `s:5:"false";` in the column "input_properties".
- Make sure, that the time part of the TV value still gets saved correctly.

### Related topic in the MODX forum
https://community.modx.com/t/date-tv-time-wont-save/7335

### DIFF
--- a/core/src/Revolution/Processors/Resource/Create.php
+++ b/core/src/Revolution/Processors/Resource/Create.php
@@ -510,7 +510,7 @@ class Create extends CreateProcessor
                         $tvProperties = $tv->get('input_properties');
                         if (!empty($value)) {
                             $dateTime = new \DateTime($value);
-                            if (array_key_exists('hideTime', $tvProperties) && (bool)$tvProperties['hideTime']) {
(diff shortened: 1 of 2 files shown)


Jako:
Maybe it is better to prepare an upgrade script that changes the value.

Mark-H (approved):
I'm fine with either this approach and/or the the migration script to update values.

Ibochkarev (approved):

Jako:
I am also fine with that solution.

Jako (approved):

PR #16515, opened 2024-01-19
Add missing system setting for profile photo media source

### LINKED ISSUE #16510: Media Browser doesn't display popup to select a media source for user profile photos in MODX 3.0
### Summary
Media browser that popups up when selecting a photo for a user profile only displays the default media source

### Step to reproduce
In MODX 3 site, edit your profile, and click the icon to select a user photo.

### Observed behavior
There is no popup to select a media source. Only the folder for the default media source is displayed.
Screen Shot 2024-01-12 at 1 20 24 PM

### Expected behavior
The Media Browser popup to select a media source should be displayed, just like it is everywhere else the media browser is used.
Screen Shot 2024-01-12 at 1 22 10 PM

### Environment
MODX 3.0.4-pl

smg6511:
### What does it do?
Adds a long-missing setting for specifying user's profile photo media source. Also, moves up (and tweaks) the associated Lexicon entries.

### Why is it needed?
The logic has been present in the backend to specify a Media Source other that the default for user profile photos since they were introduced nearly 10 years ago. However, the system setting was never added to the build, making it virtually unknown that this functionality existed.

### How to test
Run `_build/transport.core.php` to install the new setting. Then search for `photo_profile_source` in the system settings, change its value, and verify that the Media Source shown in the Media Browser (when editing/adding a user profile photo) matches that which you specified in the setting.

### Related issue(s)/PR(s)
Resolves #16510

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -434,6 +434,15 @@ $settings['default_media_source_type']->fromArray([
     'area' => 'manager',
     'editedon' => null,
 ], '', true, true);
+$settings['photo_profile_source'] = $xpdo->newObject(modSystemSetting::class);
+$settings['photo_profile_source']->fromArray([
+  'key' => 'photo_profile_source',
+  'value' => null,
+  'xtype' => 'modx-combo-source',
+  'namespace' => 'core',
+  'area' => 'manager',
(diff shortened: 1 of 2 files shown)


Mark-H (approved):

theboxer (approved):
👍

PR #16521, opened 2024-02-10
Update generate password logic to use proper methods

### LINKED ISSUE #15740: Use strong random number/string generators inside generatePassword()

Mark-H:
### What does it do?

Re-up of #15894

Changes password generation method to be more secure.

### Why is it needed?
Actually random generation. 

### How to test
Apply and see passwords still get generated.

### Related issue(s)/PR(s)
Re-up of #15894
Fixes #15740

### DIFF
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -900,20 +900,28 @@ class modUser extends modPrincipal
         if ($length === null) {
             $length = (int)$this->xpdo->getOption('password_generated_length', null, 10, true);
         }
+
         $passwordMinimumLength = (int)$this->xpdo->getOption('password_min_length', null, 8, true);
         if ($length < $passwordMinimumLength) {
             $length = $passwordMinimumLength;
         }
 
-        if (!empty($options['alphabet'])) {
-            $alphabet = array_merge(range('a', 'z'), range('A', 'Z'));
-            shuffle($alphabet);
-            return substr(implode($alphabet), 0, $length);
+        if (empty($options['allowable_characters'])) {
+            $options['allowable_characters'] = 'abcdefghjkmnpqrstuvxyzABCDEFGHJKLMNPQRSTUVWXYZ23456789';
         }
 
-        return substr(bin2hex(random_bytes($length)), $length);
-    }
+        $allowableCharactersLength = strlen($options['allowable_characters']);
 
+        $randomBytes = random_bytes($length);
+
+        $pass = '';
+        for ($i = 0; $i < $length; $i++) {
+            $randomIndex = ord($randomBytes[$i]) % $allowableCharactersLength;
+            $pass .= $options['allowable_characters'][$randomIndex];
+        }
+
+        return $pass;
+    }
 
     /**
      * Send an email to the user
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -898,25 +898,20 @@ class modUser extends modPrincipal
     public function generatePassword($length = null, array $options = [])
     {
         if ($length === null) {
-            $length = $this->xpdo->getOption('password_generated_length', null, 10, true);
+            $length = (int)$this->xpdo->getOption('password_generated_length', null, 10, true);
         }
-        $passwordMinimumLength = $this->xpdo->getOption('password_min_length', null, 8, true);
+        $passwordMinimumLength = (int)$this->xpdo->getOption('password_min_length', null, 8, true);
         if ($length < $passwordMinimumLength) {
             $length = $passwordMinimumLength;
         }
-        $options = array_merge([
-            'allowable_characters' => 'abcdefghjkmnpqrstuvxyzABCDEFGHJKLMNPQRSTUVWXYZ23456789',
-            'srand_seed_multiplier' => 1000000,
-        ], $options);
 
-        $ps_len = strlen($options['allowable_characters']);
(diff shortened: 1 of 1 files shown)


smg6511:
I've got a couple general questions on this before doing a quick review:

- I assume that since you changed the password dictionary variable from `$options['allowable_characters']` to `$options['alphabet']` that you're not worried about breaking usage of this method by custom implementations/extras. If that's the case, we might want to consider deprecating the `$options` param or go even further an deprecate both params and consider this a private method.
- Re the following block, it seems to me you'd want to define the alphabet if the `$options['alphabet']` param _is_ empty; or am I not seeing this right:
```php
    if (!empty($options['alphabet'])) {
        $alphabet = array_merge(range('a', 'z'), range('A', 'Z'));
        shuffle($alphabet);
        return substr(implode($alphabet), 0, $length);
    }
```
in other words, is something like this what you're really going for?
```php
    $alphabet = empty($options['alphabet']) 
        ? array_merge(range('a', 'z'), range('A', 'Z')) 
        : $options['alphabet']
        ;
    shuffle($alphabet);
    return substr(implode($alphabet), 0, $length);
```

Mark-H:
Hmm, I overlooked the change from `allowable_characters` to `alphabet` recreating the original PR, but the behavior seems to have been intended differently. I don't know if the old behavior is worth keeping or not. Would welcome feedback on this, as well as examples that may use this code a certain way that would break. 

I'd be fine with removing that block entirely and limiting the functionality of this method to only the new generation with `random_bytes()` to keep things simple.

smg6511:
Yeah, my guess is there's likely been no usage of that method outside of the core. If, however, there has been and a dictionary were passed in via `$options['allowable_characters']`, that would break with the new var name. That said, I'd opt for dropping the block as you suggested.

smg6511:
Looking at this again I see what you mean when you imply the change (changing the options key in question) wouldn't break anything. It might cause unexpected output if there was external use of this method where chars and/or a custom seed were passed into the $options, but would not prevent a password from being generated. I get now that your changed variable is meant to be a _boolean_ value, not a character set (in which case a key of something like 'use_alpha_chars' would be more clear). 

At any rate, I'd still say drop that block altogether because the new generation logic you added below it will create a better password anyway. Further, I'd even move toward making this method private and removing its params, making it even more clear that this is a utility method meant for internal use only. After all, I think the intention is that these passwords are temporary ones that users should immediately change.

theboxer:
I'd agree with @smg6511 and nuke everything but the length & bin2hex usage :)

However I'd keep the signature same (even though `$options` will not be used), just to keep the BC going.

theboxer:
@Mark-H @smg6511 I adjusted the PR to keep the `allowable_characters` option and use the random bytes for generating the password.

opengeek (approved):

PR #15957, opened 2022-01-06
Add modSessionHandlerInterface

theboxer:
### What does it do?
- Creates a modSessionHandlerInterface interface and uses it for modSessionHandler
- Moves flushSessions method from `Security\Flush` processor to the interface & handler
- In the `Security\Flush` processor - checks if `flushSessions` method exists in the handler class stored in `session_handler_class` system setting and executes it if so

### Why is it needed?
Custom session handlers can't flush all sessions.

### How to test
#15928

### Related issue(s)/PR(s)
Resolves #15928

PR #16522, opened 2024-02-10
Fix issues with extending modSessionHandler and flushing all sessions

### LINKED ISSUE #15928: Logout all users action doesn't work with custom session handlers

Mark-H:
### What does it do?

- Fixes the bugs from #15928, per #15934
- Refactored inititialization of the session handler and adopt the PHP core SessionHandlerInterface, per #15934 
- Move flushing sessions logic into the session handler to optionally allow that to be extended, per #15957

### Why is it needed?

In two stale PRs #15928 and #15957 we have two proposals for dealing with some bugs and extending session handlers. While trying to figure out which one to use, I found both to have a solid approach, and wanting to use one bit of one and another part of other PR. 

This proposal replaces both those PRs as a middle ground, taking the most benefit from both approachs. 

### Breaking changes

This change includes a small breaking change in order to adopt the PHP standard SessionHandlerInterface, namely adding parameters into the open() method. There's not really any way around that in order to adopt the standard. 

As extending session handlers is a pretty advanced thing that I assume not many people have done, I suggest we accept that for the sake of getting closer to the standards, but make special note of this breaking change in the docs for 3.1 and the announcement to make sure people learn about it in case custom handlers need to be adjusted. 

### How to test
Flush sessions

### Related issue(s)/PR(s)

Replaces #15934 and #15957
Fixes #15928

### DIFF
--- a/core/src/Revolution/Processors/Security/Flush.php
+++ b/core/src/Revolution/Processors/Security/Flush.php
@@ -19,6 +19,14 @@ use MODX\Revolution\modSessionHandler;
  */
 class Flush extends Processor
 {
+    /**
+     * @return array
+     */
+    public function getLanguageTopics()
+    {
+        return ['topmenu'];
+    }
+
     public function checkPermissions()
     {
         return $this->modx->hasPermission('flush_sessions');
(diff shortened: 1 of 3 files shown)


theboxer (approved):
Thanks for merging the 2 prs into a solid one :) I'm fine with introducing such a small breaking change in the minor release.

opengeek (approved):

PR #16523, opened 2024-02-10
3.0.x -> 3.x

Mark-H:
### What does it do?

### Why is it needed?

### How to test

### Related issue(s)/PR(s)


Mark-H:
@opengeek Any objections to me doing a regular merge fixing the conflicts on this? Or would you prefer to handle this yourself?

opengeek:
I absolutely do NOT want to do merges like this. Most of those changes are already in there. If something specific was merged into 3.0.x and not put into 3.x, that needs to be dealt with individually. I handle this manually with cherry-picks when merging PRs targeting one or the other when the changes need to be in both.

Mark-H:
Alright. It's just confusing that the changelog doesn't list anything past 3.0.3, so I don't know how to make sure all fixes made it into 3.1. Unless all the 3.0.x changes need to be cherry picks from things merged into 3.x? 

For what it's worth I've only done merges into 3.x today.

opengeek:
The changelog isn't updated at all. The commit log is what matters. Whatever you merged into 3.x which needs to go in 3.0.x, should be manually merged into 3.0.x. If you need me to handle this, I will, but I would prefer in the future that this get done for each PR as we go so that we are sure it applies and is tested against each branch individually.

opengeek:
It is critical that each merge is done to both branches. The commit log can then be used to generate the changelog entries at the end of the release cycle with no effort.

PR #16529, opened 2024-02-20
Fix grouping grid sorting - ACL grids

smg6511:
### What does it do?
Adds new grouping logic to ACL GetList processors and config props in the corresponding Ext grid classes.

### Why is it needed?
Aside from the Settings grid,* other grouping grids (such as the ACL ones) neither group properly, nor sort within groups properly.

### How to test

1. Rebuild may be necessary and compress_js must be off. 
2. Create a variety of ACL entries so there are at least a few main items to group (this is the Minimum Role, which is what all ACLs are grouped by as a default [can be changed in the column menu]), and a few entries within those groups (meaning you'll need multiple contexts, namespaces, resource groups, media sources, and element categories created ahead of time to assign these roles to).
3. Verify that items are grouped as expected.
4. Verify that sorting on each column works as expected.

### Related issue(s)/PR(s)
No related issue.

* The settings grid does not group correctly if the grouping field is changed. That will be addressed in a different PR.


core/src/Revolution/Processors/Model/GetListProcessor.php
```diff
@@ -187,6 +190,42 @@ public function getSortClassKey()
...
+    public function setGroupSort(xPDOQuery &$c, string $sortBy, string $groupBy, string $groupKey, string $gridCategory = '')
+    {
+        /*
+            When group sort and column sort are the same data index, sort the groups
+            based on the current column sort direction. Otherwise, add an initial sortby
+            to specify the group sort; the secondary (sorting within the groups) is subsequently
+            added later in the getData method.
+        */
+        switch ($gridCategory) {
+            case 'usergroup-acl':
```
Mark-H: Should this logic be in the generic GetListProcessor or something ACL related?

smg6511: I chose the placement of the logic based on looking to support other grouping grids down the road (e.g., the Settings one [which could use some work as mentioned in the opening comments], or any Extras that might make use of that type of grid). That said, let me know if you think there's a more appropriate place; perhaps a new processor extending the main GetListProcessor.

smg6511: @Mark-H - Hey, it's been a while since you made this comment; let me know what your thoughts are on my response above... Thx ;-)

Mark-H: If there needs to be an extra hook in the base getlist processor I don't have objections to that, but I do still think it's kinda weird to have a switch with values for specific implementations in a base class. That just seems like it's not architected properly.

smg6511: I agree it's not ideal; honestly there's a fair amount of wonkiness in the arch of the processors in general (e.g., there's inconsistency in how get data is handled [ two ways that I've come across ]) that makes it difficult to keep things DRY. 

I can take a look at adding an intermediate class if you prefer.

theboxer: @smg6511 I don't think it needs additional class, but just an additional method in the processor.

I'd add this to the model processor:
```php
public function getSecondaryGroupCondition(string $sortBy, string $groupBy, string $groupKey)
{
    return false;
}
```

and override the method where needed, so in the `AccessNamespace/GetList` you'd do:
```php
public function getSecondaryGroupCondition(string $sortBy, string $groupBy, string $groupKey)
{
    return $sortBy === 'authority' && $groupBy === 'role_display';
}
```

With this, you should be able to also drop the `$gridCategory` param.

smg6511: @theboxer - John, appreciate the feedback. Why didn't I think of that? ;-) Will incorporate and post later tonight...

smg6511: @theboxer @Mark-H - Ok guys, this should be ready to go ... please pull down the changes and take a look again when you get a chance.

theboxer (approved):
thx for the updates, I'd say this is good to go now

PR #16530, opened 2024-02-22
Reorganize main menu

Ruslan-Aleev:
### What does it do?
Changed the order of the main menu items.
The items from the "Manage" section have moved to other logically related sections and **a new "Access"** item has been added and that's it.

### What does it look like now:

**All menu**
menu

---

**Content**
content

---

**Extras**
extras

---

**Account**
user
The language switch is now in the system section. The language switch can be left in "Account" section, but it seems to me that this is more about the "System" than the "Account".

---

**Access**
access
All items related to users and access are moved to one section.

---

**System**
system_1

system_2

### Related issue(s)/PR(s)
Discussed previously in https://github.com/modxcms/revolution/pull/14855


Ruslan-Aleev:
I re-opened a previously closed PR; the most useful thing about PR, it seems to me, is a new item for setting up access.

rthrash:
I like the access menu, but to me the language switcher is now too hidden. If I find myself on a site in a language that I’m not familiar with I look for a flag icon or a two-letter language designation somewhere obvious at the top level, not buried behind a menu in a language I can’t read or an icon that doesn’t scream language.

Ruslan-Aleev:
@rthrash Yes, it’s probably worth returning the language switch to the user section, as it is now.
Although it’s unlikely that the language is switched often in the manager panel. After all, this is a manager panel, not a website.

rthrash:
That’s very fair. I typically switch it on the login page where it is fairly obvious.

Mark-H:
As far as reviewing this goes, I have no strong opinions on this either way. I'm fine with both the current and the proposed menu structure.

Ruslan-Aleev:
Yes, I understand that the changes are not critical and can be left as they are.
But I see 2 advantages in the new menu:
- New separate access menu item;
- More logical grouping of items. For example, even though I work with MODX every day, I won’t tell from memory where some items are located. With the new structure, I hope this will become more obvious.

rthrash:
I really do like this a lot and think it's a much more logical arrangement. I will get lost looking for the cache clear menu item the first few times, but I'll get used to it quickly. ;) Kudos @Ruslan-Aleev for your thoughtful work and the PR.

Ruslan-Aleev:
@rthrash Thank you for your kind words and participation!

jaygilmore:
@Ruslan-Aleev This mostly seems like an improvement, however, there are a couple issues that came to mind when reviewing to incorporate into 3.1: 

The first issue I didn't see addressed was related to ACLs. The default menu items have permissions associated with them and changing the parent/child relationships will potentially cause issues where users have limited permissions. 

image

It's not clear (without testing) what might happen for those users with limited access (non Sudo) where access is being controlled to view submenu items via ACLs and the items change parents. 

In addition, I'm not sure what impact adding a new item without corresponding permissions would have. How might someone prevent access to this new Access menu item? 

The second issue is related to the ACLs but also to upgrades. Does the menu get updated on upgrade? What does it do to people's menu customization? 

Generally, the changes aim to improve UX; that's a win. I do think it requires some investigation and testing related to the two issues I brough up. :)

Ruslan-Aleev:
@jaygilmore I didn’t quite understand the problem with the ACL, I didn’t create new policies, and I didn’t change them.
Those what menu items were available to the user will remain so; It’s just that now some menu items have moved from system settings to the main menu.

The menu will not change during the update, it seems unnecessary to me, i.e. what the user has configured in his menu will remain. The new menu should only be installed during a clean installation, it seems to me.

jaygilmore:
Hey @Ruslan-Aleev, 

I wasn't sure. If you're confident that this won't change things when upgrading and it's not going to change how existing users access things, that's fine. It just popped up as a thought when I was reviewing the PR. As I noted, I mostly see this as an improvement.

Ruslan-Aleev:
@jaygilmore Now I'm not 100% sure =)
But logically, yes, we do not have an update script that affects the menu, for example like [here](https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/common/3.0.0-update-menu-main.php). And I didn't create new policies, so there shouldn't be any problems.

opengeek:
> @jaygilmore Now I'm not 100% sure =) But logically, yes, we do not have an update script that affects the menu, for example like [here](https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/common/3.0.0-update-menu-main.php). And I didn't create new policies, so there shouldn't be any problems.

So these updates only apply to new installations? Existing ones won't get the updated menu layout?

Ruslan-Aleev:
> So these updates only apply to new installations? Existing ones won't get the updated menu layout?

Yes, I think this is correct. I wouldn't change the menu at all when updating.
Firstly, the structure changes too much, and secondly, the user could customize the menu for himself, but the update will reset this.
This only makes sense for a new installation, I think.

theboxer (changes requested):
@Ruslan-Aleev I really like the new menu ❤️ 

While I guess it's fine to have the menus just for new installs, I think most installs will be upgrades, so it's kind of shame those ppl won't get the new menus. But I can see that due to custom menu items, custom ACLs, it would be kind of tricky to have migration script to switch to new menus on upgrade, without breaking the desired visibility set from ACLs.

I have one thing though:
I don't think namespaces are such a hot item, that would have to take the real estate under extras, you rarely open it, if ever (as user). I'd move it back under settings (admin).

Ruslan-Aleev:
@theboxer Thanks for your kind words.

I agree that the namespaces item is rarely used by users, but the purpose of the new menu is to group items based on their logic. I.e. the new menu shows the relationship between addons and namespaces, although this is not really necessary.

BUT if the guys @rthrash @Mark-H @jaygilmore @opengeek  agree with your comment, then I will correct the PR.

theboxer:
@Ruslan-Aleev also one more thing, my head was going crazy trying to follow the menu structure and how it changed, this is not really related to your changes, but more to how the whole file is constructed. I adjusted it so the whole menu is defined as a single array with clear nesting of children and comments to separate everything out.

Looks something like this:
CleanShot 2024-08-21 at 11 36 22


If you'd agree, I'd push the change to the PR (if it lets me), or send you the code so you can push it? (note, the menu structure is as proposed by you -- didn't touch namespaces, until hearing from others)

Ruslan-Aleev:
@theboxer 

> If you'd agree, I'd push the change to the PR (if it lets me), or send you the code so you can push it?

Yes, I'm for it!

theboxer:
@Ruslan-Aleev pushed

rthrash:
>I agree that the namespaces item is rarely used by users [in the Extras menu], but the purpose of the new menu is to group items based on their logic. I.e. the new menu shows the relationship between addons and namespaces, although this is not really necessary.

@Ruslan-Aleev I completely understand the rationale, but I tend to lean towards keeping namespaces under the settings like @theboxer, probably 6 or 7 out of 10, since it's more of a System-level type thing and not something you regularly interact with. I spoke with @opengeek and he's leaning that way, too. 

I'm looking forward to this making it into the 3.1 release!

rthrash:
Is there any way to handle upgrading menus without overdoing it? I do like the new logic/organization a lot and it would be a shame to not get this on upgrades.

Maybe test to see if it's a default menu config and do those, and skip any that aren't otherwise? We could do a screencast on how to adjust menus for any upgrades on sites with custom menu structures.

Ruslan-Aleev:
@rthrash I moved the "Namespace" menu item inside the "System Settings" item.
I also slightly corrected the existing menu update script.

It seems to me that we can pass the entire `$menusConfig` array to this update script and update the menu, BUT it looks strange, and I'm not sure. Maybe @theboxer can help with this? Or leave a new menu for a new installation :)

matdave:
I've reviewed these and the changes make sense. I think consolidating Manage and Admin makes things a lot more logical. As with others, I'm not sure if the Language switcher belongs in admin or in the new "Users" section. I think it's fine where it is currently in this change, and it would be more "ideal" in a separate push to move/copy it to the "user" menu and have the action only affect the "user" rather than the whole system.

matdave (approved):
I'm for these changes. Thanks Ruslan!

opengeek (approved):

PR #16493, opened 2023-10-30
Fix Element handling from within Categories tree

smg6511:
### What does it do?
Adds a new tree method to properly extract Element data from the working tree Node. There are three patterns for a Node's id, but only one was being accounted for in the code. Additionally, cleans up code formatting in the affected methods.

### Why is it needed?
Within the Categories tree, these actions are currently inoperable:
- Element removal
- Plugin activation/deactivation

### How to test
1. Create a handful of Elements (at least one of each type), both within a Category and within the root (no Category)
2. Verify the ability to remove each Element (from both the Categories tree and from within the individual Element trees)
3. Verify the ability to activate/deactivate Plugins (from both the Categories tree and from within the individual Element trees)

### Related issue(s)/PR(s)
Resolves https://github.com/modxcms/revolution/issues/16487 for MODX 2.x
Port of PR https://github.com/modxcms/revolution/pull/16489

PR #16467, opened 2023-09-04
Add guard-clause to clearDirty method

sebastian-marinescu:
### What does it do?
This makes sure the JS doesn't throw an error in the case that the items-object is an array, just like in https://github.com/modxcms/revolution/pull/16404 for 3.x

### Why is it needed?
When using ExtJS to create components of the xtype: 'radiogroup' to be used in CMPs inside a form,
this function gets called and tries to iterate over it's items array with a method that doesn't exist.

This PR isn't changing the way forms or the clearDirty-method are handled, it just assures no error is thrown when using radiogroup-components.

### Related issue(s)/PR(s)
Resolves https://github.com/modmore/ClientConfig/issues/202
Resolves https://github.com/modmore/ClientConfig/pull/176
Resolves https://github.com/modmore/ClientConfig/issues/143

PR #16540, opened 2024-03-21
Update Template Build Process

smg6511:
### What does it do?
Replaces most legacy dependencies with current versions—with the exception of bourbon, neat, and fontawesome—and drops others that are no longer relevant (such as imageoptim).

### Why is it needed?
Bring 2.x more in line with 3.x, mainly allow use of modern js features.

### How to test

1. Run the rebuild processes, including `npm update` within the `_build/templates/default` directory.
2. Run `grunt build`. 
3. Clear your manager and browser cache, then browse around the manager with your console open to verify all works as expected and no errors are being reported.

Note that grunt build will spit out some warnings, as the versions of bourbon and neat we need to stick with here (for now at least) are ancient and contain some long-deprecated code. I attempted to bring these dependencies up to date (including fontawesome) but there are many breaking changes that make it difficult to unwind and get everything working. Might try that later if there's enough "life" left in the 2.x line and it's deemed beneficial to do so.

### Related issue(s)/PR(s)
Resolves issues with building after including the following PRs: #16493 and #16467.


opengeek:
Shouldn't this include the updated package-lock.json file? If I run npm update, changes are made to the package-lock.json, which then needs to be committed again.

smg6511:
> Shouldn't this include the updated package-lock.json file?

I may be mistakenly thinking I shouldn't include it, which is why I didn't. I'll add a commit with it in just a bit...

opengeek:
> > Shouldn't this include the updated package-lock.json file?
> 
> I may be mistakenly thinking I shouldn't include it, which is why I didn't. I'll add a commit with it in just a bit...

Well, I am asking. In the PHP/composer world, you want the composer.lock file to set the dependencies to avoid differences when multiple people install the dependencies. I know that we do not include the composer.lock file in the revo project, but that is a special case since revo is essentially a framework. I would think we would want that in the case of the assets build, however. And if we don't, should it be removed from the repo?

smg6511:
I think you're right to want it in there after quick investigation.

PR #16455, opened 2023-07-01
Fix issue working with non-standard ports

dimasites:
### What does it do?
I changed logic of modx working with non-standard ports because old logic in some cases works not correct: port :44333 MODX cut ":443" and "33" part breaks the links adresses. See #16428 for details.

Relying on here [this](https://stackoverflow.com/questions/4504831/serverhttp-host-contains-port-number-too ) and [this](https://stackoverflow.com/questions/2297403/what-is-the-difference-between-http-host-and-server-name-in-php ) I suggest using the answers in stackoverflow to check the port not `$_SERVER['SERVER_PORT']` but `parse_url($_SERVER['HTTP_HOST'], PHP_URL_PORT)` because it works more reliably and correctly, as can be seen from my tests.

### Why is it needed?
Currently, when using a non-standard port for https (any one other than 443) MODX modifies request URL with mistakes, and redirect users's browser to wrong url.

See my explaination on awesome handmade infographic :) image:
explaination for pull request v2

### How to test
I have prepared a test script:
<details><summary>See code: test_modx_working_with_http_ports.php</summary>

```php

var_dump(parse_url($_SERVER['HTTP_HOST'], PHP_URL_PORT)); echo ' ← parsed port from HTTP_HOST <br>';
var_dump($_SERVER['SERVER_PORT']); echo ' ← SERVER_PORT <br>';

$http_port = parse_url($_SERVER['HTTP_HOST'], PHP_URL_PORT) ?: $_SERVER['SERVER_PORT'];
var_dump($_SERVER['SERVER_PORT']); echo ' ← http_port for usage in fixed config <br>';

$http_host = array_key_exists('HTTP_HOST', $_SERVER) ? htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES) : 'test-port.ltd';

//this is for test logic with non-standart port without server changes, COMMENT IR FOR REAL SERVER TEST
$http_host = 'test-port.ltd:44333';

echo '<br>';
var_dump($http_host); echo ' ← http_host (old logic)<br>';

$http_host_new_logic = parse_url($http_host, PHP_URL_HOST); 
var_dump($http_host_new_logic); echo ' ← http_host (new logic)<br>';


/*old logic */
if ($_SERVER['SERVER_PORT'] !== 80) {
	
	$orig_http_host = str_replace(':' . $_SERVER['SERVER_PORT'], '', $http_host);
	$fixed_http_host = str_replace(':' . $http_port, '', $http_host);
[...]

PR #16541, opened 2024-03-22
Fix working with non-standard ports

dimasites:
Fixed In default config, setup process and setup wizard.

This is port from 2.x branch, all context here: https://github.com/modxcms/revolution/pull/16455


theboxer (approved):
thx for the port to 3.x

Mark-H (approved):

PR #16552, opened 2024-04-03
Add OnContextInit System Event

christianseel:
### What does it do?
Adds a new core system event `OnContextInit`, invoked after each successful context initialization (includes those when `switchContext()` is called.

### Why is it needed?
We are having a project where we need to populate some data, that is context specific. To do so, we added a plugin to run each time a (new) context is initialized. Previously this was only possible via the event `OnMODXInit`, however this is only fired after the very first initialization of MODX. If you programmatically call `switchContext()` to init another context, there was no existing event to hook into.

### How to test
Add a plugin with a log line and activate the new system event `OnContextInit`. Log gets printed during modx initialization and when you programmatically call `switchContext()` with another context key.

### Related issue(s)/PR(s)
-

### DIFF
--- a/_build/data/transport.core.events.php
+++ b/_build/data/transport.core.events.php
@@ -1025,7 +1025,7 @@ $events['OnChunkFormDelete']->fromArray([
 
 
 /* Contexts */
-$events['OnContextInit']= $xpdo->newObject(modEvent::class);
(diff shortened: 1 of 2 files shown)


Mark-H (approved):

theboxer (approved):
Nice addition, thank you :)

PR #16555, opened 2024-04-11
Form Customization: Fully support comma-separated constraints lists

smg6511:
### What does it do?
In addition to correctly handling TV customization via FC, the proposed changes:
1. Optimize the FC Set Update processor code
2. Add validation to prevent incomplete constraint specification (_e.g._, user fills in the Constraint Field but no the Constraint, or vice versa) in the Set editing panel
3. Log a warning message if an incomplete constraint spec is saved from the grid. (Here, we don't want to trigger an error [preventing save] like we do in the stand alone editing panel because you can not recover from an error in an editor grid set to autosave.) This may point to a need for an intermediary validation state to be created where, like addError/addFieldError, you could trigger an addWarning/addFieldWarning on a field to alert users to a problem but not halt the save process.
4. Provides basic cleanup of user input for the three primary FC Set text fields (trim fields and help ensure proper constraint format)
5. Makes a tiny change to the Update controller to fix a deprecation warning issue re the usage of htmlspecialchars
6. Separate commit cleans up code quality/formatting issues in the php files

### Why is it needed?
TV re-labelling, default value alteration, etc is not currently working when providing multiple constraints. Also, the UI in this area is somewhat lacking in terms of validation and input handling (not cleaning user data to any degree).

### How to test
1. Create a one or more TVs
2. Create an FC Profile and Set
3. Experiment with all FC operations (moving and re-labeling fields, creating and renaming regions, setting default values, etc), verifying all works as expected for both standard fields and TVs
4. Try saving your Set with incomplete constraint information to verify validation works and messaging is clear

### Related issue(s)/PR(s)
This is a more comprehensive treatment related to a currently-open PR for 2.x (#16417)


theboxer (approved):
Works as described, thank you!

PR #16560, opened 2024-04-18
Update of the search processor for content-based search

Bournwog:
The search processor has been replaced with the one used in MODX 3
Now the search is performed based on the content
The commented-out functionality 'searchActions' has been removed

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1477,6 +1477,24 @@ $settings['rb_base_url']->fromArray(array (
   'area' => 'file',
   'editedon' => null,
 ), '', true, true);
+$settings['quick_search_in_content']= $xpdo->newObject('modSystemSetting');
+$settings['quick_search_in_content']->fromArray(array (
+  'key' => 'quick_search_in_content',
+  'value' => true,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'manager',
+  'editedon' => null,
+), '', true, true);
+$settings['quick_search_result_max']= $xpdo->newObject('modSystemSetting');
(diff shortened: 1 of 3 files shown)


Mark-H:
From a code review this looks fine for inclusion, just needs to get sufficiently tested, and probably targeting a 2.9 release instead of a 2.8.x?

Bournwog:
I've used it on several sites, it seems ok

Ruslan-Aleev:
@Bournwog You have a `quick_search_result_max` setting in your code, but you didn’t create it; MODX 2 doesn’t have it initially.

Bournwog:
@Ruslan-Aleev yes, I replaced one setting that is hardcoded with another, without setting a variable, if there is no variable, the hardcoded value will be taken

Ruslan-Aleev:
@Bournwog I understand how it works, yes. It just seems unfinished =) I would also add a setting.

SnowCreative:
I have mixed feelings about this. I find the new search function less useful for finding certain things. There is the "advanced search" that will search content, and I find that the more appropriate place for that. For example, I used to be able to type in just a number, and the quick search would find all resources, TVs, etc. with that ID number. With the new search number, that doesn't work; all sorts of other matches get mixed in. Advanced Search has an ID field, but that only searches resources, not TVs, chunks, etc.

Bournwog:
@Ruslan-Aleev 
add lexicon and settings fix

Ruslan-Aleev:
@Bournwog Leave the lexicon only for `en`, for other languages ​​translations will be pulled from crowdin.

Bournwog:
@Ruslan-Aleev
check it

Ruslan-Aleev (approved):

theboxer (approved):
works as described

PR #16561, opened 2024-04-19
Restore ability to render multi-select listbox TV values by selection order

### LINKED ISSUE #16495: MODX 3.0.4-pl - Listbox Multiselect
### Summary
The order that the elements is chosen is saved, but when you refresh backend and look at the resource with the Template Variable the order is not shown.

### Step to reproduce
Create a template variable with atleast two items. Use that template variable on a resource. Choose another order than the elements is showing in the list. Save. Refresh resource. Check template variable.

### Observed behavior
The initial order is shown in the list, not the selected order.

### Expected behavior
The selected order should be the order it is showing after a refresh. If the page is saved again, the initial order is saved and the elements are saved in "wrong" order.

### Environment
MODX version 3.0.4-pl

smg6511:
### What does it do?
Adds a new input option for this TV type to control the sorting of saved values when rendered to a Resource.

### Why is it needed?
A community member requested this functionality be brought back, pointing to possible use cases where preservation of the order of selections made is desirable.

### How to test
Create a Listbox (Multiple-Select) TV and experiment with changing the TV behavior via the new Preserve Entry Order option and verify that when this option is set to “Yes,” randomly selected options are shown in the order selected after saving and reloading. Conversely, verify the current default behavior remains intact (i.e., saved values appear in the same order defined by the list itself, no matter the order in which they were selected).

### Related issue(s)/PR(s)
Resolves #16495.


theboxer (approved):
NIce addition, works as described

jaygilmore:
I've confirmed this works as described. 👍🏻

Mark-H (approved):

modxcommunity:

https://community.modx.com/t/after-upgrading-to-3-x-log-filled-with-listbox-errors/8987/3

PR #16562, opened 2024-04-22
[2.x] modConnectorResponse — allow dynamic properties

marcuspoehls:
### What does it do?
Added `#[\AllowDynamicProperties]` PHP attribute to the `modConnectorResponse`.

### Why is it needed?
This change silences a deprecation warning in PHP 8.2 because of dynamic property assignments in this class. This also fixes issues with the package management, because the response now contains valid JSON instead of HTML.

### Related issue(s)/PR(s)
- related https://github.com/modxcms/xpdo/issues/256

### DIFF
--- a/core/model/modx/modconnectorresponse.class.php
+++ b/core/model/modx/modconnectorresponse.class.php
@@ -17,6 +17,7 @@ require_once MODX_CORE_PATH . 'model/modx/modresponse.class.php';
  * @package modx
  * @extends modResponse
  */
+#[\AllowDynamicProperties]
 class modConnectorResponse extends modResponse {
     /**
      * The base location of the processors called by the connectors.


opengeek:
What dynamic properties are being assigned to this class? It sounds like the source of the issue may need to be addressed rather than solving it with the AllowDynamicProperties attribute.

marcuspoehls:
The dynamic property is `$response`, assigned here: https://github.com/modxcms/revolution/blob/45e4b239f72a2352e788e4f2a3c3f0f456a5e175/core/model/modx/modconnectorresponse.class.php#L144C14-L144C28

opengeek:
> The dynamic property is `$response`, assigned here: https://github.com/modxcms/revolution/blob/45e4b239f72a2352e788e4f2a3c3f0f456a5e175/core/model/modx/modconnectorresponse.class.php#L144C14-L144C28

Thanks. That seems like an oversight. I'll dig in today or tomorrow and see if I can't resolve this by defining the expected properties.

smg6511:
I believe all that needs to be done is to add `$response = null;` to the base modResponse class props. The same goes for 3.x.

opengeek:
> I believe all that needs to be done is to add `$response = null;` to the base modResponse class props. The same goes for 3.x.

I'm not sure that's the only property being dynamically assigned here. Also, these should likely be private properties, but I'll need to investigate.

marcuspoehls:
@opengeek Thank you Jason for looking into this 🙂

opengeek:
See #16563 for my proposed solution to this warning. If you agree with the solution, please close this PR, @marcuspoehls.

marcuspoehls:
@opengeek Hey Jason, thank you for your pull request that refines the actual issue with the dynamic `$response` property. I’m closing this pull request in favor of yours https://github.com/modxcms/revolution/pull/16563

PR #16568, opened 2024-05-08
Fix Role Authority Editability

### LINKED ISSUE #16565: Changes to a Role’s definition do not carry through to related ACL entries
### Summary
In working on a solution for #16048, I discovered that changes to a Role definition—specifically its authority—do not update ACL entries that use the Role.

### Step to reproduce

1. Create at least one Role (via ACLs, Roles tab).
2. Assign this Role anywhere access permissions can be defined. An easy area to observe this is in the Media Source editing panel.
3. Change the Role's authority (via ACLs, Roles tab).
4. Return to the object for which you assigned permissions using this Role.

### Observed behavior
The ACL in question (where you changed its Role's authority) will no longer appear on the object.

### Expected behavior
Updates to Role authority should also update all objects that use the Role in an ACL.

### Note
This root cause of this is what I suspect is a bit of a data design flaw: A _value_ of the Role (the `authority`) is what's saved in every ACL's db entry, not its _primary key_ (`id` in this case). This complicates matters when we need to update an assigned Role. Now, we'll need to search through every ACL table for entries where the given `authority` value exists. Perhaps this can be redesigned in the future, but I know that wouldn't be possible until a feature release at the very least, and maybe not until the next major release. Something to consider...

### Environment
MODX 2.8.7 and 3.1.0-dev (and presumably any version of 3.x)

smg6511:
### What does it do?
Add new listener routine to restrict editing based on whether a Role is assigned to an ACL entry. Also:

1. Adds new UI feedback, both a lock icon (instead of pencil) on hover of authority as well as an alert dialog when double-clicking on locked authority
2. Ensures authority cannot be deleted in grid editor
3. Remove pencil icon for cells of non-editable Roles

### Why is it needed?
The ability to edit an assigned Role's authority leads to orphaned ACL rules that not longer show up in the manager, yet remain in the database. See the referenced issue below.

### How to test

1. Rebuild template (grunt build) and clear manager and browser caches
2. Create a few Roles under the Access Control Lists section
3. Assign at least one Role to any ACL entry of your choice
4. Verify that the assigned Roles' authority is locked in the Roles grid
5. Verify that unassigned Roles remain fully editable in the Roles grid (authority is unlocked)

### Note
The initial commit contains all substantive changes, while the follow up is just code-style/optimization oriented.

### Related issue(s)/PR(s)
Resolves #16565


Mark-H (approved):

Ibochkarev (approved):

opengeek:
As mentioned out of context in #16469, I think we need to consider just letting authority be presented/editable directly as an integer authority value when editing ACLs. Or don't allow ACLs to be edited at all, only allowing them to be created (setting the authority from a role) and deleted.

opengeek:
Actually, if we make modUserGroupRole.authority a unique index, we can rely on the link between a role and an authority. If no one has objections, I'll make this change, add the appropriate upgrade scripts, and add an upgrade test to ensure there are not multiple modUserGroupRole objects with the same authority before the upgrade to 3.1 can be executed. Then this PR can stand as is.

smg6511:
> make modUserGroupRole.authority a unique index

Makes sense to me ;-)

PR #16571, opened 2024-05-13
calling a modx instance via getInstance

webnitros:
`
\MODX\Revolution\modX::getInstance(\MODX\Revolution\modX::class)
`

### What does it do?
Calling a modx instance without having to pass it through __construct(modX &$modx)
this method will be useful in cases where you need to access an instance of the modX class from different parts of your application

### Why is it needed?
The problem is reusing classes from composer as well as shortening for writing code in various classes

### How to test
The feature was tested on real modx 2.8 projects with a high load and a large backend
\MODX\Revolution\modX::getInstance(\MODX\Revolution\modX::class)
phpunit tests pass successfully

### Related issue(s)/PR(s)
[#14555](https://github.com/modxcms/revolution/pull/14574)
https://github.com/modxcms/revolution/pull/16569#issuecomment-2108080253

## Functions app
I didn’t immediately add the app() function
to shorten the call code
I wanted to clarify where it would be better to add it?
It might be worth renaming it to appModx()
Although app() would obviously be better

```
if (!function_exists('app')) {

    /**
     * Get the available modx instance.
     * @return mixed|\MODX\Revolution\modX
     */
    function app()
    {
        return \MODX\Revolution\modX::getInstance(\MODX\Revolution\modX::class);
    }
}
```

## Example

```
require 'index.php';

# Old
$example1 = $modx->getOption('site_name');
# New
$example2 = \MODX\Revolution\modX::getInstance(\MODX\Revolution\modX::class)->getOption('site_name');
$example3 = app()->getOption('site_name');
```

ide code highlighting works immediately

CleanShot 2024-05-14 at 01 47 01

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -407,7 +407,7 @@ class modX extends xPDO {
      * @throws xPDOException
      */
     public static function getInstance($id = null, $config = null, $forceNew = false) {
-        $class = class_exists($id) ? $id : __CLASS__;
+        $class = __CLASS__;
         if (is_null($id)) {
             if (!is_null($config) || $forceNew || empty(self::$instances)) {
                 $id = uniqid($class);
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -407,7 +407,7 @@ class modX extends xPDO {
      * @throws xPDOException
      */
     public static function getInstance($id = null, $config = null, $forceNew = false) {
-        $class = __CLASS__;
+        $class = class_exists($id) ? $id : __CLASS__;
         if (is_null($id)) {
             if (!is_null($config) || $forceNew || empty(self::$instances)) {
                 $id = uniqid($class);
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -403,11 +403,11 @@ class modX extends xPDO {
      * @param array|null $config An optional array of config data for the instance.
      * @param bool $forceNew If true a new instance will be created even if an instance
      * with the provided $id already exists in modX::$instances.
-     * @return modX An instance of modX.
+     * @return static An instance of modX.
      * @throws xPDOException
      */
     public static function getInstance($id = null, $config = null, $forceNew = false) {
-        $class = __CLASS__;
+        $class = static::class;
         if (is_null($id)) {
             if (!is_null($config) || $forceNew || empty(self::$instances)) {
                 $id = uniqid($class);
--- a/_build/test/MODxTestHarness.php
+++ b/_build/test/MODxTestHarness.php
@@ -71,7 +71,7 @@ class MODxTestHarness {
                     if (!defined('MODX_CONFIG_KEY')) {
                         define('MODX_CONFIG_KEY', array_key_exists('config_key', self::$properties) ? self::$properties['config_key'] : 'test');
                     }
-                    $fixture = \MODX\Revolution\modX::getInstance(null, self::$properties["{$driver}_array_options"]);
+                    $fixture = modX::getInstance(null, self::$properties["{$driver}_array_options"]);
                     if ($fixture instanceof modX) {
                         $logLevel = array_key_exists('logLevel', self::$properties) ? self::$properties['logLevel'] : modX::LOG_LEVEL_WARN;
                         $logTarget = array_key_exists('logTarget', self::$properties) ? self::$properties['logTarget'] : (XPDO_CLI_MODE ? 'ECHO' : 'HTML');
--- a/_build/test/MODxTestHarness.php
+++ b/_build/test/MODxTestHarness.php
@@ -71,10 +71,7 @@ class MODxTestHarness {
                     if (!defined('MODX_CONFIG_KEY')) {
                         define('MODX_CONFIG_KEY', array_key_exists('config_key', self::$properties) ? self::$properties['config_key'] : 'test');
                     }
-                    $fixture = new modX(
-                        null,
-                        self::$properties["{$driver}_array_options"]
-                    );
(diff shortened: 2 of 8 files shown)


core/src/Revolution/modX.php
```diff
@@ -407,7 +407,7 @@ public static function toQueryString(array $parameters = [], $numPrefix = '', $a
      * @throws xPDOException
      */
     public static function getInstance($id = null, $config = null, $forceNew = false) {
-        $class = __CLASS__;
+        $class = class_exists($id) ? $id : __CLASS__;
```
opengeek: The `$id` here is NOT the class name. It is either a random id or a named key you provide. In the default case, it should be left null. This will ALWAYS return the existing, single instance that was first instantiated in the current request, as long as you do not also pass anything other than NULL to the `$config` parameter. In the case of connectors/index.php and manager/index.php, it is fine that they pass the additional `$config` parameter.

opengeek (changes requested):
This will work if you revert the change to modX::getInstance() and remove the class name string you are passing to it, passing null instead when additional parameters are involved.

core/src/Revolution/modX.php
webnitros: added changes

opengeek: Can revert this change to the `$class` variable here?

webnitros: > Can revert this change to the `$class` variable here?

So far I see that it will work the same
But if I want to use my own modx class, for example modY, is there an option with $class = class_exists($id)? $id : __CLASS__;

There will be a better job where I only need to specify my own class name in 3 index.php files

There is a choice, but the working option is both.

webnitros: For a better understanding of how getInstance will work
I agree, the old option is more suitable
To avoid surprises and to make the entry point into the application clear

webnitros: submitted changes

webnitros: in modx 2.8 it turns out you can make similar changes to index.php $modx = modX::getInstance();


tested it, works the same
both on modx 3 and modx 2.8

For modx 3 $modx = \MODX\Revolution\modX::getInstance();
For modx 2.8 $modx = modX::getInstance();

opengeek: If you want to be able to extend modX and have it use the actual classname of your extended class to generate the uniqid prefix (which is all this is doing), we can change that to a late static binding, e.g. `static::class`.

webnitros: > If you want to be able to extend modX and have it use the actual classname of your extended class to generate the uniqid prefix (which is all this is doing), we can change that to a late static binding, e.g. `static::class`.


No)))
The fact is that the experience of using this approach was precisely in the call $modx = modY::getInstance('modY');
Based on this experience

The option that is now in the pull request is what we need.
The main idea is to call an instance (for example) in one tab through a Live template (PhpStorm)
$modx = modY::getInstance('modY');
and start working from any class
with code highlighting ($modx->) immediately without any comments or other additional code.

opengeek: The code highlighting works from the class itself, not from the string you pass to the function. I'm lost on what you are talking about here. Regardless of what string exists as the class, the unique id that is generated with that string as the prefix is what defines the instance that will be retrieved on additional calls to the method when you DO NOT PROVIDE an id. If you provide an id, the class variable is irrelevant. But, if you call getInstance from a class modY that extends modX, class is relevant. This is the ONLY time it is relevant, and in that case, it needs to use late static binding to automatically set the class that is instantiated regardless of what the `$id` is.

webnitros: CleanShot 2024-05-15 at 00 52 21

I'm talking about modifying config.core.php
if I connect in config.core.php autoloader.php from composer
then when called it will extend modX and then I can access it using

$modx = modY::getInstance('modY');


I don’t focus on this, this experience allowed me to fully use the namespace "App\\": "src/"
where I never passed public function __construct(modX $modx)

There was always a call: modY::getInstance('modY');

Let me explain, I used the modY class because it was a combat project, and it was a risk. But the risk was justified, everything works fully

You showed that modX can be used without specifying $id

opengeek: Again, you don't need to pass 'modY' to the getInstance() function if we use `static::class` (late static binding) instead of `__CLASS__` to assign the `$class` variable in modX::getInstance()… modY::getInstance() will automatically know it is modY.

opengeek: I can make that adjustment in conjunction with yours if you are okay with me doing that. I will also adjust the return type so IDEs will also know it returns the class from the late static binding.

webnitros: > I can make that adjustment in conjunction with yours if you are okay with me doing that. I will also adjust the return type so IDEs will also know it returns the class from the late static binding.

ok

opengeek:
Thank you for your contribution, @webnitros! Before we can move forward, please [sign the MODX Contributor License Agreement (CLA)](https://modx.com/community/contribute/cla/). This is to ensure MODX has your written permission to distribute projects containing your contributions under the appropriate license.

Please make sure the CLA has been signed for GitHub user(s):  @webnitros 

Once you've signed, please reply with a comment letting us know so I can verify. Then, I will work on getting this integrated.

webnitros:
> Thank you for your contribution, @webnitros! Before we can move forward, please [sign the MODX Contributor License Agreement (CLA)](https://modx.com/community/contribute/cla/). This is to ensure MODX has your written permission to distribute projects containing your contributions under the appropriate license.
> 
> Please make sure the CLA has been signed for GitHub user(s): @webnitros
> 
> Once you've signed, please reply with a comment letting us know so I can verify. Then, I will work on getting this integrated.

ready

opengeek (approved):

PR #16578, opened 2024-06-14
Support dot followed by space in directories

### LINKED ISSUE #16351: Filesystem stripping periods from folder names
### Summary

I have a client who likes to use periods in the file structure, so their images in the filesystem have paths like "assets/images/3. Category/SomeImage.png". This has worked fine for many years. However, recently, in 3.0.2, if you click that "3. Category" folder in the Filesystem browser in the admin, the hierarchy no longer expands to show the images inside the folder. Instead, you get this error:

Unable to retrieve the mime_type for file at location: assets/images/3 Category. finfo::file(/www/assets/images/3 Category): failed to open stream: No such file or directory

Notice that the error has stripped out the period after "3" -- it is looking for the folder "3 Category" instead of "3. Category", and not finding it. The folder and files are all still there and able to be used as src files on the frontend via direct URL (ex. "www.example.com/assets/images/3.%20Category/SomeImage.png"), but the backend manager can no longer edit or see these files. This whole setup works fine in an older, 3.0.1 development setup I have kicking around. One other different behavior I noticed is that if you try to create a new folder through the manager UI, say "4. Example", that will generate a "4. Example" folder in my 3.0.1 setup, but on 3.0.2 it generates a "4 Example" folder with period stripped.

This may not be a version issue -- it is very possible that a system setting somewhere has been changed to disallow periods in folder names, but if so, I've looked over the FURL settings and I don't see any specific period restrictions. Pasting my FURL settings below.

image

### Step to reproduce
Open Filesystem, click a folder with a period in its name.
### Observed behavior
This error is shown:

image

### Expected behavior
The folder expands and shows child items in the hierarchy.

### Environment
MODX 3.0.2

smg6511:
### What does it do?
Refined regex in Browser class to allow dot-space (e.g., `1. Folder Name`) within directory, but not when directory/file begins with a dot (e.g., .htaccess, .hidden-dir, etc).

### Why is it needed?
Responding to community member request.

### How to test
Create various directories and files with dot(s) and space(s) within to verify desired behavior.

### Related issue(s)/PR(s)
Resolves #16351

### DIFF
--- a/core/src/Revolution/Processors/Browser/Browser.php
+++ b/core/src/Revolution/Processors/Browser/Browser.php
@@ -143,8 +143,11 @@ abstract class Browser extends Processor
     {
         $file = rawurldecode($file);
         $file = strip_tags($file);
-        $file = preg_replace('#\.(?![\w\-\~])#u', '', $file);
-        $file = preg_replace('#\/{2,}#', '/', $file);
+        if (strpos($file, '.') === 0) {
+            $file = preg_replace('/^(\.)([\s\/]*)(.*)/', '$1$3', $file);
+        }
+        $file = preg_replace('/\.(?![\w\-\~\s])/u', '', $file);
+        $file = preg_replace('/\/{2,}/', '/', $file);
         $file = strip_tags($file);
 
         return $file;


theboxer (approved):
seems to work as described :)

PR #16579, opened 2024-06-16
Reveal current username in main menu bar (User menu)

### LINKED ISSUE #16525: Show username in menu
### Summary
Menu item labels, which are located in the "User Navigation" menu area, are hidden in the styles.
Don't know why this was done, but ok.

https://github.com/modxcms/revolution/blob/87a62645573bd3fcde291fc269db3efd987c4962/_build/templates/default/sass/_navbar.scss#L86

However, it is worth always displaying the username, because to find it out you have to click on the menu, which is not convenient, **it's just bad UX**. 

user_1
user_2

There was a discussion (https://github.com/modxcms/revolution/issues/15595) that long name in login may not fit, but we can add `word-wrap: break-word;` this will fix the problem. And it’s strange to add long names to login.

user_3

Another option is to add a title="" attribute with login to avatar, at least you don’t have to click user menu item. The UX will improve a little, but in my opinion it would be better to display the name in the menu.

user_4

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/15595

smg6511:
### What does it do?
Unhides username in menu bar, restricting to one line by truncating long names with ellipses.

### Why is it needed?
UX improvement request.

### How to test

1. Run `grunt build` at `_build/templates/default` to update CSS.
2. Test with two manager Users — one with a short username, the other with a long username. Verify that ones ~7-8 characters long show in full and the longer one displays on a single line (e.g., 'user1@m...').

### Related issue(s)/PR(s)
Resolves #16525.


theboxer (approved):
works good, thanks :)

Ruslan-Aleev (approved):

Ibochkarev (approved):

PR #16582, opened 2024-06-18
Add secondary regex option for transliterated file uploads

### LINKED ISSUE #16551: [3.x] File extension drops in media manager when friendly url pattern restricts dot
### Summary
I want to restrict dot in resource aliases, but related setting affects uploaded files.

### Step to reproduce
On fresh modx installed, add `.` to `friendly_alias_restrict_chars_pattern` system setting default value _(for exclude dot in resource aliases in my case)_
Setting value become to ```/[\0\x0B\t\n\r\f\a&=+%#<>"~:`@\?\[\]\{\}\|\^'\\.]/```

Then try to upload file with name `te  st.txt` via media manager.

### Observed behavior
After upload, you get `te-sttxt` (without dot before .txt extension)

If i disable `upload_translit` setiing, file name become to `te  st.txt` → unsafe, non-friendly-url compatible an poor looking. May contain special chars and etc.

### Expected behavior
Dots in file extensions (and other dots may be via special pattern) preserved when upload or remane file via media manager. 

It will be great to set up RegExp pattenrs for uploaded file names and for fliendly URLs separately!

### Environment
MODX 3.0.4. `upload_translit` setiing activated (by default) and `friendly_alias_restrict_chars` = `pattern` (by default)

smg6511:
### What does it do?
Adds a new system setting that allows transliterated uploads to override the regex restriction used for file names (currently the same as with Resource aliases).

### Why is it needed?
See the referenced issue. The request made there would be the most common use-case, but I'd imagine there could be others where a user wants the naming convention to be different between file names and Resource aliases.

### How to test

1. Rebuild the core via `/_build/transport.core.php`
2. Run `/setup` (required to update the db based on the new build)
3. Test differing settings for Resources (`friendly_alias_restrict_chars_pattern`) and the new setting for files (`upload_translit_restrict_chars_pattern`). On a fresh install both settings regex patterns will be the same, so be sure to change at least one and verify the desired effect.

### Related issue(s)/PR(s)
Resolves #16551

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -1148,7 +1148,15 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             }
 
             if ((bool)$this->xpdo->getOption('upload_translit')) {
-                $newName = $this->xpdo->filterPathSegment($file['name']);
+                $uploadRegex = $this->xpdo->getOption('upload_translit_restrict_chars_pattern');
+                $options = !empty($uploadRegex)
+                    ? [
+                        'friendly_alias_restrict_chars' => 'pattern',
+                        'friendly_alias_restrict_chars_pattern' => $uploadRegex
+                    ]
+                    : []
+                    ;
+                $newName = $this->xpdo->filterPathSegment($file['name'], $options);
 
                 // If the file name is different after filtering, call OnFileManagerFileRename
                 // so the change can be tracked by plugins
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1896,6 +1896,15 @@ $settings['upload_translit']->fromArray([
   'area' => 'file',
   'editedon' => null,
 ], '', true, true);
+$settings['upload_translit_restrict_chars_pattern'] = $xpdo->newObject(modSystemSetting::class);
+$settings['upload_translit_restrict_chars_pattern']->fromArray([
+  'key' => 'upload_translit_restrict_chars_pattern',
+  'value' => '/[\0\x0B\t\n\r\f\a&=+%#<>"~:`@\?\[\]\{\}\|\^\'\\\\]/',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'file',
+  'editedon' => null,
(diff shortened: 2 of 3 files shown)


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -1148,7 +1148,15 @@ public function uploadObjectsToContainer($container, array $objects = [])
...
-                $newName = $this->xpdo->filterPathSegment($file['name']);
+                $uploadRegex = $this->xpdo->getOption('upload_translit_restrict_chars_pattern');
+                $options = !empty($uploadRegex)
+                    ? [
+                        'friendly_alias_restrict_chars' => 'pattern',
+                        'friendly_alias_restrict_chars_pattern' => $uploadRegex
+                    ]
+                    : []
+                    ;
+                $newName = $this->xpdo->filterPathSegment($file['name'], $options);
```
opengeek, line 1159: I wonder if we shouldn't have a dedicated modMediaSource method instead of using filterPathSegment()…

smg6511: I don't _think_ so, as we'd end up with a lot of duplicated logic, especially taking the translit functionality into account. What I can see is moving the meat of the `filterPathSegment` fn out of `modResource` and into the one in the base `modX` class; I'm actually curious to know why it's wired the way it currently is (with `filterPathSegment` residing in both classes). 

(Not for this PR): Looking ahead, perhaps fns like this should be moved into their own (perhaps static) utility class(es)? Additionally, `filterPathSegment` does two distinct things IMO — replaces a set of illegal chars and transliterates. Because transliteration is optional, it might be best to separate that logic out into a new fn.

opengeek: Let's keep future discussion out of these comments. These methods were intended for dealing with Resource aliases only. They were never meant for file management. The modX helper method was just that—a convenience for calling the static method on the modResource class. I'm not sure why it was reused in media sources.

opengeek: Just ignore my original comment and we'll do it like this.

dimasites:
Hi @opengeek please merge it, problem too often floats in community chats! Thx

opengeek (approved):

PR #16586, opened 2024-06-22
Add Help! label to Menu Item

### LINKED ISSUE #13485: Help! Links must have discernible text

jaygilmore:
Added Help! label to the Help Menu Item that was previously represented by the just an icon.

This was to fix for #13485

### What does it do?
Provides accessible label for the Help button in the header #13485 

### Why is it needed?
The current link has no link text and no value for screen readers

### How to test
This may require a build and or rerunning setup. 

### Related issue(s)/PR(s)
Should resolve #13485 by adding aria-label and corresponding text via lexicon to the `a` of the Help button in the header.

### DIFF
--- a/_build/data/transport.core.menus.php
+++ b/_build/data/transport.core.menus.php
@@ -511,7 +511,7 @@ $userNavMenus[2]= $xpdo->newObject(modMenu::class);
 $userNavMenus[2]->fromArray([
   'menuindex' => 8,
   'text' => 'about',
-  'description' => '',
+  'description' => 'about_desc',
   'parent' => 'usernav',
   'permissions' => 'help',
   'action' => 'help',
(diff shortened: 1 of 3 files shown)


jaygilmore:
I don't know if this needs the refactoring due to the PSR level errors as I just followed the code pattern of the original file. Can do if required.

PR #16587, opened 2024-06-24
Make authority a unique index on modUserGroupRole

opengeek:
### What does it do?
Adds a unique index to the modUserGroupRole.authority field.

### Why is it needed?
modUserGroupRole records are linked to ACLs via the authority value, so those MUST be unique or referential integrity is lost.

### How to test
Run the transport build, execute the setup, and verify that after adding a modUserGroupRole record with a non-unique authority value, the upgrade stops with a failure describing what you need to do in order to resolve the situation before re-running the setup. After removing the non-unique record, verify that setup then runs successfully and a unique index is added to the authority column.

### Related issue(s)/PR(s)
This is related to and should be adopted along with #16568 — see discussion on that PR.


modxcommunity:

https://community.modx.com/t/access-denied-for-users-after-upgrading-from-modx-3-0-6-to-3-1-0-using-upgrade-add-on/8632/6

PR #16596, opened 2024-08-02
Fix max parser iterations for resources without a template

halftrainedharry:
### Why is it needed?
Currently for resources without an assigned template ("Uses Template" = `(empty)` in the manager), the parser is called with a `$depth` (maximum iterations) of `0`.

---

The second parameter of the `getOption()` function is an array of `$options` and not the default value:
https://github.com/modxcms/xpdo/blob/ffa5d883ef44bd684db748a3e454071672913e48/src/xPDO/xPDO.php#L705-L711


opengeek:
Nice catch, @halftrainedharry!

opengeek (approved):

PR #16600, opened 2024-08-16
Big trash refactor

### LINKED ISSUE #14190: Don't work middle mouse button click on Trash manager button
### Summary
Don't work **middle mouse** and **Ctrl + Left mouse** buttons click on Trash manager button

### Expected behavior
Open the Trash manager in new tab

### Environment
MODX 2.7.0-pl

GulomovCreative:
### What does it do?

1. Adds new "Trash" menu item to the top menu
2. Fixes #14190
3. Fixes trash button icon color change when its inactive
4. Removes unnecessary styles
5. Adds a purge button to the edit resource page
6. The resource tree and trash button state are now updates when a resource is deleted/undeleted on the resource update page
7. Fixes trash button tooltip information when purging resources on the trash page
8. Now when a resource have been delete/undelete in the resource tree, the state of the resource update form (checkbox and buttons) updates
9. Fixes undefined text of the trash button tooltip
10. Adds purge resource menu item to the context menu in the resource tree

### Why is it needed?

There are many errors when deleting/restoring/purging resources

### How to test

1. —
2. Click by the middle mouse button on the trash button on the toolbar
3. Hover over the trash button when its inactive, before the icon changed color, now when inactive it does not
4. —
5. Open the edit resource page, delete it and you will see a new purging button
6. If you delete or restore a resource on the edit page, the resource tree and the trash button in the toolbar remain unchanged. Now the resource tree and the trash button will change their state
7. On the trash page when purging resources, the state of the trash button in the toolbar was not updated, now it will be
8. Create a resource, open its edit page and delete the resource **in the resource tree**, the state of the checkbox and buttons in the edit form will not change. Now everything works as it should.
9. If you delete a resource and when you hover the trash button, it will be undefined in the tooltip. Now the information will be correct
10. Open the context menu of resource tree, delete the resource and open again

### Related issue(s)/PR(s)
#14190


theboxer (approved):
Looks good, thank you!

The only thing that's missing is a migration script that will add the new menu during an upgrade.

theboxer (changes requested):
Looks good, thank you!

The only thing that's missing is an update script to add the new menu during upgrade.

rthrash:
@GulomovCreative do you think you’ll have time to get the update script for adding the new menu during an upgrade done?

core/lexicon/en/resource.inc.php
```diff
@@ -45,6 +45,8 @@
 $_lang['resource_createdon'] = 'Created On';
 $_lang['resource_delete'] = 'Delete';
 $_lang['resource_delete_confirm'] = 'Are you sure you want to delete the resource: "[[+resource]]"?<br />NOTE: Any child resources will also be deleted!';
+$_lang['resource_purge'] = 'Erase resource';
```
Ruslan-Aleev, line 48: Here the word "resource" is unnecessary, you can leave only "Erase". Because for the other menu items "resource" is not specified, and the clarification here is unnecessary.

theboxer: oh, yes!, nice catch

rthrash: Or "Permanently erase"?

GulomovCreative: I just took the same text that is on the [trash page](https://github.com/modxcms/revolution/blob/9a3f82cee664c1945aeafd1b232a871c88f7f88a/core/lexicon/en/trash.inc.php#L24)

https://github.com/modxcms/revolution/blob/9a3f82cee664c1945aeafd1b232a871c88f7f88a/core/lexicon/en/trash.inc.php#L24

GulomovCreative:
> @GulomovCreative do you think you’ll have time to get the update script for adding the new menu during an upgrade done?

@rthrash Can I see some examples? I don't know how it works

theboxer:
@GulomovCreative here are current upgrade scripts https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/mysql/3.1.0-pl.php you can check those or older ones for samples

theboxer:
Added the upgrade script & adjusted the lexicon for context menu in the tree

theboxer (approved):

Ibochkarev (approved):

core/lexicon/en/resource.inc.php
Ruslan-Aleev: @GulomovCreative I would still remove the clarification, since this is text for the resource tree, and here the clarification looks strange.

tree_lang

theboxer:
@opengeek rebased to 3.x

PR #16602, opened 2024-08-20
Fix manager messages grid permissions and other improvements

smg6511:
### What does it do?
This fixes an incorrect reference to the usergroup view permissions to ensure that recipients list is available to users. A second commit makes a number of code quality improvements and a handful of style improvements to the Create window and the message display in the grid.

### Why is it needed?
The ability to create messages is potentially unavailable under certain permissions scenarios (namely one where someone has permission to view groups or roles, but not individual users).

### How to test
Make changes to a user's `view_user`, `view_role`, and `usergroup_view` and verify that when at least one of these permissions is given, the user is able to create new messages (the Create button should be enabled). 

### Related issue(s)/PR(s)
None


theboxer (approved):
Looks good!

PR #16610, opened 2024-09-03
3.x Fix TV Category not found issue in Template and TV editing panels

smg6511:
### What does it do?
Changed the (grid filtering) category combo configs to:
1. Prevent submission of their values on post
2. Not create a hidden submit field
3. Ignore the initial request value of "undefined" when creating a new Template/TV

The first commit contains the functional changes, while the remaining ones provide code style and quality fixes.

### Why is it needed?
A change somewhere in the current development branch created a condition where form panel pages with both a Category field and a Category grid filter would not save properly. (This is seen only in the current 3.1.0-dev branch.)

**Note:** I added the "Urgent" label to this PR because we'll need it incorporated into 3.1 _before_ its -pl release, as the issue it solves prevents proper operation of the referenced panels.

### How to test

1. Before pulling down this PR, observe that when trying to create a new TV or Template, you are prevented from doing so. The Category field will always have an error, not matter what you do. Also, if you go to the Template Access tab of a TV and filter the grid by Category using the dropdown and then immediately save, note how the Category of the TV itself will have been changed to the one you chose in the grid filter (not what you want).
2. After applying this PR, create and edit a Template and a TV to verify you are able to save and update each as expected, and that filtering on the TV/Template Access grids within each does not affect the saved Element's Category.

### Related issue(s)/PR(s)
None, I discovered this issue when I was working on another PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -52,6 +52,10 @@ module.exports = {
         'one-var': ['error', 'consecutive'],
         'prefer-arrow-callback': 'warn',
         'prefer-rest-params': 'warn',
+        'semi-spacing': ['warn', {
+            before: false,
+            after: true
+        }],
         'semi-style': ['warn', 'last'],
         'space-before-function-paren': ['error', 'never']
     }
--- a/manager/assets/modext/widgets/element/modx.grid.template.tv.js
+++ b/manager/assets/modext/widgets/element/modx.grid.template.tv.js
@@ -89,7 +89,9 @@ MODx.grid.TemplateTV = function(config = {}) {
                 xtype: 'modx-combo-category'
                 ,itemId: 'filter-category'
                 ,emptyText: _('filter_by_category')
-                ,value: MODx.request.category || null
+                ,value: MODx.request.category !== 'undefined' ? MODx.request.category : null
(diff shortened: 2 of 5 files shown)


theboxer:
@smg6511 Sorry, I don't really follow how to reproduce the issue and test if it's resolved...

smg6511:
@theboxer Hey - apologize that it's unclear. When running 3.1.0-dev (_not_ 3.0.6-dev), if you try to create a Template or TV, you will be prevented from doing so; there will always be an error on the Category field saying it cannot be found (even if you select one from the dropdown). Assuming you see this too, the PR fixes that problem.

theboxer (approved):
Fixes the issue, thx :)

smg6511:
Turns out the issue was my bad: I'd, in a couple places, removed the config property in question in a recently-merged PR ... not realizing that the base combo class was setting the hiddenName prop that would eventually cause the conflict! Oh well :-/

smg6511:
@opengeek - Note that when you get to merging this, you probably _won't_ want to backport to 3.0.6 as the grid filtering persistence feature (that introduced the issue I'm fixing) has only been incorporated into the 3.1 branch.

Issue #16420, opened 2023-04-12
Handling .ico files in media sources


MaticSulc:
MODX 3.0.3-pl, PHP 8.1.17.

When adding a favicon.ico file in the website root, media source preview doesn't work for it and it keeps filling the error log.
`PHP warning: exif_read_data(favicon.ico): File not supported`

I can simply make the mediasource ignore it, by adding it in the `skipFiles` setting, but since .ico isn't under `imageExtensions`, I don't expect it to generate a preview. Can't replicate this in MODX 2.x.x.

halftrainedharry:
> but since .ico isn't under `imageExtensions`, I don't expect it to generate a preview.

It generates a preview because the string 'image' is in the mime type.

https://github.com/modxcms/revolution/blob/69a7d6d14f84151ced04a552678657456c05b9cc/core/src/Revolution/Sources/modMediaSource.php#L2414-L2417

---

The same PHP warning occurs also for other image types (like e.g. PNG or WebP) when PHP 8 is used.
Probably the code has to be changed here, to run `exif_read_data` only for image types that support it.

https://github.com/modxcms/revolution/blob/69a7d6d14f84151ced04a552678657456c05b9cc/core/src/Revolution/Sources/modMediaSource.php#L2260

---

There is also a similar topic in the MODX forum: https://community.modx.com/t/exif-read-data-file-not-supported/6572

PR #16615, opened 2024-09-18
Fix error suppression for PHP 8+

halftrainedharry:
### What does it do?
Adds a check to the custom error handler function, to see if a warning has to be handled.

### Why is it needed?
MODX has a custom error handler function. When the error control operator (`@`) is used (e.g. `@exif_read_data()`), the custom error handler still gets called.

With **PHP 7.4,** when the error was suppressed by the `@` operator, `error_reporting()` called inside the custom error handler always returned 0. This gets checked here:

https://github.com/modxcms/revolution/blob/9e7bb63615f8a9921e657027501f2aaf1be07454/core/src/Revolution/Error/modErrorHandler.php#L61-L63

With **PHP 8**, this is no longer the case:
> As of PHP 8.0.0, it returns the value of this (bitwise) expression: E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE.

https://www.php.net/manual/en/language.operators.errorcontrol.php

### How to test

For example:
- Use PHP 8.0 (or higher)
- Open the media browser and navigate to a folder with a picture that doesn't have EXIF headers.
- Make sure, that the warning `PHP warning: exif_read_data(...): File not supported` doesn't get logged to the MODX error log anymore.

### Related issue(s)/PR(s)
#16420

### DIFF
--- a/core/src/Revolution/Error/modErrorHandler.php
+++ b/core/src/Revolution/Error/modErrorHandler.php
@@ -58,7 +58,7 @@ class modErrorHandler
      */
     public function handleError($errno, $errstr, $errfile = null, $errline = null, $errcontext = null)
     {
-        if (error_reporting() == 0) {
+        if (error_reporting() == 0 || ($errno === E_WARNING && (error_reporting() & E_WARNING) === 0 )) {
             return false;
         }
 
--- a/core/src/Revolution/Error/modErrorHandler.php
+++ b/core/src/Revolution/Error/modErrorHandler.php
@@ -58,7 +58,7 @@ class modErrorHandler
      */
     public function handleError($errno, $errstr, $errfile = null, $errline = null, $errcontext = null)
     {
-        if (error_reporting() == 0 || ($errno === E_WARNING && (error_reporting() & E_WARNING) === 0 )) {
+        if (error_reporting() == 0 || !(error_reporting() & $errno)) {
             return false;
         }
 


modxcommunity:

https://community.modx.com/t/exif-read-data-file-not-supported/6572/17

opengeek:
Excellent contribution, @halftrainedharry! I'll get this reviewed and integrated as soon as possible. I'm guessing this would also be helpful as a bug fix in 2.x.

opengeek:
@halftrainedharry — Is this suppressing all E_WARNING errors, even if the @ operator is not used?

halftrainedharry:
> Is this suppressing all E_WARNING errors, even if the @ operator is not used?

No.

Here is a test snippet:
```
<?php
include MODX_CORE_PATH . "FileThatDoesntExist_01.php";
@include MODX_CORE_PATH . "FileThatDoesntExist_02.php";
```
`include` without the `@` in the first line, still adds a warning to the MODX error log.

---

With `error_reporting = E_ALL` in the "php.ini", when `include` is used for a file that doesn't exist, the error handler function is called and `error_reporting()` in the handler function returns 32767 (= E_ALL).
When `@include` is used, `error_reporting()` in the handler function returns 4437 (= E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE).

core/src/Revolution/Error/modErrorHandler.php
```diff
@@ -58,7 +58,7 @@ function __construct(modX &$modx, array $stack = [])
      */
     public function handleError($errno, $errstr, $errfile = null, $errline = null, $errcontext = null)
     {
-        if (error_reporting() == 0) {
+        if (error_reporting() == 0 || ($errno === E_WARNING && (error_reporting() & E_WARNING) === 0 )) {
```
opengeek: I believe this needs to be ` || !(error_reporting() & $errno)` so that suppression works with all non-fatal errors, not only E_WARNING. For example, E_NOTICE would not be suppressed using your solution.

opengeek (changes requested):

opengeek (approved):

PR #16620, opened 2024-09-24
Fix User Date of Birth Validation and Processing

smg6511:
### What does it do?
Replace usage of `strtotime` with `DateTimeImmutable::createFromFormat` for better format compatibility.

### Why is it needed?
Certain date formats, such as m-d-Y, are incompatible with `strtotime` and will cause the value of input fields where it's currently used (such as in a User's Profile) to fail.

### How to test
Set the system setting `manager_date_format` to `m-d-Y` (or any other format where the month and day would be programmatically ambiguous), clear caches, and verify that a User's DOB can be added/updated as expected. Note that there are two distinct places to check this: under Manage/Users/[User] and via the logged in User's Profile page (at `/manager/?a=security/profile`)

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/core/src/Revolution/Processors/Security/Profile/Update.php
+++ b/core/src/Revolution/Processors/Security/Profile/Update.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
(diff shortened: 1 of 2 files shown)


theboxer (approved):
It'd be slightly better for the review process if changes unrelated to the issue itself would be in separate commits (like reformatting etc.), otherwise it's hard to spot what actually changed.

PR #16646, opened 2024-11-20
PHP 8.4 compatibility updates

opengeek:
### What does it do?
Removes usage of deprecated features, updates minimum PHP support to >= 7.4, and updates PHPUnit to version 8.5.

### Why is it needed?
Resolves various deprecation warnings for PHP 8.4 compatibility.

### How to test
Test everything (including unit tests) in PHP 8.4

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['5.6', '7.0', '7.1', '7.2', '7.3']
+        php-version: ['7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
 
     steps:
       - name: Checkout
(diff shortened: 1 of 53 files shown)


Mark-H:
Once merged we'll also need to update the docs as those still mention older server requirements.

Mark-H (approved):

opengeek:
> Once merged we'll also need to update the docs as those still mention older server requirements.

I don't think this is going to get merged, honestly. I'd like to consider 2.x closed to feature requests and in order to actually support 8.4, we would need to update to a newer version of Smarty. This would cause BC breaks, and I don't want to go down that road.

smg6511:
@opengeek - I figure you took this off the table long ago, so might as well fully close.

PR #16647, opened 2024-11-21
Fix separator for menu item without children

### LINKED ISSUE #16645: [3.x] Manager Menu Custom Item without Children missing classname
### Summary

When I add a custom menu item to the main manager navigation without children, there is no border separator:

Is there a reason for this? I think this line should always be 'top' for border between items:

https://github.com/modxcms/revolution/blob/ab804dff5ddfa7cdf2a989bc99acfaf108ca0149/manager/controllers/default/header.php#L182

### Environment

MODX 3.0.5

theboxer:
### What does it do?
Adds separator for menu items without children

### Why is it needed?
Missing separator

### How to test
Create menu entry without children

### Related issue(s)/PR(s)
Resolves #16645


manager/controllers/default/header.php
```diff
@@ -179,7 +179,7 @@ public function buildMenu($name, $placeholder)
                 $description = '<span class="description">'.$menu['description'].'</span>'."\n";
             }
 
-            $top = !empty($menu['children']) ? ' top' : '';
+            $top = ' top';
```
smg6511:
```suggestion

```

manager/controllers/default/header.php
```diff
@@ -179,7 +179,7 @@ public function buildMenu($name, $placeholder)
...
             $position = $idx <= 2 && $placeholder == 'navb' ? 'down' : 'up';
 
             $menuTpl = '<li id="limenu-'.$menu['id'].'"class="menu-'.$position.$top.'">'."\n";
```
smg6511:
```suggestion
            $menuTpl = '<li id="limenu-' . $menu['id'] . '" class="menu-' . $position . ' top">' . "\n";
```

smg6511: Additionally, replace L192 with:
```php
$menuTpl .= '<a href="?a=' . $menu['action'] . $menu['params'] . '"' . $onclick . $title . $ariaLabel . '>' . $label . $description . '</a>' . "\n";
```

Then, finally, in the `_navbar.scss` file, change L111 from `.top:not(#modx-manager-search-icon) {` to:
```css
> li {
```

smg6511 (approved):
So, this definitely corrects the issue as-is. Thus I went ahead and approved.

I made some suggestions, nonetheless, because I noticed that having the `$top` var serves no purpose so you might as well simply add the `top` class to the `<li>` directly (and, the `top-link` css class is not used in 3.x). Plus, really we want to select all top-level nav `<li>`s to apply the top border, so the selector in the scss file can be simplified (as noted in the second comment).

Anyway, up to you if you want to incorporate these changes. If you think it's too nit-picky, no problem ;-)

theboxer:
@smg6511 thx for the suggestions, added them all

smg6511 (approved):

Ibochkarev (approved):

Mark-H (approved):

opengeek:
This PR includes the compiled assets. Can you remove those @theboxer?

PR #16650, opened 2024-11-24
Add hostname to s3 source

Bournwog:
There is an option to connect to services other than Amazon


Mark-H:
For what it's worth, this is already supported in MODX 3. I don't object to some QOL in MODX 2 if people want this, but then we may need to consider the upgrade path from 2->3 where the property is called endpoint.

Bournwog:
The endpoint is not the same as the hostname. For example:
- Hostname: s3.example.com
- Endpoint: https://s3.example.com
After migration, we cannot determine whether the user is using HTTP or HTTPS

Mark-H:
Alright, I must not know enough about this particular S3 feature then.  Can you provide additional information on 1) how to test and 2) why this is needed? It may also be necessary to help [provide additional documentation](https://github.com/modxorg/docs) once approved. 

For reference, here is the template we ask people to use when submitting a PR to make sure there is sufficient context in order to review it. 

```
### What does it do?

### Why is it needed?

### How to test

### Related issue(s)/PR(s)
```

Issue #16507, opened 2024-01-05
MODExt Grid column editor tooltips are displayed even the config option editable is disabled


revocarne:
### Summary
In a grid, the icon pen is displayed even the config options editable is disabled.
It’s really unpleasant when for developing a CMP.

### Step to reproduce
Still when disable the config options editable in a column.

### Observed behavior
Easy to fix the "bug"
In the file "revolution/manager/assets/modext/widgets/core/modx.grid.js" line 475 and 1176

replace "if (c[i].editor) {"
by : "if (c[i].editor && c[i].editable) {"

### Environment
All

PR #16653, opened 2024-11-29
Editor grids permissions refactor, enhancements, and fixes

smg6511:
### Related PR
See also #15919. This is a re-packaging of that PR; this new PR seeks to cover all editor grids across the application and separates out some work (for a future PR) that had been begun around internationalizing core names and descriptions within some of the grids.

### Related Issues
#14929, #16507


opengeek:
@smg6511 Is this close to being ready?

smg6511:
@opengeek Yes, very close - been hammering away at it. It should be ready for review by early next week.

smg6511:
@opengeek @theboxer @Mark-H - Alright, I believe this is _finally_ ready. Hope folks can carve out some time to go through it all ;-) !

rthrash:
@smg6511 this is a big one—kudos on getting it to this stage! Since it's fresh on your mind and touches so many areas, it will take a while to fully review I suspect (was just discussing this with @jaygilmore and @opengeek). 

Any suggestions or checklists of all the various nooks and crannies that should be tested to get this one merged?

smg6511:
> Any suggestions or checklists of all the various nooks and crannies that should be tested to get this one merged?

Ok, here's generally how I'd suggest going through this:

### Initial Runthrough
I overhauled the base grids classes (in `modx.grid.js`) to remove duplication of existing methods that could be shared and provide access to the new methods introduced from a new central base class; the original `Grid`, `LocalGrid`, and `JsonGrid` extend from this new base. As such, _every_ grid across the app should be interacted with to ensure all works as expected.

### Grids with New "Creator" Column and Protected Records
These are the grids where MODX provides built-in records to start with and/or where an Extra installs a record (which usually should only be removed by un-installing the Extra [such as in _Namespaces_]):
- _Namespaces_
- _Roles_
- _Contexts_
- _Media Sources_
- _Dashboards_
- _Content Types_
- _ACL Policies_
- _ACL Policy Templates_

A few notable details:
- Although _System Settings_ type grids would qualify for the same treatment as the above grids, implementation of it is really not feasible at this juncture due to the current data design. The identification of any given record's creator is largely derived or based on some hard-coded value (in the db or in php), as opposed to there being explicit flags in the db (columns where this type of info is persisted).
- For the most part, I've set most built-in items to be fully protected (i.e., in addition to being non-removable, certain fields are also set immutable [`name` and sometimes `description`, or in Content Types the `mime type`]). What will be carried through separately (in a follow-up PR) is to implement translation on these immutables fields.
- Following from the above point, matching field behavior has been implemented for a protected record's editing page (i.e., any field that was immutable on the grid is also immutable in the editing page [e.g., see Media Sources and Contexts]). Additionally, a new note is shown at the top of the panel alerting users re the item being protected and that some built-in fields may not be changed.

### All Editor Grids
- Buttons and button menu items are either hidden or set to disabled when permissions for their action is not granted.
- Bulk Actions button menu items should, additionally, only be active when selections have been made (this is regardless of permissions on their respective actions).
- Some grids depend on the permissions of related objects when rendering certain items. A prime example is the _Template Variables_ grid in a Template's editing panel: it renders a link on each TV name leading to that TV's editing page (but now, only if permission to edit & save TVs is granted). The various _Access Permissions_ grids also cross link in this way. Pay attention to these direct-linked items as you review each area.
- Pencil icon should only show for editable fields (based on permissions)
- Gear menu should only show when there is at least one action on a row that the user has permission to take; otherwise it is hidden

### General Strategy
As I was working through this PR, I kept one browser open with an Admin user with full permissions and another with an alternate user with permissions I'd vary to check my work. I basically systematically went through each nav menu item and completed updates to all relevant grids for one area at a time. I'd suggest the same for reviewing. Remember there are some grids that are a couple hops away, and can be easy to miss:
- TVs: There are multiple grids within this editing page to review (primarily the links to related objects [Contexts, Templates, User Groups] is what you'll want to pay attention to)
- Trash Manager
- User Group > Access Permissions (5 grids with various cross-linked objects)

opengeek:
@smg6511 — I am trying to get a collaborative review session together for this PR so we can get it integrated sooner than later. If possible, could you resolve the current small conflict in modx.grid.js?

smg6511:
@opengeek - I fixed the conflict but, dagonit, unintentionally ended up merging instead of force pushing because I was trying to fix a minor format error at the same time as resolving the conflict ... grrr. Hopefully this'll be ok.

opengeek:
> @opengeek - I fixed the conflict but, dagonit, unintentionally ended up merging not force pushing because I was trying to fix a minor format error at the same time as resolving the conflict ... grrr. Hopefully this'll be ok.

Unfortunately, that merge commit is problematic as it now shows completely unrelated changes in the diff.

smg6511:
How to back out of that is the question then :-/  Maybe I can revert the merge?

opengeek:
I think you should be able to go back to [8c6a110](https://github.com/modxcms/revolution/pull/16653/commits/8c6a110b15f3b0d23aa7bd6063c9c184bfc3bbf1) and then rebase the the work on the latest 3.x. I'm not sure exactly the situation, but maybe you could `git reset --hard HEAD~2` as it is two commits back?

smg6511:
@opengeek Ok, I think I got it squared away (I hope). Don't know why the compile assets step is choking.

modxcommunity:

https://community.modx.com/t/help-wanted-testing-a-pr-for-upcoming-revo-release/8528/1

core/src/Revolution/modNamespace.php
```diff
@@ -157,4 +180,44 @@ public function findPolicy($context = '')
...
+
+    public static function getExtrasNamespaces(modX $modx): array
+    {
+        $namespaceList = [];
+
+        $c = $modx->newQuery(modTransportPackage::class);
+        $c->select([
+            'name' => 'DISTINCT SUBSTRING_INDEX(`signature`,"-",1)'
+        ]);
+        $namespaces = $modx->getIterator(modTransportPackage::class, $c);
```
Mark-H: Shouldn't this be fetching from modNamespace instead of modTransportPackage? 🤔

smg6511: I don't think so, as the purpose of this method is to just get the Extras-related namespaces (and not others, like custom ones).

opengeek: I think @Mark-H is correct. MODX namespaces are namespaces, regardless if they are part of an extra or the core. There's no such thing as an Extras-related namespace based on the signature of a transport package.

raffy99:
i did some testing as proposed in #15919. i tested with individual permissions the the Roles grid (ACLs/Roles )
the Contexts grid and the Media Sources grid and also some cross checkings from TVs and corresponding grids. Permissions have been correctly taken into account for what I have seen.

core/src/Revolution/modNamespace.php
smg6511: The thing is I'm guessing you're both just looking at the code and not the context of what it's being used for. No, there's not literally an Extra-specific type of namespace, but I have derived that distinction (based on installed TPs) for purposes of identifying and sorting in specific grids. For example, I often create namespaces that are a home for particular types of customization and transformations. The method we're discussing is how I'm able to identify the "Creator" as "User" instead of "Core" or "Extra" (shown in the few grids that have that column). 

Am I wrong in thinking that every installed Extra does and must have its own unique namespace? I can't think of any I've installed that don't.

opengeek: You are wrong, yes. There is no requirement for an extra to have a unique namespace, nor is there any requirement preventing them from having half-a-dozen namespaces installed from a single transport package. And there is certainly no requirement that the namespace match the transport package (or extra) name.

I honestly have no idea what this Creator as User thing is about or how that is related to MODX namespaces.

jaygilmore: @smg6511 is this aspect of the PR necessary or do you tihnk it could be pulled for a future PR/improvement? It seems like it could be deferred. We really want to get this merged when we can, but if feels like this piece is optional and needs some review.

jaygilmore:
> i did some testing as proposed in #15919. i tested with individual permissions the the Roles grid (ACLs/Roles ) the Contexts grid and the Media Sources grid and also some cross checkings from TVs and corresponding grids. Permissions have been correctly taken into account for what I have seen.

Thanks for taking a look at this @raffy99. It's a fairly comprehensive PR and it's really challenging to test to avoid major issues across installs.

core/src/Revolution/modNamespace.php
smg6511: Been real busy lately, but I'll have some time to take a look at this over the next week.

smg6511: Alright guys, I pulled this logic for now and will add it to a follow-up PR. (It affected only two grids.) At that point, we can debate the "how" of what I'm wanting to achieve, and I'll demonstrate its value via a screen capture video.

matdave:
Hey @smg6511,

I just noticed a new force push. Is this ready or are you still working on it? Do you have a rundown of things to test? It's 100 files, so I'm trying to figure out where to take the first bite.

smg6511:
@matdave - Hi Mat - thanks for digging into this. To answer your questions:

- Re the force push, as this has been a long-running effort (which has been done for a while now), I've rebased every so often to ensure there are no conflicts. Short answer is, yes, it's ready ;-)
- [See this comment](https://github.com/modxcms/revolution/pull/16653#issuecomment-2540085244) earlier in the conversation thread here where I laid out a basic plan for how to review. Let me know if that's not sufficient.  I also left my [original draft ](https://github.com/modxcms/revolution/pull/15919) of this PR open for reference; you might find my opening comments there useful as well.

matdave:
So far it's looking good, I did notice one inconsistency while all of these are being adjusted. The button to create user groups is still present even if `usergroup_new` permissions aren't allowed.


image

matdave:
BTW, I'm fine if you want to consider that out of scope since it's not a regression. Just now that everything else is working, I'm noticing the things that shouldn't be there.

smg6511:
@matdave - Yeah, I didn't mess with that since it's tree-related and I was sticking specifically to the grids. There are some issues with the similarly-constructed Resource Groups trees and their toolbars/buttons I'd look to address separately.

matdave (approved):
Good work!

smg6511:
@matdave - Thanks, man! Appreciate you taking the time to review. Cheers :-)

PR #16656, opened 2024-12-04
Fix typo

Jako:
### What does it do?
Fix a lexicon typo


smg6511 (approved):
My bad ... good catch Thomas!

opengeek (approved):

Jako:
I only realised that because the translation was somehow strange.

Issue #16644, opened 2024-11-19
Add resource type icons (link, weblink) to quick search


Ruslan-Aleev:
### Summary
The current search shows **template icon**, BUT I think it would be useful to show the resource type icon, especially for **link, weblink**, etc.

![search_icon](https://github.com/user-attachments/assets/c05be64b-225b-49aa-a6be-ce278f286869)

### Why is it needed?
UX: It is easier to understand which resource is the main one and which is a link.

PR #16659, opened 2024-12-06
Fix classes for modWebLink/modSymLink in quick search

Ruslan-Aleev:
### What does it do?
Fix classes for modWebLink/modSymLink in quick search.
Since links are not exactly resources, we shouldn't apply a template icon to them in search; it's difficult to find the resource you need, especially if the names of the link and resource are the same.
p.s. In the tree, links are already displayed with default icons, not template icons.

Before:
before

After:
after

### Why is it needed?
UX: It is easier to understand now which resource is a link.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/16644

### DIFF
--- a/manager/assets/modext/widgets/core/modx.searchbar.js
+++ b/manager/assets/modext/widgets/core/modx.searchbar.js
@@ -51,6 +51,18 @@ MODx.SearchBar = function(config) {
                  * @returns {string}
                  */
                 getClass: function(values) {
+                    // Always return default classes for modWebLink/modSymLink
+                    if (values.class) {
+                        switch (values.class) {
+                            case 'MODX\\Revolution\\modSymLink':
+                                return 'files-o';
+                            case 'MODX\\Revolution\\modWebLink':
+                                return 'link';
+                            default:
+                                break;
+                        }
+                    }
+
                     if (values.icon) {
                         return values.icon;
                     }
@@ -59,10 +71,6 @@ MODx.SearchBar = function(config) {
                         switch (values.class) {
                             case 'MODX\\Revolution\\modDocument':
                                 return 'file';
-                            case 'MODX\\Revolution\\modSymLink':
-                                return 'files-o';
-                            case 'MODX\\Revolution\\modWebLink':
-                                return 'link';
                             case 'MODX\\Revolution\\modStaticResource':
                                 return 'file-text-o';
                             default:


smg6511:
@Ruslan-Aleev - I was taking a quick look at this and I think the issue might be due to something in your installation. I checked this on my local dev install (see screenshot below) and a live site; both show the icons you would expect.

Screen Shot 2024-12-11 at 12 10 37 AM

smg6511:
Let me ask folks this: What's an example use-case for applying a Template to a symlink or weblink? My feeling is that it's actually an undesirable inconsistency to apply the custom icon for some resource types and not others, as I would assume the reason to even specify a custom icon is to easily recognize _all_ resources using a particular Template. However, if others agree that applying the icon conditionally is the right thing to do, we should adjust the description for the Manager Icon Class field (which currently states that the icon will apply to “all resources using this template”).

Ruslan-Aleev:
For a SymLink , the template icon can be left (although I wouldn't leave it), since the template is used for the SymLink resource, as far as I understand. But for a WebLink, there is no point in setting a template icon at all, it will only confuse (both in the tree and in quick search).

Mark-H:
Template icons are **meant to override the default icon when set**. I'm not sure if I understand why you want to undo that override when you could just... not set the override in your site's template(s)?

As long as I don't add a manager icon to the templates, my search is showing as expected:

Scherm­afbeelding 2024-12-11 om 18 47 42

This seems like it would introduce an unnecessary inconsistency between how resources show in the tree and the search.

Ruslan-Aleev:
> This seems like it would introduce an unnecessary inconsistency between how resources show in the tree and the search.

In the tree, links are displayed **with their default icons (always), not the template icon**.
And does it make sense to apply a template icon to a link? A link is not a resource, it does not have a template, at least WebLink-resource. And if the link is to another site? Then the template icon has no meaning at all, it only confuses.
What is the use of a template icon for quick search? So that the user goes to the resource, sees that it is a link, and opens the desired resource? Extra steps.

smg6511:
> In the tree, links are displayed **with their default icons**  (always)

I don't think that is the intended behavior and that's the issue that should be solved. I think the point that Mark is making is that you have control over the appearance of custom icons by either _not_ applying a Template at all to your *links, or not specifying a custom icon for the Template(s) that you _do_ apply to *links. That said, why would you assign a Template to either one of the link Resource types anyway (the question I was getting at above re use cases)? Perhaps we're misunderstanding something?

Mark-H:
> In the tree, links are displayed with their default icons (always), not the template icon.

Ah, I overlooked that. Thanks for clarifying. 

What would need to be fixed IMO then is the tree, not the search. 

As an example of why you might still have a template assigned to weblinks, on MODX.today we use that to have different TVs that affect how the weblink is rendered in the collection of articles. The link itself may not be rendered as a standalone resource, but it may still appear on the site in different ways.

Ruslan-Aleev:
> why would you assign a Template to either one of the link Resource types anyway

I understood what you mean, I didn't think about it earlier, and we didn't understand each other.
MODX has automatic template assignment, so when you create a resource, it will have a template. Of course, you can remove the template when you change the resource type to a link, but this is an unnecessary action in my opinion.

> What would need to be fixed IMO then is the tree, not the search.

I don't agree with this, in my opinion it will reduce UX.

matdave:
I agree with @Mark-H on this. If an override template icon is set, it should be fixed in the tree to show that icon. Using custom link templates to add specific TVs to links is a thing I've done at least a few times. I don't personally mess with custom icons much though.

jaygilmore:
@Mark-H @Ruslan-Aleev where did we land on this? Is there any additional thoughts on this PR?

Ruslan-Aleev:
@jaygilmore I still stand by my opinion =) If the majority disagrees, then we can close the PR.

Ibochkarev:
## Code Review

**Summary:** The change makes quick search always use default icons (files-o, link) for modSymLink/modWebLink, bypassing the template icon override. This aligns search with the tree, where links already display default icons.

**Code quality:** The implementation is clean — early return for link classes before the template icon check. No issues with the logic itself.

**Design discussion (from the thread):**

Two approaches have been raised:

1. **This PR:** Keep links with default icons in both tree and search. Rationale: links are not full resources; template icons add little value and can confuse when link and target resource share a name.

2. **Alternative (@Mark-H, @matdave):** Fix the tree to show the template icon when set, so tree and search both respect the override. Rationale: custom link templates (e.g. for TVs) are a valid use case; if a template icon is set, it should appear everywhere for consistency.

**Assessment:** The PR achieves consistency between tree and search (both show default icons for links). The alternative would require tree changes and would treat links like other resources for icon display. The choice depends on product direction: either links are always visually distinct (current PR), or they follow template icon overrides like documents (alternative).

**Verdict:** Approve from a code perspective. The design decision (this PR vs. fixing the tree) should be confirmed by maintainers before merge, given the differing opinions in the thread.

smg6511:
Closing in favor of alternate #16958, which has been approved and is pending merge.

PR #16675, opened 2025-01-07
Fix visibility of the Install button in packages grid

### LINKED ISSUE #16672: [3.1.0] No install button after downloading new package
### Summary
I'm not seeing the install button appear in the Package Management grid, after downloading a new package in MODX 3.1.0.

### Step to reproduce

- Install MODX 3.1.0
- Install any package from package manager

### Observed behavior

image

### Expected behavior

Green install button should be present for packages that have not yet been installed.

### Environment

MODX 3.1.0
Nginx
PHP 8.2
MariaDB 11.4

Firefox / Chromium both have this issue. No errors in console.

All is well in 3.0.6.

smg6511:
### What does it do?
Adds a new prop to provide a boolean indicator of a package's installed state. The prop utilizes the recently-introduced date formatter's `isEmpty` method, which more comprehensively determines whether the `installed` field contains a meaningful value.

Note that this new prop still relies on a derived value. It would be much better IMO at some future point to persist this state in a new database field rather than basing it off of a date value.

### Why is it needed?
Install button is currently missing in the grid.

### How to test
Download, install, and uninstall a few packages to verify the install button appears as expected.

### Related issue(s)/PR(s)
Resolves #16672

### DIFF
--- a/core/src/Revolution/Processors/Workspace/Packages/GetList.php
+++ b/core/src/Revolution/Processors/Workspace/Packages/GetList.php
@@ -108,14 +108,14 @@ class GetList extends GetListProcessor
      */
     public function prepareRow(xPDOObject $object)
     {
-        if ($object->get('installed') === '0000-00-00 00:00:00') {
-            $object->set('installed', null);
-        }
         $packageArray = $object->toArray();
+
+        $isInstalled = !$this->formatter->isEmpty($packageArray['installed']);
+        $packageArray['isInstalled'] = $isInstalled;
         $packageArray = $this->getVersionInfo($packageArray);
         $packageArray = $this->formatDates($packageArray);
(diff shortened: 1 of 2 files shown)


BobRay:
I tested installing a minimal package that has separate "Install" and "Update" sections in resolvers and validators.  I can confirm that in 3.1.0, using "Reinstall" works as it should with respect to resolvers and validators. For new packages, the "Install" section executes, for new versions that update the package, the "Update" section executes. 

So, using Reinstall will be a reliable workaround until this fix is released.

modxcommunity:

https://community.modx.com/t/extjs-problems-on-newly-installed-site/8247/2

opengeek (approved):

modxcommunity:

https://community.modx.com/t/strange-issue-installing-new-extras-in-modx-3-1-0/8274/2

PR #16680, opened 2025-01-20
Fix manager menu for sql_mode "ANSI_QUOTES"

halftrainedharry:
### What does it do?
Fixes the SQL query that is used for the manager submenus.

### Why is it needed?
With the sql_mode `ANSI_QUOTES`, the generated SQL query returns the error `Unknown column 'installer' in 'order clause'`.

### How to test
* Add `ANSI_QUOTES` to the `sql_mode` setting in the MySQL config file `my.ini`
* Clear the cache
* Make sure the manager menu displays all submenus

### Related issue(s)/PR(s)
Related forum topic: https://community.modx.com/t/3-1-0-missing-left-side-menu/8236

### DIFF
--- a/core/src/Revolution/modMenu.php
+++ b/core/src/Revolution/modMenu.php
@@ -135,7 +135,7 @@ class modMenu extends modAccessibleObject
 
         if ($this->xpdo->getOption('package_installer_at_top', null, true)) {
             // make sure installer is always on top
-            $c->sortby('FIELD(modMenu.text, "installer")', 'DESC');
+            $c->sortby('FIELD(modMenu.text, \'installer\')', 'DESC');
         }
 
         $c->sortby($this->xpdo->getSelectColumns(modMenu::class, 'modMenu', '', ['menuindex']), 'ASC');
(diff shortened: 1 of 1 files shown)


core/src/Revolution/modMenu.php
```diff
@@ -135,7 +135,7 @@ public function getSubMenus($start = '')
 
         if ($this->xpdo->getOption('package_installer_at_top', null, true)) {
             // make sure installer is always on top
```
smg6511:
```suggestion
            // To support ANSI_QUOTES sql mode, string literals must be single quoted
```

smg6511 (approved):
This fix works as advertised. 

I'd also replace the comment/note as seen in the suggestion so someone doesn't come through and change the format down the road.

@opengeek - For cases like this, would it be worth looking into having xpdo transform  string literals to being single-quoted as a part of the query building process?

Ibochkarev (approved):

Mark-H (approved):

opengeek:
> @opengeek - For cases like this, would it be worth looking into having xpdo transform string literals to being single-quoted as a part of the query building process?

This is the result of someone manually building part of the query. I don't see how xPDO would have any ability to handle that intelligently. My rule of thumb when manually building queries is to use double-quotes to enclose the SQL so I can just use single-quotes as required inside the string. Or you can use the `\PDO->quote()` function itself, which is accessible from xPDO.

PR #16682, opened 2025-01-21
Fix non-existent snippet tags broken up by @

smg6511:
### What does it do?
Identifies and fixes a problem present in the getPropertySet method where the Element name is re-set, but without the necessary colon between the actual Element name and the rest of its tag.

### Why It's Needed / How to Test
See PR #16322

### Note
This is by and large the result of @Mark-H's work -- although it did take the better part of a night to track down the issue he was asking for help on ;-)

### DIFF
--- a/core/src/Revolution/modTag.php
+++ b/core/src/Revolution/modTag.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -9,7 +10,6 @@
  */
 
 namespace MODX\Revolution;
-
 
 use MODX\Revolution\Filters\modInputFilter;
 use MODX\Revolution\Filters\modOutputFilter;
@@ -119,7 +119,7 @@ abstract class modTag
      *
      * @param modX $modx A reference to the modX object
      */
-    function __construct(modX &$modx)
+    public function __construct(modX &$modx)
     {
         $this->modx =& $modx;
         $this->name =& $this->_fields['name'];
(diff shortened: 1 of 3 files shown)


Mark-H:
Thanks for taking over the PR and working on it some more. It is indeed something that takes a while to figure out - it took me like 3 days of work to get to that PR which fixed some issues, but left the one I couldn't sort out. 

I've added a note to my to do list to review and dig back into this to give it a proper review. It may take me a few weeks but is on the list.

jaygilmore:
> It may take me a few weeks but is on the list.

Hey @Mark-H any chance you might be able to take another peek at this one, since you were the OP? :)

smg6511:
@Mark-H - It's been a _few_ few weeks (just giving you a hard time). It'd be great to get all that work that mainly you did implemented   ;-) ... can you carve out some time to check this soon?

smg6511:
@Mark-H - Hey, I see you've been a little more active lately, so giving another bump on this one ;-)

smg6511:
@opengeek - Hey, I've bumped Mark a few times on this, but it seems he's got too much going on. I think this would be in your wheelhouse to review though ;-)

core/src/Revolution/modTag.php
```diff
@@ -511,19 +511,23 @@ public function getPropertySet($setName = null)
...
-                    $name .= ':' . $filters[1];
-                }
-                $this->set('name', $name);
-            }
+
+        $startFiltersIndex = strpos($name, ':');
+
+        if ($startFiltersIndex !== false) {
+            $tagStart = mb_substr($name, 0, $startFiltersIndex);
+            $tagEnd = mb_substr($name, $startFiltersIndex);
```
opengeek: My only concern with this solution is documented performance issues with using mb_substr(). Is there any other way we could solve this without degrading performance?

opengeek: Further, it the parsing methods use `substr()` in many places and never use the multibyte variation, so it seems inconsistent to have this use `mb_substr()`.

smg6511: Yeah, went ahead a swapped in the standard method; easy to do. It's been so long since submitting this that I don't remember what my motivation was for using the mb_ method ;-)

opengeek (approved):

PR #16697, opened 2025-01-28
Deprecate modUserProfile->sessionid and do not set value

opengeek:
### What does it do?
Marks the modUserProfile->sessionid field as deprecated and does not set the value when a user session is initiated or session id changes.

### Why is it needed?
Security enhancement.

### How to test
Make sure data is not added to the field for new users or updated for existing users when sessions are utilized.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -400,7 +400,6 @@ class modUser extends modPrincipal
                     $ua->set('logincount', $ua->logincount + 1);
                     $ua->set('lastlogin', $ua->thislogin);
                     $ua->set('thislogin', time());
-                    /*$ua->set('sessionid', session_id());*/
                     $ua->save();
                 }
             }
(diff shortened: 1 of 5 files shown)


Mark-H (approved):

core/src/Revolution/modUser.php
```diff
@@ -400,7 +400,7 @@ public function addSessionContext($context)
                     $ua->set('logincount', $ua->logincount + 1);
                     $ua->set('lastlogin', $ua->thislogin);
                     $ua->set('thislogin', time());
-                    $ua->set('sessionid', session_id());
+                    /*$ua->set('sessionid', session_id());*/
```
smg6511: Consider either just removing this line instead of commenting it out, or add an extra comment noting the intent.

JoshuaLuckers (approved):

PR #16702, opened 2025-02-04
Update superboxselect.js to version 1.6 and add HTML entity encoding

matdave:
### What does it do?
Updates superbox select to the latest version from Dan Humphrey (which is still old, just newer). I've also added an `htmlEncode` function to the value and display fields, so they do not break the manager if you use special characters. 

### Why is it needed?
Using special characters in the value or display field of the superbox select can lead to broken functionality. 

### How to test
The easiest way to test is to create a multi-list TV and add special characters to the values. For my testing I was having the most issues with `"`, `&`, and `<` characters in various combinations. If you test the output and save before and after this fix, you will find that the output and the stored results are not what you put in. After this fix, the output and stored results will match your options. 

### Related issue(s)/PR(s)
Found while working on a client site.


theboxer (approved):
🚀

modxcommunity:

https://community.modx.com/t/listbox-multiple-issue-re-preserveselectionorder/8622/6

Issue #16711, opened 2025-02-24
3.1.1 Menu z-index issue with tabs and toggle switch


Jako:
### Summary
In 3.1.1 the second level of the menu does not cover the MODX tabs and the toggle switch input.

### Step to reproduce
Create/use i.e. an extras menu with many entries and open that menu with mouseover.

### Observed behavior
See the following screenshot:

![Image](https://github.com/user-attachments/assets/0ea56f74-8ce6-4cad-882a-37d9f8c20868)

The tabs and the toggle switch are visible in front of the menu.

### Environment
MODX 3.1.1.

rthrash:
I suspect this can be fixed with some minor CSS updates:

```
#modx-leftbar { z-index: 3 }
#modx-action-buttons-container: { z-index: 2 }
#modx-content { z-index: 1 }
```

It needs a lot of testing! I'm going to do just that via the osTreeStyles plugin Jay created: https://community.modx.com/t/coloured-icons-in-the-resource-browser/8324/6?u=rthrash

<img width="665" alt="Image" src="https://github.com/user-attachments/assets/e208f92f-0bf5-4ab6-aad0-9aa17e552482" />

*almost there…*

That basically works, but it returns another bug that was previously addressed: when going full-screen with some plugins like Ace or Monaco (using position:fixed), the left bar and action buttons overlap the full screen editor again. I suspect the way fullscreen plugins may need some tweaking as there's a conflict with our left bar/navbar/content/buttons IDs and the fullscreen overlays. 

<img width="980" alt="Image" src="https://github.com/user-attachments/assets/d46adc39-334e-41ca-9c15-dcc7a7f017af" />

*…but a full-screen overlay breaks now*

<img width="981" alt="Image" src="https://github.com/user-attachments/assets/0f87cb69-9b4d-4282-9b35-5bd61fc158b1" />

*how it should look (remove z-index from #modx-content)*

 Maybe @matdave can offer a suggestion?

rthrash:
Related: #16690

opengeek:
This is related to another reported problem that was created by #16690 — see https://github.com/modxcms/revolution/pull/16690#issuecomment-2670602569

PR #16690, opened 2025-01-23
Fix navbar z-index

jenswittmann:
### What does it do?

The z-index is too high and prevents extras like ace/tinymce/commerce modal and so on being on top.

### Why is it needed?

See #16684.

### How to test

Replace `manager/templates/default/css/index.css` with this [demo file](https://gist.github.com/jenswittmann/9152010fe306121a2a810d2128211797).

PR #16712, opened 2025-02-24
Add z-index to x-form-item class

matdave:
### What does it do?
Primary goal is to add z-index to x-form-item class. During setup of my test environment I noticed some critical vulnerabilities were unaddressed in the package generation files, so I corrected those as well. 

### Why is it needed?
Fixes an issue where the form items can overlap the menu bar, making the menu unclickable. 

### How to test
Open a file to edit in the manager.  Scroll down until "Content" -> "Clear Cache" -> "Refresh URIs" opens over one of the field labels. Without the fix the text will appear over "Refresh URIs" making it hard to click. The same happens if you do this for checkboxes and certain other field elements.

### Related issue(s)/PR(s)
#16711


matdave:
This issue was introduced in #16690

rthrash:
Can confirm this fixes the issue.

rthrash (approved):
Z-index fix issues are solid. Thanks @matdave

opengeek:
> During setup of my test environment I noticed some critical vulnerabilities were unaddressed in the package generation files, so I corrected those as well.

@matdave — this is complicating the integration of the PR. Can you remove the package.json (and resulting package-lock.json) changes from this PR and submit those as a separate PR. Also, what version of npm are you using because I get this when running `npm install`… and that again modifies the lock file.

```
❯ npm install
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
```

In addition, after running `npm install` and running `grunt build`, the modx.jsgrps-min.js file is also changed. So I think this would be better applied as a separate PR, either before or after applying this one.

matdave:
@opengeek I've got those reverted

PR #16713, opened 2025-02-25
Fix empty attributes with i.e. german umlaut in the transport package readme

Jako:
### What does it do?
Change htmlentities to htmlspecialchars

### Why is it needed?
In one of my packages I have used a `ß` in the readme for the transport package. When this transport package is installed, the result the of `Workspace/Packages/GetAttribute` processor is empty for some reason inside of the Parsedown class in combination with the processor response and the package can't be installed anymore.

### How to test
Add a `ß` in the readme of a transport package.

### Related issue(s)/PR(s)
None known

### DIFF
--- a/core/src/Revolution/Processors/Workspace/Packages/GetAttribute.php
+++ b/core/src/Revolution/Processors/Workspace/Packages/GetAttribute.php
@@ -78,7 +78,7 @@ class GetAttribute extends Processor
         foreach ($attributesToGet as $attribute) {
             $data = $this->transport->getAttribute($attribute);
             $attributes[$attribute] = in_array($attribute,
-                ['changelog', 'license', 'readme']) ? $parseDown->text(htmlentities($data)) : $data;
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

smg6511:
You mention the _umlaut_ but show an Eszett (ß). Was that a typo by chance?

Jako:
`ß` is an umlaut in German. Like `ä`, `ö`, `ü` and so on.

smg6511:
This probably doesn't matter—as I think you're indicating that all those special characters need the same handling—but ß is a primary German character, whereas the umlaut is a vowel accent character. I'm not German, but studied the language for a little while long ago ;-)

Jako:
The word `german umlaut` is now away.

Jako:
@smg6511 Can you remove your comments about the wrong `german umlaut` usage. They don't make any sense after my PR description change.

opengeek (approved):

PR #16718, opened 2025-03-11
Fix wrong check of checkRequirement method

Jako:
### What does it do?
Fix wrong check of checkRequirement method

### Why is it needed?
checkRequirement result contains a string or true. So the check will never have a false result and the error is not sent.

### Related issue(s)/PR(s)
None known.

### DIFF
--- a/core/src/Revolution/Rest/modRestController.php
+++ b/core/src/Revolution/Rest/modRestController.php
@@ -668,8 +668,8 @@ abstract class modRestController
         $properties = $this->getProperties();
 
         if (!empty($this->postRequiredFields)) {
-            if (!$this->checkRequiredFields($this->postRequiredFields)) {
-                return $this->failure($this->modx->lexicon('error'));
+            if ($result = $this->checkRequiredFields($this->postRequiredFields) !== true) {
+                return $this->failure($result);
             }
         }
 
@@ -744,8 +744,8 @@ abstract class modRestController
         }
 
         if (!empty($this->putRequiredFields)) {
-            if (!$this->checkRequiredFields($this->putRequiredFields)) {
-                return $this->failure();
+            if ($result = $this->checkRequiredFields($this->putRequiredFields) !== true) {
+                return $this->failure($result);
             }
         }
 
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Rest/modRestController.php
```diff
@@ -668,8 +668,8 @@ public function post()
         $properties = $this->getProperties();
 
         if (!empty($this->postRequiredFields)) {
-            if (!$this->checkRequiredFields($this->postRequiredFields)) {
-                return $this->failure($this->modx->lexicon('error'));
+            if ($result = $this->checkRequiredFields($this->postRequiredFields) !== true) {
+                return $this->failure($result);
             }
```
smg6511: In each of the three request types, I believe your change still will not capture the validation string in case of an error. This definitely will:
```suggestion
            $result = $this->checkRequiredFields($this->postRequiredFields);
            if ($result !== true) {
                return $this->failure($result);
            }
```

Jako: Both code variants are valid. An assignment in the if clause is maybe not the preferred code style. This can be changed.

smg6511: The two actually return different results. In the original, `$result` will always be a boolean value while the suggested will preserve the string value if present.

Jako: The comparison runs after the assignment according to my tests (the error message contained a string with the missing fields). See the note on: https://www.php.net/manual/en/language.operators.precedence.php

Perhaps a `assignment in condition` code quality check can be enabled. But this will disallow a lot of these code shortcuts in the core code.

smg6511: You need to test it in isolation to see what I mean. See this [sandbox test](https://onlinephp.io/c/e8595). As you can see in the second case, you can get the result you're expecting if you set off the first part of the statement in parens. Regardless, it's probably better code style wise to shy away from the one-liner.

Jako: You are right. If I add the brackets around the assignment, another error message is added. I have to check my other code, because I was sure that this comparison is valid.

Jako: During these checks I noticed some `if(null === $variable = $class->method())` checks in foreign code. Maybe it is valid the other way around.

smg6511: Yes, it does work the other way around, but I'd still generally advise against it. BTW, you'll need to make the same update for the put and delete checks (roughly at L748 and L823).

opengeek (approved):

PR #16721, opened 2025-03-18
Convert invalid dates to NULL in modTransportPackage.installed

opengeek:
### What does it do?
Converts any invalid dates in the modTransportPackage.installed column to NULL.

### Why is it needed?
Empty dates should have been saved as NULL in the column and their presence can cause errors when NO_ZERO_DATE/NO_ZERO_IN_DATE sql_modes are enabled in an environment.

### How to test
In an environment with NO_ZERO_DATE/NO_ZERO_DATE sql_modes enabled, build the transport and run the upgrade process, confirming that the process completes and that any invalid dates in the modTransportPackage.installed column are now NULL. You will need to have invalid dates in that column before enabling the strict sql_modes if you want to verify that this works properly.

### Related issue(s)/PR(s)
Required for #16717 to be merged.

### DIFF
--- a/setup/includes/upgrades/mysql/3.1.2-pl.php
+++ b/setup/includes/upgrades/mysql/3.1.2-pl.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * Specific upgrades for Revolution 3.1.2-pl
+ *
+ * @var modX $modx
+ * @var modInstallVersion $this
+ * @package setup
+ * @subpackage upgrades
+ */
+
+/* Repair any datetime data that can trigger errors when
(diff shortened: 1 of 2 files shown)


Jako:
Should this change be ported to version 2.x?

opengeek:
> Should this change be ported to version 2.x?

Maybe? I wasn't planning another 2.x release right now. Considering it can't really be made compatible with the latest release of PHP, I'm not too concerned about making it compatible with super strict sql_modes enabled.

rthrash:
@Jako did you have a chance to see if this one works as advertised?

PR #16722, opened 2025-03-18
Fix wrong check of checkRequirement method

Jako:
What does it do?
Fix wrong check of checkRequirement method

Why is it needed?
checkRequirement result contains a string or true. So the check will never have a false result and the error is not sent.

Related issue(s)/PR(s)
2.x Port of #16718


opengeek (approved):

Mark-H (approved):

theboxer (approved):
nice catch!

PR #16734, opened 2025-04-17
Fix Media Browser Sorting by Modification Date

### LINKED ISSUE #16731: Incorrect sorting by "Last Modified" in MODX browser
### Summary
In the MODX browser the files don't get correctly sorted by the "last modified" date, if the system setting `manager_date_format` doesn't have the default value.

### Step to reproduce
* Change the system setting `manager_date_format` to `d.m.Y`
* Open the MODX browser (menu "Media" -> "Media Browser")
* Change the dropdown "Sort By:" to "Last Modfied"
* Verify that the order of the files is incorrect

### Environment
MODX 3.1.2-pl

### Cause
The processor `Browser/Directory/GetFiles` used to return the **timestamp** of a file (in the field "lastmod"). The formatting of the timestamps was done in JS. The sorting of the files was also done in JS using the timestamp.

With #16604 the processor now already returns the **formatted time string**. The sorting is done in JS using this already formatted strings.

smg6511:
### What does it do?
Adds raw, unformatted date data into store for use in sorting.

### Why is it needed?
Some custom formatted date strings can break proper sorting.

### How to test
Open Media Browser and select a folder with several files. When choosing to sort by Last Modified, files should show in descending order (you can only check this by selecting each file and taking note of the date in the right details panel).

### Related issue(s)/PR(s)
Resolves #16731


opengeek (changes requested):
Can we leave the reformatted code in modx.browser.js file out of this for now? Is there actually a change in that file related to this problem?

jaygilmore:
It does look as if the modx.browser.js stuff was formatting only and should be omitted, for sure unless there are specific changes. If there are, I'd think reverting the formatting for now would be useful.

jaygilmore:
@smg6511 any chance you might be able to resubmit withouth the formatting changes? We'd love to have the addition in the next release.

smg6511:
@jaygilmore @opengeek - Ok guys, I removed the cleanup of the js file.

jaygilmore:
@smg6511 thanks. Much appreciated. :)

opengeek (approved):

PR #16741, opened 2025-05-06
Moved logic to add/remove collapsed (left bar) css class in appropriate event handlers

rtripault:
### What does it do?

Moved the logic to add/remove the `collapsed` CSS class on the "trigger" to collapse/expand the manager left bar (trees)

### Why is it needed?

The logic was initially set in a "click" handler, making it only happening when clicking on the trigger (arrow).
However, ExtJS code (`Ext.getCmp('modx-layout').toggleLeftbar()`, `Ext.getCmp('modx-layout').hideLeftbar()`, `Ext.getCmp('modx-layout').showLeftbar()`) could also be executed to perform the same action without "clicking" on the trigger, resulting in visual contraction like having the tree collapsed, but the trigger offering to collapsed it anyway.

### How to test

Once the manager is loaded, in the browser console, execute `Ext.getCmp('modx-layout').toggleLeftbar()`

Without this patch, the "trigger" arrow should not change direction. With this patch applied, it should reflect the tree state (collapsed or expanded)

### DIFF
--- a/manager/assets/modext/core/modx.layout.js
+++ b/manager/assets/modext/core/modx.layout.js
@@ -427,18 +427,25 @@ Ext.extend(MODx.Layout, Ext.Viewport, {
                 },
                 afterrender: function() {
                     const trigger = Ext.get('modx-leftbar-trigger');
-                    if (this.collapsed) {
-                        trigger.addClass('collapsed');
-                    }
                     trigger.on('click', function() {
                         if (this.collapsed) {
-                            trigger.removeClass('collapsed');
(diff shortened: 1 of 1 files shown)


smg6511:
Hey @rtripault - First, good catch! You were definitely on the right track. As I was reviewing your PR, I found a couple issues but couldn't make my suggestions directly on your PR since they were out of the selectable range; so I went ahead and posted an alternate PR (#16748).

smg6511:
Alternate fix (#16748) merged, so closing.

PR #16748, opened 2025-05-24
Ensure left bar trigger and collapsed state stay in sync

smg6511:
### What does it do?
Note that this is an alternative to #16741 that addresses a couple other nuances. Like @rtripault did, I moved the add/remove class logic for the trigger elsewhere (but into the upper collapse/expand blocks for the west region). Alterations were also made to the `toggleLeftBar` method.

### Why is it needed?
As the original PR stated, the trigger state was not changing when `toggleLeftBar` was called; this method is exclusively used for the keyboard shortcut (which is ctrl-shift-h). A couple other things get resolved with this alternate solution:
1. The initial internal state is correct now. Before, the first time you trigger `toggleLeftBar` nothing happens (or the wrong thing happens); this issue was already present before (not caused by the referenced PR).
2. There were issues when using the shortcut (triggering `toggleLeftBar`), then actually clicking the trigger icon (or vice versa). This PR ensures the correct state and behavior no matter how you open/close the left bar.

### How to test
Open and close the left bar by using both the trigger (arrow icon) and using the keyboard shortcut (ctrl-shift-h), in various combinations. Additionally navigate around a bit to verify the correct state is carried through from page to page.

### Related issue(s)/PR(s)
None, as mentioned this is an alternate to #16741.


smg6511:
@matdave - Hey Mat, can you give this a quick review? I'd like to knock this off and get it into the next release. Thanks!

matdave (approved):
Seems to work!

Issue #15864, opened 2021-10-24
PHP 8.1: strftime() is deprecated


JoshuaLuckers:
The `strftime()` function is called by the core in [various places](https://github.com/modxcms/revolution/search?q=strftime). 

In PHP 8.1 this will trigger a deprecation warning because the strftime() and gmstrftime() functions have been deprecated in favor of date()/DateTime::format() (for locale-independent formatting) or IntlDateFormatter::format() (for locale-dependent formatting).

> The strftime() and gmstrftime() functions exhibit similar issues as strptime(), in that the formats they support, as well as their behavior, is platform-dependent. Unlike strptime(), these functions are available on Windows, though with a different feature set than on Linux. Musl-based distributions like Alpine do not support timezone-related format specifiers correctly. These functions are also locale-based, and as such may exhibit thread-safety issues.
> 
> Once again date() or DateTime::format() provide portable alternatives, and IntlDateFormatter::format() provides a more sophisticated, localization-aware alternative.

References:
- https://github.com/php/php-src/blob/1cf4fb739f7a4fa8404a4c0958f13d04eae519d4/UPGRADING#L379-L381
- https://wiki.php.net/rfc/deprecations_php_8_1#strftime_and_gmstrftime

### Environment
- MODX 2.x, 3.x 
- PHP 8.1

muzzwood:
I've been experimenting with this.
It looks like the PHP extension `intl` would be a requirement if we're going to use `IntlDateFormatter`.
Maybe we need a helper method/class that determines if it's installed and uses `DateTime::format` which isn't locale aware otherwise.

Usage of IntlDateFormatter would be something along the lines of:
```
$locale = $this->modx->getOption('locale') ?? '';
$timezone = $this->modx->getOption('timezone') ?? '';

$formatter = new IntlDateFormatter($locale, IntlDateFormatter::SHORT, IntlDateFormatter::SHORT, $timezone);
return $formatter->format(new DateTime());
```

rthrash:
This issue has been mentioned on **MODX Community**. There might be relevant details there:

https://community.modx.com/t/whats-happening-with-the-strftime-and-date-output-modifiers/4826/3

Mark-H:
Two weeks ago I released a SiteDash update that tracks available PHP extensions to see if requiring intl would be feasible. On a sample size so far of about 18% of sites connected to SiteDash, roughly 81% of those have the intl extension enabled. That's more than imagick (75%) but less than fileinfo (91%). 

[...]

PR #16749, opened 2025-06-01
Replace remaining usages of php strftime

smg6511:
### What does it do?
Removes usage of the deprecated `strftime` function, implementing the recently-added `modManagerDateFormatter` class formatting methods in its place.

### Why is it needed?
PHP plans to remove `strftime` in v9.0.

### How to test
There were five areas where `strftime` was replaced: 
1. Output filters (date, strftime, fuzzydate), 
2. Date TV (Input Options Disabled Dates field examples and Output Options Date Format field & examples), 
3. Resource Overview (data controller), 
4. Package Browser (search results _released_ date), and
5.  modFIle, which is limited in its implementation; it's not the primary file system handler (which is flysystem based). The one usage I played with was the Export functionality for Form Customization sets. Here I'm not sure that the date is ever used anywhere, so this seems to be more a matter of verifying the Export functionality itself is working as expected.

Note that a new static method was added to `modManagerDateFormatter` to convert existing strftime formats (e.g., `'%A, %d %B %Y'`) to their datetime equivalents (e.g., `'l, d F Y'`) for BC.

That said, just play around in these areas and ensure all works as expected.

### Related issue(s)/PR(s)
Resolves the issue raised in #15864. The subject of moving to the `Intl` extension and its `IntlDateFormatter` methods discussed in that same issue should probably be a separate feature proposal, which would ultimately be implemented within MODX's `modManagerDateFormatter`.

### DIFF
--- a/core/lexicon/en/tv_widget.inc.php
+++ b/core/lexicon/en/tv_widget.inc.php
@@ -48,7 +48,7 @@ $_lang['combo_typeahead_delay'] = 'Delay';
 $_lang['combo_typeahead_delay_desc'] = 'Milliseconds before a matched option is shown. (Default: 250)';
 $_lang['date'] = 'Date';
 $_lang['date_format'] = 'Date Format';
-$_lang['date_format_desc'] = 'Enter a format using <a href="https://www.php.net/strftime" target="_blank">php’s strftime syntax</a>.
+$_lang['date_format_desc'] = 'Enter a format using <a href="https://www.php.net/manual/en/datetime.format.php" target="_blank">php’s datetime syntax</a>.
     <div class="example-list">Common examples include:
         <ul>
             <li><span class="example-input">[[+example_1a]]</span> ([[+example_1b]]) (default format)</li>
--- a/core/src/Revolution/File/modFile.php
+++ b/core/src/Revolution/File/modFile.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -9,7 +10,6 @@
  */
 
 namespace MODX\Revolution\File;
-
 
 /**
  * File implementation of modFileSystemResource
@@ -151,7 +151,6 @@ class modFile extends modFileSystemResource
      */
     public function unpack($to = '', $options = [])
     {
-
         $results = false;
 
         if ($this->fileHandler->modx->getService('archive', 'compression.xPDOZip', XPDO_CORE_PATH, $this->path)) {
--- a/core/src/Revolution/File/modFile.php
+++ b/core/src/Revolution/File/modFile.php
@@ -188,9 +188,9 @@ class modFile extends modFileSystemResource
      *
      * @return string The formatted time
      */
-    public function getLastAccessed($timeFormat = '%b %d, %Y %I:%M:%S %p')
+    public function getLastAccessed($timeFormat = 'M d, Y h:i:s A')
     {
-        return strftime($timeFormat, fileatime($this->path));
+        return $this->fileHandler->formatter->format(fileatime($this->path), $timeFormat);
     }
 
     /**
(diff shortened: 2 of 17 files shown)


halftrainedharry:
> The subject of moving to the `Intl` extension and its `IntlDateFormatter` methods discussed in that same issue should probably be a separate feature proposal, which would ultimately be implemented within MODX's `modManagerDateFormatter`.

So what exactly is the plan here?
I'm a bit concerned about the PR is this current form. We have a lot of non-english sites and it's quite common to output the month (or day of the week) as a "full textual representation". E.g. like this:

```
[[*my_date_tv:strtotime:date=`%d. %B %Y`]]
```
With this PR, the **displayed output changes**, and the name of the month is always in english. Which shouldn't be the case.

smg6511:
@halftrainedharry - Good point, I was kind of "running with it" given that everything else had been changed away from `strftime`. The couple places I need to reconsider are the Output Filter and TV Output (Date) Option since their used to format public-facing content. Let me look at integrating `Intl` in now; does anyone object to doing that? If we're not ready to fully commit to `Intl` I could add some checks in and still use strftime if `Intl` is unavailable and php < 9. Thoughts?

smg6511:
Note that the latest commit reflects work-in-progress to address the need for localized date display. At this point I'm looking more for opinion on where I'm going with this, structure/stragegy-wise. 

I'm thinking of adding a system setting allowing for explicit adoption of `Intl` now, with options for: [1] backend (core manager areas), [2] frontend (tv output), or [3] both. Also I think a new `idate` output filter would be useful (but maybe more appropriate to put in a follow-up PR)—this would be a date filter that exclusively uses `IntlDateFormatter`.

The backend would be much easier for users to move to, as they'd only have to change one formatting setting (`manager_date_format`) to an `Intl` compatible one. The frontend change _could_ be more involved, where folks might have multiple places they've set a `strftime` (or maybe datetime, but I don't think so?) formatting string. That said, in this update I baked in an auto conversion for when `strftime` is no longer available (so far it's looking like php 9 is when it'll be removed); that might be leveraged to enable auto-converted frontend implementation, allowing users to incrementally change their `strftime` patterns to `Intl` ones. 

I don't believe Extras are a factor here (?), as ones that use a date in one way or another have their own formatting settings and implementations.

Ibochkarev:
## Code Review

**Summary:** The PR correctly addresses the deprecation of `strftime` (PHP 9) and adds Intl-based localization for frontend dates. The approach with `modFrontendDateFormatter`, `modDateFormatConverter`, and strftime→Intl/datetime mapping is sound.

**Strengths:**
- Replaces `strftime` with `modManagerDateFormatter` / `modFrontendDateFormatter` across output filters, TV configs, modFile, modTransportProvider, and resource data.
- Adds backward-compatible conversion of strftime format strings to Intl/datetime equivalents.
- Addresses @halftrainedharry's concern about localized month/day names for non-English sites.
- New `tabs2spaces` filter and lexicon loading for date-related filters are useful additions.

**Issues to fix before merge:**

1. **Debug code in `modFrontendDateFormatter`** — Constructor and `format()` log to `modX::LOG_LEVEL_ERROR` on every call, which will flood error logs. Remove the `$msg` logging blocks.

2. **`$_SESSION['manager_language']` in frontend context** — `modFrontendDateFormatter` is used for frontend output filters; `$_SESSION['manager_language']` may not exist there. Use `cultureKey` or a safe fallback instead.

3. **`modStrftimeToIntlConverter`** — Contains `echo 'CONV working...';` (debug output). Remove before merge. The class also appears to be a stub (`return false`); confirm it is either implemented or not yet wired in.

4. **Date filter input parsing** — The original logic used `strtotime($output)` for non-timestamp strings. `parseValue()` in the parent only accepts timestamps or `Y-m-d H:i:s` MySQL format. Date strings like `2024-01-15` or `January 15, 2024` may no longer parse correctly. Consider preserving `strtotime()` fallback for non-timestamp strings before passing to the formatter.

**Verdict:** The direction is good and the structure is solid. Remove debug code and fix the frontend/session and parsing edge cases before marking ready for review.

modxcommunity:

https://community.modx.com/t/strftime-errors/9044/6

PR #16754, opened 2025-06-17
Update chunk.inc.php

slaweally:
"Yığın" → "Parça" (much more understandable and functional) Since Chunks represent HTML fragments, the term "Parça" is much more appropriate


ant-tar:
@slaweally For all translations please use CrowdIn instead: https://crowdin.com/project/modx-revolution/tr
All these language requests will be declined soon.

slaweally:
> @slaweally For all translations please use CrowdIn instead: https://crowdin.com/project/modx-revolution/tr All these language requests will be declined soon.

Thank you, but it is easier and more convenient for my baby to translate manually. that's why I sent a request to fix the mistakes and omissions in the translations. if it is not accepted, I don't need to make new translations.

If you want to see the current version in Turkish: https://github.com/slaweally/ModX-revolution

You can get the full and complete version on my own page. Good work.

Ruslan-Aleev:
@slaweally Thank you for participating!
But all translations are taken from https://crowdin.com/project/modx-revolution/tr automatically via PR, that's how it's set up. So send a request to crowdin and add the necessary translation files, if it's not difficult for you, adding translations is easy, and you don't have to add/edit the entire translation.

slaweally:
> @slaweally Thank you for participating! But all translations are taken from https://crowdin.com/project/modx-revolution/tr automatically via PR, that's how it's set up. So send a request to crowdin and add the necessary translation files, if it's not difficult for you, adding translations is easy, and you don't have to add/edit the entire translation.

Okay, I understand you, but the process will work very differently for me. I will make an effort to develop the Turkish documentation and other content of ModX. so it's not just a language file for core. so it's better to move forward without being tied to a specific place.

I understand what you are saying, if we get it automatically from there, no problem.

Jako:
You can upload the translated turkish language files from your repository into Crowdin very easy. You just need a Crowdin login.

It is a lot easier to have one centralized language source, that is translated to other target languages. With such a system, the translators of each language can be informed, that the source language has changed and these strings maybe have to be changed in the target languages, too. And the translators can be informed that new language strings have been added. Also the automatic translation tools of Crowdin are quite good.

PR #16755, opened 2025-06-20
Dynamic manager path in setup

henninghh990:
Added MODX_MANAGER_URL to the setup options, so that "Files are already in-place" is checkable on sites where manager/ has been renamed.

### What does it do?
The function on the first page of the setup checks if the `manager/`-folder exists, and let's the user check "Files are already in-place". I've replaced the string `manager/` with `MODX_MANAGER_URL` so it reads the dynamic value from `config.inc.php`.

### Why is it needed?
Many of us rename the `manager/`-page to either customize the experience or for security purposes. This fix enables us to install / upgrade a site and select "Files are already in-place".

### How to test
Rename manager and manager URL in `config.inc.php`, then run setup.


opengeek (commented):
You can already achieve the ability to customize the manager path/url using the advanced distribution, along with several other paths/urls. I don't see the need for this in the standard distribution.

smg6511:
Agree with @opengeek here. I've also changed the manager folder name on some sites; in the case where you don't use the advanced dist for an initial site install you can simply manually change the manager path in `config.inc.php` when renaming the manager folder. That change won't be overwritten on future upgrade installs.

Though having a convenience method for doing this _might_ be nice, it probably encourages the idea that security-by-obscurity is effective—which I believe many if not most would agree it is not.

Ibochkarev (commented):
## What the PR does

**File:** `setup/controllers/options.php`

```diff
- if (file_exists(MODX_INSTALL_PATH . 'manager/index.php') &&
+ if (file_exists(MODX_INSTALL_PATH . MODX_MANAGER_URL . '/index.php') &&
```

When `config.inc.php` defines a custom `MODX_MANAGER_URL` (e.g. `admin` after renaming the folder), the setup correctly checks for the renamed path instead of always looking for `manager/`.

---

## Technical assessment

**1. Logic**

The change is correct: if the manager folder is renamed and `MODX_MANAGER_URL` is updated in `config.inc.php`, the setup will check the right path.

**2. Alternative: `MODX_MANAGER_PATH`**

If `MODX_MANAGER_PATH` is defined at setup time, it would be a clearer choice for a filesystem check:

```php
file_exists(MODX_MANAGER_PATH . 'index.php')
```

`MODX_MANAGER_PATH` is the filesystem path; `MODX_MANAGER_URL` is the URL segment. For a renamed folder they usually match, but `MODX_MANAGER_PATH` is more explicit for path operations.

**3. Path concatenation**

`MODX_INSTALL_PATH . MODX_MANAGER_URL . '/index.php'` is fine as long as `MODX_MANAGER_URL` has no leading slash. If it can (e.g. `/manager`), you might get `path//manager/index.php`. Worth confirming the expected format of `MODX_MANAGER_URL` in `config.inc.php`.

---

## Maintainer feedback (summary)

| Reviewer   | Position |
|-----------|----------|
| @opengeek | Custom manager path is supported via the advanced distribution; not needed in standard. |
| @smg6511  | Manual edit of `config.inc.php` after renaming is enough; that change is not overwritten on upgrades. Also notes that security-by-obscurity is generally discouraged. |

---

## Conclusion

The implementation is valid and would improve setup for sites with a renamed manager folder. The decision not to merge appears to be about scope (advanced dist vs standard) and workflow (manual config edit) rather than correctness.

If the PR is revisited, consider using `MODX_MANAGER_PATH` if it is available during setup, and ensure `MODX_MANAGER_URL` is documented as the URL segment without a leading slash.

smg6511:
@henninghh990 - We appreciate your submission. We've elected, however, to not incorporate it at this time since there's already a way to achieve this via the advanced distribution. Please consider contributing again in the future ;-)

PR #16757, opened 2025-07-11
Fix fatal error in trim output filter

jenswittmann:
This solves a PHP fatal error that occurs when using something like this:

`[[!#SESSION.user.password:trim]]`

### PHP Error

```
Fatal error: Uncaught TypeError: trim(): Argument #1 ($string) must be of type string, array given in core/src/Revolution/Filters/modOutputFilter.php:66 Stack trace: #0 core/src/Revolution/Filters/modOutputFilter.php(66): trim(Array) #1 core/src/Revolution/modTag.php(354): MODX\Revolution\Filters\modOutputFilter->filter(Object(ModxPro\PdoTools\Parsing\Tag)) #2 core/components/pdotools/src/Parsing/Tag.php(32): MODX\Revolution\modTag->filterOutput() #3 core/components/pdotools/src/Parsing/Parser.php(264): ModxPro\PdoTools\Parsing\Tag->process() #4 core/src/Revolution/modParser.php(221): ModxPro\PdoTools\Parsing\Parser->processTag(Array, true) #5 core/components/pdotools/src/Parsing/Parser.php(73): MODX\Revolution\modParser->processElementTags('', '<!DOCTYPE html>...', true, false, '[[', ']]', Array, 9) #6 core/src/Revolution/modResource.php(520): ModxPro\PdoTools\Parsing\Parser->processElementTags('', '<!DOCTYPE html>...', true, false, '[[', ']]', Array, 10) #7 core/src/Revolution/modResource.php(468): MODX\Revolution\modResource->parseContent() #8 core/src/Revolution/modResponse.php(72): MODX\Revolution\modResource->prepare() #9 core/src/Revolution/modRequest.php(154): MODX\Revolution\modResponse->outputContent(Array) #10 core/src/Revolution/modRequest.php(138): MODX\Revolution\modRequest->prepareResponse() #11 core/src/Revolution/modX.php(1509): MODX\Revolution\modRequest->handleRequest() #12 index.php(63): MODX\Revolution\modX->handleRequest() #13 {main} thrown in core/src/Revolution/Filters/modOutputFilter.php on line 66
```

### System

- MODX 3.1.2
- PHP Version 8.3.21
- Extras: pdoTools 3.0.2-pl

### DIFF
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -48,7 +48,7 @@ class modOutputFilter
         if (!is_scalar($element->_output)) {
             return;
         }
-        
+
         $usemb = function_exists('mb_strlen') && (boolean)$this->modx->getOption('use_multibyte', null, false);
         $encoding = $this->modx->getOption('modx_charset', null, 'UTF-8');
 
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -63,7 +63,7 @@ class modOutputFilter
 
                 $this->log('Processing Modifier: ' . $m_cmd . ' (parameters: ' . $m_val . ')');
 
-                $output = trim((string) $output);
+                $output = is_string($output) ? trim($output) : $output;
 
                 try {
                     switch ($m_cmd) {
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -63,7 +63,7 @@ class modOutputFilter
 
                 $this->log('Processing Modifier: ' . $m_cmd . ' (parameters: ' . $m_val . ')');
 
-                $output = trim($output);
+                $output = trim((string) $output);
 
                 try {
                     switch ($m_cmd) {
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -46,10 +46,6 @@ class modOutputFilter
      */
     public function filter(&$element)
     {
-        if (!is_scalar($element->_output)) {
-            return;
-        }
-
         $usemb = function_exists('mb_strlen') && (boolean)$this->modx->getOption('use_multibyte', null, false);
         $encoding = $this->modx->getOption('modx_charset', null, 'UTF-8');
 
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -45,6 +45,10 @@ class modOutputFilter
      */
     public function filter(&$element)
     {
+        if (!is_scalar($element->_output)) {
+            return;
+        }
+        
         $usemb = function_exists('mb_strlen') && (boolean)$this->modx->getOption('use_multibyte', null, false);
         $encoding = $this->modx->getOption('modx_charset', null, 'UTF-8');
 


rthrash:
@jenswittmann does that happen outside of pdotools?

jenswittmann:
> @jenswittmann does that happen outside of pdotools?

Yes, you can try this with a Snippet that return an Array:

```
<?php

return [
    'demo' => 'test'
];
```

When you use `[[demo:trim]]` the error occurs.

core/src/Revolution/Filters/modOutputFilter.php
```diff
@@ -63,7 +63,7 @@ public function filter(&$element)
 
                 $this->log('Processing Modifier: ' . $m_cmd . ' (parameters: ' . $m_val . ')');
 
-                $output = trim($output);
+                $output = trim((string) $output);
```
smg6511: I'd rather see the solution test for string rather than cast all values to that type. If I'm not mistaken, there would be legitimate cases where the value being worked with is not a string (_e.g._, int or bool, etc.), so maybe this:
```suggestion
                $output = is_string($output) ? trim($output) : $output;
```

smg6511 (changes requested):

core/src/Revolution/Filters/modOutputFilter.php
opengeek: It could be argued that all output filters should be operating only on strings. Perhaps it is worth looking at bypassing filtering altogether if the value is not a string?

smg6511: I don't _think_ so; what about the comparison filters (gt, lt, etc.) acting on the integer output of a snippet? The number TV, as it stands, does only output a string representation. But with that TV type  I could see there being an input option to force an integer, or a new filter that acts on a strict comparison. I think it just keeps your options (so to speak) open if you _don't_ coerce values passed to the filter to string. That's my 2¢ at least ;-)

opengeek: If I rephrase this to say scalar value, would you agree? Though technically, all elements in MODX that appear in content HAVE to return a string to be useful. You can obviously create snippets that return other (non-scalar) types in situations where you are using that snippet programmatically in PHP code, but when used in content, it is essentially useless. Since output filters are meant to act on the value returned to be used in content, I think my point is valid.

smg6511: Yeah, definitely agree with bypassing non-scalar vals. So maybe at the top of the filter method, add this:
``` php
if (!is_scalar($element->_output)) {
    return;
}
```
Then still test for string to trim instead of coercing as I suggested above?

jenswittmann:
@smg6511 here's a real-world example that might not work with the discussed fix maybe 🤔 

Check if the user has a name in a custom session variable and output a link based on it:

```
[[!#SESSION.user.password:is=``:then=`<a href="/login">Login</a>`:else=`<a href="/login?logout=true">Logout</a>`]]
```

Every output modifier gets trimmed in [line 66](https://github.com/modxcms/revolution/blob/d3d9df44cdcb35c40382f63c96ff16e5579ae072/core/src/Revolution/Filters/modOutputFilter.php#L66). So, with return, you always get the else value, right?

smg6511:
@jenswittmann - No, I did some tests with `[[#!GET.somekey:trim:is=...]]` in various ways and it works like you want it to with the early return. The benefit of the early return is that if you happen to unintentionally apply filters on an array, using `[[#GET:trim:is=...]]` for example, you'll just get back a string representation of the array (or just "Array"). Otherwise you would get the else value.

smg6511:
@jenswittmann Hi again, Jens. SInce you were active earlier today (on the file browser update I did), I thought I'd see if you wanted to take a second look and consider implementing the changes I'd suggested for your PR here. Then I can re-review and maybe we can get this one merged in soon as well ;-)

jenswittmann:
@smg6511 can confirm that both suggestion fixed it!

smg6511 (approved):
Looks good!

opengeek (changes requested):
Unfortunately, this breaks three of the unit tests with parsing and cannot be accepted as is.

```
There were 3 failures:

1) MODX\Revolution\Tests\Model\modParserTest::testProcessElementTags with data set #27 (array(1, 'Doesn't exist'), '[[!+invalid_tag:notempty=`<p>...ist`]]', array('', true, false, '[[', ']]', array(), 0))
Did not get expected results from parsing [[!+invalid_tag:notempty=`<p>[[!+[[!+invalid_tag]]]]</p>`:default=`Doesn't exist`]].
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    'processed' => 1
-    'content' => 'Doesn't exist'
+    'processed' => 0
+    'content' => '[[!+invalid_tag:notempty=`<p>[[!+[[!+invalid_tag]]]]</p>`:default=`Doesn't exist`]]'
 )

/home/runner/work/revolution/revolution/_build/test/Tests/Model/modParserTest.php:137

2) MODX\Revolution\Tests\Model\modParserTest::testProcessElementTags with data set #28 (array(1, 'Tag2'), '[[!+invalid_tag:notempty=`<p>...]]]`]]', array('', true, false, '[[', ']]', array(), 0))
Did not get expected results from parsing [[!+invalid_tag:notempty=`<p>[[!+tag[[+is1]]]]</p>`:default=`[[!+tag[[+is2]]]]`]].
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'processed' => 1
-    'content' => 'Tag2'
+    'content' => '[[!+invalid_tag:notempty=`<p>Tag1</p>`:default=`Tag2`]]'
 )

/home/runner/work/revolution/revolution/_build/test/Tests/Model/modParserTest.php:137

3) MODX\Revolution\Tests\Model\modParserTest::testDefaultNonExistingTvValue
Did not parse non-existing TV with default modifier correctly
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'[[*foo:default=`bar`]]'
+'bar'

/home/runner/work/revolution/revolution/_build/test/Tests/Model/modParserTest.php:927

--
```

core/src/Revolution/Filters/modOutputFilter.php
```diff
@@ -45,6 +46,10 @@ function __construct(modX &$modx)
      */
     public function filter(&$element)
     {
+        if (!is_scalar($element->_output)) {
+            return;
+        }
+
```
smg6511:
```suggestion
```

Ok, I see what we need to do. The short story is that we need null values to be preserved in certain instances, which is why limiting to scalar vals breaks some of the tests (I didn't go deep enough down the rabbit hole to discover _exactly_ why this is). So we need to kill this new conditional.

smg6511: To still catch incompatible values passed to the modifiers, the way to do it would be something like the following, inserted right before the for loop (this passes the unit tests):
```php
$commandName = $element->get('name');
try {
    // Check for invalid data types
    if (is_array($output)) {
        throw new \Exception("An element ($commandName) in document #{#} is attempting to pass its value(s) to an output modifier, but its data type (array) is not compatible. Only string, number, boolean, and null types may be passed.");
    }
} catch (\Exception $e) {
    $this->modx->log(modX::LOG_LEVEL_ERROR, $e->getMessage());
    return;
}
```
There are a couple to-dos with this block of code if we do it this way:

1. Get the reference data for the error (maybe warning) message (the doc id at minimum [how is this most easily done?])*
2. Convert this to a lexicon string with a couple placeholders instead of hard-coding the message.

* @opengeek (or anyone else looking who can provide the answer) - Is there an easy way to get the original full tag being worked with (e.g., `[[+something:is=X:then=Y:else=Z]]`)? I see how the deconstructed parts are gotten, but including at least a part of the full tag in the error message would help folks to find buggy tags.

smg6511 (changes requested):

core/src/Revolution/Filters/modOutputFilter.php
smg6511: @jenswittmann - Hey, even though you reverted the scalar conditional we still need to implement further handling as noted above to fully avoid fatal errors (at least when an array is the subject of a modifier). Keeping this thread unresolved will keep it more visible (I'm hoping Jason will chime in on my questions here).

Just FYI, as it stands now we avoid the fatal error so long as no string functions are called as modifiers (on an array subject). But if you try something like `[[demo:ucfirst]]` (referring to your earlier test example) where the output of subject `demo` is an array, the error will occur.

opengeek (approved):

core/src/Revolution/Filters/modOutputFilter.php
opengeek: @smg6511 — Sorry—I had no idea there was a question here for me. Can we move forward with this one and attack additional issues separately?

smg6511: @opengeek - Sure thing ;-)

smg6511 (approved):

PR #16759, opened 2025-07-16
Revert #14832 and Improve Tree Resource Contrast to Fix #16465

### LINKED ISSUE #16465: Usability Issue with Determining Unpublished vs Hidden
### Summary
The unpublished and hide from menu state are reversed on MODX 3 as compared to MODX 2. 

### Step to reproduce
Open a site tree in MODX 2 and then compare it to a matching site tree in MODX 3

### Observed behavior
On 3.x: 

Hidden  = italic
Unpublished = light color

On 2.x

Hidden = lite color
Unpublished = italic

### Expected behavior
The design should be consistent between the two, or possibly improved on 3.x to make it more obvious what is hidden vs unpublished.

jaygilmore:
*Requires asset build to see changes.*

### What does it do?
This PR reverts #14832 and implements improved color contrast and icon styling to resolve the usability issue described in #16465.

### Why is it needed?
- Fixes #16465
- Restores the previous SCSS files removed in #14832
- Adjusts color variables for better contrast and clarity (bringing it closter to acceptable contrast and difference ratio for accessibility)
- Aligns MODX 3 tree display with MODX 2 behavior for hidden/unpublished resources so that both versions have a consistent UX convention. 

### How to test
Checkout this PR, build assets and view the Resource Tree to see the changes. 

### Related issue(s)/PR(s)
Resolves #16465.

### Other Notes
For folks wishing to have the conventions within #14832, I'll follow in that with code for a plugin to handle that.

### Difference Image

after-and-before-pr

Commments and feedback welcome.

### DIFF
--- a/_build/templates/default/sass/_colors-and-vars.scss
+++ b/_build/templates/default/sass/_colors-and-vars.scss
@@ -132,7 +132,7 @@ $treeHeaderBorder: 1px solid #CDCDCD;
 $treeHeaderBg: $lightestGray;
 $treeBg: $lightestGray;
 $treeBgSelected: #D6E7F8;
-$treeColor: $darkGray;
+$treeColor: darken($darkGray,10%);
 $treeColorDesaturated: lighten(desaturate($treeColor,100%), 25%);
 $treeColor_2: #728DA7;
 $treePseudoRootBg: #F1F1F1;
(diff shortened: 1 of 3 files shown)


Ruslan-Aleev:
> Aligns MODX 3 tree display with MODX 2 behavior for hidden/unpublished resources so that both versions have a consistent UX convention.

It needs to be changed in the 2.x branch =)
I made a https://github.com/modxcms/revolution/pull/14832 and the complaint was that "Published & Visible" looks similar to "Unpublished & Visible", i.e. Published and Unpublished look almost the same, which is strange, because resource publication is a parameter that should be visible in the tree immediately. In my opinion, italic is not enough for visual separation.

My logic is as follows:
- Bright resource - published;
- Pale resource - unpublished;
- Italic - hidden in the menu (+ publication status);
In the 2.x branch it is mixed up, and bright can be both published and not, which, in my opinion, is illogical and breaks the UX.

jaygilmore:
Hey @Ruslan-Aleev, 

Thanks for your feedback.

Both approaches have merit, but I'm prioritizing consistency across the platform. The 2.x convention (italics for unpublished, light text for hidden from menu, standard for shown in menu) is established and functional.

Power users working with both 2.x and 3.x daily rely on this visual consistency. I've confirmed this with several developers who find the current inconsistency disruptive to their workflow.

Since 2.x is mostly in maintenance mode with limited changes, and many 2.x users won't upgrade to latest releases, it's logical to align 3.x with established 2.x conventions rather than modify 2.x. Users on 3.x are more likely to keep installations current.

The current solution addresses a real usability problem for active developers. For teams preferring the #14832 approach, I'll provide a plugin solution.

What are your thoughts on this direction?

Ruslan-Aleev:
@jaygilmore Hi!
I think the problem is not worth discussing, only a few people pay attention to it =)
I made the previous PR because both I and users had a problem with visual understanding of the publication status in the tree (and even now, I can hardly remember the logic of this design on 2.x, if the tree is not in front of my eyes =) ).
BUT if the behavior in 3.x creates more problems than benefits, then it is worth returning to the 2.x version, I think the plugin is not needed.
Thank you for your participation and suggested solutions!

jaygilmore:
@Ruslan-Aleev, @smg6511 I'm going to actually delete this PR and keep the current 3.x color scheme and increase the contrast slightly and then use the outlines for unpublished and deleted Resources. Again, it will only use font-weight to handle the icon outline display.

jaygilmore:
Closing in favour of #16760 16761

PR #16761, opened 2025-07-25
Resource Tree Icons Adjustment and Contrast Increase

jaygilmore:
### What does it do?
Differentiates Unpublished/Deleted Content from Published via the icons. Publish uses solid, unpublished/deleted uses outline versions. This helps folks quickly identify resources that are published to work on since unpublished and deleted are not generally active. 

Darkens the colors of the text for both the normal, unpublished and deleted item Resource Tree text to improve readibility and accessibility.

### Why is it needed?
The 3.x implementation of the tree adjustment was generally an improvement that brought a logical change to the text colors, but it made it slightly harder to identify published resorurces due to the Hidden from menu items being greyed out—even if published.

### Screenshots 
Side-by-side comparison of the MODX Revolution Manager interface. The top image (‘This PR’) shows updated resource tree icons with improved contrast and clearer visual distinction between resource types. The bottom image (‘Original 3.x’) shows the original icons with lower contrast and subtler visual differences.

### How to test
Check out this PR and run the build.

### Related issue(s)/PR(s)
This is an alternative to #16760 and replaces #16759 entirely by retaining the changes done originally by @Ruslan-Aleev in #14832. This is an alternative implementation that should also adequately address the issue in #16465.


SnowCreative:
I love using the outlined icons for unpublished resources, by why not keep the grayed-out versions for published & hidden ones? That gives an extra visual cue beyond just the italic text.

smg6511:
@SnowCreative - I think the idea is to avoid any crossover of meaning, which would require the gray be used to _either_ indicate unpublished _or_ hidden. In this case the choice was for gray to be the visual cue for unpublished in addition to the new outline style, and that makes sense to me.

SnowCreative:
> In this case the choice was for gray to be the visual cue for unpublished in addition to the new outline style, and that makes sense to me.

But the gray isn't what is indicating unpublished status for the icons. The outline is. So why not ALSO use gray in the icons, so that they can actually indicate all four of the possible statuses? Then outline in solid black would mean unpublished and visible, and outline in gray would mean unpublished and hidden.

The more I think about this, the more the method used in 2.X makes the most sense to me. Gray = hidden makes an awful lot of visual sense. The italics always bothered me a bit for unpublished, but using outlined icons instead of solid would add enough extra cue. This would also be the least confusing for people who are administering sites on both 2.x and 3.x.

menu

      
But I can totally see the argument for using gray to indicate published status, since that means it's not accessible on the web at all. If you really want to switch to using gray = unpublished, then use the outlines to indicate hidden status (ghost = hidden), not published status:

menu2

 If we were to go in this direction, I'd add some custom manager CSS to 2.x sites for clients who work on sites in both 2.x and 3.x so that the trees are consistent with each other.

smg6511:
Guys - I don't really think there's a wrong way with either Jay's submission, my suggestion, or Michael's suggestion (I like the 2nd one better there). If anything, it shows that this might be a highly personal preference (that perhaps gets appeased by way of a plugin, or new settings). It may also point to a need to work a different angle. 

I mocked up a little different way to consider going about this that uses a pattern already present in the 3.x design and adds a little convenience. Mind you, this is not to suggest a change to this PR — I think we should just make this basic incremental improvement, whatever it ends up being, and tend to a more refined change afterward.



### The base state
The idea here is to have a clickable icon that both indicates and allows 1-click control over menu visibility. This removes the need for italicized titles, which I've never loved. 

State - Base


### Hover state over hidden doc
The create new document appears as it always has, just bumped to the left to allow for the new menu visibility button (circle compass icon).

State - Hover

### Hover over create button
There's no change here, aside from the position of the + button.

State - Hover – 1

### Hover over menu visibility button
Clicking here would toggle visibility on.

State - Hover – 2


Or, clicking an already-visible document would toggle it off.

State - Hover – 3



Other subtle changes were made here too, like making doc ids a bit smaller and adjusting sizes and weights to the context title and tree tab headers. More fine-tuning would be in order, but this gets another idea out there.

Ruslan-Aleev:
@jaygilmore Thanks for the implementation, I like this option, it seems the logical and does not require special code edits (although changing icons adds unnecessary visual states).
The only thing is, I'm not sure that all icons of the current "Font awesome" set support a smaller font-weight.

SnowCreative:
> The idea here is to have a clickable icon that both indicates and allows 1-click control over menu visibility. This removes the need for italicized titles, which I've never loved.

@smg6511  Wow, that would be great. I'd love to be able to change visible status just by clicking a little icon in the tree, in addition to that icon showing the visibility status so we don't need italics for anything. I have a whole set of sites where admins are constantly updating the visibility status of pages, and this would make it so they don't have to load the whole page for editing just to do that.

You've also introduced another visual element — color. I've thought of using that in the past, and actually added some CSS to one site to color the titles and icons depending on status. That makes it super-easy to to see status without needing italics, although multi-colored titles might not be to everyone's liking.

smg6511:
> The only thing is, I'm not sure that all icons of the current "Font awesome" set support a smaller font-weight.

This thought crossed my mind too; the free FA5 version only gives you 89 of the 672 icons in outline form. The basic ones we need are there though. We may want to consider a future switch to Google's Material Symbols font, as it offers a lot more flexibility at no cost.

jaygilmore:
@SnowCreative 
> why not keep the grayed-out versions for published & hidden ones? That gives an extra visual cue beyond just the italic text.

This is a fair point. @rthrash felt unpublished needed a stronger visual cue. @opengeek and @rthrash and I discussed that hidden is a meta value and only matters in the cases where and extra actually chooses to. 

Also, if you want a plugin that preserves/returns the exact 2.x standard, I have a POC for that and could update it to do just that. I can have it in a day or so. Just say the word, @SnowCreative. 

---
@smg6511 I like the problem you're aiming to solve. I do think adding more icons has some potential downsides in terms of cognitive load and usability around a lot of little click targets in a small space but it does provide more feedback at a glance also.

I'm not sure if toggling menu visibility in the tree makes a tonne of sense where pub/unpub would always make sense to me. When I'm building a site I set menu visibility to off if it will be omitted but rarely after that point. Publishing is something I do more often—especially from a published to unpublished state. 

In any event, it's probably worth a standalone proposal.

smg6511:
@jaygilmore - Re your feedback to my proposed idea: Yeah, I definitely factored in the question of "is this going to be too much visual input" as I was thinking through this. It'd have to be actually prototyped in a proposal PR to really work it through. On the menu show/menu hide toggling, it'd simply be a nice feature for those who might do that a lot; you don't, Michael does ... just a matter of one's work style and nature of the sites they handle. We already have the contextual menu for pub/unpub; not as quick as a toggle button, but still a convenience that's already there.

Re your PR, I'm good with what it's doing, but you'll need to go about it differently to cover the issue that @Ruslan-Aleev raised (that only a small subset of solid icons are available in outline form). Take the file I've attached for a spin and see what you think. (I couldn't make suggested changes in the PR itself, as some of my changes are too far outside of the scope of yours.)

[_tree.scss.zip](https://github.com/user-attachments/files/21493846/_tree.scss.zip)

smg6511:
@jaygilmore Jay - giving you a little bump here. Did you get a chance to take a look at the solution in the file I attached?

jaygilmore:
> @jaygilmore Jay - giving you a little bump here. Did you get a chance to take a look at the solution in the file I attached?

I'm going to look at this tomorrow-ish. I don't get these notifications as I should.

smg6511:
@jaygilmore - I see you had eyes on this PR yesterday - I wanted to remind you to check out the file I attached with suggested changes ;-)

jaygilmore:
@smg6511 I started looking at your file and comparing mine. I can't see any difference in the results. Font weights are ignored if unsupported by the font and the next state is shown. So, if I switch to an icon that is only solid, the icons are still shown, when I use my file vs yours. Let me know if I'm missing something or if you think there's anything I should test that I didn't? 

Screenshot of MODX Revolution 3.2 interface showing the Resource tree with sections for Published & Visible, Published & Hidden, Unpublished & Visible, Unpublished & Hidden, and Deleted pages, each with corresponding icons indicating their visibility and publication status. The Edit Template panel for ‘BaseTemplate’ is open on the right.
Screenshot of MODX Revolution 3.2 interface showing the Resource tree with sections for Published & Visible, Published & Hidden, Unpublished & Visible, Unpublished & Hidden, and Deleted pages, each with corresponding icons indicating their visibility and publication status. The Edit Template panel for ‘BaseTemplate’ is open on the right.

smg6511:
@jaygilmore - I get the feeling you _may_ have run rebuild to inspect my updates but then did not rebuild again when you went back to compare them to yours (??). Or maybe you have something special in your environment that is handling missing icons? Whatever the case, there's really no falling back in terms of the FA icons (at least not out of the box); if a "weight" of a specific icon isn't in the free set the icon will not show. Here are a couple screenshots from my end to illustrate:

Without my change (this PR as-is)
Screen Shot 2025-11-03 at 11 06 17 PM

With change
Screen Shot 2025-11-03 at 11 15 24 PM

jaygilmore:
@smg6511 I'll take another look. This is really strange that the icons are not showing at the lower weights—unless the font supports the lower weight and we just didn't include it in the set we have with Revo. 

Out of curiosity, what browser were you using for testing?

smg6511:
@jaygilmore - In my experience here and elsewhere there's no such thing as FA fonts falling back to an available version; if you request a pro-level icon and only have free-level access you'll get no icon. But maybe you've seen different behavior. FWIW I don't think you've forgotten to include the available outline fonts; you can always verify on the FA site, which is what I did when taking a look at this.

In any case, I do most dev on Chrome/Mac (latest, or near latest version).

jaygilmore:
@smg6511 I figured it out, indeed, it's Chrome not behaving as Safari. The spec for fonts (including any glyph) is to match the weight with the next available weight and if not it should synthesize it. This is what Safari is doing. This is not what Chrome is doing. 

Anyhow, your fixes make sense. I'm not quite sure what the process might be to handle this for each font class. For example, (as shown in the image below) the deleted item icon vanishes as well. 

A screenshot showing the icons (house) in the MODX Revolution Manager's Resource tree. The icons are there except for the icon beside the deleted resource.

I'd be happy to figure out how to incorporate these but I could use your guidance, Jim. :)

smg6511:
@jaygilmore - Oh, interesting (re Safari doing swaps). I can take a closer look at dealing with other missing glyphs in the next couple days. I think we will want to stick to whitelisting the supported outline classes that are spec'd/needed by the core, instead of doing it for the entire supported subset (roughly 89 glyphs).  Cheers ;-)

smg6511:
@jaygilmore - Alright, I think this updated file should do it. I just needed to reorganize the changes a bit to cover the deleted state as well as the unpublished one. I think the "whitelist" of FA5 classes that have outline weights available is a good starting point (covers at the very least the icons specified by the core). And it would be easy to add in new classes as needed/requested by users.

Ultimately we probably want to look at replacing the use of FA with Google Material Symbols, as we won't have to deal with the free vs pro glyphs issue there.

Here's the updated file:
[_tree.scss.zip](https://github.com/user-attachments/files/23513594/_tree.scss.zip)

jaygilmore:
@smg6511, I've reviewed your changes, and they seem mostly alright in practice but they do not take into consideration the `.fa-folder` use to define a custom icon for a Template. When you change the custom font, it's ignoring the fonts that don't have an outline version. I'm not sure if there's an easy way to catalogue which custom icons have the availalbe outline. I'd be happy to incorporate this here or if you have an other way to handle it, I'd be happy to do that as well.  

At around line 214 you're referencing `&.icon-file` which is used for the default state classes but not for the overrides which expect the `.fa-file` format and as such, the rule for the `font-weight: 400` is not applied unless we define those `.fa-…` fonts. 

With your latest file: 
A screenshot of the MODX Resource tree showing the solid (not outlined) icons for folder on the unpublished and deleted Resources.


Adding in the `&.fa-folder` and `&.fa-file` to your section here: 

```scss
  /*
    Outline only free fa icons (89 of the 672 free solid icons)
    See https://fontawesome.com/v5/search?s=regular&ic=free&o=r
  */
  .unpublished &,
  .deleted & {
    &.fa-file,
    &.fa-folder,
    &.icon-file,
    &.icon-folder,
    &.icon-folder-open,
    &.icon-file-code,
    &.icon-file-pdf,
    &.icon-file-video,
    &.icon-file-audio,
    &.icon-file-image,
    &.icon-file-excel,
    &.icon-file-word,
    &.icon-file-powerpoint {
      &::before {
        font-weight: 400;
      }
    }
  }
```
Yields the following: 

a screenshot of the MODX Revolution Resource tree showing the custom defined Template tree icons for unpublished and deleted resources are outlined.

smg6511:
@jaygilmore - Right, and that's the idea with how I set up that "whitelist"—you found some cases that were missing and useful and easily added them to the list. I leave it up to you guys as to how expansive you want to make that list.

I don't know that doing something like creating a sass function to generate selectors based on the full list of 89 would be worthwhile; I thought about the possibility, but it might be more clear to list them explicitly. Although, maybe you gain something since there'd be two selectors needed for each icon (.icon-xxx and .fa-xxx).

How much effort gets put into this detail might depend on how determined you are to stick with FA in the future. The icons are great, but the free/pro divide is maybe an unnecessary challenge.

Anyway, hope that's helpful input. Let me know if you're wanting to go all out on this; I can help if so, just might take a little time ;-)

**UPDATE:** Actually it might not be too hard to put together a function to generate the inner "whitelist" selectors; the question is do you want all the css that'll get generated to support 178 (89 x 2) new selectors (given that arguably a decent amount of those are icons one would never specify for the tree)?

jaygilmore:
@smg6511 I've added the full list of the FA5 Free icons that support the regular (outlined) in the file. Did you want to check to make sure it makes sense for me? 
[_tree.scss.zip](https://github.com/user-attachments/files/23535412/_tree.scss.zip)

smg6511:
@jaygilmore - Yeah, looks good to me!

smg6511 (approved):

SnowCreative:
Looking at the latest version, I can say for certain that I don't think italics is enough of a visual cue for unpublished, and gray by itself is plenty of a cue for unpublished without any extra cue from the icons. So if/when this gets implements, I'll be creating a custom CSS to use the outlined icons for hidden status!

smg6511:
@SnowCreative - Also, at some point soon I'll put in a PR with my proposed solution you liked earlier in this discussion.

PR #16762, opened 2025-07-28
Add categoryId condition in GetListElement processor

catsmeatman:
### What does it do?
More powerful processor

### Why is it needed?
In CMP or custom Resources sometimes need filter elements by categoryId in GetListElement Processor.

### DIFF
--- a/core/src/Revolution/Processors/Element/GetList.php
+++ b/core/src/Revolution/Processors/Element/GetList.php
@@ -37,10 +37,19 @@ abstract class GetList extends GetListProcessor
         $c->select([
             'category_name' => 'Category.category',
         ]);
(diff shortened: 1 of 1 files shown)


smg6511:
Let's see what others say, but I think what you're wanting to do is best done by extending the appropriate GetList classes in your CMP to implement the filtering of Elements by category. To add this to the core code without there being any direct use of it in the core/manager UI can be problematic—imagine, at some time later, one of the core devs comes across these unused lines of code and removes them. That would break part of your CMP's functionality.

smg6511:
@catsmeatman @opengeek - Closing for reason stated in my earlier comment.

PR #16763, opened 2025-08-02
Make modNamespace's translatePath method null-safe

### LINKED ISSUE #16513: Namespace with a path set to null crashes ajax calls
### Summary
This is very much an edge case.

The code below will fail in PHP >= 8.1 if the path is null. If the 'path' or 'assets_path' of a namespace are null, this throws a deprecation warning (no warning if the value is empty). If it happens in a processor and the deprecation warning is returned, the JS will crash because the warning it not valid JSON. I think this only happens with an empty cache, because then MODX re-creates the namespace cache on every request.

This error also shows up in page content when the cache is empty.

It does not occur if display_errors is off or E_DEPRECATED errors are not caught. 

The two fields allow null to be set. Some extras that don't need an assets_path have it set to null. The only ones I know about are mine, but there may be others.

```
public static function translatePath(xPDO &$xpdo,$path) {
        return str_replace(array(
            '{core_path}',
            '{base_path}',
            '{assets_path}',
        ),array(
            $xpdo->getOption('core_path',null,MODX_CORE_PATH),
            $xpdo->getOption('base_path',null,MODX_BASE_PATH),
            $xpdo->getOption('assets_path',null,MODX_ASSETS_PATH),
        ),$path);
    }
```

The simplest way to solve it is to add this at the top of the function:

```
$path = $path===null? '' : $path;
```

It might also be a good idea to do that same conversion in modNamespace->save(), and if necessary, in the transport package code that saves the namespace (if it doesn't call the namespace's save() method).

### Step to reproduce
Set a namespace assets_path value to null, turn on display_errors, with error reporting set to display E_DEPRECATED errors.

You can also see this by setting up test code with the translatePath() code above and calling:

```
translatePath($modx, null);
```

### Observed behavior
error message

### Expected behavior
No error message

smg6511:
### What does it do?
Replaces NULL `$path` with empty string.

### Why is it needed?
The `str_replace` fn no longer quietly allows a null value to be passed to its subject param. Deprecation notices show in php logs when null is passed in php 8.1+.

### How to test

1. Change a namespace's core or assets path to NULL via sql query. 
2. Observe the php logs after clearing modx's cache (before loading this PR); you should see the deprecation notice(s). 
3. Load the PR and clear cache again. There should be no deprecation notices (of the type seen in step 2).

### Future Consideration
This could also be avoided by altering the _path_ and _assets_path_ columns in the namespaces table to be not null with default of empty string.

### Related issue(s)/PR(s)
No related issue.

### DIFF
--- a/core/src/Revolution/modNamespace.php
+++ b/core/src/Revolution/modNamespace.php
@@ -97,6 +97,9 @@ class modNamespace extends modAccessibleObject
 
     public static function translatePath(xPDO &$xpdo, $path)
     {
+        if ($path === null) {
+            $path = '';
+        }
         return str_replace([
             '{core_path}',
             '{base_path}',
--- a/core/src/Revolution/modNamespace.php
+++ b/core/src/Revolution/modNamespace.php
@@ -11,17 +11,17 @@ use xPDO\xPDO;
  * Represents a Component in the MODX framework. Isolates controllers, lexicons and other logic into the virtual
  * containment space defined by the path of the namespace.
  *
- * @property string                $name The key of the namespace
- * @property string                $path The absolute path of the namespace. May use {core_path}, {base_path} or {assets_path} as
+ * @property string $name The key of the namespace
+ * @property string $path The absolute path of the namespace. May use {core_path}, {base_path} or {assets_path} as
  * placeholders for the path.
- * @property string                $assets_path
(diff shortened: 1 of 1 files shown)


theboxer (approved):
looks good to me :)

PR #16769, opened 2025-09-17
Fix addEventListener in core/components/*/bootstrap.php and allow anonymous functions

### LINKED ISSUE #16768: Support persistent event listeners in core/components/*/bootstrap.php and allow anonymous functions
### Summary

Support persistent event listeners registered by `$modx->addEventListener($eventname, $pluginId)` in `bootstrap.php` files and allow attaching anonymous functions as event handlers.

### Why is it needed?

In MODX 3, packages can use `bootstrap.php` for initialization, which is very convenient. However:

* When adding events with `$modx->addEventListener($event, $pluginId)` inside `bootstrap.php`, they do not work as expected because later the `eventMap` is reloaded and overwrites these changes.
* This forces developers to always create plugins even for simple cases, which adds unnecessary boilerplate.

Having the ability to:

1. Register persistent events directly in `bootstrap.php`.
2. Attach closures or callable functions as event handlers.

…would make package development cleaner and more flexible.

### Suggested solution(s)

1. Ensure that event listeners registered in `bootstrap.php` are not removed/overwritten when the `eventMap` is reinitialized.
2. Extend `$modx->addEventListener()` to support callbacks/closures, for example:

```php
$modx->addEventListener('OnWebPageInit', function($modx, $event) {
    // custom logic here
});
```

or create new method like `addEventListenerClosure`.

This would allow developers to bind logic directly in code without creating a separate plugin element.

### Related issue(s)/PR(s)

Not found yet.

OlegAnTo2000:
### What does it do?

This PR introduces **runtime and closure-based event listeners** with support for **priority** and **optional names** in `modX`.

I tried to leave 100% backward compatibility - we get a similar addEventListener syntax and correct operation of plugins and events, while getting a powerful tool for managing listeners in packages.

Main technical changes:

* Added two new internal maps:

  * `runtimeEventMap` — stores runtime listeners added via `addEventListener()` so they are not overwritten by `_initEventMap()`.
  * `closureEventMap` — stores closure-based listeners with `priority`, `name`, and `callback`.

* Added new method:

  ```php
  public function addEventListenerClosure(
      string $event,
      \Closure $callback,
      int $priority = 0,
      ?string $name = null,
      bool $replace = true
  ): bool
  ```

  This allows developers to register closure-based listeners with configurable execution order and optional unique names.

* Updated `invokeEvent()`:

  1. Executes persistent DB plugins (`eventMap`) as before.
  2. Executes runtime listeners (`runtimeEventMap`).
  3. Executes closure listeners (`closureEventMap`), sorted by priority (lower = earlier).

* Added universal removal method:

  ```php
  public function removeEventListenerClosure(
      ?string $event = null,
      ?int $priority = null,
      ?string $name = null
  ): bool
  ```

  Behavior:

  * If only `$event` is provided → remove listeners for that event.
  * If `$event` is null and `$name` provided → remove all listeners with that name across all events.
  * If `$event` is null and `$priority` provided → remove all listeners with that priority across all events.
  * If everything is null → clear all closure listeners.
  
 * Improved the `removeEventListener` method:
   - It can now remove listeners by plugin ID or all handlers if the plugin ID is not specified.
   - It also removes from the new runtime and closure event maps.
  
 * Clearing of `runtimeEventMap` and `closureEventMap` has been added to the old `removeAllEventListener` method.
  
 * Added `removeAllRuntimeEventListeners` method which clears new event maps `runtimeEventMap` and `closureEventMap`.

### Why is it needed?

Currently, `addEventListener()` writes into `$eventMap`, which is later reset by `_initEventMap()` when loading from DB. As a result, listeners registered in `bootstrap.php` or other runtime code are lost.

With this PR:

* Runtime listeners persist throughout the request lifecycle.
* Developers can attach closures without needing to define plugins in the database.
* Execution order is predictable thanks to `priority`.
* Closure Listeners can be easily replaced or removed by `name`, which simplifies package development and testing.

This feature makes it possible to extend MODX in a more lightweight and programmatic way.

---

### How to test

- Use regular modx project

- Go to project `core` directory and change `modX.php` file from my repo
  ```
  cd yourmodxroot/core
  curl https://raw.githubusercontent.com/OlegAnTo2000/revolution/fix-bootstrap-events/core/src/Revolution/modX.php -o src/Revolution/modX.php
  ```

- Download and install (or create your own) package with `bootstrap.php` file, for example pdoTools.

- **Register a closure listener in `core/components/yourextra/bootstrap.php`:**

   ```php
   $modx->addEventListenerClosure(
       'OnWebPagePrerender',
       function (array $params, \MODX\Revolution\modX $modx) {
           $output = &$modx->resource->_output;
           $output .= "\n<!-- TEST CLOSURE: " . date('Y-m-d H:i:s') . " -->";
           return null;
       },
       priority: 5,
       name: 'appendTestComment'
   );
   ```

   * Open any frontend page.
   * View page source.
   * Expected: the HTML ends with a comment like:

     ```html
     <!-- TEST CLOSURE: 2025-09-17 20:45:00 -->
     ```

- **Verify priority order:**

   ```php
   $modx->addEventListenerClosure(
       'OnWebPagePrerender',
       function (array $params, \MODX\Revolution\modX $modx) {
           $output = &$modx->resource->_output;
           $output .= "\n<!-- TEST CLOSURE 1: " . date('Y-m-d H:i:s') . " -->";
           return null;
       },
       priority: 5,
       name: 'appendTestComment',
       replace: false
   );
   $modx->addEventListenerClosure(
       'OnWebPagePrerender',
       function (array $params, \MODX\Revolution\modX $modx) {
           $output = &$modx->resource->_output;
           $output .= "\n<!-- TEST CLOSURE 2: " . date('Y-m-d H:i:s') . " -->";
           return null;
       },
       priority: 10,
       name: 'appendTestComment',
       replace: false
   );
   ```

   * Refresh the page.
   * Expected order:

     ```
     <!-- TEST CLOSURE 1 ... -->
     <!-- TEST CLOSURE 2 ... -->
     ```

   Swap priorities and confirm that order changes accordingly.
   
- **Test replace by name:**
   ```php
   $modx->addEventListenerClosure(
       'OnWebPagePrerender',
       function (array $params, \MODX\Revolution\modX $modx) {
           $output = &$modx->resource->_output;
           $output .= "\n<!-- TEST CLOSURE 1: " . date('Y-m-d H:i:s') . " -->";
           return null;
       },
       priority: 5,
       name: 'appendTestComment',
       replace: true // can be replaced
   );
   $modx->addEventListenerClosure(
       'OnWebPagePrerender',
       function (array $params, \MODX\Revolution\modX $modx) {
           $output = &$modx->resource->_output;
           $output .= "\n<!-- TEST CLOSURE 2: " . date('Y-m-d H:i:s') . " -->";
           return null;
       },
       priority: 10,
       name: 'appendTestComment',
       replace: true // can be replaced
   );
   ```
   * Refresh the page.
   * Only the second comment was expected:

     ```
     <!-- TEST CLOSURE 2 ... -->
     ```

- **Test removal by event:**

   ```php
   $modx->removeEventListenerClosure('OnWebPagePrerender');
   ```

   * Refresh the page.
   * Expected: no closure-based comments at all.

- **Test removal by name (global):**

   ```php
   $modx->removeEventListenerClosure(null, null, 'appendTestComment');
   ```

   * Only listeners named `appendComment` should be removed across all events.

- **Test removal by priority:**

   ```php
   $modx->removeEventListenerClosure(null, 5, null);
   ```

   * All closure listeners with priority 5 should be removed.

- **Test error handling:**

   ```php
   $modx->addEventListenerClosure('OnWebPagePrerender', function () {
       throw new \Exception("Boom!");
   }, name: 'errorTest');
   ```

   * Refresh the page.
   * Expected: page renders normally, but error log contains:

     ```
     [OnWebPagePrerender] Closure failed: Boom!
     ```

---

### Related issue(s)/PR(s)

Closes #16768

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -119,6 +119,14 @@ class modX extends xPDO {
      * @var array A map of elements registered to specific events.
      */
     public $eventMap= null;
+    /** 
+     * @var array<string, array<int, string>> Runtime listeners added via addEventListener() 
+     */
+    protected $runtimeEventMap = [];
+    /**
+     *  @var array<string, array<int, \Closure>> Closure-based listeners added via addEventListenerClosure() 
+     */
+    protected $closureEventMap = [];
     /**
      * @var array A map of already processed Elements.
      */
@@ -1679,61 +1687,94 @@ class modX extends xPDO {
      * @param array $params Optional params provided to the elements registered with an event.
      * @return bool|array
      */
-    public function invokeEvent($eventName, array $params= []) {
-        if (!$eventName)
-            return false;
-        if ($this->eventMap === null && $this->context instanceof modContext)
+    public function invokeEvent($eventName, array $params = []) {
+        if (!$eventName) return false;
+    
+        // Initialize the map from the database if necessary
+        if ($this->eventMap === null && $this->context instanceof modContext) {
             $this->_initEventMap($this->context->get('key'));
-        if (!isset ($this->eventMap[$eventName])) {
-            //$this->log(modX::LOG_LEVEL_DEBUG,'System event '.$eventName.' was executed but does not exist.');
-            return false;
         }
-        $results= [];
-        if (count($this->eventMap[$eventName])) {
-            $this->event= new modSystemEvent();
-            foreach ($this->eventMap[$eventName] as $pluginId => $pluginPropset) {
+    
+        $results = [];
+    
+        // 1) Build a combined list of plugins: persistent (from the database) + runtime
+        $persistent = (isset($this->eventMap[$eventName]) && is_array($this->eventMap[$eventName]))
+            ? $this->eventMap[$eventName]
+            : [];
+    
+        $runtime = (!empty($this->runtimeEventMap[$eventName]) && is_array($this->runtimeEventMap[$eventName]))
+            ? $this->runtimeEventMap[$eventName]
+            : [];
+    
+        // Combine them so that persistent ones have priority (we do not allow duplicates for the same pluginId)
+        $combinedPlugins = $persistent + $runtime;
+    
+        // 2) Calling plugins (as before), but now using $combinedPlugins
+        if (!empty($combinedPlugins)) {
+            $this->event = new modSystemEvent();
+    
+            foreach ($combinedPlugins as $pluginId => $pluginPropset) {
                 /** @var modPlugin $plugin */
-                $plugin= null;
+                $plugin = null;
                 $this->Event = clone $this->event;
                 $this->event->resetEventObject();
-                $this->event->name= $eventName;
-                if (isset ($this->pluginCache[$pluginId])) {
-                    $plugin= $this->newObject(modPlugin::class);
+                $this->event->name = $eventName;
+    
+                if (isset($this->pluginCache[$pluginId])) {
+                    $plugin = $this->newObject(modPlugin::class);
                     $plugin->fromArray($this->pluginCache[$pluginId], '', true, true);
                     $plugin->_processed = false;
                     if ($plugin->get('disabled')) {
-                        $plugin= null;
+                        $plugin = null;
                     }
                 } else {
-                    $plugin= $this->getObject(modPlugin::class, ['id' => intval($pluginId), 'disabled' => '0'], true);
+                    $plugin = $this->getObject(modPlugin::class, ['id' => intval($pluginId), 'disabled' => '0'], true);
                 }
+    
                 if ($plugin && !$plugin->get('disabled')) {
                     $this->event->plugin =& $plugin;
-                    $this->event->activated= true;
-                    $this->event->activePlugin= $plugin->get('name');
-                    $this->event->propertySet= (($pspos = strpos($pluginPropset, ':')) >= 1) ? substr($pluginPropset, $pspos + 1) : '';
-
-                    /* merge in plugin properties */
-                    $eventParams = array_merge($plugin->getProperties(),$params);
-
-                    $msg= $plugin->process($eventParams);
-                    $results[]= $this->event->_output;
+                    $this->event->activated = true;
+                    $this->event->activePlugin = $plugin->get('name');
+                    $this->event->propertySet = (($pspos = strpos($pluginPropset, ':')) >= 1) ? substr($pluginPropset, $pspos + 1) : '';
+    
+                    $eventParams = array_merge($plugin->getProperties(), $params);
+    
+                    $msg = $plugin->process($eventParams);
+                    $results[] = $this->event->_output;
+    
                     if ($msg && is_string($msg)) {
                         $this->log(modX::LOG_LEVEL_ERROR, '[' . $this->event->name . ']' . $msg);
                     } elseif ($msg === false) {
                         $this->log(modX::LOG_LEVEL_ERROR, '[' . $this->event->name . '] Plugin ' . $plugin->name . ' failed!');
                     }
+    
                     $this->event->plugin = null;
-                    $this->event->activePlugin= '';
-                    $this->event->propertySet= '';
+                    $this->event->activePlugin = '';
+                    $this->event->propertySet = '';
                     if (!$this->event->isPropagatable()) {
                         break;
                     }
                 }
             }
         }
-        return $results;
+    
+        if (!empty($this->closureEventMap[$eventName])) {
+            usort($this->closureEventMap[$eventName], function ($a, $b) {
+                return $a['priority'] <=> $b['priority'];
+            });
+        
+            foreach ($this->closureEventMap[$eventName] as $listener) {
+                try {
+                    $results[] = $listener['callback']($params, $this);
+                } catch (\Throwable $e) {
+                    $this->log(modX::LOG_LEVEL_ERROR, "[{$eventName}] Closure failed: " . $e->getMessage());
+                }
+            }
+        }
+    
+        return $results ?: false;
     }
+    
 
     /**
      * Loads and runs a specific processor.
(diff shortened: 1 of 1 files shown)


core/src/Revolution/modX.php
```diff
@@ -2062,25 +2111,110 @@ public function logManagerAction($action, $class_key, $item, $userId = null) {
...
+     *                          If null/empty, and $name=null — clear entire closureEventMap.
+     * @param ?int    $priority If set, remove only listeners with this priority.
+     * @param ?string $name     If set, remove only listeners with this name.
+     * @return bool True if something was removed.
+     */
+    public function removeEventListenerClosure(?string $event = null, ?int $priority = null, ?string $name = null): bool
+    {
+        $removed = false;
+
+        // Ветка 1: полный wipe
```
rthrash, line 2177: Can we get all code comments in English, please?

core/src/Revolution/modX.php
```diff
@@ -2062,25 +2111,110 @@ public function logManagerAction($action, $class_key, $item, $userId = null) {
...
+        if (empty($event) && $name === null && $priority === null) {
+            $this->closureEventMap = [];
+            return true;
+        }
+
+        // Ветка 2: поиск по всем событиям (если event пустой, а имя или приоритет заданы)
```
rthrash: Same as above (in English please!)

core/src/Revolution/modX.php
Ibochkarev: @OlegAnTo2000 ping

Ibochkarev:
## Summary

Introduces **runtime and closure-based event listeners** in `modX` so that listeners registered in `bootstrap.php` or other runtime code are not overwritten by `_initEventMap()`. Adds `addEventListenerClosure()` for closure-based listeners with priority and optional names, plus `removeEventListenerClosure()` and `removeAllRuntimeEventListeners()` for management.

**Main changes:**
- `addEventListener()` now writes to `runtimeEventMap` instead of `eventMap`, preserving listeners across `_initEventMap()` calls.
- New `closureEventMap` and `closureEventSeq` for closure listeners with priority and stable ordering.
- `invokeEvent()` executes: 1) persistent DB plugins, 2) runtime listeners, 3) closure listeners (sorted by priority, then sequence).
- New methods: `addEventListenerClosure()`, `removeEventListenerClosure()`, `removeAllRuntimeEventListeners()`.

---

## Suggestions

### 1. Code comments must be in English

As noted by @rthrash, several comments are in Russian and should be translated:

- `// Ветка 1: полный wipe` → `// Branch 1: full wipe`
- `// Ветка 2: поиск по всем событиям (если event пустой, а имя или приоритет заданы)` → `// Branch 2: search across all events (when event is empty but name or priority is set)`
- `return true; // оставляем` → `return true; // keep`
- `return false; // совпал — удаляем` → `return false; // match — remove`

---

### 2. `invokeEvent()` return value — possible breaking change

**Current (original):** Returns `[]` when the event exists but no plugins run; returns `false` only when the event is not in `eventMap`.

**New (PR):** `return $results ?: false` — when `$results` is empty (`[]`), `false` is returned.

So for existing events with no listeners or no output, the PR now returns `false` instead of `[]`. Code that distinguishes “event not found” (`false`) from “event ran with no output” (`[]`) would be affected.

**Recommendation:** Preserve `[]` when the event exists and has run (even with no output):

```php
// At the end of invokeEvent()
return $results;
```

The original only returned `false` when `!isset($this->eventMap[$eventName])`. The new logic no longer has that early return; it builds `$persistent` and `$runtime` from the maps. If the event exists in any map, we should return `$results` (possibly `[]`). If it exists in none, return `false`. Suggested logic:

```php
$hasListeners = !empty($combinedPlugins) || !empty($this->closureEventMap[$eventName] ?? []);
return $hasListeners ? $results : false;
```

Or, to keep behavior closer to the original: return `false` only when the event is not registered anywhere:

```php
$eventExists = isset($this->eventMap[$eventName])
    || !empty($this->runtimeEventMap[$eventName] ?? [])
    || !empty($this->closureEventMap[$eventName] ?? []);
return $eventExists ? $results : false;
```

---

### 3. `removeEventListenerClosure()` — filter logic

The filter uses `$priority !== null && (int)$l['priority'] !== $priority` → keep. When both `$priority` and `$name` are provided, the logic keeps the listener if **either** condition fails (priority doesn’t match **or** name doesn’t match). So removal is by `$priority` **and** `$name` together.

**Recommendation:** Clarify in the docblock whether removal is by `$priority` **and** `$name` together, or by `$priority` **or** `$name`. If the intent is “remove all listeners matching both `$priority` and `$name` when both are set”, the current logic is correct. If the intent is “remove by priority OR by name when either is set”, the filter logic should be adjusted.

---

### 4. `closureEventMap` — in-place sort on every invoke

`usort()` is called on `$this->closureEventMap[$eventName]` on every `invokeEvent()` for that event. That mutates the array and re-sorts each time.

**Recommendation:** For performance, consider sorting only when listeners are added (or when a listener with a specific priority is added). For typical use cases, the current approach is acceptable.

---

### 5. `addEventListenerClosure()` — default priority

**Current:** `$priority = 10`.

**PR description:** “Lower = earlier execution (default 10)”.

**Recommendation:** Ensure the docblock is consistent with the PR description and that the default is intentional for the intended use cases.

---

### 6. `removeEventListener()` — closure removal when `$pluginId` is set

When `$pluginId` is non-zero, `removeEventListener()` removes only runtime listeners for that plugin. Closures are removed only when `$event` is set and `$pluginId` is 0:

```php
if (!empty($event) && isset($this->closureEventMap[$event]) && !intval($pluginId)) {
    unset($this->closureEventMap[$event]);
    $removed = true;
}
```

So `removeEventListener('OnWebPagePrerender', 123)` does not remove closures for that event. That is consistent with the idea that closures are not tied to plugin IDs.

---

## Assessment

### Strengths

- Fixes the core issue: listeners added in `bootstrap.php` or other runtime code are no longer lost when `_initEventMap()` runs.
- Closure-based listeners are a useful addition for packages and extensions.
- Priority and ordering are well-defined.
- Named listeners can be replaced or removed.
- Error handling for closures: exceptions are caught and logged, and the event continues.
- `removeEventListener()` and `removeAllEventListener()` correctly clear the new maps.

### Concerns

- Return value change: `[]` vs `false` for events with no listeners.
- Russian comments in code (should be English).
- Possible ambiguity in `removeEventListenerClosure()` when both `$priority` and `$name` are provided.

---

## Verdict

**Approve with minor changes** — recommend addressing:

1. Translate all Russian comments to English.
2. Clarify or adjust `invokeEvent()` return value to preserve backward compatibility when the event exists but no listeners run.
3. Add tests for `addEventListenerClosure`, `removeEventListenerClosure`, and priority ordering.

The feature is well-designed and aligns with the goals of #16768. The suggested changes are small and should improve clarity and compatibility.

PR #16773, opened 2025-09-21
Fix and Enhance Resource Groups Functionality

### LINKED ISSUE #16274: Bug when drag a resource from one resource-group into another resource-group
### Summary
In the resourcegroup manager page (in ressourcegroup tree), drag & drop resource into other resourcegroup does not work.

### Step to reproduce
Always when trying to drag and drop a ressource in other ressourcegroup.

### Observed behavior
In MODx core processor (security/ressourcegroup/updateresourcesin.class.php), in line 36, the var $resourceId get not the real ressourceId.

### Expected behavior
How it should behave after following steps above.

### Environment
MODX version 2.8.3, 2.8.4 (I think MODx3 also), Apache 2.4.53 , MariaDB 10.4.24, Firefox 100

### Exemple
http://www.revoweb.ch/modx/modx_ressourcesgroups-2022-05-18_20.49.29.mp4

smg6511:
### What does it do?
Fix the parsing of differing node id formats to allow moving and copying of Resources _between_ Resource Groups.

Along with fixing the original issue, this solution adds the ability to _copy_ a Resource from one group into another by use of a modifier key (opt/alt - click) before/while dragging. New language in the panel's main description has been added to make this clear to users.

### Why is it needed?
There was no group to group functionality. Attempting to move a Resource between groups would result in nothing happening and/or a endless progress dialog box.

### How to test
Go to the Resource Groups editing panel (manager/?a=security/resourcegroup) and ensure the following works:
- Moving Resources from the right Contexts tree into one or more groups (this is unchanged)
- Moving a Resource from one group to another (drag drop _without_ modifier key)
- Copying a Resource from one group to another (opt/alt - click - drag drop)
- The create and update windows; there's no new functionality here, just some code clean ups and little nuances fixed (such as the field error message shown when the group name is not specified)

### Related issue(s)/PR(s)
Resolves #16274 for MODX 3.x.


theboxer (approved):
works good, thank you

PR #16777, opened 2025-09-29
Fix Resource Preview Functionality

### LINKED ISSUE #16195: use_weblink_target breaks weblink preview if target is ID or link tag
### Summary
When `use_weblink_target` is enabled, weblinks that use an ID or [[~ID]] format no longer work.  Issue stems from the modContext class's makeUrl's handling of that variable. 

### Step to reproduce
Enable `use_weblink_target` in the system settings. Create a new Weblink with the ID of an existing resource. Save and Preview.

### Observed behavior
The Weblink will take you to example.com/[[~ID]] or example.com/ID instead of the parsed link. 

### Expected behavior
It should work like standard weblinks and allow for the content to be parsed as a resource ID or dynamically created link. 

### Environment
Observed using 2.8.4 in MODX Cloud.

smg6511:
### What does it do?

- Updates processors and frontend js to properly set the state and visibility of the preview ("View") button (in the Resource panel UI) and Resource tree contextual menus.
- Adds new logic to determine validity of web and sym link content to more reliably build the correct (or empty) preview URL.
- Fixes the logic setting a Resource's deleted state in the UI (changing via the tree and/or panel button was not setting the switch in the settings area).

### Why is it needed?
Primarily, as noted in the issue solved, weblink URLs were not being built because the target Resource's URL was not being parsed/built. Also, discovered along the way of fixing this, the view button/tree item was not always showing/hiding/being disabled as expected. Lastly, one could pass obviously malformed content for weblinks resulting in bad links and previews.

### How to test

- Create various Resource types and toggle their deleted setting (from both the panel UI and tree), noting that this action should affect the "View" button states and tree menu items as expected (a deleted Resource should hide its View option).
- Create valid web and sym links and verify that the correct document is shown when clicking "View"
- Create invalid weblinks in various ways and verify the "View" button is disabled and the tree menu item is hidden.

### Note
I made the choice to disable instead of hide the View button for invalid web/symlinks, in contrast to how it's hidden for deleted Resources. I'm of the mind that action buttons/links shouldn't be hidden at all, just disabled. (It seems somewhat jarring for items to appear and disappear from the main UI.) This would be similar to the common pattern of showing, say, a Submit or Save button in a disabled state until the form it relates to is completely filled out. Let me know your thoughts on this, as I'd move to make a few more changes to make this behavior consistent.

### Related issue(s)/PR(s)
Resolves #16195


theboxer:
@smg6511 hey, it's really hard to see what actually changed with the formatting changes mixed up with everything else. can u nuke those and if needed just create separate pr with just the formatting updates?

smg6511:
@theboxer - Greetings - yeah, just look at the first few commits on the tree class to see the functional changes. I did separate out those versus formatting/etc. so folks could see them easily ;-)

opengeek:
> @theboxer - Greetings - yeah, just look at the first few commits on the tree class to see the functional changes. I did separate out those versus formatting/etc. so folks could see them easily ;-)

It becomes difficult to review the PR as a whole this way. Moving forward, let's avoid reformatting when changing code. Reformatting should only occur on lines that change or should be done in PRs specifically meant to reformat code without making any logic changes.

opengeek (approved):

modxcommunity:

https://community.modx.com/t/clientconfig-setting-not-working-in-weblinks/8986/2

PR #16778, opened 2025-09-29
Fix z-index issue in the menu of MODX 3.x.

Jako:
### What does it do?
Add a z-index: 0 to #modx-content and remove some not needed z-index in the navbar.

### Why is it needed?
Fix the z-index issue in the menu of MODX 3.x. 

The reason for this issue can be found in https://coder-coder.com/z-index-isnt-working/#3-youre-setting-opacity-transform-or-other-css-properties-that-put-the-element-in-a-new-stacking-context

The flyouts of the menu are hidden with opacity: 0. This will put these elements into its own, new stacking context, which does not play well with #modx-content items having a z-index. This will be fine when #modx-content has a z-index: 0.

Hopefully this will solve other z-index issues too.

### How to test

Compile the CSS and check custom manager pages using elements with a z-index (i.e. in dbAdmin, when Ace is installed).

### Related issue(s)/PR(s)
#16711 and others

#16690 and #16712 can be maybe reverted.


smg6511 (approved):
For others who review this, you likely won't see a difference unless you have a specific scenario where visibility issues are caused by the previous z-index values (or lack thereof in the case of the #modx-content element). In my dev setup I don't see a difference before and after these changes, but those changes may be useful for preventing future issues.

SIde note: It occurs to me that maybe we should establish a range of z-indexes that core MODX uses, say 0-100 and maybe one high one meant to “override” all page elements, including anything generated by an Extra (maybe 999 for system modals and such). There's no reason to have ridiculously-high z-indexes if there's a rule for how we use them. Then it would be up to Extra devs to adhere to these ranges, using 101-998 for placing their elements above the core layout or below that range to weave something between core elements (unlikely to be done, but possible). Anyway, just an idea for future consideration ;-)

Jako:
You can test the change of the menu rendering in dbAdmin, when Ace is installed.

Before: 
 Screenshot before

After:
Screenshot after

modxcommunity:

https://community.modx.com/t/tinymce-rte-fullscreen-issue/8949/2

smg6511:
@jako and others - This PR fixes the issue of stacking when not in fullscreen mode. The fullscreen mode problems still exist to varying degrees, but I believe those are up to the Extra's devs to fix. 

- The latest TinyMCE Rich Text Editor (3.1.6) gets very close, but with the open/close carrot icons in the main UI still peaking through; 3.1.5 does not stack properly for the most part
- Ace is not stacking right - several main UI elements are exposed still
- Monaco has same problems as Ace
- Redactor gets it right, completely hiding the UI below

@opengeek - The above said, I think this PR is good to go.

Jako:
Maybe the visible elements are caused by the same issue as the menu. The caret icons are transformed by css and put into its own, new stacking context, that way. It can be solved by adding a wrapping element into a defined z-index.

PR #16779, opened 2025-10-14
Email logo to png extension

hylkest:
### What does it do?
I changed the extension from .svg to .png and uploaded a new png logo

### Why is it needed?
Some email providers for example Gmail are blocking .svg images in their email. So the image won't load if you view it in gmail.

### How to test
Use the "Let the user choose their own password via email" option for example and check the email in gmail. The image won't load

Screenshot 2025-10-14 at 14 13 59


manager/controllers/default/security/login.class.php
```diff
@@ -90,7 +90,7 @@ public function process(array $scriptProperties = []) {
         }
         $this->setPlaceholder('background', $background);
 
-        $logo = $this->modx->getOption('login_logo', null, $managerUrl . 'templates/' . $managerTheme . '/images/modx-logo-color.svg', true);
+        $logo = $this->modx->getOption('login_logo', null, $managerUrl . 'templates/' . $managerTheme . '/images/modx-logo-color.png', true);
```
smg6511, line 93: This is actually the wrong place to change this, as this is for the web UI not the email html. You need to alter the email template at `manager/templates/default/email/default.tpl`. But, it'd be preferable to provide the png as a fallback to the svg, maybe trying something like:
```html
<img 
     src="{$_config.url_scheme}{$_config.http_host}{$_config.manager_url}templates/default/images/modx-logo-color.png" 
     alt="{$_config.site_name}"
     srcset="{$_config.url_scheme}{$_config.http_host}{$_config.manager_url}templates/default/images/modx-logo-color.svg" 
>
```

I'd also suggest providing the PNG at a minimum of 2x size (~440px wide), maybe even 3x (~650px). It's sized down in the css, so no worries on physical size; this will give it a sharper appearance on the mostly high-dpi devices the message will be read with.

opengeek: Any chance you can follow-up on @smg6511's comment @hylkest ?

opengeek (changes requested):

Ibochkarev:
## What the PR does

1. Changes the default `login_logo` in `login.class.php` from `.svg` to `.png`
2. Adds `modx-logo-color.png` (3.7 KB)

---

## The issue: wrong target

The **email template** at `manager/templates/default/email/default.tpl` has its own fallback:

```smarty
{if $_config.login_logo}
    <img src="{$_config.login_logo}" alt="{$_config.site_name}"/>
{else}
    <img src="{$_config.url_scheme}{$_config.http_host}{$_config.manager_url}templates/default/images/modx-logo-color.svg" alt="{$_config.site_name}"/>
{/if}
```

When `login_logo` is not set, the email template uses the hardcoded `.svg` path. Changing the default in `login.class.php` does not affect this fallback, so emails will still use SVG when no custom logo is configured.

---

## Recommended approach

**1. Update the email template**

Change the fallback in `manager/templates/default/email/default.tpl` from `.svg` to `.png`:

```smarty
<img src="{$_config.url_scheme}{$_config.http_host}{$_config.manager_url}templates/default/images/modx-logo-color.png" alt="{$_config.site_name}"/>
```

**2. Optional: PNG with SVG fallback for capable clients**

As @smg6511 suggested, use PNG as the main source and SVG in `srcset` for clients that support it:

```html
<img src="{$_config.url_scheme}{$_config.http_host}{$_config.manager_url}templates/default/images/modx-logo-color.png" 
     alt="{$_config.site_name}"
     srcset="{$_config.url_scheme}{$_config.http_host}{$_config.manager_url}templates/default/images/modx-logo-color.svg" />
```

**3. Login page**

The login page can stay on SVG (better for web: scalable, smaller). If you prefer consistency, keep the PNG default in `login.class.php`, but the main fix must be in the email template.

**4. PNG resolution**

For retina displays, consider a 2x (~440px) or 3x (~650px) PNG. The CSS keeps the display size at ~217px, so a larger source will look sharper on high-DPI screens.

---

## Summary

| Change | Current PR | Should be |
|--------|------------|-----------|
| `login.class.php` | .svg → .png | Optional (web can stay SVG) |
| `email/default.tpl` | Not changed | Change fallback to .png |
| `modx-logo-color.png` | Added ✓ | Keep |

**Verdict:** The PNG asset is fine, but the fix must be applied in `manager/templates/default/email/default.tpl`, not in `login.class.php`. Please update the email template as described above.

smg6511:
Closing in favor of #16953 since no response here to requested changes.

PR #16782, opened 2025-10-29
Fix and enhance checkbox and radio TV rendering

### LINKED ISSUE #16801: TV checkboxes width is set to fixed narrow width
### Summary
A simple checkobox list TV is displayed incorrectly  in manager. The problem is that some elements have a fixed width value.

Image

### Expected behavior
Full width label

### Environment
MODX 3.1.2

smg6511:
### What does it do?
Updated css and added new input properties to the two TV types to give users better control over their display, particularly for multi-column rendering. The new properties added are:
- Column Direction
- Column Width
- Wrap Column Text
Chooser Opts

### Why is it needed?
When these TV types are set to more than one column, their options can overlap visually when their text is longer than just a couple/few words. Extjs 3 has a known issue with how it calculates the column widths and, because it inlines those widths on render, it can be difficult to control the problem outside of the core. Fixing that in Ext via overrides is more trouble than it's worth, so I elected to unset the hard-coded widths and provide some general controls to improve the layout of these TVs. 

Before this fix:
Before 1

After this fix (just a couple sample configs):
After 1
After 2
After 3

### How to test
1. Create a couple multi-column TVs (a radio and a checkbox) and play with the new properties, giving each TV varying lengths of text in their options.
2. Verify that there's no overlap between columns.

### Related issue(s)/PR(s)
I believe I saw something in the Community thread on Slack that alerted me to the problem; there doesn't seem to be an Issue posted on GitHub though.

**Update:** Resolves #16801 - The issue was just posted by a user today (12/19).


theboxer (approved):
Works as described, thank you!

PR #16786, opened 2025-11-19
Fix image rotation on upload

### LINKED ISSUE #16785: JPG is rotated back to original
### Summary
When a JPG/JPEG had been edited by user and not stored with correct metadata the image is rotated back to before the edit after uploading.

A fix has been done before to apply it to the preview. This fix will make that redundant. 
https://github.com/modxcms/revolution/pull/15484

### Step to reproduce
Rotate and image and save it, upload to website.
 (Doesn't always apply, depends on the tools used to edit)

### Observed behavior
The edited image which was changed from landscape to portrait is landscape again.

### Expected behavior
The images keep its edited state.

### Environment
All previous Modx versions.

MrRoco:
### What does it do?
Added an Exif check for edited images and if the have been Orientated.
If Orientated, make sure the image is also saved with the correct data.

### Why is it needed?
Certain software doesn't save the edit data correctly which will cause rotated imaged to be reverted back after upload.

### How to test
Rotate an image, upload it to the site, check if the image is still in correct position.

### Related issue(s)/PR(s)
Resolves: #16785 and also resolves https://github.com/modxcms/revolution/pull/15484

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -1182,6 +1182,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
                 $this->xpdo->log(modX::LOG_LEVEL_ERROR, $e->getMessage());
             }
 
+            // Check if Orientation is correct due to EXIF issues in JPG/JPEG and rotate if needed.
             try {
                 $image_extensions = ['jpg', 'jpeg'];
                 $file_extension = strtolower(pathinfo($file['name'], PATHINFO_EXTENSION));
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -32,6 +32,7 @@ use PDO;
 use xPDO\Cache\xPDOCacheManager;
 use xPDO\Om\xPDOCriteria;
 use xPDO\xPDO;
+use MODX\Revolution\modPhpThumb;
 
 /**
  * An abstract base class extend to implement loading your League\Flysystem\AbstractAdapter
@@ -1182,7 +1183,27 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             }
 
             try {
-                $this->filesystem->write($newPath, file_get_contents($file['tmp_name']));
+                $image_extensions = ['jpg', 'jpeg'];
+                $file_extension = strtolower(pathinfo($file['name'], PATHINFO_EXTENSION));
+
+                if (in_array($file_extension, $image_extensions)) {
+                    $exif = @exif_read_data($file['tmp_name']);
+                    if (!empty($exif['Orientation']) && $exif['Orientation'] > 1) {
+                        $phpThumb = new modPhpThumb($this->xpdo, []);
+                        $phpThumb->setSourceFilename($file['tmp_name']);
+                        $phpThumb->setParameter('ar', 'x');
+                        if ($phpThumb->GenerateThumbnail()) {
+                            $image_string = $phpThumb->OutputThumbnailData();
+                            $this->filesystem->write($newPath, $image_string);
+                        } else {
+                            $this->filesystem->write($newPath, file_get_contents($file['tmp_name']));
+                        }
+                    } else {
+                        $this->filesystem->write($newPath, file_get_contents($file['tmp_name']));
+                    }
+                } else {
+                    $this->filesystem->write($newPath, file_get_contents($file['tmp_name']));
+                }
             } catch (FilesystemException | UnableToWriteFile $e) {
                 $this->addError('path', $this->xpdo->lexicon('file_err_upload'));
                 $this->xpdo->log(modX::LOG_LEVEL_ERROR, $e->getMessage());


SnowCreative:
Hey, really nice! About once a year I get a client asking me why the photo they just uploaded is rotated on their website.

smg6511:
@MrRoco - Can you specify at least one software package that does not correctly save the data? That'll help in testing the PR.

MrRoco:
> @MrRoco - Can you specify at least one software package that does not correctly save the data? That'll help in testing the PR.

Hello, it's really random, most of the times its from clients who use windows.

matdave:
> @MrRoco - Can you specify at least one software package that does not correctly save the data? That'll help in testing the PR.

It used to happen on phones all the time when people uploaded directly. It's why it's a built in helper on my MODX Buddy package (along with resizing).

jaygilmore:
@MrRoco It possibly would have been preferrable to do this with Imagick and a fallback to GD and not add another dependent call to phpThumb but that's not a huge deal. It's doubtful we'll ever remove it and I dont' think it's commands will ever change. 

@smg6511 here are some images I created with the altered exif data using exiftools on the commandline. 
[exif-tests.zip](https://github.com/user-attachments/files/24535230/exif-tests.zip)

All but the original have the exif data altered but not the pixel orientation. The original has no orientation tag. These should help if you wanted to test.

smg6511:
@jaygilmore Thanks for the files. The trouble with those is they do not have the `Orientation` exif tag, which is what this solution is reliant on. What I'm unclear on is exactly what exif data needs to be incorrect to create this issue.

I've noticed that when you edit the orientation of an image it's `Orientation` flag seems to always get set to 1. Same with if you save an image as anything other than its raw form (e.g., for iPhone from HEIC to jpg) ... sets it to 1. Makes sense, it's reset to the default indicating it's the intended viewing orientation. 

So, is the referenced software glitch that it _doesn't_ reset Orientation to 1 when editing/exporting? And is `Orientation` only meant to record the _original_ way the device (be it a smartphone or digital SLR) was oriented upon taking a given picture?

jaygilmore:
@smg6511 it appears that I did it incorrectly when I set the Orientation flag. Here's a new set of files. 

It's my understanding that if you set the rotation and the pixels are then rotated, the new position is 1. So, if you're using anything that would set the orientation and it saves the oriented pixels on output it will set that Orientation flag to 1 because it's now in a new state. 

I verified as follows: 

```
exiftool -n -Orientation compass-rose-orient*.jpeg
======== compass-rose-orient-1-normal.jpeg
Orientation                     : 1
======== compass-rose-orient-3-rotate-180.jpeg
Orientation                     : 3
======== compass-rose-orient-6-rotate-90.jpeg
Orientation                     : 6
======== compass-rose-orient-8-rotate-270.jpeg
Orientation                     : 8
    4 image files read
```
[test-orientations.zip](https://github.com/user-attachments/files/24568165/test-orientations.zip)

smg6511:
@jaygilmore - Thanks for the updated files. With these, there is no difference in the output (i.e., the resulting image after upload). It could be because the images are square (? ... but probably not) or that there's other information that triggers the issue.

@MrRoco - Did you test this PR against an image(s) with bad exif info and, if so, can you share it/them here?

rthrash:
On an older (3.1.2) version of Revo, the test images @jaygilmore supplied appear to have the correct orientation based on how they're named:

Screenshot 2026-02-03 at 2 51 58 PM

It's not clear how to test this PR. Anyone have an image with bad EXIF data handy?

opengeek:
I do not think MODX should be introducing a dependency on phpthumb during the upload of an image to a media source. Not all software that works with images respect the Orientation metadata and most image exports to the JPEG format will output an image based on the Orientation metadata and then reset it to 1. If there is a way to handle this without adding a hard dependency on phpthumb inside the media source, that would be a preferrable solution. But I do not know if that is possible without further research.

Ibochkarev (commented):
## Architecture concerns (align with @opengeek)

**1. phpThumb dependency in media source**

Introducing a hard dependency on `modPhpThumb` during upload couples the media source layer to a specific image library. Considerations:

- Not all MODX installations may have phpThumb configured or available
- Media sources are meant to handle storage; image processing is a separate concern
- As @jaygilmore noted, Imagick with GD fallback would be more standard and avoid this coupling

**Suggestion:** Implement EXIF correction using GD (`imagecreatefromjpeg` + `imagerotate` based on Orientation) or Imagick, which are core PHP extensions. This keeps the media source independent of phpThumb.

**2. Alternative implementation sketch**

```php
// GD-based approach (no phpThumb dependency)
if (extension_loaded('gd') && in_array($file_extension, ['jpg', 'jpeg'])) {
    $exif = @exif_read_data($file['tmp_name']);
    if (is_array($exif) && !empty($exif['Orientation']) && $exif['Orientation'] > 1) {
        $image = imagecreatefromjpeg($file['tmp_name']);
        if ($image) {
            $rotated = $this->rotateByExifOrientation($image, $exif['Orientation']);
            if ($rotated) {
                ob_start();
                imagejpeg($rotated, null, 95);
                $this->filesystem->write($newPath, ob_get_clean());
                imagedestroy($rotated);
                imagedestroy($image);
                continue; // skip default write
            }
            imagedestroy($image);
        }
    }
}
```

---

## Potential bugs

**1. `$exif` may not be an array**

`exif_read_data()` returns `false` on failure (e.g. no EXIF, corrupted file, or missing `exif` extension). Accessing `$exif['Orientation']` when `$exif` is `false` causes:

- PHP 8+: `TypeError` (cannot use array offset on bool)
- PHP 7.x: deprecation notice

**Fix:** Add a type check:

```php
$exif = @exif_read_data($file['tmp_name']);
if (is_array($exif) && !empty($exif['Orientation']) && $exif['Orientation'] > 1) {
```

**2. Memory usage for large images**

`$phpThumb->OutputThumbnailData()` loads the full image into memory. For large uploads (e.g. 20+ MP), this can cause memory issues. Consider adding a max dimension/size check before processing, or documenting that very large images may behave differently.

---

## Minor suggestions

**1. EXIF extension check**

If the `exif` extension is not loaded, `exif_read_data()` is undefined and will cause a fatal error. Consider:

```php
if (extension_loaded('exif') && in_array($file_extension, $image_extensions)) {
```

**2. Comment placement**

The doc comment is inside the `try` block; it would be clearer as a method-level or block-level docblock above the `try`.

**3. Error logging**

When `GenerateThumbnail()` fails and the fallback is used, the original file is written. If the intent is to correct orientation, consider logging a warning so admins can debug why correction failed.

---

## Testing

As noted in the discussion, test images from @jaygilmore showed no visible difference with this PR (possibly due to square images or other factors). Before merge:

- [ ] Test with a real smartphone photo (portrait/landscape) that has Orientation 3, 6, or 8
- [ ] Verify behavior when `exif` extension is disabled
- [ ] Verify behavior when phpThumb is unavailable or misconfigured
- [ ] Test with a large image (e.g. 10+ MP) to check memory usage

---

## Summary

| Aspect | Status |
|--------|--------|
| Problem addressed | Valid — common issue with phone/camera uploads |
| Approach | Correct — detect and re-encode |
| Implementation | phpThumb dependency is a concern |
| Robustness | Fix `$exif` type check; add `exif` extension check |

**Verdict:** The fix targets a real problem, but the architectural and robustness concerns should be addressed. Prefer a GD/Imagick-based solution over phpThumb in the media source layer. At minimum, add the `is_array($exif)` check and `extension_loaded('exif')` guard before merging.

PR #16789, opened 2025-11-30
Form Customization: Reveal Content Above & Below Regions

### LINKED ISSUE #16442: Please put "modx-content-above" and "modx-content-below" in the list of regions in manager form customization
### Summary
In the first versions of MODX 3 we couldn't put TVs underneath the content field. Assigning them to "modx-resource-main-left" added them below the content field. Now, we can't put TVs above the content field any more. In 2.8, assigning a TV to modx-resource-main-left put it above the content field, below the main top fields, and assigning it to modx-resource-content put it below the content field. In MODX 3, assigning TVs to modx-resource-main-left or modx-resource-content put them in the same place, below the content field, so there's no difference between these now in form customization.

### Why is it needed?
There need to be options to put TVs either above or below the content field. If above, they should show up after the main fields in that panel (pagetitle, etc.)

### Suggested solution(s)
It's weird that "modx-resource-content" is used both for the content field AND for the content panel in MODX versions below 3. That makes it impossible to distinguish between the field and the area. Since there is no real "modx-resource-content" region in 3.0, I suggest adding two regions, modx-resource-content-top and modx-resource-content-bottom for 3.0 to control where TVs can be located.

smg6511:
### What does it do?
Adds missing regions to FC set schema and renames those regions’ IDs to be more consistent with current naming convention.

### Why is it needed?
It's unclear that two additional regions exist above and below the main content field. Also, even though you can currently assign TVs to these regions, you can not control the overall visibility of those regions.

### How to test

1. Because there's a change to a schema, a full rebuild (_build/transport.core.php) and running of the setup (install) must first be done
2. Create at least one TV for testing
3. Create a Form Customization set, and note that the `modx-resource-content-above` and `modx-resource-content-below` now show in the Regions tab
4. Move a TV to both of these regions and verify they appear in the expected position in a Resource
5. Uncheck each of these regions in the FC Set Region tab and verify each region is hidden when viewing a Resource
6. Also, verify that usage of the legacy Region IDs work as well (`modx-content-above` and `modx-content-below`)

### Related issue(s)/PR(s)
Resolves #16442

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -1046,7 +1046,9 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,hideMode: 'offsets'
             ,items: [{
                 id: 'modx-resource-content-above'
-                ,border: false
+            },{
+                /** @deprecated To be removed in 3.3, reference new region id 'modx-resource-content-above' */
+                id: 'modx-content-above'
             },{
                 xtype: 'textarea'
                 ,name: 'ta'
@@ -1057,8 +1059,10 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
                 ,grow: false
                 ,value: (config.record.content || config.record.ta) || ''
             },{
+                /** @deprecated To be removed in 3.3, reference new region id 'modx-resource-content-below' */
+                id: 'modx-content-below'
+            },{
                 id: 'modx-resource-content-below'
-                ,border: false
             }]
         };
     }
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -1045,7 +1045,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,autoHeight: true
             ,hideMode: 'offsets'
             ,items: [{
-                id: 'modx-content-above'
+                id: 'modx-resource-content-above'
                 ,border: false
             },{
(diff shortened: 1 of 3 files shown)


manager/assets/modext/widgets/resource/modx.panel.resource.js
```diff
@@ -1045,7 +1045,7 @@ Ext.extend(MODx.panel.Resource,MODx.FormPanel,{
             ,autoHeight: true
             ,hideMode: 'offsets'
             ,items: [{
-                id: 'modx-content-above'
+                id: 'modx-resource-content-above'
```
theboxer, line 1048: This is a breaking change, I'm not sure if big enough, but can break some extras that are injecting to those places and using the ID for identifying the space. (same goes for the modx-content-below)

smg6511: It _shouldn't_ be: There's BC support added in `modActionDom` that transforms the old id to the new one as the rules are actually being built. Is there a scenario I'm not thinking of that would be problematic?

Although I suppose I'm thinking of this specifically in the context of Form Customization; guessing you're thinking one might have reason to directly access these two components outside of the FC process?

smg6511: @theboxer - Ok, to cover your concern I went ahead and added back in two items using the original ids with a deprecation notice.

opengeek: @theboxer — does this resolve your concerns?

theboxer: Yep, it's fine now. thank you

theboxer (approved):

PR #16793, opened 2025-12-04
feat(2.x): Filter special characters in error messages

matdave:
### What does it do?
Add filtering to messages sent to the smarty error template

### Why is it needed?
Prevent special characters from being added to error messages

### DIFF
--- a/core/model/modx/modmanagercontroller.class.php
+++ b/core/model/modx/modmanagercontroller.class.php
@@ -180,7 +180,12 @@ abstract class modManagerController {
 
         $tpl = $this->getTemplateFile();
         if ($this->isFailure) {
-            $this->setPlaceholder('_e', $this->modx->error->failure($this->failureMessage));
+            $this->setPlaceholder('_e',
+                filter_var_array(
+                    $this->modx->error->failure($this->failureMessage),
+                    FILTER_SANITIZE_FULL_SPECIAL_CHARS
+                )
+            );
             $content = $this->fetchTemplate('error.tpl');
         } else if (!empty($tpl)) {
             $content = $this->fetchTemplate($tpl);
--- a/core/model/modx/modmanagercontrollerdeprecated.class.php
+++ b/core/model/modx/modmanagercontrollerdeprecated.class.php
@@ -76,7 +76,12 @@ class modManagerControllerDeprecated extends modManagerController {
 
         /* assign later to allow for css/js registering */
         if (is_array($cbody)) {
-            $this->setPlaceholder('_e', $cbody);
+            $this->setPlaceholder('_e',
+                filter_var_array(
+                    $cbody,
+                    FILTER_SANITIZE_FULL_SPECIAL_CHARS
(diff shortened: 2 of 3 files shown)


theboxer (approved):

Ibochkarev (commented):
## What works well

**1. Security improvement**

Encoding special characters (`<`, `>`, `'`, `"`, `&`) prevents reflected XSS when error messages are rendered in HTML. This is especially important for the manager login/error pages, which may display user input or exception messages.

**2. Consistent application**

All three error output paths are covered:
- `modManagerController` — failure messages
- `modManagerControllerDeprecated` — `$cbody` array
- `modManagerResponse` — `$this->body` array

**3. Appropriate filter choice**

`FILTER_SANITIZE_FULL_SPECIAL_CHARS` encodes quotes (unlike `FILTER_SANITIZE_SPECIAL_CHARS` without `ENT_QUOTES`), which is safer for attributes and inline scripts.

---

## Suggestions and potential issues

**1. Return type of `modx->error->failure()`**

In `modManagerController`, the code passes `$this->modx->error->failure($this->failureMessage)` directly to `filter_var_array()`, which expects an **array** as the first argument. If `failure()` ever returns a string or other non-array type, this will cause:
- PHP 8+: `TypeError`
- PHP 7: unexpected behavior

Consider adding a type check or ensuring the value is always an array before filtering:

```php
$failure = $this->modx->error->failure($this->failureMessage);
$this->setPlaceholder('_e', is_array($failure)
    ? filter_var_array($failure, FILTER_SANITIZE_FULL_SPECIAL_CHARS)
    : htmlspecialchars((string) $failure, ENT_QUOTES, 'UTF-8'));
```

**2. Nested arrays not filtered**

`filter_var_array()` only sanitizes top-level values. If the error structure contains nested arrays (e.g. `['message' => '...', 'errors' => ['field' => 'msg']]`), nested string values will not be encoded. If such structures are possible, a recursive sanitization helper may be needed.

**3. Handling `filter_var_array` failure**

`filter_var_array()` returns `false` if the filter fails. In that case, `false` would be passed to `setPlaceholder` / `smarty->assign`, which could lead to odd template output. Consider a fallback:

```php
$filtered = filter_var_array($data, FILTER_SANITIZE_FULL_SPECIAL_CHARS);
$this->setPlaceholder('_e', $filtered !== false ? $filtered : $data);
```

**4. Formatting**

The multi-line `filter_var_array` call could be slightly more readable with consistent indentation, though this is subjective.

---

## Testing checklist

- [ ] Trigger a failure with special chars in message (e.g. `<>"'&`)
- [ ] Verify error page renders correctly and chars are encoded
- [ ] Test deprecated controller path
- [ ] Test `modManagerResponse` path (e.g. login failure)
- [ ] Confirm legitimate Unicode/international chars are not broken

---

**Verdict:** Good security improvement. LGTM with the suggestions above, especially verifying the return type of `failure()` and handling edge cases.

PR #16794, opened 2025-12-05
PHP 8.5 Compatibility

opengeek:
### What does it do?
Improves PHP 8.5 compatibility.

### Why is it needed?
PHP 8.5 introduces some deprecations that resulted in a large number of deprecation warnings.

### How to test
Apply, make sure to composer update, and then check for deprecation warnings concerning usages of non-canonical cast names.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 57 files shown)


smg6511 (approved):
Looks good - log is clear of notices (both cast and pimple-related ones)

PR #16796, opened 2025-12-11
Increase default minimum and generated passsword lengths

### LINKED ISSUE #16756: Increase password minimum and generated length defaults
### Summary
Increase default password length and generated password length

### Why is it needed?
The National Institute of Standards and Technology (NIST) has updated its 2025 guidelines, moving away from mandatory resets and complex character requirements towards practices like longer passwords, real-time blocklists, and Multi-Factor Authentication (MFA). 

### Suggested solution(s)
[Security Experts recommend](https://www.army.mil/article/280417/secure_our_world_cecom_recommends_strong_passwords_and_password_managers#:~:text=Length%3A%20The%20longer%20the%20password,16%20characters%20for%20strong%20passwords.) 12-16 character passwords for an optimal balance of convenience and security. 

Minimum length: 12
Generated length: 16 

### Related issue(s)/PR(s)
n/a

opengeek:
### What does it do?
Increases the default minimum and generated password lengths.

### Why is it needed?
The National Institute of Standards and Technology (NIST) has updated its 2025 guidelines, moving away from mandatory resets and complex character requirements towards practices like longer passwords, real-time blocklists, and Multi-Factor Authentication (MFA).

### How to test
Make sure the defaults are respected if you set the values to empty strings in system settings and that new installs get the new default values.

### Related issue(s)/PR(s)
Resolves #16756

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1240,7 +1240,7 @@ $settings['proxy_username']->fromArray([
 $settings['password_generated_length'] = $xpdo->newObject(modSystemSetting::class);
 $settings['password_generated_length']->fromArray([
   'key' => 'password_generated_length',
-  'value' => 10,
+  'value' => 16,
   'xtype' => 'numberfield',
   'namespace' => 'core',
   'area' => 'authentication',
@@ -1249,7 +1249,7 @@ $settings['password_generated_length']->fromArray([
 $settings['password_min_length'] = $xpdo->newObject(modSystemSetting::class);
 $settings['password_min_length']->fromArray([
   'key' => 'password_min_length',
-  'value' => 8,
+  'value' => 12,
   'xtype' => 'numberfield',
   'namespace' => 'core',
   'area' => 'authentication',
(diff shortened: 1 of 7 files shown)


smg6511:
@opengeek - Giving this a look and testing. I'm doing so on an existing install which scenario begs the question of whether we should also run a db update on the two system settings (perhaps only if their values are the old defaults of 8 and 10); to do so may or may not be a "no-no," in terms of technically altering user data. What do you think? 

That aside, maybe we add a note/warning/info blurb that appears during an _update_ install that alerts users re the new defaults and a gentle suggestion to update existing passwords that don't meet the new criteria.

Not trying to complicate things, it's just that this change would only affect new installs — and it depends on whether that's your only intent with this change.

Oh, only other thing is it'd be good to update the `user_err_password_too_short` lex to inform user of min length. That could be done in a separate fix though if you want ;-)

rthrash:
Given that it should only affect newly created (or reset) passwords, and not existing ones, I vote for updating the settings and letting them know in the readme and/or upgrade UI. Suggesting they address exisitng password lengths is a good idea, also.

opengeek:
The intention here was only to update the system setting defaults for new installations. I do not think assuming existing deployments want the new settings applied is a good idea. As I saw with the unit test failures, if someone had a custom generated length setting that no longer meets the new min length while never changing their install's min length setting, and we update that value, their passwords would fail to generate.

smg6511:
> The intention here was only to update the system setting defaults for new installations.

Ok, in that case this PR is good to go.

smg6511 (approved):

JoshuaLuckers (approved):

PR #16802, opened 2025-12-19
Bump minimum PHP requirement to 8.1

opengeek:
### What does it do?
Bumps the minimum PHP requirement to the 8.1 release.

### Why is it needed?
A security vulnerability in the AWS-SDK-PHP dependency was fixed only in releases that require PHP >= 8.1 which prevents installation of dependencies via composer without ignorning vulnerabilities. I believe the best course of action is dropping support for PHP < 8.1.

### How to test
Make sure everything still works with all of the dependency updates that will come with bumping our minimum PHP to >= 8.1.

### Related issue(s)/PR(s)
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


matdave (approved):
Hold onto your butts

PR #16806, opened 2025-12-30
Fix expired session login dialog

### LINKED ISSUE #16742: Cannot login using the "Login expired" popup form.
### Summary
When my session times out in the manager a modal login dialog pops up but if I enter my credentials it just keeps returning. I can login by refreshing the page and using the main dashboard login fields.

### Step to reproduce
Open the error log. Take a break. Come back and see the modal login popup. Enter credentials.

### Observed behavior
Shows "Please Wait ..." with a "Saving ..." progress bar for about a second then back to the login form. The progress bar is at about 10% and doesn't grow while displayed. Clicking the "Logout" button does indeed log out.

### Expected behavior
Logged in.

### Environment
MODX 3.1.2
Brave browser.
Kubuntu 24.04 Desktop
Apache (I think)
mysql Ver 15.1 Distrib 10.6.21-MariaDB
That's my current environment but the problem has been with me for years, and using any browser.

smg6511:
### What does it do?
Maps the login window action "Security/Login" to the controller one "security/login" (lowercase).

Additionally, some minor css changes were made to the status message toast, as its content was being clipped with short messages (such as that which is shown after successfully extending a session).

### Why is it needed?
Login is probably the only action where it can occur both directly via the controller (main login page) and via a window (extend session). The window, however, _must_ use the newer (3.x) casing scheme in its action to call its corresponding processor. The core connector, however, was only testing for the presence of the lowercase formatted action to make certain decisions.

### How to test

1. Run grunt build to see the style changes
2. Use one of two ways to force-expire your session: 1) change your `session_cookie_lifetime` system setting to a very low number like 60 (1 minute) and login with the remember me checkbox ticked; or 2) manually delete the session cookie from your browser's dev tools pane.
3. After normal login, go to the Reports/Error Log page. Once you know your current session has expired click the refresh button; this should bring up the login window to extend your session. Verify that clicking login from there allows you to continue your work.

### Related issue(s)/PR(s)
Resolves #16742

### Special Note
The process of fixing the underlying cause of this issue raises another, which will be separately addressed. Some requests, like when trying to save a Resource after your session has expired, result in a mangled 401 response which prevents the login window from ever showing.


opengeek (commented):
While it technically worked as advertised, once I submitted the login through the dialog, it displayed a message saying "session extended" but the error log just had a "working…" dialog and never actually refreshed until I clicked refresh again. Is this expected? It doesn't feel correct.

smg6511:
@opengeek - Yes, to me that's a separate issue to address. There are a couple things to address: The mangled response you're already aware of and, re the stuck "working" toast and lack of refresh of the log, I don't know whether the extension of a session _ever_ had the ability to re-engage an action attempted during a stale session (it does the same thing you see here in 2.x). That, in my mind, is a bigger fish to fry — honestly I'm not even sure it's possible or at least practical (am I thinking this is more complicated than it really is?) — and it'd be good to just get the basic functionality of extending the session working. I thought about being more verbose in the success toast upon extension, alerting the user to re-take whatever their previous action was. Looking forward to your insight ... ;-)

opengeek:
> @opengeek - Yes, to me that's a separate issue to address. There are a couple things to address: The mangled response you're already aware of and, re the stuck "working" toast and lack of refresh of the log, I don't know whether the extension of a session _ever_ had the ability to re-engage an action attempted during a stale session (it does the same thing you see here in 2.x). That, in my mind, is a bigger fish to fry — honestly I'm not even sure it's possible or at least practical (am I thinking this is more complicated than it really is?) — and it'd be good to just get the basic functionality of extending the session working. I thought about being more verbose in the success toast upon extension, alerting the user to re-take whatever their previous action was. Looking forward to your insight ... ;-)

No insight, I'm afraid. I was just making sure I wasn't missing expected behavior. We'll go with this for now. 😅

smg6511:
> We'll go with this for now.

Ok, do you want me to add a little to the success message for clarity, or just run with it as-is for now?

opengeek:
> > We'll go with this for now.
> 
> Ok, do you want me to add a little to the success message for clarity, or just run with it as-is for now?

As-is for now.

opengeek (approved):

PR #16807, opened 2025-12-31
Improve manager forgot login activation error messaging 

### LINKED ISSUE #13951: MODX 3: Login - Forgot your password error message if no mail service installed
### Summary
I get an uncompleted inline error message when I trigger the "Forgot your password" process

**UPDATE: It's worth noting my dev server doesn't have any ability to send outbound email.**

### Step to reproduce
Type an email address of a manager user in the "Forgot your password" form.

### Observed behavior
An error appears, saying "Error while sending mail to"

screen shot 2018-07-03 at 11 13 33

Possibly inherited from MODX2.x

### Expected behavior
I expected an error message such as "Error while sending mail to user@address.com", or "A mail service is required to send email - please contact your administrator" 

### Environment
MODX3 Alpha v 3.0.0-dev
From https://github.com/Sterc/modx3builds/raw/master/latest/modx3-alpha-regular.zip
Chrome Version 67.0.3396.99

smg6511:
### What does it do?
Created a new, more explicit lexicon for condition where a mail failure occurs when attempting to send a new password activation email.

### Why is it needed?
The original message shown appears incomplete.

### How to test

1. While logged in, clear modx cache to ensure new lexicon is picked up
2. Ensure your system settings for sending mail are set in a way that will trigger a mail server error (e.g., specify an incorrect port)
3. Logout and click "Forgot your password?"
4. Enter an email of any legitimate manager user
5. Verify a new, more clear error message shows when clicking "Send Activation Email"

### Related issue(s)/PR(s)
Resolves #13951

### DIFF
--- a/core/lexicon/en/default.inc.php
+++ b/core/lexicon/en/default.inc.php
@@ -259,6 +259,7 @@ $_lang['login'] = 'Login';
 $_lang['logout'] = 'Logout';
 $_lang['logout_confirm'] = 'Are you sure you want to logout?';
 $_lang['long_title'] = 'Long title';
+$_lang['mail_to_err_server_error'] = 'A mail server error prevented delivery to <em>[[+recipient]]</em>. Please contact your system administrator if the problem persists.';
 $_lang['male'] = 'Male';
 $_lang['manage_files'] = 'Manage Files';
 $_lang['manager'] = 'Manager';
--- a/manager/controllers/default/security/login.class.php
+++ b/manager/controllers/default/security/login.class.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
@@ -28,7 +29,8 @@ class SecurityLoginManagerController extends modManagerController
     public $loadHeader = false;
     public $loadFooter = false;
 
-    public function initialize() {
+    public function initialize()
+    {
         $this->handleLanguageChange();
         return true;
     }
@@ -36,7 +38,8 @@ class SecurityLoginManagerController extends modManagerController
      * Check for any permissions or requirements to load page
      * @return bool
      */
-    public function checkPermissions() {
+    public function checkPermissions()
+    {
         return true;
     }
 
@@ -44,14 +47,17 @@ class SecurityLoginManagerController extends modManagerController
      * Register custom CSS/JS for the page
      * @return void
      */
-    public function loadCustomCssJs() {}
+    public function loadCustomCssJs()
+    {
+    }
 
     /**
      * Custom logic code here for setting placeholders, etc
      * @param array $scriptProperties
      * @return void
      */
-    public function process(array $scriptProperties = []) {
+    public function process(array $scriptProperties = [])
+    {
         $this->handleForgotLoginHash();
         $this->handleMagicLoginLink();
         $this->preserveReturnUrl();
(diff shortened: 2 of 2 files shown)


core/lexicon/en/default.inc.php
```diff
@@ -259,6 +259,7 @@
 $_lang['logout'] = 'Logout';
 $_lang['logout_confirm'] = 'Are you sure you want to logout?';
 $_lang['long_title'] = 'Long title';
+$_lang['mail_to_err_server_error'] = 'A mail server error prevented delivery to <em>[[+recipient]]</em>. Please contact your system administrator if the problem persists.';
```
opengeek, line 262: Are we certain this should echo back the recipient? It feels more appropriate to just present a simple message like "An error occurred attempting to send an email. Please contact your system administrator if the problem persists."

opengeek (changes requested):
I'm having trouble reviewing this as I am seemingly unable to cause a mail server error even before checking out this PR to test.

core/lexicon/en/default.inc.php
smg6511: It just spits back the email that was submitted when starting the forgot password process. I can make the message more generic if you want; I tend to like being less terse with feedback messaging and this particular instance reflects a server/config error rather than user input error (and I don't think it gives away any particularly useful info to a bad actor). 

I've always found our error messaging and logging to be fairly lacking, thus my more verbose messaging. But, perhaps the best of both worlds would be to go general on the front end and add a backend log message with a little more detail.

smg6511:
> seemingly unable to cause a mail server error

I just set my `mail_smtp_hosts` sys setting to a non-valid host (in my case localhost:1020)

core/lexicon/en/default.inc.php
opengeek: Alright. The message seems fine and is properly stripping tags…

opengeek (changes requested):
Now my concern is that the error only shows if you enter a valid username/email address. If I put in an invalid username/email address, the regular success message shows. This could be a way of allowing bad actors to test for usernames/email addresses.

smg6511:
> Now my concern is that the error only shows if you enter a valid username/email address.

Yeah, and that's the way it has been; I didn't make a change toward that behavior. Here's the thing though: IMO being vague about whether someone's login attempt is being made with a valid user (name/email) probably isn't providing much security. 

I actually find it incredibly annoying as a user (in reference to other sites, particularly ones I have an account for but may not have visited for a long time) when I go through a forgot/send me my password process that responds with "if you have an account with us you'll receive an email with..." — just tell me if I _don't_ have an account by X name; it gives you the hint to look more carefully at what you submitted (was there a small typo, etc?). A decent amount of the time I'll know I entered the correct info for at least a previously active account and get no response, presumably because of problems in the site's notification system. 

Anyway, that's my soapbox on the matter—just giving my perspective and not necessarily a hard argument one way or the other ;-) In any case, how about I go ahead and make the frontend feedback a little more general and provide a log message that gives the sys admin a better hint of what's going on?

opengeek:
> Yeah, and that's the way it has been; I didn't make a change toward that behavior. Here's the thing though: IMO being vague about whether someone's login attempt is being made with a valid user (name/email) probably isn't providing much security.
>
> I actually find it incredibly annoying as a user (in reference to other sites, particularly ones I have an account for but may not have visited for a long time) when I go through a forgot/send me my password process that responds with "if you have an account with us you'll receive an email with..." — just tell me if I _don't_ have an account by X name; it gives you the hint to look more carefully at what you submitted (was there a small typo, etc?). A decent amount of the time I'll know I entered the correct info for at least a previously active account and get no response, presumably because of problems in the site's notification system.

I won't argue how much security it provides, but it is a commonly known vulnerability when a site exposes a valid username/email address from user input like this. This gives an attacker a way to narrow down attempts to crack the password for a known user.
 
> Anyway, that's my soapbox on the matter—just giving my perspective and not necessarily a hard argument one way or the other ;-) In any case, how about I go ahead and make the frontend feedback a little more general and provide a log message that gives the sys admin a better hint of what's going on?

I'll take this as-is and figure out if exposing a valid user this way can be avoided here some other way.

opengeek (approved):

Issue #13977, opened 2018-07-05
Manager customization — TVs cause right sidebar to explode


SnowCreative:
### Summary
Placing TVs in the right sidebar of the Document tab don't conform to the width of this column

### Step to reproduce
Create TV of type "Listbox".

### Observed behavior
If placed in the "modx-resource-main-right" Region, TV sits below the three widgets, but not at the right width:

![TV in right column](https://user-images.githubusercontent.com/2528374/44619404-c96e4880-a853-11e8-81f7-d566207e21f7.png)

If placed in the "modx-resource-main-right-bottom" Region, TV sits inside the third widget, but still doesn't conform to the width:

![TV in right column, inside widget](https://user-images.githubusercontent.com/2528374/44619408-e60a8080-a853-11e8-9e2c-61ae780c8093.png)

### Expected behavior
TV widths should be determined by the column they occupy.

### Environment
MODX 3.0.0 alpha

jonleverrier:
@rainbowtiger screenshots return a 404

SnowCreative:
You caught me during an edit. They work now.

muzzwood:
This *should* be able to be fixed with
`#modx-resource-main-right .x-form-field-wrap.x-form-field-trigger-wrap {max-width:100%;}` however this means the inner drop-down fields may be out of alignment. I'll edit this with an update if I figure it out.

*Update*
The above CSS works however on the initial click the dropdown is offset by the same amount the field has been narrowed. On the second click it works correctly.
I'm guessing this will need an ExtJS `doLayout()` or similar on the panel after the field is rendered but I'm not exactly sure where the code for that is. 
Any ideas? 

![screenshot from 2018-07-06 12-59-17](https://user-images.githubusercontent.com/5160368/42360740-3fcafd1a-811d-11e8-9018-8c5b60592852.png)

jonleverrier:
@digitalpenguin @rainbowtiger  Side thought  - I wonder if TV's in the `modx-resource-main-right` region, should have a white background like the "menu" panel above? 

They looked naked dropped in on a grey background.

muzzwood:
True.. Though it would need some logic to add a wrapper panel if there are elements there and hide it if not.  
It'd be great for fields to be inserted into the three panels above via manager/form customization too.

jonleverrier:
[...]

PR #16809, opened 2026-01-06
Fix TV render sizing for non-TV panels

### LINKED ISSUE #12839: Combo Boxes cut off with Manager Form Customization
Below is the problem.
image

.x-form-field-wrap {
   width: auto!important;
}

If I add this CSS it seems to fix it.  This might be overly broad of a rule though.

image

smg6511:
### What does it do?
Implement more reliable and flexible sizing specs via updated css, removing explicit widths from Ext configs. The changes allow every built-in TV type to display correctly, no matter where they are placed.

### Why is it needed?
Many TV types do not render well when moved to the various Resource regions, overflowing the region boundaries. In some cases the inner input (holding the current value) also obscured the right border because of incorrect sizing.

### How to test
1. Run `grunt build` while in the `_build/templates/default` folder
2. Create each TV type (two of each if wanting to test the render to the TV panel and to Resource regions at the same time)
3. Setup Form Customization to move each TV type to one of the Resource regions.
4. Verify width and general sizing is consistent with other built-in fields as well as among moved TVs

### Related issue(s)/PR(s)
Resolves #12839 and #13977

### Special Note
More work could be done to improve the styling overall (_e.g._, maybe more space vertically between fields, fine-tuning spacing and sizing of the descriptions/labels, etc.) and I did not focus much on the mobile view as there are problems to fix there beyond TV rendering which are out of scope of this fix.


smg6511:
@theboxer - Greetings - I see you've reviewed a couple of my UI-related PRs today. Thanks for that! It'd be great if you could check this one out as well ;-)

theboxer (approved):
Seems good to me.

Before:

Image

After:

Image

opengeek (changes requested):
@smg6511 — unfortunately, I cannot merge this PR as is. The two merge commits you added cause conflicts when I perform the rebase I have to do before the grunt build. Can you remove those merge commits and rebase [7b09ed7](https://github.com/modxcms/revolution/pull/16809/commits/7b09ed7976e6b6ec1dbd0e2ae2f3d8b0fd1e71e0) on top of 3.x? You'll have to force push.

smg6511:
@opengeek Ok, I think I got it sorted. It was odd though that the conflict was coming from one of the non-English lexicon files 🤷 

Ehhh, strike that: What the h#ll is going on with this one !? I don't get why the rebase is behaving this way. Will try again...

smg6511:
@theboxer - Hey, when you get a chance can you jump over to #16819 and give it a quick peak and approval (as you did here)? My rebasing of _this_ PR got all botched up and it was easier to just re-sumbit these changes on a clean branch. No new changes were made. Thanks!

PR #16810, opened 2026-01-06
Fix sorting logic in Security/Group/GetList.php

intersel:
According to the header comment, `$sort` and `$dir` should be used to configure the **sort by** column and the sort direction.

## What does it do?
The header file comment states:

```php
@param string $sort (optional) The column to sort by. Defaults to name.
@param string $dir  (optional) The direction of the sort. Defaults to ASC.
```
However, the actual implementation was sorting by id, without using $sort or $dir, and the sort direction was always ASC.

## Why is it needed?

This PR fixes the issue by using name as the default sort column and $dir to properly configure the sort direction.

## How to test

Open the Groups ACL and check that groups are now sorted by their names.

### Related issue(s)/PR(s) 

not found

### DIFF
--- a/core/src/Revolution/Processors/Security/Group/GetList.php
+++ b/core/src/Revolution/Processors/Security/Group/GetList.php
@@ -106,8 +106,8 @@ class GetList extends GetListProcessor
                 'OR:description:LIKE' => '%' . $query . '%',
             ]);
         }
-        $sortby = $this->getProperty('sort','name');
-        $dir = $this->getProperty('dir','ASC');
+        $sortby = $this->getProperty('sort', 'name');
+        $dir = $this->getProperty('dir', 'ASC');
         $c->sortby('parent', 'asc');
         $c->sortby($sortby, $dir);
         return $c;
--- a/core/src/Revolution/Processors/Security/Group/GetList.php
+++ b/core/src/Revolution/Processors/Security/Group/GetList.php
@@ -106,9 +106,10 @@ class GetList extends GetListProcessor
                 'OR:description:LIKE' => '%' . $query . '%',
             ]);
         }
+        $sortby = $this->getProperty('sort','name');
+        $dir = $this->getProperty('dir','ASC');
         $c->sortby('parent', 'asc');
-        $c->sortby('id', 'asc');
-        
+        $c->sortby($sortby, $dir);
         return $c;
     }
 }


opengeek (commented):
Can you provide more details on where this sorting problem exists in the manager? I cannot reproduce this under Access Control Lists in the User Groups and Users tab.

smg6511:
@opengeek & @intersel - It took a little digging, but I see the issue — a few panels have a combo that filters a grid by UGs (see Users and Dashboards for example) where the list isn't ordered in an immediately recognizable way (particularly if you've got subgroups). I'll take a closer look at this PR as well; I can say that it might need a different approach if we want to properly reflect multi-level groups in the combos.

To illustrate, from my dev install where I've got a large variety of all kinds of object (including a bunch of ACL groups and subgroups):

The tree looks good:

Screen Shot 2026-01-08 at 11 41 50 PM


But notice what happens in the combo list:

Screen Shot 2026-01-08 at 11 46 55 PM

The combos in Ext 3 don't have a built-in grouping view, so accurately reflecting the groups and sorting requires the query to return results in flat order (i.e., using the fully-expanded tree view as a reference, in top to bottom order as if there was only one level) with an additional data point used to identify which level each item is on. Then, the hierarchy is built via the combo's tpl config.

intersel:
> Can you provide more details on where this sorting problem exists in the manager? I cannot reproduce this under Access Control Lists in the User Groups and Users tab.

I use this processor in my applications, and I was expecting to use it as it is specified in its file... that's all... sorting fields sound like normal.
Now, I feel that having the possibility to do it should be useful in the manager too... and I don't see why modx processors should only be used in a manager context...? isn't modx a framework?

opengeek:
> > Can you provide more details on where this sorting problem exists in the manager? I cannot reproduce this under Access Control Lists in the User Groups and Users tab.
> 
> I use this processor in my applications, and I was expecting to use it as it is specified in its file... that's all... sorting fields sound like normal. Now, I feel that having the possibility to do it should be useful in the manager too... and I don't see why modx processors should only be used in a manager context...? isn't modx a framework?

It is. I was simply asking for clarification on how to test this PR easily since you included these instructions in the PR:
> Open the Groups ACL and check that groups are now sorted by their names.

But the User Groups under Access Control were already sorted by their names before I even applied the PR. 🤷🏼‍♂️

smg6511:
@opengeek See my [earlier comment](https://github.com/modxcms/revolution/pull/16810#issuecomment-3727012320) which points to the places where the sorting is incorrect.

intersel:
> It is. I was simply asking for clarification on how to test this PR easily since you included these instructions in the PR:
> 
> > Open the Groups ACL and check that groups are now sorted by their names.
> 
> But the User Groups under Access Control were already sorted by their names before I even applied the PR. 🤷🏼‍♂️

Oh! yes, that's right... 

Well, so thanks to  [smg6511](https://github.com/smg6511), I can tell it will fix the option list shown by [smg6511](https://github.com/smg6511) in the manager ;)

smg6511 (dismissed):
Actually, to correct the sorting of a set of single-level UGs, the two original `$c->sortby` lines should be removed. I think this would be acceptable as a basic, initial improvement. Note that the default/incoming sort props are handled in the parent class (`GetListProcessor`), so there's no need for any getting/setting of sort or dir in the child class.

But, like I said in my earlier comments, the sorting ultimately needs to be implemented in a different way to support the display of multi-level/nested UGs. If you want to tackle that take a look at how it's done for the UGs tree—there's a separate class that gets called to sort the results data (it can't be done [at least in a practical manner] in the db query because of how the data is structured).

opengeek (approved):
This corrects the processor to work like its documentation claims, so I am going to accept this as is and we can address improvements on multi-level group sorting separately.

smg6511:
@opengeek - Yeah, we can further improve the sorting in a new PR but, FWIW, as I was saying in my previous comment the actual problem is that the original couple `sortby` (on parent and id) statements in this `GetList` file were not necessary and were in fact the bug. The sortby behavior "as advertised" is established in the parent class, which is essentially duplicated in this solution.

I'm really just pointing this out so contributors might be motivated to look at other places further up the inheritance chain where missing or errant functionality may exist _before_ adding code.

PR #16813, opened 2026-01-09
Send mails with Office 365 SMTP accounts

Jako:
### What does it do?

Add an option to send mails with Office 365 SMTP accounts using OAuth2.

### Why is it needed?

It is currently not possible to connect with these SMTP accounts.

### How to test

The code is the result of the examples in the [MailTrap blog](https://mailtrap.io/blog/phpmailer-office-365/) and in the [PhpMailer docs](https://github.com/PHPMailer/PHPMailer/blob/master/examples/azure_xoauth2.phps).

It can be extended with other OAuth providers by uncommenting the other provider classes in the [GetToken processor](https://github.com/Jako/revolution/blob/phpmailer-azure/core/src/Revolution/Processors/System/Email/Oauth2/GetToken.php) and in the [settings controller](https://github.com/Jako/revolution/blob/phpmailer-azure/manager/controllers/default/system/settings.class.php) and adding other types in `mail_smtp_auth_type` and changing the OAuth provider class and other settings with the value in there.

To configure the options, a new tab is available on the system settings page, when the `mail_smtp_auth_type` system setting is not empty.

Since it uses additional composer packages, these have to be installed before the tests.

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -920,36 +920,36 @@ $settings['mail_smtp_auth_type']->fromArray([
     'area' => 'mail',
     'editedon' => null,
 ], '', true, true);
-$settings['mail_smtp_outh2_azure_client_id'] = $xpdo->newObject(modSystemSetting::class);
-$settings['mail_smtp_outh2_azure_client_id']->fromArray([
-    'key' => 'mail_smtp_outh2_azure_client_id',
+$settings['mail_smtp_oauth2_azure_client_id'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_oauth2_azure_client_id']->fromArray([
+    'key' => 'mail_smtp_oauth2_azure_client_id',
     'value' => '',
     'xtype' => 'textfield',
     'namespace' => 'core',
     'area' => 'mail',
     'editedon' => null,
 ], '', true, true);
-$settings['mail_smtp_outh2_azure_tenant_id'] = $xpdo->newObject(modSystemSetting::class);
-$settings['mail_smtp_outh2_azure_tenant_id']->fromArray([
-    'key' => 'mail_smtp_outh2_azure_tenant_id',
+$settings['mail_smtp_oauth2_azure_tenant_id'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_oauth2_azure_tenant_id']->fromArray([
+    'key' => 'mail_smtp_oauth2_azure_tenant_id',
     'value' => '',
     'xtype' => 'textfield',
     'namespace' => 'core',
     'area' => 'mail',
     'editedon' => null,
 ], '', true, true);
-$settings['mail_smtp_outh2_azure_client_secret'] = $xpdo->newObject(modSystemSetting::class);
-$settings['mail_smtp_outh2_azure_client_secret']->fromArray([
-    'key' => 'mail_smtp_outh2_azure_client_secret',
+$settings['mail_smtp_oauth2_azure_client_secret'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_oauth2_azure_client_secret']->fromArray([
+    'key' => 'mail_smtp_oauth2_azure_client_secret',
     'value' => '',
     'xtype' => 'textfield',
     'namespace' => 'core',
     'area' => 'mail',
     'editedon' => null,
 ], '', true, true);
-$settings['mail_smtp_outh2_azure_refresh_token'] = $xpdo->newObject(modSystemSetting::class);
-$settings['mail_smtp_outh2_azure_refresh_token']->fromArray([
-    'key' => 'mail_smtp_outh2_azure_refresh_token',
+$settings['mail_smtp_oauth2_azure_refresh_token'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_oauth2_azure_refresh_token']->fromArray([
+    'key' => 'mail_smtp_oauth2_azure_refresh_token',
     'value' => '',
     'xtype' => 'textfield',
     'namespace' => 'core',
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -911,6 +911,51 @@ $settings['mail_smtp_user']->fromArray([
   'area' => 'mail',
   'editedon' => null,
 ], '', true, true);
+$settings['mail_smtp_auth_type'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_auth_type']->fromArray([
+    'key' => 'mail_smtp_auth_type',
+    'value' => '',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'mail',
+    'editedon' => null,
+], '', true, true);
+$settings['mail_smtp_outh2_azure_client_id'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_outh2_azure_client_id']->fromArray([
+    'key' => 'mail_smtp_outh2_azure_client_id',
+    'value' => '',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'mail',
+    'editedon' => null,
+], '', true, true);
+$settings['mail_smtp_outh2_azure_tenant_id'] = $xpdo->newObject(modSystemSetting::class);
+$settings['mail_smtp_outh2_azure_tenant_id']->fromArray([
+    'key' => 'mail_smtp_outh2_azure_tenant_id',
+    'value' => '',
+    'xtype' => 'textfield',
+    'namespace' => 'core',
+    'area' => 'mail',
+    'editedon' => null,
+], '', true, true);
+$settings['mail_smtp_outh2_azure_client_secret'] = $xpdo->newObject(modSystemSetting::class);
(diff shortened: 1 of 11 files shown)


modxcommunity:

https://community.modx.com/t/website-email-sending-failures/7797/16

matdave:
This is a great idea. My only nitpick is the use of an unofficial dependency to integrate it. However, since it is foundationally based around league/oauth2-client, it should make it easier to integrate the other services. 

Additionally, it might be worth taking a look at the other integrations compatible with that endpoint to see if the system settings can become more "generalized" instead of specific to a single service, in case there is any advantage to using the same fields for other possibilities and having a selector for `Oath2 Provider`

Jako:
The other integrations are prepared in the [GetToken processor](https://github.com/Jako/revolution/blob/phpmailer-azure/core/src/Revolution/Processors/System/Email/Oauth2/GetToken.php) and in the [settings controller](https://github.com/Jako/revolution/blob/phpmailer-azure/manager/controllers/default/system/settings.class.php). All of them are mentioned in the [phpMailer getOAuthToken docs](https://github.com/PHPMailer/PHPMailer/blob/master/get_oauth_token.php) and in the [suggestions of the phpMailer composer json](https://github.com/PHPMailer/PHPMailer/blob/master/composer.json#L49).

The system settings can be generalized, but then we lose the ability to switch between the different services.

A selector for the different services can be added with a custom combo xtype.

Jako:
With the latest additions a mail was successful sent with an Office 365 SMTP account of a client. 

The following settings have to be set to send a mail with an Office 365 SMTP account in the modx sytem setting:

`mail_smtp_auth` -> `Yes`
`mail_smtp_auth_type` -> `AZURE`
`mail_smtp_autotls` -> `Yes`
`mail_smtp_hosts` -> `<the.mail.account.host>` (i.e. smtp-mail.outlook.com)
`mail_smtp_pass` -> `<the.mail.account.password>`
`mail_smtp_secure` -> `tls`
`mail_smtp_user` -> `<the.mail.account.user>`
`mail_use_smtp` -> `Yes`

When `mail_smtp_auth_type` is set to AZURE, the system settings page contains a new panel in the OAuth2 Token tab. In this panel you can set the SMTP Azure Client ID, the SMTP Azure Client Secret and the SMTP Azure Tenant ID. If the tab is not visible, you have to select English as manager language and/or reload the settings page.

To create these Azure settings you have to look at the Microsoft docs. Fill the settings with the according values and click on Save Settings. After that you can click on Get OAuth2 Token to be redirected to a secured page at Microsoft where the OAuth2 token is generated and sent back to the MODX installation. If it was successful, the `mail_smtp_oauth2_azure_refresh_token` system setting is filled with a value.

After that you should be able to send a mail with an Office 365 SMTP account i.e. with QuickEmail or FormIt.

Jako:
Is it possible to get further with this? Is anything missing?

opengeek:
> Is it possible to get further with this? Is anything missing?

I'm personally not a fan of having core dependencies on a Microsoft-specific implementation of a standard, as @matdave already stated. I also have zero access to any Microsoft services to test anything about this. So I will defer to reviews from folks that do have such access.

Jako:
I can also add the code for the other services I mentioned above: Google OAuth2, Yahoo OAuth2 and Microsoft OAuth2. However, I cannot test them because I don't have access to any accounts. Perhaps other users are able to do that. However, since I don't know whether the PR will ever be accepted, I won't do any work on this.

The implementation works for one of my clients. We have two large providers in Germany that are currently switching to Azure OAuth2 (HostEurope/DomainFactory — I think they were bought by GoDaddy). I think this is a useful and necessary addition.

Mark-H:
I think Office365 is big enough, especially in the business world, that it would be worth adding support for it to the core. AFAIK, Google and others already can use the standard SMTP based integrations so I wouldn't necessarily push for the other ones also. 

This _could_ also be a custom modMail implementation shipped as an extra, but as it builds upon the support in phpmailer and only requires fairly minimal changes to that, exposing that support seems reasonable to add to core to me. 

The one thing I might suggest alternatives on is the way the oauth form is presented as a tab when azure is set... that seems a little disconnected. It requires changing the value, then refreshing, and then you can use it. I'm also getting an empty alert on the settings page. 

I might suggest turning the tab into a standalone controller ?a=system/oauth that we link to from the auth type description, or something like that? 

Sadly, I also don't have Office365 to actually test the integration with.

Jako:
My client can maybe help with an Office365 account to test the implementation. But that's only one account for testing - is that enough?

I don't have a good idea where to add the oauth form. There are several options:
- Create a standalone controller - Where should it be added in the menu? Only visible when the setting is AZURE?
- Create a tab in the settings controller - Maybe the tab can be shown directly, when the setting is enabled.
- Create a popup window that uses the existing processor - Where should it be added in the menu? Only visible when the setting is AZURE?

mkschell:
@Jako can you explain why this is better as core functionality as opposed to being an Extra? I'm looking at the "if this is Azure then do this" in the core code and it feels like the kind of thing we _wouldn't_ want in the core code of a modular/extensible system. Also because of the bespoke/ad hoc nature of testing, we're going to have some code paths that can never be covered by automated tests. That said, I'm open to a convincing argument. So far Mark has said that it "seems reasonable".

Jako:
It would add that to the core, since the core also sends emails too, such as password reset emails. Although the core code is prepared for the use of a different mailer class, it is tied to modPHPMailer::class in MODX3 and in the standard extras, such as FormIt. Therefore, it cannot be changed with a plugin or bootstrap file. Making this part modular/extensible could be a task for MODX 4, since a lot of extra code would also need to be changed. Until then, this part should remain in the core as a temporary solution.

Creating an own extra for this would also mean adding an own Composer dependency for phpMailer. What would happen if the core dependency used a different version? Do you remember the Guzzle issue that was introduced with MODX 3.2.3? Something like that shouldn’t happen so often.

opengeek:
Let's not conflate things here. 

The Guzzle issue is that extras are shipping different versions of dependencies that already exist in the core. That is not the same thing. That will always be a problem without being able to install components via the composer ecosystem.

It absolutely would not require your own dependency for phpmailer. You would need to override modMail with your own custom class. It could use the PHPMailer dependency just as modPHPMailer does. This was the intended design.

Core code and extras being tied directly to the default implementation of modMail (implemented in modPHPMailer) is a bug that seems to have been copied and reapplied in various places. That seems fairly embedded and unfixable for 3.x without BC implications.

That all in mind, I suppose this is the best solution, but I still don't like including this dependency in core. I don't have a good alternative at this point.

pbowyer:
I think we could make this work as an extra in MODX 3.x without breaking backwards compatibility.

The problem is that `modMail` isn’t the contract callers actually use. MODX only asks for a mailer in one place, in `modUser::sendEmail()` at `core/src/Revolution/modUser.php:941`, but it calls `$mail->setHTML()` and reads `$mail->mailer->ErrorInfo`. FormIt does the same and also calls `$mail->mailer->addAttachment()` (`Hook/Email.php:200,220,328`). A class implementing only the declared `modMail` API would fail in both.

But. Extending `modPHPMailer` avoids that. The extra can override `_getMailer()`, call the parent method, set `AuthType` to `XOAUTH2`, then call `$this->mailer->setOAuth(...)`. Existing callers still get the PHPMailer-specific API they expect. That’s roughly the same code this PR adds to `modPHPMailer::send()`, and the extra can use MODX’s copy of PHPMailer. It only needs the Azure provider, which brings in `league/oauth2-client`.

`$modx->services` is a Pimple container, so an extra can register a lazy service from a plugin on `OnMODXInit` using early priority:

```php
$modx->services->add(
    'mail',
    fn() => $modx->mail = new Microsoft365Mailer($modx)
);
```

`getService()` checks the container first (`xPDO.php:1271`), so FormIt’s hardcoded `modPHPMailer::class` is only a fallback. The catch is that the plugin must run before anything asks for the mail service.

A `mail_class` system setting, registered lazily during `modX::initialize()`, would remove that dependency on plugin order. It would not repair the leaky `modMail` contract in 3.x, but it would provide a backwards-compatible way to work with the de facto contract until MODX 4.

There’s one backwards-compatibility detail I'd preserve until 4.0. The deprecated `getService()` assigned the service to `$modx->mail`, while container resolution doesn’t. IMO the core factory should keep that side effect. Yes, FormIt should use its local `$mail` variable rather than `$this->modx->mail` on the error path, but other extras may also assume that `$modx->mail` exists.

pbowyer:
Any feedback on the idea?

opengeek:
> Any feedback on the idea?

This definitely sounds like the right path to take on this. Your analysis of the modMail contract being broken is spot on. Further, any namespace could register the lazy service in the bootstrap, no plugin required.

That said, continuing to build out this complex dance of class definitions in settings that are essentially handling one part of container configuration while the rest is hardcoded is just extending an architectural problem that is systemic here.

But if you want to put effort into this, it would certainly be a welcome alternative.

pbowyer:
> That said, continuing to build out this complex dance of class definitions in settings that are essentially handling one part of container configuration while the rest is hardcoded is just extending an architectural problem that is systemic here.

I agree.

> But if you want to put effort into this, it would certainly be a welcome alternative.

I can't say I _want_ to 🙃, but I think it's the only approach that can be taken in v3.x while sticking to semver guarantees so I'll do it if @Jako doesn't want to.

Jako:
I have created an Extra out of this. It will be compatible with MODX 2 and 3, so I use a plugin there. @pbowyer Thanks for the idea. Its currently tested with Google and Azure/Office 365.

PR #16816, opened 2026-01-21
Resolve various deprecation warnings

opengeek:
### What does it do?
Resolves various deprecation warnings in PHP 8.1+

### Why is it needed?
When running the unit tests in various PHP 8.x releases, a number of deprecation warnings are encountered.

### How to test
Run the phpunit test suite in the latest PHP releases. Note that there are a number of warnings still being emitted by the league/flysystem dependency.

### Related issue(s)/PR(s)
n/a

### DIFF
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -63,7 +63,7 @@ class modOutputFilter
 
                 $this->log('Processing Modifier: ' . $m_cmd . ' (parameters: ' . $m_val . ')');
 
-                $output = trim($output);
+                $output = trim((string)$output);
 
                 try {
                     switch ($m_cmd) {
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -63,7 +63,7 @@ class modOutputFilter
 
                 $this->log('Processing Modifier: ' . $m_cmd . ' (parameters: ' . $m_val . ')');
 
-                $output = trim((string)$output);
+                $output = is_string($output) ? trim($output) : $output;
 
                 try {
                     switch ($m_cmd) {
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -306,8 +306,9 @@ class modElement extends modAccessibleSimpleObject
                 }
             }
             $tag = '[[';
+            $name = (string)$this->get('name');
             $tag .= $this->getToken();
-            $tag .= strlen((string)$this->get('name')) > 0 ? $this->get('name') : md5(uniqid(rand()));
+            $tag .= strlen($name) > 0 ? $name : md5(uniqid(rand()));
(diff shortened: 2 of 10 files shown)


core/src/Revolution/modElement.php
```diff
@@ -307,7 +307,7 @@ public function getTag()
             }
             $tag = '[[';
```
smg6511, line 308: Apologies for this suggestion being broken up (the latter part of it is at the end of this block of suggestions); Github just wouldn't let me mark this particular block as a single suggestion :-/
```suggestion
            $name = (string)$this->get('name');
            $tag = '[[';
```

core/src/Revolution/Filters/modOutputFilter.php
```diff
@@ -63,7 +63,7 @@ public function filter(&$element)
 
                 $this->log('Processing Modifier: ' . $m_cmd . ' (parameters: ' . $m_val . ')');
 
-                $output = trim($output);
+                $output = trim((string)$output);
```
smg6511: IMO it'd be better to keep this flexible and not cast the incoming value. We can improve this in further Jens' PR (which looked at a while back); I had a [couple questions/suggestions] there (https://github.com/modxcms/revolution/pull/16757#discussion_r2617465856) ...
```suggestion
                $output = is_string($output) ? trim($output) : $output;
```

core/src/Revolution/modElement.php
```diff
@@ -307,7 +307,7 @@ public function getTag()
...
             $tag .= $this->getToken();
-            $tag .= strlen($this->get('name')) > 0 ? $this->get('name') : md5(uniqid(rand()));
+            $tag .= strlen((string)$this->get('name')) > 0 ? $this->get('name') : md5(uniqid(rand()));
```
smg6511: This is the remainder of the suggestion from above.
```php
$tag .= $name > 0 ? $name : md5(uniqid(rand()));
```

smg6511 (changes requested):
Jason - just a couple little suggestions, otherwise all good. 

An Aside: One thing I found though is that one test fails (not related to changes in this PR): In the `Processors/Resource/ResourceCreateTest` test, item 8 fails because, after successfully trimming the pagetitle, the untrimmed title is used as criteria to grab the test Resource; it's not found of course. After thinking about it a bit, I concluded we need to trim the `$pagetitle` used as criteria on line 101.

opengeek:
> An Aside: One thing I found though is that one test fails (not related to changes in this PR): In the `Processors/Resource/ResourceCreateTest` test, item 8 fails because, after successfully trimming the pagetitle, the untrimmed title is used as criteria to grab the test Resource; it's not found of course. After thinking about it a bit, I concluded we need to trim the `$pagetitle` used as criteria on line 101.

Let's take this elsewhere, but I'm not seeing any failed tests?

smg6511 (approved):

smg6511:
> Let's take this elsewhere, but I'm not seeing any failed tests?

Ok, I'll double-check what I was seeing and take this up in another PR if confirmed.

PR #16819, opened 2026-01-30
Fix TV render sizing for non-TV panels

### LINKED ISSUE #12839: Combo Boxes cut off with Manager Form Customization

smg6511:
### What does it do?
Implement more reliable and flexible sizing specs via updated css, removing explicit widths from Ext configs. The changes allow every built-in TV type to display correctly, no matter where they are placed.

### Why is it needed?
Many TV types do not render well when moved to the various Resource regions, overflowing the region boundaries. In some cases the inner input (holding the current value) also obscured the right border because of incorrect sizing.

### How to test
1. Run `grunt build` while in the `_build/templates/default` folder
2. Create each TV type (two of each if wanting to test the render to the TV panel and to Resource regions at the same time)
3. Setup Form Customization to move each TV type to one of the Resource regions.
4. Verify width and general sizing is consistent with other built-in fields as well as among moved TVs

### Related issue(s)/PR(s)
Resolves #12839 and #13977

### Special Note
More work could be done to improve the styling overall (_e.g._, maybe more space vertically between fields, fine-tuning spacing and sizing of the descriptions/labels, etc.) and I did not focus much on the mobile view as there are problems to fix there beyond TV rendering which are out of scope of this fix.


manager/templates/default/css/login.css
```diff
@@ -1,20 +1,3 @@
...
-* GNU General Public License as published by the Free Software Foundation, either version 2 of the
-* License, or (at your option) any later version.
-* 
-* MODX Revolution is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
-* without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
-* 
-* See the GNU General Public License for more details. You should have received a copy of the GNU
-* General Public License along with MODX Revolution. If not, see <http://www.gnu.org/licenses/>.
-* 
-*/
```
opengeek, line 17: This should not be committed here.

opengeek (changes requested):

manager/templates/default/css/login.css
smg6511: This fix is bedeving me ;-) Sorry, I missed omitting this css file ... removing ...

smg6511: I get the feeling I went about the removal the wrong way, via:
```
git rm --cached manager/templates/default/css/login.css
git commit --amend
git push --force
```
(?)

opengeek: Yeah—you removed the entire file from the repository with that. 😅

opengeek: I'll get this one resolved.

smg6511: Well, sh*t. 

I see you'll resolve it, but, what's the correct way to do that (in this case when it's the most recent or only commit)?

opengeek: I'll let you know when I figure that out. 😂

opengeek:
```
git restore --source=3.x manager/templates/default/css/login.css
git add .
git commit --amend
git push --force
```

This is what I did to solve it once you deleted the file. I think the same would have worked to restore it to the previous state before you did the `git rm` command.

opengeek (approved):

PR #16825, opened 2026-02-19
Fix: array_diff_key() TypeError in upgrade.install.php

Ibochkarev:
# Fix: array_diff_key() TypeError in upgrade.install.php (#16825)

## What it does

- **`modAccessPolicy::get()` override** — For key `'data'`, the result is normalized to an array. xPDO’s handling of `phptype` `json` can return a string (e.g. when the value is empty or not decoded) or `null` (e.g. on invalid JSON). The override aligns the model with the `@property array $data` contract.
- **Validation** — A new `JsonArrayValidationRule` ensures the `data` field is either an array or a valid JSON string that decodes to an array; invalid values are rejected with the lexicon key `policy_err_invalid_data`. Validation is wired in the schema and in the MySQL map for `modAccessPolicy`.
- **`setup/includes/upgrade.install.php`** — Uses `$adminPolicy->get('data')` and `$policy->get('data')` directly. The defensive `is_array` / `fromJSON` logic was removed, since the model now guarantees an array. Line length and style were adjusted for phpcs.

## Why it is needed

During upgrade (e.g. MODX 3.12 → 3.20), `modAccessPolicy` `data` can be read from the database as a raw JSON string. Passing that string as the first argument to `array_diff_key()` causes a PHP 8 `TypeError`: `array_diff_key(): Argument #1 ($array) must be of type array, string given`, which breaks the upgrade when non-standard policies (e.g. from MiniShop3) exist.

The fix is applied at the source (the model) and reinforced with validation, as suggested in review.

## How to test

1. Install MODX 3.1.2 and add a component that creates custom access policies (e.g. MiniShop3).
2. Run the upgrade to MODX 3.2.0 via the setup (e.g. `setup/` or CLI).
3. Confirm the upgrade completes without the `array_diff_key()` TypeError and that policies are migrated as expected.

## Related

- Resolves [modx-pro/MiniShop3#100](https://github.com/modx-pro/MiniShop3/issues/100) (reported there; fix is in modxcms/revolution).
- Addresses review: `get('data')` should always return an array (fix at source); invalid `data` is handled with a validator on the field.

### DIFF
--- a/setup/includes/upgrade.install.php
+++ b/setup/includes/upgrade.install.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
@@ -69,10 +70,10 @@ $settings_version = $modx->getObject(modSystemSetting::class, [
 ]);
 if ($settings_version == null) {
     $settings_version = $modx->newObject(modSystemSetting::class);
-    $settings_version->set('key','settings_version');
-    $settings_version->set('xtype','textfield');
-    $settings_version->set('namespace','core');
-    $settings_version->set('area','system');
+    $settings_version->set('key', 'settings_version');
+    $settings_version->set('xtype', 'textfield');
+    $settings_version->set('namespace', 'core');
+    $settings_version->set('area', 'system');
 }
 $settings_version->set('value', $currentVersion['full_version']);
 $settings_version->save();
@@ -83,10 +84,10 @@ $settings_distro = $modx->getObject(modSystemSetting::class, [
 ]);
 if ($settings_distro == null) {
     $settings_distro = $modx->newObject(modSystemSetting::class);
-    $settings_distro->set('key','settings_distro');
-    $settings_distro->set('xtype','textfield');
-    $settings_distro->set('namespace','core');
-    $settings_distro->set('area','system');
+    $settings_distro->set('key', 'settings_distro');
+    $settings_distro->set('xtype', 'textfield');
+    $settings_distro->set('namespace', 'core');
+    $settings_distro->set('area', 'system');
 }
 $settings_distro->set('value', trim($currentVersion['distro'], '@'));
 $settings_distro->save();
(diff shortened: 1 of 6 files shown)


setup/includes/upgrade.install.php
```diff
@@ -174,7 +174,11 @@
     /* get admin policy and list of standard policies */
     $standards = ['Administrator','Resource','Load Only','Load, List and View','Object','Element'];
     $adminPolicy = $modx->getObject(modAccessPolicy::class, ['name' => 'Administrator']);
-    $adminPolicyData = $adminPolicy ? $adminPolicy->get('data') : [];
+    $adminPolicyData = [];
+    if ($adminPolicy) {
+        $raw = $adminPolicy->get('data');
+        $adminPolicyData = is_array($raw) ? $raw : ($modx->fromJSON($raw, true) ?: []);
+    }
```
opengeek: This should NEVER be necessary. The get() method should always return an array. Something else is wrong here.

Ibochkarev: Fixed at the source. `modAccessPolicy` now overrides `get()` so that `get('data')` always returns an array: xPDO’s handling of `phptype` `json` can return a string (e.g. when the value is empty or not decoded) or `null` (e.g. on invalid JSON), so the model was not guaranteeing the `@property array $data` contract. The override normalizes the result for key `'data'` to `[]` when it’s not already an array. The upgrade script no longer needs the defensive `is_array` / `fromJSON` logic—it just uses `$adminPolicy->get('data')` and `$policy->get('data')`.

opengeek: I see the issue, but it is in xPDO and should be addressed there.

Ibochkarev: Ok, then we are waiting for your decision in xPDO?

opengeek: Also, the default value for data is `{}`, so if this is getting set to null or another value that doesn't result in an array, that seems like a bug in the extra that is setting this invalid value. This could be solved with a validator on the field.

Ibochkarev:
The problem was in Minishop 3 - I made a fix in the add-on

PR #16826, opened 2026-02-19
Fix: preserve file extension when upload_translit is enabled

### LINKED ISSUE #16787: File upload transliteration can break filename using FURL rules
### Summary
Enabling upload_translit and setting friendly_alias_max_length can break filenames upon upload

### Step to reproduce
In system settings, enable upload_translit and set friendly_alias_max_length. Upload for example jpg file with filename length exceeding friendly_alias_max_length chars.

### Observed behavior
Uploaded filename including extension is after transliteration trimmed to friendly_alias_max_length chars, thus losing its extension. While in some configurations file type is derived from file content itself and not from file extension, a lot of functionality is broken. For example, trying to delete such file leads to error saying that '' is not allowed file type. Also third party components are affected, like MoreGallery.
This results from modMediaSource calling xpdo->filterPathSegment() on uploaded filename.

### Expected behavior
In the first place, FURL setting like friendly_alias_max_length should not be used for uploaded file names (there can be a separate setting). But the main point is that transliterating and trimming filenames should keep extensions intact, affecting only filename itself.

Ibochkarev:
### What does it do?
With `upload_translit` on, we run FURL filtering on the basename only and put the original extension back on. Alias rules no longer chew through `.jpg`.

`applyUploadTranslitToFile()` does the work: `pathinfo()` split, `friendly_alias_max_length => 0` (upload names ignore the FURL length cap), `filterPathSegment()` on the basename, then rebuild the name and fire `OnFileManagerFileRename` if it changed. Dotfiles like `.htaccess` and empty filter results keep the original name.

`modMediaSourceTest` covers the long-name + max_length case, alphanumeric restrict (`photo.jpg`), compound names, lowercase basename, and names with no extension.

### Why is it needed?
`upload_translit` + `friendly_alias_max_length` used to truncate the whole filename. `verylongfilename.jpg` became `verylongfilenam`, lost the extension, and broke delete / MoreGallery. With `friendly_alias_restrict_chars` set to `alpha` or `alphanumeric`, the same full-name path turned `photo.jpg` into `photo-jpg`.

Fixes #16787.

### How to test
1. Turn on `upload_translit`.
2. Set `friendly_alias_max_length` to something small (15).
3. Upload `verylongfilename.jpg`.
4. Check the stored name is still `verylongfilename.jpg`.
5. Set `friendly_alias_restrict_chars` to `alphanumeric`, upload `photo.jpg`, check it stays `photo.jpg`.
6. Delete the file and confirm that still works.
7. Run `modMediaSourceTest` if you want the automated check.

### Related issue(s)/PR(s)
- #16787

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -1148,27 +1148,7 @@ abstract class modMediaSource extends modAccessibleSimpleObject implements modMe
             }
 
             if ((bool)$this->xpdo->getOption('upload_translit')) {
-                $uploadRegex = $this->xpdo->getOption('upload_translit_restrict_chars_pattern');
-                $options = !empty($uploadRegex)
-                    ? [
-                        'friendly_alias_restrict_chars' => 'pattern',
-                        'friendly_alias_restrict_chars_pattern' => $uploadRegex
-                    ]
-                    : []
-                    ;
-                $newName = $this->xpdo->filterPathSegment($file['name'], $options);
-
-                // If the file name is different after filtering, call OnFileManagerFileRename
-                // so the change can be tracked by plugins
-                if ($newName !== $file['name']) {
-                    $path = $container . $this->sanitizePath($newName);
-                    $file['name'] = $newName;
-
-                    $this->xpdo->invokeEvent('OnFileManagerFileRename', [
-                        'path' => $path,
-                        'source' => &$this,
-                    ]);
-                }
+                $this->applyUploadTranslitToFile($file, $container);
             }
 
             $newPath = $container . $this->sanitizePath($file['name']);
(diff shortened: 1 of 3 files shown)


core/src/Revolution/Sources/modMediaSource.php
```diff
@@ -1156,7 +1156,12 @@ public function uploadObjectsToContainer($container, array $objects = [])
                     ]
                     : []
                     ;
-                $newName = $this->xpdo->filterPathSegment($file['name'], $options);
+                $options['friendly_alias_max_length'] = 0;
+                $pathInfo = pathinfo($file['name']);
+                $basename = $pathInfo['filename'] ?? $pathInfo['basename'];
+                $extension = isset($pathInfo['extension']) ? '.' . $pathInfo['extension'] : '';
+                $filteredBasename = $this->xpdo->filterPathSegment($basename, $options);
```
opengeek: This method is intended for filtering URI path segments, not filenames. I'm confused why this is even being used here.

Ibochkarev: You’re right that `filterPathSegment` is meant for URI path segments, not filenames, and using it here was misleading.

I’ve addressed it by:

1. Adding `modX::filterFilename($basename, array $options = [])`
This method is intended specifically for filename basenames (no extension). It still uses the same transliteration and character-restriction logic (the lexicon says uploads are “transliterated according to the global transliteration rules”), but it’s now behind a filename-oriented API and applies upload-specific options (e.g. `upload_translit_restrict_chars_pattern, friendly_alias_max_length => 0`).

2. Replacing the call in `modMediaSource`
The upload path now uses `$this->xpdo->filterFilename($basename, $options) instead of filterPathSegment`, and the translit logic is in a dedicated `applyUploadTranslitToFile()` method.

So the behaviour (same global transliteration + upload pattern) is unchanged, but the intent is clear: we use a filename filter for uploads, not a URI path segment filter.

milanidavide:
I ran into this issue too on a project with a custom transliteration service. It intentionally converts dots to hyphens for resource aliases, but because `upload_translit` passes the complete filename through the same process, uploading `hello.jpg` produced `hello-jpg`, leaving the file without an extension.

My first thought was essentially the same as the core fix in this PR: separate the final extension, filter only the filename stem, and then reattach the original extension.

I saw @opengeek's comment questioning why `filterPathSegment()` is used for uploads. Looking through the history, uploads have used this method since `upload_translit` was introduced in #15457, so this PR is not introducing that dependency. BUT it does raise the question of whether the existing approach should be reconsidered.

I see two possible approaches:
1. Keep this simple: pass only the filename stem through the existing filtering pipeline and preserve the final extension.
2. Rework upload filtering so it no longer relies on `filterPathSegment()`. That method is documented for URL path segments and does **considerably more** than transliteration, so this would require defining dedicated filename-filtering behavior and deciding which existing settings should still apply to uploads.

Personally, I think the fix in this PR makes sense. The extension needs to be kept out of the filtering process either way. Any larger discussion about replacing `filterPathSegment()` for uploads could happen separately, without holding up this fix.

Could the maintainers clarify which scope they expect for this PR?

PR #16829, opened 2026-02-19
FIx: Display of readme.txt in Package Manager seriously crippled

### LINKED ISSUE #16206: Display of readme.txt in Package Manager seriously crippled.
### Summary
When installing a package in Package Manager, the rendering of Markdown content is seriously strange when viewing the Readme tab.
 
 - There's no apparent way to separate paragraphs. All newlines are reduced to a single newline so there is no space between paragraps (and as far as I can tell, there's no way to separate them).
 - Backticks are either ignored and removed, or rendered literally.
 - Text surrounded by single backticks is rendered as regular text, rather than highlighted monospace.
 - As far as I can tell, all HTML tags outside of code sections are rendered literally, as are entities. 
 - Opening HTML beginning tag tokens are rendered as literal entities (&lt ;) inside of backticks.
 - Code sections between three backticks are rendered in the regular text font with no highlighting.

### Step to reproduce
It's difficult to illustrate the problem here, but try this code in a readme.txt for a package (without the triple backticks at each end), then look at the Readme tab when installing it. Add a code section between sets of three backticks somewhere below.

```
<p>line1</p>


<p>Line2</p>



Line 3.

Put this tag (`<br/>`) in the `<head>` section of your template.

&nbsp;&nbsp;
<p>&nbsp;</p>
<br />
<br>
Some text.



xxx
```

### Observed behavior
Output is nothing like what you would expect.

### Expected behavior
Should be rendered like normal Markdown with HTML tags outside of code sections interpreted as HTML tags.

### Environment
MODX 3.0.1 makes more of a mess of things.
MODX 2.8.4 also renders backticks literally, but does not convert opening HTML tag tokens as entities. It preserves space between paragraphs, but it also renders text between single backticks the same as regular text (and displays the backticks). It also does not display code between three backticks as monospace and without any highlighting.

Ibochkarev:
### What does it do?
- **Backend:** Removes `htmlspecialchars()` before Parsedown in `GetAttribute` for readme/changelog/license. Markdown is parsed from raw text; empty attributes are handled with `$data ?? ''`.
- **Frontend:** For the readme, changelog, and license tabs in Package Manager, the template uses raw substitution (`{[values.id]}`) so HTML from Parsedown is rendered without escaping.
- **CSS:** Adds Markdown styles under `.meta-wrapper`: `pre`, `code`, `pre code`, and `p` (code blocks, inline code, paragraph spacing). Values use `rem` units.

### Why is it needed?
Readme content during package installation was broken: paragraphs ran together, backticks and code blocks were not styled, and HTML was shown as literal text. This was caused by calling `htmlspecialchars()` before Parsedown and by escaping the resulting HTML in the template.

### How to test
1. Use a package whose readme.txt has paragraphs, `` `inline code` ``, a fenced code block, and HTML (e.g. `<br/>`).
2. Go to Extras → Install, choose the package, and open the Readme tab.
3. Confirm: paragraphs are spaced, inline and block code are visually distinct, and HTML is rendered instead of shown as text.

### Related issue(s)/PR(s)
Resolves #16206

### DIFF
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -2645,6 +2645,30 @@ iframe[classname="x-hidden"] {
   h6 {
     font-size: 0.95em;
   }
+
+  pre {
+    background: #f5f5f5;
+    padding: 0.625rem;
+    overflow-x: auto;
+    border-radius: 0.25rem;
+    margin: 1rem 0;
+  }
+
+  code {
+    font-family: monospace;
+    background: #f5f5f5;
+    padding: 0.125rem 0.375rem;
+    border-radius: 0.1875rem;
+  }
+
+  pre code {
+    padding: 0;
+    background: none;
+  }
+
+  p {
+    margin: 0 0 1rem;
+  }
 }
 
 .window-no-padding {
--- a/_build/templates/default/sass/index.scss
+++ b/_build/templates/default/sass/index.scss
@@ -2645,30 +2645,6 @@ iframe[classname="x-hidden"] {
   h6 {
     font-size: 0.95em;
   }
-
-  pre {
-    background: #f5f5f5;
-    padding: 0.625rem;
-    overflow-x: auto;
-    border-radius: 0.25rem;
-    margin: 1rem 0;
-  }
-
-  code {
-    font-family: monospace;
-    background: #f5f5f5;
-    padding: 0.125rem 0.375rem;
-    border-radius: 0.1875rem;
-  }
-
-  pre code {
-    padding: 0;
-    background: none;
-  }
-
-  p {
-    margin: 0 0 1rem;
-  }
 }
 
 .window-no-padding {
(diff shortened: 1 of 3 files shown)


smg6511 (changes requested):
Hi @Ibochkarev - Just a couple more adjustments and we're good:

1. I took some of what you did in the SCSS and incorporated it into a bunch of style changes for this packages area. It was much easier to just create a new PR (#16952) for that than ask you to incorporate it here. (Please take a look at it when you get a chance.) Because of that, you can drop your changed `index.scss` file here.
2. From what I can tell from testing, what you did in the package.panels.js isn't necessary; it didn't have any effect one way or another for me. If you agree that's the case, then you can also drop changes to that file, leaving only the changes to `GetAttribute.php` (which is the primary fix for sure).

smg6511:
@Ibochkarev - Hi Ivan, just a bump to take a look at the 2 suggestions above ;-)

Ibochkarev:
> @Ibochkarev - Hi Ivan, just a bump to take a look at the 2 suggestions above ;-)

Done

smg6511 (approved):
Good to go. Further stylistic changes pending in supplemental PR.

opengeek (approved):

Issue #12983, opened 2016-04-26
Modx 2.5 "anonymous_sessions" = 0 prevents logging in to the Manager


ilyatut:
### Summary

After updating from my previous Modx 2.4.4 to Modx 2.5, setting the "anonymous_sessions" option to "0" and logging out, I was unable to log in back to the Manager. Newly created users were unable to log in to the Manager too.
### Step to reproduce

Go to the Manager, System settings, Sessions and cookies.
Set "anonymous_sessions" = No.
Log out and try to log in again.
### Observed behavior

In my case I was unable to log in again. No error message, nothing. The login form just redirected me back to the login form page with all the fields empty.
So, I had to go to my phpMyAdmin panel, manually reset "anonymous_sessions" to 1, clear all the cache files and only then I was allowed to enter the Manager using my login/password.
### Expected behavior

It should have allowed me to log in regardless of the "anonymous_sessions" value.
### Environment

The latest MODX 2.5-pl on a shared hosting where have never occured such problems with the Manager before. Tested in the latest Chrome browser.

sottwell:
Just upgraded an installation on SkyToaster, turned off the anonymous_session to test it, no problems, both with my existing user and a new testing user.

jaygilmore:
@ilyatut thanks for the report. Something that popped into my mind is the user permissions. What are the access permissions set for the users created. I don't know if new Admin users are created with the Admin /Super User with Group/Role still or not, or if this has any bearing at all.

sottwell:
I specifically gave the new user Administrator with SuperUser role. Not sudo.

jaygilmore:
@sottwell I am curious about @ilyatut's users too. But that definitely helps clarify for sure. 

@ilyatut is there any info in the error_log or on your server logs related to errors such as 40X or similar?

sottwell:
I even created another user, using the supplied Editor policy, and that also works fine.

ilyatut:
All my users were created with the Admin /Super User access and sudo. No restrictions.
That's strange, but there are no errors in my core/cache/logs/ folder.

ilyatut:
[...]

PR #16836, opened 2026-02-21
fix(core): Manager login with anonymous_sessions = No

### LINKED ISSUE #16424: MODX 3: Manager login not possible with anonymous_sessions = No
### Summary
When the system setting `anonymous_sessions` is set to `No`, it's no longer possible to log in to the MODX manager (when the cookies are cleared beforehand).

### Step to reproduce
In MODX 3, set the system setting `anonymous_sessions` to `No`. Clear the browser cookies. Then try to log in to the manager.

### Observed behavior
Sending the login form, just reloads the page.

In MODX 2.x this is not the case, but I couldn't figure out what the difference between the versions is.
When you log in, the "security/login" processor gets executed and a redirect happens. In MODX 2.x `$_COOKIE` has content after the redirect, in MODX 3 `$_COOKIE` is still empty.

https://github.com/modxcms/revolution/blob/69a7d6d14f84151ced04a552678657456c05b9cc/core/src/Revolution/modX.php#L2793

### Environment
MODX 3.0.3

Ibochkarev:
### What does it do?

Keeps PHP sessions available for the manager when `anonymous_sessions` is turned off at the system level.

Instead of special-casing `mgr` inside `modX::_initSession()`, this adds an `anonymous_sessions=Yes` context setting on `mgr` (install data + 3.3.0 upgrade). Context settings override system settings, so manager login can start a session before any cookie exists. Public contexts can still set `anonymous_sessions=No` on their own.

The English setting description now says to disable the option on public contexts, not system-wide, and that `mgr` needs sessions for login.

### Why is it needed?

With system `anonymous_sessions=No` and cleared cookies, manager login reloads the form and never authenticates. Session bootstrap only starts a session when anonymous sessions are allowed or a session cookie is already present, so first login never gets a cookie.

Discussion on #16424 / #12983 (and community threads) treated this as a context setting for public frontends. Turning it off in System Settings accidentally disables sessions for `mgr` too. This change matches that model: keep manager sessions on, leave public contexts free to opt out.

### How to test

1. Upgrade or apply the new `mgr` context setting `anonymous_sessions=Yes` (or run setup upgrade to 3.3.0-dev / 3.3.0-pl).
2. Set system `anonymous_sessions` to No.
3. Clear browser cookies.
4. Open the manager login page, submit valid credentials.
5. Confirm you land in the manager dashboard.
6. On the `web` context, set `anonymous_sessions` to No and confirm anonymous frontend requests still skip starting a session when no cookie is present.

### Related issue(s)/PR(s)

Resolves #16424

Addresses feedback from @Jako, @mkschell, and @smg6511 on keeping the per-context intent instead of hardcoding `mgr` in core session bootstrap.

### DIFF
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -98,7 +98,7 @@ $_lang['setting_allow_tv_eval'] = 'Enable eval in TV bindings';
 $_lang['setting_allow_tv_eval_desc'] = 'Select this option to enable or disable eval in TV bindings. If this option is set to no, the code/value will just be handled as regular text.';
 
 $_lang['setting_anonymous_sessions'] = 'Anonymous Sessions';
-$_lang['setting_anonymous_sessions_desc'] = 'If disabled, PHP sessions start only when a session cookie is already present (typically for authenticated users). Disable this as a Context Setting on public contexts such as web to skip sessions for anonymous visitors. Leave it enabled for the mgr context: manager login needs a session before authentication. A system-wide No is inherited by every context that does not override it. If session_enabled is No, this setting has no effect because sessions are never started.';
+$_lang['setting_anonymous_sessions_desc'] = 'When authentication or other persistence features that require a PHP Session are not needed, set this to “No.” The Manager (mgr) Context must set this to “Yes” to operate correctly. Note that in multi-Context setups, the system-wide value for this setting can be overridden at the Context level.';
 
 $_lang['setting_archive_with'] = 'Force PCLZip Archives';
 $_lang['setting_archive_with_desc'] = 'If true, will use PCLZip instead of ZipArchive as the zip extension. Turn this on if you are getting extractTo errors or are having problems with unzipping in Package Management.';
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -98,7 +98,7 @@ $_lang['setting_allow_tv_eval'] = 'Enable eval in TV bindings';
 $_lang['setting_allow_tv_eval_desc'] = 'Select this option to enable or disable eval in TV bindings. If this option is set to no, the code/value will just be handled as regular text.';
 
 $_lang['setting_anonymous_sessions'] = 'Anonymous Sessions';
-$_lang['setting_anonymous_sessions_desc'] = 'If disabled, only authenticated users will have access to a PHP session. This can reduce overhead for anonymous users and the load they impose on a MODX site if they do not need access to a unique session. If session_enabled is false, this setting has no effect as sessions would never be available.';
+$_lang['setting_anonymous_sessions_desc'] = 'If disabled, PHP sessions start only when a session cookie is already present (typically for authenticated users). Disable this as a Context Setting on public contexts such as web to skip sessions for anonymous visitors. Leave it enabled for the mgr context: manager login needs a session before authentication. A system-wide No is inherited by every context that does not override it. If session_enabled is No, this setting has no effect because sessions are never started.';
 
 $_lang['setting_archive_with'] = 'Force PCLZip Archives';
 $_lang['setting_archive_with_desc'] = 'If true, will use PCLZip instead of ZipArchive as the zip extension. Turn this on if you are getting extractTo errors or are having problems with unzipping in Package Management.';
--- a/_build/data/transport.core.context_settings.php
+++ b/_build/data/transport.core.context_settings.php
@@ -1,4 +1,5 @@
 <?php
+
 use MODX\Revolution\modContextSetting;
 
 $collection['0']= $xpdo->newObject(modContextSetting::class);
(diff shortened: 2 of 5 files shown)


smg6511 (approved):
@opengeek @mkschell - I'm good with this and my tests confirm the session init works correctly when anon is on and off. Do you guys see any issues with this implementation?

mkschell:
@smg6511 Overall I think this could be a fairly safe "fix" to the issue but I have a couple of concerns that I might want to explore more deeply.

1) The current intention for the setting is that it's set _per context_, not set to False in the _system settings_, and this PR would weaken the argument for that. If we normalized the usage of the setting in _system settings_ again, is this actually an improvement?

2) My spidey sense is whispering "second order effects" .. I can imagine at least one unlikely scenario where this could lead to problems. It's unlikely enough that I won't bother laying it out here ;-)

Since there was some discussion on the Issue for this one, I feel like getting some consensus over in the issue might be a better next step than going ahead and merging it here.

smg6511:
> 1. The current intention for the setting is that it's set _per context_, not set to False in the _system settings_, and this PR would weaken the argument for that. If we normalized the usage of the setting in _system settings_ again, is this actually an improvement?

Hey @mkschell - I'd argue that the way it's designed now makes the wrong assumption—that more sites than not require an active session for one reason or another. At least from my experience, that's not the case. Maybe I'm wrong, but opting in for sessions seems more intuitive anyway. Also I think the change is safe (in terms of not affecting current setups) and doesn't require current sites to change anything. Perhaps there's something else you're alluding to in item 2 above that I'm not thinking of ;-)

Anyway, if you end up being persuaded/comfortable with this change, I'd take it one more step further and make the setting false by default (no change to existing installs of course).

Ibochkarev:
@mkschell @smg6511 Hi guys! I rewrote the implementation. Please watch it again

pbowyer:
Nice work @Ibochkarev I like the reworked shape of this patch.

smg6511:
@Ibochkarev - I'll recheck this in the next day or two.

core/lexicon/en/setting.inc.php
```diff
@@ -98,7 +98,7 @@
 $_lang['setting_allow_tv_eval_desc'] = 'Select this option to enable or disable eval in TV bindings. If this option is set to no, the code/value will just be handled as regular text.';
 
 $_lang['setting_anonymous_sessions'] = 'Anonymous Sessions';
-$_lang['setting_anonymous_sessions_desc'] = 'If disabled, only authenticated users will have access to a PHP session. This can reduce overhead for anonymous users and the load they impose on a MODX site if they do not need access to a unique session. If session_enabled is false, this setting has no effect as sessions would never be available.';
+$_lang['setting_anonymous_sessions_desc'] = 'If disabled, PHP sessions start only when a session cookie is already present (typically for authenticated users). Disable this as a Context Setting on public contexts such as web to skip sessions for anonymous visitors. Leave it enabled for the mgr context: manager login needs a session before authentication. A system-wide No is inherited by every context that does not override it. If session_enabled is No, this setting has no effect because sessions are never started.';
```
smg6511: Clarifying this a bit. I'm not sure if the passage about `session_enabled` should be included, as it's not exposed and is likely only ever used for specialized purposes.
```suggestion
$_lang['setting_anonymous_sessions_desc'] = 'When authentication or other persistence features that require a PHP Session are not needed, set this to “No.” The Manager (mgr) Context must set this to “Yes” to operate correctly. Note that in multi-Context setups, the system-wide value for this setting can be overridden at the Context level.';
```

smg6511 (changes requested):
For the most part this is looking good. One problem I'm seeing in the tests is that they remove _pre-existing_ `anonymous_sessions` settings on the web and mgr Contexts. In the fixtures, I think you're going to want to persist those in the setup and replace them in the teardown.

smg6511:
@Ibochkarev, _et al._: After a recent discussion with @opengeek, it became quite clear that we've been going about this the wrong way. Honestly (and a little embarrassingly), after earnestly debating this problem/solution with Jason I realized I'd become overcommitted to this idea that `anonymous_sessions` was a root cause for the issue. It's not; it only tangentially offers a way to hack a fix to a problem that lies in a design flaw in the session handling. As such, I'm going to move to close this in the next few days unless there are strong objections to doing so.

Jako:
Can disabling this system setting be made impossible in the manager with an error message? If you have to disable it, this can be done directly in the database. Is that enough?

opengeek:
> Can disabling this system setting be made impossible in the manager with an error message? If you have to disable it, this can be done directly in the database. Is that enough?

Setting this should break nothing. You can still login when anonymous sessions are disabled. And the manager should be irrelevant to anonymous sessions since you cannot access it anonymously. The issue is tangental and will be addressed.

Jako:
Then prepare a context setting for the mgr context (that maybe even cannot be changed). Since context settings normally supersede system settings, this should be enough.

opengeek:
> Then prepare a context setting for the mgr context (that maybe even cannot be changed). Since context settings normally supersede system settings, this should be enough.

That IS NOT the problem. The login SHOULD work no matter if this setting is set in ANY context. As @smg6511 described, it is a session handling issue that needs to be addressed.

Jako:
Ok, then the session cookie has to be created in the login processor to bypass the test in _initSession.

PR #16839, opened 2026-02-21
feat(manager): add key bindings for Error Log Refresh and Clear buttons

### LINKED ISSUE #16706: [Feature request] Manager error log report - add key bindings to the Refresh and Clear buttons
It would be nice to have key bindings on the Refresh and Clear buttons while viewing the Error Log in the Manager.
I and maybe many dev-ops use the error log frequently and sprinkle code under development with log statements. Having to fumble for the mouse to click "Clear" in the error log slows down my workflow when I am using the keyboard for almost everything else.
I realize that I can perform a browser refresh (F5) to refresh the log but most times I need to clear as well.

I guess any bindings would need to be customizable because available key combos vary between systems. Maybe defaults of Alt+[ and Alt+] might work.

Ibochkarev:
### What does it do?
Adds keyboard shortcuts to the Error Log page in the Manager: **Alt+]** (or Option+] on macOS) for Refresh and **Alt+[** (or Option+[ on macOS) for Clear. Uses Ext.KeyMap via existing `keys` support on action buttons; keyCode 219/221 so the same physical keys work across Windows and macOS and keyboard layouts.

### Why is it needed?
Frequently using the error log with the keyboard (e.g. during development) is slowed down by having to use the mouse for Refresh and Clear. Shortcuts keep the workflow keyboard-first.

### How to test
1. Open Manager → System → Error Log.
2. Press **Alt+]** (Windows) or **Option+]** (macOS) — log should refresh.
3. Press **Alt+[** (Windows) or **Option+[** (macOS) — log should clear (if you have `error_log_erase` permission).
4. Confirm shortcuts work with focus inside the log textarea.

### Related issue(s)/PR(s)
Resolves #16706

### DIFF
--- a/manager/assets/modext/sections/system/error.log.js
+++ b/manager/assets/modext/sections/system/error.log.js
@@ -17,12 +17,14 @@ MODx.page.ErrorLog = function(config) {
             ,handler: this.refreshLog
             ,scope: this
             ,hidden: config.record.tooLarge
+            ,keys: [{ key: 221, alt: true }]
         },{
             text: _('clear')
             ,id: 'modx-abtn-clear'
             ,handler: this.clear
             ,scope: this
             ,hidden: MODx.hasEraseErrorLog ? false : true
+            ,keys: [{ key: 219, alt: true }]
         },{
             text: _('cancel')
             ,id: 'modx-abtn-cancel'


mkschell (commented):
Thanks for the PR! I'm not sure we're going to address this issue. See original issue for my comments. If we were going to implement this, however, I would have wanted to see some consensus built around the default and a mechanism for overriding.

Ibochkarev:
> Thanks for the PR! I'm not sure we're going to address this issue. See original issue for my comments. If we were going to implement this, however, I would have wanted to see some consensus built around the default and a mechanism for overriding.

This can be resolved using the plugin you suggested + documentation.

mkschell:
Relevance of feature being reconsidered (see original issue).

PR #16840, opened 2026-02-21
Change getVersionData() to use include instead of include_once

### LINKED ISSUE #16655: Issue with getVersionData()
### Summary
Because `$modx->getVersionData()` uses `include_once()` and returns that result, it will return `true` if called more than once. This doesn't happen in normal MODX operations because the include only executes if `$modx->version === null`.

In unit, integration, and acceptance tests, there are cases where MODX is instantiated, and the version is checked in more than one test.  Because the tests are run in the same PHP scope, it appears that PHP knows the file was included, but `$modx->version` is `null`.

In those cases, the include returns `true`, so  `getVersionData()['version']`  becomes `1['version']`, and throws a PHP error (which goes unreported because of the @ sign). This causes problems that are very difficult to track down.

A workaround is to include the version file directly without calling the function, but I can't think of any reason `include_once` couldn't be replaced by just `include`. In normal MODX operation, the include will only happen once anyway, and including the version file a second time is completely harmless. It will return the same value each time, and the difference between the time to include the file and the time to check to see if it's already been included is probably trivial.

### Step to reproduce

```
$x = include_once 'C:\xampp\htdocs\addons\core\docs\version.inc.php';
echo "\nX: " . $x['version'];

$y = include_once 'C:\xampp\htdocs\addons\core\docs\version.inc.php';
echo "\n\nY";
var_dump($y);
```

### Observed behavior
Result: 
```
X: 2 (or 3)
Y: (bool) true
```

### Expected behavior
Return values should be the same

Ibochkarev:
### What does it do?
Replaces `include_once` with `include` in `modX::getVersionData()` so the method always returns the version array. Adds PSR12 spacing (space before `=`) on the changed line.

### Why is it needed?
When the version file was already loaded in the same PHP scope (e.g. in tests with multiple MODX instances), `include_once` returns `true` on the second run. Callers doing `getVersionData()['version']` or `getVersionData()['full_version']` then hit `true['version']`, causing a PHP error (suppressed by `@`), which is hard to track down.

### How to test
1. In a test or script, instantiate MODX (or get `$modx`) and call `$modx->getVersionData()` twice.
2. Assert both returns are arrays with keys `version`, `full_version`, etc.
3. Optionally run existing integration/acceptance tests that create multiple MODX instances.

### Related issue(s)/PR(s)
Resolves #16655

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1427,7 +1427,7 @@ class modX extends xPDO {
      */
     public function getVersionData() {
         if ($this->version === null) {
-            $this->version= @ include_once MODX_CORE_PATH . "docs/version.inc.php";
+            $this->version = @ include MODX_CORE_PATH . "docs/version.inc.php";
         }
         return $this->version;
     }


Mark-H (approved):
Only affects instantiating MODX twice, in normal usage $this->version will not be null, so no objections.

opengeek:
Can you remove changes that are formatting only on lines that are otherwise unchanged, please? We are not formatting unrelated code in PRs as it creates friction, especially with a large volume of changes that can lead to conflicts. This needs to be done on all the recent PRs you have submitted. Thanks!

opengeek (changes requested):
Changes to formatting on lines that are otherwise unchanged by the PR needs to be removed.

theboxer (approved):

Ibochkarev:
@opengeek I rolled back the changes, but now the linter crashes with errors on these lines.

opengeek:
@Ibochkarev — Ivan, please do not just add new commits reverting the formatting changes because when replayed, they will still create conflicts.  Those commits that are formatting only need to be removed from the PR completely. Making the changes and then unmaking them will not avoid conflicts when I need to rebase the changes.

opengeek:
> @opengeek I rolled back the changes, but now the linter crashes with errors on these lines.

I don't care about the linter unless it is erroring on a line that was changed by the PR.

Ibochkarev:
@opengeek It seems like everything should be ok now, the only thing is that it's complaining about formatting =)

opengeek (approved):

PR #16845, opened 2026-02-21
fix(sources): avoid loading mgr Context in MediaSource findPolicy

### LINKED ISSUE #16212: findPolicy() in modMediaSource always looks at the mgr context

Ibochkarev:
### What does it do?

`modMediaSource::findPolicy()` still looks up Media Source ACLs with `context_key = mgr` (same as the manager UI). It no longer calls `getContext('mgr')` from a frontend request just to read `access_media_source_enabled`.

If the current context is already `mgr`, or `mgr` is already in `$modx->contexts`, that context’s option is used. Otherwise the system option is used.

### Why is it needed?

On the frontend, forcing a load of the `mgr` Context object can INFO-log ACL denials for anonymous users (`Principal 0 does not have permission to load object of class modContext with primary key: mgr`). See #16212.

An earlier approach swapped the hard-coded `mgr` ACL scope for the request context. Maintainers pointed out Media Source ACLs are manager-only (`#16214`, comments from @Mark-H, @theboxer, @smg6511, @opengeek), so that change is not used here.

### How to test

1. Frontend: load a page that touches a media source as an anonymous user. Confirm the log no longer fills with mgr Context load INFO lines from `findPolicy()`.
2. Manager: open Media Sources / file browser as a user with and without source ACL. Access should match previous manager behavior.
3. `phpunit --filter MediaSourceFindPolicyTest`

### Related issue(s)/PR(s)

Resolves #16212  
Supersedes the approach discussed in #16214 for this specific INFO-noise path.

### DIFF
--- a/_build/test/Tests/Model/Sources/MediaSourceFindPolicyTest.php
+++ b/_build/test/Tests/Model/Sources/MediaSourceFindPolicyTest.php
@@ -1,0 +1,61 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+
+namespace MODX\Revolution\Tests\Model\Sources;
+
+use MODX\Revolution\modContext;
+use MODX\Revolution\MODxTestCase;
+use MODX\Revolution\Sources\modFileMediaSource;
+use MODX\Revolution\Sources\modMediaSource;
+
+/**
+ * findPolicy() must keep Media Source ACLs mgr-scoped without loading mgr Context (#16212).
+ *
+ * @group Model
+ * @group Sources
+ * @group modMediaSource
+ */
+class MediaSourceFindPolicyTest extends MODxTestCase
+{
+    public function testFindPolicyKeepsMgrScopeWithoutLoadingMgrContext()
+    {
+        /** @var modMediaSource $source */
+        $source = $this->modx->newObject(modMediaSource::class);
+        $source->fromArray([
+            'name' => 'UnitTestFindPolicySource',
+            'description' => '',
+            'class_key' => modFileMediaSource::class,
+            'properties' => [],
+        ], '', true);
+        $this->assertTrue((bool)$source->save());
+
+        unset($this->modx->contexts['mgr']);
+
+        $web = $this->modx->getObject(modContext::class, ['key' => 'web']);
+        $this->assertNotEmpty($web, 'web context fixture required');
+        $previous = $this->modx->context;
+        $this->modx->context = $web;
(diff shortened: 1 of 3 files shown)


smg6511:
There was a lot of debate about this in #16214 (which also seeks to fix this issue and is still open). It seems to me the questions posed there by @Mark-H and @theboxer — re uncovering why the line in question is there and the that media source ACLs only are relevant in the mgr context — still need to be answered.

opengeek:
> There was a lot of debate about this in #16214 (which also seeks to fix this issue and is still open). It seems to me the questions posed there by @Mark-H and @theboxer — re uncovering why the line in question is there and the that media source ACLs only are relevant in the mgr context — still need to be answered.

Correct.

smg6511:
@Ibochkarev @opengeek @theboxer @Mark-H :

I did a little digging into this and, from my findings, the issue this is trying to fix cannot be replicated. One of the scenarios that did trigger the log (INFO level only) was one in which it'd be expected—when you try to access a Resource in the front end that's assigned to a Resource Group (making it non-accessible to an anonymous user, or any non-Group user for that matter).

I did see this notice triggered by the Agenda Extra, but _maybe_ that's a small bug in the Extra's policy handling (I saw it once and now seem to be unable to make it happen again).

As to why the `$context` var hard-coded within not only the `modMediaSource` class but also the `modNamespace` class, it's (as @Mark-H was pointing out) because policies only have direct meaning for these objects in the `mgr` Context. Because of the class structure and that `findPolicy` in these classes is an override of the parent class' (`modAccessibleObject`) method its signature needs to match—so it's not that a $context is always meant to be passed. If anything, there might better way to design the policy flow in the future that makes it less confusing.

BTW, the log referenced is generated when an object's class instance is loaded (ultimately via `modAccessibleObject::_loadInstance`).

Unless someone can replicate this and give others detailed instruction on how to do the same, we should close the PRs and issue related to this. I'd like to do so in the next few days.

Ibochkarev:
@smg6511 @opengeek Hi guys! I rewrote the implementation. Please watch it again

smg6511:
Before any fix can be accepted (or even tested) there needs to be evidence of the issue that can be consistently replicated. So far there's no detailed instruction on how to do that and I've not been able to trigger this issue intentionally (or really at all outside of the expected way I mention in my note above). If someone can provide a scenario that testers can verify against, very soon, we'll keep this open. Otherwise, with all due respect to those working on this, I think we need to close the PR.

smg6511:
Closing for now; over the course of ~4 years no one has been able to (or at least has not shown how to) reproduce the logging problem raised in the original issue. As explained in [my note above](https://github.com/modxcms/revolution/pull/16845#issuecomment-4888784696), there's a reason for why it was done the way it was.

PR #16846, opened 2026-02-21
fix(user): update lastlogin/thislogin only on first session context per login

### LINKED ISSUE #16817: "Last Login" field on user page shows second-to-last login
### Summary
In the MODx manager, the "Last Login" field shows the user's second-to-last login instead of the last one.

### Step to reproduce
In the MODx manager, navigate to any user (Users > User). In the "General Information" tab, and view the "Last Login" field. The date in the field can be compared to the user's last recorded login in the manager logs. 

### Environment
MODX 3.1.2

Ibochkarev:
### What does it do?
Updates `addSessionContext()` in `modUser` so that `lastlogin` and `thislogin` are only set when adding the **first** session context for the current login (i.e. when `sessionContexts` is empty). `failedlogincount`, `logincount`, and `save()` still run on every new context as before.

### Why is it needed?
When a user had multiple manager contexts and logged into a second one, `lastlogin` and `thislogin` were updated again. That made "Last Login" show the previous session instead of the one before it. By updating these fields only on the first context per login, "Last Login" correctly reflects the prior session.

### How to test
1. Log in to the manager (first context) and note "Last Login" / "This Login".
2. Log in to another manager context (or trigger `addSessionContext` again in the same session).
3. Confirm "Last Login" still shows the previous session and "This Login" the current one, without being overwritten by the second context.

### Related issue(s)/PR(s)
Resolves #16817


smg6511:
After double-checking this I still say, as I did in the referenced issue, that there is no bug and this field reflects the correct date as-is. It's just that "Last Login" can be misleading (depends on how you interpret what _last_ means) and maybe "Previous Login" would be more clear. Note that I also logged in to a second session on a different browser as well while testing.

@opengeek @mkschell - Any thoughts on this?

mkschell:
@smg6511 I agree this is not really a bug fix, it's more of a semantic correction. It's reasonable based on what a user would expect from a "last login" field, but it's also equally valid imo to update the label as you've suggested. Since the system has behaved this way for a long time, perhaps since the beginning, and perhaps even carried over from 1.x, I'd lean toward the latter. "Previous Login", "Previous Session Login" or something like that.

smg6511 (changes requested):
Hey @Ibochkarev - It looks like consensus (including from the author of the original issue) is that the evaluation and storage of the dates are correct and it's just the label that's not quite right. To solve this, please revert your change to `modUser.php` and submit an update to the Lexicon for the field label to "Previous Login" -- Thanks!

Ibochkarev:
> Hey @Ibochkarev - It looks like consensus (including from the author of the original issue) is that the evaluation and storage of the dates are correct and it's just the label that's not quite right. To solve this, please revert your change to `modUser.php` and submit an update to the Lexicon for the field label to "Previous Login" -- Thanks!

Ok

smg6511 (approved):

PR #16848, opened 2026-02-21
feat(manager): warn when closing or leaving with unsaved changes (Resources, Elements, Files)

### LINKED ISSUE #16277: Warn when closing all unsaved objects
### Summary
When I close a changed resource without saving, I sometimes get a warning. This is inconsistent as it seems to require a content change as well as a change in field focus.

This should be made consistent in that any change should issue a warning if a close attempt is made before saving.

Additionally, this should apply to objects in the Elements and Files tabs as well as the Resources tab.

### Why is it needed?
This is needed to avoid work lost by closing a tab if the content isn't saved. 

It's particularly important for accessibility as a consistent interface throughout would improve confidence in the system.

Also, people with disability concerns might accidentally close a window or not see the difference between the tabs and wonder why they behave differently.

Ibochkarev:
### What does it do?

Unsaved-change warnings work the same for Resources, Elements, and Files when `confirm_navigation` is on:

1. Browser close/refresh: one `beforeunload` via `MODx.bindUnsavedBeforeUnload()`, checking `hasDirtyContentForm()` (any content form dirty; no field blur required).
2. Left sidebar tab switch: confirm when dirty; after Yes, a one-shot bypass avoids re-prompt loops.
3. Cancel (toolbar redirect): confirm when the form is dirty, then `MODx.loadPage` sets `suppressUnsavedWarning` so the browser does not ask again.

The resource-only `warnUnsavedChanges` beforeunload handler is removed.

### Why is it needed?

Closing or leaving a dirty resource was inconsistent, and Elements/Files did not warn the same way (#16277). That loses work and hurts accessibility.

### How to test

1. Enable **Confirm Navigation with unsaved changes**.
2. Resource: change a field without blurring → refresh/close → warning; switch left tab → Yes once (no loop); Cancel → one confirm then leave; after save → no warning.
3. Chunk/Snippet/Template/Plugin/TV: same for refresh, left tab, Cancel.
4. File create/edit: same.
5. Trash tab still opens trash when already active.

### Related issue(s)/PR(s)

Resolves #16277

### DIFF
--- a/manager/assets/modext/core/modx.component.js
+++ b/manager/assets/modext/core/modx.component.js
@@ -297,9 +297,18 @@ Ext.extend(MODx.toolbar.ActionButtons,Ext.Toolbar,{
             }
         } else {
             // if just doing a URL redirect
-            var params = itm.params || {};
+            const params = itm.params || {};
             Ext.applyIf(params, o.baseParams || {});
-            MODx.loadPage('?' + Ext.urlEncode(params));
+            const fp = o.formpanel ? Ext.getCmp(o.formpanel) : null;
+            if (fp && fp.isDirty && fp.isDirty()) {
+                Ext.Msg.confirm(_('warning'), _('resource_cancel_dirty_confirm'), function(btn) {
+                    if (btn === 'yes') {
+                        MODx.loadPage('?' + Ext.urlEncode(params));
+                    }
+                });
+            } else {
+                MODx.loadPage('?' + Ext.urlEncode(params));
+            }
         }
         return false;
     }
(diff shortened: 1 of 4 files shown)


smg6511:
@Ibochkarev - I'll take a look at this soon. A couple things: 

- I removed you as an Assignee since this is _your_ PR
- Only use `const` and `let` for variable definitions; the global `var` is almost never appropriate these days. It'll help if you run a javascript linter, as we have a bunch of rules set up that'll catch things like this.

Ibochkarev:
> @Ibochkarev - I'll take a look at this soon. A couple things:
> 
> * I removed you as an Assignee since this is _your_ PR
> * Only use `const` and `let` for variable definitions; the global `var` is almost never appropriate these days. It'll help if you run a javascript linter, as we have a bunch of rules set up that'll catch things like this.

Done — replaced var with const/let in the PR-touched code and used === for the trash tab id.

PR #16854, opened 2026-02-22
fix(core): re-init session and culture in switchContext

### LINKED ISSUE #14962: Add a public initCulture and initSession method

Ibochkarev:
### What does it do?

After a successful `switchContext()`, MODX calls `_initSession(null)` then `_initCulture(null)` so the new context’s `cultureKey` / locale (and session options when a session is not already active) take effect.

Also returns `false` when `$this->context` is null instead of fatalling on `->key`.

Does **not** expose public `initCulture()` / `initSession()` — that was rejected in review; session switching mid-request is still discouraged, and `_initSession` remains a no-op when a session is already initialized.

### Why is it needed?

Context-routing plugins call `switchContext()` after the original init path. Without re-init, lexicon/locale (and sometimes session enablement via context settings) stay on the previous context (#14962).

### How to test

1. Two contexts with different `cultureKey` (e.g. `en` / `de`).
2. From a request on one context, `$modx->switchContext('other', true)` → `$modx->cultureKey` and lexicon match the target.
3. `phpunit --filter modXSwitchContextCultureTest`

### Related issue(s)/PR(s)

Resolves #14962

### DIFF
--- a/_build/test/Tests/Model/modXSwitchContextCultureTest.php
+++ b/_build/test/Tests/Model/modXSwitchContextCultureTest.php
@@ -1,0 +1,83 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+
+namespace MODX\Revolution\Tests\Model;
+
+use MODX\Revolution\modContextSetting;
+use MODX\Revolution\MODxTestCase;
+
+/**
+ * Regression for culture re-init inside switchContext (#14962).
+ *
+ * @group Model
+ * @group modX
+ */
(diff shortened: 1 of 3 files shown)


opengeek (changes requested):
I do not see the purpose of making these methods public. First, you should never "switch" sessions in a single request, so having this as a public method is encouraging things that should not be done. If initCulture needs to be considered when usuing `switchContext()` then I think it would be better addressed in `switchContext()` rather than by exposing the `initCulture()` method publicly.

PR #16858, opened 2026-02-22
feat(tv): add min/max length options for textarea TVs

### LINKED ISSUE #16503: Give option for maximum length for textarea TVs
### Summary
Currently, text TVs have options for minimum and maximum length, but textarea TVs do not.

### Why is it needed?
textareas are just as much in need of maximum length as text fields. In some situations, like for short descriptions, text TVs don't provide enough visible space for entering a short description; even though the text in a text field would scroll to the right as you enter it, it's hard to edit like that. I have cases where we need a two- or three-line description for something and need to limit the entry to that length, and there is no way to set that up.

Ibochkarev:
### What does it do?
Adds Min length and Max length to Textarea TV input properties, with the same client validators as Text TVs, and passes them into the resource form ExtJS textarea.

### Why is it needed?
#16503 asked for textarea length limits to match existing text TV options.

### How to test
1. Edit a Textarea TV → Input Options
2. Set Min/Max length; confirm Min cannot exceed Max
3. Open a resource with that TV and confirm ExtJS enforces the limits
4. Leave both empty and confirm no min/max constraints

### Related issue(s)/PR(s)
Resolves #16503


biz87:
## Code Review

**Summary**

Adds min/max length options for textarea TV inputs. The input properties config (`textarea.php`) gets `minLength` and `maxLength` number fields with cross-validation (min < max). The textarea Smarty template conditionally includes `minLength`/`maxLength` attributes on the ExtJS text area. The TV panel JS adds validation rules for textarea matching the existing text input pattern.

**Assessment**

Consistent implementation that mirrors the existing min/max length support already available for text TVs. The cross-validation (`minLtMax`/`maxGtMin`) reuses the same validator pattern from text inputs. The Smarty template conditionals properly handle empty values. Tab index renumbering (2→4, 3→5, 4→6) accounts for the new fields correctly.

**Verdict**: Approve

manager/templates/default/element/tv/renders/input/textarea.tpl
```diff
@@ -29,6 +29,8 @@ Ext.onReady(function() {
         enableKeyEvents: true,
         msgTarget: 'under',
         allowBlank: {if $params.allowBlank == 1 || $params.allowBlank == 'true'}true{else}false{/if},
+        {if $params.minLength|default},minLength: {$params.minLength|default}{/if}
+        {if $params.maxLength|default},maxLength: {$params.maxLength|default}{/if}
```
smg6511: We're moving away from the non-standard comma first style.
```suggestion
       {if $params.minLength|default}
            minLength: {$params.minLength|default},
        {/if}
        {if $params.maxLength|default}
            maxLength: {$params.maxLength|default},
        {/if}
```

smg6511 (changes requested):
Testing showed these changes function as expected, I just have a couple small changes and optional suggestions:

I'd like to tighten up the input options layout a bit as done in the attached version of `textarea.php`. I realize not all TVs reflect this, but the general movement is toward not having Allow Blank (or other fields with small space requirements) filling its own column when it makes sense.

Optional too is adding a description for the new fields. Again, I see you're following what's been done so far but I'd personally like descriptions across the board for consistency. For these fields it might seem obvious, but length could refer to characters or words for a text-based field. So if you do add descriptions, I'd suggest:
Min - "The minimum character count for this field’s content. (Default: none)"
Max - "The maximum character count for this field’s content. (Default: none)"

[textarea.php](https://github.com/user-attachments/files/25918168/textarea.php)

smg6511 (approved):
Yup, still working as expected. Thanks for making the suggested changes!

PR #16859, opened 2026-02-22
fix(packages): handle corrupted package zip gracefully

### LINKED ISSUE #16454: Package manager should fail gracefully if package zip is corrupted
### Summary
If a package is not downloaded completely or has been corrupted some other way, the "saving" modal animation will stay on screen forever. 

```
Fatal error: Uncaught ValueError: Invalid or uninitialized Zip object in /var/www/html/core/vendor/xpdo/xpdo/src/xPDO/Compression/xPDOZip.php on line 152
```

### Step to reproduce
Try to install a corrupted package.

### Observed behavior
There is no feedback to the user. It appears like it's just taking a long time.

### Expected behavior
If possible, there should be an error communicated to the user.

### Environment
MODX version 3.0.3

Ibochkarev:
### What
Install processor catches install failures and returns a user-facing error instead of leaving the manager modal stuck on a fatal `ValueError` from a corrupted `.transport.zip`.

### Why
Truncated or corrupted package zips triggered an uncaught PHP 8 `ValueError` during install with no feedback in the UI.

### How
- Wrap `$this->package->install()` in `Install` with `try/catch (\Throwable)` and map zip/`ValueError` to `package_err_zip_invalid`.
- Add the English lexicon entry for the corrupted zip message.
- Leave `modTransportPackage::getTransport()` without a try/catch so retrieve errors propagate to the Install processor.
- Add a small processor contract test for the error mapping.

### Related
Resolves #16454

### DIFF
--- a/_build/test/Tests/Processors/Workspace/Packages/InstallErrorHandlingTest.php
+++ b/_build/test/Tests/Processors/Workspace/Packages/InstallErrorHandlingTest.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+
+namespace MODX\Revolution\Tests\Processors\Workspace\Packages;
+
+use MODX\Revolution\MODxTestCase;
+use MODX\Revolution\Processors\Workspace\Packages\Install;
+
+/**
+ * Contract for corrupted package zip error handling (#16454).
+ *
+ * @group Processors
+ * @group Workspace
+ */
+class InstallErrorHandlingTest extends MODxTestCase
+{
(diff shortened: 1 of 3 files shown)


biz87:
## Code Review

**Summary**

Adds graceful handling for corrupted package ZIP files across multiple layers: wraps `xPDOTransport::retrieve()` in try/catch in `modTransportPackage::getTransport()`, wraps `$this->package->install()` in try/catch in the Install processor, catches `\Throwable` instead of `Exception` in the File Unpack processor, and adds a new lexicon key `package_err_zip_invalid` for user-friendly error messages.

**Assessment**

Good defensive improvement. The `\Throwable` catch handles both `\Exception` and `\Error` (e.g. `\ValueError` from ZipArchive in PHP 8.x). The `getInstallErrorMessage()` helper distinguishes ZIP-related errors from general install failures, providing better user feedback. Setting `$this->package = null` on transport retrieval failure ensures the existing null check below works correctly. The `COMPLETED` log marker is properly emitted on failure so the UI console finishes.

**Verdict**: Approve

Issue #15653, opened 2021-04-08
Bug with getting information about the context


Semdevmaster:
### Summary
In templates or snippets
`$modx->context->key;//works fine`
`$modx->context->rank;//works fine`
`$modx->context->name;//does not work (always empty string)`
`$modx->context->description;//does not work (always empty string)`
### Step to reproduce
create a snippet that returns data from the current context

### Observed behavior
in all templates and snippets when requesting current context data, only key and rank available
Name and description are always empty, although filled in the admin panel
But
`$res = $modx->getObject('modContext',['key'=>'web']);
print_r($res->toArray());`
`Array
(
    [key] => web
    [name] => Москва
    [description] => Москва
    [rank] => 0
)`
Works fine
### Expected behavior
`$modx->context->name;//should output the data from admin panel`
`$modx->context->description;//should output the data from admin panel`

### Environment
MODX 2.8.1

opengeek:
The name and description of the context have never been part of the data stored in the context cache which is used to load a context when initialized. This may seem like a bug as you might expect that data to be available in the templates/snippets, but it is a feature request.

JoshuaLuckers:
If I'm not mistake the `rank` property isn't part of the data that's stored in the cache either. In the reporters example `$modx->context->rank` returns `0` because it's the default value for the property.

PR #16860, opened 2026-02-22
docs(cache): clarify why context metadata is excluded from generateContext (#15653)

Ibochkarev:
### What does it do?
Documents why `modCacheManager::generateContext()` omits mutable context row fields (`name`, `description`, `rank`, etc.) and how to load them from the DB when needed.

### Why is it needed?
#15653 asked for those fields in context cache; putting them there would stale manager UI after context edits. The doc records the contract instead of changing the cache payload.

### How to test
1. Read the PHPDoc on `generateContext()`
2. Confirm context cache still has no `name`/`description` after refresh
3. Load a context via `modContext::getObject()` and confirm those fields come from the DB

### Related issue(s)/PR(s)
Related to #15653

### DIFF
--- a/core/src/Revolution/modCacheManager.php
+++ b/core/src/Revolution/modCacheManager.php
@@ -62,6 +62,11 @@ class modCacheManager extends xPDOCacheManager
      * loading the various listings and maps in the modX class, including resourceMap, aliasMap,
      * and eventMap.  It can also be used to setup or transform any other modX properties.
      *
+     * The cached array deliberately omits mutable context row fields (name, description, rank, etc.);
+     * including them would serve stale labels in the manager after edits until full cache clears.
+     * To read those in front-end code, load the context object from the database when needed, e.g.
+     * $row = $modx->getObject(modContext::class, ['key' => $modx->context->get('key')]);
+     *
      * @todo Further refactor the generation of aliasMap and resourceMap so it uses less memory/file size.
      *
      * @param string $key     The modContext key to be cached.
@@ -79,11 +84,6 @@ class modCacheManager extends xPDOCacheManager
                 $cacheKey = $obj->getCacheKey();
                 $contextKey = is_object($this->modx->context) ? $this->modx->context->get('key') : $key;
                 $contextConfig = array_merge($this->modx->_systemConfig, $options);
-
-                $results['key'] = $obj->get('key');
-                $results['name'] = $obj->get('name');
-                $results['description'] = $obj->get('description');
-                $results['rank'] = $obj->get('rank');
 
                 /* generate the ContextSettings */
                 $results['config'] = [];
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

**Summary**

Adds `key`, `name`, `description`, and `rank` fields to the context cache data in `modCacheManager::generateContext()`. These fields are now available in the cached context results alongside `config`, `aliasMap`, `resourceMap`, etc.

**Assessment**

Small, clean addition. Including these core context fields in the cache is logical — they are part of the context object and may be needed without loading the full context from the database. No existing behavior is changed; only new keys are added to the results array.

**Verdict**: Approve

smg6511 (changes requested):
While this seems to work on the surface, it interferes with post-edit updates to objects in the Resource tree—note how edits to a Context's name fails to be reflected in the tree. The cache mgr is _probably_ not the place to be trying this anyway, especially when you take multi-context setups into account.

IMO, it's more flexible and safer to not try to build this in as a feature; it's easy enough to just do this in a Snippet:
```php
$ctx = $modx->getObject('modContext',['key' => $modx->context->key]); 
$name = $ctx->get('name');
// ... etc ...
```

Ibochkarev:
@smg6511 Thanks for the thoughtful review — totally see your point.

I’ve added a follow-up commit that drops `key`, `name`, `description`, and `rank` from what `generateContext()` puts in the context cache, so we’re not freezing those row fields in there anymore (and hopefully no more weird/stale labels in the resource tree after you edit a context).

I also expanded the `generateContext()` docblock to spell out why those fields stay out of the cache, and pointed to loading them from the DB when you need them — same idea as your snippet, e.g.
`$modx->getObject(modContext::class, ['key' => $modx->context->get('key')])`.

Happy to tweak anything if this still doesn’t match what you had in mind — whenever you get a moment for another pass, much appreciated.

core/src/Revolution/modCacheManager.php
```diff
@@ -62,6 +62,11 @@ function __construct(& $xpdo, array $options = [])
      * loading the various listings and maps in the modX class, including resourceMap, aliasMap,
      * and eventMap.  It can also be used to setup or transform any other modX properties.
      *
+     * The cached array deliberately omits mutable context row fields (name, description, rank, etc.);
+     * including them would serve stale labels in the manager after edits until full cache clears.
+     * To read those in front-end code, load the context object from the database when needed, e.g.
+     * $row = $modx->getObject(modContext::class, ['key' => $modx->context->get('key')]);
```
smg6511: Just a couple minor tweaks and we're good to go. Note that, for your example code, you don't have to use the getter for the key. Also, in general, prefer more natural descriptive var names when possible (so $context instead of $row).
```suggestion
     * $context = $modx->getObject(modContext::class, ['key' => $modx->context->key]);
     * $contextName = $context->get('name');
     * [ etc... ]
```

smg6511 (changes requested):

smg6511 (approved):
Yup, looks good. BTW @Ibochkarev, good idea to document this so others don't try the same thing ;-)

PR #16861, opened 2026-02-22
feat(install): redirect homepage to setup when config is missing

### LINKED ISSUE #15581: Redirect homepage to setup, before install
### Summary
After uploading Modx files to website, if you try to access the site's homepage it should redirect to the Modx setup page.

### Why is it needed?
Currently after uploading Modx and browsing to the site's homepage, before running setup, the user is shown a Error 503 page for "Could not load MODX config file." This is an unfriendly design.

### Suggested solution(s)
It would be a better UX design if the person is redirected directly to the Modx setup page.

### Related issue(s)/PR(s)
Maybe be related to the new setup/installer design.

Ibochkarev:
### What does it do?
If `core/config/{key}.inc.php` is missing, `index.php` responds with a 302 to the setup directory (including subfolder installs) instead of a 503.

### Why is it needed?
Fresh uploads hit an unfriendly 503 before setup (#15581).

### How to test
1. Deploy files without creating the config
2. Open the site root — expect redirect to `/setup/` (or `{subdir}/setup/`)
3. Confirm a forged `SCRIPT_NAME` with `//host` still yields a root-relative Location

### Related issue(s)/PR(s)
Resolves #15581

### DIFF
--- a/index.php
+++ b/index.php
@@ -29,8 +29,15 @@ if (!defined('MODX_CONFIG_KEY')) {
 /* redirect to setup if config does not exist (pre-install) */
 $configPath = MODX_CORE_PATH . 'config/' . MODX_CONFIG_KEY . '.inc.php';
 if (!file_exists($configPath)) {
-    $basePath = !empty($_SERVER['SCRIPT_NAME']) ? dirname($_SERVER['SCRIPT_NAME']) : '';
-    $setupUrl = rtrim($basePath, '/\\') . '/setup/';
+    if (defined('MODX_BASE_URL')) {
+        $setupUrl = rtrim(MODX_BASE_URL, '/') . '/setup/';
+    } elseif (!empty($_SERVER['SCRIPT_NAME'])) {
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

**Summary**

Adds a pre-install redirect in `index.php`: if the MODX config file doesn't exist, the user is redirected to `setup/` with a 302 response. This provides a better UX for fresh installations.

**Suggestions**

- `index.php:33-36`: The redirect URL is built from `$_SERVER['SCRIPT_NAME']`, which could potentially be manipulated. Consider using `MODX_BASE_URL` if available, or hardcoding `/setup/` as a safer fallback. In practice, since MODX isn't initialized at this point, `SCRIPT_NAME` is reasonable, but worth noting.

**Assessment**

Simple, useful improvement. The `file_exists()` check is appropriate since the config file must exist for MODX to function. The 302 redirect is the correct status code. The check runs before any MODX bootstrapping, so there's minimal overhead. The existing "MODX not installed" message further down in the file is preserved as a fallback for when the config exists but MODX can't initialize.

**Verdict**: Approve

PR #16862, opened 2026-02-22
feat(core): add error_log_max_size setting and trim on append

### LINKED ISSUE #15313: Add a setting that would limit the size of the error log
### Summary
The error log often grows to an incredible size.
It makes sense to add a setting that would limit the growth of the log, by deleting old entries.

log-size

### Why is it needed?
Limit the size of the error log.

### Suggested solution(s)
There are even add-ons that solve this problem (see discussion in https://modx.pro/components/17837), but I think it would be useful to have this in the core.

### Related issue(s)/PR(s)
N/A

Ibochkarev:
### What does it do?
Adds `error_log_max_size` (bytes, `0` = unlimited). Appends to the error log under one exclusive lock: if the next write would exceed the cap, older content is trimmed (about half the limit, line-aligned) before writing. A single entry larger than the cap keeps only its trailing `maxSize` bytes.

### Why is it needed?
#15313: the error log can grow without bound and fill the disk.

### How to test
1. Set `error_log_max_size` to a small value (e.g. 1024)
2. Trigger enough log lines to pass the limit
3. Confirm the file stays near the cap and recent lines remain
4. Set `0` and confirm no trimming

### Related issue(s)/PR(s)
Resolves #15313

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2149,6 +2149,15 @@ $settings['error_log_filepath']->fromArray([
     'area' => 'system',
     'editedon' => null,
 ], '', true, true);
+$settings['error_log_max_size'] = $xpdo->newObject(modSystemSetting::class);
+$settings['error_log_max_size']->fromArray([
+    'key' => 'error_log_max_size',
+    'value' => '0',
+    'xtype' => 'numberfield',
+    'namespace' => 'core',
+    'area' => 'system',
+    'editedon' => null,
+], '', true, true);
 $settings['passwordless_activated'] = $xpdo->newObject(modSystemSetting::class);
 $settings['passwordless_activated']->fromArray([
     'key' => 'passwordless_activated',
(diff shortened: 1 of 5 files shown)


biz87:
## Code Review

**Summary**

Adds an `error_log_max_size` system setting and implements automatic log trimming in `modCacheManager::writeFile()`. When the error log exceeds the configured size, it trims to the last 50% of the limit before appending. Uses file locking (`LOCK_EX`) to prevent concurrent write issues during trimming.

**Suggestions**

- `modCacheManager.php`: The `writeFile()` override checks `isErrorLogPath()` on every append-mode write. This involves `getOption()` calls and `realpath()` comparisons on each log write, which could impact performance under heavy logging. Consider caching the resolved error log path.
- `modCacheManager.php`: The trim operation reads the last portion of the file, truncates, then writes back. If the process crashes between `ftruncate()` and `fwrite()`, the log content is lost. Consider writing to a temporary file first, then renaming.
- The PR includes many unrelated formatting changes (spacing, line wrapping) mixed with functional changes in `modCacheManager.php`, making the diff harder to review.

**Assessment**

Useful feature for preventing error log files from growing unbounded. The file locking approach is correct for preventing concurrent corruption. The 50% trim threshold is reasonable — it avoids trimming on every write once the limit is reached. The `isErrorLogPath()` check correctly handles custom log paths and normalizes path separators.

**Verdict**: Approve — functional change is sound, but consider separating formatting changes from functional changes in future PRs.

Issue #14753, opened 2019-09-24
MODX3 - Default dashboard not respecting removal of widgets


muzzwood:
### Summary
Fresh install of latest MODX 3 dev. Removed some widgets from the default dashboard then saved. Viewed dashboard and all widgets are still there.
Went back to the edit dashboard screen and widgets are still removed.
So it's saving the changes, it must just be a problem with loading the dashboard.

It appears this only happens with the default dashboard. I created a new dashboard and assigned it to the admin group and it displayed correctly.

### Step to reproduce
Remove widgets from default dashboard then view the dashboard.

### Observed behavior
All Widgets showed on dashboard screen even though some had been removed.

### Expected behavior
Dashboard should not show widgets that have been removed.

### Environment
MODX 3.0dev

Mark-H:
Isn't editing the default dashboard a no-no anyway because that'll be overwritten on upgrade? Maybe the dashboard should be properly locked now that it's much easier to customise.

muzzwood:
Yeah that's a good point. I suppose on the dashboard list page, the "Update dashboard" option should be disabled on the default one.
It would be good to somehow make it more clear from first glance that the default needs to be duplicated too. 
Hmm.. is there a technical (or other) reason the default dashboard is overwritten on upgrade?

ghost:
I've never had my additions or removals of default widgets changed on upgrade.

muzzwood:
My understanding is the default dashboard will have the default widgets re-added if they've been removed after a MODX upgrade. Other dashboards are left as is.

ghost:
Well, in any case a better practice would be to create dashboards specific to groups that require modified dashboards.

BobRay:
I agree with Susan (Upgrade MODX gets added to the default dashboard and has always stayed there through all upgrades). I can confirm that removing widgets from the default dashboard does not remove their records from the dashboard_widget_placement table.

But something else weird seems to be going on with dashboards. The dashboard_widget_placement table now has a 'user' field, though it never has before. AFAIK, individual users don't even have dashboards, much less their own widgets, only user groups do, and there's no way to set a dashboard or a widget for an individual user in the Manager.  What, then, is the point of the 'user' field in that table?

Four of the widgets on my fresh MODX3 site appear twice in the dashboard_widget_placement table in the DB, once with a user of 0, and again with a user of 1. Widgets that only appear with a user of 0 don't show up on the dashboard they are attached to.

[...]

PR #16864, opened 2026-02-22
test(dashboard): regression for widget removal sync (#15390)

Ibochkarev:
### What does it do?
Adds `testRemovedWidgetsNotShownAfterUpdate()` to guard the widget-removal sync path from #15390.

### Why is it needed?
#14753 reported default dashboard widgets persisting after removal. #15390 fixed runtime behavior in `Dashboard\Update::setWidgets()`. This test locks that path so it does not regress.

Fixtures use xPDO-assigned IDs and `try/finally` cleanup so failed assertions do not leave dashboard data in the test DB.

### How to test
With a configured test DB:

```
core/vendor/bin/phpunit -c _build/test/phpunit.xml --filter testRemovedWidgetsNotShownAfterUpdate
```

### Related issue(s)/PR(s)
- Related: #14753
- Behavior under test: #15390

### DIFF
--- a/_build/test/Tests/Model/Dashboard/modDashboardTest.php
+++ b/_build/test/Tests/Model/Dashboard/modDashboardTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -11,9 +12,11 @@
 */
 namespace MODX\Revolution\Tests\Model\Dashboard;
 
-
 use MODX\Revolution\modDashboard;
+use MODX\Revolution\modDashboardWidget;
+use MODX\Revolution\modDashboardWidgetPlacement;
 use MODX\Revolution\modManagerController;
+use MODX\Revolution\Processors\System\Dashboard\Update;
 use MODX\Revolution\MODxTestCase;
 use xPDO\xPDOException;
 
@@ -26,7 +29,8 @@ use xPDO\xPDOException;
  * @group Dashboard
  * @group modDashboard
  */
-class modDashboardTest extends MODxTestCase {
+class modDashboardTest extends MODxTestCase
+{
     /**
      * Load some utility classes this case uses
      *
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

**Summary**

Fixes the default dashboard not respecting widget removal (#14753). When a template dashboard (user=0) is updated and a widget is removed, user-specific placements are now also updated. The Dashboard `Update` processor's `setWidgets()` is refactored to track added/removed widgets and propagate changes to user placements. The `addUserWidgets()` method in `modDashboard` is fixed to sort by rank and use `getCollection()` instead of `getMany()`. Includes a comprehensive regression test.

**Assessment**

Well-structured fix. The `setWidgets()` refactoring properly separates concerns: `syncTemplatePlacements()` for the template, `propagateAddedWidgets()` and `propagateRemovedWidgets()` for user-specific placements. The test is thorough — creates dashboard, widgets, placements, verifies initial state, performs update, and checks that removed widgets disappear from both template and user placements. Proper cleanup in the test ensures no test data leaks.

**Verdict**: Approve

mkschell (dismissed):
This solution does not appear to take into account the discussion and the lack of consensus on the solution in #14753 . It looks like a low resolution shortcut to resolve the issue as stated, without considering the nuance of the issue.

mkschell:
On further exploration, I could not reproduce the issue. It appears to have already been solved in 2021 via https://github.com/modxcms/revolution/pull/15390

The new test `testRemovedWidgetsNotShownAfterUpdate` is defensible, so perhaps re-submitting the PR with a test that covers the 2021 change would be welcome.

Thanks!

Ibochkarev:
@mkschell Thanks for the detailed review. I agree the runtime behavior is already covered by #15390 in `Dashboard\\Update::setWidgets()`.

Following your suggestion, I narrowed this PR to a regression test only (`testRemovedWidgetsNotShownAfterUpdate`), with the docblock explicitly referencing #15390 as the behavior under test and #14753 as the original report context. The earlier refactor, `addUserWidgets` / `getCollection` change, and empty-`widgets` handling are removed so the diff does not duplicate the 2021 fix.

If we want to follow up on the edge case where `widgets` is an empty array (the old `if ($widgets)` guard), I am happy to propose that as a separate small PR after maintainer consensus.

PR #16865, opened 2026-02-22
fix(core): add missing resource default system settings

### LINKED ISSUE #14727: MODX 3: Some resource settings in "System settings" are missing
### Summary
Some resource settings in "System settings" are missing.

Ther missing settings:

- show_in_tree_default
- hide_children_in_tree_default
- alias_visible_default

### Environment
MODX 3

Ibochkarev:
### What does it do?

Registers two system settings that the resource create form already reads, but that were never added to transport:

- `show_in_tree_default` — default for the "Show in tree" checkbox on new resources
- `alias_visible_default` — default for the "Use current alias in alias path" checkbox on new resources

Adds transport definitions and English lexicon entries. The `alias_visible_default` description reflects friendly URL alias path behavior, not the resource tree.

Per review feedback, this PR no longer adds `hide_children_in_tree_default`, `resource_tree_num_search_results`, an upgrade script, or manager processor/controller changes.

### Why is it needed?

`create.class.php` already calls `$this->context->getOption('show_in_tree_default', ...)` and `alias_visible_default`, but the settings were missing from transport, so administrators could not configure them in System Settings or override them per context.

Issue #14727 listed four missing settings; after review, only these two have a clear use case as configurable defaults.

### How to test

1. **Fresh install**: Run setup, go to System Settings → Site. Verify `show_in_tree_default` and `alias_visible_default` appear.
2. **Upgrade**: On an existing 3.x install, run upgrade. Verify both settings are added automatically from transport objects.
3. **Create resource**: Change the defaults (system-wide or per context), create a new resource, and verify the "Show in tree" and "Use current alias in alias path" checkboxes match the configured defaults.

### Related issue(s)/PR(s)

Resolves #14727


Mark-H:
Adding new settings don't require a migration script. They get packed into the transport as objects and are added with their defaults automatically if they don't exist. 

Should this target a 3.3 cause of new features/behavior?

biz87:
## Code Review

**Summary**

Adds missing resource default system settings: `show_in_tree_default`, `hide_children_in_tree_default`, `alias_visible_default`, and `resource_tree_num_search_results`. Includes transport data definitions, English lexicon entries, controller changes to apply defaults on resource creation, context node data for the resource tree, and a 3.2.1 upgrade script to add settings to existing installations.

**Suggestions**

- `setup/includes/upgrades/mysql/3.2.1-pl.php`: This file is also created by PRs #16886 and #16873. Whichever merges last will need to combine all includes. Consider coordinating to avoid conflicts.
- The `show_in_tree_default` setting is read in the controller (line 137) but was already there before this PR — only `hide_children_in_tree_default` is genuinely new in the controller. The diff correctly adds `hide_children_in_tree_default` to the defaults array and the checkbox fields list.

**Assessment**

Clean feature addition that fills a gap in configurable resource defaults. The upgrade script is properly idempotent (checks for existing settings before creating). Lexicon entries are clear. The context node data additions ensure the resource tree JavaScript has access to the new defaults.

**Verdict**: Approve

smg6511:
@Mark-H @mkschell @theboxer - Just hitting a few of you to get your perspectives on this. I looked at this closely even though it seems like a pretty trivial update on the surface. What I'm coming away with is this:

1. Unless I'm missing something, the setting `resource_tree_num_search_results` isn't there because it's not used. The only search functions I see are those of the quick search bar (which is just a list and uses the system setting `quick_search_result_max`) and the advanced search panel where the results reside in a grid.
2. It occurs to me that the other three settings being added here _may_ have intentionally been left out of the system settings; maybe not, but I could see why they might be —
    - `show_in_tree_default `: This could be real confusing if inadvertently set to "No" on a system- or context-wide level; the only ways to access the Resource would be directly in the URL (by id) or via the search functions (if the user knows part of the title)
    - `hide_children_in_tree_default`: Not as hazardous, but similar to above in terms of potential for confusion. Also, I'm failing to recognize why it'd be useful to hide child docs on a global level
    - `alias_visible_default`: Again, I fail to see why one would set the to "No" globally — what would be the point of setting FURLs on and this off?

Anyway, I may just not be seeing the use cases. But, if I'm preaching to the choir, it may be that a couple bits need to be taken _out_ of the current code as opposed to being added in.

smg6511:
@opengeek - Seeing if you have a position on the above comments/questions ... (?)

opengeek:
> @opengeek - Seeing if you have a position on the above comments/questions ... (?)

I honestly don't really understand the purpose of any of those settings, so I'll defer to others that might be more familiar  with the functionality of the manager UI.

rthrash:
@Ibochkarev what use case would those sytem-wide settings make sense for?

Mark-H:
`show_in_tree` is something that is typically used with a custom resource type, like Collections or Articles. Those control the value for their child resources to determine what does and doesn't show in tree, and offer an alternative (grid) UI to manage them instead. 

I agree with @smg6511 that turning that off globally is a great way to lose track of resources - unless, again, you have a Collections that controls it for you and lists resources below it in a grid instead of tree. 

`alias_visible` is something that can be used to "hide" a path, one common pattern is having a `System Pages` resource parent that contains resources you don't want typical end-users managing, maybe like a login page or sitemap. Turning off `alias_visible` for the system page makes that your friendly url becomes `/sitemap` instead of `/system-pages/sitemap`. But definitely not something I can imagine wanting globally, only for specific resources like that. 

My conclusion is that I agree with @smg6511 that these were not included in the build for a reason and don't serve a global use that we should add them now.

PR #16866, opened 2026-02-22
fix(tv): allow Smarty in date TV config for min/max values

### LINKED ISSUE #14575: Allow template variables input type date configurations to be smarty

Ibochkarev:
### What does it do?
- Uses `{eval}` in date.tpl for `minDateValue`, `maxDateValue`, `minTimeValue`, `maxTimeValue` so Smarty placeholders are processed at render time
- Skips `strtotime`/`date` conversion in the date processor when the value contains a Smarty placeholder (`{$`), avoiding incorrect output for dynamic expressions

### Why is it needed?
In JSON TV config (MIGX, CMP), date constraints were output as literal strings. Dynamic values like `"maxDateValue":"{$smarty.now|date_format:'%Y-%m-%d'}"` were not evaluated, making it impossible to limit date selection to today or other runtime values.

### How to test
1. Create a date TV with JSON config: `{"maxDateValue":"{$smarty.now|date_format:'%Y-%m-%d'}","allowBlank":true,"hideTime":true}`
2. Edit a resource with that TV — the date picker should restrict selection to today or earlier
3. Verify static values (e.g. `"maxDateValue":"2025-12-31"`) still work

### Related issue(s)/PR(s)
Resolves #14575

### DIFF
--- a/_build/test/Tests/Processors/Element/TemplateVar/Renders/mgr/input/DateInputRenderTest.php
+++ b/_build/test/Tests/Processors/Element/TemplateVar/Renders/mgr/input/DateInputRenderTest.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+
+namespace MODX\Revolution\Tests\Processors\Element\TemplateVar\Renders\mgr\input;
+
+use MODX\Revolution\modTemplateVar;
+use MODX\Revolution\MODxTestCase;
+
+/**
+ * Tests for date TV input render (minTimeValue/maxTimeValue with Smarty placeholder handling).
+ *
+ * @package modx-test
+ * @subpackage modx
+ * @group Processors
+ * @group Element
+ * @group TemplateVar
+ */
+class DateInputRenderTest extends MODxTestCase
+{
+    /**
+     * @before
+     */
+    public function setUpFixtures()
+    {
+        parent::setUpFixtures();
+        $this->modx->getService('smarty', 'MODX\Revolution\Smarty\modSmarty', '');
+    }
+
+    /**
+     * Test that static time values are converted to date format.
+     */
+    public function testProcessConvertsStaticTimeValues()
+    {
+        $placeholders = [];
+        $controller = new \stdClass();
+        $controller->setPlaceholder = function ($k, $v) use (&$placeholders) {
+            $placeholders[$k] = $v;
+        };
+        $this->modx->controller = $controller;
+
+        $tv = $this->modx->newObject(modTemplateVar::class);
+        $tv->fromArray(['id' => 1, 'name' => 'test'], '', true, true);
+
+        $dateClassPath = dirname(__DIR__, 9)
+            . '/core/src/Revolution/Processors/Element/TemplateVar/Renders/mgr/input/date.class.php';
+        $renderClass = include $dateClassPath;
+        $render = new $renderClass($tv, []);
+        $render->process('2025-01-15 14:30:00', [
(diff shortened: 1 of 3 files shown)


biz87:
## Code Review

**Summary**

Fixes the date TV input render to allow Smarty expressions (e.g. `{\$smarty.now|date_format:'%H:%M'}`) in `minTimeValue`/`maxTimeValue` config fields. The render now checks for `{\$` prefix before converting time values with `date()` + `strtotime()`, preserving Smarty placeholders for runtime evaluation. Includes tests for both static time conversion and Smarty placeholder pass-through.

**Assessment**

Targeted fix. The `strpos($params['maxTimeValue'], '{\$') === false` guard correctly identifies Smarty expressions and skips the PHP date conversion. The test approach (including the render class file directly) is pragmatic given the legacy include-based render system. Both test cases (static values converted, Smarty values preserved) cover the key scenarios.

**Verdict**: Approve

smg6511:
@Ibochkarev - Hey Ivan - there is actually no fix needed here, just a different way of implementing this type of json config—by setting the dynamic date via a Snippet, _e.g._: 
`{"maxDateValue":"[[!my-date-snippet]]","allowBlank":true,"hideTime":true}`)

I confirmed this works and generates an uncached result for the date.

A couple reasons I'd avoid the Smarty strategy this PR relies on:

- It's ideal to avoid any form of `eval` when possible
- Smarty's date implementation (even in 5.x) still relies on the deprecated `strftime` php function

smg6511:
@Ibochkarev - Friendly bump to take a look at my comment above. I don't think we're going to dig into supporting dynamic constraints via Smarty, especially given that the template engine is not likely to be the choice for the future. 

That said, it could be useful to add a note in the relevant [MODX Docs page](https://docs.modx.com/3.x/en/building-sites/elements/template-variables/input-types) re usage of snippets to provide such dynamic values.

Respectfully suggest closing this one.

smg6511:
@Ibochkarev - Hi Ivan, I'm closing this in favor of documenting the alternative. I've written up something to add to the TV Input Types page that I'll post a PR for over in the Docs repo later on today.

PR #16867, opened 2026-02-22
feat(core): add async/defer support to regClientScript and regClientStartupScript

### LINKED ISSUE #14520: regClientScript async defer

Ibochkarev:
### What does it do?
Optional `async` / `defer` for `regClientScript` and `regClientStartupScript`, plus `:jsToBottom=`0,async`` / `:jsToHead=`0,defer``. Loading is passed as a third call argument and read with `func_get_arg()` so the declared two-parameter signatures stay compatible with existing `modX` subclasses. Script URLs still dedupe; the first registration wins.

### Why is it needed?
Callers need non-blocking script tags without hand-writing HTML (#14520).

### How to test
```
core/vendor/bin/phpunit -c _build/test/phpunit.xml --filter 'testRegClientScriptAsync|testRegClientStartupScriptDefer|testClientScriptsDeduplicateByUrl|testRegClientScriptIgnoresInvalidLoading|testRegClientScriptDeclaredSignatureStaysTwoParameters|testJsToBottomAsync|testJsToHeadDefer'
```

### Related issue(s)/PR(s)
Resolves #14520

### DIFF
--- a/_build/test/Tests/Model/Filters/modOutputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modOutputFilterTest.php
@@ -800,6 +800,30 @@ goes here'
     }
 
     /**
+     * Test :jsToBottom with async loading attribute
+     */
+    public function testJsToBottomAsync() {
+        $url = 'assets/js/async-script.js';
+        $this->modx->setPlaceholder('utp', $url);
+        $this->tag->set('name', 'utp:jsToBottom=`0,async`');
+        $this->tag->process();
+        $expected = '<script src="' . $url . '" async></script>';
+        $this->assertContains($expected, $this->modx->jscripts);
+    }
+
+    /**
+     * Test :jsToHead with defer loading attribute
+     */
+    public function testJsToHeadDefer() {
+        $url = 'assets/js/defer-script.js';
+        $this->modx->setPlaceholder('utp', $url);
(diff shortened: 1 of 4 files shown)


Mark-H (approved):

biz87:
## Code Review

**Summary**

Adds `async` and `defer` loading attribute support to `modX::regClientScript()`, `modX::regClientStartupScript()`, and the `jsToHead`/`jsToBottom` output filter modifiers. The new `$loading` parameter accepts `'async'` or `'defer'` (validated, case-insensitive). The deduplication key is extended to include the loading attribute so the same script can be registered with different loading strategies. Includes tests for both the modX methods and the output filter modifiers.

**Assessment**

Well-implemented feature. The validation whitelist (`'async'`/`'defer'` only) prevents arbitrary attribute injection. The modifier parser (`parseJsModifierValue`) cleanly handles the comma-separated format (e.g. `0,async`). The dedup key extension (`$src . '|' . $loading`) is the right approach. Tests cover both the direct API and the output filter path. The method signature change adds an optional parameter — backward compatible.

**Verdict**: Approve

PR #16869, opened 2026-02-23
feat(OutputFilter): add notinarray modifier for placeholder conditions

### LINKED ISSUE #16476: [OutputFilter] Add 'notinarray' modifier
### Summary
Filter results by excluding some results

### Why is it needed?
I'm frustrated when I want to filter result by exclude some resource.
Example : [[+id:notin=`5,15,22`:then=`Yes, it's not in array`]]` 
The modifier "inarray" exists and works well, but in some case it could be more efficient to use the oposite "notinarray"

### Suggested solution(s)
Add the the modifiers notin, NOTIN, notinarray, notinArray  with this purpose : Check to see if the value is not in an array (comma seperated)

Ibochkarev:
### What does it do?
Adds `notin` / `NOTIN` / `notinarray` / `notinArray` output modifiers for exclusion checks with `:then` / `:else`.

### Why is it needed?
Templates need the inverse of `inarray` when branching on list membership (#16476).

### How to test
```
core/vendor/bin/phpunit -c _build/test/phpunit.xml --filter testNotin
```
Or in a template: `[[*id:notinarray=`1,2,3`:then=`ok`:else=`no`]]`

### Related issue(s)/PR(s)
Resolves #16476

### DIFF
--- a/_build/test/Tests/Model/Filters/modOutputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modOutputFilterTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -10,7 +11,6 @@
  * @package modx-test
 */
 namespace MODX\Revolution\Tests\Model\Filters;
-
 
 use MODX\Revolution\modPlaceholderTag;
 use MODX\Revolution\MODxTestCase;
@@ -24,7 +24,8 @@ use MODX\Revolution\MODxTestCase;
  * @group Filters
  * @group modOutputFilter
  */
-class modOutputFilterTest extends MODxTestCase {
+class modOutputFilterTest extends MODxTestCase
+{
     /** @var modPlaceholderTag $tag */
     public $tag;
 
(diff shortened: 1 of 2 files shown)


jenswittmann:
I can confirm that the output modifier works in MODX 3.2.1-dev ✅

However, I am not familiar with the other two steps for PHPUnit and PHP-CS tests, which produce errors.

biz87:
## Code Review

**Summary**

Adds a `notinarray` (aliases: `notin`, `NOTIN`, `notinArray`) output modifier that checks if a placeholder value is NOT in a comma-separated list. Returns `1` (truthy) when the value is absent from the list, `0` when present. The bulk of the diff is PSR-12 formatting cleanup of the test file.

**Assessment**

Clean addition following the existing `inarray` modifier pattern. The implementation is consistent — explode by comma, check with `in_array()`, invert with `!`. The multiple aliases match the convention used by the existing `inarray`/`inArray`/`IN` modifier. Test formatting changes (braces, spacing) are PSR-12 normalization.

**Verdict**: Approve

PR #16870, opened 2026-02-23
fix(manager): Improve query criteria in ResourceCreate controller to avoid empty IN condition

### LINKED ISSUE #16376: Missing sanity check in resource/create controller

Ibochkarev:
### What does it do?
Builds the Form Customization user-group `IN` criterion in Resource Create only when the user has groups. An empty group list no longer produces an empty `IN ()` that TypeErrors in xPDO.

### Why is it needed?
Users with no groups hit a crash opening Resource Create (#16376).

### How to test
1. Log in as a user with no user groups
2. Create a resource — page loads without TypeError
3. Repeat with a user that has groups — FC default template still applies

### Related issue(s)/PR(s)
Resolves #16376

### DIFF
--- a/manager/controllers/default/resource/create.class.php
+++ b/manager/controllers/default/resource/create.class.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of MODX Revolution.
  *
@@ -25,7 +26,6 @@ require_once __DIR__ . '/resource.class.php';
  */
 class ResourceCreateManagerController extends ResourceManagerController
 {
-
     /**
      * Check for any permissions or requirements to load page
      *
@@ -162,8 +162,11 @@ class ResourceCreateManagerController extends ResourceManagerController
             $this->resourceArray = array_merge($this->resourceArray, $reloadData);
             $this->resourceArray['resourceGroups'] = [];
             $this->resourceArray['parents'] = $this->getParents();
-            $this->resourceArray['resource_groups'] = $this->modx->getOption('resource_groups',
-                $this->resourceArray, []);
+            $this->resourceArray['resource_groups'] = $this->modx->getOption(
+                'resource_groups',
+                $this->resourceArray,
+                []
+            );
             $this->resourceArray['resource_groups'] = is_array($this->resourceArray['resource_groups'])
                 ? $this->resourceArray['resource_groups']
                 : json_decode($this->resourceArray['resource_groups'], true);
@@ -226,8 +229,9 @@ class ResourceCreateManagerController extends ResourceManagerController
         } else {
             switch ($this->context->getOption('automatic_template_assignment', 'parent', $this->modx->_userConfig)) {
                 case 'parent':
-                    if (!empty($this->parent->id))
+                    if (!empty($this->parent->id)) {
                         $defaultTemplate = $this->parent->get('template');
+                    }
                     break;
                 case 'sibling':
                     if (!empty($this->parent->id)) {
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

**Summary**

Fixes an empty `IN` condition in the resource create controller's default template query. When a user belongs to no groups, `$userGroups` is empty, causing `ProfileUserGroup.usergroup:IN => []` to generate invalid SQL. The fix builds the `$userGroupCriteria` array conditionally: the `IN` clause is only added when the user has groups. Also adds PSR-12 braces around single-line `if` statements.

**Assessment**

Correct fix for a real SQL issue. The conditional criteria construction properly handles both cases (user with groups and user without groups). The `null` fallback for `ProfileUserGroup.usergroup` is preserved so users without group assignments still get form customization applied. PSR-12 brace additions are appropriate cleanup of the touched code.

**Verdict**: Approve

mkschell (approved):
Clean fix bringing resource-create controller in line with identical guard already present in modManagerController and modTemplateVar. Resolves #16376. Passed smoke test (I verified the bug and verified the fix.) Ready to merge. Thanks!

mkschell (changes requested):
Appreciate the effort to optimize the query, but this now appears to violate the intent of the RBAC system. The MODX RBAC model is "open until locked" — once you restrict an FC profile to specific user groups, users outside those groups should NOT see those customizations. The recent commit in this PR breaks that by giving groupless users more access than grouped users — they'd see FC rules from every profile, including ones meant only for specific groups.

Ibochkarev:
> Appreciate the effort to optimize the query, but this now appears to violate the intent of the RBAC system. The MODX RBAC model is "open until locked" — once you restrict an FC profile to specific user groups, users outside those groups should NOT see those customizations. The recent commit in this PR breaks that by giving groupless users more access than grouped users — they'd see FC rules from every profile, including ones meant only for specific groups.

I made it based on a Copilot comment, but I can just delete it.

mkschell:
@Ibochkarev if you're able to reset and force push to remove that commit, that would be awesome. Thanks!

Ibochkarev:
> @Ibochkarev if you're able to reset and force push to remove that commit, that would be awesome. Thanks!

Done

smg6511:
@mkschell - Ivan made your requested changes; you good with this one? I do have one tiny comment re L267-8: It's not at all incorrect, but in terms of consistency we haven't typically used the `+` op to merge arrays (from what I recall seeing). Because there's only one item, I would've just done `$userGroupCriteria['ProfileUserGroup.usergroup:IN'] = $userGroups`. Totally up to you guys...

smg6511 (approved):
I ran through this, manually setting $userGroups to an empty array (pre- and post-PR changes), while also observing the compiled SQL and this works as expected.

@mkschell - When you get a chance take another look; I think the changes made meet your request ;-)

PR #16871, opened 2026-02-23
fix(login): return correct message when block expires after failed attempts

### LINKED ISSUE #16171: Confusing message about blocking a user
### Summary
It happens that if a user has a simple login, for example, `admin`, then, as it seems to me, the login form is blocked due to "brute force" on the manager.

And in the login form a message pops up:
`You have been blocked from the Manager by an administrator.`

login_error

In the `login.inc.php` lexicon:
`$_lang['login_blocked_admin'] = 'You have been blocked from the Manager by an administrator.';`

_But it turns out that admin blocked himself, even if there are no other users in the manager at all._

**Those either the message is displayed incorrectly, or the logic is broken in the file** https://github.com/modxcms/revolution/blob/3.x/core/src/Revolution/Processors/Security/Login.php.

### Environment
MODX 2.x >

Ibochkarev:
### What does it do?
Fixes `Login::checkIsBlocked()`: when a brute-force temp block has expired (`blockeduntil` in the past), the profile is unblocked and the method returns `false` so login continues. Permanent admin blocks (`blocked=1`, `blockeduntil=0`) still return `login_blocked_admin`.

### Why is it needed?
After auto-block from failed attempts, the first login once the block expired showed "blocked by administrator" and was rejected until a second try. See #16171.

### How to test
1. Set `failed_login_attempts` to 3; fail login 3 times → temp block message.
2. Wait for expiry (or lower `blocked_minutes`); log in with valid credentials → succeeds on first try.
3. User with `blocked=1`, `blockeduntil=0` → still gets admin block message.
4. PHPUnit: `LoginCheckIsBlockedTest` under `_build/test/Tests/Processors/Security/`.

### Related issue(s)/PR(s)
Resolves #16171

### DIFF
--- a/_build/test/Tests/Processors/Security/LoginCheckIsBlockedTest.php
+++ b/_build/test/Tests/Processors/Security/LoginCheckIsBlockedTest.php
@@ -1,0 +1,135 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+
+namespace MODX\Revolution\Tests\Processors\Security;
+
+use MODX\Revolution\modUser;
+use MODX\Revolution\modUserProfile;
+use MODX\Revolution\MODxTestCase;
+use MODX\Revolution\Processors\Security\Login;
+
+/**
+ * Tests for Login::checkIsBlocked() block-expiry behavior.
+ *
+ * @package modx-test
+ * @subpackage Processors
+ * @group Processors
+ * @group Security
+ * @group Login
+ */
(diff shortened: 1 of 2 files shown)


biz87:
## Code Review

**Summary**

Fixes `Login::checkIsBlocked()` so that when a temporarily blocked user's block period has expired, the method unblocks the user and returns `false` (allowing login) instead of falling through to the "blocked by administrator" message. Also reformats long lines in `prepareMgrResponse()` and `prepareWebResponse()`.

**Assessment**

Clear, correct bug fix. The original code unblocked the profile (reset `failedlogincount`, set `blocked = 0`, saved) but then fell through to `return $this->modx->lexicon('login_blocked_admin')`, which prevented login with a misleading message. Adding `return false;` after `$profile->save()` is the right fix — it lets the login flow continue. The `elseif` to `if` refactoring improves readability. The formatting changes in `prepareMgrResponse` and `prepareWebResponse` are minor and don't affect behavior.

**Verdict**: Approve

PR #16872, opened 2026-02-23
fix(propertyset): refresh cache after UpdateFromElement so plugin properties stay in sync

### LINKED ISSUE #16023: plugin property not cached correctly after change in the manager
### Summary
When you change the value of a default property for a plugin in the manager, the value in the cache is still the old value until you clear the cache again.

### Step to reproduce
Create a plugin with a default property (e.g. `my_prop`).
Edit the value and click the "Save" button.
Check the cached property value in the file `core/cache/context_settings/.../context.cache.php` (under `pluginCache`).

Or use some plugin code like this to compare the cached value and the value in the database.
```
$my_property = $scriptProperties['my_prop']; //cached value

//load property from DB
$plugin_db = $modx->getObject('modPlugin', $modx->event->plugin->get('id'));
$props_db = $plugin_db->getProperties();
$my_property_db = $props_db['my_prop'];

if ($my_property != $my_property_db){
    $modx->log(modX::LOG_LEVEL_ERROR,'cached plugin property ' . $my_property . ' NOT equal to value in db ' . $my_property_db);
}
```

### Observed behavior
The cached value of the property is different from the property value in the database.

### Expected behavior
The current value of the property should be cached.

### Related thread in the community forum
[https://community.modx.com/t/scriptproperties-caching-question/4907](https://community.modx.com/t/scriptproperties-caching-question/4907)

When you click the "Save" button, it generates 2 XHR-requests.
The first one with the action `element/plugin/update` clears the cache (function `afterSave()`) and `core/cache/context_settings/.../context.cache.php` is regenerated.
The second one with the action `element/propertyset/updatefromelement` sets the properties-field but doesn't refresh the cache.

### Environment
MODX 2.8.3 (same behaviour in MODX 3)

Ibochkarev:
### What does it do?
After a successful Property Set save from an element (`UpdateFromElement`), refreshes the cache when `clearCache` is not false so `pluginCache` picks up new properties.

### Why is it needed?
Saving plugin properties in the manager could leave stale plugin cache until a manual refresh (#16023).

### How to test
1. Edit a plugin property set and save
2. Confirm plugin behavior / cache reflects the new values without a separate Clear Cache

### Related issue(s)/PR(s)
Resolves #16023

### DIFF
--- a/core/src/Revolution/Processors/Element/PropertySet/UpdateFromElement.php
+++ b/core/src/Revolution/Processors/Element/PropertySet/UpdateFromElement.php
@@ -114,7 +114,9 @@ class UpdateFromElement extends Update
      */
     public function afterSave()
     {
-        if ($this->getProperty('clearCache', true)) {
+        $requestProps = $this->getProperties();
+        $clearCache = !array_key_exists('clearCache', $requestProps) || $this->modx->paramValueIsTrue($requestProps, 'clearCache');
+        if ($clearCache) {
             $this->modx->cacheManager->refresh();
         }
     }
--- a/core/src/Revolution/Processors/Element/PropertySet/UpdateFromElement.php
+++ b/core/src/Revolution/Processors/Element/PropertySet/UpdateFromElement.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -9,7 +10,6 @@
  */
 
 namespace MODX\Revolution\Processors\Element\PropertySet;
-
 
 use MODX\Revolution\modElement;
 use MODX\Revolution\modPropertySet;
@@ -109,6 +109,18 @@ class UpdateFromElement extends Update
     }
 
     /**
+     * Refresh context cache so plugin/element properties in cache match the database.
+     *
+     * @return void
+     */
+    public function afterSave()
+    {
+        if ($this->getProperty('clearCache', true)) {
+            $this->modx->cacheManager->refresh();
+        }
+    }
+
+    /**
      * {@inheritDoc}
      * @return mixed
      */
@@ -116,7 +128,9 @@ class UpdateFromElement extends Update
     {
         if (!$this->object) {
             $this->element->setProperties($this->getData());
-            $this->element->save();
+            if ($this->element->save() === false) {
+                return $this->failure($this->modx->lexicon($this->objectType . '_err_save'));
+            }
         } else {
             /* Run the beforeSave method and allow stoppage */
             $canSave = $this->beforeSave();
@@ -131,6 +145,7 @@ class UpdateFromElement extends Update
             }
         }
 
+        $this->afterSave();
         $this->logManagerAction();
 
         return $this->success();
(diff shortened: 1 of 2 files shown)


biz87:
## Code Review

**Summary**

Adds `afterSave()` to the `UpdateFromElement` processor that calls `$this->modx->cacheManager->refresh()` after saving property set changes. This ensures `pluginCache` in `context.cache.php` is rebuilt with updated property values. Also adds error handling when `$this->element->save()` fails in the default properties path.

**Assessment**

Correct fix for a real cache synchronization issue. The `element/plugin/update` request runs `afterSave()` and refreshes cache before properties are updated in the DB; then `element/propertyset/updatefromelement` updates the DB but previously skipped cache refresh. The `clearCache` property check (defaulting to `true`) follows the pattern used in other processors. The added error handling for `$this->element->save()` returning false is a good defensive improvement.

**Verdict**: Approve

core/src/Revolution/Processors/Element/PropertySet/UpdateFromElement.php
```diff
@@ -108,6 +108,18 @@ public function beforeSave()
...
     }
 
+    /**
+     * Refresh context cache so plugin/element properties in cache match the database.
+     *
+     * @return void
+     */
+    public function afterSave()
+    {
+        if ($this->getProperty('clearCache', true)) {
```
smg6511: Because the `clearCache` value will come in via js and to ensure the boolean is correctly read, need to do this a little differently:
```suggestion
        $requestProps = $this->getProperties();
        $clearCache = !array_key_exists('clearCache', $requestProps) || $this->modx->paramValueIsTrue($requestProps, 'clearCache');
        if ($clearCache) {
```

core/src/Revolution/Processors/Element/PropertySet/UpdateFromElement.php
```diff
@@ -108,6 +108,18 @@ public function beforeSave()
         return parent::beforeSave();
     }
 
+    /**
+     * Refresh context cache so plugin/element properties in cache match the database.
+     *
```
smg6511:
```suggestion
     * {@inheritdoc}
```

smg6511 (changes requested):
A little extra work had to be done on this to ensure the process has access to and respects the `clearCache` option. The current implementation will return true no matter what for a couple reasons:

1. Most importantly, the `clearCache` status of the Element being saved needs to be explicitly passed to the `UpdateFromElement` class because it's not a part of that class' object properties (an incoming commit with a change to the modx.grid.element.properties.js will fix this)
2. The value of the incoming `clearCache` value will ultimately be a string, so the `getProperty` method will not accurately assess its value.

smg6511 (approved):
Re-tested and works as expected.

Note for anyone else who may test this: The issue author gives example code for testing against props set on a plugin. I used the `OnPluginFormPrerender` event for this and observed the logging after making test prop edits, saving, and then doing a browser refresh (this being what triggers the chosen plugin event). Be aware that `OnPluginFormSave` is problematic because that event fires before the new `UpdateFromElement->afterSave` is run.

The new code covers all Elements with properties, so technically just verifying against the test plugin should cover it. But if you want to observe this on other Elements, an easy way to do it is create each Element and use it in a way that outputs your test prop value in a cacheable Resource (but with its _Empty Cache_ setting off). Then play with updating each Element's test prop to verify the change is reflected in its host Resource when the Element's _Empty Cache_ is on, and that it does not when that setting is off.

smg6511:
@halftrainedharry - Since you posted the original issue, could you review this solution when you get a chance?

PR #16873, opened 2026-02-23
fix(setup): remove upload_check_exists legacy system settings

### LINKED ISSUE #15877: There is no description for the upload_check_exists system setup

Ibochkarev:
### What changed and why

After upgrading from MODX 2.x, orphan system settings `upload_check_exists` / `upload_check_exist` can remain in the DB. They have no lexicon descriptions in MODX 3 (which uses `upload_file_exists`), so they show up blank in System Settings (#15877).

This PR adds a 3.2.1 upgrade step that deletes those legacy keys, following the same pattern as `3.0.0-cleanup-system-settings.php`.

`mysql/3.2.1-pl.php` also includes the password/cachepwd widen step from #16886 so both PRs share one version entry point and merge cleanly in either order.

### How to test

1. On a DB that still has `upload_check_exists` and/or `upload_check_exist`, run the 3.2.1 upgrade (or include the common script in a test upgrade run).
2. Confirm both legacy keys are gone.
3. Confirm `upload_file_exists` is unchanged.
4. Re-run the step: no errors when the keys are already absent.
5. Gate E:
   - `php -l setup/includes/upgrades/common/3.2.1-cleanup-upload-check-exists.php` → exit 0
   - `php -l setup/includes/upgrades/mysql/3.2.1-pl.php` → exit 0
   - `core/vendor/bin/phpcs --standard=phpcs.xml setup/includes/upgrades/common/3.2.1-cleanup-upload-check-exists.php setup/includes/upgrades/common/3.2.1-password-cachepwd-255.php setup/includes/upgrades/mysql/3.2.1-pl.php` → exit 0

### Related issue(s)/PR(s)

Resolves #15877  
Coordinates with #16886 (shared `mysql/3.2.1-pl.php`)

### Compatibility notes

Runs only during setup upgrade to 3.2.1+. Fresh 3.x installs never create the legacy keys. Safe for sites that already deleted them (idempotent remove).

No public API changes. Removes obsolete system setting rows that MODX 3 does not use. Runtime already reads `upload_file_exists` only.

### Test coverage

No PHPUnit for installer upgrade scripts in this repo (same as other `setup/includes/upgrades/common/*` steps). Verification is syntax/phpcs plus the manual upgrade checklist above.

Lexicon keys used (`system_setting_cleanup_success`, `system_setting_cleanup_failed`) already exist in `setup/lang/*/upgrades.inc.php`.

### Contributors

Thanks @JoshuaLuckers for pointing at the cleanup-upgrade approach, and @Ruslan-Aleev for confirming the 2.x→3.x leftover.

### AI tool use

Cursor agent assisted review, simplification of the cleanup script, and PR description rewrite. Human review still required before merge.

### DIFF
--- a/setup/includes/upgrades/common/3.2.1-cleanup-upload-check-exists.php
+++ b/setup/includes/upgrades/common/3.2.1-cleanup-upload-check-exists.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * Remove legacy MODX 2.x system settings replaced by upload_file_exists in MODX 3.
+ * After a normal 2.x→3.x upgrade the MODX 3 setting already exists; these orphans
+ * only linger without lexicon descriptions (#15877).
+ *
+ * @var modX $modx
+ */
+
+use MODX\Revolution\modSystemSetting;
+
+$settings = [
+    'upload_check_exists',
+    'upload_check_exist',
+];
+
+$messageTemplate = '<p class="%s">%s</p>';
+
+foreach ($settings as $key) {
+    /** @var modSystemSetting|null $setting */
+    $setting = $modx->getObject(modSystemSetting::class, ['key' => $key]);
+    if (!($setting instanceof modSystemSetting)) {
+        continue;
+    }
+
+    if ($setting->remove()) {
+        $this->runner->addResult(
+            modInstallRunner::RESULT_SUCCESS,
+            sprintf(
+                $messageTemplate,
+                'ok',
+                $this->install->lexicon('system_setting_cleanup_success', ['key' => $key])
+            )
+        );
+    } else {
+        $this->runner->addResult(
+            modInstallRunner::RESULT_WARNING,
+            sprintf(
+                $messageTemplate,
+                'warning',
+                $this->install->lexicon('system_setting_cleanup_failed', ['key' => $key])
+            )
+        );
+    }
+}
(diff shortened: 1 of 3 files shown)


biz87:
## Code Review

**Summary**

Adds a 3.2.1 upgrade script that removes legacy MODX 2.x system settings `upload_check_exists` and `upload_check_exist` from the database, since MODX 3 uses `upload_file_exists` instead.

**Issues**

- `3.2.1-cleanup-upload-check-exists.php:21-22`: The lexicon keys `system_setting_cleanup_success` and `system_setting_cleanup_failure` are used but not defined anywhere in this PR. This will result in blank/missing messages during the upgrade process. Either add these lexicon entries to `setup/lang/*/default.inc.php` or use existing lexicon keys.
- `setup/includes/upgrades/mysql/3.2.1-pl.php`: This file is also created by PR #16886. Whichever PR merges second will need to combine both includes. Consider coordinating to avoid the conflict.

**Assessment**

The cleanup logic is correct and follows existing upgrade patterns. The `getObject` + `remove()` approach is idempotent and safe. However, the missing lexicon entries for the success/failure messages need to be added for the upgrade UI to display meaningful feedback.

**Verdict**: Request changes — add the missing lexicon entries for the upgrade messages.

PR #16874, opened 2026-02-23
fix(core): initialize modError in modX::initialize() for use outside …

### LINKED ISSUE #15807: MODX 3:  $modx->initialize() doesn't initialize error handler outside of MODX
### Summary
When initializing MODX 3 outside of MODX, after `$modx->initialize()`, `$modx->error` is null.

### Step to reproduce
```
<?php
require 'C:/xampp/htdocs/modx3/config.core.php';

$cliMode = false;
$outsideModx = false;

/* Set a user (set to an actual username) */
$myUserName = 'JoeBlow';

/* Instantiate MODX if necessary */
if (!isset($modx)) {
    $outsideModx = true;

    /* Set path to MODX core directory  (not executed due to require above) */
    if (!defined('MODX_CORE_PATH')) {
        define('MODX_CORE_PATH', 'C:/xampp/htdocs/modx3/core/');
    }
    /* get the vendor autoload file */
    require_once MODX_CORE_PATH . "vendor/autoload.php";

    /* instantiate the $modx object */
    $modx = new \MODX\Revolution\modX();
    if ((!$modx) || (!$modx instanceof modX)) {
        die ('Could not create MODX class');
    }

    // require 'C:\xampp\htdocs\modx3\core\vendor\composer/../../../core/src\Revolution\Error\modErrorHandler.php';

    $modx->initialize('web');
    // $modx->initialize('mgr');
    $modx->getService('error', 'error.modError', '', '');
    if ($modx->error === null) {
        die("Error class is null");
    } else {
        die('Error class is not null');
    }
}
```

### Observed behavior
The code above will die with "Error class is null" when initializing either 'web' or 'mgr',

Uncommenting the `getService()` call solves the problem, but `getService()` is deprecated.

The long require line that includes the Error class is successful, but doesn't solve the problem. This is the path used in `_initErrorHandler()`.

### Expected behavior

$modx->initialize calls `$this->_initErrorHandler($options);` which should do what it says, doesn't. It throws no error even if it uses `require` instead of `include` for the Error class file.

`$modx->error->addError()` works fine in a snippet, but generated a call on Null error with the code above.

Question: What is the MODX 3 version of the getService() call?

### Environment
MODX 3, Windows 10, XAMPP localhost

Ibochkarev:
### What does it do?
Initializes `$modx->error` (modError) during `$modx->initialize()` so it is available when using MODX outside the web context (CLI, unit tests, extras). Adds `_initError()` to register modError in the services container and set `$modx->error`. Updates `modRequest::loadErrorHandler()` to use the container, and replaces deprecated `getService('error.modError')` in modPHPMailer with the services API.

### Why is it needed?
After `$modx->initialize()` outside MODX, `$modx->error` was null because modError was only created in `runProcessor()` and `modRequest::handleRequest()`. Any code calling `$modx->error->addError()` or `$modx->error->hasError()` before those paths caused a fatal error.

### How to test
1. Run: `$modx = new \MODX\Revolution\modX(); $modx->initialize('web');`
2. Assert `$modx->error !== null` and e.g. `$modx->error->addError('test')` does not throw.

### Related issue(s)/PR(s)
Resolves #15807

### DIFF
--- a/_build/test/MODxTestCase.php
+++ b/_build/test/MODxTestCase.php
@@ -40,6 +40,8 @@ abstract class MODxTestCase extends XTestCase {
         $this->modx = MODxTestHarness::getFixture(modX::class, 'modx');
         if ($this->modx->request) {
             $this->modx->request->loadErrorHandler();
+        }
+        if ($this->modx->error !== null) {
             $this->modx->error->reset();
         }
         /* setup some basic test-environment options to allow us to simulate a site */
--- a/_build/test/Tests/Model/Request/modRequestTest.php
+++ b/_build/test/Tests/Model/Request/modRequestTest.php
@@ -152,8 +152,9 @@ class modRequestTest extends MODxTestCase {
      * Test the loadErrorHandler method
      */
     public function testLoadErrorHandler() {
+        $this->modx->error = null;
         $this->request->loadErrorHandler();
-        $this->assertInstanceOf(modError::class, $this->modx->error,'modRequest.loadErrorHandler did not load a modError-derivative class!');
+        $this->assertInstanceOf(modError::class, $this->modx->error, 'modRequest.loadErrorHandler did not load a modError-derivative class!');
     }
 
     /**
--- a/core/src/Revolution/Mail/modPHPMailer.php
+++ b/core/src/Revolution/Mail/modPHPMailer.php
@@ -114,8 +114,8 @@ class modPHPMailer extends modMail
             case modMail::MAIL_SMTP_SECURE :
                 $this->mailer->SMTPSecure = $this->attributes[$key];
                 break;
-            case modMail::MAIL_SMTP_AUTOTLS :
-                $this->mailer->SMTPAutoTLS= $this->attributes[$key];
+            case modMail::MAIL_SMTP_AUTOTLS:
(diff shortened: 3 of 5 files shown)


biz87:
## Code Review

**Summary**

Initializes `modError` during `modX::initialize()` via a new `_initError()` method, making `$modx->error` available in CLI, unit tests, and extras that use MODX outside the web request flow. Refactors `modRequest::loadErrorHandler()` to use the services container and skip re-initialization. Updates `modPHPMailer` to use the services API instead of the deprecated `getService('error.modError')`.

**Suggestions**

- `modX.php`: The formatting changes to `regClientCSS()`, `regClientStartupScript()`, and `regClientScript()` (spacing around `=`) are unrelated to the fix. While they conform to PSR-12, they add noise to the diff and could cause merge conflicts with other PRs touching `modX.php`. Consider splitting them out.

**Assessment**

Well-structured fix addressing a real problem — `$modx->error` being null outside web requests caused fatal errors in CLI and testing contexts. The `_initError()` method follows the existing `_init*()` pattern. The guard in `loadErrorHandler()` (`instanceof modError` check) prevents double initialization. The `modPHPMailer` update correctly replaces the deprecated `getService()` pattern with the services container. Test changes are appropriate.

**Verdict**: Approve

PR #16875, opened 2026-02-23
fix(filters): evaluate TV elements bindings in tvLabel output modifier

### LINKED ISSUE #13607: tvLabel output modifier does not work with @ bindings
### Summary
`tvLabel` [output modifier](https://docs.modx.com/revolution/2.x/making-sites-with-modx/customizing-content/input-and-output-filters-(output-modifiers)#InputandOutputFilters(OutputModifiers)-Stringoutputmodifiers) (added in MODX 2.4.0) does not support `@SELECT` binding _(and presumably other @ bindings)_ for TV Input Options.

To function correctly, it would presumably need to evaluate the [`site_tmplvars.elements`](https://github.com/modxcms/revolution/pull/12122/files#diff-3b43f92f730c95fab1bd225d72f76928R655-L655) binding _first_, then proceed with the label lookup.

### Step to reproduce

1. Create a TV (Check Box, Listbox, etc)
2. Define "Input Option Values" using an `@SELECT` binding.
    
    **Rough Example:**
    
    ```sql
    @SELECT 'Example Title' AS title, 'example'
    ```
3. Map TV to a Template
4. Edit Resource in Template of assigned TV, and select available TV option, then **Save.**
5. Output label for TV option, using aforementioned output filter: `[[*example:tvLabel]]`

### Observed behavior
Output: 
_(Nothing is output)_

### Expected behavior
Output: Example Title

### Related
#12122 (introduced `tvLabel` output modifier)

### Environment
MODX 2.5.8-dev

Ibochkarev:
### What does it do?
`tvLabel` resolves TV input option values through `modTemplateVar::processBindings()` before building the label map, so `@SELECT` and other @ bindings work like in the manager.

### Why is it needed?
Front-end `tvLabel` showed raw option keys when Input Option Values used bindings (#13607).

### How to test
1. TV with `@SELECT`-based input options
2. Output `[[*tvName:tvLabel]]` on a resource
3. Confirm the label text appears, not the raw value

### Related issue(s)/PR(s)
Resolves #13607

### DIFF
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -717,11 +717,15 @@ class modOutputFilter
                                 break;
                             }
                             $o_prop = $tv->get('output_properties');
-                            $options = explode('||', $tv->get('elements'));
+                            $resourceId = ($this->modx->resource && $this->modx->resource->get('id'))
+                                ? (int) $this->modx->resource->get('id')
+                                : 0;
+                            $elementsProcessed = $tv->processBindings($tv->get('elements'), $resourceId);
+                            $options = is_string($elementsProcessed) ? explode('||', $elementsProcessed) : [];
                             $lookup = [];
                             foreach ($options as $o) {
-                                list($name, $value) = explode('==', $o);
-                                $lookup[$value] = $name;
+                                $parts = array_pad(explode('==', $o, 2), 2, '');
+                                $lookup[$parts[1]] = $parts[0];
                             }
                             if (isset($o_prop['delimiter'])) {
                                 $delimiter = $o_prop['delimiter'];
@@ -732,7 +736,7 @@ class modOutputFilter
                             }
                             $return_values = [];
                             foreach ($values as $v) {
-                                $return_values[] = $lookup[$v];
+                                $return_values[] = isset($lookup[$v]) ? $lookup[$v] : $v;
                             }
                             $output = implode($delimiter, $return_values);
                             break;
--- a/core/src/Revolution/Filters/modOutputFilter.php
+++ b/core/src/Revolution/Filters/modOutputFilter.php
@@ -724,8 +724,12 @@ class modOutputFilter
                             $options = is_string($elementsProcessed) ? explode('||', $elementsProcessed) : [];
                             $lookup = [];
                             foreach ($options as $o) {
-                                $parts = array_pad(explode('==', $o, 2), 2, '');
-                                $lookup[$parts[1]] = $parts[0];
+                                if (strpos($o, '==') === false) {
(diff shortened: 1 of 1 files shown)


jenswittmann:
Hey @Ibochkarev,
thanks a lot for your work! I tested your pull request in MODX 3.2.1-dev and everything worked perfectly ✅

jenswittmann (approved):

biz87:
## Code Review

**Summary**

Fixes the `tvLabel` output modifier to evaluate TV element bindings (`@SELECT`, `@CHUNK`, `@SNIPPET`, etc.) via `processBindings()` before building the label lookup. Previously, raw binding strings (e.g. `@SELECT ...`) were treated as literal option text. Also fixes option parsing to handle labels containing `==` (via `explode('==', $o, 2)` + `array_pad`) and returns the raw value when no matching label is found instead of generating a notice.

**Assessment**

Correct fix. The `processBindings()` call aligns with how `modTemplateVarInputRender::getInputOptions()` handles the same field, ensuring consistency between the manager input and the output modifier. Passing the current resource ID enables resource-specific bindings. The `array_pad` defensive coding prevents errors when options lack the `==` separator. Falling back to `$v` instead of accessing an undefined key is the right behavior.

**Verdict**: Approve

core/src/Revolution/Filters/modOutputFilter.php
```diff
@@ -717,11 +717,15 @@ public function filter(&$element)
...
+                                ? (int) $this->modx->resource->get('id')
+                                : 0;
+                            $elementsProcessed = $tv->processBindings($tv->get('elements'), $resourceId);
+                            $options = is_string($elementsProcessed) ? explode('||', $elementsProcessed) : [];
                             $lookup = [];
                             foreach ($options as $o) {
-                                list($name, $value) = explode('==', $o);
-                                $lookup[$value] = $name;
+                                $parts = array_pad(explode('==', $o, 2), 2, '');
+                                $lookup[$parts[1]] = $parts[0];
```
smg6511: Suggest doing this a little differently to support all typical options scenarios, where the prepared options consist of:

- labels only (Sun||Moon||Star)
- label-value pairs (Sun==1||Moon==2||Star==3)
- mixed label/label-value strings (Sun==1||Moon||Star==3)

```suggestion
                                if (strpos($o, '==') === false) {
                                    $lookup[$o] = $o;
                                } else {
                                    $parts = explode('==', $o);
                                    $lookup[$parts[1]] = $parts[0];
                                }
```

smg6511 (changes requested):
Verified the solution works for bindings. Suggested change will further support  options strings that weren't working as expected with the original modifier code or the updated code in this PR.

smg6511 (approved):
Retested and good to go!

PR #16876, opened 2026-02-23
fix(core): enforce context load permission and return 404 when denied

### LINKED ISSUE #15145: Bug with "load" context permission
### Summary
If we look at the [documentation](https://docs.modx.com/current/en/building-sites/client-proofing/security/security-tutorials/more-on-the-anonymous-user-group) we can see this

> By default, Anonymous Users are granted Load Only permissions in every Context except the "mgr" Context. Without Load Only permissions, requests to the Context would result in a 404 Page Not Found response.

Actually we can deny  Load Only permission or all permissions (but don't remove this ACL) nothing happens. No 404. This is because [context initialization](https://github.com/modxcms/revolution/blob/2.x/core/model/modx/modx.class.php#L556-L558) occurs before user initialization. And [this condition](https://github.com/modxcms/revolution/blob/8dc604e8c55ff7b18c78c1aa194bf28ac6ce30fe/core/model/modx/modx.class.php#L2437) is always true. Besides if you set **false** there no errors will be thrown. Only message "Could not load context: web" in the log file.  And no checks for this.

### Environment
MODX 2.7.3.

Ibochkarev:
### What does it do?
After the error handler loads, refuse requests when the current context is missing or the user lacks `load`, and send the site 404 page.

### Why is it needed?
Docs say denying Anonymous `load` on a context should yield 404. Context init ran before that check was enforced on the request, so denied contexts still served content (#15145).

### How to test
1. Remove or deny Anonymous `load` on the `web` context (keep the ACL row).
2. Request a front-end URL in that context.
3. Confirm a 404 (error page / header), not the normal resource.

### Related issue(s)/PR(s)
Fixes #15145

### DIFF
--- a/core/src/Revolution/modRequest.php
+++ b/core/src/Revolution/modRequest.php
@@ -73,6 +73,11 @@ class modRequest
     {
         $this->loadErrorHandler();
 
+        if (!$this->modx->context || !$this->modx->context->checkPolicy('load')) {
+            $this->modx->sendErrorPage();
+            return;
+        }
+
         // If enabled, send the X-Powered-By header to identify this site as running MODX, per discussion in #12882
         if ($this->modx->getOption('send_poweredby_header', null, true)) {
             $version = $this->modx->getVersionData();
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1613,7 +1613,7 @@ class modX extends xPDO {
             } elseif (strpos(strtolower($src), "<script") !== false) {
                 $this->sjscripts[count($this->sjscripts)]= $src;
             } else {
-                $this->sjscripts[count($this->sjscripts)]= '<script src="' . $src . '"></script>';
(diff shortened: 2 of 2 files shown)


biz87:
## Code Review

**Summary**

Enforces context `load` permission in `modRequest::handleRequest()` — if the current context is null or the user lacks the `load` policy, the request is terminated with `sendErrorPage()` (404). In `modX::_initContext()`, sets `$this->context = null` when context loading fails with no fallback, preventing the instance from being left in an inconsistent state.

**Assessment**

Important security fix. Previously, a context with `load` denied could still serve content because the permission was checked during `_initContext()` but the request continued even after the check failed. The new guard in `handleRequest()` ensures denied contexts result in a 404. Setting `$this->context = null` on failure is correct — it ensures the null check in `handleRequest()` catches the case properly. The `sendErrorPage()` call comes after `loadErrorHandler()`, so error handling is available. The formatting changes in `modX.php` (spacing normalization) are consistent with PSR-12.

**Verdict**: Approve

PR #16877, opened 2026-02-23
fix(core): preserve plugin returnedValues across invokeEvent loop

### LINKED ISSUE #15056: Returned values from plugins resets before each new plugin

Ibochkarev:
### What does it do?
In `modX::invokeEvent()`, the event object is now reset once before the plugin loop (`resetEventObject()` and `name` set), and inside the loop only `_output` is cleared between plugins. `returnedValues` and `_propagate` are no longer reset per plugin, so values set by one plugin remain available to the next and to code after `invokeEvent()`. The `$this->Event = clone $this->event` line is kept for backward compatibility.

### Why is it needed?
Previously, `resetEventObject()` was called at the start of each plugin iteration, which set `returnedValues` to `null`. So when multiple plugins were attached to the same event, any value assigned to `$modx->event->returnedValues` by an earlier plugin was lost before the next plugin ran and was not available after the event. This fix preserves `returnedValues` across plugins while still resetting per-plugin output. Minimal change; no refactor of `modSystemEvent` or event API (unlike closed PR #15577).

### How to test
1. Create two plugins bound to the same event (e.g. OnBeforeDocFormSave).
2. In the first plugin, set e.g. `$modx->event->returnedValues = ['key' => 'from plugin 1'];`.
3. In the second plugin (or in code that runs after `$modx->invokeEvent('OnBeforeDocFormSave', ...)`), read `$modx->event->returnedValues` and confirm it still contains the value set by the first plugin.

### Related issue(s)/PR(s)
Resolves #15056

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1613,7 +1613,7 @@ class modX extends xPDO {
             } elseif (strpos(strtolower($src), "<script") !== false) {
                 $this->sjscripts[count($this->sjscripts)]= $src;
             } else {
-                $this->sjscripts[count($this->sjscripts)]= '<script src="' . $src . '"></script>';
+                $this->sjscripts[count($this->sjscripts)] = '<script src="' . $src . '"></script>';
             }
         }
     }
@@ -1636,7 +1636,7 @@ class modX extends xPDO {
         } elseif (strpos(strtolower($src), "<script") !== false) {
             $this->jscripts[count($this->jscripts)]= $src;
         } else {
-            $this->jscripts[count($this->jscripts)]= '<script src="' . $src . '"></script>';
+            $this->jscripts[count($this->jscripts)] = '<script src="' . $src . '"></script>';
         }
     }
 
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

**Summary**

Fixes `modX::invokeEvent()` so that `returnedValues` and `_propagate` are preserved across plugins in the event loop. Moves `resetEventObject()` and `$this->event->name` assignment before the loop; inside the loop, only `_output` is cleared between plugins.

**Suggestions**

- This PR modifies the same `invokeEvent()` method as PR #16892 (restore event context in nested calls). The two changes overlap and will conflict on merge — whichever merges second will need to integrate both fixes. Consider coordinating with #16892 to avoid merge conflicts.

**Assessment**

Correct fix for a real bug — `resetEventObject()` was called per-plugin, wiping `returnedValues` set by earlier plugins. The change is minimal and preserves backward compatibility (the `$this->Event = clone $this->event` line is retained). The per-plugin `_output` reset is appropriate since each plugin's output should be captured independently. The formatting changes (spacing normalization) are minor and consistent.

**Verdict**: Approve — note the merge conflict with #16892.

PR #16879, opened 2026-02-24
feat(ci): add PHP-CS-Fixer and auto-fix workflow

Ibochkarev:
### What does it do?
Adds `friendsofphp/php-cs-fixer` plus a pull_request workflow that dry-runs the fixer on changed PHP under `core/`, `manager/`, and `connectors/`. The job uses `contents: read` only (no auto-commit).

### Why is it needed?
Catch style drift in CI without giving a write token to a job that installs Composer dependencies from the PR.

### How to test
1. Open a PR that touches PHP under `core/` with non-compliant style
2. Confirm the PHP-CS-Fixer check fails with a diff
3. Run `composer fix` locally, push, and confirm the check passes

### Notes
Auto-push was removed on purpose. Pin Actions by commit SHA. Local fix: `composer fix` / `composer fix:dry`.

### Related issue(s)/PR(s)
CI / DX.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
       paths-ignore:
           - "_build/templates/**"
 
-          # Allows you to run this workflow manually from the Actions tab
+  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/php-cs-fixer-auto.yml
+++ b/.github/workflows/php-cs-fixer-auto.yml
@@ -12,15 +12,18 @@ on:
 
 permissions:
     contents: write
+    pull-requests: read
 
 jobs:
     auto-fix:
+        # Same-repo PRs only: never checkout or push fork branches with write token.
+        if: github.event.pull_request.head.repo.full_name == github.repository
         runs-on: ubuntu-latest
         steps:
             - name: Checkout PR
-              uses: actions/checkout@v3
+              uses: actions/checkout@v4
               with:
-                  ref: refs/pull/${{ github.event.pull_request.number }}/head
+                  ref: ${{ github.head_ref }}
                   token: ${{ secrets.GITHUB_TOKEN }}
                   fetch-depth: 0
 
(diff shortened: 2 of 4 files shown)


opengeek (commented):
I am against this. If the codebase were already PSR-12, sure. But that is not the case and we do not need this right now. This can be considered for future repositories that are PSR-12 compliant from the start.

biz87:
## Code Review

**Summary**

Adds PHP-CS-Fixer as a dev dependency with a config file (`.php-cs-fixer.dist.php`) and a GitHub Actions workflow that auto-fixes code style in PRs by committing directly to the PR branch.

**Issues**

- `.php-cs-fixer.dist.php`: `setRiskyAllowed(true)` enables rules that can change code behavior (not just formatting). Combined with auto-committing to PR branches, this could introduce subtle bugs without the author's awareness.
- `.php-cs-fixer.dist.php`: Rules like `trailing_comma_in_multiline`, `single_quote`, and `ordered_imports` go beyond PSR-12 and introduce new style conventions that haven't been established in the project. This could create conflicts with the existing PHPCS PSR-12 enforcement and cause style ping-pong between the two tools.
- `.github/workflows/php-cs-fixer-auto.yml`: Uses `actions/checkout@v3` and `actions/cache@v3` — these should be updated to `@v4` per current GitHub Actions best practices.

**Suggestions**

- Consider limiting the fixer config to `'@PSR12' => true` only (without additional opinionated rules) to align with the existing PHPCS configuration and avoid introducing conflicting style changes.
- Consider running PHP-CS-Fixer in `--dry-run` mode in CI (reporting violations rather than auto-committing fixes), similar to how PHPCS currently works. Auto-committing style fixes to contributor PRs can be confusing and disruptive.
- Set `setRiskyAllowed(false)` unless specific risky rules are intentionally needed and documented.

**Assessment**

The intent is good — automating code style enforcement reduces review friction. However, the current configuration introduces style rules beyond the project's established PSR-12 standard and auto-commits potentially risky changes to PRs. The overlap with PHPCS needs careful consideration to avoid conflicting enforcement.

**Verdict**: Request changes — limit rules to PSR-12, disable risky rules, and consider dry-run mode instead of auto-commit.

PR #16882, opened 2026-02-24
test(security): regression for principal 0 in UserGroup ACL processors

### LINKED ISSUE #15062: [bug] Error running (anonymous) UserGroup ACL Processors
### Summary
Four security/access/usergroup processors use the equal operator when comparing the principal ID to null instead of the identical operator, preventing the processor from running successfully for an ID of 0.

### Step to reproduce
`runProcessor` on one of:
- `security/access/usergroup/context/create`
- `security/access/usergroup/context/update`
- `security/access/usergroup/resourcegroup/create`
- `security/access/usergroup/resourcegroup/update`
with `principal` set to `(int)0`

### Observed behavior
A `usergroup_err_ns` error is returned, as `0 == null`

### Expected behavior
The processor runs successfully, as `0 !== null`

Note that when running via the manager, the processor succeeds, as `'0' != null`.

### Environment
MODX 2.7.3, PHP 7.3.5

Ibochkarev:
### What does it do?
Adds a regression test that UserGroup ACL processors accept `principal=0` (Anonymous) under strict `=== null` checks.

### Why is it needed?
#15062: loose `== null` treated principal `0` as missing. Production processors on `3.x` already use `=== null`; this test locks that behavior.

### How to test
```
core/vendor/bin/phpunit -c _build/test/phpunit.xml --filter UserGroupPrincipalZeroTest
```

### Related issue(s)/PR(s)
Related: #15062

### DIFF
--- a/_build/test/Tests/Processors/Security/Access/UserGroupPrincipalZeroTest.php
+++ b/_build/test/Tests/Processors/Security/Access/UserGroupPrincipalZeroTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
--- a/_build/test/Tests/Processors/Security/Access/UserGroupPrincipalZeroTest.php
+++ b/_build/test/Tests/Processors/Security/Access/UserGroupPrincipalZeroTest.php
@@ -1,0 +1,83 @@
+<?php
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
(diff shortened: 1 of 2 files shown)


biz87:
## Code Review

**Summary**

Changes `==` to `===` for principal ID comparison in `modAccessibleObject::checkPolicy()`. This prevents principal `0` (e.g. anonymous UserGroup) from being treated as equal to `null`, which caused incorrect ACL evaluation.

**Assessment**

Important security fix. In PHP, `0 == null` evaluates to `true`, which means a principal ID of `0` would incorrectly match a `null` principal in the policy check. The strict comparison ensures proper ACL evaluation. This aligns with the same `=== null` pattern already used in the UserGroup ACL processors. Minimal, targeted change in security-critical code.

**Verdict**: Approve

PR #16884, opened 2026-02-24
fix(lexicon): each search replaces the previous

### LINKED ISSUE #14456: Lexicon stepwise refinement for search results
### Summary
I've found interesting behavior of step-by-step refinement of search results for MODX i18n.

### Step to reproduce
1. Please open Lexicon Management at /manager/?a=workspaces/lexicon
2. Leave default Namespace, Topic and Language values  - "core", "default" and "en"
3. Try to enter "Search" word, press "Enter"(for Windows) on keyboard, after (this is important - don't press "Clear filter" button!) - copy/paste or enter "Results" and press "Enter" again, and thirdly use the same way for third criteria - "More".

Please check this small video also:

steps_gif

### Observed behavior
You can see that results can be "clarified" step-by-step without "Clear filter" button

### Expected behavior
I'm confused, if this is feature I believe it should be documented well as I don't know that and believe many of us too :)
But if we have bug here previous results should be cleared and all lexicons from "core" namespace should be reloaded again according to new criteria after each form submit . 

### Environment
MODX 2.7.1

Ibochkarev:
### What does it do?

Normalizes the `query` parameter in `Workspace/Lexicon/GetList` before filtering:

- If `query` arrives as an array (duplicate request params), the last value is used.
- The value is trimmed to a single string; non-strings become an empty string.

Each lexicon search request therefore sends one criterion to the processor. No JavaScript changes in this PR.

### Why is it needed?

In Lexicon Management, repeated searches without clearing the filter could leave duplicate `query` values on the request. The processor then saw stacked criteria instead of a replacement search.

### How to test

1. Open Lexicon Management (`namespace=core`, `topic=default`, `language=en`).
2. Search for `Search` and press Enter. The grid should list entries matching `Search`.
3. Without clearing the filter, search for `Results` and press Enter. The grid should show only entries matching `Results`.
4. Repeat with `More`. Each search should replace the previous result set.

### Related issue(s)/PR(s)

Resolves #14456

### DIFF
--- a/core/src/Revolution/Processors/Workspace/Lexicon/GetList.php
+++ b/core/src/Revolution/Processors/Workspace/Lexicon/GetList.php
@@ -83,7 +83,14 @@ class GetList extends Processor
             'language' => $this->getProperty('language'),
         ];
 
+        // Normalize query to a single string so each search replaces the previous (no stepwise refinement)
         $query = $this->getProperty('query');
+        if (is_array($query)) {
+            $query = end($query);
+        }
+        $query = is_string($query) ? trim($query) : '';
+        $this->setProperty('query', $query);
+
         if (!empty($query)) {
             $where[] = [
                 'name:LIKE' => '%' . $query . '%',


biz87:
## Code Review

**Summary**

Fixes lexicon search so each new search replaces the previous one instead of refining results. Backend (`GetList.php`) normalizes the `query` parameter to a single string when an array is received. Frontend (`modx.grid.js`) calls `store.clearFilter()` before loading and resets pagination to page 1 when the search query changes.

**Assessment**

Clean fix targeting both sides of the problem. The backend guard against array query params (`is_array($query) ? end($query)`) is defensive and correct. The frontend changes are properly scoped to `param === 'query'` so other filter operations are unaffected. The `store.load({ params: { start: 0 } })` combined with `bottomToolbar.changePage(1)` ensures pagination resets consistently. This benefits all grids using the search field (Lexicon, System Settings, Namespaces, etc.).

**Verdict**: Approve

PR #16885, opened 2026-02-24
feat(core): add chunk debug placeholders (this.name, this.id, this.placeholders)

### LINKED ISSUE #14159: Enhance debugging of chunks
### Summary
Knowing what placeholders are available inside a chunk is more or less a gamble. You can read the  documentation or study the code. Both methods are not very reliable and can be quite time-consuming.

I propose to add a generic placeholder to getChunk ([[+chunk.properties]]). It will contain all properties given to getChunk and can be output inside the chunk:

pdoMenu / wayfinder chunk:
```
<!-- [[+chunk.properties]] -->
<li>[[+menutitle]]<li>
```

### Solution
Add a new placeholder to getChunk:
#1872 - https://github.com/modxcms/revolution/blob/4beca05219ea5e3b4011ff54268dc54e8623163a/core/model/modx/modx.class.php

```
    public function getChunk($chunkName, array $properties= array ()) {
        // add debug placeholder here
       $properties['chunk']['properties'] = print_r($properties,true);
        //
        $output= '';
        if ($this->getParser()) {
            $chunk= $this->parser->getElement('modChunk', $chunkName);
            if ($chunk instanceof modChunk) {
                $chunk->setCacheable(false);
                $output= $chunk->process($properties);
            }
        }
        return $output;
```

Questions:
* I'm not sure about the placeholder name. Do you have a better one?
* Can this possibly break something?

Regards,
pepebe

Edit: I guess this should also be added to parseChunk()

Ibochkarev:
### What does it do?
Optional `chunk_debug_placeholders` setting. When on, `getChunk` injects `[[+this.name]]`, `[[+this.id]]`, and `[[+this.placeholders]]` (escaped dump). `parseChunk` applies the same after load so debug keys match the caller property bag. New installs get the setting from transport; upgrades from `3.2.1-pl`.

### Why is it needed?
Chunk debugging without editing every template. Off by default.

### How to test
1. Enable `chunk_debug_placeholders`
2. `getChunk` / `parseChunk` with a chunk that prints `[[+this.name]]` / `[[+this.placeholders]]`
3. Disable the setting and confirm those placeholders stay empty

### Related issue(s)/PR(s)
Chunk debug placeholders feature.

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -308,6 +308,15 @@ $settings['cache_scripts']->fromArray([
   'area' => 'caching',
   'editedon' => null,
 ], '', true, true);
+$settings['chunk_debug_placeholders'] = $xpdo->newObject(modSystemSetting::class);
+$settings['chunk_debug_placeholders']->fromArray([
+  'key' => 'chunk_debug_placeholders',
+  'value' => false,
+  'xtype' => 'combo-boolean',
+  'namespace' => 'core',
+  'area' => 'system',
+  'editedon' => null,
+], '', true, true);
 $settings['clear_cache_refresh_trees'] = $xpdo->newObject(modSystemSetting::class);
 $settings['clear_cache_refresh_trees']->fromArray([
   'key' => 'clear_cache_refresh_trees',
--- a/core/lexicon/en/setting.inc.php
+++ b/core/lexicon/en/setting.inc.php
@@ -182,6 +182,9 @@ $_lang['setting_cache_resource_expires_desc'] = 'This value (in seconds) sets th
 $_lang['setting_cache_scripts'] = 'Enable Script Cache';
 $_lang['setting_cache_scripts_desc'] = 'When enabled, MODX will cache all Scripts (Snippets and Plugins) to file to reduce load times. MODX recommends leaving this set to \'Yes\'.';
 
+$_lang['setting_chunk_debug_placeholders'] = 'Chunk Debug Placeholders';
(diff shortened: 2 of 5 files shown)


jenswittmann:
I'm happy to confirm that this worked in MODX 3.2.1-dev, which is great for development ✅

jenswittmann (approved):

biz87:
## Code Review

**Summary**

Adds opt-in debug placeholders for chunks: `[[+this.name]]`, `[[+this.id]]`, and `[[+this.placeholders]]`, controlled by a new `chunk_debug_placeholders` system setting (default: off). Modifies both `modX::getChunk()` and the deprecated `modX::parseChunk()`. Includes the system setting definition in transport data and an English lexicon entry.

**Suggestions**

- `modX.php:1956`: In `getChunk()`, the debug placeholders are only added when `!empty($properties)`, but in `parseChunk()` (line 1979) there's no such guard. Consider aligning the behavior — either always inject debug placeholders when the setting is on, or always require non-empty properties.
- `modX.php:1956`: In `getChunk()`, `$properties['this']` is set as a nested array (which MODX flattens via dot notation), while in `parseChunk()` flat keys `this.name`, `this.id`, `this.placeholders` are used directly. Both work, but the inconsistency could confuse developers reading the code.

**Assessment**

Useful debugging feature, cleanly gated behind a system setting. The `print_r()` output for `this.placeholders` is practical for development. The private `buildChunkDebugPlaceholders()` helper keeps the logic DRY. The setting is correctly defined as `combo-boolean` in the `system` area.

**Verdict**: Approve

PR #16886, opened 2026-02-24
fix(setup): extend modUser password/cachepwd to 255 chars in 3.2.1 upgrade

### LINKED ISSUE #13929: Update modUser.password and modUser.cachepwd to 255 characters
#13926 introduces a new password hasher based on native PHP functions. The [relevant documentation](http://php.net/manual/en/function.password-hash.php) recommends ensuring hashes of up to 255 characters are supported for future hash algorithms. MODX currently only supports 100 characters. 

I've considered including this change in #13926, however opted against that because:

- It is not strictly necessary for a first implementation, as PHP now (and in the forseeable future, at least not before PHP 7.4) uses PASSWORD_BCRYPT as hashing algorithm, which will always return 60 characters. The modNative implementation also does not, by design to keep crypto simple and secure, make that configurable, so MODX will not be generating any longer hashes any time soon.

- The current state of MODX 3 is in a bit of limbo, with important PRs like #13888 / #13900 potentially conflicting with any model change I try to do now. The core build.model.php is broken due to path and class name changes and when fixed causes the core model to be updated to an xPDO 3 model, which is not the intention of a (relatively minor) fix and has much more farreaching consequences that the MODX 3 project team is, to my limited knowledge, still working through.

Ibochkarev:
### What does it do?
3.2.1 upgrade: widen `modUser.password` and `modUser.cachepwd` to `varchar(255)`.

Also includes the shared `mysql/3.2.1-pl.php` entry point used by #16873 (upload settings cleanup), so both PRs merge without file conflicts.

### Why is it needed?
Schema already declares 255 chars; older DBs may still truncate hashes. Issue #13929.

### How to test
Upgrade a 3.2.x DB; confirm both columns are `varchar(255)`. Re-run the step; it stays idempotent.

### Related issue(s)
Fixes #13929

### DIFF
--- a/setup/includes/upgrades/common/3.2.1-cleanup-upload-check-exists.php
+++ b/setup/includes/upgrades/common/3.2.1-cleanup-upload-check-exists.php
@@ -2,8 +2,8 @@
 
 /**
  * Remove legacy MODX 2.x system settings replaced by upload_file_exists in MODX 3.
- * Migrates a legacy value only when upload_file_exists is missing; never overwrites
- * an existing MODX 3 setting the admin may already have changed.
+ * After a normal 2.x→3.x upgrade the MODX 3 setting already exists; these orphans
+ * only linger without lexicon descriptions (#15877).
  *
  * @var modX $modx
  */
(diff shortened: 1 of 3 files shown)


biz87:
## Code Review

**Summary**

Adds a 3.2.1 upgrade script that alters `modUser.password` and `modUser.cachepwd` columns to `varchar(255)` to support longer password hashes from future PHP algorithms. Uses the standard `alterField` + `processResults` pattern. Idempotent and safe for any upgrade path.

**Assessment**

Simple, well-structured upgrade script following existing conventions. The common/platform split (`common/3.2.1-password-cachepwd-255.php` included from `mysql/3.2.1-pl.php`) is the correct pattern used throughout the upgrade system.

Note: PR #16873 also creates `setup/includes/upgrades/mysql/3.2.1-pl.php`. Whichever merges second will need to integrate the other's include into the shared file.

**Verdict**: Approve

PR #16887, opened 2026-02-24
fix(core): resolve SQL error when removing user or user group

### LINKED ISSUE #13725: SQL error in log when removing a user
### Summary
The following error is logged when removing/deleting a user.
```[2017-12-19 12:29:12] (ERROR @ /home/vagrant/Code/modx/public_html/core/xpdo/xpdo.class.php : 1320) Could not get table class for class: modAccess
[2017-12-19 12:29:12] (ERROR @ /home/vagrant/Code/modx/public_html/core/xpdo/xpdo.class.php : 1289) Could not get table name for class: modAccess
[2017-12-19 12:29:12] (ERROR @ /home/vagrant/Code/modx/public_html/core/xpdo/om/xpdoobject.class.php : 240) Error 42000 executing statement: 
Array
(
    [0] => 42000
    [1] => 1064
    [2] => You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `modAccess` WHERE `modAccess`.`principal` = 3' at line 1
)
```

### Step to reproduce
Create a test user then remove it via the manager then check the error log.

### Observed behavior
This was originally thought to be a problem with the Login Extra. The same error was appearing when adding a user and assigning a usergroup. See here: https://github.com/modxcms/Login/issues/117
However I noticed it also happens when deleting a user account so the cause may be in the core.
It's worth noting that everything works as expected - the user is removed- it's just that the error is logged.

### Expected behavior
If removing the user is successful there should not be an error logged.

### Environment
MODX 2.6.1, PHP7.1, Tried on both Apache and Nginx, Multiple versions of MySQL.

Ibochkarev:
### What does it do?
Remove the composite `Acls` relation from `modPrincipal` in the schema and mysql map, because `modAccess` is abstract and has no table. xPDO cascade-delete attempted to query a non-existent table, generating invalid SQL (`FROM AS modAccess`). Add manual ACL deletion in `modPrincipal::remove()` before `parent::remove()`, iterating over `principal_targets` tables (Context, ResourceGroup, Category, MediaSource, Namespace). Refactor `modUserGroup::remove()` to remove its duplicate ACL deletion block and rely on `modPrincipal::remove()`.

### Why is it needed?
When deleting a user (or user group) via the Manager, xPDO tried to cascade-delete the composite `Acls` of class `modAccess`. `modAccess` is abstract—real tables are `access_context`, `access_resource_groups`, etc. xPDO could not resolve a table name and produced broken SQL, causing errors in the log even though the user was removed successfully.

### How to test
1. Create a test user in Manager (Security → Users).
2. Delete the user from the grid.
3. Check the error log—there should be no "Could not get table name for class: modAccess" or SQL syntax errors.
4. Verify user removal and User Group removal still work correctly.

### Related issue(s)/PR(s)
Fixes #13725

### DIFF
--- a/core/model/schema/modx.mysql.schema.xml
+++ b/core/model/schema/modx.mysql.schema.xml
@@ -799,7 +799,6 @@
     </object>
 
     <object class="modPrincipal" extends="xPDO\Om\xPDOSimpleObject">
-        <composite alias="Acls" class="modAccess" local="id" foreign="principal" cardinality="many" owner="local" />
     </object>
 
     <object class="modPropertySet" table="property_set" extends="xPDO\Om\xPDOSimpleObject">
--- a/core/src/Revolution/modPrincipal.php
+++ b/core/src/Revolution/modPrincipal.php
@@ -45,8 +45,7 @@ abstract class modPrincipal extends xPDOSimpleObject
             \MODX\Revolution\Sources\modAccessMediaSource::class,
             modAccessNamespace::class,
         ]);
-        $targets = explode(',', $this->xpdo->getOption('principal_targets', null, $defaultTargets));
-        array_walk($targets, 'trim');
+        $targets = array_map('trim', explode(',', $this->xpdo->getOption('principal_targets', null, $defaultTargets)));
 
         foreach ($targets as $target) {
             $fields = $this->xpdo->getFields($target);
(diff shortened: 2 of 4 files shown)


biz87:
## Code Review

**Summary**

Fixes SQL errors when removing users or user groups by removing the invalid composite `Acls` relation from `modPrincipal` schema (since `modAccess` is abstract with no table) and adding manual ACL deletion in `modPrincipal::remove()` that iterates over concrete `principal_targets` tables. Refactors `modUserGroup::remove()` to rely on the parent's new deletion logic instead of duplicating it.

**Suggestions**

- `modPrincipal.php:50`: `array_walk($targets, 'trim')` is called with the built-in `trim` function, but `array_walk` passes `(&$value, $key)` as arguments. This means `trim` receives the numeric index as the second parameter (character mask), which could inadvertently strip digits from the beginning/end of class names. Consider using `$targets = array_map('trim', $targets)` instead. Note: this is an existing pattern in `getAttributes()`, not introduced by this PR, but since you're writing new code with the same pattern, it would be good to fix it here.

**Assessment**

Correct fix for a real problem — xPDO cannot cascade-delete through an abstract class. Moving ACL cleanup to `modPrincipal::remove()` is architecturally sound since both `modUser` and `modUserGroup` extend `modPrincipal`. The raw SQL approach is appropriate here since xPDO can't resolve the abstract `modAccess` table. The `modUserGroup::remove()` refactoring properly removes duplicate logic. Schema and generated model changes are consistent.

**Verdict**: Approve

PR #16889, opened 2026-02-24
Fix template previews when installed in subdirectory

### LINKED ISSUE #16256: Create document does not render template preview
image

Does not render image 

Path is
URL/connectors/system/phpthumb.php?src=PATH_TO_IMAGE&w=335&h=236&HTTP_MODAUTH=AUTH_ID&zc=1

Tried different php versions 7.4 8.0 8.1
Tried diff path absolute vs media source

Ibochkarev:
### What does it do?

Updates `Template/GetList` to pass the template's stored `preview_file` path and `source` id directly to phpThumb, instead of building a preview URL via `getPreviewUrl()`.

phpThumb expects a filesystem path (plus optional media source), not a full HTTP URL. When MODX runs in a subdirectory, `getPreviewUrl()` produced a URL that phpThumb could not resolve, so template previews in the Create document picker appeared blank.

### Why is it needed?

Template preview images did not render when creating a new document on sites installed in a subdirectory. The template picker called phpThumb with `src` set to a full URL; the default media source expects a path, so the image request failed.

### How to test

1. Install MODX in a subdirectory (e.g. `base_url` = `/subdir/`).
2. Create a template with a preview image (Template → Preview Image).
3. Open Create document and open the template picker.
4. Select a template with a preview image and confirm the preview loads (no blank area).
5. On a root install (`base_url` = `/`), repeat and confirm previews still work.

### Related issue(s)/PR(s)

Resolves #16256


biz87:
## Code Review

**Summary**

Fixes template preview images when creating documents. Adds `getPreviewPath()` and `getPreviewSourceId()` to `modElement` to provide filesystem paths (instead of URLs) to phpThumb. Updates `Template/GetList` to use these new methods. Adds URL-to-path conversion in `modMediaSource::prepareSrcForThumb()` for subdirectory installs. Removes `error_reporting(E_ALL)` from the PhpThumb processor and fixes a PHP 8.1 deprecation (`pathinfo` returning null).

**Suggestions**

- `modElement.php`: `getPreviewPath()` and `getPreviewSourceId()` share significant logic (fetching source, initializing, checking `basePath + previewfile`). Consider extracting the shared source resolution into a private helper to reduce duplication.
- `modMediaSource.php:1585-1607`: The URL-to-filesystem conversion in `prepareSrcForThumb()` does `file_exists()` checks — ensure this doesn't introduce a performance concern when called frequently (e.g. in a large template list), since `file_exists()` hits the filesystem on each call.

**Assessment**

Solid fix addressing a real issue with phpThumb expecting filesystem paths rather than URLs. The PhpThumb processor cleanup (removing `error_reporting(E_ALL)`, guarding `pathinfo` return) is a good improvement. The `is_string($ext)` guard before `strtolower()` properly handles PHP 8.1 deprecations. Multi-file change is well-coordinated across the stack.

**Verdict**: Approve

core/src/Revolution/Processors/Element/Template/GetList.php
```diff
@@ -73,18 +73,20 @@ public function beforeIteration(array $list)
...
-        $preview = $object->getPreviewUrl();
-
-        if (!empty($preview)) {
+        $preview = '';
+        $previewPath = $object->getPreviewPath();
+        if (!empty($previewPath)) {
             $imageQuery = http_build_query([
-                'src'           => $preview,
+                'src'           => $previewPath,
+                'source'        => $object->getPreviewSourceId(),
```
smg6511: The fix for this really only needs a net 2-line change from the original file:
```suggestion
        $preview = $object->get('preview_file') ?? '';

        if (!empty($preview)) {
            $imageQuery = http_build_query([
                'src'           => $preview,
                'source'        => $object->get('source'),
```

You don't need the new methods in `modElement` and `modMediaSource`. Also, the guard in `modPhpThumb` isn’t necessary because, when accessing a specific part of the pathinfo using the 2nd param flag, an empty string is always returned if  that part (in this case _extension_) has no value.

smg6511 (changes requested):
For others testing, I found that the preview problem solved here only occurs when MODX is installed in a subdirectory.

core/src/Revolution/Processors/Element/Template/GetList.php
```diff
@@ -73,18 +73,20 @@ public function beforeIteration(array $list)
...
                 'w'             => 335,
                 'h'             => 236,
                 'HTTP_MODAUTH'  => $this->modx->user->getUserToken($this->modx->context->get('key')),
                 'zc'            => 1
             ]);
 
-            $preview = $this->modx->getOption('connectors_url', MODX_CONNECTORS_URL) . 'system/phpthumb.php?' . urldecode($imageQuery);
+            $connectorsUrl = $this->modx->getOption('connectors_url', MODX_CONNECTORS_URL);
+            $preview = $connectorsUrl . 'system/phpthumb.php?' . urldecode($imageQuery);
```
smg6511: This isn't a big deal, but personally I'd revert this back to the original one-liner since it's not a super long concatenation.

smg6511:
@Ibochkarev - Hi Ivan, just a bump to take a look at the suggestions here. I'll review again once you have ;-)

Ibochkarev:
> @Ibochkarev - Hi Ivan, just a bump to take a look at the suggestions here. I'll review again once you have ;-)

Done

smg6511 (approved):
Ok, looks good.

opengeek:
Once again, this PR title/description does not seem to reflect what this PR ended up being after all the review/changes. Please update this.

Ibochkarev:
@opengeek Updated the PR description to match the final scope after review: a single change in `Template/GetList` that passes `preview_file` and `source` to phpThumb instead of a full URL from `getPreviewUrl()`. The title was already updated to reflect the subdirectory install case.

opengeek (approved):

PR #16892, opened 2026-02-25
fix(core): restore event context in nested invokeEvent calls

### LINKED ISSUE #14063: Incorrect event name when calling static plugin
### Summary
Incorrect event name when calling static plugin.

### Step to reproduce
1. Create **static** plugin "showEventName". Select event _OnWebPageInit_ or _OnWebPagePrerender_, for instance.
```
<?php
die($modx->event->name);
```
2. Open website. Event name is correct - The name of the event that was selected for plugin.
3. Create empty plugin "checkMediaSource". Select event **OnMediaSourceGetProperties**.
4. Open website.

### Observed behavior
Displayed event name is **OnMediaSourceGetProperties** - incorrect.

### Expected behavior
The name of the selected event should be displayed.

### Environment
MODX 2.6.5, PHP 7.0, Nginx 1.13.6

Ibochkarev:
### What changed and why

Nested `$modx->invokeEvent()` calls overwrote `$modx->event`. Outer plugins then read the inner event name (for example `OnMediaSourceGetProperties` instead of `OnWebPageInit`) (#14063).

This PR saves and restores `$this->event` in `invokeEvent()` with try/finally.

### How to test

`$modx->event` is one shared object. A nested `invokeEvent()` used to replace it, so after the inner call returned the outer plugin still saw the inner event. This branch keeps the previous `$this->event` and restores it in `finally`.

You do not need a custom media source class. The stock Filesystem source (usually id 1) is enough.

**Plugin `showEventName`**, event `OnWebPageInit` (static file or inline):

```php
$before = $modx->event->name;
$source = $modx->getObject(\MODX\Revolution\Sources\modMediaSource::class, 1);
$source->initialize();
$source->getProperties();
$after = $modx->event->name;
die("before={$before} after={$after}");
```

`getProperties()` fires `OnMediaSourceGetProperties` from inside the outer plugin.

**Plugin `checkMediaSource`**, event `OnMediaSourceGetProperties`. Empty body is enough.

Clear the cache, then open any frontend resource.

| | `before` | `after` |
|---|---|---|
| 3.x | `OnWebPageInit` | `OnMediaSourceGetProperties` |
| this PR | `OnWebPageInit` | `OnWebPageInit` |

The original report in #14063 used `die($modx->event->name)` on a static `OnWebPageInit` plugin. Adding the empty `OnMediaSourceGetProperties` plugin was enough on that site because core already loaded media source properties during the request.

### Related issue(s)/PR(s)

Resolves #14063

### Compatibility notes

All contexts using `invokeEvent()`. Plugins that accidentally relied on nested overwrite may behave differently.

Restores documented isolation for nested events. No signature changes. Patch-safe for well-behaved plugins; plugins that depended on leaked inner event state will see the outer event again.

### Test coverage

No new PHPUnit in this PR. Use the dummy plugins above.

### Contributors

@rthrash asked for a before/after and dummy plugins for review.

### AI tool use

Cursor helped expand the How to test section after review.

### DIFF
--- a/core/src/Revolution/modX.php
+++ b/core/src/Revolution/modX.php
@@ -1705,51 +1705,64 @@ class modX extends xPDO {
             //$this->log(modX::LOG_LEVEL_DEBUG,'System event '.$eventName.' was executed but does not exist.');
             return false;
         }
-        $results= [];
-        if (count($this->eventMap[$eventName])) {
-            $this->event= new modSystemEvent();
-            foreach ($this->eventMap[$eventName] as $pluginId => $pluginPropset) {
-                /** @var modPlugin $plugin */
-                $plugin= null;
-                $this->Event = clone $this->event;
-                $this->event->resetEventObject();
-                $this->event->name= $eventName;
-                if (isset ($this->pluginCache[$pluginId])) {
-                    $plugin= $this->newObject(modPlugin::class);
-                    $plugin->fromArray($this->pluginCache[$pluginId], '', true, true);
-                    $plugin->_processed = false;
-                    if ($plugin->get('disabled')) {
-                        $plugin= null;
+        $results = [];
+        $previousEvent = $this->event;
+        try {
+            if (count($this->eventMap[$eventName])) {
+                $this->event = new modSystemEvent();
+                foreach ($this->eventMap[$eventName] as $pluginId => $pluginPropset) {
+                    /** @var modPlugin $plugin */
+                    $plugin = null;
+                    $this->Event = clone $this->event;
+                    $this->event->resetEventObject();
+                    $this->event->name = $eventName;
+                    if (isset($this->pluginCache[$pluginId])) {
+                        $plugin = $this->newObject(modPlugin::class);
+                        $plugin->fromArray($this->pluginCache[$pluginId], '', true, true);
+                        $plugin->_processed = false;
+                        if ($plugin->get('disabled')) {
+                            $plugin = null;
+                        }
+                    } else {
+                        $plugin = $this->getObject(
+                            modPlugin::class,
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

**Summary**

Fixes `modX::invokeEvent()` to save and restore `$this->event` using try/finally, preventing nested `invokeEvent()` calls from corrupting the outer event context. Previously, when a plugin triggered another event (e.g. `OnMediaSourceGetProperties` during `OnWebPageInit`), the inner call overwrote `$this->event` and the outer plugin would see the wrong event name after the nested call returned.

**Assessment**

Clean, correct fix. The try/finally pattern is the right approach — it guarantees `$this->event` is restored even if an exception is thrown during plugin execution. The `return $results` inside the try block works correctly with finally in PHP. The formatting changes (spacing normalization, multi-line `getObject` call) are minor and consistent with PSR-12. No behavioral side effects — the method returns the same results, it just properly restores state afterward.

**Verdict**: Approve

rthrash:
@Ibochkarev I was going to test/review this now but it wasn't immediately clear what the before failure state and the after correct state would look like. Example dummy plugins and media sources would be super-helpful in getting this signed off on, too.

Ibochkarev:
> @Ibochkarev I was going to test/review this now but it wasn't immediately clear what the before failure state and the after correct state would look like. Example dummy plugins and media sources would be super-helpful in getting this signed off on, too.

@rthrash I updated the description and added an example

milanidavide:
I hit this exact issue today.

Loading a static plugin through the Filesystem media source triggered `OnMediaSourceGetProperties`. Because MODX did not restore the outer event context after the nested invocation, the nested event replaced it.

PR #16894, opened 2026-02-25
fix(resource): Return full resource from resource/create getObject()

### LINKED ISSUE #13912: resource/create processor no long returns fields other than id
### Summary
The modProcessor $response->getObject() method for the resource/created processor used to return an array of the object's fields. It still does so for elements.

For resources, $response->getObject() returns an array containing only the ID

### Step to reproduce
Create a snippet called Test94 with this code:

```
<?php
$output = '';
$fields = array(
    'pagetitle' => 'My resource 94',
    'description' => 'Resource 94 description',
    'content' => 'This is the content',
    /* any other fields here */
);
$response =
    $modx->runProcessor(
        'resource/create',
        $fields);

if ($response->isError()) {
    if ($response->hasFieldErrors()) {
        $fieldErrors =
            $response->getAllErrors();
        $errorMessage = implode("\n",
            $fieldErrors);
    } else {
        $errorMessage =
            'An error occurred: ' .
            $response->getMessage();
    }
    $output = $errorMessage;
} else {
    $fields = $response->getObject();
    $output = print_r($fields, true);
}
return $output;
```
Create a resource with this code and view it:

[[Test94]]

### Observed behavior
Output is 

`Array ( [id] => 6 )`

### Expected behavior
The array should contain the other resource fields

### Environment
MODX 3.x, MODX 2.6.3, XAMPP windows, PHP 7

Ibochkarev:
### What changed and why

`resource/create` returned only `['id' => …]` from `getObject()` while other Create processors return the full record. API consumers expected pagetitle, content, and other fields (#13912).

This PR makes `Resource\Create::cleanup()` return the full resource object.

### How to test

1. `runProcessor('resource/create', …)`; inspect `getObject()` for full fields.
2. Run `ResourceCreateTest.php`.

### Related issue(s)/PR(s)

Resolves #13912

### Compatibility notes

Connector/API callers of `resource/create`.

**Intentional behavior change:** `getObject()` now includes all resource fields, not only `id`. Code that assumed a single-key array must use `['id']` explicitly or check keys. Not patch-safe for callers that depended on the minimal shape.

### Test coverage

`ResourceCreateTest.php` (existing or updated in branch).

### Contributors

None beyond author.

### AI tool use

### DIFF
--- a/core/src/Revolution/Processors/Resource/Create.php
+++ b/core/src/Revolution/Processors/Resource/Create.php
@@ -237,7 +237,7 @@ class Create extends CreateProcessor
     {
         $this->object->removeLock();
         $this->clearCache();
-        return $this->success('', ['id' => $this->object->get('id')]);
+        return $this->success('', $this->object);
     }
 
     /**


biz87:
## Code Review

**Summary**

One-line change in `Resource\Create::cleanup()` that returns the full resource object (`$this->object`) instead of a minimal `['id' => $this->object->get('id')]` array. This aligns resource creation with how other element create processors already work, providing all resource fields via `$response->getObject()`.

**Assessment**

Clean, minimal change. The `success()` method calls `toArray()` on xPDO objects, so all resource fields will be serialized in the response. This is consistent with how `Element\Chunk\Create`, `Element\Snippet\Create` and similar processors behave.

Worth noting this is a **behavioral change** — any existing code that relies on `$response->getObject()` returning only `['id' => N]` will now receive the full resource array. This should be a non-breaking improvement since the `id` field is still present, but callers that use strict equality checks on the response structure could theoretically be affected.

**Verdict**: Approve — straightforward fix that improves API consistency.

PR #16895, opened 2026-02-25
fix(manager): remove URI maxLength and show URI field on validation errors

### LINKED ISSUE #13815: Fixed URI too long when using "use_alias_path".

Ibochkarev:
### What changed and why

Long friendly URL aliases hit server-side limits (#13815) while the manager capped URI at 255 chars and hid validation errors when URI was collapsed ("Freeze URI").

This PR adds server-side URI limits (`uri_max_length`, `cache_alias_map`, lexicon errors) per maintainer review in #16977, removes the client `maxLength` cap, and shows the URI field when validation fails on a hidden field.

### How to test

1. Friendly URLs on; test deep aliases with `cache_alias_map` on and off.
2. Exceed limits: server error messages, not a blank form.
3. Freeze URI, trigger validation: URI field becomes visible.

### Related issue(s)/PR(s)

Resolves #13815

Supersedes manager/server portions of #16977 for a single merge path.

### Compatibility notes

Manager resource form and alias/index limits. Depends on `uri_max_length` and DB index length.

Stricter server validation for long URIs. Client 255 cap removed. Review alias length policy before merge.

### Test coverage

No new PHPUnit; manual resource save cases.

### Contributors

@smg6511 for #16977 review direction.

### AI tool use


biz87:
## Code Review

### Summary

Removes client-side `maxLength: 255` from the URI field (the `uri` column is TEXT, not limited to 255) and auto-reveals the hidden URI field when validation errors occur. +30/-3.

### Assessment

Correct fix. `hasUriError` handles both array and object error formats. `revealUriField` is reused in two scenarios (client-side and server-side validation). The `failure` override correctly calls `superclass.failure`.

### Verdict

**Approve**

smg6511:
So, this doesn't need new methods to be achieved, just a listener in the URI field config:
```javascript
           listeners: {
                invalid: {
                    fn: function(cmp, msg) {
                        if (cmp.hidden && this.initialized) {
                            cmp.show();
                            // optionally toggle Freeze URI switch here (see considerations below)
                        }
                    },
                    scope: this
                }
            }
```

But, really other considerations need to be taken into account ... @opengeek @Mark-H (like to get your insight on the following):

- If we're going to show/hide this field (instead of just always showing it), isn't it a little odd to toggle the Freeze URI on error when the intent is _not_ to freeze it? I get the general logic of them being paired fields, but it introduces an incongruity in this context.
- The URI value is written whether Freeze is on or off; is that value used when constructing paths to Resources no matter what or is it only used for frozen URIs? If the latter, why are we writing this value in the background in a hidden field?

Ibochkarev:
> So, this doesn't need new methods to be achieved, just a listener in the URI field config:
> 
> ```js
>            listeners: {
>                 invalid: {
>                     fn: function(cmp, msg) {
>                         if (cmp.hidden && this.initialized) {
>                             cmp.show();
>                             // optionally toggle Freeze URI switch here (see considerations below)
>                         }
>                     },
>                     scope: this
>                 }
>             }
> ```
> 
> But, really other considerations need to be taken into account ... @opengeek @Mark-H (like to get your insight on the following):
> 
> * If we're going to show/hide this field (instead of just always showing it), isn't it a little odd to toggle the Freeze URI on error when the intent is _not_ to freeze it? I get the general logic of them being paired fields, but it introduces an incongruity in this context.
> * The URI value is written whether Freeze is on or off; is that value used when constructing paths to Resources no matter what or is it only used for frozen URIs? If the latter, why are we writing this value in the background in a hidden field?

Good call — switched to an invalid listener on the URI field and dropped the extra methods. On error we just show the field; Freeze URI stays untouched.

On freeze: with uri_override off, `modResource::save()` rebuilds the URI from the alias path anyway (`!$this->get('uri_override')`). The hidden input still posts a value, but the server overwrites it unless frozen. Showing the field here is only so you can see the error, not to freeze anything.

Duplicate URIs with freeze off still hit alias, not uri (`Resource/Update.php`). Same as before.

smg6511:
@Ibochkarev - Ivan, thanks for working with me on this. After thinking about this a bit more, a different approach might be better; I put up an alternate PR (#16977). Check it out and let me know what you think...

Ibochkarev:
> @Ibochkarev - Ivan, thanks for working with me on this. After thinking about this a bit more, a different approach might be better; I put up an alternate PR (#16977). Check it out and let me know what you think...

Good work!

Ibochkarev:
Rebased on current `3.x`. This branch keeps only the manager UX from the thread (drop `maxLength`, `invalid` listener to reveal the URI field). Server-side limits and clearer errors are in your #16977 — happy to close this one if you prefer a single PR.

Ibochkarev:
closing https://github.com/modxcms/revolution/pull/16895 in favour of https://github.com/modxcms/revolution/pull/16977

PR #16896, opened 2026-02-25
fix(element): Add label to Property Set dropdown on Properties tab

### LINKED ISSUE #13901: modx 3.x property set dropdown has no label
### Summary
I've only tested this with chunks. On the Properties tab of the chunk the drop down for created property sets has no label and no content.

Ibochkarev:
### What does it do?
Adds a visible label **"Property Sets:"** for the Property Set dropdown on the Properties tab of elements (Chunk, Snippet, Template, etc.). The label is rendered in the toolbar in `modx.grid.element.properties.js` and in the tbar in `modx.panel.property.set.js`, using the existing lexicon key `propertysets`.

### Why is it needed?
The Property Set dropdown had no label, so users could not tell what the control was for. This change makes the UI clearer and fixes the missing label reported in the related issue.

### How to test
1. Edit a Chunk (or any element that has a Properties tab with Property Sets).
2. Open the **Properties** tab.
3. Confirm that the label **"Property Sets:"** appears next to the Property Set dropdown in the toolbar.

### Related issue(s)/PR(s)
Resolves #13901

### DIFF
--- a/manager/assets/modext/widgets/element/modx.grid.element.properties.js
+++ b/manager/assets/modext/widgets/element/modx.grid.element.properties.js
@@ -134,7 +134,7 @@ MODx.grid.ElementProperties = function(config = {}) {
             pressed: true,
             disabled: !MODx.perm.unlock_element_properties,
             scope: this
-        }, '->', {
+        }, '->', _('propertysets') + ': ', {
             xtype: 'modx-combo-property-set',
             id: 'modx-combo-property-set',
             baseParams: {
--- a/manager/assets/modext/widgets/element/modx.panel.property.set.js
+++ b/manager/assets/modext/widgets/element/modx.panel.property.set.js
@@ -68,7 +68,7 @@ MODx.grid.PropertySetProperties = function(config = {}) {
     Ext.applyIf(config, {
         autoHeight: true,
         lockProperties: false,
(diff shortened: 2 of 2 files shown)


biz87:
## Code Review

### Summary

Adds a "Property Sets:" label from the existing `propertysets` lexicon key to the Property Set dropdown on the Properties tab. 2 JS files, +2/-2.

### Verdict

**Approve** — minimal, correct change using existing lexicon.

mkschell (approved):
Agreed, clean, minimal UX improvement. Lexicon key already in use so i18n is covered. Good to merge imo.

Ibochkarev:
@smg6511, hello! Please check

smg6511:
@Ibochkarev - I began looking at this a while back and am looping back to it now. I'm going to send you an alternative very soon (which also covers this in the Property Sets page). It follows the style I began to establish where the labels for these dropdowns appear above the field to conserve horizontal space and give a cleaner look (see the current Lexicons grid).

UPDATE: Because it was easier to do than try to mark up this PR, I created a replacement. Please see #16970. @mkschell - Mike, please take a look there too if you have time.

smg6511:
Closing in favor of #16970.

PR #16899, opened 2026-02-25
Increase drag-n-drop image visibility in package manager

### LINKED ISSUE #15989: [3.0.0-rc1] Image in package manager
### Summary

If you go to the package manager, you can see a pale image on the main background =)

изображение

Tinted the background color to make it easier to see:

изображение

This style that shows the picture is in the following location:

`index.css:26338`

```css
#modx-panel-packages.drag-n-drop::before {
    background: transparent url("../images/restyle/dragndrop.svg");
    ...
}
```

### Environment

- MODX 3.0.0-rc1
- Firefox 96.0.1

Ibochkarev:
### What does it do?
Increases the opacity of the drag-and-drop hint image in the Package Manager from the default (effectively ~0.1) to `0.25` for the `#modx-panel-packages.drag-n-drop::before` pseudo-element in `_build/templates/default/sass/index.scss`. The compiled CSS is produced by `grunt sass` during build; this PR only changes the SCSS source.

### Why is it needed?
With the previous opacity, the dragndrop.svg hint was barely visible and could be mistaken for a visual bug or artifact. Making the hint clearly visible (while still subtle) improves UX and matches the intent of the drag-and-drop zone.

image

### How to test
1. Open the Package Manager in the MODX manager.
2. Ensure the packages list is in a state where the drag-and-drop overlay can appear (e.g. empty or when dragging).
3. Confirm the drag-and-drop hint image is visibly present and looks intentional, not like a bug.

### Related issue(s)/PR(s)
Resolves #15989


biz87:
## Code Review

### Summary

Adds `opacity: 0.25` to the drag-and-drop hint SVG in Package Manager. Single SCSS change, +1/-0.

### Assessment

Minimal CSS change, correct and harmless. Improves visibility of the drag-and-drop hint without being intrusive.

### Verdict

**Approve**

mkschell (approved):
Simple scss/css change, no superfluous whitespace changes, manually verified working.
Thanks!

Ruslan-Aleev:
I would also add the another one SVG, since user can only drag and drop archives, and here the images are in SVG.
Or make the icon a little more neutral, without creating a separate one for the package manager. But that might be unnecessary.

smg6511:
Guys - I'm of the mind that in the normal state there should be no icon shown. I believe the typical pattern out there is to indicate the drop zone when dragging an object over it. The other thing is, and maybe this is what Ruslan's getting at, the icon itself indicates the movement of literal images — if a graphic is to be shown (preferably on an over state) it should communicate the type of content that's droppable. 

I do appreciate the effort to fix what's currently showing — it's a little thing that's bothered me for a while ;-)

mkschell:
@smg6511 Any harm in leaving this visibility improvement in flow to be merged for now, and we get a new Issue logged with the additional thoughts raised above? I think those are excellent points and agree with moving to more common patterns. (I don't mind adding the new Issue if you agree.)

smg6511:
@mkschell - Hey Mike - not gonna block it ;-) Go for it on posting new issue...

PR #16900, opened 2026-02-25
feat(core): add SOCKS5 proxy support for package management and remote providers

### LINKED ISSUE #16815: Add SOCKS5 proxy support for MODX3 package management and remote providers
Problem Description
Currently, MODX3 supports HTTP proxies through system settings (proxy_host, proxy_port, etc.), but it lacks native support for the SOCKS5 protocol. In some environments or geographical regions, users face network restrictions where SOCKS5 (especially with remote DNS resolution) is the only stable way to connect to the official MODX repository or other remote providers.

When using a SOCKS5 proxy with current settings, MODX (via Guzzle or native cURL) treats it as an HTTP proxy, leading to errors like cURL error 56: Proxy CONNECT aborted.

Proposed Solution
I suggest expanding the existing proxy system settings to allow users to specify the proxy type.

Add a new System Setting: proxy_type.

Options: HTTP, SOCKS4, SOCKS5, SOCKS5_HOSTNAME (to support remote DNS resolution).

Update modX::buildHttpClientOptions: Ensure the Guzzle client initialization incorporates the CURLOPT_PROXYTYPE option based on the new setting.

Update modTransportPackage::transferPackage: Ensure the native cURL implementation for downloading transport packages also respects the proxy_type setting.

Implementation Details (Example)
In modX.php, the buildHttpClientOptions method should be updated to handle the proxy protocol:

```
PHP

// Existing logic in buildHttpClientOptions
if (!empty($proxyHost)) {
    // ... logic to build proxy string ...
    $opts['proxy'] = $proxy_str; // For Guzzle, this could be "socks5h://..." 
    
    // Explicitly set cURL options for broader compatibility
    $opts['curl'] = [
        CURLOPT_PROXYTYPE => $this->getOption('proxy_type', null, CURLPROXY_HTTP),
        CURLOPT_HTTPPROXYTUNNEL => true
    ];
}
```
Benefit
This change will make MODX3 more resilient and flexible in complex network environments, ensuring that users can always download and update extras regardless of local network limitations or blocks.

Ibochkarev:
### What changed and why

Sites behind SOCKS or HTTP proxies could not reach package providers or RSS feeds using existing `proxy_*` settings alone (#16900).

This PR routes package downloads and remote provider requests through SOCKS5/HTTP proxy configuration from system settings.

### How to test

1. Set `proxy_host`, `proxy_port`, and related proxy settings.
2. Refresh Package Management provider list.
3. Load a dashboard RSS widget that pulls a remote feed.

### Related issue(s)/PR(s)

Resolves #16900

### Compatibility notes

Requires proxy settings and network path to provider. Guzzle/cURL behavior depends on PHP build.

No public API signature changes. Opt-in via proxy settings. Patch-safe.

### Test coverage

No automated tests; manual proxy environment.

### Contributors

Discussion on #16900.

### AI tool use

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -1201,6 +1201,15 @@ $settings['proxy_auth_type']->fromArray([
   'area' => 'proxy',
   'editedon' => null,
 ], '', true, true);
+$settings['proxy_type'] = $xpdo->newObject(modSystemSetting::class);
+$settings['proxy_type']->fromArray([
+  'key' => 'proxy_type',
+  'value' => 'HTTP',
+  'xtype' => 'textfield',
+  'namespace' => 'core',
+  'area' => 'proxy',
+  'editedon' => null,
+], '', true, true);
 $settings['proxy_host'] = $xpdo->newObject(modSystemSetting::class);
 $settings['proxy_host']->fromArray([
   'key' => 'proxy_host',
--- a/_build/test/Tests/Model/modXTest.php
+++ b/_build/test/Tests/Model/modXTest.php
@@ -416,7 +416,7 @@ class modXTest extends MODxTestCase
             ['socks5', 'SOCKS5'],
             ['SOCKS4', 'SOCKS4'],
             ['SOCKS5_HOSTNAME', 'SOCKS5_HOSTNAME'],
-            ['socks5h', 'HTTP'],
+            ['socks5h', 'SOCKS5_HOSTNAME'],
             ['invalid', 'HTTP'],
             ['', 'HTTP'],
         ];
--- a/_build/test/Tests/Model/modXTest.php
+++ b/_build/test/Tests/Model/modXTest.php
@@ -109,6 +109,8 @@ class modXTest extends MODxTestCase
         $this->modx->placeholders = [];
         $this->modx->resourceMap = [[1]];
         unset($this->modx->contexts['custom']);
+        $this->modx->setOption('proxy_host', '');
+        $this->modx->setOption('proxy_type', 'HTTP');
     }
     /**
      * Test getting the modCacheManager instance.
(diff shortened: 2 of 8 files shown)


biz87:
## Code Review

### Summary

Adds SOCKS5/SOCKS4 proxy support for Package Manager, remote providers, and dashboard feed. Introduces `proxy_type` system setting and `modX::getProxyType()` / `modX::getProxyUrl()` methods, centralizing proxy logic that was previously duplicated across three files. 8 files, +189/-43. Includes unit tests.

### Issues

**`modTransportPackage.php:477-483` — HTTP proxy credentials are encoded twice.**
`getProxyUrl()` already includes `rawurlencode(user):rawurlencode(pass)@host` in the URL. But for HTTP proxies, `CURLOPT_PROXYUSERPWD` is set with another `rawurlencode()` call on the same values. This double-encoding will break authentication for passwords containing special characters (e.g. `p@ss` becomes `p%2540ss` instead of `p%40ss`).

**`Feed.php:148` and `modTransportPackage.php:472` — `CURLOPT_HTTPPROXYTUNNEL` is set unconditionally for all proxy types.**
HTTP CONNECT tunneling is appropriate for HTTP proxies with HTTPS targets but is unnecessary for SOCKS proxies and may cause issues with some implementations. Consider setting it only when `proxy_type === 'HTTP'`.

### Suggestions

- `proxy_type` is created as a `textfield`. For a fixed set of values (HTTP, SOCKS4, SOCKS5, SOCKS5_HOSTNAME), a dropdown/combobox xtype would be more user-friendly and prevent invalid input.

- The test for `socks5h` → `HTTP` (fallback) suggests `socks5h` is not accepted as input. Since it's a standard curl notation for SOCKS5 with remote DNS, it might be worth mapping it to `SOCKS5_HOSTNAME` instead of falling back to HTTP.

### Assessment

Good refactoring — proxy logic is centralized, tests are solid. The double-encoding issue for HTTP proxy credentials is a real bug that will affect users with special characters in passwords. PHPStan clean.

### Verdict

**Request changes** — double `rawurlencode` for HTTP proxy credentials will break authentication.

PR #16901, opened 2026-02-25
feat(workspace): add logManagerAction for package install/upgrade/remove

### LINKED ISSUE #16453: Log Workspace Actions
### Summary
Log actions in the package installer

### Why is it needed?
We currently log all sorts of individual elements and setting adjustments, but there is no logged action when someone updates, installs or removes a package which can have major sweeping changes on a website. 

### Suggested solution(s)
Log manager actions for update / install / remove on packages.

Ibochkarev:
### What changed and why

Package install, upgrade, remove, and uninstall did not appear in Manager Log, so audits of workspace actions were incomplete (#16453).

This PR calls `logManagerAction()` from Install, Remove, Uninstall, Update, and Version/Remove processors. Uses `getPrimaryKey()` for signature. Fixes Uninstall lexicon placeholders (array instead of `sprintf()`).

### How to test

1. Install → log shows `package_install`.
2. Upgrade → `package_upgrade`.
3. Remove → `package_remove`.
4. Uninstall → no errors; log entry present.
5. Manager Log lists each action with correct package name.

### Related issue(s)/PR(s)

Resolves #16453

### Compatibility notes

Manager Log and workspace processors only.

No API changes. New log entries only. Patch-safe.

### Test coverage

No new PHPUnit; manual workspace actions above.

### Contributors

None beyond author.

### AI tool use

### DIFF
--- a/core/src/Revolution/Processors/Workspace/Packages/Install.php
+++ b/core/src/Revolution/Processors/Workspace/Packages/Install.php
@@ -86,10 +86,14 @@ class Install extends Processor
         $this->modx->log(modX::LOG_LEVEL_WARN, $msg);
         $this->modx->log(modX::LOG_LEVEL_INFO, 'COMPLETED');
 
+        $isUpgrade = $this->package->previousVersionInstalled();
         $this->modx->invokeEvent('OnPackageInstall', [
             'package' => $this->package,
-            'action' => $this->package->previousVersionInstalled() ? xPDOTransport::ACTION_UPGRADE : xPDOTransport::ACTION_INSTALL,
+            'action' => $isUpgrade ? xPDOTransport::ACTION_UPGRADE : xPDOTransport::ACTION_INSTALL,
         ]);
+
+        $action = $isUpgrade ? 'package_upgrade' : 'package_install';
+        $this->modx->logManagerAction($action, modTransportPackage::class, $this->package->getPrimaryKey());
 
         return $this->success($msg);
     }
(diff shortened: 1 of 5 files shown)


biz87:
## Code Review

### Summary

Adds `logManagerAction()` calls for package install, upgrade, remove, and uninstall operations. Also fixes a bug where `sprintf` was used instead of `lexicon()` with placeholders in Uninstall, and replaces incorrect `get('id')` with `getPrimaryKey()` since `modTransportPackage` uses `signature` as its primary key. 5 files, +22/-5.

### Issues

**`Install.php` — logging is done inline, while `Remove`, `Uninstall`, `Update`, and `Version/Remove` use a dedicated `logManagerAction()` method.**
For consistency, Install should also use a method:
```php
public function logManagerAction()
{
    $action = $this->package->previousVersionInstalled() ? 'package_upgrade' : 'package_install';
    $this->modx->logManagerAction($action, modTransportPackage::class, $this->package->getPrimaryKey());
}
```

**`Remove.php` — `logManagerAction()` is called after removing transport files but before `cleanup()` (which fires the event and returns success).** If zip/directory removal partially fails, the action is still logged. Consider placing the log call right after the successful `removePackage()` or inside `cleanup()` before the event — similar to how Install.php places the log after the event. Non-blocking.

### Suggestions

- `Uninstall.php:82` — The `sprintf` → `lexicon()` fix is correct. Worth verifying that the `package_err_uninstall` lexicon entry actually contains the `[[+signature]]` placeholder.

### Assessment

Good targeted fixes. The `get('id')` → `getPrimaryKey()` change is important since `modTransportPackage` uses `signature` as PK, meaning `get('id')` would return `null`. The `sprintf` fix resolves a real bug where the placeholder array was silently ignored. PHPStan clean across all 5 files.

### Verdict

**Approve** — consistency and ordering remarks are optional.

rthrash (approved):
This worked exactly as described for install/uninstall/remove. I didn't have an Extra that needed to be upgraded installed, so was not able to test that. Great update that provides excellent additional information. 

Image

Note: the bottom two log entries (with "unknown") were the result of running through the same steps as the top entries.

PR #16902, opened 2026-02-25
feat(search): include Template Variable values in manager search

### LINKED ISSUE #16703: Include the values of TVs when using manager's search functionality
### Summary
I'm not sure if this has already been discussed, but I often run into the issue that I can't search for TVs in the manager search bar.

### Why is it needed?
It would greatly improve the user experience, especially since TVs are one of the main selling points of MODX.
Right now, I just do it inconveniently directly in the database.

### Suggested solution(s)
~

### Related issue(s)/PR(s)
~

Ibochkarev:
### What changed and why

Manager search indexed resource fields but not Template Variable values. Content stored only in TVs was invisible to search (#16703).

This PR includes TV values in manager search results for resources.

### How to test

1. Create a resource with a distinctive string only in a TV (not pagetitle/content).
2. Run manager search for that string.
3. Confirm the resource appears in results.

### Related issue(s)/PR(s)

Resolves #16703

### Compatibility notes

Manager search only. May increase query work on large sites with many TVs.

No API changes. Broader search results only. Patch-safe.

### Test coverage

No automated tests in PR; manual search case above.

### Contributors

None beyond author.

### AI tool use

### DIFF
--- a/core/src/Revolution/Processors/Search/Search.php
+++ b/core/src/Revolution/Processors/Search/Search.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -10,7 +11,6 @@
 
 namespace MODX\Revolution\Processors\Search;
 
-
 use MODX\Revolution\modChunk;
 use MODX\Revolution\modContext;
 use MODX\Revolution\modElement;
@@ -20,6 +20,7 @@ use MODX\Revolution\modResource;
 use MODX\Revolution\modSnippet;
 use MODX\Revolution\modTemplate;
 use MODX\Revolution\modTemplateVar;
+use MODX\Revolution\modTemplateVarResource;
 use MODX\Revolution\modUser;
 use MODX\Revolution\modUserProfile;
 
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

### Summary

Adds TV value search to manager quick search, refactors `searchResources()` into 5 focused methods, and implements relevance-based sorting (exact pagetitle > starts with > other fields > content > TV match > newest). Single file, +121/-33.

### Issues

**`Search.php:131` — `$this->query` is used in LIKE without escaping wildcard characters.**
If a user searches for a string containing `%` or `_`, they are interpreted as LIKE wildcards. This is not SQL injection (values are quoted by xPDO), but it affects search correctness. The same issue exists in the original code, but since the TV query is new, it's worth fixing at least there:
```php
$escaped = addcslashes($this->query, '%_');
$c->where(['value:LIKE' => '%' . $escaped . '%']);
```
Non-blocking — this is a pre-existing issue.

**`Search.php:185-196` — Sort expressions duplicate field logic from `buildResourceSearchCriteria()`.**
If a new search field is added later, it must be updated in both places. Not a bug, but a maintenance concern. A comment linking these two locations would help.

### Suggestions

- `Search.php:132` — The limit `getMaxResults() * 2` for TV query is a magic number. A single resource can have many TVs, so for sites with heavy TV usage this may miss results. Consider `GROUP BY contentid` in the query to limit by unique resource IDs instead.

- `Search.php:179` — Parameter `$c` has PHPDoc type but no native type hint, unlike the other new methods.

### Assessment

Clean refactoring — the monolithic method is split into logical parts, each with PHPDoc and return types. TV search uses a separate query with deduplication and limit to prevent oversized IN clauses. Relevance sorting is a solid UX improvement. All queries use xPDO criteria API, `::class` constants, and `getIterator`. PHPStan clean.

### Verdict

**Approve** — issues are non-blocking, PR is ready to merge.

halftrainedharry:
> * when "Search in content" is enabled in manager search

It's a bit counterintuitive that the `quick_search_in_content` setting is now also used to search in TVs.

> * exact pagetitle match → pagetitle starts with query

So if the pagetitle contains the search term but not at the beginning, then this resource is not at the top of the result-list. Why?

> * Limits TV-matched resource IDs to `getMaxResults() * 2`

For what reason is a multiplying factor of 2 used here?

---

* This PR doesn't handle the **default values** of TVs. If the search term is the default value of a TV, the resources that use that default value are not in the result.

* In my opinion, it would be cleaner and faster to use a LEFT JOIN for the TV table, than to run 2 queries (the first just to get some resource-IDs of matching TV values).

PR #16903, opened 2026-02-25
fix(media): only call exif_read_data for JPEG/TIFF

### LINKED ISSUE #16420: Handling .ico files in media sources

Ibochkarev:
### What does it do?
Call `exif_read_data()` only for file extensions that support EXIF in PHP (jpg, jpeg, tiff, tif). Adds `isExifSupportedExtension()` helper so .ico and other image types no longer trigger "File not supported" warnings in media source previews.

### Why is it needed?
`exif_read_data()` in PHP only supports JPEG and TIFF; calling it on .ico, PNG, WebP, etc. causes warnings and pollutes logs (Issue #16420).

### How to test
Open Media Source in the manager, browse to a folder containing .ico (or PNG/WebP) and confirm no PHP warning `exif_read_data(...): File not supported` appears; preview and rotation for JPEG/TIFF should still work.

### Related issue(s)/PR(s)
Resolves #16420

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2300,7 +2300,7 @@ QTIP;
             } else {
                 $absolute_path = $path;
             }
-            if (function_exists('exif_read_data')) {
+            if (function_exists('exif_read_data') && $this->isExifSupportedExtension($ext)) {
                 $exif = @exif_read_data($absolute_path);
                 if (!empty($exif) && array_key_exists('Orientation', $exif) && $exif['Orientation'] >= 5) {
                     // This image was rotated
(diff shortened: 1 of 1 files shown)


biz87:
## Code Review

### Summary

Fix for spurious PHP warnings from `exif_read_data()` when browsing .ico, .png, .webp files in the media manager. Adds an `isExifSupportedExtension()` helper that only allows JPEG/TIFF through. Single file change, +13/-1.

### Suggestions

- `modMediaSource.php:2356` — The method is `private`, while all neighboring methods (`buildManagerImagePreview`, `getImageDimensions`) are `protected`. Since `modMediaSource` is subclassed by various media source types, `protected` would be more consistent and extensible.

- Same method — native type hints (`string $ext` and `: bool`) could be added since PHPDoc already declares them.

### Assessment

Clean minimal fix. The extension list accurately matches what PHP's `exif_read_data()` supports. Existing JPEG/TIFF behavior is unaffected. PHPStan clean, no new PHPCS violations.

### Verdict

**Approve** — suggestions are optional, PR is ready to merge.

mkschell (approved):
Clean bug fix. Guard correctly limits exif_read_data to JPEG/TIFF per PHP docs. Good to merge.

Mark-H (approved):

PR #16911, opened 2026-02-26
Fix package metadata format

matdave:
### What does it do?
Adjust Processors\Workspace\Packages\GetList class's function getPackageMeta

### Why is it needed?
At some point in the last 14-17 years, the provider started responding with a different model than what was expected. Rather than changing the provider and potentially breaking a lot more areas that work with the metadata field, I propose to change the processor. 

### How to test
Open your package management page `/manager/?a=workspaces` and try to expand an extra that was installed with the provider. It will expose an empty field. 

### Related issue(s)/PR(s)
This was pointed out in Community Slack

### DIFF
--- a/core/src/Revolution/Processors/Workspace/Packages/GetList.php
+++ b/core/src/Revolution/Processors/Workspace/Packages/GetList.php
@@ -170,9 +170,9 @@ class GetList extends GetListProcessor
     {
         $metadata = $packageArray['metadata'];
         if (!empty($metadata)) {
-            foreach ($metadata as $k => $v) {
-                if ($k === 'description') {
-                    $packageArray['readme'] = nl2br($v);
+            foreach ($metadata as $key => $value) {
+                if ($key === 'description') {
+                    $packageArray['readme'] = nl2br($value);
                     break;
                 }
             }
--- a/core/src/Revolution/Processors/Workspace/Packages/GetList.php
+++ b/core/src/Revolution/Processors/Workspace/Packages/GetList.php
@@ -170,9 +170,9 @@ class GetList extends GetListProcessor
     {
         $metadata = $packageArray['metadata'];
         if (!empty($metadata)) {
-            foreach ($metadata as $row) {
-                if (!empty($row['name']) && $row['name'] === 'description') {
-                    $packageArray['readme'] = nl2br($row['text']);
+            foreach ($metadata as $k => $v) {
+                if ($k === 'description') {
+                    $packageArray['readme'] = nl2br($v);
                     break;
                 }
             }


matdave:
@Mark-H I don't know if you are responding from your provider differently. I noticed at least one extra that was empty, so I assume it's not using the weird method the GetList processor is expecting.

core/src/Revolution/Processors/Workspace/Packages/GetList.php
```diff
@@ -170,9 +170,9 @@ public function getPackageMeta(modTransportPackage $package, array $packageArray
     {
         $metadata = $packageArray['metadata'];
         if (!empty($metadata)) {
-            foreach ($metadata as $row) {
-                if (!empty($row['name']) && $row['name'] === 'description') {
-                    $packageArray['readme'] = nl2br($row['text']);
+            foreach ($metadata as $k => $v) {
```
JoshuaLuckers: Even although it might be obvious what the variables mean I would still prefer them to be a bit more descriptive.

Ibochkarev: `$k/$v` -> `$metaKey/$metaValue`

core/src/Revolution/Processors/Workspace/Packages/GetList.php
```diff
@@ -170,9 +170,9 @@ public function getPackageMeta(modTransportPackage $package, array $packageArray
...
+                if ($k === 'description') {
+                    $packageArray['readme'] = nl2br($v);
```
smg6511: Basically like Joshua and Ivan suggested:
```suggestion
            foreach ($metadata as $key => $value) {
                if ($key === 'description') {
                    $packageArray['readme'] = nl2br($value);
```

smg6511 (changes requested):
This fix works as expected and will be good to go with the requested change.

Another couple of changes to clean up the display within the grid would be nice too:
In `index.scss` L2335, change def to:
```css
.package-readme {
  padding: 0 11px 8px 11px;
  font-style: italic;
}
```

And in `package.grid.js` L12, change XTemplate def to:
```javascript
'<div class="package-readme">{readme}</div>'
```

On the last item, the original parent tag was `<p>`, which causes malformed html when a readme contains `<p>` tags as well.

core/src/Revolution/Processors/Workspace/Packages/GetList.php
smg6511: Guys, I'm resolving these comments since I implemented your comment into a code suggestion below ;-)

JoshuaLuckers (approved):

Ibochkarev (approved):
👍

smg6511:
@matdave - Nudge ;-)

smg6511 (approved):
This is good to go. We should make the suggested CSS changes in a separate PR if not here.

PR #16919, opened 2026-03-08
Use FQCN in modResource and modUser

Ibochkarev:
### What does it do?

Replaces two remaining short `mod*` class name strings in core xPDO calls with FQCN:

- `modResource::isPreviewable()` — `getObject('modResource', …)` → `getObject(modResource::class, …)`
- `modUser::getResourceGroups()` — `loadAttributes(['modAccessResourceGroup'], …)` → `loadAttributes([modAccessResourceGroup::class], …)`

Earlier iterations that changed the schema or added `modX::loadClass()` normalization were reverted per review feedback.

### Why is it needed?

MODX 3.0 deprecates short `mod*` class names in favor of PSR-4 FQCN. These two core call sites still passed bare strings into xPDO, which could trigger deprecation notices in the profiler. Fixing them at the call site follows the intended v3.0 migration path without masking warnings globally.

### How to test

1. Enable the profiler or deprecation log.
2. Exercise resource preview checks (`modResource::isPreviewable()`).
3. Exercise user resource group loading (`modUser::getResourceGroups()`, used via `modResourceGroup::hasAccess()`).
4. Confirm no v3.0 short-class deprecation warnings from these paths.

### Related issue(s)/PR(s)

None.

### DIFF
--- a/core/src/Revolution/modResource.php
+++ b/core/src/Revolution/modResource.php
@@ -1691,7 +1691,7 @@ class modResource extends modAccessibleSimpleObject implements modResourceInterf
             return false;
         }
         $targetIsSelf = $sourceId === null || (int)$targetId === (int)$sourceId;
-        if ($doc = $this->xpdo->getObject('modResource', $targetId)) {
+        if ($doc = $this->xpdo->getObject(modResource::class, $targetId)) {
             if (!$doc->deleted) {
                 return true;
             }
--- a/core/src/Revolution/modUser.php
+++ b/core/src/Revolution/modUser.php
@@ -623,7 +623,7 @@ class modUser extends modPrincipal
         if (isset($_SESSION["modx.user.{$id}.resourceGroups"][$ctx])) {
             $resourceGroups = $_SESSION["modx.user.{$id}.resourceGroups"][$ctx];
         } else {
-            $this->loadAttributes(['modAccessResourceGroup'], $ctx, true);
+            $this->loadAttributes([modAccessResourceGroup::class], $ctx, true);
             if (isset($_SESSION["modx.user.{$id}.resourceGroups"][$ctx])) {
                 $resourceGroups = $_SESSION["modx.user.{$id}.resourceGroups"][$ctx];
             }
--- a/core/model/schema/modx.mysql.schema.xml
+++ b/core/model/schema/modx.mysql.schema.xml
@@ -11,7 +11,7 @@
  -->
 <!-- The following xPDO model represents an object-relational map structure of MODX -->
 <model package="MODX\Revolution\" baseClass="xPDO\Om\xPDOObject" platform="mysql" defaultEngine="InnoDB" phpdoc-package="MODX\Revolution" version="3.0">
-    <object class="modAccess" extends="xPDO\Om\xPDOSimpleObject">
+    <object class="MODX\Revolution\modAccess" extends="xPDO\Om\xPDOSimpleObject">
         <field key="target" dbtype="varchar" precision="100" phptype="string" null="false" default="" index="fk" />
         <field key="principal_class" dbtype="varchar" precision="100" phptype="string" null="false" default="MODX\Revolution\modPrincipal" index="index" />
         <field key="principal" dbtype="int" precision="10" attributes="unsigned" phptype="integer" null="false" default="0" index="fk" />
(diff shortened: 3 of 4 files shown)


Mark-H:
I'm confused about the schema changes - all `object`s are already in the `\MODX\Revolution` package so this is already getting prepended as namespace when the model is built. It seems like this would cause that it to get doubled up. 

Any deprecation warnings should come from specific _usages_ using the short name, not the core schema. But as there's no issue linked or deprecation warning shown, it's unclear what specifically we're fixing. 

The loadClass fix I could get behind but would want to test with third-party extras that have (perhaps unwisely so) used the `mod` prefix in their models.

smg6511:
On the `loadClass` change: Isn't the whole idea of having these deprecation warnings to remind (badger) us to adopt the appropriate replacement method? I don't think it's a fix to mute those.

smg6511:
@Ibochkarev - Hey, just wanted to get your response to Mark's and my comments above and whether you want to continue to pursue this change. Thx ;-)

Ibochkarev:
Thanks @Mark-H and @smg6511 — really helpful feedback, and sorry for the slow reply.

After re-reading your comments I agree the original approach was off:

**Schema FQCN** — you're right. Objects already live in the `MODX\Revolution\` package, and xPDO prepends that namespace when the model is built. Putting FQCN in `class="..."` risks doubling the namespace. I've reverted all schema changes.

**`loadClass()` normalization** — also reverted. @smg6511 is right that those deprecation notices are intentional; muting them in `loadClass()` would work against the v3.0 migration path rather than help it.

**What I kept instead** — fixing actual short-name usages in core code. A repo-wide search turned up two remaining call sites that still passed bare `mod*` strings into xPDO:

- `modResource::isPreviewable()` — `getObject('modResource', …)` → `getObject(modResource::class, …)`
- `modUser::getResourceGroups()` — `loadAttributes(['modAccessResourceGroup'], …)` → `loadAttributes([modAccessResourceGroup::class], …)`

Pushed in `7c7b3df9e`.

Happy to dig further if you know of other profiler warnings I missed — I didn't have a concrete stack trace in the original PR, which was part of the problem. If there are more, I'd rather fix them at the call site than in schema or `loadClass()`.

@Mark-H — on third-party extras using the `mod` prefix: with the `loadClass()` shim gone, those should behave exactly as before. Worth a quick smoke test if anyone has a known extra that does this.

Let me know if this direction works or if you'd prefer closing the PR and opening a smaller one with just the two call-site fixes.

core/src/Revolution/modResource.php
```diff
@@ -1691,7 +1691,7 @@ public function canPreviewResource($targetId, $sourceId = null): bool
             return false;
         }
         $targetIsSelf = $sourceId === null || (int)$targetId === (int)$sourceId;
-        if ($doc = $this->xpdo->getObject('modResource', $targetId)) {
+        if ($doc = $this->xpdo->getObject(modResource::class, $targetId)) {
```
smg6511, line 1694: Although this works just fine, it may be preferable for clarity to use `self::class` since we're in the class this is referencing.

core/src/Revolution/modUser.php
```diff
@@ -623,7 +623,7 @@ public function getResourceGroups($ctx = '')
         if (isset($_SESSION["modx.user.{$id}.resourceGroups"][$ctx])) {
             $resourceGroups = $_SESSION["modx.user.{$id}.resourceGroups"][$ctx];
         } else {
-            $this->loadAttributes(['modAccessResourceGroup'], $ctx, true);
+            $this->loadAttributes([modAccessResourceGroup::class], $ctx, true);
```
smg6511, line 626: To be sure, you will want to add a `use` statement pointing to this class at the top of this file.

But, further (and maybe @opengeek or @Mark-H can answer this) it appears that this method may no longer be relevant in 3.x. I couldn't find usage of this anywhere.

smg6511 (commented):
@Ibochkarev My first comment is optional (but maybe preferrable), but it'd be good to know more re the `modUser->getResourceGroups` method — unless it's used behind the scenes somehow, we may be able to just remove that (?).

smg6511:
> more re the `modUser->getResourceGroups` method

Answering my own question, this is used in one place in `modResourceGroup->hasAccess`.

smg6511 (approved):
Ok, we're good here!

opengeek:
Please update this PR title/description to reflect the actual state of the PR.

Ibochkarev:
@opengeek Updated the PR description to match the final scope after review. The title was already updated by @smg6511. This PR now only contains the two call-site FQCN fixes in `modResource` and `modUser`; schema and `loadClass()` changes were reverted.

opengeek (approved):

PR #16921, opened 2026-03-10
fix(setup): improve PHP version error message and require mbstring extension

Ibochkarev:
### What does it do?
- **bootstrap.php**: Updates the PHP version requirement error message: title to "PHP version is not supported", description to "Your PHP version (%s) is no longer supported. Please upgrade to PHP %s or higher to install MODX."
- **requirements.php**: Adds `mbstring` to `MODX_REQUIRED_EXTENSIONS` so the installer checks for the extension before proceeding.

### Why is it needed?
- Clearer, more neutral wording when the user's PHP version is below the minimum.
- MODX relies on multibyte string handling; requiring `mbstring` at setup prevents runtime errors on systems where it is missing.

### How to test
1. Run the installer with PHP &lt; 7.4 (or the configured minimum): the new PHP version message should appear.
2. Disable the `mbstring` extension (or use a PHP build without it) and run the installer: the requirements screen should list "MODX requires the PHP mbstring extension".

### Related issue(s)/PR(s)
None.

### DIFF
--- a/setup/provisioner/bootstrap.php
+++ b/setup/provisioner/bootstrap.php
@@ -37,9 +37,9 @@ require_once __DIR__ . DIRECTORY_SEPARATOR . 'requirements.php';
 $phpVersionSatisfiesRequirement = version_compare(PHP_VERSION, MODX_MINIMUM_REQUIRED_PHP_VERSION, '>=');
 if (!$phpVersionSatisfiesRequirement) {
     $unsatisfiedRequirementsErrors[] = [
-        'title' => 'Wrong PHP Version!',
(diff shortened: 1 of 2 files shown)


mkschell (changes requested):
The PHP version error message is an improvement. Changes to PHP extension requirements are non-trivial imo. Could you please separate these concerns, and open an Issue to discuss changing the PHP extension requirements?

Ibochkarev:
> The PHP version error message is an improvement. Changes to PHP extension requirements are non-trivial imo. Could you please separate these concerns, and open an Issue to discuss changing the PHP extension requirements?

Ok

PR #16923, opened 2026-03-14
fix(setup): require mbstring extension in installer

### LINKED ISSUE #14990: MODX3 Setup:  Uncaught Error: Call to undefined function mb_split()
### Summary
When trying out the new MODX3 Alpha 2 I get the following error message on the set-up screenn:

`Fatal error: Uncaught Error: Call to undefined function mb_split() in /home/mytest/public_html/core/vendor/smarty/smarty/libs/plugins/shared.mb_str_replace.php:47 Stack trace: #0 /home/mytest/public_html/core/vendor/smarty/smarty/libs/plugins/modifier.replace.php(34): smarty_mb_str_replace('[[+current_year...', '2020', '&copy; 2005-[[+...') #1 /home/mytest/public_html/core/cache/setup/smarty/f408b021d2b46489800138ede826932f9da88e74_0.file.footer.tpl.php(37): smarty_modifier_replace('&copy; 2005-[[+...', '[[+current_year...', '2020') #2 /home/mytest/public_html/core/vendor/smarty/smarty/libs/sysplugins/smarty_template_resource_base.php(123): content_5e37fe9d7cd3c4_44651287(Object(Smarty_Internal_Template)) #3 /home/mytest/public_html/core/vendor/smarty/smarty/libs/sysplugins/smarty_template_compiled.php(114): Smarty_Template_Resource_Base->getRenderedTemplateCode(Object(Smarty_Internal_Template)) #4 /home/mytest/public_html/core/vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php(216): Smarty_Template_C in /home/mytest/public_html/core/vendor/smarty/smarty/libs/plugins/shared.mb_str_replace.php on line 47`

### Step to reproduce
This appears to be because my server defaults to php v7.0.x which I know is no longer supported but couldn't we have a check in place for this to warn the user instead as if anyone new is trying out MODX for the first time, it's not the most welcome of starts.

### Observed behavior
Error as above.

### Expected behavior
PHP < 7.2 I should see a warning message that older versions are not supported and I should upgrade.
PHP >= 7.2 I should see the lovely new installation page.

### Environment
Ubuntu server
Apache 2.x
Php 7.0.x

This works correctly when I switch to PHP 7.2.

Thanks!

Ibochkarev:
### What does it do?
Adds `mbstring` to `MODX_REQUIRED_EXTENSIONS` in the setup provisioner so the installer checks for the extension before proceeding. If mbstring is missing, the requirements screen lists it and blocks installation.

### Why is it needed?
MODX and Smarty use multibyte string functions (e.g. `mb_split()`). Without the mbstring extension, the setup fails with a fatal error (e.g. "Call to undefined function mb_split()") instead of a clear requirement message. Checking at installer time prevents this and matches the expected behaviour described in #14990.

### How to test
1. Disable the `mbstring` extension (or use a PHP build without it).
2. Run the MODX installer.
3. Confirm the requirements screen lists "MODX requires the PHP mbstring extension" and setup does not proceed until it is enabled.

### Related issue(s)/PR(s)
Resolves #14990.  
Split from #16921 per review (extension requirements as a separate change for discussion).

### DIFF
--- a/setup/provisioner/requirements.php
+++ b/setup/provisioner/requirements.php
@@ -16,6 +16,7 @@ define('MODX_REQUIRED_EXTENSIONS', [
     "fileinfo",
     "gd",
     "json",
+    "mbstring",
     "pdo",
     "simplexml",
     "xml",


mkschell (commented):
Thanks for splitting this out.

I agree there is a real issue here, but I'm not comfortable changing environment requirements in a patch release, and ideally not even in a minor release unless we've clearly signaled that ahead of time.

In general, for open-source maintenance releases, the safest path is to keep them focused on backward-compatible bug fixes and avoid changes that alter deployment/runtime expectations for existing users. Requiring an additional PHP extension moves us into "environment contract changed" territory, which has a wider blast radius than the original bug.

For the issue as stated, I think the better fix is to remove our use of Smarty's |replace filter in the setup/templates where it currently appears. As far as I can tell, there are only 3 locations to address, and that would solve the installer/setup failure without introducing a new hard requirement for all patch adopters.

That still leaves us with a potentially hampered |replace filter for userland code on environments that do not have mbstring installed. Because of that, I think it still makes sense to keep mbstring in the suggest section rather than promoting it to a required extension, and we should see if there's an opportunity to improve the docs around that behavior/limitation.

Separately, we're also reviewing our dependency on Smarty in the system overall for future releases, so I'd rather handle broader requirement changes in that larger context than through a patch-level installer change.

Given that, I'm inclined not to merge this PR as-is, and instead pursue the narrower fix of removing the |replace usage in setup.

mkschell:
Closing this and I'll update #14990 to clarify the direction.

PR #16924, opened 2026-03-14
fix(dashboard): update widget only stable releases, opt-in prereleases, To Download button

### LINKED ISSUE #15483: Update Dashboard widget shows unreleased (and unstable versions)
### Summary
The update widget does not show the correct versions to update to. 

### Step to reproduce
Either install a previous 3.x version, or mock your local version by changing the local values in `/core/docs/version.inc.php`, e.g. to 3.0.0-alpha2. See that the widget is showing you an update, although the latest published release is -alpha2. 

### Observed behavior
- the widget shows you that an update is available to version 3.0.0-alpha3. But this is the current build version, and has not yet been released (at the time of writing here). Therefore also the click on the update button (wrong label!), which leads you to the download page, is kind of useless, as it does not offer a zip to download for the version (you would need the nightly build). 
- as the update information is gathered from the build file in the 3.x branch, it will _always_ show you that an update is available - unless you are using a git installation, which potentially - if you pulled recently, contains the same version information. In all other cases, mainly normal installations, the repo buildfile will always be ahead of the recently released tarballs...
- the widget will show you core updates available regardless of the patch level here. Most people will only be interested in stable versions, so only those updates should be shown. 

### Expected behavior
- widget should get its information for released versions (exception: if you have a git based install, it may be ok to show the nightly upcoming version)
- widget should show you only stable versions (with a system setting allowing you to optin to unstable versions if you want)
- as long as no update is possible inplace (probably never), the button should be renamed to "To Download" or something. We could also directly offer the download for the correct tarball, because those links are constructable easily. 

### Environment
MODX version: 3.x (git)

Ibochkarev:
### What does it do?
- Adds system setting `updates_show_prereleases` (default `false`) so the Updates dashboard widget can optionally include prerelease/unstable MODX core versions. When disabled, only stable releases are requested from the Sentinel API.
- Updates `SoftwareUpdate/GetList` to pass `prereleases` from the setting into the API request instead of a hardcoded `0`.
- Updates the dashboard widget cache key to include the prereleases flag (`modx-core` vs `modx-core-prereleases`) so changing the setting yields fresh data.
- Renames the core update action button from "Download" to "To Download" via new lexicon key `updates_to_download` in the dashboard (en only), since in-place update is not supported and the action leads to downloading the release.

### Why is it needed?
Addresses [issue #15483](https://github.com/modxcms/revolution/issues/15483): the widget was showing unreleased/unstable versions as available updates, and the button label was misleading. The fix keeps stable-only by default, adds an opt-in for prereleases, and clarifies the button as "To Download".

### How to test
1. Open Dashboard and ensure the Updates widget is visible.
2. With default settings, confirm only stable release(s) appear when an update is available; button label is "To Download".
3. In System Settings, enable "Show Unstable Core Updates in Dashboard" (`updates_show_prereleases`). Reload the dashboard and confirm the widget can show prerelease versions when the API returns them.
4. Disable the setting again and confirm the widget shows only stable again (cache key change should show correct data after reload).

### Related issue(s)/PR(s)
Resolves #15483

### DIFF
--- a/_build/data/transport.core.system_settings.php
+++ b/_build/data/transport.core.system_settings.php
@@ -2194,5 +2194,14 @@ $settings['static_elements_html_extension']->fromArray([
     'area' => 'static_elements',
     'editedon' => null,
 ], '', true, true);
+$settings['updates_show_prereleases'] = $xpdo->newObject(modSystemSetting::class);
+$settings['updates_show_prereleases']->fromArray([
+    'key' => 'updates_show_prereleases',
+    'value' => false,
+    'xtype' => 'combo-boolean',
+    'namespace' => 'core',
+    'area' => 'system',
+    'editedon' => null,
+], '', true, true);
 
 return $settings;
--- a/core/lexicon/en/dashboard.inc.php
+++ b/core/lexicon/en/dashboard.inc.php
@@ -31,6 +31,7 @@ $_lang['updates_available'] = 'Updates available';
 $_lang['updates_update'] = 'Update';
 $_lang['updates_ok'] = 'Up to date';
 $_lang['updates_extras'] = 'Extras';
+$_lang['updates_to_download'] = 'To Download';
 
 $_lang['quicklinks'] = 'Quicklinks';
 $_lang['security_notices'] = 'Security Notices';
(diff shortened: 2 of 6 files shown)


smg6511:
I think this update widget needs to be left as-is for now. The issue you reference was fixed with #16608 for 3.0.6. I think we'll be moving toward integrating an actual upgrade widget that will have the ability to choose alternate packages instead of building on this non-interactive one. There's too much else to do to put work into including non-pl downloads, etc for this iteration of the widget IMO.

smg6511:
@Ibochkarev - See my note above. Recommend closing.

PR #16925, opened 2026-03-14
docs: add AGENTS.md and gitignore for Cursor/Claude Code

Ibochkarev:
### What does it do?
Adds `AGENTS.md` in the project root with instructions for AI coding assistants (Cursor, Claude Code): project structure, tech stack, tools and scripts, configuration files, architecture, code style, testing, CI/CD, and quick reference. Updates `.gitignore` to exclude `.cursor/`, `.claude/`, and `.claudeignore` so local AI tooling files are not committed.

### Why is it needed?
Developers using Cursor or Claude Code get consistent project context and conventions from a single file; ignoring editor/AI directories keeps the repo clean and avoids accidental commits of local config.

### How to test
- Open the repo in Cursor or Claude Code and confirm `AGENTS.md` is used as project instructions.
- Run `git check-ignore .cursor .claude .claudeignore` and confirm they are ignored.

### Related issue(s)/PR(s)
None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,8 @@ composer.lock
 
 # ignore .ddev dir
 /.ddev/
+
+# ignore AI assistant files
+.cursor/
+.claude/
+.claudeignore
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,212 @@
+# Project Instructions
+
+MODX Revolution 3.0.0 — open-source PHP CMS (GPL-2.0+). Architecture: **core** (PHP backend, xPDO ORM, Smarty), **manager** (ExtJS 3 admin UI), **setup** (web installer), **_build** (tests, Grunt/Sass build). This file is the single source of truth for AI assistants (Cursor, Claude Code). Follow paths, conventions, and DO NOT rules below.
+
+---
+
+## Table of Contents
+
+1. [Tech Stack](#tech-stack)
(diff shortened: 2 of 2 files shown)


pbowyer:
-1 from me on adding `AGENTS.md` because this will conflict with our customised file, which will then be overwritten on every MODX upgrade.

If kept I'd rather it's named `AGENTS.example.md` or similar - but unless it's shown to perform better than an auto-generated `AGENTS.md` I think we should leave it to users to generate the most up-to-date one via Codex.

opengeek:
-1 here on the AGENTS.md, as well. I believe this area is a little too volatile at this juncture to adopt a definitive AGENTS.md file. The freedom to experiment and iterate on AI guidance is going to serve us better here for now.

Could I plead that we consider changing paths a little here and have discussions on things before submitting PRs like this? A feature request on this would be ideal. Or even a forum post to get more eyes on it. IMO, collaboration that starts with ideas before moving to implementation would reduce noise and potential decision fatigue.

I'm also -1 on the .gitignore with specific exclusions like this. We could consider a generic rule to ignore all files and directories beginning with a `.`, but that feels heavy handed. You can exclude those AI agent support directories for the tools you are using in your local global gitignore rules, and that is what I am currently doing.

Also you are adding multiple barely related things in a single commit. Please keep PRs atomic.

PR #16926, opened 2026-03-14
fix(forms): Add templatename to Form Customization Set list (avoid N+1)

### LINKED ISSUE #16300: Form Customization in MODX 3: Template names don't initially appear, or are wrong
### Summary
Template names in the list of manager customizations is sometimes missing, or displays the wrong template names.

### Step to reproduce
Go to Form Customization, then click on customization name to edit the group.

### Observed behavior
The template name column is empty.
Screen Shot 2022-11-16 at 4 50 12 PM

Reload the screen, and the names appear. Sometimes I have to reload twice.
Screen Shot 2022-11-16 at 4 50 21 PM

This happens only the first time. If you close out of the customization group, then go back to it, the template names are still there. If, however, you clear the browser cache, then go back to the customization group, the template names are gone again. I've tried with 3 different browsers.

HOWEVER, sometimes the template names that do show up are wrong, so the same template name may appear multiple times, even though there is only one customization set for that template.

### Expected behavior
Template names should appear on first viewing.

### Environment
MODX 3 (all versions)

Ibochkarev:
### What does it do?
In `Security/Forms/Set/GetList` processor, `prepareRow()` now populates `templatename` in the data returned for each Form Customization Set. The value is taken from the already-selected joined column `Template.templatename` when present on the object (no extra query); otherwise it falls back to `getOne('Template')`. The value is always cast to string for consistent JSON/API type.

### Why is it needed?
Template names were missing or incorrect in the Form Customization UI because the list processor did not include `templatename` in the row data. The query already joins `modTemplate` and selects `Template.templatename`; using it when available avoids N+1 queries.

### How to test
1. Open Manager → Form Customization.
2. Ensure the list of sets shows the correct template name for each set (column or tooltip).
3. Create/edit a set with a template assigned and confirm the name appears correctly in the list.

### Related issue(s)/PR(s)
Resolves #16300

### DIFF
--- a/core/src/Revolution/Processors/Security/Forms/Set/GetList.php
+++ b/core/src/Revolution/Processors/Security/Forms/Set/GetList.php
@@ -93,6 +93,14 @@ class GetList extends GetListProcessor
     {
         $fcSetArray = $object->toArray();
 
+        // Use joined column from query when available to avoid N+1; fallback to relation if not hydrated
+        $templatename = $object->get('templatename');
+        if ($templatename === null || $templatename === '') {
+            $template = $object->getOne('Template');
+            $templatename = $template ? $template->get('templatename') : '';
+        }
+        $fcSetArray['templatename'] = (string) $templatename;
+
         $constraint_field = $object->get('constraint_field');
         $constraint = $object->get('constraint');
         if (!empty($constraint_field)) {


smg6511:
@Ibochkarev - Hey, this problem was already solved in a different way with #16653, which was merged before the 3.2.0 release.

Ibochkarev:
@smg6511 Hi. OK, please close the issue with a comment and I will close this PR.

smg6511:
Will do ;-)

Ibochkarev:
Fixed in https://github.com/modxcms/revolution/pull/16653

PR #16928, opened 2026-03-16
fix(manager): persist manager language across logout/login

### LINKED ISSUE #16072: The selected language of the manager panel is reset after exiting the manager
### Summary
My current system language is "Russian", I choose, for example, "German", I work with the manager in "German", everything is fine. But then I exit the manager, and when I enter the manager, on the login form, everything will be in "Russian" again, and I again need to select "German".

> When you log out from the manger, the session is destroyed and the manager_language setting is gone. That's the issue here.

> The only option is to save this in a cookie or in the local storage, but thats not part of this fix and has to be done elsewhere...

_Originally posted by @Jako in https://github.com/modxcms/revolution/pull/16069#issuecomment-1047782380_

### Expected behavior
The selected language must be saved, otherwise the meaning of changing the language is lost

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/issues/14694
https://github.com/modxcms/revolution/pull/16069

### Environment
MODX 3

Ibochkarev:
### What changed and why

Manager language lived only in `$_SESSION`. Logout destroyed the session, so the locale reset on the next login (#16072).

This PR persists the choice in cookie `modx_manager_language`, reads it on login, sets it on language change, and passes `manager_language` through the logout redirect (PHP + `modx.js`). Adds `modX::setManagerLanguageCookie()`.

### How to test

1. Log in, switch language (e.g. German).
2. Log out, log in again.
3. Confirm the manager stays on the chosen language.

### Related issue(s)/PR(s)

Resolves #16072

### Compatibility notes

Manager only. Cookie name `modx_manager_language`; value validated against available lexicons.

No API signature changes. Additive persistence for manager locale. Safe for patch releases.

### Test coverage

No automated tests added; manual login/logout flow.

### Contributors

None beyond author.

### AI tool use

### DIFF
--- a/manager/assets/modext/core/modx.js
+++ b/manager/assets/modext/core/modx.js
@@ -362,7 +362,12 @@ Ext.extend(MODx,Ext.Component,{
                 ,listeners: {
                     'success': {fn:function(r) {
                         if (this.fireEvent('afterLogout',r)) {
-                            location.href = './';
+                            var lang = Ext.util.Cookies && Ext.util.Cookies.get("modx_manager_language");
+                            var url = "./";
+                            if (lang) {
+                                url = "./?manager_language=" + encodeURIComponent(lang);
+                            }
+                            location.href = url;
                         }
                     },scope:this}
                 }
--- a/manager/controllers/default/security/login.class.php
+++ b/manager/controllers/default/security/login.class.php
@@ -205,16 +205,21 @@ class SecurityLoginManagerController extends modManagerController
 
         $ml = $this->modx->sanitizeString($this->modx->getOption('manager_language', $_REQUEST));
         if (!$ml || !in_array($ml, $languages)) {
-            $ml = $this->modx->getOption('manager_language', $_SESSION);
-            if (!$ml) {
-                // Try to detect default browser language
-                $accept_languages = strtolower($_SERVER['HTTP_ACCEPT_LANGUAGE']);
-                preg_match_all('#([\w-]+)(?:[^,\d]+([\d.]+))?#', $accept_languages, $matches, PREG_SET_ORDER);
-                foreach ($matches as $match) {
-                    $lang = trim(explode('-', $match[1])[0]);
-                    if (in_array($lang, $languages)) {
-                        $ml = $lang;
-                        break;
+            $ml = isset($_COOKIE['modx_manager_language'])
+                ? $this->modx->sanitizeString($_COOKIE['modx_manager_language'])
+                : null;
+            if (!$ml || !in_array($ml, $languages)) {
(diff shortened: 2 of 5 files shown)


manager/assets/modext/core/modx.js
```diff
@@ -362,7 +362,12 @@ Ext.extend(MODx,Ext.Component,{
                 ,listeners: {
                     'success': {fn:function(r) {
                         if (this.fireEvent('afterLogout',r)) {
-                            location.href = './';
+                            var lang = Ext.util.Cookies && Ext.util.Cookies.get("modx_manager_language");
+                            var url = "./";
+                            if (lang) {
+                                url = "./?manager_language=" + encodeURIComponent(lang);
+                            }
+                            location.href = url;
```
smg6511: Reminder: Never use var unless the variable has to be in the global scope. Otherwise, I just suggest cleaning up the declaration in general as shown ;-)
```suggestion
                            const
                                lang = Ext.util.Cookies && Ext.util.Cookies.get('modx_manager_language'),
                                url = lang ? `./?manager_language=${encodeURIComponent(lang)}` : './'
                            ;
                            window.location.href = url;
```

manager/controllers/default/security/login.class.php
```diff
@@ -205,16 +205,21 @@ public function handleLanguageChange()
...
+                    $accept_languages = strtolower($_SERVER['HTTP_ACCEPT_LANGUAGE'] ?? '');
+                    preg_match_all('#([\w-]+)(?:[^,\d]+([\d.]+))?#', $accept_languages, $matches, PREG_SET_ORDER);
+                    foreach ($matches as $match) {
+                        $lang = trim(explode('-', $match[1])[0]);
+                        if (in_array($lang, $languages)) {
+                            $ml = $lang;
+                            break;
+                        }
                     }
                 }
```
smg6511: I realize this was part of the original, but consider removing this block that attempts to detect and apply the lang. I looked into this and did a bit of testing and found the method very unreliable due to behavior varying pretty widely among browsers and likely OSs.

smg6511 (changes requested):
Ok, this definitely works. I have one change request and an optional suggestion for you to look at, then it's good to go.

manager/controllers/default/security/login.class.php
Ibochkarev: Removed the Accept-Language block. If the request, cookie, and session have no usable language, login falls back to `en`.

Ibochkarev: @smg6511 done, thanks

smg6511 (approved):
Rechecked and works well.

PR #16930, opened 2026-03-21
Add createdon/editedon timestamps to element and category tables

pbowyer:
### What does it do?
MODX tracks when resources, users, and settings were created and last modified, but not elements. There's no way to tell when a snippet was last changed or which templates haven't been touched in years.

As a developer, this information is useful. This PR adds `createdon` and `editedon` columns to chunks, snippets, plugins, templates, TVs, categories, and their association tables.

### How does it work?
`createdon` is set in PHP when an element or category is first saved, the same way `modUser` already handles it. `editedon` is handled by MySQL (`ON UPDATE CURRENT_TIMESTAMP`), so it requires no application code. Both use the `datetime` type rather than `timestamp` to avoid timezone conversion quirks and the 2038 year limit.

Existing records get `NULL` for both columns. `editedon` stays `NULL` until an actual edit is made, which matches how resources behave.

For static elements, `editedon` reflects when MODX last synced the file content to the database, which may differ from when the file was actually modified. `editedon` reflecting when the DB row was last updated is at least honest and consistent.

### How to test
1. Run MODX setup upgrade to verify the upgrade script adds columns
2. Create a new element in the manager and verify createdon appears
3. Edit it and verify editedon is populated

I've extended the test suite to verify `createdon` is set on creation, `editedon` is `null` after creation, `createdon` is preserved after update.

### Related issue(s)/PR(s)
https://github.com/modxcms/revolution/pull/13857 was a previous PR to do this. It added `createdby` and `editedby` as well, and as it went the scope expanded. I have kept my PR focused because I am not using the other features. It ran into problems with static elements, and think I have taken a pragmatic approach to solving that.

#13770 #13626 #13641 #2027

### DIFF
--- a/_build/test/Tests/Model/Element/modPluginTest.php
+++ b/_build/test/Tests/Model/Element/modPluginTest.php
@@ -68,6 +68,35 @@ class modPluginTest extends MODxTestCase {
     }
 
     /**
+     * Loading an unchanged static Plugin's content must not save the object.
+     *
+     * @return void
+     */
+    public function testGetUnchangedStaticContentDoesNotSetEditedon() {
+        $staticFile = MODX_CORE_PATH . 'cache/unit-test-static-plugin.php';
+        $plugin = $this->modx->newObject(modPlugin::class);
+        $plugin->fromArray([
+            'name' => 'Unit Test Static Plugin',
+            'plugincode' => 'return "Static content";',
+            'static' => true,
+            'static_file' => $staticFile,
+        ]);
+
+        try {
+            $this->assertTrue($plugin->save());
+            $this->assertEmpty($plugin->get('editedon'));
+
+            /** @var modPlugin $reloaded */
+            $reloaded = $this->modx->getObject(modPlugin::class, $plugin->get('id'), false);
+            $this->assertSame('return "Static content";', $reloaded->getContent());
+            $this->assertEmpty($reloaded->get('editedon'));
+        } finally {
+            $plugin->remove();
+            @unlink($staticFile);
+        }
+    }
+
+    /**
      * @param string $content
      * @dataProvider providerSetContent
      * @depends testGetContent
--- a/_build/test/Tests/Model/Element/modSnippetTest.php
+++ b/_build/test/Tests/Model/Element/modSnippetTest.php
@@ -70,6 +70,35 @@ class modSnippetTest extends MODxTestCase {
     }
 
     /**
+     * Loading an unchanged static Snippet's content must not save the object.
+     *
+     * @return void
+     */
+    public function testGetUnchangedStaticContentDoesNotSetEditedon() {
+        $staticFile = MODX_CORE_PATH . 'cache/unit-test-static-snippet.php';
+        $snippet = $this->modx->newObject(modSnippet::class);
+        $snippet->fromArray([
+            'name' => 'Unit Test Static Snippet',
+            'snippet' => 'return "Static content";',
+            'static' => true,
+            'static_file' => $staticFile,
+        ]);
+
+        try {
+            $this->assertTrue($snippet->save());
+            $this->assertEmpty($snippet->get('editedon'));
+
+            /** @var modSnippet $reloaded */
+            $reloaded = $this->modx->getObject(modSnippet::class, $snippet->get('id'), false);
+            $this->assertSame('return "Static content";', $reloaded->getContent());
+            $this->assertEmpty($reloaded->get('editedon'));
+        } finally {
+            $snippet->remove();
+            @unlink($staticFile);
+        }
+    }
+
+    /**
      * @param string $content
      * @dataProvider providerSetContent
      * @depends testGetContent
--- a/_build/test/Tests/Processors/Element/CategoryTest.php
+++ b/_build/test/Tests/Processors/Element/CategoryTest.php
@@ -193,9 +193,10 @@ class CategoryProcessorsTest extends MODxTestCase {
         $category = $this->modx->getObject(modCategory::class, ['category' => 'UnitTestCatTimestamp']);
         $this->assertNotNull($category, 'Category not found after creation');
 
-        $this->assertTrue(is_numeric($category->get('createdon')), 'createdon should be a Unix timestamp');
-        $this->assertGreaterThan(0, (int)$category->get('createdon'), 'createdon should be set on new Category');
-        $this->assertEquals(0, (int)$category->get('editedon'), 'editedon should start empty on new Category');
+        // createdon is an int-backed timestamp field, so get() reads back a formatted date string.
+        $this->assertNotEmpty($category->get('createdon'), 'createdon should be set on new Category');
+        $this->assertGreaterThan(0, strtotime($category->get('createdon')), 'createdon should be a valid timestamp');
+        $this->assertEmpty($category->get('editedon'), 'editedon should start empty on new Category');
     }
 
     /**
--- a/_build/test/Tests/Processors/Element/CategoryTest.php
+++ b/_build/test/Tests/Processors/Element/CategoryTest.php
@@ -180,7 +180,7 @@ class CategoryProcessorsTest extends MODxTestCase {
     }
 
     /**
-     * Tests that createdon is set when creating a Category and editedon remains null
+     * Tests that createdon is set when creating a Category and editedon starts empty
      */
     public function testCategoryCreateSetsCreatedon() {
         /** @var ProcessorResponse $result */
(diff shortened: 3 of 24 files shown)


smg6511:
@pbowyer Hey Peter - Absolutely agree the timestamp additions should be done. However, for consistency with the current data structure, I'd much rather see them implemented as unix timestamps. 

For next-gen MODX (presumable 4.0) the optimal date/time storage type can be debated and, if changed, implemented across the whole codebase. Note that in the context of web apps, the 32-bit limitation is really a non-issue as, from my research, you're pretty much only going to find legacy non 64-bit hardware in industrial settings, especially 12 years from now.

pbowyer:
Thanks for your review @smg6511. I disagree with you (I would improve the column type now for new additions, so there's less to migrate in MODX v4) but I've changed them to be Unix timestamps as requested.

smg6511 (commented):
@pbowyer - Hey Peter, thanks for being flexible and adjusting to UTs for this PR. I've confirmed that, for the most part, timestamps are recorded as expected. I have a handful of initial review questions and requests to share ahead of a more final review:

1. I'm unsure about the value of adding these cols to the association tables, especially the TVRs one. I'd like to get your take on how it'd be useful. @opengeek @Mark-H and others: It'd be great to get your opinions on this as well.
2. Same as above goes for the Plugin Event table.
3. With Snippets and Plugins, after the first save something seems to trigger the save again (or maybe it's just the reloading of the page) — resulting in `editedon` being recorded as well (a second or two later than `createdon`). It'd be ideal to find a way for this to not happen.
4. Nuance 1: I think logic should also be added for the Duplicate processors, adding a `beforeSave` method to reset the duplicated element's `createdon` to the current `time()` and `editedon` to empty.
5. Nuance 2: This may or may not be a job for another PR but I noted that when a static element's content is externally edited you have to at least revisit/reload the element's editing page for `editedon` to be updated. I was wondering if clearing cache could also trigger this update. This may be too fine a point, but just wanted to at least bring it up.

pbowyer:
> * I'm unsure about the value of adding these cols to the association tables, especially the TVRs one. I'd like to get your take on how it'd be useful. @opengeek @Mark-H and others: It'd be great to get your opinions on this as well.
> 
> * Same as above goes for the Plugin Event table.

Two I did because I have a use for them: for plugin events it's helpful to see when new events were added to a plugin. The same to a lesser extent for `site_tmplvar_access` because it helps with auditing direct from the database. 

The other two (TVRs and TVTpls) I included for consistency, not because I currently would use that field. 

> * With Snippets and Plugins, after the first save something seems to trigger the save again (or maybe it's just the reloading of the page) — resulting in `editedon` being recorded as well (a second or two later than `createdon`). It'd be ideal to find a way for this to not happen.

I cannot reproduce this. Do you perhaps still have the `ON UPDATE CURRENT_TIMESTAMP` present after changing the `datetime` column type to `int`?

> * Nuance 1: I think logic should also be added for the Duplicate processors, adding a `beforeSave` method to reset the duplicated element's `createdon` to the current `time()` and `editedon` to empty.

Good catch, added in 97b73d45df4c60817cede2561243131d6f76f45a.

> * Nuance 2: This may or may not be a job for another PR but I noted that when a static element's content is externally edited you have to at least revisit/reload the element's editing page for `editedon` to be updated. I was wondering if clearing cache could also trigger this update. This may be too fine a point, but just wanted to at least bring it up.

I purposefully kept this out of scope of this PR because it's what stopped the former PR (https://github.com/modxcms/revolution/pull/13857#issuecomment-392159213 onwards). Addressing it in a follow-on PR will keep the discussion focused and allow the side-discussions static elements seem to generate(!) to happen without affecting this change.

smg6511:
@pbowyer - Hello again! Ok, as I continue to look this over, I realize I failed to specify that the effect I was seeing with Snippets and Plugins seemingly double-saving had to do with static ones. That's _probably_ why you couldn't replicate what I was describing (?).

Thanks for making the changes for the Duplicate processor, that's working as expected.

On the association tables, I'm not saying “no” but just want you to consider whether these changes have broader appeal. If it's pretty niche or for mostly a personal need, then it might be the job of a custom component/plugin or and Extra (thinking of auditing).

pbowyer:
Hi @smg6511 and thanks for the bug report. I was able to reproduce and fix the problem, it was happening because static Snippet/Plugin files retained a leading newline after stripping `<?php`. MODX interpreted the content as changed, saved the element during reload, and populated `editedon`.

Tangentially the same extra newline has caused us problems when using Gitify, as it leads to unexpected changes after exporting an imported snippet saved via the Manager. I'd never investigated, but this explains why it happens and the fix will sort that too.

> On the association tables, I'm not saying “no” but just want you to consider whether these changes have broader appeal. If it's pretty niche or for mostly a personal need, then it might be the job of a custom component/plugin or and Extra (thinking of auditing).

I don't think that's for me to consider. I've presented a PR and @opengeek needs to make a call on their inclusion or not as he's steering the direction of the project.

smg6511:
@pbowyer - Ok, so let's work on moving this forward _without_ the addition to the association tables. I talked with Jason a while back and, while he didn't have a strong opinion on this, he said that it was a bit much adding them there. I personally think that dates should only be persisted in the db for the objects we create and tracking activity re peripheral tables' data might better be left to an auditing Extra/CMP.

pbowyer:
Hi @smg6511 I've removed the association timestamps and rebased the branch too.

smg6511:
@pbowyer - Hey Peter, thanks for the update.

@opengeek @JoshuaLuckers @Mark-H - and others who were/are involved in this and the original solution: The small issue of `editedon` not being tracked when a static file is updated outside of MODX still stands, but I think that's a separate issue to tackle in a follow-up PR. Everything else is in order here, so I say let's move forward.

smg6511 (approved):
Re-tested and works.

JoshuaLuckers (approved):

opengeek (changes requested):
Thank you for your work on this, Peter. The model and processor changes look good and the maps match a regeneration from the schema. One significant issue with the setup side, though, was found while running a build + upgrade against a snapshot.

### Blocking: the upgrade script never runs

`setup/includes/upgrades/common/3.2.1-element-timestamps.php` is not included by anything. `modInstallVersion::_getUpgradeScripts()` scans `upgrades/mysql/`, and each `mysql/X.Y.Z-pl.php` there includes the relevant `common/` scripts. Without a wrapper the columns are never added.

After running the upgrade, any element or category save fails, and since xPDO selects columns explicitly, element grids fail:

```
Unknown column 'modSnippet.createdon' in 'SELECT'
Unknown column 'createdon' in 'INSERT INTO'
```

Separately, the `3.2.1` prefix is wrong for the release this ships in. `version_compare('3.2.2-pl', '3.2.1-element-timestamps', '<')` is false, thus installs on 3.2.1 or 3.2.2 would skip it even if it were wired up.

Fix (verified locally: columns added to all six tables, saves work, `createdon` set on create, `editedon` set on update):

1. Rename to `common/3.3.0-element-timestamps.php`
2. Add `setup/includes/upgrades/mysql/3.3.0-pl.php`:

```php
<?php
/**
 * Specific upgrades for Revolution 3.3.0-pl
 *
 * @var modX $modx
 * @package setup
 * @subpackage upgrades
 */

/* run upgrades common to all db platforms */
include dirname(__DIR__) . '/common/3.3.0-element-timestamps.php';
```

### Cleanup Requests/Considerations

- **PR description is out of date.** It describes `datetime` columns with `ON UPDATE CURRENT_TIMESTAMP`, `NULL` for existing rows, and association tables. The implementation is int Unix timestamps set in PHP with default `0`, and association tables were dropped. Please freshen this in order to maintain accurate merge history.
- **`editedon` is stamped on every non-new `save()`**, including category moves via the Sort processor, elements orphaned by a category removal, static file resyncs in `getContent()`, and package installs. That fits the "last time the row changed" semantic you chose, but it differs from `modResource`, where processors set it. I wanted to note this to ensure this behavior is a deliberate decision.
- **`modScript::getFileContent()` trim change:** Good fix for the leading newline issue. One-time side effect after upgrade: existing static scripts that stored the newline will be re-saved once on first load and get an `editedon` stamp. I think this is fine, but worth a line in the description.
- Whitespace-only changes remain on `modPluginEvent`, `modTemplateVarResource*`, `modTemplateVarTemplate` maps and a blank line in `PasswordResetToken.php` and its test, left over from the reverted association table commit. Harmless, but could be dropped to keep this clean.

pbowyer:
Hi @opengeek,

Thanks for running the build and upgrade against a snapshot, and for the fix. I've renamed the script to `common/3.3.0-element-timestamps.php` and added `mysql/3.3.0-pl.php` to include it.

On `editedon`: yes, stamping it on every save of an existing row is deliberate. I wanted it to mean "the last time this row changed", whichever code path saved it, so category moves, orphaning, static resyncs and package installs all count. I've spelled this out in the description, along with the one-off re-save of static scripts that stored the leading newline.

I've also rewritten the description to match the implementation, and dropped the leftover whitespace changes from the maps, `PasswordResetToken.php` and its test.

Issue #12262, opened 2015-01-11
RFC: Enhanced logger support


pbowyer:
# Abstract & Motivation

I would like to make MODX Revolution accept a PSR3-compliant logger. For my use-cases the biggest win is flexibility, as you can log to different streams (so I can feed everything directly to Splunk/Logstash, but have critical errors emailed to the team), each of which can record different information.

The proposal maintains total backwards compatibility, and any changes which presented this would not be compatible with this proposal.
# Proposal
### Step 1

Currently the responsibility for logging is tightly entangled into xPDO (the ORM) and modX (the core MODX class). Step 1 is to refactor the code and extract this into its own class or library.
### Step 2

Step 2 is to handle the autoloading of a PSR3 library. This library may be distributed with MODX or may be installed via Composer (for licensing compatibility, it may have to be the latter). @rtripault is working on a Composer PoC; I have left this section vague on purpose as I want a pragmatic and tightly-focused proposal [translated: I don't care how this bit is done for now, so long as it works]
### Step 3

Step 3 is to write a compatibility layer between the MODX logging class and the PSR3-compatible library.
### Step 4

Step 4 is to provide a mechanism to define and load whichever logging library. I propose that additional logging engines can be defined in `core/config/config.inc.php`, the logger is initialised in `index.php` and passed as a constructor argument to `modX()`. 

This however is a big change and can cause backward-compatibility issues - it may be necessary to make it optional and have the logger initialised within `xPDO()` as at present _unless_ an instance is passed in to the constructor.  I can only think of a Dependency Injection Container as an alternative...
### Step 5

Test :)
# Rationale

The rationale for separating the logging out from xPDO and modX is simple: the responsibility for writing logs does not belong with them, and is the responsibility of a separate object.

The rationale for choosing a PSR3-compatible logger is less clear. They are popular at the moment, the libraries widely used and well-tested (one less component to maintain), and the approach they take (of streams and renderers) is both flexible and used in PHP and Python, possibly other language communities too.

netProphET:
Something to ponder re: flexibility. Would users want or need to be able to define the logger itself at the context level, or will defining a single logger at the application level, and utilizing context settings suffice?

pbowyer:
Good thinking @netProphET, I can see it being useful to separate them out sometimes; equally global errors are global errors.

[...]

PR #16931, opened 2026-03-21
Add PSR-3 logging support in MODX

pbowyer:
## What this does

Previously MODX hasn't supported PSR-3 logging. This was added to xPDO (commit https://github.com/modxcms/xpdo/commit/55fd34318a86d9ebe2cf01564c7126d90034af2c) but `modX::_log()` overrides it in several code paths, so a logger registered on xPDO never sees all of MODX's log output.

This PR adds `setLogger(LoggerInterface $logger)` to MODX, restructures `_log()` to dispatch to a PSR-3 logger when one is set, and adds 20 new tests (11 baseline, 9 PSR-3).

```php
use Monolog\Logger;
use Monolog\Handler\StreamHandler;

$logger = new Logger('modx');
$logger->pushHandler(new StreamHandler('/path/to/modx.log', Logger::DEBUG));
$logger->pushProcessor(new Monolog\Processor\IntrospectionProcessor());

$modx->setLogger($logger);
$modx->setLogLevel(modX::LOG_LEVEL_DEBUG); // control what reaches the logger
```

## Backward compatibility

When no PSR-3 logger is registered (`$this->logger === null`, the default), all logging behaviour is identical to the previous implementation. I added 11 tests as a baseline to verify this.

## Design decisions and alternatives

Three questions came up during implementation where the answer isn't obvious. I went back and forth on them @opengeek and picked one to finish the code, but without a strong view on them.

### 1. Should the PSR-3 logger respect MODX's `logLevel`, or receive everything?

- **A. Respect `logLevel` (chosen):** The PSR-3 logger only sees messages that pass the `logLevel` threshold. Users call `$modx->setLogLevel(modX::LOG_LEVEL_DEBUG)` when they want full visibility, and can filter further in their logger config (e.g. Monolog handler levels).

- **B. Send everything to the logger:** The PSR-3 logger receives all messages regardless of `logLevel`. The logger's own filtering decides what to keep.

I chose A because it seemed right, but have been doubting that since. Initially I chose 2B (below) so it made sense for performance, as every `log()` call that isn't filtered out early triggers `debug_backtrace()` in the legacy path to resolve the calling file and line. The early `logLevel` check (an integer comparison) skips all that work for messages below threshold. With option B and the default `LOG_LEVEL_ERROR`, every DEBUG/INFO/WARN call would still build context and dispatch to the logger. The cost of A is one extra line of config (`setLogLevel`) when setting up a logger.

But then I switched to 2A so the performance impact of 1B would be minimal.

**To switch to B:** Remove the early return at the top of `_log()` for the PSR-3 path.

### 2. Should `debug_backtrace()` be called for the PSR-3 path?

- **A. Skip backtrace for PSR-3 (chosen):** The `file` and `line` context fields are only populated if the caller passed them (e.g. `__FILE__`, `__LINE__`). When not provided, they're empty strings. Monolog users add `IntrospectionProcessor` for automatic caller info.

- **B. Always resolve backtrace:** Every log call resolves the backtrace so that `file`/`line` are always in the PSR-3 context, matching legacy behavior.

I chose A because `debug_backtrace()` is the most expensive operation in the log path. I felt this should be left to the developer as e.g. Monolog's `IntrospectionProcessor` does the same thing but only for messages that survive handler-level filtering. So a DEBUG message discarded by the handler never triggers a backtrace.

**To switch to B:** Move the backtrace resolution to before the `_dispatchToPsrLogger()` call, so the PSR-3 context always has file/line. Simpler for users who don't want to configure logging processors, but costs a backtrace on every dispatched log call.

### 3. Should the PSR-3 logger replace legacy output, or be additive?

- **A. Replace legacy output (chosen):** When a PSR-3 logger is set, legacy FILE/ECHO/HTML/ARRAY output is suppressed. Only modRegister logging is preserved alongside the PSR-3 logger.

- **B. Additive (both):** Both the PSR-3 logger and legacy targets receive messages. The existing `error.log` continues alongside the custom logger.

- **C. Configurable:** Add an option (e.g. `psr_logger_exclusive`) to let users choose at runtime.

I chose A because if you've set up a PSR-3 logger, you're taking control of where logs go. Writing to `error.log` at the same time is confusing and wasteful. modRegister is preserved because it's a message queue, not traditional logging.

**To switch to B:** In the restructured `_log()`, remove the early return after the PSR-3 dispatch and let execution fall through to the legacy path. The `$this->logger = null` / `try/finally` pattern already exists for the legacy `parent::_log()` call and would prevent double-dispatch.

**To switch to C:** Add a config option check before the early return, e.g.:
```php
if ($hasPsrLogger && $this->getOption('psr_logger_exclusive', null, true)) {
    // skip legacy
    return;
}
// fall through to legacy path
```

### Other notes

- **xPDOLogger exclusion:** The bundled `xPDOLogger` (standalone xPDO's default logger) is excluded from MODX's PSR-3 dispatch. Its `handleXpdo()` calls `exit()` for FATAL, while MODX uses `sendError('fatal')` which renders a proper error page. Allowing `xPDOLogger` would change the current MODX behaviour, so I've assumed it is for standalone xPDO use only.

- **FATAL handling:** MODX uses `sendError('fatal')` for fatal log messages, not `exit()`. The PSR-3 logger receives the FATAL message (mapped to `LogLevel::CRITICAL`) before `sendError()` runs.

- **Backward compatibility:** When no PSR-3 logger is set (`$this->logger === null`, the default), behavior is identical to the previous implementation. The baseline tests verify this.

### Related issue(s)/PR(s)
- https://github.com/modxcms/revolution/issues/12262

### DIFF
--- a/_build/test/Tests/Model/TestLogger.php
+++ b/_build/test/Tests/Model/TestLogger.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+
+namespace MODX\Revolution\Tests\Model;
+
+use Psr\Log\AbstractLogger;
+
+class TestLogger extends AbstractLogger
+{
+    /** @var array */
+    public $records = [];
+
+    public function log($level, $message, array $context = []): void
+    {
+        $this->records[] = [
+            'level' => $level,
+            'message' => $message,
+            'context' => $context,
+        ];
+    }
+}
--- a/_build/test/Tests/Model/modXLoggingTest.php
+++ b/_build/test/Tests/Model/modXLoggingTest.php
@@ -1,0 +1,444 @@
+<?php
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+namespace MODX\Revolution\Tests\Model;
+
+use MODX\Revolution\modX;
+use MODX\Revolution\MODxTestCase;
+use Psr\Log\AbstractLogger;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use xPDO\xPDO;
+use xPDO\Cache\xPDOCacheManager;
+
+/**
+ * Tests related to modX logging behavior.
+ *
+ * @package modx-test
+ * @subpackage modx
+ * @group Model
+ * @group modX
+ * @group Logging
+ */
+class modXLoggingTest extends MODxTestCase
+{
+    /**
+     * @var int Original log level to restore after each test.
+     */
+    private $originalLogLevel;
+
+    /**
+     * @var mixed Original log target to restore after each test.
+     */
+    private $originalLogTarget;
+
+    /**
+     * @before
+     * @throws \xPDO\xPDOException
+     */
+    public function setUpFixtures()
+    {
+        parent::setUpFixtures();
+        $this->originalLogLevel = $this->modx->getLogLevel();
+        $this->originalLogTarget = $this->modx->getLogTarget();
+        // Ensure no PSR-3 logger is set for baseline tests
+        $this->modx->logger = null;
+    }
+
+    /**
+     * @after
+     */
+    public function tearDownFixtures()
+    {
+        $this->modx->setLogLevel($this->originalLogLevel);
+        $this->modx->setLogTarget($this->originalLogTarget);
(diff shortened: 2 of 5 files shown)


opengeek:
Thanks for working on this, @pbowyer! I'm working out a release plan for a new minor of xPDO. Let's keep this in draft until we have that available.

pbowyer:
Since posting this I _think_ I have swung in favour of 1B... 😊

modxcommunity:

https://community.modx.com/t/thinking-through-contribution-processes-and-seeking-input/9051/9

opengeek:
@pbowyer — xpdo ^3.2 is available with the work you did there — if you get some time, can we get the composer version restraint updated here and get this PR ready for inclusion in a 3.3 release?

pbowyer:
There we are @opengeek, done!

pbowyer:
The CI failure is a Docker pull failure, and I can't find a way to rerun the action without adding a commit?

opengeek:
> The CI failure is a Docker pull failure, and I can't find a way to rerun the action without adding a commit?

I'll dig into this. Dependency on the mysql 5.6 images need to be addressed.

opengeek (approved):

PR #16933, opened 2026-03-26
Fix PHP warning: Undefined variable $image

Jako:
### What does it do?
Use the null coalescing operator to assign an empty string when $image is not defined

### Why is it needed?
PHP warning: Undefined variable $image

### DIFF
--- a/core/src/Revolution/Sources/modMediaSource.php
+++ b/core/src/Revolution/Sources/modMediaSource.php
@@ -2341,7 +2341,7 @@ QTIP;
         }
 
         return [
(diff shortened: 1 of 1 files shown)


Ibochkarev (approved):

smg6511:
@Jako - Is there an open issue this is related to? If so, just want to identify and close it upon this getting merged in. Thx!

Jako:
I don't think so. It just popped up in the logs of one installation that has warnings enabled.

opengeek (approved):

PR #16935, opened 2026-03-26
Fix PHP warning: Undefined array key "updateable"

Jako:
### What does it do?
Initialise the array key updateable before using it.

### Why is it needed?
PHP warning: Undefined array key "updateable"


smg6511:
Hey @Jako - Consider handling this the same way it is done in the `process` method L40, changing L128 to:
```php
$categoryData = [
    'updateable' => 0
];
```
...and leave the line you changed as it was before.

Also, let me know if there happens to be a related Issue. Thx!

Jako:
You can fix this however you want.

opengeek (approved):
Approved with the changes that were suggested.

PR #16937, opened 2026-03-28
fix(manager): correct OnBeforeManagerPageInit / OnManagerPageInit order

### LINKED ISSUE #9461: Wrong events order
fi1osof created Redmine issue ID 9461

OnManagerPageInit event fired before OnBeforeManagerPageInit.

OnManagerPageInit fired in modManagerRequest, OnBeforeManagerPageInit fired in modManagerController.

Maybe better will remove OnManagerPageInit from modManagerRequest and register new evenm name?

Ibochkarev:
### What changed and why

`OnManagerPageInit` fired in `modManagerRequest` before `OnBeforeManagerPageInit` in `modManagerController::render()`. Plugin order did not match documented intent (#9461).

This PR fires `OnBeforeManagerPageInit` first in `modManagerController::render()`, then `OnManagerPageInit` with `array_merge($this->config, ['action' => …, 'namespace' => …])`. The early call in `modManagerRequest` is removed.

### How to test

```bash
core/vendor/bin/phpunit -c _build/test/phpunit.xml --filter ManagerPageInitEventsTest
```

Confirm plugins on both events run in the expected order and `OnManagerPageInit` receives action/namespace in the event payload.

### Related issue(s)/PR(s)

Resolves #9461

### Compatibility notes

Manager only. Plugins that depended on `OnManagerPageInit` firing in the request handler before controller config existed will see the event later, with a richer payload. Matches the approach suggested in the issue.

Intentional event timing change. No signature changes. Review plugins that hook both events before merge.

### Test coverage

`_build/test/Tests/Controllers/ManagerPageInitEventsTest.php`

### Contributors

Issue reporter and discussants on #9461.

### AI tool use

### DIFF
--- a/_build/test/Tests/Controllers/ManagerPageInitEventsTest.php
+++ b/_build/test/Tests/Controllers/ManagerPageInitEventsTest.php
@@ -1,0 +1,109 @@
+<?php
+
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For the full copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ */
+
+namespace MODX\Revolution\Tests\Controllers;
+
+use MODX\Revolution\modManagerRequest;
+use MODX\Revolution\modX;
+use Yoast\PHPUnitPolyfills\TestCases\TestCase;
+
+/**
+ * Ensures manager page system events fire in semantic order (OnBeforeManagerPageInit before OnManagerPageInit, #9461).
+ *
+ * @group Controllers
+ * @group ManagerController
+ */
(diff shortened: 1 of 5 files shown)


smg6511:
@opengeek - Was taking a look at this and wanted your stance on the potential BC implications and whether they would cause this change to be blocked anyway. Otherwise it of course makes sense for these events to run in the expected order, although I find the wrapping of them in a new method just to accomodate a test to be unnecessary.

Ibochkarev:
@smg6511 Inlined both events in `render()` and dropped `invokeManagerPageInitEvents()`. Tests now mirror that block without a production helper.

`OnManagerPageInit` still fires later than pre-3.x (`modManagerRequest`) and carries the merged `config` payload. I treat that as an intentional fix for #9461; shout if you want the old call site kept for BC.

PR #16940, opened 2026-04-01
Catch failure checking S3 file visibility

matdave:
### What does it do?
Adds a catch around the `$this->filesystem->visibility` function call in S3 media sources.

### Why is it needed?
Visibility can throw a `FilesystemException` exception if it is not implemented in the user's permissions. https://github.com/thephpleague/flysystem/blob/8aaffb653c5777781b0f7f69a5d937baf7ab6cdb/src/FilesystemAdapter.php#L68

### How to test
Ran into this on a site where the user did not have permissions to retrieve object metadata. However, I'm unsure how to replicate that.

### DIFF
--- a/core/src/Revolution/Sources/modS3MediaSource.php
+++ b/core/src/Revolution/Sources/modS3MediaSource.php
@@ -663,7 +663,7 @@ class modS3MediaSource extends modMediaSource
         if ($this->visibility_dirs) {
             try {
                 return $this->filesystem->visibility($path);
-            } catch (\Exception | FilesystemException $exception) {
+            } catch (FilesystemException | UnableToRetrieveMetadata $exception) {
                 $this->xpdo->log(modX::LOG_LEVEL_INFO, $exception->getMessage());
             }
         }
--- a/core/src/Revolution/Sources/modS3MediaSource.php
+++ b/core/src/Revolution/Sources/modS3MediaSource.php
@@ -652,7 +652,7 @@ class modS3MediaSource extends modMediaSource
             try {
                 return $this->filesystem->visibility($path);
             } catch (\Exception | FilesystemException $exception) {
-                $this->xpdo->log(modX::LOG_LEVEL_ERROR, $exception->getMessage());
+                $this->xpdo->log(modX::LOG_LEVEL_INFO, $exception->getMessage());
             }
         }
         return false;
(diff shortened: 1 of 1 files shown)


core/src/Revolution/Sources/modS3MediaSource.php
```diff
@@ -649,15 +649,23 @@ public function getVisibility($path)
     public function getVisibilityFile($path)
     {
         if ($this->visibility_files) {
-            return $this->filesystem->visibility($path);
+            try {
+                return $this->filesystem->visibility($path);
+            } catch (\Exception | FilesystemException $exception) {
```
JoshuaLuckers: Is there a reason why the \Exception is also a log level info? In my opinion this means something important is broken right?

matdave: @JoshuaLuckers 

> Is there a reason why the \Exception is also a log level info?

This is to prevent spamming the error log on a valid site. 

>In my opinion this means something important is broken right?

No, it can be valid that your S3 settings do not allow for retrieving visibility.

opengeek:
```suggestion
            } catch (FilesystemException | UnableToRetrieveMetadata $exception) {
```

This matches the file's established convention exactly and aligns with the documented exception path in Flysystem's `FilesystemAdapter`. Real bugs from the AWS SDK side will continue to propagate (correct), and the known-harmless permission-denied case (which surfaces as `UnableToRetrieveMetadata extends FilesystemException`) is still caught.

core/src/Revolution/Sources/modS3MediaSource.php
```diff
@@ -649,15 +649,23 @@ public function getVisibility($path)
...
         return false;
     }
 
     public function getVisibilityDirectory($path)
     {
         if ($this->visibility_dirs) {
-            return $this->filesystem->visibility($path);
+            try {
+                return $this->filesystem->visibility($path);
+            } catch (\Exception | FilesystemException $exception) {
```
opengeek:
```suggestion
            } catch (FilesystemException | UnableToRetrieveMetadata $exception) {
```

same as above

opengeek (changes requested):

core/src/Revolution/Sources/modS3MediaSource.php
opengeek: @matdave ?

core/src/Revolution/Sources/modS3MediaSource.php
opengeek: @matdave ?

opengeek (approved):

PR #16827, opened 2026-02-19
Fix: convert refresh_site to modManagerController class

Ibochkarev:
### What does it do?

Replaces the script-style `refresh_site.php` with a proper `SystemRefreshSiteManagerController` extending `modManagerController`. The controller implements `checkPermissions()`, `process()`, `getPageTitle()`, `loadCustomCssJs()`, and `getTemplateFile()`.

### Why is it needed?

The action `/manager/?a=system/refresh_site` did not work correctly because the old script used `$this->checkFormCustomizationRules()` without a controller context. The loader expects a `.class.php` controller, so the page could not load. As a result, cache was not cleared (e.g. frontendManager).

### How to test

1. Open `/manager/?a=system/refresh_site` in the manager.
2. Confirm the page loads and shows the cache refresh result.
3. Confirm cache is cleared (e.g. frontendManager, lexicon).
4. Confirm the "Clear Cache" menu item still works via `MODx.clearCache()`.

### Related issue(s)/PR(s)

Resolves #16282

PR #16947, opened 2026-04-28
Remove unreachable refresh_site controller

### LINKED ISSUE #16282: Revo 3: the action /manager/?a=system/refresh_site does not work properly
### Summary
The action **/manager/?a=system/refresh_site** does not work properly 

### Why is it needed?
for example frontendManager component does not clear the cache

### Suggested solution(s)
Need to rename file **/manager/controllers/default/system/refresh_site.php** to **/manager/controllers/default/system/refresh_site.class.php**
and change the code to:

```
use MODX\Revolution\modManagerController;

class SystemRefreshSiteManagerController extends modManagerController {

    public function process(array $scriptProperties = []) {

        /* invoke OnBeforeCacheUpdate event */
        $this->modx->invokeEvent('OnBeforeCacheUpdate');

        $results= [];
        $this->modx->cacheManager->refresh([], $results);

        /* invoke OnSiteRefresh event */
        $this->modx->invokeEvent('OnSiteRefresh');

        $num_rows_pub = isset($results['publishing']['published']) ? $results['publishing']['published'] : 0;
        $num_rows_unpub = isset($results['publishing']['unpublished']) ? $results['publishing']['unpublished'] : 0;
        $this->modx->smarty->assign('published',$this->modx->lexicon('refresh_published', ['num' => $num_rows_pub]));
        $this->modx->smarty->assign('unpublished',$this->modx->lexicon('refresh_unpublished', ['num' => $num_rows_unpub]));

        $this->modx->smarty->assign('results', $results);

        $this->checkFormCustomizationRules();
        return $this->modx->smarty->fetch('system/refresh_site.tpl');
    }

    public function checkPermissions()
    {
        return $this->modx->hasPermission('empty_cache');
    }

    public function getPageTitle()
    {
        return $this->modx->lexicon('empty_cache');
    }

    public function loadCustomCssJs()
    {
 		return;
    }

    public function getTemplateFile()
    {
        return;
    }
}
```

opengeek:
### What does it do?

Removes the unused `system/refresh_site` manager controller path:

  - Deletes `manager/controllers/default/system/refresh_site.php`
  - Deletes `manager/templates/default/system/refresh_site.tpl`

Lexicon entries `refresh_site` / `refresh_site_desc` (in `core/lexicon/en/topmenu.inc.php`) are retained — they're used as the **Clear Cache** topmenu item's label and description, not as path references.

### Why is it needed?

The controller is dead code, unreachable in-tree:

  - The topmenu **Clear Cache** entry in `_build/data/transport.core.menus.php` uses `'action' => ''` with `'handler' => 'MODx.clearCache(); return false;'` — it routes to the `System/ClearCache` connector processor, not to this controller.
  - No PHP `include`/`require` references the file, no JS or URL constructions reference `?a=system/refresh_site`, and no other code references the would-be `SystemRefreshSiteManagerController` class.
  - The only entrypoint was direct URL navigation to `/manager/?a=system/refresh_site`, which has been broken since the `modManagerController` loader started requiring `.class.php` files.

PR #16827 proposed converting the script-style file into a controller class to fix the broken URL. Triage determined the maintenance surface isn't worth preserving an entrypoint with no callers and no menu route. Closing #16827 in favor of retirement.

### How to test

1. On a fresh 3.x install, click **Manage → Clear Cache** from the topmenu — confirm the cache is cleared (`OnBeforeCacheUpdate` / `OnSiteRefresh` still fire from the connector processor path).
2. Visit `/manager/?a=system/refresh_site` directly — confirm the response is a clean "unknown action" from `modManagerResponse` rather than a fatal.
3. Confirm no other manager screens regress (search the manager UI for any incidental references — none expected).

### Related issue(s)/PR(s)

Resolves #16282 (by retirement rather than fix). Supersedes #16827.


Ibochkarev (approved):

smg6511 (approved):
Confirmed all works as expected. No found references to controller.

JoshuaLuckers:
Shouldn’t there be an upgrade script that removes these files as well?

opengeek:
> Shouldn’t there be an upgrade script that removes these files as well?

Maybe? IDK. I'm torn since I generally use with git and removing files in a script feels redundant.

JoshuaLuckers:
> > Shouldn’t there be an upgrade script that removes these files as well?
> 
> Maybe? IDK. I'm torn since I generally use with git and removing files in a script feels redundant.

If they were part of a release, we used to remove them during setup. 
E.g: https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/common/3.0.0-cleanup-files.php

opengeek:
> > > Shouldn’t there be an upgrade script that removes these files as well?
> > 
> > 
> > Maybe? IDK. I'm torn since I generally use with git and removing files in a script feels redundant.
> 
> If they were part of a release, we used to remove them during setup. E.g: https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/common/3.0.0-cleanup-files.php

I recall — though that was specifically cleaning up old 2.x paths abandoned in 3.x. 🤔

smg6511:
But, there's nothing that would remove them in an typical upgrade situation otherwise, right?

opengeek:
> But, there's nothing that would remove them in an typical upgrade situation otherwise, right?

There's not, but we've never established a standard for this other than the one cleanup script for 3.0.0 release that I can find.

opengeek:
Further, it looks like the way it was handled in that script would not work if you used the advanced installer to put the manager in a different location (or manually moved it to a different location).

JoshuaLuckers:
I wasn’t totally sure what we used to do with left over files in the past, it has been a while for me!

JoshuaLuckers (approved):

PR #16951, opened 2026-05-07
Update FC Set grids - Correct group headers config

### LINKED ISSUE #16948: Additional translation interpolation in grouping templates
### Summary
Two grouping templates in `manager/assets/modext/widgets/fc/modx.panel.fcset.js` use the same template-literal-with-interpolated-`_()` pattern that #16823 / #16830 fixed in `MODx.grid.GridBase`, but with **double quotes** around the interpolation. They are dormant — none of today's translations of `field`/`fields`/`tv`/`tvs` contain a `"` — but a future translation that does will reproduce the same `Uncaught SyntaxError: missing ) in parenthetical` crash on the Form Customization Set editor.

The two sites:

- `manager/assets/modext/widgets/fc/modx.panel.fcset.js:289`

  ```js
  groupTextTpl: `{[values.rs[0].data.tab]} ({[values.rs.length]} {[values.rs.length > 1 ? "${_('fields')}" : "${_('field')}"]})`,
  ```

- `manager/assets/modext/widgets/fc/modx.panel.fcset.js:454`

  ```js
  groupTextTpl: `{group} ({[values.rs.length]} {[values.rs.length > 1 ? "${_('tvs')}" : "${_('tv')}"]})`,
  ```

Both panels build their own `groupTextTpl` directly rather than going through `MODx.grid.GridBase`'s grouping config, so the fix from #16831 (`MODx.util.Format.encodeQuotes`) does not apply here automatically.

### Step to reproduce
Today, the bug is latent — there is no in-tree translation that triggers it. To reproduce manually:

1. Add a translation override (or temporarily edit a lexicon entry) so that one of `field`, `fields`, `tv`, or `tvs` contains a `"` character, e.g.:

   ```php
   $_lang['tvs'] = 'TVs (e.g. "metaDescription")';
   ```

2. Open the manager in that language.
3. Edit any element with a Form Customization Set, or open the TV grouping panel where the FC Set rendering is shown.
4. Observe the JavaScript console.

### Observed behavior
`Uncaught SyntaxError: missing ) in parenthetical` (or equivalent), originating from the `groupTextTpl` evaluation. The grouping panel fails to render and the editor is unusable for that locale, identical in mechanism to #16823 / #16830.

### Expected behavior
The grouping header renders correctly with the translated label, regardless of whether the translation contains `"` or `'`.

### Suggested fix

Scope of this issue covers both the two known sites and an exhaustive sweep:

**1. Fix the two known sites** in `modx.panel.fcset.js`. Two options, in order of preference:

a. **Apply `MODx.util.Format.encodeQuotes(..., 'double')`** at both sites, mirroring the pattern introduced by #16831:

```js
groupTextTpl: `{[values.rs[0].data.tab]} ({[values.rs.length]} {[values.rs.length > 1 ? "${MODx.util.Format.encodeQuotes(_('fields'), 'double')}" : "${MODx.util.Format.encodeQuotes(_('field'), 'double')}"]})`,
```

b. **Refactor these grids to inherit `MODx.grid.GridBase`'s grouping config** so the encoding is centralized and any future fix to the base applies here automatically.

**2. Sweep for other latent sites.** The same crash pattern can hide anywhere a JS template literal interpolates `_()` (or any author-controlled string) inside quote delimiters. A spot-check during the #16831 review did not find more, but the sweep was not exhaustive. Suggested patterns to grep for under `manager/assets/modext/`:

- `groupTextTpl` constructions using template literals (backticks) with `${_(...)}` inside `'...'` or `"..."` delimiters.
- Any other XTemplate / Ext template strings built from template literals that interpolate `_()`, `config.*Text`, or other lexicon-derived values inside quote delimiters.
- Existing `htmlEncode` / `{group:htmlEncode}` usages (idiomatic alternative path) — note these for consistency, but they are already safe.

Apply the same fix (option **1a** or **1b**) to any additional sites found.

### Environment
- MODX 3.x (head)
- Affected file: `manager/assets/modext/widgets/fc/modx.panel.fcset.js`
- Browser: any (JavaScript syntax error is parse-time, not runtime-conditional)
- Discovered during review of #16831; see context there for the underlying mechanism.

### Related
- Fixed sibling: #16831
- Original reports: #16823, #16830

smg6511:
### What changed and why

- Removed custom `groupTextTpl` for two grid configs, deferring to the already-established base config
- Make columns with primarily unique values ungroupable
- Fix the rendering when grouping by visibility
- Update Lexicons to make display consistent (i.e., previously the plural form for the TVs grid read "Template Variables" and the singular "TV;" added entries for abbreviated and long forms of this term, changing the singular in this grid to "Template Variable."

Aside from minor enhancements, the change is to guard against future problems as noted in the related issue.

### How to test

1. Clear browser and MODX caches
2. Visit a Form Customization Set (preferably with multiple entries) to verify grids operate as expected.
3. Change the Lexicon via the manager for the default topic's `tvs_long` value to `Template's Variables` (adding a single quote as shown or double quote if you wish, with or without backslash escape), refresh the cache and ensure the Template Variables grid (3rd tab) still displays as expected. Note that this text is used in each group's header, _e.g._, “Category: None (2 Template Variables)” or, in the case of the first tab's grid, “Region (ID): modx-page-settings-box-left (6 Fields)” if changing the `field` or `fields` Lexicon
4. Extra credit: Manually edit the actual Lexicon file entry for `tvs_long` and add an escaped single quote in the value, again clearing cache and verifying grid operation. I add this as the custom db entries might already be encoded (I didn't look into this, but I wouldn't be surprised if they are)

### Related issue(s)/PR(s)

Resolves #16948

### Compatibility notes

n/a

n/a

### Test coverage

n/a for Extjs

### Contributors

@opengeek 

### AI tool use

n/a


manager/assets/modext/widgets/fc/modx.panel.fcset.js
```diff
@@ -279,17 +282,19 @@ MODx.grid.FCSetFields = function(config = {}) {
         ],
         autoHeight: true,
         grouping: true,
-        groupBy: 'tab_rank',
+        groupBy: 'tab',
```
Ibochkarev, line 285: This is the one I’d change. `groupBy: 'tab'` fixes the header text but loses rank ordering.

Keep `groupBy: 'tab_rank'`, and on the `tab_rank` column add something like:

```js
groupRenderer: function(v, unused, record) {
    return record.data.tab;
}
```

Base `groupTextTpl` will then show the tab id via `{text}`, `singleText`/`pluralText` still apply, and groups stay in form order. Also means `hideGroupedColumn` keeps hiding `tab_rank` instead of the Region (ID) column.

manager/assets/modext/widgets/fc/modx.panel.fcset.js
```diff
@@ -279,17 +282,19 @@ MODx.grid.FCSetFields = function(config = {}) {
...
         plugins: [this.vcb],
         stateful: false,
         remoteSort: false,
         sortBy: 'rank',
         sortDir: 'ASC',
         hideGroupedColumn: true,
-        groupTextTpl: `{[values.rs[0].data.tab]} ({[values.rs.length]} {[values.rs.length > 1 ? "${_('fields')}" : "${_('field')}"]})`,
+        singleText: _('field'),
+        pluralText: _('fields'),
```
Ibochkarev, line 293: Good call setting these. Without them GridBase falls back to `record`/`records`, which is what these grids have effectively been showing since the custom `groupTextTpl` wasn’t wired into `groupingConfig`.

manager/assets/modext/widgets/fc/modx.panel.fcset.js
```diff
@@ -260,7 +260,10 @@ MODx.grid.FCSetFields = function(config = {}) {
         header: _('visible'),
         dataIndex: 'visible',
         width: 40,
-        sortable: false
+        sortable: false,
+        groupRenderer: function(v) {
+            return v ? _('yes') : _('no');
+        }
     });
```
Ibochkarev, line 267: This is the right Ext pattern for boolean grouping. Same on the TVs grid below.

core/lexicon/en/default.inc.php
```diff
@@ -600,3 +599,9 @@
 $_lang['tv_description_desc'] = 'Usage information for this TV shown next to its caption in Resource editing pages and as a tooltip in the Elements tree.';
 $_lang['tv_elements'] = 'Input Option Values';
 $_lang['tv_elements_short_desc'] = 'Defines the selectable options for this TV, which may be manually entered or built with a one-line <a href="https://docs.modx.com/current/en/building-sites/elements/template-variables/bindings/select-binding" target="_blank">database query</a>.';
+
+// Temporarily match old keys to new ones to ensure compatibility
+$_lang['clear_cache_on_save_msg'] = $_lang['clear_cache_on_save_desc'];
+$_lang['rrg_drag'] = $_lang['resource_groups_panel_desc'];
+$_lang['tv'] = $_lang['tv_abbr'];
+$_lang['tvs'] = $_lang['tvs_long'];
```
Ibochkarev, line 607: EN-only + aliases looks right for this repo (Crowdin owns the rest). Minor nit: `tvs_abbr` isn’t referenced anywhere yet — fine if you want the set complete, otherwise it can wait until something needs it.

Ibochkarev (changes requested):
Nice direction overall — dropping the custom `groupTextTpl` and riding GridBase’s `singleText` / `pluralText` path is the right fix for #16948. That’s the code-judo version of the issue (option 1b), not another `encodeQuotes` copy-paste.

One blocker before this lands: `groupBy: 'tab'` on FCSetFields. The old `groupBy: 'tab_rank'` kept groups in region-rank order; the custom tpl only existed to show the tab id in the header. Grouping by the tab string sorts alphabetically by id, so regions can reorder relative to the form. Please keep `groupBy: 'tab_rank'` and put a `groupRenderer` on the `tab_rank` column that returns `record.data.tab`. Then you still get the base tpl + `encodeQuotes`, and `hideGroupedColumn` keeps behaving as before.

Visibility `groupRenderer` and `groupable: false` on the high-cardinality columns look good. Lexicon split into abbr/long with the old-key aliases is fine for EN/Crowdin.

Happy to approve once the Fields grouping order is restored.

core/lexicon/en/default.inc.php
smg6511: I'd like to keep the set even though `tvs_abbr` isn't assigned yet. There is one other instance that can be dropped and point to this new key (_e.g._, `search_resulttype_tvs` [topmenu.inc.php]), which could be done in a separate PR that handles other Lexicons-related clean up.

manager/assets/modext/widgets/fc/modx.panel.fcset.js
smg6511: Reverted default groupby. When I originally did this I was looking at the rank as not being very useful (because it's not changeable in the UI), but I see what you mean re the field order on the page. I didn't change the rendering though, as that would introduce an inconsistency.

Ibochkarev: Thanks

core/lexicon/en/default.inc.php
Ibochkarev: Ok, thanks

Ibochkarev (approved):

Ibochkarev:
### Tested MODX 3.2.1-dev / PHP 8.2 (Docker) + grunt build.

### Result: fixes the issue
- FC Set Fields/TVs grids group headers OK
- Lexicon with ' and " in tvs_long/tv_long: no JS SyntaxError; headers render

**Fields tab** — 31 fields, grouped by `tab_rank`; headers e.g. `Region Rank: 1 (6 Fields)`:

Fields group headers

**TVs tab (default lexicon)** — 3 TVs; headers `Category: None (1 Template Variable)`, `Category: PR16951 (2 Template Variables)`:

TVs category groups

**Lexicon with apostrophe** (`Template's Variables`) — no JS errors; headers show apostrophe; `groupTextTpl` uses `&#39;`:

TVs apostrophe lexicon

**Group by Visible** — `Visible: Yes (3 Template's Variables)`:

TVs visibility grouping

**Lexicon with double quotes** (`TVs (e.g. "metaDescription")`) — no SyntaxError; headers render with quotes:

TVs double-quote lexicon

**Profiles list** — test profile present:

FC profiles

PR #16956, opened 2026-05-27
Fix a typo in the download success message

Jako:
### What changed and why

Fix a typo in the download success message.

### How to test

The current key of the lexicon call starts with a $ sign. This has to be removed to retrieve the right lexicon entry.


Ibochkarev (approved):

smg6511 (approved):
Yup, good catch.

Mark-H (approved):

PR #16316, opened 2022-12-02
Fix parsing long template tags under some conditions [#16263]

Mark-H:
### What does it do?

Improves the regex pattern used in modParser::collectElementTags() to avoid hitting a backtrack limit under some environments (PHP 7.2, 7.3). Also adds a log message in case an error is detected. 

Credit to @JoshuaLuckers for providing the patch.

### Why is it needed?

Make sure long tags can be parsed correctly and make debugging easier if something does fail. Please see #16263 for more details.

### How to test

Please see #16263 for more details. It's worth noting I've so far seen 3 people refer to the fix in the issue as fixing the issue they had. As with all parser tweaks, trying it out on different real sites would be preferred to ensure this doesn't introduce any new issues. 

[Here's a CI run of the unit test prior to applying the fix](https://github.com/Mark-H/revolution/actions/runs/3597851277/jobs/6060109440), to show the unit test and fix work. 

Locally I've also been able to reproduce it by running the test on 7.2 before:

``` 
➜  3.x git:(issue-16263) ✗ php core/vendor/bin/phpunit  _build/test/Tests/Model/modParserTest.php
PHPUnit 8.5.31 by Sebastian Bergmann and contributors.

..........................................................[1]    78526 segmentation fault  /Applications/MAMP/bin/php/php7.2.34/bin/php -c  core/vendor/bin/phpunit 
```

and after applying the fix:

```
➜  3.x git:(issue-16263) ✗ php core/vendor/bin/phpunit  _build/test/Tests/Model/modParserTest.php
PHPUnit 8.5.31 by Sebastian Bergmann and contributors.

...........................................................       59 / 59 (100%)

Time: 181 ms, Memory: 14.00 MB

OK (59 tests, 70 assertions)
```

(I'm doing `php core/vendor/bin/phpunit` because otherwise php can't get to mysql on my local machine.)

### Related issue(s)/PR(s)

Fixes #16263

PR #16957, opened 2026-05-28
Fix modInputFilter root-level backticks in modifier values

### LINKED ISSUE #16942: Parser errors introduced in MODX 3.2.0
### Summary
Nesting in conjunction with output modifiers breaks in MODX3.2.0 in cases that worked in 2.8.8

### Step to reproduce

This is an example to create a filtering statement only in case the placeholder `[[+filter]]` is not empty:
```
[[+filter:isnot=``:then=`
  &tvFilters=`news-kategorie==%[[+filter:commaToString=`%||news-kategorie==%`]]%`
`]]
```

### Observed behavior
The plain content of the placeholder was printed:
```Pressemitteilung```

### Expected behavior
The expected output should be:
```
&tvFilters=`news-kategorie==%Pressemitteilung%`
```
This was also the observed output in MODX 2.8.8

### Environment
MODX 3.2.0, PHP 8.4

### PR DESCRIPTION
**Pull request by opengeek**
### What changed and why

Fixes a long-standing tokenization regression in `modInputFilter` where root-level backticks inside a modifier value (e.g. `then=`&tvFilters=`...``) were closing the outer modifier prematurely, shredding the remainder into bogus commands. This broke complex output modifiers reported in #16942.

The regression has been present in every 3.x release since v3.0.0-alpha3. It was introduced by [PR #14458](https://github.com/modxcms/revolution/pull/14458) ("Fix the parser when using nested output filters"), which replaced the 2.x recursive regex with a different recursive regex that mishandled property-style backticks. [PR #16288](https://github.com/modxcms/revolution/pull/16288) ("Refactor input filter parsing to properly support nested tags") later rewrote the regex as a character-scan, fixing other nested-tag cases but preserving the same blind spot for property-value backticks. The 2.x parser handled this shape correctly.

This PR adds property-value depth tracking alongside the existing `[[..]]` depth tracking inside `modInputFilter::filter()`. A backtick after `=` inside a modifier opens a property value; subsequent root-level backticks pair against that property value rather than closing the outer modifier. The outer modifier closes only when the property stack is empty and the previous character is not `=`.

Root cause is localized to one function. No changes to `modParser`, `modOutputFilter`, or `parsePropertyString()` were needed — downstream behaviors already handle correctly-tokenized output.

### How to test

Run the focused parser/filter suites:

```bash
core/vendor/bin/phpunit -c _build/test/phpunit.xml \
  _build/test/Tests/Model/Filters/modInputFilterTest.php \
  _build/test/Tests/Model/modParserTest.php
```

Before the fix: 5 failures, all keyed `issue-16942-*`.
After the fix: 97 tests, 0 failures.

Broader sanity check:

```bash
core/vendor/bin/phpunit -c _build/test/phpunit.xml --group Filters
```

125 tests, 0 failures.

### Related issue(s)/PR(s)

Resolves #16942.

Related history:
- PR #16288 — nested-tag input-filter parsing.
- PR #16316 — `collectElementTags()` backtrack and long-tag stability.
- Community thread examples of mosquito-style conditionals and chunk/snippet output modifiers (captured as green regression guards in the test additions).

### Compatibility notes

Universal change. `modInputFilter::filter()` is core parser code with no platform-specific dependencies. PHP 8.1+ compatibility preserved (no new language features). No database schema changes; 2.x → 3.x upgrade path unaffected.

No public API signatures or return types change. The behavior change is a bug fix: complex output modifiers that worked before 3.2.0 work again. Callers that somehow depended on the broken tokenization (treating an inner property-value backtick as a modifier delimiter) would behave differently, but the documented and historical contract is what this fix restores. Safe for patch-level consumers.

### Test coverage

Adds 16 regression cases across `modInputFilterTest` and `modParserTest`:

- Input-filter tokenization for complex output modifiers, including the #16942 patterns (red before fix, green after).
- Tag collection regression guards covering PR #16288 / PR #16316 nesting and long-tag patterns.
- End-to-end parser output for community-thread mosquito conditionals, nested placeholders in `then`/`else`, uncached prefix variants, and the #16942 manifestations.
- Snippet property delivery for the conditional-injection pattern from #16942 comments, including a sanity case and a placeholder-as-property variant.

### Contributors

Thanks to @frischnetz for reporting #16942 with reproduction cases that drove the regression tests and fix.

### AI tool use

Used Claude Code (Claude Opus 4.7, 1M context) to:
- Survey the parser and filter codebase to map the call paths for `modInputFilter::filter()`, `modParser::processElementTags()`, `modParser::parsePropertyString()`, and `modScript::loadScript()`.
- Draft the regression test battery, including the dataset shapes for `collectElementTags`, `processElementTags`, and the new snippet-execution test.
- Trace the #16942 manifestation through the input filter character scan and identify property-value backtick nesting as the missing concept.
- Draft the fix in `modInputFilter::filter()` and verify it against the test suite.

Each step was reviewed and validated before commit; PHPUnit runs were observed end-to-end. The 5 red-then-green dataset transitions, the absence of new failures elsewhere, and the surgical scope of the fix were all confirmed before submission.

### DIFF
--- a/_build/test/Tests/Model/Filters/modInputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modInputFilterTest.php
@@ -172,6 +172,31 @@ class modInputFilterTest extends MODxTestCase
                     '',
                 ],
             ],
+
+            'issue-16957-replace-with-equals-separator' => [
+                '+phone:replace=` ==`:before=`<a href="tel:`:after=`">[[+phone]]</a>`',
+                '+phone',
+                ['replace', 'before', 'after'],
+                [
+                    ' ==',
+                    '<a href="tel:',
+                    '">[[+phone]]</a>',
+                ],
+            ],
+
+            'issue-16957-replace-literal-equals' => [
+                '+value:replace=`=== `',
+                '+value',
+                ['replace'],
+                ['=== '],
+            ],
+
+            'question-mark-first-property-opener' => [
+                '+x:then=`SomeSnippet?prop=`val``',
+                '+x',
+                ['then'],
+                ['SomeSnippet?prop=`val`'],
+            ],
         ];
     }
 }
--- a/_build/test/Tests/Model/Filters/modInputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modInputFilterTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -11,7 +12,6 @@
 */
 
 namespace MODX\Revolution\Tests\Model\Filters;
-
 
 use MODX\Revolution\Filters\modInputFilter;
 use MODX\Revolution\modFieldTag;
--- a/_build/test/Tests/Model/Filters/modInputFilterTest.php
+++ b/_build/test/Tests/Model/Filters/modInputFilterTest.php
@@ -122,7 +122,56 @@ class modInputFilterTest extends MODxTestCase
                     '',
                     '',
                 ]
-            ]
+            ],
+
+            'community-then-with-nested-mosquito' => [
+                '+content:isnot=``:then=`[[+segment_type:is=`header`:then=`<h1>[[+content]]</h1>`:else=`<p>[[+content]]</p>`]]`',
+                '+content',
+                ['isnot', 'then'],
+                [
+                    '',
+                    '[[+segment_type:is=`header`:then=`<h1>[[+content]]</h1>`:else=`<p>[[+content]]</p>`]]',
+                ],
+            ],
+
+            'pr-16288-uncached-nested-tag-in-default' => [
+                'pagetitle:default=`[[!+fallback:default=`none`]]`',
+                'pagetitle',
+                ['default'],
+                ['[[!+fallback:default=`none`]]'],
+            ],
+
+            'pr-16316-chained-then-else-deeply-nested' => [
+                '+pagetitle:is=`a`:then=`A`:else=`[[+x:is=`b`:then=`B`:else=`[[+y:is=`c`:then=`C`:else=`D`]]`]]`',
+                '+pagetitle',
+                ['is', 'then', 'else'],
+                [
+                    'a',
+                    'A',
+                    '[[+x:is=`b`:then=`B`:else=`[[+y:is=`c`:then=`C`:else=`D`]]`]]',
+                ],
+            ],
+
+            'issue-16942-tvfilters-nested-placeholder-modifier' => [
+                '+filter:isnot=``:then=`&tvFilters=`news-kategorie==%[[+filter:commaToString=`%||news-kategorie==%`]]%``',
+                '+filter',
+                ['isnot', 'then'],
+                [
+                    '',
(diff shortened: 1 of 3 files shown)


smg6511:
@opengeek - I'm beginning to take a look. One interesting thing is that my report of failures is different than what you show here. Note that, as a baseline, I've copied in only your new unit tests so I can see the failures before applying the fix in the PR. I get 2 failures in the Input Filters test and the broad test. 

However, in a separate run of the main Parser test, I get 10 failures. If I run this same test on the current 3.x dev branch (no changes from this PR) I'm getting 5 failures.

Any reason we might get different results there?

opengeek:
> @opengeek - I'm beginning to take a look. One interesting thing is that my report of failures is different than what you show here. Note that, as a baseline, I've copied in only your new unit tests so I can see the failures before applying the fix in the PR. I get 2 failures in the Input Filters test and the broad test.
> 
> However, in a separate run of the main Parser test, I get 10 failures. If I run this same test on the current 3.x dev branch (no changes from this PR) I'm getting 5 failures.
> 
> Any reason we might get different results there?

Make sure you are using a clean install for the unit tests?

smg6511:
Let me ask you this: Is it not enough to, in the context of Github, be on the 3.x branch (or a clean branch based off of it) and run `composer update` to arrive at the right state? Or is there something else I'm missing that would affect the tests?

(Note that I'd initially forgotten to run `composer update`.)

opengeek:
> Let me ask you this: Is it not enough to, in the context of Github, be on the 3.x branch (or a clean branch based off of it) and run `composer update` to arrive at the right state? Or is there something else I'm missing that would affect the tests?
> 
> (Note that I'd initially forgotten to run `composer update`.)

I honestly don't know. I've always used a separate configuration and db locally for my clean install unit testing so that it matches the CI runs.

smg6511 (approved):
I never could get my pre PR application failure counts to line up with yours, but no matter. There were a number of failures before and none after. I also manually replicated  a handful of the solved patterns before and after as an extra measure to confirm the fix. Interestingly, at least two of the patterns referenced in the issue worked just fine without the PR fix (I didn't verify if these were a part of your new test cases though).

frischnetz:
I might have found a slight regression with the new code: I use this to remove spaces from the phone number before building a link:
```
[[+phone:replace=` ==`:before=`<a href="tel:`:after=`">[[+phone]]</a>`]]
```
AFAIK the the new code looks for 
```
=`
``` 
so it gets confused and returns this:
```
040`:before=`<a href="tel:1234567890">040 1234567890</a>
```
instead of 
```
<a href="tel:0401234567890">040 1234567890</a>
```

opengeek:
@frischnetz — thanks for reporting this regression. See if the changes in 8455a9f resolve it for you?

jaygilmore:
Heya, @frischnetz. Thanks as always for helping test this. I'm wondering if you might have a chance to take a look to see if the  changes Jason did above in his last comment resolve this for you? It would be good to be able to include this in a sooner-than-later patch for revo.

frischnetz:
Hey @jaygilmore and @opengeek, Thanks for looking into it! I only tested it just now and found that it works for my (simple) test cases!

PR #16958, opened 2026-05-29
Fix: Make search results and tree items display in consistent manner

smg6511:
### What changed and why

Added status information to search result data and a new method to the front end XTemplate to match how the tree displays the various states of Resources and Elements. Some specific updates of interest:

- Displays custom template icon when applicable
- Displays folder when Resource is a container

#### Design Note
I only touched css minimally here, but will propose a number of style enhancements in a separate PR.

### How to test

Test in an environment with many Resources and Elements, having various published/deleted/disabled states and compare the search display to the tree display. The two should match. 

Note that some Extras override the tree icon (_e.g._, Collections, Gallery), but do not override the search one; that disconnect would need to be addressed by the relevant Extras’ maintainers.

#### Example Screenshots
On the left are a couple clips of my tree; on the right, a few search result examples showing various states. You can spot where specific ones match up.

search and tree

### Related issue(s)/PR(s)

This is a different approach than and supersedes the proposed change in #16659. 

### Compatibility notes

n/a

None

### Test coverage

n/a

### Contributors

n/a

### AI tool use

n/a

### DIFF
--- a/_build/templates/default/sass/_uberbar.scss
+++ b/_build/templates/default/sass/_uberbar.scss
@@ -158,6 +158,27 @@
         color: $searchResultsColor;
       }
     }
+
+    &.disabled,
+    &.unpublished {
+      em {
+        opacity: 1;
+      }
+    }
+
+    &.disabled {
+      color: $unpublished;
+      i.icon {
+        color: $unpublished;
+      }
+    }
+
+    &.deleted {
+      a {
+        text-decoration: $delTextDeco;
+      }
+    }
+
   }
 
   .icon-user {
--- a/core/src/Revolution/Processors/Search/Search.php
+++ b/core/src/Revolution/Processors/Search/Search.php
@@ -161,7 +161,7 @@ class Search extends Processor
                 'name' => $this->modx->hasPermission('tree_show_resource_ids')
                     ? $record->get('pagetitle') . ' (' . $record->get('id') . ')'
                     : $record->get('pagetitle'),
-                '_action' => 'resource/update&id=' . $record->get('id'),
+                'resultLinkAction' => 'resource/update&id=' . $record->get('id'),
                 'description' => $record->get('description'),
                 'type' => static::TYPE_RESOURCE . 's',
                 'class' => $record->get('class_key'),
@@ -214,7 +214,7 @@ class Search extends Processor
             $data = [
                 'name' => $record->get($nameField),
                 'description' => $record->get($descriptionField),
-                '_action' => 'element/' . $type . '/update&id=' . $record->get('id'),
+                'resultLinkAction' => 'element/' . $type . '/update&id=' . $record->get('id'),
                 'type' => $type . 's',
                 'class' => $class,
                 'attributes' => $attributes
@@ -265,7 +265,7 @@ class Search extends Processor
             $data = [
                 'name' => $record->get('username'),
                 'description' => $record->get('fullname') . ' / ' . $record->get('email'),
-                '_action' => 'security/user/update&id=' . $record->get('internalKey'),
+                'resultLinkAction' => 'security/user/update&id=' . $record->get('internalKey'),
                 'type' => static::TYPE_USER . 's',
                 'attributes' => $attributes
             ];
--- a/core/src/Revolution/Processors/Search/Search.php
+++ b/core/src/Revolution/Processors/Search/Search.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
@@ -9,7 +10,6 @@
  */
 
 namespace MODX\Revolution\Processors\Search;
-
 
 use MODX\Revolution\modChunk;
 use MODX\Revolution\modContext;
@@ -117,14 +117,14 @@ class Search extends Processor
         $c->select('modTemplate.icon as icon');
 
         $querySearch = [
-            'modResource.pagetitle:LIKE' => '%' . $this->query .'%',
-            'OR:modResource.longtitle:LIKE' => '%' . $this->query .'%',
-            'OR:modResource.alias:LIKE' => '%' . $this->query .'%',
-            'OR:modResource.description:LIKE' => '%' . $this->query .'%',
-            'OR:modResource.introtext:LIKE' => '%' . $this->query .'%',
+            'modResource.pagetitle:LIKE' => '%' . $this->query . '%',
+            'OR:modResource.longtitle:LIKE' => '%' . $this->query . '%',
+            'OR:modResource.alias:LIKE' => '%' . $this->query . '%',
+            'OR:modResource.description:LIKE' => '%' . $this->query . '%',
+            'OR:modResource.introtext:LIKE' => '%' . $this->query . '%',
         ];
         if ($this->searchInContent()) {
-            $querySearch['OR:modResource.content:LIKE'] = '%' . $this->query .'%';
+            $querySearch['OR:modResource.content:LIKE'] = '%' . $this->query . '%';
         }
         $querySearch['OR:modResource.id:='] = $this->query;
         $queryContext = [
@@ -178,10 +178,10 @@ class Search extends Processor
         $c = $this->modx->newQuery($class);
         $querySearch = [
             $nameField . ':LIKE' => '%' . $this->query . '%',
-            'OR:' . $descriptionField . ':LIKE' => '%' . $this->query .'%',
+            'OR:' . $descriptionField . ':LIKE' => '%' . $this->query . '%',
         ];
         if ($this->searchInContent() && !empty($contentField)) {
-            $querySearch['OR:' . $contentField . ':LIKE'] = '%' . $this->query .'%';
+            $querySearch['OR:' . $contentField . ':LIKE'] = '%' . $this->query . '%';
         }
         $querySearch['OR:id:='] = $this->query;
         $c->where($querySearch);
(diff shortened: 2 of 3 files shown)


smg6511:
@Ibochkarev @mkschell - Hey guys, can you check this out in the near future? Thx ;-)

Ibochkarev:
> @Ibochkarev @mkschell - Hey guys, can you check this out in the near future? Thx ;-)

Hi!

I'll take a look today.

core/src/Revolution/Processors/Search/Search.php
```diff
@@ -137,18 +138,30 @@ protected function searchResources()
...
-            $this->results[] = [
+            $attributes = [
+                'isFolder' => $record->get('isfolder'),
+                'isStatic' => $record->get('class_key') === modStaticResource::class,
+                'status' => [
+                    'published' => $record->get('published'),
+                    'deleted' => $record->get('deleted'),
+                    'hidemenu' => $record->get('hidemenu')
+                ]
+            ];
```
Ibochkarev, line 152: Could we lean on `$record->getStatusClasses()` here instead of rebuilding published / deleted / hidemenu by hand?

Same idea for icons via `getIconClasses()` — that path already covers folder semantics (`childrenCount > 0 || isfolder`), content-type icons, and the `mgr_tree_icon_*` settings the tree uses.

Rough shape:
```php
'cls' => implode(' ', $record->getStatusClasses()),
'icon' => /* normalize prefix from getIconClasses() */,
```
Then the client does not need this `attributes` bag for resources.

core/src/Revolution/Processors/Search/Search.php
```diff
@@ -179,14 +192,33 @@ protected function searchElements($class, $type = '', $nameField = 'name', $desc
...
+        $isTemplate = $class === modTemplate::class;
+
         /** @var modElement $record */
         foreach ($collection as $record) {
-            $this->results[] = [
+            $attributes = [
+                'isElement' => true,
+                'isStatic' => $record->get('static'),
+                'status' => [
+                    'disabled' => $record->get('disabled') ?? false
```
Ibochkarev: `disabled` only really lives on plugins. `Element\GetNodes` gates it with `$elementClassKey === modPlugin::class` and emits `element-node-disabled`.

Calling `get('disabled')` on chunks / snippets / tvs / templates papers over that boundary. Worth matching GetNodes here (and using the same class name), plus treating static the way the tree does — via `icon-file-code-o` — instead of a `.static` status class the tree never paints.

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -40,10 +40,31 @@ MODx.SearchBar = function(config) {
...
+                            case 'users':
+                                classes += status.disabled ? ' disabled' : '' ;
+                                break;
+                            // no default
+                        }
+                        classes += status.deleted ? ' deleted' : '' ;
+                        classes += values?.attributes?.isStatic ? ' static' : '' ;
+                    }
+                    return classes;
+                },
```
Ibochkarev, line 66: If the processor hands back a ready `cls` string, this helper can go away entirely.

As written it also invents class names the tree does not use (`disabled` vs `element-node-disabled`, and `.static` as a status class). That is how search and tree drift apart even when both "have status info".

manager/assets/modext/widgets/core/modx.searchbar.js
```diff
@@ -58,7 +79,7 @@ MODx.SearchBar = function(config) {
                     if (values.class) {
                         switch (values.class) {
                             case 'MODX\\Revolution\\modDocument':
-                                return 'file';
+                                return values.attributes.isFolder ? 'folder' : 'file' ;
```
Ibochkarev, line 83: Tree folder icons come from `getIconClasses()` (`parent-resource` when there are children or `isfolder`). Hardcoding `folder` only when `isfolder` is set will diverge when `auto_isfolder` is off or a container already has kids.

If icon resolution moves server-side, this branch disappears with it.

_build/templates/default/sass/_uberbar.scss
```diff
@@ -158,6 +158,27 @@
...
+      i.icon {
+        color: $unpublished;
+      }
+    }
+
+    &.deleted {
+      a {
+        text-decoration: $delTextDeco;
+      }
+    }
```
Ibochkarev, line 180: Tree styles for these states live in `_tree.scss` (`$unpublished`, `$hidden`, `$delTextColor`, `element-node-disabled`).

Here `.unpublished` only bumps `em` opacity, `.hidemenu` gets no rules, and `.deleted` skips the color token. If the goal is visual parity with the tree, I would reuse those tokens / class names in this PR — or keep this PR to icons/folder and leave status styling for the CSS follow-up you mentioned.

Ibochkarev (changes requested):
Hey Jim — thanks for picking this up. Matching search results to the tree is the right problem.

I looked at how `Resource\GetNodes` / `Element\GetNodes` already build `cls` and `iconCls`, and I think we can get much closer to the tree with less new surface area. Right now `Search.php` ships a nested `attributes` bag and `searchbar.js` re-derives CSS from it. That second presentation path will drift the next time someone tweaks tree status styling.

Ask: have the processor emit ready-to-use `cls` (and a resolved `icon`) via `modResource::getStatusClasses()` / `getIconClasses()`, and follow the same conventions `Element\GetNodes` already uses (`element-node-disabled`, static via icon). Then the XTemplate can print `{cls}` and we can drop `getStatusClasses` plus most of the `attributes` envelope.

Line notes below. One more thing: `_uberbar.scss` does not yet mirror `_tree.scss` for unpublished / hidemenu / deleted, so a lot of the new status classes barely show even when JS attaches them. Either bring those tokens/class names into this PR, or narrow the claim to icons/folder and leave status styling for the follow-up.

Happy to chat if that framing feels off.

manager/assets/modext/widgets/core/modx.searchbar.js
smg6511: Keep in mind that this was originally only returning `'file'` anyway. This at least covers correctly displaying a doc as a folder when `isfolder` is on. The question this raises for me is what is the purpose of setting a document's `isfolder` to off when it has children?

_build/templates/default/sass/_uberbar.scss
smg6511: The small divergences are done on purpose to compensate for how small the text is in the search result. So for the most part search is picking up style info from the tree definitions.

smg6511:
> I think we can get much closer to the tree with less new surface area

Yes, normally I'd be leaning on/creating more shared code. It'd probably help to know the logic behind what I'm doing here:

1. Keep as much of the view decisions/specs out of the model as possible, passing on status information instead of compiling class names there. Aside from a more strict separation of concerns, this also provides more easily reusable data for the view. I'd begun making changes like this a while ago (moving permissions specs into object data instead of marking them with css class names; it's a change that's being incrementally done).
2. I'll be working on a proposed tree design/feature update that would end up implementing its items' statuses based on how this PR does. At that point, the goal of more parity will be more clearly met.

Keep this in mind when reviewing my responses to your requests ;-)

core/src/Revolution/Processors/Search/Search.php
smg6511: See my [comment below](https://github.com/modxcms/revolution/pull/16958#issuecomment-5242383893).

core/src/Revolution/Processors/Search/Search.php
smg6511: I made some adjustments in the new commit re setting disabled. Also, see my [general note below](https://github.com/modxcms/revolution/pull/16958#issuecomment-5242383893).

manager/assets/modext/widgets/core/modx.searchbar.js
smg6511: Again, [see note below](https://github.com/modxcms/revolution/pull/16958#issuecomment-5242383893) ;-) The classes will eventually match after adjustments to the tree in a separate PR; and as you see I'm preferring more simple, direct class names.

Ibochkarev (approved):
Good work!

smg6511:
Thanks, Ivan. Let me know if you're good with me resolving the requests above (or if there's more to be done).

Ibochkarev:
> Thanks, Ivan. Let me know if you're good with me resolving the requests above (or if there's more to be done).

Unfortunately, I don't have permissions to use Resolve in this repo. And for some reason, I can't even close the comments I created.

You have higher permissions than me—please close everything yourself. Thanks in advance!

PS: I can just delete only comments - I think I can do that )))

rthrash:
Was working on a signoff review on this one and discovered that it appears to be ignoring the content type icons, e.g., for RSS.


Should this case be covered?

smg6511:
@rthrash - Yes, good catch Ryan. The PR now factors the CTs in. They take precedence over the template and default icons. Let me know if that makes the most sense to you; maybe there's a case for template icons being "top-dog" (?).
@Ibochkarev - Please give this another run through as well.

One other behavior change was introduced in a separate commit: Now you can search by the content type name.

rthrash (approved):
## Test Pilot result: Pass

- Pull request: [#16958](https://github.com/modxcms/revolution/pull/16958)
- Revision tested: [`3aad49fac4cc00ee5290b4db6cee154f25073942`](https://github.com/modxcms/revolution/commit/3aad49fac4cc00ee5290b4db6cee154f25073942)
- MODX: 3.2.3-pl
- PHP: 8.3.13
- Site type: Staging

### Test limitation notes

- Manager JavaScript was tested from source files; the compressed production bundle was not tested.
- Manager styling was tested with a reviewed CSS overlay; the compiled production stylesheet was not tested.

### Testing notes

The tree state now shows in the search box, including custom resource icons. Great update!

_Generated by Test Pilot_

Ibochkarev (approved):

PR #16959, opened 2026-05-29
Delete the sqlsrv database tables widget

Jako:
### What changed and why

Delete the sqlsrv database tables widget. Microsoft SQL Server support was dropped in MODX 3.0 (https://docs.modx.com/3.x/en/getting-started/upgrading-to-3.0/sqlsrv). So it should be safe to remove that widget.

Detected during checking missing and superfluous lexicon entries.


smg6511 (approved):

Ibochkarev (approved):

Mark-H:
Would be nice to also add the upgrade script that removes the file on update, i.e. a file like https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/common/3.0.0-cleanup-files.php

Jako:
The upgrade script is added.

Mark-H (approved):

opengeek:
> Would be nice to also add the upgrade script that removes the file on update, i.e. a file like https://github.com/modxcms/revolution/blob/3.x/setup/includes/upgrades/common/3.0.0-cleanup-files.php

If we are going to make this a standard requirement for upgrade (this has ONLY ever been done in the upgrade to go from 2.x to 3.x, unless I am unaware of it), then the functions being duplicated in each one of these scripts needs to be incorporated into the setup. I'm happy for us to do that, but I think this is something that needs helper functions for use in the upgrade scripts or some type of automation when not using the code from git. I'll be happy to work out the helper functions and then I can update the script here to use them, unless someone has an objection.

Issue #16955, opened 2026-05-27
[Bug]: Missing lexicon entries


Jako:
### Summary

According to _build/lexicon/checklexicon.php some lexicon entries are missing

### Steps to Reproduce

1. run `_build/lexicon/checklexicon.php` in a Git based MODX installation
2. check the content of the file `_missing.php` afterwards
3. it contains the missing lexicon entries

### Expected Behavior

No missing lexicon entries.

### Actual / Observed Behavior

Currently missing lexicon entries:

```
$_lang['alert'] = '';
$_lang['cache_err_write'] = '';
$_lang['class_err_ns'] = '';
$_lang['confirm_delete_context'] = '';
$_lang['context_duplicate_confirm'] = '';
$_lang['createedit_symlink'] = '';
$_lang['database_table_reserved'] = '';
$_lang['database_table_unused'] = '';
$_lang['element_class_ns'] = '';
$_lang['emailsubject'] = '';
$_lang['err.obj_nf'] = '';
$_lang['file_err_download_timeout'] = '';
$_lang['login_cancelled_install_in_progress'] = '';
$_lang['login_cancelled_site_was_updated'] = '';
$_lang['name_err_ns'] = '';
$_lang['namespace_err_ns_name'] = '';
$_lang['optimize_database_err'] = '';
$_lang['package_err_save'] = '';
$_lang['policy_template_err_name_ns'] = '';
$_lang['provider_err_nf'] = '';
$_lang['record'] = '';
$_lang['records'] = '';
$_lang['resource_group_err_ns_name'] = '';
$_lang['resource_group_err_save'] = '';
$_lang['resource_group_ns_name'] = '';
$_lang['resource_type_help'] = '';
$_lang['resources_active'] = '';
$_lang['set_err_remove'] = '';
$_lang['setting_manager_tooltip_delay_desc'] = '';
$_lang['setting_manager_tooltip_enable_desc'] = '';
$_lang['source_err_remove'] = '';
$_lang['source_err_remove_default'] = '';
$_lang['template_err_create'] = '';
$_lang['template_preview_description'] = '';
$_lang['tv_err_create'] = '';
$_lang['tv_err_remove'] = '';
$_lang['tv_output_type_desc'] = '';
$_lang['tv_type_desc'] = '';
$_lang['ugc_mutate'] = '';
$_lang['unknown_error'] = '';
$_lang['user_err_not_found'] = '';
$_lang['user_new'] = '';
[...]

PR #16960, opened 2026-05-29
Fix missing lexicon entries

Jako:
### What changed and why

Add missing lexicon entries and rename keys to use existing lexicon entries

### Related issue(s)/PR(s)

#16955


core/lexicon/en/context.inc.php
```diff
@@ -9,6 +9,7 @@
 $_lang['context'] = 'Context';
 $_lang['context_add'] = 'Add Context';
 $_lang['context_data'] = 'Context Data';
+$_lang['context_duplicate_confirm'] = 'Are you sure you want to duplicate this context?';
```
smg6511, line 12: This entry appears to be a legacy one—there is no confirmation dialog when duplicating a Context. The controller (context.view.php) and view in which this key was found (manager/assets/modext/sections/context/view.js) also look to be unused and should be removed as there's no such view in 3.x.
```suggestion
```

core/lexicon/en/user.inc.php
```diff
@@ -170,6 +171,7 @@
 $_lang['user_male'] = 'Male';
 $_lang['user_management_msg'] = 'Here you can choose which user you wish to edit.';
 $_lang['user_mobile'] = 'Mobile phone number';
+$_lang['user_new'] = 'Create User';
```
smg6511, line 174: This already exists via the key `new_user`; so, update the key in `widgets/security/modx.panel.user.js` to match.
```suggestion
```

core/lexicon/en/default.inc.php
```diff
@@ -30,6 +30,7 @@
 $_lang['add_tv'] = 'Create TV';
 $_lang['add_weblink'] = 'Create Weblink';
 $_lang['add_symlink'] = 'Create Symlink';
+$_lang['alert'] = 'Alert';
```
smg6511, line 33: It certainly doesn't hurt to have this but, from what I found,   there are nuances in three places where this key is used:

- `downloadFile` method in `modx.tree.directory.js` and `modx.browser.js`. Of note here is that the `MODx.msg.alert `call seems to be overridden by something and is not actually used, although the presence of the failure listener is required, even if its declared like so: `failure: { fn: Ext.emptyFn }`
- `processResponse` in `modx.tree.treeloader.js` where the msg.alert is actually used. Here however, "Error" or "Warning" is probably the better dialog title rather than "Alert," in which case we should change the key, i.e.: `MODx.msg.alert(_('error'), json.message);`

core/lexicon/en/default.inc.php
```diff
@@ -41,6 +42,7 @@
 $_lang['backup'] = 'Backup';
 $_lang['bk_manager'] = 'Backup';
 $_lang['bulk_actions'] = 'Bulk Actions';
+$_lang['cache_err_write'] = 'Cannot cache the file to download!';
```
smg6511, line 45: The class this is called in is specific to the Console, so we should be more clear about that. Further we should change the key to be more clear as well (_e.g._, `console_download_cache_err`
```suggestion
$_lang['console_download_cache_err'] = 'Could not download the file due to a problem caching the Console’s content for output!';
```

Note that there are two additional todos to complete this fix:

1. Change the key in `Processors/System/DownloadOutput.php` from `cache_err_write` to `console_download_cache_err`
2. Add an empty failure to the modx.console.js file, i.e., add the following to the `download` method's listeners: `failure: { fn: Ext.emptyFn },`; otherwise the failure message that this Lex is used in never gets shown.

core/lexicon/en/default.inc.php
```diff
@@ -58,6 +60,7 @@
 $_lang['category_remove'] = 'Delete Category';
 $_lang['chunk'] = 'Chunk';
 $_lang['chunks'] = 'Chunks';
+$_lang['class_err_ns'] = 'Class not specified.';
```
smg6511, line 63: I understand that this was the original key, but it is exclusively used in relation to Derivatives (in the current codebase) so a more specific key and message would be better:
```suggestion
$_lang['derivatives_parent_class_err_empty'] = 'Can not get this object’s child classes because its class was not specified.';
```
Again, a couple changes would be needed to the calling class (`System/Derviatives/GetList.php`) at ~L49:

1. Change the Lex key, and
2. Add a return to the statement

So, `return $this->failure($this->modx->lexicon('derivatives_parent_class_err_empty'));`

Without the return, the error will never show in the front end.

core/lexicon/en/default.inc.php
```diff
@@ -137,6 +140,7 @@
 $_lang['element_caption_new'] = 'Caption of New Element';
 $_lang['elements'] = 'Elements';
 $_lang['email'] = 'Email';
+$_lang['emailsubject'] = 'Default Email Subject';
```
smg6511, line 143: Actually, the default should just be an empty string on ~L953 in `modUser.php`. IMO it makes no sense to set some arbitrary subject for an email subject. Mail readers automatically assign a subject when it's blank (Apple Mail “No Subject”, Gmail “(no subject)”, and so on). So,
```
$mail->set(modMail::MAIL_SUBJECT, $this->xpdo->getOption('subject', $options, ''));
```

As such, I'd revert to having no Lex for this
```suggestion
```

core/lexicon/en/default.inc.php
```diff
@@ -431,6 +437,7 @@
 $_lang['resource_overview'] = 'Overview';
 $_lang['resource_settings'] = 'Resource Settings';
 $_lang['resource_type'] = 'Resource Type';
+$_lang['resources_active'] = 'Active Resources';
```
smg6511, line 440: This is only referenced in `modx.grid.resource.active.js`, but in 3.x grids do not use headers (thus the title is disregarded). We actually need to remove the title props from _all_ grids at some point. In this case, ~L4 of this js file.
```suggestion
```

core/lexicon/en/element.inc.php
```diff
@@ -34,7 +34,9 @@
 $_lang['tv_elements'] = 'Input Option Values';
 $_lang['tv_default'] = 'Default Value';
 $_lang['tv_type'] = 'Input Type';
+$_lang['tv_type_desc'] = 'Allows you to select a different input type for a TV, e.g. a text field, a dropdown field, etc.';
```
smg6511, line 37:
```suggestion
$_lang['tv_type_desc'] = 'The type of input this TV will render in the Resource editing panel (text, dropdown, date, etc.). (Default: “Text”)';
```

core/lexicon/en/element.inc.php
```diff
@@ -34,7 +34,9 @@
...
 $_lang['tv_output_type'] = 'Output Type';
+$_lang['tv_output_type_desc'] = 'Allows you to format the value of a TV to a different kind of output.';
```
smg6511, line 39:
```suggestion
$_lang['tv_output_type_desc'] = 'Applies special front-end formatting to this TV’s value, based on the specified parameters or a predefined format. (Default: “Default”)';
```

core/lexicon/en/file.inc.php
```diff
@@ -66,7 +67,9 @@
 $_lang['file_last_modified'] = 'Last Modified';
 $_lang['file_msg_download_error'] = 'Download error!';
 $_lang['file_msg_download_success'] = 'Download success!';
+$_lang['file_msg_download_timeout'] = 'Download timeout!';
 $_lang['file_name'] = 'File Name';
+$_lang['file_name_err_ns'] = 'Please specify a valid file name.';
```
smg6511, line 72: We're moving toward not using abbreviated var/key names; let's transition "ns" to "empty":
```suggestion
$_lang['file_name_err_empty'] = 'Please specify a valid file name.';
```

core/lexicon/en/login.inc.php
```diff
@@ -15,6 +15,8 @@
 $_lang['login_blocked_too_many_attempts'] = 'You have been blocked due to too many failed login attempts.';
 $_lang['login_button'] = 'Login';
 $_lang['login_cannot_locate_account'] = 'The username or password you entered is incorrect. Please check the username, re-type the password, and try again.';
+$_lang['login_cancelled_install_in_progress'] = 'Login cancelled! Install in progress.';
+$_lang['login_cancelled_site_was_updated'] = 'Login cancelled! Site was updated!';
```
smg6511, line 19: To anyone who might know: What is the meaning of this second case (~L368 in the `login.class.php` controller) that uses the `...was_updated` Lex? This message is quite ambiguous.

core/lexicon/en/policy.inc.php
```diff
@@ -56,6 +56,7 @@
 $_lang['policy_template_err_ae'] = 'A Policy Template already exists with the name `[[+name]]`. Please select another name.';
 $_lang['policy_template_err_nf'] = 'Policy Template not found.';
 $_lang['policy_template_err_ns'] = 'Policy Template not specified.';
+$_lang['policy_template_err_name_ns'] = 'Policy Template name not specified.';
```
smg6511, line 59: For empty value errors, we just use the generic `field_required` Lex key. So the fix is to change ~L36 in `Processors/Security/Access/Policy/Template/Create.php`
```suggestion
```

core/lexicon/en/resource.inc.php
```diff
@@ -143,6 +143,7 @@
 $_lang['resource_syncsite_help'] = 'When enabled, this will make MODX empty the cache after you save the resource. This way your visitors will not see an older version of the resource.';
 $_lang['resource_template'] = 'Uses Template';
 $_lang['resource_template_help'] = 'The template in use by the resource.';
+$_lang['resource_type_help'] = 'The type of the resource.';
```
smg6511, line 146: Let's be a little more descriptive, _e.g._,
```suggestion
$_lang['resource_type_help'] = 'Indicates the type of content this Resource contains (Document, Weblink, Symlink, Static Resource, etc.) (Default: “Document”)';
```

core/lexicon/en/source.inc.php
```diff
@@ -28,6 +28,8 @@
 $_lang['source_err_nfs'] = 'No Media Source can be found with the id: [[+id]].';
 $_lang['source_err_ns'] = 'Please specify the Media Source.';
 $_lang['source_err_ns_name'] = 'Please specify a name for the Media Source.';
+$_lang['source_err_remove'] = 'An error occurred while trying to delete the Media Source!';
+$_lang['source_err_remove_default'] = 'The default Media Source cannot be deleted!';
```
smg6511, line 32: Although the key contains "_default" it is not necessarily the default that is protected. Suggest more specific message:
```suggestion
$_lang['source_err_remove_default'] = 'The built-in Filesystem Media Source is protected and cannot be deleted!';
```

core/lexicon/en/user.inc.php
```diff
@@ -56,6 +56,7 @@
 $_lang['role_warn_authority_locked'] = 'This Role’s authority can not be changed because this Role is currently assigned to one or more Access Control List entries.';
 $_lang['roles'] = 'Roles';
 $_lang['ugc_grid_title'] = 'User Group Access to Contexts';
+$_lang['ugc_mutate'] = 'User Group Access to Context';
```
smg6511, line 59:
```suggestion
```

This Lex isn't directly related to Users and should be added in the `default.inc.php` file instead of here (as these generalized titles could be used in a couple places). Additionally, we need two keys (action dependent) to be consistent with other windows. I'd suggest:
```
$_lang['usergroup_access_create'] = 'Add User Group Access';
$_lang['usergroup_access_update'] = 'Edit User Group Access';
```
Finally, the `ugc_mutate` keys need to be updated to match those suggested above in `modx.grid.access.context.js`.

core/src/Revolution/Processors/Browser/Directory/Rename.php
```diff
@@ -39,7 +39,7 @@ public function process()
         }
         $name = $this->sanitize($this->getProperty('name'));
         if (!strlen($name)) {
-            $this->addFieldError('name', $this->modx->lexicon('name_err_ns'));
+            $this->addFieldError('name', $this->modx->lexicon('file_name_err_ns'));
```
smg6511, line 42:
```suggestion
            $this->addFieldError('name', $this->modx->lexicon('file_name_err_empty'));
```

core/src/Revolution/Processors/Browser/File/Rename.php
```diff
@@ -39,7 +39,7 @@ public function process()
         }
         $name = $this->sanitize($this->getProperty('name'));
         if (empty($name)) {
-            $this->addFieldError('name', $this->modx->lexicon('name_err_ns'));
+            $this->addFieldError('name', $this->modx->lexicon('file_name_err_ns'));
```
smg6511, line 42:
```suggestion
            $this->addFieldError('name', $this->modx->lexicon('file_name_err_empty'));
```

core/src/Revolution/Processors/Element/GetListByClass.php
```diff
@@ -49,7 +49,7 @@ public function process()
     {
         $className = $this->getProperty('element_class');
         if (empty($className)) {
-            return $this->failure($this->modx->lexicon('element_class_ns'));
+            return $this->failure($this->modx->lexicon('class_err_ns'));
```
smg6511, line 52:
```suggestion
            return $this->failure($this->modx->lexicon('propertyset_element_class_err_empty'));
```

A couple things here:

1. There's already an entry in `propertyset.inc.php` that was meant for this (`propertysets_err_item_class_ns`; it's currently unused other than in the Lexes)
2. Let's change that entry with a little better key name (suggested above) and message:

```php
$_lang['propertyset_element_class_err_empty'] = 'A class name must be specified to attach this Element to the Property Set';
```

core/src/Revolution/Processors/Security/Access/UserGroup/AccessNamespace/Create.php
```diff
@@ -34,7 +34,7 @@ public function beforeSet()
     {
         $principal = $this->getProperty('principal');
         if ($principal === null) {
-            $this->addFieldError('principal', $this->modx->lexicon('usergroup_err_ns'));
+            $this->addFieldError('principal', $this->modx->lexicon('user_group_err_ns'));
```
smg6511, line 37: Not suggesting re-keying from "ns" to "empty" for now, since this is pointing to an already-existing key

core/lexicon/en/user.inc.php
smg6511, line 90:
```suggestion
$_lang['user_err_not_found'] = 'User not found.';
```

smg6511, line 210: Just adding a couple aliases for re-keyed items...
```suggestion
$_lang['user_group_user_create'] = $_lang['user_group_user_add'];
$_lang['user_err_nf'] = $_lang['user_err_not_found'];
$_lang['user_profile_err_nf'] = $_lang['user_profile_err_not_found'];
```

core/src/Revolution/Processors/Security/Profile/Get.php
```diff
@@ -53,7 +53,7 @@ public function initialize()
     {
         $this->user = $this->modx->user;
         if (!$this->user) {
-            return $this->modx->lexicon('user_err_not_found');
+            return $this->modx->lexicon('user_err_nf');
```
smg6511, line 56: Let's keep this as it was, but re-key as suggested in the Lex file:
```suggestion
            return $this->modx->lexicon('user_err_not_found');
```

Note that the alias I added in the Lex suggestion would cover any existing usage of the old key, but we could also update the key in the other two places it's found: the `security/profile.class.php` and `security/user/update.class`.php (below) controllers

core/src/Revolution/Processors/Security/Profile/Update.php
```diff
@@ -47,7 +47,7 @@ public function initialize()
     {
         $this->profile = $this->modx->user->getOne('Profile');
         if ($this->profile === null) {
-            return $this->modx->lexicon('user_profile_err_not_found');
+            return $this->modx->lexicon('user_profile_err_nf');
```
smg6511, line 50: Again, let's keep the newer key style, and re-key the Lex entry
```suggestion
            return $this->modx->lexicon('user_profile_err_not_found');
```

core/lexicon/en/user.inc.php
smg6511, line 181: Re-key to match the Profile/Update processor. While we're doing so, no need for the error type in the message
```suggestion
$_lang['user_profile_err_not_found'] = 'User profile not found.';
```

core/src/Revolution/Processors/Workspace/PackageNamespace/Update.php
```diff
@@ -34,7 +34,7 @@ public function beforeSave()
     {
         $name = $this->getProperty('name');
         if (empty($name)) {
-            $this->addFieldError('name', $this->modx->lexicon('namespace_err_ns_name'));
+            $this->addFieldError('name', $this->modx->lexicon($this->objectType . '_err_ns'));
```
smg6511, line 37: This is exclusive to one object, so no use for dynamic Lex key IMO
```suggestion
            $this->addFieldError('name', $this->modx->lexicon('namespace_err_ns'));
```

smg6511 (changes requested):
Thomas: Thanks for putting this together—I promise I'm not trying to be a PITA, but after going through this very carefully I do have a bunch of requests. Most are easily applied, but there are a couple which would need a change to files not currently in this PR.

Jako:
It makes no sense to add the changes you have requested, since they are no lexicon changes and change other parts of the core. I suggest the following handling: Create a cherry pick of all commits that are not part of the ones that have a requested change. Each lexicon change in my PR is done with a single commit, so cherry picking them should be quite easy.

PR #16962, opened 2026-05-31
Check for the real default media source value and don't use a fixed v…

Jako:
### What changed and why

Check for the real default media source value and don't use a fixed value.

### DIFF
--- a/core/src/Revolution/Processors/Source/Remove.php
+++ b/core/src/Revolution/Processors/Source/Remove.php
@@ -32,7 +32,7 @@ class Remove extends RemoveProcessor
      */
     public function beforeRemove()
     {
-        if ($this->object->get('id') === 1) {
+        if ($this->object->get('id') == $this->modx->getOption('default_media_source')) {
             return $this->modx->lexicon('source_err_remove_default');
         }
         return true;


Ibochkarev (approved):

smg6511:
@jako @Ibochkarev: You're misinterpreting what the purpose of this line is—It's to prevent the _built-in_ Filesystem Source (which should always be id 1) from being deleted. That source is not necessarily the default source in a multi media source setup. As such, no change is needed. Perhaps a short doc block above the line in question explaining this could be helpful though.

Jako:
@smg6511 After I have seen the installations of one client, I think a bit different about this: the media source 1 of several installations was changed into a 'Client Media Source' and pointed to something like `{base_path}user/whatever`. The filesystem media source was created as a second media source with a custom name. 

If we don't want that media source 1 to be deleted, we also have to make sure that it can't be edited, too. Since this would be a bit too restrictive to me, the check for the `default_media_source` setting is more suitable.

smg6511:
@jako - We actually already more clearly established the `Filesystem` media source as being protected a couple or so releases ago. While it cannot be removed or renamed, its settings can be edited. Although this change may eventually cause some confusion for the likely small percentage of folks who renamed this source (and subsequently try to rename it again), their setups would be otherwise unaffected. Remember that it's possible to have no source as the default media source, so there's not fool-proof protection by disallowing the deletion of the default source as you propose.

Jako:
Then at least the default_media_source system/context/user setting check should be added there to disallow these sources too.

smg6511:
@Jako You raise a good point and the UX can definitely be improved. I think the best way, rather than disallow the removal of a non-built-in source that's specified as a default, would be to mark the current default in the Sources grid (maybe via an icon or asterisk after the name) and _maybe_ add a variation on the current warning/confirmation dialog that addresses the implications of removing that source.

Jako:
The getlist processor can check for the default media sources in the system/context/user settings and return these values. The warning message can be extended with the value of the processor.

PR #16965, opened 2026-06-09
Remove E_STRICT error handling

Jako:
### What changed and why

E_STRICT has no meaning since PHP 7 (https://wiki.php.net/rfc/reclassify_e_strict). So it can be safely removed. MODX 2 and 3 are using PHP 7+

### DIFF
--- a/core/src/Revolution/Error/modErrorHandler.php
+++ b/core/src/Revolution/Error/modErrorHandler.php
@@ -110,13 +110,6 @@ class modErrorHandler
                 $this->modx->log(modX::LOG_LEVEL_WARN, $errmsg, '', '', $errfile, $errline);
                 break;
             default:
-                if (version_compare(PHP_VERSION, '8.4.0', '<') && $errno == E_STRICT) {
-                    $handled = true;
-                    $errmsg = 'E_STRICT information: ' . $errstr;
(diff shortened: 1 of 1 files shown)


smg6511 (approved):
Makes sense as what was reported in E_STRICT now gets reported in E_NOTICE and none of the code base makes use of this constant. Also consider removing the reference to it in `modX.php` (comment on line 710).

@opengeek - You may want to weigh in. Is there any reason to keep a ref to E_STRICT (maybe after E_NOTICE on line 87) as BC for external extenders/callers of this class? IOW:
```php
case E_NOTICE:
case E_STRICT:
    $handled = true;
    [...]
```
I'm guessing that's a non-issue though (?).

opengeek:
@smg6511 — I'm really not sure. The constant exists, whether it is triggered or not, in PHP versions we support, so I'm not convinced we should remove this code yet. But I'm not heavily invested either way.

Jako:
The constant is not used anymore since PHP 7. And using the constant throws a deprecation error since PHP 8.4. This deprecation error is avoided with a version check.

An error with that constant is never triggered in PHP 7+. So it makes no sense to check for the error with that constant in MODX.

PR #16963, opened 2026-06-07
JS code quality, modernization for tree, elements, and resources

smg6511:
### What changed and why

This is a formatting/style/syntax only set of changes to bring this area of the JS up to par with coding standards and our linting rules. Don't let the number of changed lines fool you—a majority of these are due to whitespace and comma placement changes.

I'm doing this ahead of working on a few issue-related PRs that touch these areas.

### How to test

Navigate around via the Trees and edit various Resources and Elements; all should work as expected and throw no errors in the console. Requires a `grunt/build` if you are reviewing with `compress_js` on.

### Related issue(s)/PR(s)

n/a

### Compatibility notes

n/a

### Breaking change assessment

None

### Test coverage

n/a

### Contributors

n/a

### AI tool use

None

PR #16967, opened 2026-06-16
JS code quality, modernization for core modx and component classes

smg6511:
### What changed and why

This is a formatting/style/syntax only set of changes to bring these JS classes up to par with coding standards and our linting rules. As with my previous merged update (#16963) a majority of lines changed are due to whitespace and comma placement changes.

I'm doing this ahead of working on a few issue-related PRs that touch these areas.

### How to test

Navigate around the manager and view and edit various items; all should work as expected and throw no errors in the console. Requires a `grunt/build` if you are reviewing with `compress_js` on.

### Related issue(s)/PR(s)

n/a

### Compatibility notes

n/a

Strictly-speaking, the update to the `getURLParameters` method (modx.js) could have BC implications due to:

1. Shallow array values are now collected in an array instead of being kept in string form, and
2. The URL API is now being used, which natively decodes string values. Previously, encoded strings were left as-is.

However, I'm highly confident that there are no issues core-wise. I reviewed the nature of the request variables used and tested the specific places where the encoding could come into play (namely the trees and the file browser).

On the array value handling, I don't anticipate any Extras issues as the method's current handling is broken anyway: a flaw in the logic results in only the last array value being retained. Because this bug exists already and we've not been getting complaints about it, it's reasonable to assume that either Extras are handling URL params on their own or they aren't making use of arrays (shallow ones specifically) in their request values. 

Note that, for complex arrays, the output before and after is the same, e.g.:
```javascript
?param4[0][item1]=Item1&param4[0][item2]=Item2
```
would yield:
```javascript
{
    'param4[0][item1]': 'Item1',
    'param4[0][item2]': 'Item2'
}
```

### Test coverage

While we don't have a mechanism to unit test our legacy ext js code, more sensitive changes like converting for-ins to forEach and re-organization/optimization of a couple methods were all manually tested to ensure correct scope and output was maintained.

To examine these changes more easily, take a look at the individual commits.

### Contributors

n/a

### AI tool use

None

### DIFF
--- a/manager/assets/modext/core/modx.component.js
+++ b/manager/assets/modext/core/modx.component.js
@@ -72,7 +72,7 @@ Ext.extend(MODx.Component, Ext.Component, {
             count = this.config.components.length,
             contentPanel = Ext.getCmp('modx-content')
         ;
-        for (let i = 0; i < count; i = i + 1) {
+        for (let i = 0; i < count; i++) {
             const component = MODx.load(this.config.components[i]);
             if (contentPanel) {
                 contentPanel.add(component);
@@ -145,9 +145,9 @@ Ext.extend(MODx.toolbar.ActionButtons, Ext.Toolbar, {
             const
                 el = args[i],
                 id = el.id || Ext.id(),
-                ex = ['-', '->', '<-', '', ' ']
+                exclude = ['-', '->', '<-', '', ' ']
             ;
-            if (ex.indexOf(el) != -1 || (el.xtype && el.xtype == 'switch')) {
+            if (exclude.indexOf(el) !== -1 || (el.xtype && el.xtype === 'switch')) {
                 MODx.toolbar.ActionButtons.superclass.add.call(this, el);
                 continue;
             }
@@ -155,7 +155,7 @@ Ext.extend(MODx.toolbar.ActionButtons, Ext.Toolbar, {
                 xtype: 'button',
                 cls: (el.icon ? 'x-btn-icon bmenu' : 'x-btn-text bmenu'),
                 scope: this,
-                disabled: el.checkDirty ? true : false,
+                disabled: el?.checkDirty,
                 listeners: {},
                 id: id
             });
@@ -184,7 +184,7 @@ Ext.extend(MODx.toolbar.ActionButtons, Ext.Toolbar, {
             }
 
             /* if checkDirty, disable until field change */
-            if (el.xtype == 'button') {
+            if (el.xtype === 'button') {
                 el.listeners.render = {
                     fn: function(btn) {
                         if (el.checkDirty && btn) {
@@ -250,12 +250,11 @@ Ext.extend(MODx.toolbar.ActionButtons, Ext.Toolbar, {
         }
 
         Ext.Msg.confirm('', itm.confirm, function(e) {
-            /* if the user is okay with the action */
             if (e === 'yes') {
                 if (callback === null) {
                     return true;
                 }
-                if (typeof(callback) === 'function') { /* if callback is a function, run it, and pass Button */
+                if (typeof callback === 'function') {
                     Ext.callback(callback, scope || this, [itm]);
                 } else {
                     window.location.href = callback;
@@ -266,7 +265,9 @@ Ext.extend(MODx.toolbar.ActionButtons, Ext.Toolbar, {
         return true;
     },
 
+    /** @todo Remove? No usage found in core js */
     reloadPage: function() {
+        // eslint-disable-next-line no-restricted-globals, no-self-assign
         location.href = location.href;
     },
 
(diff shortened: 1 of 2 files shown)


manager/assets/modext/core/modx.component.js
```diff
@@ -11,319 +10,374 @@ MODx.Component = function(config) {
...
+        if (!formPanel) {
+            return false;
+        }
+
+        Object.entries(listeners).forEach(([event, value]) => {
+            if (typeof typeof value === 'function') {
+                formPanel.on(event, value, this);
+            } else if (typeof value === 'object' && value.fn) {
+                formPanel.on(event, value.fn, value.scope || this);
+            }
```
Ibochkarev, line 100: Blocker: `typeof typeof value` is always `'string'`, so the function-listener branch never fires.

Looks like a typo for `typeof value === 'function'`. The object branch also lost the old `listeners[i] &&` guard — `typeof null === 'object'` would throw on `value.fn`.

Suggested shape:
```js
if (typeof value === 'function') {
    formPanel.on(event, value, this);
} else if (value && typeof value === 'object' && value.fn) {
    formPanel.on(event, value.fn, value.scope || this);
}
```

manager/assets/modext/core/modx.component.js
```diff
@@ -11,319 +10,374 @@ MODx.Component = function(config) {
...
+            } else {
                 el.handler = this.checkConfirm;
-            } else if (el.confirm && el.handler) {
-                el.handler = function() {
-                    Ext.Msg.confirm(_('warning'),el.confirm,function(e) {
-                      if (e === 'yes') { Ext.callback(el.handler,this); }
-                    },el.scope || this);
-                };
-            } else if (el.handler) {} else { el.handler = this.handleClick; }
+            }
```
Ibochkarev: Blocker: `hasHandler` via `typeof el.handler === 'object'` treats `null` as a real handler (`typeof null === 'object'`).

That changes the old decision tree. On 3.x today:
- `handler === null && menu === null` → `checkConfirm`
- `confirm && handler` → wrap
- else if no handler → `handleClick`

With the new tree, `handler: null` (common Ext default) skips both `checkConfirm` and `handleClick`, and a confirm wrap can end up calling `null`.

I would keep the original predicates (or explicitly exclude null: `handler != null && …`) so the outcomes match what was there before the reorder. Sorting frequent cases first is fine only if the state machine stays equivalent.

Side note while you are in here: the confirm wrap still overwrites `el.handler` and then `Ext.callback(el.handler)` — that was already recursive/self-calling before. Capturing `const originalHandler = el.handler` before the wrap would fix it for real.

manager/assets/modext/core/modx.js
```diff
@@ -103,428 +111,453 @@ Ext.extend(MODx,Ext.Component,{
...
+            if (key.indexOf('[]') !== -1) {
+                paramsObject[key.replace('[]', '')] = params.getAll(key);
+            } else {
+                paramsObject[key] = value;
             }
-        }
-        return arg;
-    }
+        });
+        return paramsObject;
```
Ibochkarev, line 200: This is a real API change, not formatting. `URLSearchParams` is a nice upgrade, but:

1. Keys like `foo[]` used to land as a scalar under `'foo[]'`. Now they become an array under `'foo'`. Anything reading `MODx.request['foo[]']` (or expecting a string) breaks.
2. Values get decoded (`+` / `%20` → space). Usually better, still a behavior shift vs the old split.
3. `key.replace('[]', '')` only strips the first match — `endsWith('[]')` + `slice(0, -2)` is safer.

Since `MODx.request = this.getURLParameters()` runs on init, this hits the whole manager. Either document it in Breaking / Compatibility, or keep scalar-compatible keys in this PR and add array capture in a follow-up.

Ibochkarev (changes requested):
Hey Jim — thanks for the cleanup pass ahead of the issue PRs. Most of the whitespace / declaration work looks fine, and dropping the IE8 branch in `status` is the right call for 3.x.

A couple of the later commits are not style-only though, and two of them look like real regressions. Calling those out below.

Also: `getURLParameters` now uses `URLSearchParams` and remaps `foo[]` into arrays under `foo`. That is a behavior/API change for anything reading `MODx.request`. Worth updating the Breaking assessment (or keeping scalar keys if you want this PR to stay non-breaking).

Happy to re-review once the two blockers are sorted.

manager/assets/modext/core/modx.js
smg6511: Thanks for pointing out the above. I'll put a note in the BC section. In response to your notes:
- 1 & 3: The array handling is actually broken in the current method (detail in BC note), which makes me comfortable with making the behavior change. Specifically on 3, we don't support multiple [] pairs (in the form of key[][]) in any given key anyway.
- 2: I made sure this wasn't a breaker across the core; although we already explicitly decode strings, the result is the same even if those strings have already been decoded natively via the URL API.

Ibochkarev: Hi!

Great, if everything's ok, can you resolve the threads?

manager/assets/modext/core/modx.component.js
smg6511: Good catch on handling null. I'll handle that in follow-up commit later today.

On the confirm wrap/callback: After looking at this more closely and tracking where this could have been triggered, there's only one place (context/view) and that appears to be a legacy/unused view (which is why the problematic re-writing of `el.handler` and then calling itself instead of the original handler never came up as a bug). I'd suggest we leave this block alone and handle removal of it and other related unused code in another PR.

manager/assets/modext/core/modx.component.js
smg6511: Typo and null handling fixed in recent commit.

manager/assets/modext/core/modx.component.js
smg6511: Made adjustments for null and added note about legacy code block.

Ibochkarev (approved):

PR #16980, opened 2026-07-30
Work around LiteSpeed update issues on HTTP2/3

Mark-H:
> This patch is unfinished. It occasionally causes requests in the manager to stop processing. Sharing for collaborative purposes.

### What changed and why

- The early session_write_close() makes sure the feeds aren't keeping a lock on the session
- The modx.js implements queuing and max connection throthling on the manager requests

LiteSpeed seems to be rejecting the bulk of requests when using HTTP2/3, likely related to its "security enhancements". 

On HTTP1.1, browsers already throttle requests to typically 6 per domain, so this is not a problem. HTTP2/3 however open as many connections as needed, and MODX does load a whole bunch of stuff over AJAX that seems to be tripping up LiteSpeed's recent changes. 

### How to test

On LiteSpeed, make sure the manager works.

### Related issue(s)/PR(s)

Community discussions:
- https://community.modx.com/t/http-1-2-3-and-modx-followup/9211
- https://community.modx.com/t/resource-tree-disappeared-suddenly/9213

### Compatibility notes

Specifically recent LiteSpeed versions. 

None.

### Test coverage

None.

### Contributors

None. 

### AI tool use

Problem analysed and patch developed with help from Cursor.


pbowyer:
@Mark-H is a commercial LiteSpeed licence required to reproduce the problem and test the fix?

opengeek:
> @Mark-H is a commercial LiteSpeed licence required to reproduce the problem and test the fix?

I'm wondering the same, as I have been unable to reproduce on OpenLiteSpeed so far.

Mark-H:
I was giving access to a site experiencing the problem by dejaya, not sure about the full configuration. Does seem likely from what you and others have said.

Mark-H:
Rebased to change the WIP commit message to have an actual description, plus two more additions to this PR:

## [rebuild] Ensuring backslashes in connector requests are escaped (for LiteSpeed)
This builds upon the report and suggested fix in https://community.modx.com/t/resource-tree-disappeared-suddenly/9213/16, which explains a hosts' investigation into the LiteSpeed issue. They pinpoint it to backslashes which trip up LiteSpeed's tweaked security handling. The most notable instance of that is the resource tree, which includes the resource type in the `type` URL parameter: /connectors/index.php?action=Resource/GetNodes&id=web_0&type=MODX\Revolution\modContext

**ExtJS actually does automatically encode backslashes already**, however, the tree and various other places were manually drafting endpoint URLs instead of going through the extjs `params` or utility methods - leading to unescaped backslashes.

This fix adds a new MODx.getConnectorUrl utility which uses Ext.urlAppend to generate proper parameterised URLs. That utility is then implemented into various places that manually craft URLs to ensure consistent behavior. Not all of these would necessarily fail in litespeed today as the chance they include special characters can be quite low; but being consistent is useful in case future tweaks are needed.

## Improve queue handling, ensuring aborted requests decrease the active count so queue continues

When requests would get aborted (one such instance that's easy to reproduce is the system settings grid: repeatedly filter namespace or area, that seems to trigger duplicate requests that get aborted) they would not be taken out of the queue or call drain() to trigger next queued requests. After 6 of those, requests would halt and spin indefinitely.

lanceC99:
I would just like to confirm that the problem still exists with OpenLiteSpeed 1.9.1, ModX v 3.2.4-pl
And the plugin (https://community.modx.com/t/resource-tree-disappeared-suddenly/9213/18)
The plugin doesn't catch this because it only overrides XHR and fetch()
The failure is happening on the main Manager navigation URL;
manager/?a=resource/update&id=67&reload=...&class_key=MODX\Revolution\modDocument&context_key=web

PR #16982, opened 2026-08-02
fix(error): return HTTP 500 for uncaught frontend errors

### LINKED ISSUE #16932: PHP errors return HTTP 200 instead of 500 (blank white page)
For discussion - I am intentionally not proposing code at this stage.

### Summary

When a PHP error occurs during resource rendering -- whether in a snippet, plugin, template variable, or anywhere else in the rendering pipeline -- the visitor receives a blank white page with an **HTTP 200 OK** status code. MODX does not register any exception handler or shutdown function to catch these errors and set an appropriate HTTP 500 response.

### Steps to reproduce

The simplest reproduction is a snippet with a type error:

1. Create a snippet with code that triggers a TypeError, e.g.:

```php
<?php
function testTyped(int $x): string {
    return $x;
}
return testTyped("not an int");
```

2. Call the snippet on a published resource: `[[testSnippet]]`
3. Visit the page in a browser
4. Observe: blank white page, but the HTTP response status is `200 OK`

But this is not limited to snippets. Any uncaught exception or fatal error anywhere in the rendering chain will produce the same result: a TypeError in a TV render, a database connection failure during resource loading, a ParseError in a static element file, a missing class in a plugin, etc.

### Observed behavior

The visitor sees a blank page (or partial output if the error occurs mid-render) with an **HTTP 200 OK** status code. The consequences:

- **Silent failures**: A single bad snippet or plugin can take down a page (or the entire site if it runs in a sitewide plugin/event) with no HTTP-layer indication that anything is wrong
- **Cache poisoning**: Reverse proxies (Varnish, Cloudflare, nginx proxy_cache) that cache HTTP 200 responses will serve the blank page to all visitors until the cache expires (this is how I discovered the problem)
- **Monitoring blind spots**: Uptime monitors that check HTTP status codes will not detect the outage
- **SEO damage**: Search engines may deindex pages that return blank HTTP 200 responses
- **Difficult debugging**: Without a HTTP 500 status, developers must check PHP error logs to discover the problem -- there is no indication at the HTTP level. Depending on `display_errors` / `log_errors` PHP settings, the error may or may not appear in PHP's error log, and may or may not appear in the MODX error log either.

### Expected behavior

When an unrecoverable PHP error occurs during page rendering, the response should have an **HTTP 500** status code (or at minimum, not 200).

### Why this happens

Nothing in the frontend rendering pipeline catches exceptions or detects fatal errors after the fact. The call chain from `index.php` through `modRequest::handleRequest()`, `modResponse::outputContent()`, `modResource::prepare()`, and `modParser::processElementTags()` down to individual element execution is entirely unprotected.

Specifically:

**1. No exception handler is registered**

`modX::_initErrorHandler()` calls `set_error_handler()` to register `modErrorHandler`, but never calls `set_exception_handler()`. In PHP 7+, TypeErrors, ValueErrors, ParseErrors, and other `\Error` subclasses are not routed through `set_error_handler` -- they are thrown exceptions that bypass it entirely. Without a registered exception handler, PHP handles them with its default behavior: log or display the error and terminate with whatever HTTP status was implicit (200).

**2. No shutdown function checks for fatal errors**

MODX does not register a `register_shutdown_function()` that calls `error_get_last()` to detect fatal errors and set a HTTP 500 status. The two shutdown functions that exist serve other purposes:
- `modResponse::outputContent()` registers one for cache post-processing (`_postProcess`)
- `modX::deprecated()` registers one for saving deprecation records

Neither checks for fatal errors or sets HTTP status codes.

**3. `modErrorHandler` only handles recoverable errors**

The registered `modErrorHandler::handleError()` handles warnings, notices, deprecations, and `E_USER_ERROR` by logging them and returning `true` (suppressed). But `set_error_handler` callbacks are never invoked for fatal errors (`E_ERROR`, `E_PARSE`, `E_CORE_ERROR`, `E_COMPILE_ERROR`) or for thrown `\Error`/`\Exception` objects. These are the exact error types that cause the white screen.

**4. No try/catch anywhere in the rendering chain**

The problem is not limited to element execution. The entire rendering pipeline lacks exception handling:

- **`index.php`** calls `$modx->handleRequest()` with no try/catch
- **`modRequest::handleRequest()`** calls `invokeEvent()`, `findResource()`, `getResource()`, and `prepareResponse()` with no try/catch
- **`modResponse::outputContent()`** calls `resource->prepare()`, `invokeEvent('OnWebPagePrerender')`, and processes content with no try/catch
- **`modResource::prepare()`** calls `process()` and `parseContent()` with no try/catch
- **`modParser::processElementTags()`** recursively processes all tags with no try/catch
- **`modScript::process()`** executes snippet/plugin code via a bare `include` with no try/catch
- **`modX::invokeEvent()`** calls `plugin->process()` with no try/catch
- **`modTemplateVar::getRender()`** includes TV render files via `include` with no try/catch

A thrown exception or fatal error at any of these points will terminate PHP execution with no opportunity to set an HTTP error status.

The only part of the rendering pipeline that has a try/catch is `modOutputFilter::filter()` for output modifiers -- everything else is unprotected.

**5. `sendError()` exists but is never called automatically**

`modX::sendError()` properly clears output buffers and sets an HTTP error header, but it is only called explicitly for known error conditions (site unavailable, configuration missing, etc.). There is no mechanism that calls it when an unexpected PHP error occurs during rendering.

### How I've worked around it

On a project where this caused downtime (a reverse proxy cached a blank 200 response), I wrote a class that I register from my extra's namespace bootstrap file (`bootstrap.php`).

The bootstrap calls `FrontendRequestFailsafe::register($modx)`, which -- for frontend requests only (skipping CLI, manager, and connector requests for now) -- registers:

1. `set_exception_handler()` to catch uncaught exceptions (TypeError, ValueError, etc.)
2. `register_shutdown_function()` with `error_get_last()` to catch fatal errors (E_ERROR, E_PARSE, E_COMPILE_ERROR, etc.)
3. A custom `modErrorHandler` subclass that intercepts `E_USER_ERROR` (which `set_error_handler` does receive, unlike the fatal types above)

Each handler clears output buffers, sends `HTTP 500` with no-cache headers (`Cache-Control: no-store`, `X-Accel-Expires: 0`, etc.), logs the error, and exits. On a healthy request, none of this fires and the response is unaffected.

This works as a per-project workaround, but I think it should be built into MODX itself so that all sites get correct HTTP error status codes by default.

### Environment

- MODX 3.x (all versions, including current 3.1.x / 3.2-dev)
- Any PHP (the problem worsened with PHP 7's Error hierarchy, but existed before for fatal errors)
- Any web server

Ibochkarev:
### What changed and why

When a snippet, plugin, TV, or other element throws during web rendering, MODX often returned a blank page with HTTP 200. Proxies cached that response. Uptime checks saw 200 and missed the failure.

This PR adds `modUncaughtErrorHandler`, registered from `_initErrorHandler()`:

- `set_exception_handler()` for uncaught `\Throwable` (`TypeError`, `ParseError`, etc.)
- `register_shutdown_function()` with `error_get_last()` for `E_ERROR`, `E_PARSE`, `E_CORE_ERROR`, `E_COMPILE_ERROR`
- `E_USER_ERROR` in `modErrorHandler` routes to `LOG_LEVEL_FATAL` on web requests when `send_http_500_on_error` is enabled

Each path calls the existing `sendError('fatal')` flow. `fatal.include.php` now sends no-cache headers so reverse proxies do not store 500 responses.

Opt out with system setting `send_http_500_on_error` (default `true`).

### How to test

1. Create a snippet that throws a `TypeError` (pass a string to a typed `int` parameter).
2. Put `[[snippetName]]` on a published resource and open the page.
3. Expect HTTP 500 and the generic fatal error page, not a blank 200.
4. Check response headers for `Cache-Control: no-store`.
5. Set `send_http_500_on_error` to `false` and confirm the handler no longer forces 500.

Unit tests:

```bash
core/vendor/bin/phpunit -c _build/test/phpunit.xml --group modUncaughtErrorHandler
core/vendor/bin/phpunit -c _build/test/phpunit.xml --group modErrorHandler
```

CI (phpcs, PHPUnit on MySQL 8.1–8.5) is green on this branch.

### Related issue(s)/PR(s)

Resolves #16932

### Compatibility notes

- Web requests only. CLI skips handler registration; `sendError()` behavior in CLI is unchanged.
- Applies to frontend, manager, and connector contexts that initialize `modX`.
- PHP 7+ (`\Throwable` hierarchy). No new Composer dependencies.

**Behavior change:** `E_USER_ERROR` on web requests now terminates with HTTP 500 when `send_http_500_on_error` is enabled (default). Code that logged a user error and kept running should use `E_USER_WARNING` or disable the setting.

No public API signature changes. Patch-safe for sites that do not rely on `E_USER_ERROR` continuing after the handler runs.

### Test coverage

- Added `_build/test/Tests/Model/Error/modUncaughtErrorHandlerTest.php`: fatal type detection, error formatting, exception logging, shutdown behavior for non-fatal errors, registration (web only).
- Extended `_build/test/Tests/Model/Error/modErrorHandlerTest.php`: `E_USER_ERROR` → `LOG_LEVEL_FATAL` when enabled (web only; skipped under CLI PHPUnit).

### Contributors

Thanks to @pbowyer for #16932 and the documented workaround that shaped this approach.

### AI tool use

Claude (Cursor) helped draft the implementation, tests, and PR text from the issue analysis and codebase review.

### DIFF
--- a/_build/test/Tests/Model/Error/modErrorHandlerTest.php
+++ b/_build/test/Tests/Model/Error/modErrorHandlerTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
--- a/_build/test/Tests/Model/Error/modErrorHandlerTest.php
+++ b/_build/test/Tests/Model/Error/modErrorHandlerTest.php
@@ -8,11 +8,14 @@
  * files found in the top-level directory of this distribution.
  *
  * @package modx-test
-*/
+ */
 namespace MODX\Revolution\Tests\Model\Error;
 
-
+use MODX\Revolution\Error\modErrorHandler;
+use MODX\Revolution\Error\modUncaughtErrorHandler;
+use MODX\Revolution\modX;
 use MODX\Revolution\MODxTestCase;
+use RuntimeException;
 
 /**
  * Tests related to the modErrorHandler class.
@@ -23,8 +26,58 @@ use MODX\Revolution\MODxTestCase;
  * @group Error
  * @group modErrorHandler
  */
-class modErrorHandlerTest extends MODxTestCase {
-    public function testExample() {
-        $this->assertTrue(true);
+class modErrorHandlerTest extends MODxTestCase
+{
+    /** @var modErrorHandler */
+    public $handler;
+
+    /**
+     * @before
+     */
+    public function setUpFixtures()
+    {
+        parent::setUpFixtures();
+        $this->handler = new modErrorHandler($this->modx);
+    }
+
+    /**
+     * @after
+     */
+    public function tearDownHandlerState()
+    {
+        modUncaughtErrorHandler::resetForTesting();
+        restore_error_handler();
+    }
+
+    public function testHandleErrorUserErrorTriggersFatalResponseWhenEnabled()
+    {
+        if (defined('XPDO_CLI_MODE') && XPDO_CLI_MODE) {
+            $this->markTestSkipped('HTTP fatal handling is disabled in CLI mode.');
+        }
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Fatal log invoked');
+
+        $modx = $this->getMockBuilder(modX::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['getOption', 'log'])
+            ->getMock();
+        $modx->method('getOption')->willReturnMap([
+            ['send_http_500_on_error', null, true, true],
+        ]);
+        $modx->expects($this->once())
+            ->method('log')
+            ->with(
+                modX::LOG_LEVEL_FATAL,
+                $this->stringContains('User error: simulated user error'),
+                '',
+                '',
+                '/tmp/test.php',
+                10
+            )
+            ->willThrowException(new RuntimeException('Fatal log invoked'));
+
+        $handler = new modErrorHandler($modx);
+        $handler->handleError(E_USER_ERROR, 'simulated user error', '/tmp/test.php', 10);
     }
 }
--- a/_build/test/Tests/Model/Error/modUncaughtErrorHandlerTest.php
+++ b/_build/test/Tests/Model/Error/modUncaughtErrorHandlerTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the MODX Revolution package.
  *
--- a/_build/test/Tests/Model/Error/modUncaughtErrorHandlerTest.php
+++ b/_build/test/Tests/Model/Error/modUncaughtErrorHandlerTest.php
@@ -1,0 +1,137 @@
+<?php
+/*
+ * This file is part of the MODX Revolution package.
+ *
+ * Copyright (c) MODX, LLC
+ *
+ * For complete copyright and license information, see the COPYRIGHT and LICENSE
+ * files found in the top-level directory of this distribution.
+ *
+ * @package modx-test
+ */
+namespace MODX\Revolution\Tests\Model\Error;
+
+use MODX\Revolution\Error\modUncaughtErrorHandler;
+use MODX\Revolution\modX;
+use MODX\Revolution\MODxTestCase;
+
+/**
+ * Tests related to the modUncaughtErrorHandler class.
+ *
+ * @package modx-test
+ * @subpackage modx
+ * @group Model
+ * @group Error
+ * @group modUncaughtErrorHandler
+ */
+class modUncaughtErrorHandlerTest extends MODxTestCase
+{
+    /**
+     * @after
+     */
+    public function tearDownHandlerState()
+    {
+        modUncaughtErrorHandler::resetForTesting();
+        restore_exception_handler();
+        restore_error_handler();
+    }
+
+    /**
+     * @param int  $type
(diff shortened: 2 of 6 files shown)


pbowyer:
Thanks for this PR @Ibochkarev.

I haven't tested so this is a question where I don't know the answer: if I switch the [logger implementation to Monolog](https://github.com/modxcms/revolution/pull/16931) and register its exception handler, which one wins here?

Ibochkarev:
> Thanks for this PR @Ibochkarev.
> 
> I haven't tested so this is a question where I don't know the answer: if I switch the [logger implementation to Monolog](https://github.com/modxcms/revolution/pull/16931) and register its exception handler, which one wins here?

Good question. #16931 only adds `setLogger()` — core does not register Monolog's `ErrorHandler`.

This PR registers `modUncaughtErrorHandler` during `initialize()`, before `OnMODXInit`. If you call `Monolog\ErrorHandler::register()` in a plugin after init, Monolog runs first and chains to our handler, so you get Monolog logging plus `sendError('fatal')` / HTTP 500. If you register Monolog before `initialize()`, our handler replaces it and Monolog's exception handler never runs.

`setLogger()` alone does not conflict — it only changes log output.

PR #16983, opened 2026-08-03
Make Resource content regions independently hideable via Form Customization

### LINKED ISSUE #15711: Form customization can't hide just the content field
### Summary
When modx-resource-content is disabled the whole section goes away. So it is not possible to hide just the content field and move template variables to modx-resource-content.

### Step to reproduce
Disable modx-resource-content in form customization and move template variables into modx-resource-content. The section is invisible.

### Expected behavior
There should be a way just to hide the content field and move TVs into that section.

### Environment
MODX 2.8.2

smg6511:
### What changed and why

Re-factored the component structure for the content-related regions/field to allow them to be individually shown/hidden. Note that the change in how these are defined also makes the above and below regions renderable for _all_ Resource types instead of just the Document type.

### How to test

Create two TVs and a Form Customization set to move those TVs into the `modx-resource-content-above` and `modx-resource-content-below` Regions. Toggle the visibility of those and the main `modx-resource-content` field to verify all work as expected.

### Related issue(s)/PR(s)

Resolves #15711

### Compatibility notes

n/a

n/a

### Test coverage

n/a

### Contributors

Issue authored by @Jako

### AI tool use

None

### DIFF
--- a/manager/assets/modext/widgets/resource/modx.panel.resource.js
+++ b/manager/assets/modext/widgets/resource/modx.panel.resource.js
@@ -714,8 +714,7 @@ Ext.extend(MODx.panel.Resource, MODx.FormPanel, {
                     value: config.record.introtext || ''
                 }]
             }]
-
-        }, this.getContentField(config)];
+        }, ...this.getContentField(config)];
     },
 
     getMainRightFields: function(config = {}) {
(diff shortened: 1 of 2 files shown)


Ibochkarev:
Sibling regions look like the right fix for #15711.

**Blocking:** `...this.getContentField(config)` breaks Static / SymLink / WebLink. Those overrides still return a single config object, and spreading a non-iterable throws when the left column builds.

Suggested fix in `getMainLeftFields`:

```js
const contentFields = this.getContentField(config);
// ...
}, ...(Ext.isArray(contentFields) ? contentFields : [contentFields])];
```

Or make Static / SymLink / WebLink return `[field]` as well.

Please also:
1. Keep `layout: 'form'`, `autoHeight: true`, and `hideMode: 'offsets'` on the `modx-resource-content` wrapper.
2. Keep deprecated `modx-content-above` / `modx-content-below` until 3.3, or note the early removal under Compatibility notes.

Happy to re-review after that.

smg6511:
@Ibochkarev  - Good catch Ivan, I'd completely forgotten the extending document types. Will fix in a bit... Note that this will target 3.3, which is why I removed the deprecated elements.

smg6511:
> 1. Keep `layout: 'form'`, `autoHeight: true`, and `hideMode: 'offsets'` on the `modx-resource-content` wrapper.

Note that these are established in the upper defaults and/or don't have any effect, so they're not needed on this wrapper. For example, if layout: form were not in the defaults, the content field label wouldn't show up.

Ibochkarev (approved):
👍

rthrash (approved):
I created two TVs, assigned them to the relevant areas, and toggled each area per the instructions. All show as expected without FC rules. No Manager error log messages during testing. Before:

Screenshot 2026-08-14 at 11 55 39 AM

(Note: before image added after changing the `above-tv` label for another PR test. Functionality works and the label change was not caused by the PR.)

### `modx-resource-content` unchecked visibile
Image
 
 
### `tv-above` and `tv-below` unchecked visible
Image
 
 
### `tv-above`-only unchecked visible
Image

PR #16987, opened 2026-08-10
Add CI check against re-bundled phpmailer/Smarty/AWS (MAB-01)

Ibochkarev:
### What changed and why

MAB-01 already put PHPMailer, Smarty, and the AWS SDK behind Composer in `core/vendor/`. The old trees under `core/model/` were cleaned up in 3.0. Nothing stops someone from copying them back in a later PR.

This adds a small hygiene gate:

- `_build/check_no_bundled_libs.sh` — fails on the historical paths and on library signature files (`PHPMailer.php`, `Smarty.class.php`, `Aws/Sdk.php`, …) under `core/` outside `vendor` / `cache` / `packages`
- a dedicated job in `.github/workflows/ci.yml` (no PHP matrix, no Composer)
- `composer check-bundled-libs` for the same check locally

It looks at paths and file fingerprints, not the words “phpmailer” / “smarty” / “aws”, so wrappers like `modPHPMailer` and `MODX\Revolution\Smarty` do not trip it.

### How to test

```bash
bash _build/check_no_bundled_libs.sh
# or
composer check-bundled-libs
```

Expect `OK`. To see it fail: `mkdir -p core/model/aws` then run again; remove the dir after.

### Related issue(s)/PR(s)

- Residual risk close-out for MAB-01 (Composer dependency management)
- Sibling status paperwork: modxcms/mab-recommendations#26

### Compatibility notes

CI-only. No runtime behaviour change. Needs bash + `find` (already true on GitHub-hosted runners and typical Mac/Linux dev machines).

No public API or default behaviour change. A PR that re-introduces a forbidden tree will fail CI until that copy is removed.

### Test coverage

No PHPUnit case — the gate is the test. Manual negative checks: forbidden path and planted `Aws/Sdk.php` both exit 1; clean tree exits 0.

### Contributors

Builds on the MAB-01 Composer work already in 3.x.

### AI tool use

Cursor helped draft the script and CI job. I chose path fingerprints over keyword search, ran pass/fail smoke locally, and kept the change to one script + one workflow job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,16 @@ on:
   workflow_dispatch:
 
 jobs:
+  bundled-libs-check:
+    name: No bundled phpmailer/Smarty/AWS
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Reject copies outside core/vendor
+        run: bash _build/check_no_bundled_libs.sh
+
   mysql-check:
     name: MySQL Check
     runs-on: ubuntu-latest
--- a/_build/check_no_bundled_libs.sh
+++ b/_build/check_no_bundled_libs.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+#
+# MAB-01 hygiene: fail if phpmailer, Smarty, or AWS SDK are re-bundled under
+# core/ outside Composer-managed core/vendor/.
+#
+# Checks tree fingerprints (historical 3.0.0 cleanup paths + library signature
+# files). Does not search for the words "phpmailer"/"smarty"/"aws" — those hit
+# MODX wrappers, lexicons, and docs.
+#
+set -euo pipefail
+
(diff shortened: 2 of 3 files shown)


mkschell:
I don't see this as materially benefitting the CMS in a way that makes it worth a reviewer's time and focus, or the small additional complexity in the CI layer. The risk that this PR guards against is not well discussed or explored as a problem.

Mark-H:
Yeah I'm not sure why this is needed. Are we at risk of people randomly adding manual dependencies? Is there context to this that I'm missing?

Ibochkarev:
Thanks for looking at this.

@mark-h — no, I’m not seeing people copy these libs back into core/ by hand. There’s no incident or near-miss behind the PR.

I was treating leftover risk from [MAB-01](https://github.com/modxcms/mab-recommendations/blob/master/01-replace-3rd-party-libraries-with-composer-dependencies.md) (Composer for AWS / phpmailer / smarty). That recommendation is from 2016. Composer already owns those packages in 3.x, so a CI guard for re-bundling doesn’t buy us much now.

@mkschell — agreed on cost/benefit. Extra CI and review time for a problem we aren’t hitting isn’t worth it.

I’ll close this.

PR #16988, opened 2026-08-11
Fix undefined array key "desc" in element property lexicon translation

Ibochkarev:
### What changed and why

`modElement::get('properties')` and `modPropertySet::get('properties')` read `$property['desc']` without checking the key. PHP 8 logs `Undefined array key "desc"` when a property definition has no description field.

The fix resolves the label with `$property['desc'] ?? $property['description'] ?? ''` and calls the lexicon only when the result is non-empty. Both classes share the same translation loop, so both get the same guard.

### How to test

```bash
php -l core/src/Revolution/modElement.php
php -l core/src/Revolution/modPropertySet.php
XDEBUG_MODE=coverage core/vendor/bin/phpunit --filter testGetPropertiesWithoutDescKey -c ./_build/test/phpunit.xml
```

Exit 0 on PHP 8.4 locally.

To see the original warning: open an element in the manager whose stored properties omit `desc` (imported or legacy definitions). Before this change, the log pointed at `modElement.php:165`.

### Related issue(s)/PR(s)

Found in the manager error log. Part of the broader PHP 8 cleanup tracked in #16018.

### Compatibility notes

Applies on PHP 8+. When `desc` is present, behavior is unchanged. When both `desc` and `description` are missing, `desc_trans` is an empty string instead of a warning.

No API or signature changes. Safe for patch releases.

### Test coverage

`testGetPropertiesWithoutDescKey` with two data-provider cases: missing `desc` (no warnings, empty `desc_trans`) and `description` fallback only.

### Contributors

N/A

### AI tool use

Cursor agent wrote the patch, regression test, and review notes.

### DIFF
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -162,7 +162,7 @@ class modElement extends modAccessibleSimpleObject
                     }
                     $this->xpdo->lexicon->load($property['lexicon']);
                 }
-                $desc = $property['desc'] ?? $property['description'] ?? '';
+                $desc = $property['desc'] ?? '';
                 $property['desc_trans'] = $desc !== '' ? $this->xpdo->lexicon($desc) : '';
                 $property['area'] = !empty($property['area']) ? $property['area'] : '';
                 $property['area_trans'] = !empty($property['area']) ? $this->xpdo->lexicon($property['area']) : '';
--- a/core/src/Revolution/modElement.php
+++ b/core/src/Revolution/modElement.php
@@ -162,7 +162,8 @@ class modElement extends modAccessibleSimpleObject
                     }
                     $this->xpdo->lexicon->load($property['lexicon']);
                 }
-                $property['desc_trans'] = $this->xpdo->lexicon($property['desc']);
+                $desc = $property['desc'] ?? $property['description'] ?? '';
+                $property['desc_trans'] = $desc !== '' ? $this->xpdo->lexicon($desc) : '';
                 $property['area'] = !empty($property['area']) ? $property['area'] : '';
                 $property['area_trans'] = !empty($property['area']) ? $this->xpdo->lexicon($property['area']) : '';
 
--- a/core/src/Revolution/modPropertySet.php
+++ b/core/src/Revolution/modPropertySet.php
@@ -75,7 +75,8 @@ class modPropertySet extends xPDOSimpleObject
                     }
                     $this->xpdo->lexicon->load($property['lexicon']);
                 }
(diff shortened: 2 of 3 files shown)


Mark-H (approved):

smg6511:
@Ibochkarev @Mark-H - Guys, just one little thing re the unit test: I don't think it's needed or really appropriate to basically be checking the _correctness_ of how something is coded (in this case that there's a null guard to avoid php errors/warnings) via a UT (at least not one that's to be included in the code base's suite). IMO, that part of this PR can be dropped.

Ibochkarev:
> @Ibochkarev @Mark-H - Guys, just one little thing re the unit test: I don't think it's needed or really appropriate to basically be checking the _correctness_ of how something is coded (in this case that there's a null guard to avoid php errors/warnings) via a UT (at least not one that's to be included in the code base's suite). IMO, that part of this PR can be dropped.

Hi!

I've removed the changes from the test.

mkschell:
For me, that was a bit of a grey area. The test wasn't directly testing the input/output contract of the method, but I saw it as testing for an unwanted side effect (as opposed to testing the correctness of the code). I see it's been modified and approved, and that's ok with me. IMO we _should_ be working to reduce side effects of functions (can you tell I'm into FP?) but I'm also ok with not explicitly testing for them (or their absence) if that's the general sentiment.

core/src/Revolution/modElement.php
```diff
@@ -162,7 +162,8 @@ public function get($k, $format = null, $formatTemplate = null)
                     }
                     $this->xpdo->lexicon->load($property['lexicon']);
                 }
-                $property['desc_trans'] = $this->xpdo->lexicon($property['desc']);
+                $desc = $property['desc'] ?? $property['description'] ?? '';
```
smg6511:
```suggestion
                $desc = $property['desc'] ?? '';
```

core/src/Revolution/modPropertySet.php
```diff
@@ -75,7 +75,8 @@ public function get($k, $format = null, $formatTemplate = null)
                     }
                     $this->xpdo->lexicon->load($property['lexicon']);
                 }
-                $property['desc_trans'] = $this->xpdo->lexicon($property['desc']);
+                $desc = $property['desc'] ?? $property['description'] ?? '';
```
smg6511:
```suggestion
                $desc = $property['desc'] ?? '';
```

smg6511 (changes requested):
Maybe this is a nit, but `'description'` is not an expected key on the properties array; it's only used in defining a non-default prop set object and isn't accessed here.

smg6511 (approved):

PR #17002, opened 2026-08-17
fix(manager): keep settings grid usable after delete

### LINKED ISSUE #14280: Quick editing & deleting of context settings from the grid doesn't works after deleting some other setting
### Summary
Quick editing & deleting of context settings from the grid doesn't works after deleting any settings

### Step to reproduce

1. Create 2 or more some settings for context and fill them
2. Remove one of them
3. And try to edit some setting via double clicking or delete it
4. Gotcha! We get the error "Setting not found"

Gif picture to demonstration

ux

### Environment
MODX 2.6.5-pl, JS errors is empty

I tested this bug in 2.7.0-pl and it still remains

Ibochkarev:
### What changed and why

After you delete a context setting from the grid, the next inline edit or delete often returned “Setting not found”. `MODx.grid.Grid.remove()` wrote `action` and the deleted `key` straight onto shared `config.saveParams`. Later autosave (`saveRecord` → `UpdateFromGrid`) merged those polluted params into another row and sent the wrong key.

`remove()` now clones `saveParams` before mutating. Settings grids also return `false` from `afterRemoveRow` after `refreshFilterOptions()` (which already reloads the store), so `removeActiveRow` does not race with a second `store.remove(selection)`.

### How to test

1. Manager → Contexts → edit `web` → Context Settings.
2. Create three settings (e.g. `test_a`, `test_b`, `test_c`) with any values.
3. Delete `test_a` from the grid menu.
4. Double-click `test_b` value, change it, blur/save — should succeed (no “Setting not found”).
5. Delete `test_c` — should succeed without a page reload.

```
node --check manager/assets/modext/widgets/core/modx.grid.js
node --check manager/assets/modext/widgets/core/modx.grid.settings.js
# exit 0
```

Browser smoke on revolution.test: after remove, `config.saveParams` stays `{context_key:"web"}`; UpdateFromGrid for the next row returns success.

### Related issue(s)/PR(s)

Resolves #14280

### Compatibility notes

Manager ExtJS grids that use `config.saveParams` with `remove()` (context / system / user / usergroup settings). No PHP or connector API change.

No public API change. Behavior fix only: remove no longer mutates the shared `saveParams` object.

### Test coverage

No ExtJS unit harness in-repo. Verified with `node --check` and a manager smoke that deletes one context setting then UpdateFromGrid on another.

### Contributors

@GulomovCreative reported the original bug.

### AI tool use

Cursor (Composer) traced the saveParams leak, applied the clone + afterRemoveRow veto, and drafted this PR body. Browser checks on revolution.test confirmed the polluted params before the fix and clean autosave after.

### DIFF
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -1581,13 +1581,8 @@ Ext.extend(MODx.grid.GridBase, Ext.grid.EditorGridPanel, {
     },
 
     removeActiveRow: function(record) {
-        if (!this.fireEvent('afterRemoveRow', record)) {
-            return;
-        }
-        const selection = this.getSelectionModel().getSelected();
-        if (selection) {
-            this.getStore().remove(selection);
-        }
+        this.getStore().remove(record);
+        this.fireEvent('afterRemoveRow', record);
     },
 
     refresh: function() {
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -1554,7 +1554,9 @@ Ext.extend(MODx.grid.GridBase, Ext.grid.EditorGridPanel, {
         }
         const
             { record } = this.menu,
-            saveParams = this.config.saveParams || {},
+            // Clone — mutating config.saveParams leaks remove action/key into later
+            // autosave UpdateFromGrid payloads (#14280).
+            saveParams = Ext.apply({}, this.config.saveParams || {}),
             primaryKey = this.config.primaryKey || 'id'
         ;
         text = text || 'confirm_remove';
@@ -1580,8 +1582,11 @@ Ext.extend(MODx.grid.GridBase, Ext.grid.EditorGridPanel, {
     },
 
     removeActiveRow: function(record) {
-        if (this.fireEvent('afterRemoveRow', record)) {
-            const selection = this.getSelectionModel().getSelected();
+        if (!this.fireEvent('afterRemoveRow', record)) {
+            return;
+        }
+        const selection = this.getSelectionModel().getSelected();
+        if (selection) {
             this.getStore().remove(selection);
         }
     },
--- a/manager/assets/modext/widgets/core/modx.grid.js
+++ b/manager/assets/modext/widgets/core/modx.grid.js
@@ -1554,8 +1554,7 @@ Ext.extend(MODx.grid.GridBase, Ext.grid.EditorGridPanel, {
         }
         const
             { record } = this.menu,
-            // Clone — mutating config.saveParams leaks remove action/key into later
-            // autosave UpdateFromGrid payloads (#14280).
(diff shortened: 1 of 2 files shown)


manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -1554,7 +1554,9 @@ Ext.extend(MODx.grid.GridBase, Ext.grid.EditorGridPanel, {
         }
         const
             { record } = this.menu,
-            saveParams = this.config.saveParams || {},
+            // Clone — mutating config.saveParams leaks remove action/key into later
+            // autosave UpdateFromGrid payloads (#14280).
```
smg6511: Suggesting a little more straight-forward note here
```suggestion
            // Clone config.saveParams here to avoid modification of the original config object, which would break subsequent calls to updateFromGrid
```

manager/assets/modext/widgets/core/modx.grid.js
```diff
@@ -1580,8 +1582,11 @@ Ext.extend(MODx.grid.GridBase, Ext.grid.EditorGridPanel, {
...
     removeActiveRow: function(record) {
-        if (this.fireEvent('afterRemoveRow', record)) {
-            const selection = this.getSelectionModel().getSelected();
+        if (!this.fireEvent('afterRemoveRow', record)) {
+            return;
+        }
+        const selection = this.getSelectionModel().getSelected();
+        if (selection) {
             this.getStore().remove(selection);
         }
```
smg6511: The original method was oddly constructed with the `afterRemoveRow` firing before the row is actually removed. This reflects the pattern used elsewhere:
```suggestion
        this.getStore().remove(record);
        this.fireEvent('afterRemoveRow', record);
```
Note that the original method's retrieval of the selection isn't needed because the relevant data is already passed in the `record`.

manager/assets/modext/widgets/core/modx.grid.settings.js
```diff
@@ -300,7 +300,11 @@ MODx.grid.SettingsGrid = function(config = {}) {
             }
         },
         afterRemoveRow: function() {
+            // Reload filters + grid. Return false so removeActiveRow does not also
+            // store.remove(selection) — that races the refresh and can drop another
+            // row, so later edit/delete hits setting_err_nf (#14280).
```
smg6511: I didn't observe this being a problem before, but the correction to the `afterRemoveRow` method in the base class should take care of this.
```suggestion
```

manager/assets/modext/widgets/core/modx.grid.settings.js
```diff
@@ -300,7 +300,11 @@ MODx.grid.SettingsGrid = function(config = {}) {
...
             this.refreshFilterOptions(gridFilterData);
+            return false;
```
smg6511:
```suggestion
```

smg6511 (changes requested):
My testing verifies that the cloning of the saveParams fixes the issue. The other changes weren't really needed from my observations, although I did suggest an update to the `afterRemoveRow` method, as the original implementation didn't make sense.

smg6511 (approved):
Looks good!

Issue #17014, opened 2026-09-03
[Bug]: OnWebLogin: $modx->getUser() still returns previous/anonymous user


Ibochkarev:
### Summary

During `OnWebLogin`, session contexts for the new user are already set, but `$modx->getUser()` still returns the request-cached user (often anonymous / id 0, or the previous visitor). Manager login refreshes `$modx->user` before `OnManagerLogin`; web login does not.

### Steps to Reproduce

1. Create a plugin on `OnWebLogin` that runs:
   ```php
   /** @var modX $modx */
   /** @var modUser $user */
   $fromGetUser = $modx->getUser();
   $modx->log(modX::LOG_LEVEL_ERROR, 'event user id=' . $user->get('id')
       . '; getUser id=' . $fromGetUser->get('id')
       . '; username=' . $fromGetUser->get('username'));
   ```
2. Log in on the front end as a real user (Login snippet / processor, context `web`).
3. Check the error log.

### Expected Behavior

After a successful web login, `$modx->getUser()` (or at least `$modx->getUser($loginContext, true)`) returns the same user that was just authenticated, consistent with `OnManagerLogin` / manager login.

### Actual / Observed Behavior

`$user` from the event is the authenticated user, but `$modx->getUser()` still returns the cached anonymous (or previous) user.

In `Processors/Security/Login.php` → `afterLogin()`:

1. `addSessionContexts()` runs for every login context.
2. For `mgr` only: `$modx->user = null` then `$modx->getUser('mgr', true)` before the event.
3. `fireAfterLoginEvent()` fires `OnWebLogin` / `OnManagerLogin`.

Web never clears/reloads `$modx->user`, so `getUser()` short-circuits on the old in-memory object.

Docs: [modxorg/Docs#428](https://github.com/modxorg/Docs/issues/428), clarification PR [modxorg/Docs#586](https://github.com/modxorg/Docs/pull/586) (Mark-H suggested fixing this in core).

Workaround for plugins today: use the `$user` event argument, or force a reload (e.g. `$modx->user = null; $modx->getUser($loginContext, true);`).

#### MODX Version

3.x (`Processors/Security/Login.php`); same `afterLogin()` pattern on 2.x

#### PHP Version

N/A (logic bug)

#### Database Type & Version

N/A

#### Server / Web Server / Browser (if relevant)

N/A

### Logs / Error Output

Typical plugin log after web login as a normal user:

```
event user id=2; getUser id=0; username=(anonymous)
```

[...]

PR #17022, opened 2026-09-09
fix: refresh $modx->user before OnWebLogout

### LINKED ISSUE #17015: [Bug]: OnWebLogout: $modx->getUser() still returns the logged-in user
### Summary

During `OnWebLogout` / `OnManagerLogout`, session contexts have already been removed, but `$modx->user` is never cleared or reloaded. `$modx->getUser()` therefore still returns the user who just logged out. Docs describe the event as firing after logout, so plugins that call `getUser()` (or fire other code that does) behave as if the user is still authenticated.

Related login-side mismatch: #17014.

### Steps to Reproduce

1. Create a plugin on `OnWebLogout` that runs:
   ```php
   /** @var modX $modx */
   /** @var modUser $user */
   $fromGetUser = $modx->getUser();
   $modx->log(modX::LOG_LEVEL_ERROR, 'event userid=' . $user->get('id')
       . '; getUser id=' . $fromGetUser->get('id')
       . '; isAuthenticated(web)=' . ($modx->user->isAuthenticated('web') ? '1' : '0'));
   ```
2. Log in on the front end, then log out (Logout snippet / processor, context `web`).
3. Check the error log.

### Expected Behavior

After session contexts are removed, `$modx->getUser()` should reflect a logged-out state for that context (anonymous / id 0, or a forced reload), so plugins during `OnWebLogout` do not keep treating the request as authenticated via `getUser()`.

### Actual / Observed Behavior

`removeSessionContexts()` runs, then `fireAfterLogoutEvent()` fires `OnWebLogout` / `OnManagerLogout` while `$modx->user` is still the previous authenticated object. `getUser()` short-circuits on that cached object.

In `Processors/Security/Logout.php` → `process()`:

1. `fireBeforeLogoutEvent()`
2. `removeSessionContexts()`
3. `fireAfterLogoutEvent()` — no `$modx->user = null` / `getUser(..., true)` in between

Docs report: [modxorg/Docs#427](https://github.com/modxorg/Docs/issues/427).

Workaround for plugins today: use the event `$user` / `userid` for who logged out, and do not rely on `$modx->getUser()` for “current session user” during this event. Or force a reload after clearing `$modx->user`.

#### MODX Version

3.x (`Processors/Security/Logout.php`); same order on 2.x

#### PHP Version

N/A (logic bug)

#### Database Type & Version

N/A

#### Server / Web Server / Browser (if relevant)

N/A

### Logs / Error Output

Typical plugin log after logout as a normal user:

```
event userid=2; getUser id=2; isAuthenticated(web)=0
```

(Session context is gone, but `getUser()` still returns the same user object.)

Ibochkarev:
### What changed and why

After logout, session contexts were already removed, but `$modx->user` stayed the previous authenticated object. During `OnWebLogout` / `OnManagerLogout`, `$modx->getUser()` therefore still returned the user who just left.

`Logout::process()` now snapshots that user onto the processor, clears `$modx->user`, and reloads via `getUser($loginContext)` before the after-logout event. The event still receives `$user` / `userid` / `username` for who logged out.

`$forceLoadSettings` is not used on reload, so user settings are not rewritten into the cleared context session.

### How to test

Automated:

```bash
composer run-script phpunit -- --filter LogoutTest --no-coverage
```

Exit code: `0` (1 test, 10 assertions).

Red-green: without the refresh step, the test fails with `getUser id` equal to the logged-out user id.

Manual:

1. Create a plugin on `OnWebLogout` that logs `$user->get('id')` and `$modx->getUser()->get('id')`.
2. Log in on the front end, then log out (`web` context).
3. Confirm event `$user` is the departed user and `getUser()` is anonymous (`id` 0) when no other context session remains.

### Related issue(s)/PR(s)

Closes #17015

Related login-side mismatch: #17014 / #17021

Docs: modxorg/Docs#427

### Compatibility notes

Universal for 3.x logout through `Processors\Security\Logout`.

If the visitor still has an active `mgr` session after leaving `web`, `getUser('web')` may still resolve to that mgr user via the existing `getUser()` fallback. That path is unchanged and outside this fix.

Observable change for after-logout plugins that previously read a stale `$modx->getUser()` (still the departed user). Event `$user` / `userid` remain the source for who logged out. Public method signatures are unchanged. Safe for a patch release for consumers already using the event arguments.

### Test coverage

Added `_build/test/Tests/Processors/Security/LogoutTest.php` for the web path: event identity preserved, `getUser()` anonymous, `isAuthenticated('web')` false.

Not covered here: dedicated mgr logout case; concurrent web+mgr session after partial logout (pre-existing `getUser()` fallback).

### Contributors

@Ibochkarev

### AI tool use

Cursor agent (issue-to-pr pipeline) drafted the fix, regression test, and this PR description from #17015.


ShevArtV:
Tested PR #17022 locally with `modx-pr-tester`.

Environment:
- MODX PR HEAD: `be840f1f7dc0fc286e809bfda2d95f2a22ed1f47`
- Local URL: `http://127.0.0.1:8081/`
- PHPUnit command: `composer run-script phpunit -- --filter LogoutTest --no-coverage --do-not-cache-result`

Results:
- `LogoutTest` passed: 1 test, 10 assertions.
- Web-only logout: `OnWebLogout` still receives the departing web user (`event_user_id=4`, `event_username=logout-web-fixture`), while after logout `getUser()` resolves to anonymous user id `0` and `isAuthenticated('web')` is `false`.
- Web logout with an active Manager session: `OnWebLogout` still receives the departing web user, web authentication is cleared, and `getUser()` resolves to the active Manager user, which matches the separate mgr session.
- Manager login/logout smoke check passed.

I did not run a red/green comparison against the parent commit because the parent does not contain `LogoutTest.php`; the working PR checkout was not modified.

Conclusion: within the tested scope, this PR behaves as intended and looks good to accept.

PR #17024, opened 2026-09-09
feat: Update Flysystem to 3.x

matdave:
### What changed and why

Update Flysystem to version 3. Currently MODX Is on Flysystem 2.x which has many broken features that have been left unmaintained. A critical broken feature is the inability to rename, move or copy objects in S3. 

### How to test

Apply the new composer.json and test file operations

### Compatibility notes

Flysystem 3.x is stated as being backwards compatible with 2.x, with the primary difference being it requires PHP 8, which MODX also now requires. 

This is likely not patch level safe, as it is a full library upgrade. 

### Related Issues

https://github.com/thephpleague/flysystem/issues/1759

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,9 +51,9 @@
     "require": {
         "php": ">=8.1",
         "xpdo/xpdo": "^3.2",
-        "league/flysystem": "^2.0",
-        "league/flysystem-aws-s3-v3": "^2.0",
-        "league/flysystem-ftp": "^2.0",
+        "league/flysystem": "^3.0",
+        "league/flysystem-aws-s3-v3": "^3.0",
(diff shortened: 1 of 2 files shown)


Ibochkarev:
Tested with modx-pr-tester.

Browser: Chrome browser / macOS (Docker PHP 8.2).
Result: Pass.
Packages: flysystem 3.36.0, aws-s3-v3 3.35.3, ftp 3.35.3.

Local Filesystem media source: create/list/rename/move/remove OK (19/19 smoke).
Manager media browser loads.

Did not test live S3/FTP remotes.

Ibochkarev (approved):

